Issue #2935, opened 2024-05-16
Ansible prepare step of imported plan does not work


mkluson:
When you use import plan feature [1] and the imported plan has one of prepare steps implemented via Ansible, the execution fails, because it can not find the playbook. E.g.:
`ERROR! the playbook: /var/ARTIFACTS/git-masternbv9uido/ansible/setup_to_run_leapp.yml could not be found`

[1] https://tmt.readthedocs.io/en/stable/spec/plans.html#import-plans

mkluson:
I found out, that commenting out  `_ansible_playbook_path` at https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/__init__.py#L1474 fix this issue and does not break use case when the imported plan is run directly.

PR #2989, opened 2024-06-05
Use ansible playbook from worktree

mkluson:
During `tmt run` execution there are two copies of the playbook
available:
- One in the original repository with tmt plan
- Second in the `worktree` under /var/tmp/tmt...

`_ansible_playbook_path` uses the second one instead the first one now.

Resolves: https://github.com/teemtee/tmt/issues/2935

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -956,11 +956,9 @@ class Guest(tmt.utils.Common):
         self.debug(f"Applying playbook '{playbook}' on guest '{self.primary_address}'.")
         # FIXME: cast() - https://github.com/teemtee/tmt/issues/1372
         parent = cast(Provision, self.parent)
-        assert parent.plan.my_run is not None  # narrow type
-        assert parent.plan.my_run.tree is not None  # narrow type
-        assert parent.plan.my_run.tree.root is not None  # narrow type
+        assert parent.plan.fmf_root is not None  # narrow type
         # Playbook paths should be relative to the metadata tree root
-        playbook = parent.plan.my_run.tree.root / playbook.unrooted()
+        playbook = parent.plan.fmf_root / playbook.unrooted()
         self.debug(f"Playbook full path: '{playbook}'", level=2)
         return playbook
 


mkluson:
Hi there,
I would like to kick off a discussion about fix for https://github.com/teemtee/tmt/issues/2935. This PR propose one possible solution, but I am not sure if it is the best one.

tmt/steps/provision/__init__.py
```diff
@@ -1471,8 +1471,6 @@ def _run_ansible(
         :param silent: if set, logging of steps taken by this function would be
             reduced.
         """
-        playbook = self._ansible_playbook_path(playbook)
```
happz, line 1474: IIRC, the idea was that `_ansible_playbook_path()` should be fixed rather than removed. It's there to make sure the playbook lives in directories tmt controls, and the side-effect is it produces an absolute path to the playbook which you are unhappy about. Let's fix that, but not by throwing away the check like this. We'll investigate & find something more complicated and uglier :)

mkluson:
I confirm, solution proposed by @happz works for me too.

psss:
@mkluson, @happz, thanks for looking into this and sorry for the slow response.

> @psss @lukaszachy any idea whether `my_run.tree.root` would be the only possible choice? I can imagine the `plan.fmf_root` would not be pushed to guest or something, but ansible tests passed even with the `PROVISION_HOW=virtual` set.

Using `plan.fmf_root` sounds like a reasonable choice here. I think `my_run.tree.root` was added here long time before the plan import feature has been implemented. At that time all plans could be only in the single `fmf` tree from which `tmt run` is called. Imported plans are stored under the `import` folder of the run directory, for example like this:

    /var/tmp/tmt/run-024/import/plans/minimal

The `fmf_root`s of the imported plans are thus stored under the `/var/tmp/tmt/run-024/import` directory. These folders do not have to be copied to guest as the `ansible-playbook` command is always executed locally. I don't see anything which could be broken by this approach. +1 from me to go this way.

mkluson:
I updated the code with the proposal from @happz and consider the PR ready to review.

tmt/steps/provision/__init__.py
mkluson: Your recommendation has been implemented.

psss (approved):
Now looks good, thanks for looking into this!

tmt/steps/provision/__init__.py
psss: Thanks!

janhavlin (approved):

happz:
The failure seems to be unrelated, I've been able to re-run the test and it works. Merging.

PR #2991, opened 2024-06-05
Allow `tmt try IP@connect`

lukaszachy:
Pull Request Checklist

* [x] implement the feature
* [x] adjust plugin docstring

### DIFF
--- a/tmt/cli.py
+++ b/tmt/cli.py
@@ -1705,11 +1705,8 @@ def try_command(context: Context, image_and_how: str, **kwargs: Any) -> None:
             options = {"image": image_and_how}
 
         # For 'connect' rename 'image' to 'guest'
-        try:
-            if options['how'] == 'connect':
-                options['guest'] = options.pop('image')
-        except KeyError:
-            pass
+        if options.get('how') == 'connect':
+            options['guest'] = options.pop('image')
 
         tmt.steps.provision.Provision.store_cli_invocation(context=None, options=options)
 
--- a/tmt/cli.py
+++ b/tmt/cli.py
@@ -1681,6 +1681,11 @@ def try_command(context: Context, image_and_how: str, **kwargs: Any) -> None:
     \b
         tmt try fedora@container
         tmt try centos-stream-9@virtual
+
+    Or connect to the running guest by specifying FQDN or IP address:
+
+    \b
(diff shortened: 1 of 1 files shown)


lukaszachy:
@psss how should be this documented? Having to change all `IMAGE@HOW` into `IMAGE_OR_IP@HOW` doesn't seem a good idea to me as it will just be confusing.

KwisatzHaderach (approved):

janhavlin (approved):

tmt/cli.py
```diff
@@ -1666,6 +1666,11 @@ def try_command(context: Context, image_and_how: str, **kwargs: Any) -> None:
     \b
         tmt try fedora@container
         tmt try centos-stream-9@virtual
+
+    Or connect to the running guest by specifying FQDN or IP address:
```
psss, line 1670:
```suggestion
    Or connect to the running guest by specifying hostname or IP address:
```

Let's use rather `hostname` than the abbreviation?

psss (changes requested):
Looks good, I'd just avoid the abbreviation.

therazix (approved):

tmt/cli.py
```diff
@@ -1684,6 +1689,13 @@ def try_command(context: Context, image_and_how: str, **kwargs: Any) -> None:
         else:
             options = {"image": image_and_how}
 
+        # For 'connect' rename 'image' to 'guest'
+        try:
+            if options['how'] == 'connect':
+                options['guest'] = options.pop('image')
+        except KeyError:
+            pass
```
thrix, line 1697: with `get` we can avoid the try / except handling?

```suggestion
            if options.get('how') == 'connect':
                options['guest'] = options.pop('image')
```

martinhoyer (approved):
lgtm, works as expected.

tmt/cli.py
lukaszachy: it's not the same, is it? https://serverfault.com/questions/269838/what-is-the-difference-between-a-hostname-and-a-fully-qualified-domain-name

martinhoyer: Agreed to keep FQDN on hacking session.

happz:
Unrelated failures, merging.

PR #2978, opened 2024-06-02
Add primitives for loading and saving step results [Step results #3]

happz:
Pull Request Checklist

Depends on https://github.com/teemtee/tmt/pull/2977

* [x] implement the feature

PR #2977, opened 2024-06-02
Introduce "phase result" for prepare/finish plugins to return [Step results #2]

happz:
No plugin returns it yet, not any of the results is saved, the patch changes return value types and adds the structure itself.

Depends on https://github.com/teemtee/tmt/pull/2963

Pull Request Checklist

* [x] implement the feature

PR #2996, opened 2024-06-07
Save results before raising exception

pholica:
This is essential for storing at least partial results and making report work if there was some issue with the system (e.g. became unreachable) during the execution.

Pull Request Checklist

* [x] implement the feature
* [x] include a release note


tmt/steps/execute/__init__.py
```diff
@@ -946,6 +939,13 @@ def go(self, force: bool = False) -> None:
         self.status('done')
         self.save()
 
+        if failed_tasks:
+            # TODO: needs a better message...
+            raise tmt.utils.GeneralError(
+                'execute step failed',
+                causes=[outcome.exc for outcome in failed_tasks if outcome.exc is not None]
+                )
```
happz, line 947: Makes sense to save collected results before raising an exception, `summary()` seems good as well, but I'm less sure about changing the status to `done`.

@psss @lukaszachy what is the expected status of `execute` when an error is raised while it runs? I wouldn't expect it to be `done`, and if nothing else, currently it stays in `todo`.

I'd maybe extract just the saving of results, call that, then raise an exception before getting to `status()` and `save()` (maybe https://github.com/teemtee/tmt/pull/2978/ before this PR?)

pholica: I've tried leaving it in todo, but in that case, then report removes all the files.

happz: Neither seems right...

lukaszachy: I have no idea ;) there is just 'todo' and 'done'. Time for 3rd one?

happz: > Well, I don't have the secret knowledge, but the done says:
> 
> > done ... the final result of the step stored to workdir (we are done)
> 
> And that corresponds with the point in the execution, nothing else will be done with the data anymore...
> 
> I'm sorry to be pushy, but if there's no real issue, could we merge this? Or is there anything I need to do?

I'd argue discussing possible undesirable side effects and settling the question of whether `done` is correct or not *is* a real issue: before your patch, in case of an exception, the state would be `todo`, and your patch changes it to `done`. That's a change that may affect other workflows.

To me, it looks like a change we shouldn't make, but since I am not an owner of any secret knowledge either, I asked others to chime in during the last hacking session on Tuesday. If things won't move, I'll ping the community again, but I for one do not feel comfortable merging it as-is.


> Another reason is that, if e.g. `exit-first: True` is used, it also ends in `done` state if there was a failure but the execution didn't "finish" completely, right?

Good point, although I'd say there's one important difference: `exit-first` applies to *failed* tests, not exceptions. A failed test raises no exception in tmt, it's a "normal", recoverable state - store the result and move on to another test. On the other hand, an exception is raised in the other scenario, things went south so far that it was impossible to recover and continue with another test or exit gracefully.

pholica: > > Well, I don't have the secret knowledge, but the done says:
> > > done ... the final result of the step stored to workdir (we are done)
> > 
> > 
> > And that corresponds with the point in the execution, nothing else will be done with the data anymore...
> > I'm sorry to be pushy, but if there's no real issue, could we merge this? Or is there anything I need to do?
> 
> I'd argue discussing possible undesirable side effects and settling the question of whether `done` is correct or not _is_ a real issue: before your patch, in case of an exception, the state would be `todo`, and your patch changes it to `done`. That's a change that may affect other workflows.

I see, and I'm sorry about my tone. The thing is that what I'm trying to also say is that with the current design (according to my observations), the `todo` state is actually the thing that causes the reporting to be broken. I'm just little bit afraid that this PR could become blocked by possible modification of multiple places by introducing a new state and implementing different kind of behaviour for that while we'd be suffering in situation when things go wrong losing all the data that was collected during the execution...

I believe we can agree that nothing will change in case of situation when there are no `failed_tasks`. And if there are some, the execution already ended anyway at this point, but I understand the cautiousness and it's better be safe than sorry. I just wish we could have a fork of TMT with this fix we could use in the TF in the meantime until it's verified it would work for everyone... 

> 
> To me, it looks like a change we shouldn't make, but since I am not an owner of any secret knowledge either, I asked others to chime in during the last hacking session on Tuesday. If things won't move, I'll ping the community again, but I for one do not feel comfortable merging it as-is.

Thank you for that and let me apologize again.
> 
> > Another reason is that, if e.g. `exit-first: True` is used, it also ends in `done` state if there was a failure but the execution didn't "finish" completely, right?
> 
> Good point, although I'd say there's one important difference: `exit-first` applies to _failed_ tests, not exceptions. A failed test raises no exception in tmt, it's a "normal", recoverable state - store the result and move on to another test. On the other hand, an exception is raised in the other scenario, things went south so far that it was impossible to recover and continue with another test or exit gracefully.

Yes, and there's also a situation of timeout that behaves (in the context of `exit-first`) the same way as failure, but that's another topic, timeout is far from recoverable as there's no recover code executed, that's why I used that argument - sorry for not mentioning the timeout upfront. But yeah, this behaviour is not directly related to this PR.

pholica: Hello, I'd like to most politely remind about this as it's now almost 3 weeks since the last communication. This is blocking us in improving our efficiency as we cannot store log files and have to print everything as in our tests the failures are often in situations when the system is no longer accessible.

happz: Yeah, there seems to be not much of a reaction :/ I'll try to play with that status today between meetings, hopefully, I'll get a better idea of how to avoid `todo` removing things - I still believe `todo` is more fitting than `done`, given we're talking error, not fail, so if we can keep it and avoid trashing your data, it would be nice. Facing the lack of guidance, the decision is left for taking...

pholica: Thank you

happz: Ok, trying it now. My time management sucks.

happz: @pholica "I've tried leaving it in todo, but in that case, then report removes all the files." - can you elaborate on what files are removed? Is "report" the `report` step or just a generic "report" emitted by the `execute` step?

Also, a reproducer would be handy, if you have a simple one. I started by injecting exceptions into the `execute` code, and I went with keeping `todo` status in case of an exception, I strongly feel that `done` is not the right one in this case. Now I'm looking into what files get removed and by which part of the code if the status remains `todo`, but I think I need more pointers.

pholica: Ok, here are the steps:

1. Use this repo for TMT plan: https://github.com/pholica/tmt_testrepo/tree/main
2. Get some RHEL/Fedora VM and create a snapshot of it (you'll make it unbootable later)
3. Run: `tmt run -a provision --how connect --guest ${VM_IP}` - tmt "fails" on `Failed to pull workdir from the guest.` and `ssh: connect to host ${VM_IP} port 22: No route to host`
4. Check log files:
```
$ find /var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/ -name \*.txt
/var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/01_log_plan_data-1/output.txt
/var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/02_log_test_data-2/data/test_log.txt
/var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/02_log_test_data-2/output.txt
/var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/03_do_damage_and_reboot-3/data/ls_boot.txt
/var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/03_do_damage_and_reboot-3/output.txt
```
5. Run `tmt run --last report --how html` - now you get `summary: no results found`
6. Check log files again:
```
$ find /var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/ -name \*.txt
find: ‘/var/tmp/tmt/run-056/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/’: No such file or directory
```
7. Check files after tmt report is run:
```
$ tree /var/tmp/tmt/run-056/
/var/tmp/tmt/run-056/
├── log.txt
├── plans
│   └── example
│       └── broken_reboot
│           ├── data
│           │   └── variables.env
│           ├── discover
│           │   ├── step.yaml
│           │   └── tests.yaml
│           ├── execute
│           │   ├── results.yaml
│           │   └── step.yaml
│           ├── finish
│           │   └── step.yaml
│           ├── prepare
│           │   └── step.yaml
│           ├── provision
│           │   ├── guests.yaml
│           │   └── step.yaml
│           ├── report
│           │   ├── default-0
│           │   │   └── index.html
│           │   └── step.yaml
│           └── tree
│               ├── LICENSE
│               └── plans
│                   └── example.fmf
└── run.yaml

14 directories, 15 files
```

pholica: Any luck reproducing the issue? Is any more information needed?

psss:
This is an important item for Upgrades & Conversions, marking as a `must`.

tmt/steps/execute/__init__.py
psss: Thank you for the detailed instructions. Here's a simplified reproducer:

    tmt run --id /tmp/lost --until execute
    sed -i s/done/todo/ /tmp/lost/*/execute/step.yaml
    tmt run --id /tmp/lost report

I believe the main problem is actually in the step workdir cleanup. The result removal part should be fixed by 9d67ce86. Could you please try it?

martinhoyer: Thanks!
I've reduced the `tmt-reboot` timeout to 1 and used the default provisioning VM, so with just `tmt run -advv`.
```
 progress:                                                   
        fail: Failed to pull workdir from the guest. This usually means that login as 'root' to the guest does not work.
        fail: Command 'rpm -q --whatprovides rsync || dnf install -y  rsync' returned 255.
    finish
    
        guest: stopped
libvirt: QEMU Driver error : Requested operation is not valid: Refusing to undefine while domain managed save image exists

plan failed

Cause number 1:

    Requested operation is not valid: Refusing to undefine while domain managed save image exists
```
with no results in report.

While I don't have that much knowledge of the code, I'd agree with @happz to move the save part above the `if failed task`, incl. the merge results bit, as I'd expect 'done' status to be set only after 'save'.
I've tried this and the report is more or less identical to this PR patch, just without the printed summary of course (`summary: 3 tests executed`).

Checking https://github.com/teemtee/tmt/pull/2977 and https://github.com/teemtee/tmt/pull/2978 now.

psss: In fa2c32addc4313c930fab074a797dfd58a2ac806 I've updated the fix to change status to `done` only if there were no errors. So this original comment from @happz should be addressed.

psss:
Except for missing `save()` there were two issues which caused the problem:

* workdir cleanup of disabled steps fixed in 972febbe0a519e8e0cb29f27629fdf239e9ea54b
* missing plan data fixed in b615f25815b86d77d38d04b584ed84eb1ea5b045

Test coverage in fa2c32addc4313c930fab074a797dfd58a2ac806 suggests it's now working as expected.

martinhoyer (commented):
Looks good to me and works. I don't know enough this code to 'approve' stuff though.

thrix (approved):
LGTM, no comments from me

tmt/steps/execute/__init__.py
happz: Yes indeed, resolving.

happz (approved):

Issue #2997, opened 2024-06-07
Test executed twice instead of executing the next test


pholica:
I've executed following test plan using TMT: https://github.com/pholica/tmt_testrepo/blob/6087865da91737f8bf9c9ff2d2f19da1311f6cac/plans/example.fmf
from repo: https://github.com/pholica/tmt_testrepo/tree/6087865da91737f8bf9c9ff2d2f19da1311f6cac

During the execution, I hit a situation when the `01_log_plan_data` was actually executed twice, second time as if it was `02_log_test_data`. See the result of `03_list_root` and output from `02_log_test_data`.

I'm not hitting this behavior consistently, but rather randomly but quite frequently.

```
$ tmt run -a provision --how connect --guest 192.168.122.111
/var/tmp/tmt/run-033

/plans/example/broken_reboot
    discover
        how: fmf
        directory: /home/pholica/git/pholica/tmt_testrepo
        summary: 3 tests selected
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: connect
        primary address: 192.168.122.111
        topology address: 192.168.122.111
        multihost name: default-0
        arch: x86_64
        distro: Red Hat Enterprise Linux 9.4 (Plow)
    
        summary: 1 guest provisioned
    prepare
        queued push task #1: push to default-0
        
        push task #1: push to default-0
    
        queued prepare task #1: requires on default-0
        
        prepare task #1: requires on default-0
        how: install
        summary: Install required packages
        name: requires
        where: default-0
        package: /usr/bin/flock
    
        queued pull task #1: pull from default-0
        
        pull task #1: pull from default-0
    
        summary: 1 preparation applied
    execute
        queued execute task #1: default-0 on default-0
        
        execute task #1: default-0 on default-0
        how: tmt
        progress:                                            
    
        summary: 3 tests executed
    report
        how: display
        summary: 2 tests passed and 1 test failed
    finish
    
        summary: 0 tasks completed

total: 2 tests passed and 1 test failed
```
```
$ cat /var/tmp/tmt/run-033/plans/example/broken_reboot/execute/data/guest/default-0/plans/example/tests/02_log_test_data-2/output.txt 
TMT_PLAN_DATA='/var/tmp/tmt/run-033/plans/example/broken_reboot/data'
plan_data_foobar
```
```
$ tmt --version
tmt version: 1.33.0
$ rpm -q tmt
tmt-1.33.0-1.fc39.noarch
```

[...]

PR #2998, opened 2024-06-07
Make test shell wrapper filename even more unique

happz:
To prevent accidental reuse of the wrapper, reported in #2997, and improve investigation by saving all wrappers, make the test shell wrapper filename even more unique that it already was. Besides the phase and guest names, include also test name and its serial number. That should make the filename unlike any other filename created for other tests.

See https://github.com/teemtee/tmt/issues/2997

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -49,6 +49,7 @@ from tmt.utils import (
     field,
     key_to_option,
     option_to_key,
+    render_template,
     )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


lukaszachy (approved):

psss (approved):
I was also able to reproduce with the `connect` provision. The fix looks good and works as expected. Thanks for quickly addressing this!

happz:
Unrelated test failures, enough approvals, merging.

PR #3003, opened 2024-06-11
Add option to skip output log in junit and polarion reports

KwisatzHaderach:
We need the option to lower the amount of lines in xml files, so this patch adds the ability to skip the standard output and just keep failures if they appear.
The default is kept on showing the standard output.

Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema

### DIFF
--- a/tmt/steps/report/__init__.py
+++ b/tmt/steps/report/__init__.py
@@ -9,6 +9,7 @@ import tmt.steps
 from tmt.options import option
 from tmt.plugins import PluginRegistry
 from tmt.steps import Action
+from tmt.utils import field
 
 if TYPE_CHECKING:
     import tmt.cli
@@ -16,7 +17,12 @@ if TYPE_CHECKING:
 
 @dataclasses.dataclass
 class ReportStepData(tmt.steps.StepData):
-    pass
+    output_log: bool = field(
+        default=True,
+        option=('--output-log / --no-output-log'),
+        is_flag=True,
+        show_default=True,
+        help='Show full output in resulting xml file.')
 
 
 ReportStepDataT = TypeVar('ReportStepDataT', bound=ReportStepData)
--- a/tmt/steps/report/__init__.py
+++ b/tmt/steps/report/__init__.py
@@ -17,12 +17,12 @@ if TYPE_CHECKING:
 
 @dataclasses.dataclass
 class ReportStepData(tmt.steps.StepData):
-    output_log: bool = field(
+    include_output_log: bool = field(
         default=True,
-        option=('--output-log / --no-output-log'),
+        option=('--include-output-log / --no-include-output-log'),
         is_flag=True,
         show_default=True,
-        help='Show full output in resulting xml file.')
+        help='Include full standard output in resulting xml file.')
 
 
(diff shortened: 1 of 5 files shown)


tmt/steps/report/__init__.py
```diff
@@ -9,14 +9,20 @@
...
 if TYPE_CHECKING:
     import tmt.cli
 
 
 @dataclasses.dataclass
 class ReportStepData(tmt.steps.StepData):
-    pass
+    output_log: bool = field(
+        default=True,
+        option=('--output-log / --no-output-log'),
```
happz, line 22: Maybe `include-onput-log`? `output-log` alone sounds like there would be an output log created by the plugin rather than including output in XML.

happz (approved):

tmt/steps/report/__init__.py
```diff
@@ -9,14 +9,20 @@
...
 
 @dataclasses.dataclass
 class ReportStepData(tmt.steps.StepData):
-    pass
+    include_output_log: bool = field(
+        default=True,
+        option=('--include-output-log / --no-include-output-log'),
+        is_flag=True,
+        show_default=True,
+        help='Include full standard output in resulting xml file.')
```
psss, line 25: So this new option is relevant for `junit` and `polarion` but not other report plugin, right? This means that `display` or `html` plugin help now suggests this is also supported but in fact it is not. I guess this could be confusing. What about adding the option only to plugins which actually support it?

psss (changes requested):
The change looks good, I'm just not sure about adding the new option to all available `report` plugins as it might be confusing to users.

tmt/steps/report/__init__.py
KwisatzHaderach: This is problematic down the line since I'm using this objects type at one place, but I'll try to improve it. Maybe add common parent data class for junit and polarion since they share this argument.

happz: Ouch, I missed that, +1. This belongs to `junit` plugin, maybe also to `polarion`, but not to shared `report` base.

KwisatzHaderach: it needs to be in both, but since I'm calling junit from polarion then types don't really match so the check there needs to change, but IMO it should be fairly easy, having it in two places is fine as long as typing stops complaining (I'll make it stop complaining)

KwisatzHaderach: Yay, trial and error FTW, fixed in  https://github.com/teemtee/tmt/pull/3003/commits/4bcab7f01f50c20ee5da353142b388c5a2b26f85

happz: Don't worry about copy & paste: there's at least one other shared option, one telling a plugin which file to save output into, `junit` and `html` have this one, so solving this issue is on my TODO list. Providing some classes to add shared options, to avoid duplication, is the current idea, and I can easily resolve not just the output file, but also this one when the time comes.

psss (approved):
Now looks good, thanks!

lukaszachy (approved):

psss:
Test failures irrelevant --> merging.

psss:
Seems we haven't mentioned the newly added option in the release notes: Proposing in #3093.

Issue #2657, opened 2024-01-30
Review documentation of individual plugins 


happz:
## Rules

**General ones**

* design the docstrings and help texts as if they are destined to be rendered by Sphinx, i.e. make use of ReST goodies: literals for literals - metavars, values, envvar names, commands, keys, etc., `code-block` for blocks of code or examples. It leads to better HTML docs and we will have a nice CLI renderer as well, there is no need to compromise for the sake of CLI.
* use full sentences, i.e. capital letters at the beginning & a full stop at the end.
* use Python's multiline strings rather than joining multiple strings.

**Plugin docs**

* plugin docstring provides the bulk of its CLI help and HTML documentation. It should describe what the plugin does.
* less than trivial use cases and keys deserve an example or two.
* unless there's an important difference, describe the plugin's configuration in terms of fmf rather than CLI. It's easy to map fmf to CLI options, fmf makes a better example for someone writing fmf files.

**Plugin fields**

* when referring to fields in user-facing docs, speak about "keys", i.e. "`playbook` key of `prepare/ansible` plugin". Keys are mapped 1:1 to CLI options, but let's make sure we avoid "fields", "settings" and other synonyms.
* metavar should represent the semantic of the expected value, i.e. `--file PATH` is better than `--file FILE`, `--playbook PATH|URL` is better than `--playbook PLAYBOOK`.
* if there's a default value, it belongs to the `default=` key of `field()`, and the help text should not mention it as the "`Default is ...`" can be easily added automatically and rendered correctly with `show_default=True`.

## Pull requests

* `discover`
  * `fmf`
  * `shell`
* `execute`
  * `internal`
  * `upgrade`
* `finish`
  * `ansible`
  * `shell`
* `prepare`
  * `ansible`
  * `install`
  * `shell`
* `provision`
  * `artemis`
  * `connect`
  * `local`
  * `mrack`
  * `podman`
  * `testcloud`
* `report`
  * `display`
  * `html`
  * `junit`
  * `polarion`
  * `reportportal`

PR #3007, opened 2024-06-12
Add documention best practices, adjust release notes

happz:
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note

### DIFF
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -351,6 +351,25 @@ good to include respective documentation. You can add or update
 the :ref:`specification`, extend the :ref:`examples` or write a
 new chapter for the user :ref:`guide`.
 
+tmt documentation is written with `reStructuredText`__ and built
+with `Sphinx`__. Various features of both reST and Sphinx are used
+widely in tmt documentation, from inline markup to references, feel
+free them as well to link new or updated documentation to relevant
+parts, to highlight important points, or to provide helpful examples.
+
+A couple of best practices when updating documentation:
+
+* when referring to a plugin, its options or documentation, prefer
+  reference to ``plugin/STEP/PLUGIN`` rather than to
+  ``/spec/plans/STEP/PLUGIN``. ``:ref:`plugin/prepare/ansible``` is
+  better than ``:ref:`/spec/plans/prepare/ansible```.
+
+__ https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html
+__ https://www.sphinx-doc.org/en/master/
+
+Examples
+------------------------------------------------------------------
+
 By default, examples provided in the specification stories are
 rendered as ``yaml``. In order to select a different syntax
 highlighting schema add ``# syntax: <format>``, for example:
@@ -367,6 +386,9 @@ Building documentation is then quite straightforward:
 
 Find the resulting html pages under the ``docs/_build/html``
 folder.
+
+Visual themes
+------------------------------------------------------------------
 
 Use the ``TMT_DOCS_THEME`` variable to easily pick custom theme.
 If specified, ``make docs`` would use this theme for documentation
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -374,6 +374,53 @@ A couple of best practices when updating documentation:
 
     # This should be avoided:
     :ref:`/spec/plans/prepare/ansible`
+* Design the plugin docstrings and help texts as if they are to be
+  rendered by Sphinx, i.e. make use of ReST goodies: literals for
+  literals - metavars, values, names of environment variables, commands,
+  keys, etc., ``code-block`` for blocks of code or examples. It leads to
+  better HTML docs and tmt has a nice CLI renderer as well, therefore
+  there is no need to compromise for the sake of CLI.
+* Use full sentences, i.e. capital letters at the beginning & a full
+  stop at the end.
+* Use Python multiline strings rather than joining multiple strings over
+  several lines. It often leads to leading and/or trailing whitespace
+  characters that are easy to miss.
+* Plugin docstring provides the bulk of its CLI help and HTML
+  documentation. It should describe what the plugin does.
+* Other than trivial use cases and keys deserve an example or two.
+* Unless there's an important difference, describe the plugin's
+  configuration in terms of fmf rather than CLI. It is easy to map fmf
+  to CLI options, and fmf makes a better example for someone writing fmf
+  files.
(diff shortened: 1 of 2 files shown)


docs/contribute.rst
```diff
@@ -351,6 +351,25 @@ good to include respective documentation. You can add or update
 the :ref:`specification`, extend the :ref:`examples` or write a
 new chapter for the user :ref:`guide`.
 
+tmt documentation is written with `reStructuredText`__ and built
+with `Sphinx`__. Various features of both reST and Sphinx are used
+widely in tmt documentation, from inline markup to references, feel
+free them as well to link new or updated documentation to relevant
```
therazix, line 357: It looks like a few words are missing here.
```suggestion
widely in tmt documentation, from inline markup to references. Feel
free to use them as well to link new or updated documentation to relevant
```

docs/contribute.rst
```diff
@@ -351,6 +351,25 @@ good to include respective documentation. You can add or update
...
+tmt documentation is written with `reStructuredText`__ and built
+with `Sphinx`__. Various features of both reST and Sphinx are used
+widely in tmt documentation, from inline markup to references. Feel
+free to use them as well to link new or updated documentation to relevant
+parts, to highlight important points, or to provide helpful examples.
+
+A couple of best practices when updating documentation:
+
+* when referring to a plugin, its options or documentation, prefer
+  reference to ``plugin/STEP/PLUGIN`` rather than to
```
lukaszachy, line 363: Is it 'plugin' or 'plugins'?

docs/contribute.rst
```diff
@@ -351,6 +351,25 @@ good to include respective documentation. You can add or update
 the :ref:`specification`, extend the :ref:`examples` or write a
 new chapter for the user :ref:`guide`.
 
+tmt documentation is written with `reStructuredText`__ and built
+with `Sphinx`__. Various features of both reST and Sphinx are used
+widely in tmt documentation, from inline markup to references. Feel
+free to use them as well to link new or updated documentation to relevant
+parts, to highlight important points, or to provide helpful examples.
+
+A couple of best practices when updating documentation:
```
martinhoyer, line 360: This says "couple of best practices", but is followed by only one bullet point. Also the bullet point starts with lowercase letter and ends with dot, which afaik is not good?

docs/contribute.rst
```diff
@@ -351,6 +351,25 @@ good to include respective documentation. You can add or update
...
+
+* when referring to a plugin, its options or documentation, prefer
+  reference to ``plugin/STEP/PLUGIN`` rather than to
+  ``/spec/plans/STEP/PLUGIN``. ``:ref:`plugin/prepare/ansible``` is
+  better than ``:ref:`/spec/plans/prepare/ansible```.
```
falconizmi, line 365: If you check the documentation, it doesnt look as it should. If you remove the double \`, it will fix it but even after fixing it, I dont know if we want to look like this: ... `ansible` is better than `ansible`. Maybe adding some text link to make it more readable?

docs/releases.rst
```diff
@@ -89,7 +89,7 @@ New environment variable ``TMT_REPORT_ARTIFACTS_URL`` has been added
 to :ref:`command-variables`. It can be used to provide a link for
 detailed test artifacts for report plugins to pick.
 
-:ref:`Beaker</spec/plans/provision/beaker>` provision plugin gains
+:ref:`Beaker<plugins/provision/beaker>` provision plugin gains
```
martinhoyer, line 92: Is this by design different than the syntax above?

docs/contribute.rst
happz: I was worried it might get rendered as a mess, but it seemed good on my end. I will try a `code-block`, that should be safe.

docs/contribute.rst
happz: Changed to `plugins` in 477d5926

docs/contribute.rst
happz: * changed to upper-case in 477d5926, no problem.
* I would like to keep "practices" because I hope we could add more points - any quick idea what to ask contributors when it comes to docs?

docs/releases.rst
happz: It's the same syntax, i.e. pointing to `plugins/provision/beaker`, but defines specific label, `Beaker` instead of `beaker` derived from the target. It's to make the sentence look better, added as an example in 477d5926

martinhoyer (approved):

docs/contribute.rst
martinhoyer: No quick ideas here

docs/contribute.rst
martinhoyer: Hmm it rendered fine for me

falconizmi: Okay, I wanted to make sure if you are okay that the ref links are not clickable

falconizmi: Does it rendered fine in this [doc](https://tmt--3007.org.readthedocs.build/en/3007/contribute.html#docs)

martinhoyer: ah, I understand now. This is an example how to write docs, so I believe this is how it supposed to look, yeah.

falconizmi (approved):

docs/contribute.rst
happz: In 69bc8908 I proposed a couple more points I collected in https://github.com/teemtee/tmt/issues/2657 originally.

docs/contribute.rst
```diff
@@ -351,6 +351,83 @@ good to include respective documentation. You can add or update
...
+A couple of best practices when updating documentation:
+
+* When referring to a plugin, its options or documentation, prefer
+  reference to ``plugins/STEP/PLUGIN`` rather than to older
+  ``/spec/plans/STEP/PLUGIN``:
+
+  .. code-block:: rest
+
+    # This is good:
+    :ref:`plugin/prepare/ansible`
```
therazix, line 369: Shouldn't it be `plugins`?
```suggestion
    :ref:`plugins/prepare/ansible`
```

docs/contribute.rst
thrix: @lukaszachy pls rereview

psss: Seems to be covered, resolving.

psss (approved):
Very nice documentation hints! Thanks for putting this together.

psss:
I think it's worth to mention the new docs best practices in the release notes as well, done in 88b6434a.

PR #3010, opened 2024-06-12
Add more tag separator options to polarion import

### LINKED ISSUE #2969: polarion import does not correctly parse Polarion:tags field.
Polarion:tags field separates different tags with a semi-colon. When tmt imports this field it imports it as a single tag.

e.g.
TAG1;TAG2 in Polarion is placed in the main.fmf as is.
```
tag:
  -  TAG1;TAG2
```

But it should be:
```
tag:
  -  TAG1
  -  TAG2
```

KwisatzHaderach:
Fixes #2969 

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/convert.py
+++ b/tmt/convert.py
@@ -922,6 +922,8 @@ def read_polarion_case(
     if polarion_case.tags:
         if not current_data.get('tag'):
             current_data['tag'] = []
+        # Space is the default separator, but let's try to autodetect couple different options
+        # as Polarion tag field is free text so it can have anything
         separator = ' '
         if ',' in polarion_case.tags:
             separator = ','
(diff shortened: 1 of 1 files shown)


tmt/convert.py
```diff
@@ -922,7 +922,12 @@ def read_polarion_case(
     if polarion_case.tags:
         if not current_data.get('tag'):
             current_data['tag'] = []
-        for tag in polarion_case.tags.split():
+        separator = ' '
```
happz, line 925: A comment would be nice, something along the lines of "the space is the default, but let's try to autodetect a separator, maybe we could do better". The "why" is not clear from this sequence of tests.

happz (approved):

therazix (approved):

happz:
Unrelated failures, merging.

PR #3011, opened 2024-06-12
Fix docstrings of regex template filters

happz:
They we a bit messy, especially examples.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -6,7 +6,6 @@ from requre.cassette import StorageKeysInspectSimple
 from requre.helpers.guess_object import Guess
 from requre.helpers.requests_response import RequestResponseHandling
 
-import tmt.utils
 import tmt.utils.git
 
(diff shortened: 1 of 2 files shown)


tests/integration/__init__.py
```diff
@@ -7,6 +7,7 @@
 from requre.helpers.requests_response import RequestResponseHandling
 
 import tmt.utils
+import tmt.utils.git
```
martinhoyer, line 10: hmm, isn't this redundant?

martinhoyer (approved):

KwisatzHaderach (approved):

happz:
Added `blocked` label: depends on a branch that's been pushed out to 1.36, the rebase would be a mess, and it's not critical.

tests/integration/__init__.py
happz: Addressed in 23590472

martinhoyer (approved):

PR #3015, opened 2024-06-13
Fix normalization of `hardware` key/option

happz:
A bug was introduced in the patching adding multiple disks to testcloud plugin, and the test was missing.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/unit/test_hardware.py
+++ b/tests/unit/test_hardware.py
@@ -89,6 +89,20 @@ def test_constraint_components_pattern(value: str, expected: tuple[Any, Any]) ->
     assert match.groups() == expected
 
 
+def test_normalize_hardware(root_logger) -> None:
+    # All major classes of requirements:
+    spec = (
+        # Simple name.child_name=value
+        'cpu.processors=1',
+        # The same but with cpu.flags which have special handling
+        'cpu.flag!=avc',
+        # name[peer_index].child_name=value
+        'disk[1].size=1'
+        )
+
+    tmt.steps.provision.normalize_hardware('', spec, root_logger)
+
+
 @pytest.mark.parametrize(
     ('spec', 'expected_exc', 'expected_message'),
     [
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -531,7 +531,7 @@ def normalize_hardware(
 
             elif components.name == 'cpu' and components.child_name == 'flag':
                 if components.name not in merged:
-                    merged[components.name] = []
+                    merged[components.name] = {}
 
                 if 'flag' not in merged['cpu']:
                     merged['cpu']['flag'] = []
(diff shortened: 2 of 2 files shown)


happz:
@martinhoyer to my horror, I broke `hardware` normalization, and since I didn't bother to add a unit test, nobody noticed till I used `--hardware` just this evening. So, I suppose one more to include in 1.34, this time it's a genuine bug fix :/

martinhoyer:
Nice catch. Thanks for the heads-up.  
I'd say we can do the release on Monday anyway. (DevConf)  
Unless you really want to do this week :)

martinhoyer (commented):
It looks good to me, but clearly I don't understand the code enough to be able to click approve with clear conscience.  
@happz If you could help me understand, it would be much appreciated :pray:  
I get that line 519 is `merged[components.name] = []` as there is peer_index, but am not sure why `merged[components.name]` is otherwise `{}`. Am I looking correctly at https://github.com/teemtee/tmt/blob/e4991dd46c765151a1280f9e0f89ce3ebf38202e/tmt/hardware.py#L230 ?

martinhoyer:
@psss @lukaszachy Could you take a look please, so we can do the release? :)

lukaszachy:
Hm, there is missing fmf on rawhide (as it cannot be built https://kojipkgs.fedoraproject.org//work/tasks/6207/118796207/build.log) which blocks tmt.

martinhoyer:
> Hm, there is missing fmf on rawhide (as it cannot be built https://kojipkgs.fedoraproject.org//work/tasks/6207/118796207/build.log) which blocks tmt.
> 
> Hm, we should monitor bugzilla as well, it was opened in May :/ (https://bugzilla.redhat.com/show_bug.cgi?id=2279994)

whoops. I wonder if https://github.com/teemtee/fmf/issues/194 has something to do with it.  
Probably shouldn't block this PR though?

martinhoyer:
> whoops. I wonder if [teemtee/fmf#194](https://github.com/teemtee/fmf/issues/194) has something to do with it. Probably shouldn't block this PR though?

actually, I think it's the use of pkg_resources. Looking into it.

happz:
> It looks good to me, but clearly I don't understand the code enough to be able to click approve with clear conscience. @happz If you could help me understand, it would be much appreciated 🙏 I get that line 519 is `merged[components.name] = []` as there is peer_index, but am not sure why `merged[components.name]` is otherwise `{}`. Am I looking correctly at
> 
> https://github.com/teemtee/tmt/blob/e4991dd46c765151a1280f9e0f89ce3ebf38202e/tmt/hardware.py#L230
> 
> ?

`merged` is basically this structure:

```yaml
merged:
  # merged[component.name] = [] for these:
  disk:
    - size: 10GB
    - size: 10GB
  network:
    - type: eth

  # merged[component.name] = {} for these:
  cpu:
    flags:
      - avs
      - avs2
    model: 79
  device:
    vendor: ...

  # and merged[component.name] is whatever the value is for these:
  hostname: foo.bar.com
  memory: 4GB
```

The issue was in `merged[components.name] = []` being used for all three kinds of requirements.

martinhoyer (approved):
Thanks @happz, much appreciated!

lukaszachy (approved):

Issue #2939, opened 2024-05-17
RFE: `tmt test id` option to check for duplicate IDs


engelmi:
The tmt CLI can be used to generate UUIDs for tests via `tmt test id`. Unfortunately, it does not check for duplicates. This can happen, for example, when tests are being copied and the ID is not adjusted - which can happen quite easily. 

Therefore, I'd like to propose to add an additional check for duplication to the `tmt test id` command. Possible options I can think of are:
1. checking implicitly, generating new UUIDs for identified duplicates
2. provide another option to check and overwrite identified duplicates, e.g. `--overwrite-duplicates` 
3. provide another option just to check for duplicates, e.g. `--check-duplicates`

For option 1. and 2. the `--dry` option would only print out the identified tests with duplicates and not change the file. But maybe there are other/better option? If this RFE is accepted and refined, I can also start working on it and provide a PR for it.

On a side note, it seems that `tmt test id` is missing in the documentation:
https://tmt.readthedocs.io/en/latest/stories/cli.html#stories-cli-test

happz:
Sounds also like a good idea for yet another `tmt * lint` check, one that looks not just at a single object at a time, but also keeps some notion of the global state, i.e. the existing IDs that were previously detected in the set of objects already examined.

engelmi:
I agree, I think it would fit good as another lint check. 
And yes, if I understand it correctly (but only had a brief look at it), I'd need to introduce state into the `tmt test id` command (keeping track of the IDs). Is this acceptable or should this just be a lint check? It would be convenient to auto-fix duplicate IDs, though. Is there a `tmt * lint --auto-fix` or something similar? @happz

happz:
> I agree, I think it would fit good as another lint check. And yes, if I understand it correctly (but only had a brief look at it), I'd need to introduce state into the `tmt test id` command (keeping track of the IDs). Is this acceptable or should this just be a lint check? It would be convenient to auto-fix duplicate IDs, though. Is there a `tmt * lint --auto-fix` or something similar? @happz

[...]

PR #3017, opened 2024-06-14
Added lint check for duplicate test ids

engelmi:
Fixes: https://github.com/teemtee/tmt/issues/2939

In order to be able detect duplicate ids of tests for a given tmt.Tree, a new lintable class Tests has been introduced. It doesn't reflect a specific node in the tree, but rather a collection of test nodes. On this collection, the duplicate lint check is being executed. 

The lintable `Tests` implements the lint function for the duplicate test id check - this can be further extended in the future with additional lint checks.

Example usage triggering the new lint check:
```bash
$ tmt lint <path>
$ tmt tests lint <path>
$ tmt stories lint <path>
$ tmt plans lint <path>

# example output - detected no duplicate
Lint checks on all tests
pass G001 no duplicate test id detected

# example output - detected duplicate
Lint checks on all
fail G001 duplicate id "6886e7c8-7548-4cf7-8dd4-d2d46b125fd3" in test "/tests/clean/chain"
fail G001 duplicate id "6886e7c8-7548-4cf7-8dd4-d2d46b125fd3" in test "/tests/clean/runs"

Lint checks on all
fail G001 duplicate id "5cf92c54-e073-475c-970f-b8e090ec4da2" in "/duplicates/tests/duplicateid"
fail G001 duplicate id "5cf92c54-e073-475c-970f-b8e090ec4da2" in "/duplicates/tests/duplicateid"
fail G001 duplicate id "5cf92c54-e073-475c-970f-b8e090ec4da2" in "/duplicates/tests/duplicateid"
```

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation


happz:
> @happz I wanted to ask if this is going into the right direction as we discussed in the RFE #2939? (only a draft PR for now that implements the rough functionality)

In general, it looks good, it is indeed along the lines of some collection-wide linter. I'll add just a couple of minor minor comments.

tmt/base.py
```diff
@@ -1529,6 +1529,45 @@ def lint_require_type_field(self) -> LinterReturn:
         yield LinterOutcome.FIXED, 'added type to requirements'
 
 
+@dataclasses.dataclass(repr=False)
+class Tests(tmt.lint.Lintable['Tests']):
```
happz, line 1533: I would make it very clear that this `Tests` class is not representing some set of test for any other purpose than lining them as a collection of objects. We have `Test` class representing a test, and `Tests` suggests it has more to do with test collections than just the lining.

Since plans and stories also have IDs, maybe we should have a generic collection to hold `tmt.base.Core` instances, no matter what exactly they are, and lint them. `tmt test lint` would fill it with `Test` instances, `tmt lint` would fill it with tests, plans and stories - I suppose test ID should not conflict with story ID either.

tmt/base.py
```diff
@@ -1529,6 +1529,45 @@ def lint_require_type_field(self) -> LinterReturn:
...
+
+        has_duplicates = False
+        for id, test_names in duplicate_ids.items():
+            if len(test_names) > 1:
+                has_duplicates = True
+                for name in test_names:
+                    yield LinterOutcome.FAIL, f'duplicate id "{id}" in test "{name}"'
+
+        if has_duplicates:
+            return
```
happz, line 1563: We could simplify a bit, how about the following? (`id` would conflict with a built-in function, hence calling it `test_id`, no need for an extra flag, `defaultdict` to simplify collecting phase:

```python
ids: collection.defaultdict[str, list[str]] = collection.defaultdict(list)

for test in self.tests:
    if test.id is None:
        continue
    ids[test.id].append(test.name)

duplicates = {
    test_id: test_names for test_name, test_ids in duplicate_ids.items() if len(test_names) > 1
}

for test_id, test_names in duplicates.items():
    for name in test_names:
        yield LinterOutcome.FAIL, f'duplicate id "{test_id}" in test "{name}"'

if duplicates:
    return
```

tmt/cli.py
```diff
@@ -566,9 +566,64 @@ def _lint_class(
...
+
+    echo()
+
+    return exit_code
+
+
 def do_lint(
         context: Context,
         klasses: list[Union[type[tmt.base.Test], type[tmt.base.Plan], type[tmt.base.Story]]],
+        klass_lists: list[Union[type[tmt.base.Tests], None]],
```
happz, line 626: Is a new parameter needed? What's preventing adding this class to `klasses`?

tmt/base.py
engelmi: Looks great, changed it accordingly.

tmt/cli.py
engelmi: I refactored it a bit and was able to remove this additional parameter. Basically, the collection will use the type in `klasses` to determine what the collection to get from the tree.

tmt/base.py
```diff
@@ -1529,6 +1529,45 @@ def lint_require_type_field(self) -> LinterReturn:
...
+    def __init__(self, tests: list['tmt.Test'], *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.tests = tests
+
+    @classmethod
+    def from_tree(cls, tree: 'tmt.Tree') -> 'Tests':
+        return Tests(tree.tests())
+    
+    def lint_no_duplicate_test_ids(self) -> LinterReturn:
+        """ X001: no duplicate test ids """
```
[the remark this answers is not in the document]
engelmi: Good catch! Updated it. 
Do you think the `X` prefix is ok? I wasn't sure what to choose.

engelmi:
> In general, it looks good, it is indeed along the lines of some collection-wide linter. I'll add just a couple of minor minor comments.
> 
> Edit: BTW plans and stories also have IDs :) Edit2: hm, can story have same id as a test? Maybe we should write this from the perspective of a collection of tmt objects, not just the collection of tmt tests. A container of not only tests, but any objects loaded by `lint` subcommand - `tmt test lint` would fill it with tests, while `tmt lint` would fill it with everything. `tmt.base.Core` instead of `tmt.base.Test`.

Thanks for your feedback! I tried to apply the changes. PTAL @happz 

I was also looking into updating the (unit) tests and it seems like when adding test files with the duplicate ids, the other tests will break because of this. What do you think about adding a flag to enable/disable such linters on collections?

tmt/base.py
happz: > > Since plans and stories also have IDs, maybe we should have a generic collection to hold tmt.base.Core instances
> 
> Good point! I changed it to hold either a list of `tmt.base.Test`, `tmt.base.Plan` or `tmt.base.Story`. And since naming is always the most difficult part: I am not sure I found a good name with `LintableCollection`. What do you think?

`LintableCollection` sounds good.

WRT test, plan or story, can you try it out with `list[Core]` instead of union of homogenous lists of subclasses? All three classes are derived from `Core`, `Core` defines `id`, it should work.

> > I suppose test ID should not conflict with story ID either.
> 
> Ah, didn't know about this. With the current approach I can define only lint rules for collections of one type of object like Tests or Stories.

Switching to `Core` might help.

> Thinking about using it: Should `tmt test lint` then inspect all `tmt.base.Test` and `tmt.base.Story`, and show an error if a duplicate id is found? In that case it wouldn't lint only tests.

I'd stick to what `tmt * lint` inspects: `tmt test|plan|story lint` fills the collection with all kinds of objects, and I would get them validated all against each other, i.e. I'd be able to see a test conflicting with a story. `tmt test lint` would fill it with tests only, therefore no validation between kinds of tmt objects.

But this is a subjective opinion, I can also imagine this cross-examination might be interesting enough that it would be enabled in general, for all subcommand, and `tmt test lint` would run test linters for tests and collection linters for everything too. Let's see what others think.

tmt/base.py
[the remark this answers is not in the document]
happz: No idea. Naming is hard. I wouldn't pick `X`, but I have no better idea.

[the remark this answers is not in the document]
engelmi: Renamed it to `G` (for Group) for now. We can change it later if needed.

tmt/base.py
engelmi: > I'd stick to what tmt * lint inspects: tmt test|plan|story lint fills the collection with all kinds of objects, and I would get them validated all against each other, i.e. I'd be able to see a test conflicting with a story. tmt test lint would fill it with tests only, therefore no validation between kinds of tmt objects.

That makes sense - and also from a user point of view might be quite helpful to identify duplicate ids across multple types.  

> WRT test, plan or story, can you try it out with list[Core] instead of union of homogenous lists of subclasses? All three classes are derived from Core, Core defines id, it should work.

Yes, you are right. Switching to `Core` required a bit of refactoring, but now it is also quite easy to extend with further lint rules on such sets.

engelmi:
> Did you try putting them into a dedicated, separate fmf root? We already have https://github.com/teemtee/tmt/tree/main/tests/lint/test/data, we could add something like data-duplicate-id or something, just for files to test this linter. Then tests for other linters would not consider these at all.

Thanks for the hint! I think I was able to create a few new tests based on this.

engelmi:
@happz Ping :)

tmt/base.py
```diff
@@ -1539,6 +1541,40 @@ def lint_require_type_field(self) -> LinterReturn:
...
 
 
+@dataclasses.dataclass(repr=False)
+class LintableCollection(tmt.lint.Lintable['LintableCollection']):
+    """ Linting rules applied to a collection of Tests, Plans or Stories """
+
+    def __init__(self, objs: list["tmt.base.Core"], *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.objs = objs
+        self.cls_name = type(objs[0]).__name__ if len(objs) > 0 else ""
```
happz, line 1551: Is this field used? I can find it neither in this patch nor in tmt code base.

tmt/cli.py
```diff
@@ -530,6 +530,41 @@ def finito(
...
 
+def _apply_linters(lintable: Union[
+            tmt.lint.Lintable[tmt.base.Test], 
+            tmt.lint.Lintable[tmt.base.Plan], 
+            tmt.lint.Lintable[tmt.base.Story], 
+            tmt.lint.Lintable[tmt.base.LintableCollection]], 
+            linters: list[tmt.lint.Linter],
+            failed_only: bool,
+            enforce_checks: list[str],
+            outcomes: list[tmt.lint.LinterOutcome],) -> tuple[bool, Union[list[tmt.lint.LinterRuling], None]]:
```
happz, line 541: * one spurious comma after `outcomes` type,
* instead of `Union[foo, None]`, use `Optional` - it's the same, but we use `Optional` consistently throughout the code.

tmt/cli.py
```diff
@@ -530,6 +530,41 @@ def finito(
...
+        """ 
+        Local function applying linters on a lintable instance and filter the
+        resulting rulings based on users desired ones.
+        """
```
happz, line 545: Docstring should have a short, single line at its beginning, e.g. "Apply linters on a lintable and filter out disallowed outcomes"?

tmt/cli.py
engelmi: Changed :heavy_check_mark:

tmt/base.py
engelmi: No, its not used anymore. It is a leftover from a previous refactoring that I missed to remove. 
Removed :heavy_check_mark:

happz:
@engelmi we highly recommend installing pre-commit in your local repo, see https://tmt.readthedocs.io/en/stable/contribute.html#develop. With pre-commit in place, `git commit` would be blocked because of errors now reported by Github check, see https://github.com/teemtee/tmt/actions/runs/10075730826/job/27854800266?pr=3017

engelmi:
> @engelmi we highly recommend installing pre-commit in your local repo, see https://tmt.readthedocs.io/en/stable/contribute.html#develop. With pre-commit in place, `git commit` would be blocked because of errors now reported by Github check, see https://github.com/teemtee/tmt/actions/runs/10075730826/job/27854800266?pr=3017

I installed the pre-commit locally and fixed the detected issues. 
Could you approve the GitHub CI run again? @happz

engelmi:
Tests are failing on testing-farm. I'll try to figure out what I broke and how to fix it. 
Do you have maybe any hints? @happz

happz:
> Tests are failing on testing-farm. I'll try to figure out what I broke and how to fix it. Do you have maybe any hints? @happz

Hmm, all seems unrelated to your changes, it seems there's some problem with debuginfo in rawhide, so, ruining all results.

engelmi:
> > Tests are failing on testing-farm. I'll try to figure out what I broke and how to fix it. Do you have maybe any hints? @happz
> 
> Hmm, all seems unrelated to your changes, it seems there's some problem with debuginfo in rawhide, so, ruining all results.

Ah ok. In that case, can we re-run the pipeline again since a few days have passed?

martinhoyer (approved):
Nice. Maybe worth adding a release note.

engelmi:
> Nice. Maybe worth adding a release note.

Thanks! Should I add a release note to `docs/releases.rst` or is this done by the tmt maintainers on release? @martinhoyer

martinhoyer:
> Thanks! Should I add a release note to `docs/releases.rst` or is this done by the tmt maintainers on release? @martinhoyer

If you'll have time, feel free to add it there, thanks!  
Oh, and if you're not yet in docs/overview.rst list of contributors, feel free to add yourself there, especially if you have any special name preference, otherwise we can add your name (from github account) during the release. :)

engelmi:
> > Thanks! Should I add a release note to `docs/releases.rst` or is this done by the tmt maintainers on release? @martinhoyer
> 
> If you'll have time, feel free to add it there, thanks! Oh, and if you're not yet in docs/overview.rst list of contributors, feel free to add yourself there, especially if you have any special name preference, otherwise we can add your name (from github account) during the release. :)

Updated :heavy_check_mark: 
@martinhoyer PTAL

martinhoyer:
> Updated ✔️ @martinhoyer PTAL

Thanks @engelmi, looks like you missed the release note in https://github.com/teemtee/tmt/pull/3017/commits/d06fd8eacf79b76b5b824c54fca8055fa70e3cee

engelmi:
> > Updated ✔️ @martinhoyer PTAL
> 
> Thanks @engelmi, looks like you missed the release note in [d06fd8e](https://github.com/teemtee/tmt/commit/d06fd8eacf79b76b5b824c54fca8055fa70e3cee)

Ups, good catch! Added it.

KwisatzHaderach (approved):

therazix (approved):

happz (approved):

Issue #2687, opened 2024-02-15
Possible race condition in the `testcloud` plugin


psss:
Seems that `/tests/prepare/multihost` sometimes fails to connect to the guest.

```
[guest-3]         multihost name: guest-3
[guest-3]         arch: x86_64
[guest-3]         distro: Fedora Linux 39 (Cloud Edition)
[guest-3]         kernel: 6.5.6-300.fc39.x86_64
[guest-3]         package manager: dnf
[guest-3]         selinux: yes
[guest-3]         is superuser: yes
[guest-1]         finished
[guest-1]         fail: Failed to connect in 300s.
    finish
    
[guest-2]         guest: stopped
[guest-2]         guest: removed
[guest-3]         guest: stopped
[guest-3]         guest: removed
```

Here's an [example job](http://artifacts.osci.redhat.com/testing-farm/bc8558d7-1e6e-4caa-8baa-e02119467978/) and [one more](http://artifacts.osci.redhat.com/testing-farm/4dad9954-0c7e-48af-a7ae-9560bd237b84/). As @happz [mentioned](https://github.com/teemtee/tmt/pull/2677#issuecomment-1941970750) in #2677 this stinks with race conditions. @frantisekz, could you please have a look?

frantisekz:
So, I digged into this a bit:

I was able to reproduce the issue with the following tmt plan (on some attempts):
```
/test:
    test: echo

/plan:
    execute:
        how: tmt
    discover:
        how: fmf

    provision:
        - how: virtual
        - how: virtual
        - how: virtual
        - how: virtual
        - how: virtual
        - how: virtual
        - how: virtual
```

The thing is, in cases where it fails like in the mentioned jobs, the ssh succeeds on another try. cloud-init data are generated and appended properly in the affected VMs, and my best guess is that the ssh connection is attempted before cloud-init finishes its job in the VM.

I've tried if something like disabling the ssh early boot would help (it would) and let cloud-init restart it only after it finishes what it needs to. The problem is that it seems to be impossible to pass grub arguments via libvirt (we would have to restructure it to use a direct kernel boot which is can of worms on its own).

The another possible way to handle it would be to append (tmt-side) "-o PasswordAuthentication=no" to the ssh connections that should be using ssh key. This way, the connection would fail instead of a password prompt and that should be handled just fine via tmt's retry mechanism already present.

I'll try to come up with a PR for this.

psss:
[...]

Issue #2348, opened 2023-09-22
Add virtiofs to `tmt provision` local qemu


cgwalters:
See https://dustymabe.com/2023/09/08/using-virtiofs-with-libvirt/virt-install/

Basically virtiofs is really really useful for interacting with local guests as it makes getting data in and out not completely suck.  

It looks like this is really a rfe for https://pagure.io/testcloud but because this project wraps it it'd need to be plumbed through here too.

(This may actually turn into a bit of unprivileged-libvirt virtiofsd work, similar to what I did in coreos-assembler)

thrix:
Nice. I assume this would be an alternative to copying data over network and "share" it similarly as we do with podman, thus speeding up the test data synchronization.

cgwalters:
Yes, it's dramatically more ergonomic and efficient for interacting with local VM than copying files over SSH.

frantisekz:
testcloud maint here, I see "Currently the virtiofs integration doesn’t work as non-root via a qemu:///session connection." in the linked blogpost, which would be a problem since tmt uses user session in the most cases (by default).

cgwalters:
> Currently the virtiofs integration doesn’t work as non-root via a qemu:///session connection." 

Right, that's on track to be fixed https://gitlab.com/libvirt/libvirt/-/issues/535#note_1549107870

Anyways actually the immediate thing I'm trying to do is use `tmt provision`/`testcloud` as a "development shell" provisioner, I'd actually be OK with using the system libvirt too in some cases.

frantisekz:
This is now WIP testcloud side: https://pagure.io/testcloud/pull-request/160#

psss:
Seems that the `testcloud` part has been implemented:

* https://github.com/teemtee/testcloud/commit/c902d0a

So, I guess, we should be unblocked here. @cgwalters or @frantisekz, would you be interested in preparing the patch?

psss:
Ah, sorry, just noticed there's already:

* https://github.com/teemtee/tmt/pull/3143

@frantisekz, let's finish that one?

PR #3036, opened 2024-06-20
Switch testcloud plugin to the Workarounds API

frantisekz:
This will allow to:
- expose a proper API to share commands to be executed during boot/cloud-init between tmt and testcloud, and customize them tmt-side
- open way to address https://github.com/teemtee/tmt/issues/2687 in an atomic way
- open way to address https://github.com/teemtee/tmt/issues/2348

Requires https://pagure.io/testcloud/pull-request/173 ; built in https://copr.fedorainfracloud.org/coprs/frantisekz/testcloud-wip/build/7630902/

Known issues so far:
- el9 host doesn't work - resolved
- TBD

This is meant also as a RFC for the api itself, source and available methods can be seen at https://pagure.io/testcloud/blob/284f136fcbca4e9166d687e90f0108962c1bc8f1/f/testcloud/workarounds.py

### DIFF
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -10,6 +10,7 @@ import re
 import threading
 import types
 from collections.abc import Iterator
+from string import Template
 from typing import TYPE_CHECKING, Any, Optional, Union, cast
 
 import click
@@ -43,6 +44,7 @@ testcloud: Optional[types.ModuleType] = None
 
 # To silence mypy
 DomainConfiguration: Any
+Workarounds: Any
 X86_64ArchitectureConfiguration: Any
 AArch64ArchitectureConfiguration: Any
 S390xArchitectureConfiguration: Any
@@ -58,6 +60,7 @@ def import_testcloud() -> None:
     """ Import testcloud module only when needed """
     global testcloud
     global libvirt
+    global Workarounds
     global DomainConfiguration
     global X86_64ArchitectureConfiguration
     global AArch64ArchitectureConfiguration
@@ -85,6 +88,7 @@ def import_testcloud() -> None:
             UserNetworkConfiguration,
             X86_64ArchitectureConfiguration,
             )
+        from testcloud.workarounds import Workarounds
     except ImportError as error:
         raise ProvisionError(
             "Install 'tmt+provision-virtual' to provision using this method.") from error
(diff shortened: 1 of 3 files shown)


tmt/steps/provision/testcloud.py
```diff
@@ -9,6 +9,7 @@
 import threading
 import types
 from collections.abc import Iterator
+from string import Template
```
happz, line 12: If dropping `format()` in favor of "templates", how about using Jinja? It's the "lingua franca" when it comes to rendering text blobs in tmt, it's used for docs and test/plan/story templates and paths and export, and using it would keep the number of involved "templating solutions" on the same number it is now.

ha said, I'm not sure whether `safe_substitute()` is important for your use case. Jinja would replace an unknown variable with an empty string rather than preserving it, no idea whether you need to run the template through several rounds of evaluation.

frantisekz: Yeah, the per partes replacement is used right now in the code paths (first, the tmt replaces bits like ssh key, then testcloud replaces the rest in its instance.prepare). 

Sigh... it's a good point, I am thinking about something in lines of having this as an interrim solution to allow other work to land, and then continue with replacing/adding/amending api so it can be converted to jinja properly. idk how you'd feel about that?

happz: > Yeah, the per partes replacement is used right now in the code paths (first, the tmt replaces bits like ssh key, then testcloud replaces the rest in its instance.prepare()).

Is there some custom/default relationship between these two? Like `testcloud` plugin ignoring keys that `testcloud` package would then replace if they remain available?

I believe Jinja could support this too, there are two options:

* using `default` filter, `{{ runcommands | default("{{ runcommands }}") }}`. When undefined, it would emit `{{ runcomands }}` for the next round of rendering. Ugly, but might work.
* `DebugUndefined`

  ```python
  env = Environment(undefined=DebugUndefined)
  "{{ a }} is defined, but {{ b}} is undefined".render({"a":"a"})

  # produces:
  "a is defined, but {{ b }} is undefined"
  ```
  We would have to fix `tmt.utils.render_template()` to support this "mode", but doable.

> Sigh... it's a good point, I am thinking about something in lines of having this as an interrim solution to allow other work to land, and then continue with replacing/adding/amending api so it can be converted to jinja properly. idk how you'd feel about that?

Not perfect, but I do it too. As long as I have a good idea how the eventual result should look like, it makes sense to produce less-than-ideal intermediate steps to unblock the overall progress. With comments, describing why things are what they are, mentioning the roadmap ("yep, will fix it in the next decade"), in my experience it's a viable approach.

Anyway. If it's a part of a longer project, it makes sense to wrap up things later, although Jinja may still be usable. On the other hand, Jinja is tmt's pick, not testcloud's - if testcloud package developers decide `string.Template` is the best option, `testcloud` plugin developers would need to work with it :)

psss:
> * open way to address [Possible race condition in the `testcloud` plugin #2687](https://github.com/teemtee/tmt/issues/2687) in an atomic way

That would be nice, seems the problem still randomly appears. Here's a [recent job](https://artifacts.osci.redhat.com/testing-farm/ce8838a1-1619-41a6-acf9-95aaf0b6a9b6/) for example. If possible, let's try to finish this in `1.36`?

tmt/steps/provision/testcloud.py
frantisekz: Yeah, after some experimenting, I'd prefer to jump to string.Template as an interim solution, I've added a TODO around the respective area.

frantisekz:
> > * open way to address [Possible race condition in the `testcloud` plugin #2687](https://github.com/teemtee/tmt/issues/2687) in an atomic way
> 
> That would be nice, seems the problem still randomly appears. Here's a [recent job](https://artifacts.osci.redhat.com/testing-farm/ce8838a1-1619-41a6-acf9-95aaf0b6a9b6/) for example. If possible, let's try to finish this in `1.36`?

I'd consider this one ready now, I've finished the integration part in testcloud - if it's okay with the one TODO item left, I'd go ahead and merge the changes testcloud side, release a new testcloud, and then this could get merged.

For testing - testcloud -800 is being built at https://copr.fedorainfracloud.org/coprs/frantisekz/testcloud-wip/builds/ .

frantisekz:
@psss @martinhoyer @happz This one can now land into 1.35 if possible.

seberm (approved):
I've tested the changes only locally with `testcloud==0.9.13` and it seems everything is working fine.

psss (approved):
Looks nice! Thanks for the improvement.

happz (approved):

PR #3040, opened 2024-06-24
Move abort detection into test invocation

happz:
Similar to reboot detection, hiding the details from individual execution plugins.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -712,14 +712,6 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
 
         return invocation.test.test_framework.extract_results(invocation, logger)
 
-    def check_abort_file(self, invocation: TestInvocation) -> bool:
-        """
-        Check for an abort file created by tmt-abort
-
-        Returns whether an abort file is present (i.e. abort occurred).
-        """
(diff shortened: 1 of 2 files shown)


therazix (commented):
LGTM. Just one question. Shouldn't the `check_abort_file` method be removed from the `ExecutePlugin` class as it is no longer used?

happz:
@therazix absolutely, removed in e0d050cb

therazix (approved):

happz:
> Nice, I just wonder how it worked before, when there was:
> 
> ```
> if (abort or self.data.exit_first and
>     result.result not in (ResultOutcome.PASS, ResultOutcome.INFO)):
> ```
> 
> and now we're iterating `for result in invocation.results`. Was that a bug?

I believe it was a bug, because it was checking `result.result`, and `result` was whatever the loop saw as the last item, most likely `invocation.results[-1]`:

```python
                if abort:
                        # In case of aborted all results in list will be aborted
                        result.note = 'aborted'

...

                    logger.verbose(
                        f"{_format_duration(result)} {result.show()} [{progress}]",
                        shift=shift)

...

                if (abort or self.data.exit_first and
                        result.result not in (ResultOutcome.PASS, ResultOutcome.INFO)):
```

So only the last result was considered, which does not seem right.

PR #3064, opened 2024-07-01
Add support for `kickstart` to `mrack` plugin

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema
* [x] mention the version
* [x] include a release note

### DIFF
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -74,6 +74,34 @@ if TYPE_CHECKING:
     from tmt._compat.typing import Self, TypeAlias
 
 
+def _normalize_user_data(
+        key_address: str,
+        raw_value: Any,
+        logger: tmt.log.Logger) -> dict[str, str]:
+    if isinstance(raw_value, dict):
+        return {
+            str(key).strip(): str(value).strip() for key, value in raw_value.items()
+            }
+
+    if isinstance(raw_value, (list, tuple)):
+        user_data = {}
+
+        for datum in raw_value:
+            try:
+                key, value = datum.split('=', 1)
+
+            except ValueError as exc:
+                raise tmt.utils.NormalizationError(
+                    key_address, datum, 'a KEY=VALUE string') from exc
+
+            user_data[key.strip()] = value.strip()
+
+        return user_data
+
+    raise tmt.utils.NormalizationError(
+        key_address, value, 'a dictionary or a list of KEY=VALUE strings')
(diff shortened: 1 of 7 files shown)


tmt/utils/__init__.py
```diff
@@ -78,6 +78,34 @@
     from tmt._compat.typing import Self, TypeAlias
 
 
+def _normalize_user_data(
```
happz, line 81: By moving into `tmt.utils`:

* the function is no longer "private", used by `artemis` plugin and no one else, therefore it should no longer start with an underscore,
* "user data" is a concept understood by Artemis and therefore by `artemis`, but it does not mean much to the rest of the code base. I would suggest finding a different name and adding a docstring. It was clear what was happening when it was in `tmt.steps.provision.artemis`, but what is `tmt.utils._normalize_user_data` doing? tmt does not accept any "user data"...

skycastlelily:
> I would suggest finding a different name and adding a docstring

Thanks for your mentor,How about"normalize_dict"? :)

happz:
> I would suggest finding a different name and adding a docstring
> Thanks for your mentor,How about"normalize_dict"? :)
> […](#)

Still way too generic, I'm afraid, it can mean anything, any input, any type of keys and values, who knows, it's very open. We need a name that reflects it is normalizing `key=value` inputs, and strings on both sides, sometimes the value would be optional...

skycastlelily:
Updated^^

happz:
@skycastlelily I tried to polish it a bit, I was not very happy about `beaker` field, that leaks Mrack's weird API into our code, so `to_mrack()` now exists. Check https://github.com/teemtee/tmt/pull/3064/commits/466b5686d9d73b3e8c1b591289be8ebc372983e1, WDYT?

I also change the name of that normalization callback, to `normalize_string_dict`, and I'm pretty happy about it, because it's what it does, normalizes two different forms of key/value inputs into a `dict[str, str]` :)

skycastlelily:
> WDYT?

I  really think it's brilliant, thanks do me the favor to co-author this
merge request and make the code look pretty while I'm on vacation^^

therazix (approved):

psss (approved):
Nice, thanks for implementing this!

psss:
Updated spec and added a release note in 5220d375.

skycastlelily:
Thanks:)

PR #3071, opened 2024-07-02
Add `install` prepare plugin using `install` into tmt try

falconizmi:
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] mention the version
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -132,12 +132,8 @@ the ``clean`` subcommand. Users can clean resources from the last
 run or from a run with a given id. Users can also choose to skip
 cleaning ``guests``, ``runs`` or ``images``.
 
-<<<<<<< HEAD
-``tmt try`` now supports :ref:`/stories/cli/try/option/install` option backed by :ref:`prepare/feature<plugins/prepare/install>` plugin
-=======
 ``tmt try`` now supports :ref:`/stories/cli/try/option/install` option
-backed by :ref:`prepare/feature<plugins/prepare/install>` plugin
->>>>>>> 89de3a2a (Adjust text length to match the header)
+backed by :ref:`prepare/feature</plugins/prepare/install>` plugin
 
 
 tmt-1.34.0
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -132,8 +132,7 @@ the ``clean`` subcommand. Users can clean resources from the last
 run or from a run with a given id. Users can also choose to skip
 cleaning ``guests``, ``runs`` or ``images``.
 
-``tmt try`` now supports :ref:`/stories/cli/try/option/epel` option.
-``tmt try`` now supports :ref:`/stories/cli/try/option/install` option.
+``tmt try`` now supports :ref:`/stories/cli/try/option/install` option backed by :ref:`prepare/feature<plugins/prepare/install>` plugin
 
 
 tmt-1.34.0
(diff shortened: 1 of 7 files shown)


lukaszachy:
Isn't it more natural to call that option install? As in `tmt try --install zsh fedora@container....`

docs/releases.rst
```diff
@@ -84,6 +84,9 @@ the ``clean`` subcommand. Users can clean resources from the last
 run or from a run with a given id. Users can also choose to skip
 cleaning ``guests``, ``runs`` or ``images``.
 
+``tmt try`` now supports :ref:`/stories/cli/try/option/install` option
```
happz, line 87: Needs to be moved to the right release.

stories/cli/try.fmf
```diff
@@ -146,3 +146,18 @@ link:
...
           - implemented-by: /tmt/trying.py
           - implemented-by: /tmt/steps/prepare/feature.py
+
+    /install:
+        story:
+            As a user I want an easy way to install local rpm package on the guest.
+        description: |
+            Install local rpm package on the guest.
+
+            .. versionadded:: 1.35
```
happz, line 156: Needs to be updated just like the release note.

happz (changes requested):

stories/cli/try.fmf
falconizmi: Done

docs/releases.rst
falconizmi: Done

happz: Is it? I see it in 1.35.0 release section, shouldn't it be in the upcoming one, 1.37.0?

falconizmi: Hm, I believe that I changed that .. maybe due some rebase from other branch or change in format of release number it went back to 1.35 .. for now its fixed

docs/releases.rst
```diff
@@ -31,6 +31,8 @@ execution. Use ``tmt run discover -vvv`` to see the details.
 
 ``tmt try`` now supports :ref:`/stories/cli/try/option/epel` option
 backed by :ref:`prepare/feature</plugins/prepare/feature>` plugin.
+``tmt try`` now supports :ref:`/stories/cli/try/option/install` option
```
happz, line 34:
```suggestion

``tmt try`` now supports :ref:`/stories/cli/try/option/install` option
```

happz (approved):

tmt/cli.py
```diff
@@ -1752,6 +1752,9 @@ def init(
 @option(
     "--epel", is_flag=True, default=False,
     help="Enable epel repository.")
+@option(
+    "--install", default=[], metavar="PACKAGE", multiple=True,
+    help="Install local rpm package on the guest.")
```
psss: It's not `local` packages only, right? We can install packages from repositories as well. Plus, the `install` plugin supports non-rpm packages as well.

psss (approved):
Thanks for implementing this! Looks good. In 1b7d1f59 I've just adjusted the option description (we support non-local and non-rpm packages as well) and the release note a bit.

docs/releases.rst
psss: Seems to be covered.

tmt/cli.py
psss: Fixed in 1b7d1f59.

PR #3076, opened 2024-07-03
Coerce dimensionless values of SizeConstraint to values with dimensions

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tests/unit/test_hardware.py
+++ b/tests/unit/test_hardware.py
@@ -67,6 +67,26 @@ def test_constraint_name_pattern(value: str, expected: tuple[Any, Any]) -> None:
     assert match.groups() == expected
 
 
+_size_constraint_pattern_input = [
+    ({'name': 'num_with_default', 'raw_value': '10', 'default_unit': 'GiB'},
+     'num_with_default: == 10 gibibyte'),
+    ({'name': 'num_without_default', 'raw_value': '1024'}, 'num_without_default: == 1024 byte'),
+    ({'name': 'num_with_unit', 'raw_value': '10 GiB', 'default_unit': 'GiB'},
+     'num_with_unit: == 10 GiB'),
+    ]
+
+
+@pytest.mark.parametrize(
+    ('value', 'expected'),
+    _size_constraint_pattern_input,
+    )
+def test_constraint_default_unit(value: dict, expected: tuple[Any, Any]) -> None:
+    constraint_out = tmt.hardware.SizeConstraint.from_specification(**value)
+
+    assert constraint_out is not None
+    assert str(constraint_out) == expected
+
+
 _constraint_components_pattern_input = [
     ('memory 10 GiB', ('memory', None, None, None, '10 GiB')),
     ('cpu.processors != 4 ', ('cpu', None, 'processors', '!=', '4')),
--- a/tests/unit/test_hardware.py
+++ b/tests/unit/test_hardware.py
@@ -71,7 +71,7 @@ _size_constraint_pattern_input = [
     ({'name': 'num_with_default', 'raw_value': '10', 'default_unit': 'GiB'},
      'num_with_default: == 10 gibibyte'),
     ({'name': 'num_without_default', 'raw_value': '1024'}, 'num_without_default: == 1024 byte'),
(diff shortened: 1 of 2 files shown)


tmt/hardware.py
```diff
@@ -672,8 +672,13 @@ def from_specification(
...
             raw_value: str,
             original_constraint: Optional['Constraint[Any]'] = None,
-            allowed_operators: Optional[list[Operator]] = None
+            allowed_operators: Optional[list[Operator]] = None,
+            unit: Optional[Any] = 'bytes'
             ) -> T:
+        if not isinstance(
+                UNITS(raw_value),
+                pint.Quantity):  # type: ignore[reportUnnecessaryIsInstance,unused-ignore]
+            raw_value = str(pint.Quantity(raw_value, unit))
```
happz, line 681: I would recommend doing it *after* `_from_specification()` call, inspecting the returned constraint. Doing it *before* has the disadvantage of various input types, for example, your test will not trigger for `memory: "1024"`, because `raw_value` would be a string. On the other hand, *after* the call, the `value` attribute of constraint returned by `_from_specification()` would always be a `Quantity` instance, and much safer to work with, with no exceptions.

skycastlelily:
Updated^^

tmt/hardware.py
```diff
@@ -507,7 +507,8 @@ def _from_specification(
             as_quantity: bool = True,
             as_cast: Optional[Callable[[str], ConstraintValueT]] = None,
             original_constraint: Optional['Constraint[Any]'] = None,
-            allowed_operators: Optional[list[Operator]] = None
+            allowed_operators: Optional[list[Operator]] = None,
+            unit: Optional[Any] = "bytes"
```
happz, line 511: How about calling it a `default_unit`? Seeing `unit`, one might suspect there will be some conversion performed, to enforce this unit, which is not what's going to happen.

happz:
@skycastlelily please, add also a unit test, a small one should be more than enough. A simple call to `SizeConstraint.from_specification` & checking the magnitude and unit of the returned constraint value should work. I would make it parametrized, with raw value and default unit as inputs and magnitude and unit as expected values, because we want to check various combination - number without a default value (`bytes` are added), number with a default value but not `bytes` (the default value is used, not bytes), number with unit already included (`4 GB`) (default value is ignored), and maybe there are others.

skycastlelily:
Updated^^

tests/unit/test_hardware.py
```diff
@@ -67,6 +67,26 @@ def test_constraint_name_pattern(value: str, expected: tuple[Any, Any]) -> None:
     assert match.groups() == expected
 
 
+_size_constraint_pattern_input = [
+    ({'name': 'num_with_default', 'raw_value': '10', 'default_unit': 'GiB'},
+     'num_with_default: == 10 gibibyte'),
+    ({'name': 'num_without_default', 'raw_value': '1024'}, 'num_without_default: == 1024 byte'),
+    ({'name': 'num_with_unit', 'raw_value': '10 GiB', 'default_unit': 'GiB'},
```
therazix, line 74: I would suggest using a different `default_unit` to make it clear that the unit from `raw_value` is being used.
```suggestion
    ({'name': 'num_with_unit', 'raw_value': '10 GiB', 'default_unit': 'MiB'},
```

tmt/hardware.py
```diff
@@ -507,7 +507,8 @@ def _from_specification(
             as_quantity: bool = True,
             as_cast: Optional[Callable[[str], ConstraintValueT]] = None,
             original_constraint: Optional['Constraint[Any]'] = None,
-            allowed_operators: Optional[list[Operator]] = None
+            allowed_operators: Optional[list[Operator]] = None,
+            default_unit: Optional[Any] = "bytes"
```
therazix, line 511: Just a minor suggestion. Could you also add a docstring for the `default_unit`, as every other parameter has one as well?

skycastlelily:
Updated:)

therazix (approved):

happz (approved):

happz:
Usual suspects broke tests once again, but it's nothing related to the patch. Merging.

PR #3084, opened 2024-07-10
Fix collection of know keys for test, plan and story T001 linter

happz:
The original code did not handle fields with dashes (`restart-max-count`) and internal fields (`serial-number`).

Pull Request Checklist

* [x] implement the feature


psss (approved):
Thanks for the improvement!

psss:
Extended the test coverage a bit in 6a69b436 to check that internal keys are correctly reported as invalid (which was not the case before the change).

PR #3091, opened 2024-07-17
Skip testing debuginfo install on coreos images

psss:
This does not work, it's not an important use case and we currently don't have time to investigate this.

### DIFF
--- a/tests/unit/test_package_managers.py
+++ b/tests/unit/test_package_managers.py
@@ -1733,7 +1733,19 @@ def _parametrize_test_install_debuginfo() -> Iterator[
             Optional[str]]]:
 
     for container, package_manager_class in CONTAINER_BASE_MATRIX:
-        if package_manager_class is tmt.package_managers.dnf.Dnf5:
+        # Skip testing debuginfo install on coreos images
+        if 'coreos' in container.url:
+            yield pytest.param(
+                container,
+                package_manager_class,
+                (Package('tree'), Package('dos2unix')),
+                "",
+                None,
+                None,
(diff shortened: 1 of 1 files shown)


happz (approved):
@psss would it make sense to add a "support matrix" to `prepare/install` docs? It starts to seem there are some combinations that are not supported, both yet and never, it might be nice to document them. On the other hand, there's still the all-mighty-install plugin that would take this over one day, so maybe there's no point in documenting this particular bit for `prepare/install`. If you think it's worth the effort, I'd propose an issue to have it tracked for future PR.

Maybe we could encode and generate the matrix from package manager info/docstrings/classes, to have it always up-to-date. I was also thinking of moving parameters (package, expected message, ...) from this test file into respective package manager implementations, to keep them close to code, because the test file has grown a lot - adding the info for "X supported/unsupported" matrix there as well might be nice to keep all info at one place. What do you think?

thrix (approved):
LGTM

psss:
> @psss would it make sense to add a "support matrix" to `prepare/install` docs? It starts to seem there are some combinations that are not supported, both yet and never, it might be nice to document them.

Yeah, sounds good to me. Probably just some simple form would be sufficient. Not spend too much time on that, 'cause, as you mentioned, the almighty install will deprecated that.

> I was also thinking of moving parameters (package, expected message, ...) from this test file into respective package manager implementations, to keep them close to code

This also is quite reasonable, it took me a bit of time to find the right place for the exception, so +1 from me.

psss:
Oh! Green tests! Finally. Haven't seen this for quite some time! ;-)

PR #3092, opened 2024-07-18
Retry on error 504 from ReportPortal

kkaarreell:
Introduces two changes to the reportportal plugin:

- enable request retry on HTTP error 504
- strip unprintable characters from uploaded logs as those were causing errors on RP side

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -36,6 +36,13 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
     return str(os.getenv(env_var))
 
 
+def _filter_invalid_chars(data: str) -> str:
+    return re.sub(
+        '[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+',
+        '',
+        data)
+
+
 @dataclasses.dataclass
 class ReportReportPortalData(tmt.steps.report.ReportStepData):
 
@@ -231,7 +238,8 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
 
         if not response.ok:
             raise tmt.utils.ReportError(
-                f"Received non-ok status code {response.status_code} from ReportPortal: {response.text}")
+                f"Received non-ok status code {response.status_code} "
+                f"from ReportPortal: {response.text}")
 
     def check_options(self) -> None:
         """ Check options for known troublesome combinations """
@@ -534,7 +542,7 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
                         response = session.post(
                             url=f"{self.get_url()}/log/entry",
                             headers=self.get_headers(),
-                            json={"message": log,
+                            json={"message": _filter_invalid_chars(log),
                                   "itemUuid": item_uuid,
                                   "launchUuid": launch_uuid,
                                   "level": level,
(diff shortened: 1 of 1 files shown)


Athwale:
I can confirm this works with using the modified file in my run.
```
Read file '/var/tmp/tmt/run-061/plans/plan-1/execute/data/guest/default-0/rhel_tcpdump/Security/tcpdump-Buffer-over-read-in-function-20/output.txt'.
        Response code from the endpoint: 504
        Message from the endpoint:
            <html><body><h1>504 Gateway Time-out</h1>
            The server didn't respond in time.
            </body></html>
        Ignoring RP timeout response 
    Read file '/var/tmp/tmt/run-061/plans/plan-1/execute/data/guest/default-0/rhel_tcpdump/Security/tcpdump-Buffer-over-read-in-function-20/journal.txt'.
```
The results from this failed report were uploaded to RP anyway.

tmt/steps/report/reportportal.py
```diff
@@ -226,12 +226,15 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
     def handle_response(self, response: requests.Response) -> None:
         """ Check the endpoint response and raise an exception if needed """
 
-        if not response.ok:
+        self.debug("Response code from the endpoint", response.status_code)
+        self.debug("Message from the endpoint", response.text)
+
+        if response.status_code == 504:
+            self.debug("Ignoring RP timeout response")
```
happz, line 233: Please, make it a warning. IIUIC, the user runs tests, the user expects results in RP, but the results are not there, the user turns to the TF log and sees nothing out of the ordinary because the message is in `log.txt` as a mere debug note. On the other hand, a warning in tmt output reporting there was a timeout and the submission was not repeated would provide an immediate hint.

And a question: is it really safe to not repeat the request? At least once or twice, before giving up? Seems like the results are not that important after all to me when the plugin gives up on uploading them after the first hurdle.

kkaarreell: According to my observation, data are in RP despite the timeout. Retry is worth trying, hopefully it won't duplicate test results in RP.

tmt/steps/report/reportportal.py
```diff
@@ -393,7 +394,14 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
 
         # Communication with RP instance
-        with tmt.utils.retry_session() as session:
+        with tmt.utils.retry_session(status_forcelist=(
+                    429,   # Too Many Requests
+                    500,   # Internal Server Error
+                    502,   # Bad Gateway
+                    503,   # Service Unavailable
+                    504,   # Gateway Timeout
+                    5000,  # Unclassified error - weird error from RP appearing sometimes
```
thrix, line 403: `5000` is not an http error?

kkaarreell: right, I thought it gets propagated to the log message but it is not. I will try to reproduce the issue to figure it out.

kkaarreell: So that was a 500 error in the end.

tmt/steps/report/reportportal.py
kkaarreell: So for now we have updated the `status_forcelist` with 504.

4N0body5:
I can confirm, this resolves the following problem:
```
Received non-ok status code from ReportPortal: {"errorCode":5000,"message":"Unclassified error [org.hibernate.exception.DataException: could not execute statement]"}
```

Tested on another reproducer - reporting [scrub test](https://gitlab.cee.redhat.com/special-projects/tests/scrub/-/blob/master/Sanity/bz907173-RFE-Add-option-for-scrub-utility-to-only-scrub/runtest.sh).

lukaszachy (approved):

happz (approved):

happz:
Unrelated failure, merging.

PR #3093, opened 2024-07-18
Mention the `include-output-log` key in release notes

psss:
It's worth to mention the option newly added in #3003 in the release notes as well.

Pull Request Checklist

* [ ] include a release note


KwisatzHaderach (approved):
somehow I thought I had a release note, thanks for creating it

PR #2933, opened 2024-05-15
Extend results schema to allow test phases

happz:
Related to #2826.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] update the specification

Issue #2826, opened 2024-04-04
Add support for test phases


psss:
As discussed on the stakeholder meeting, there are use cases related to tracking individual results of test "phases". These can be created using one of the following ways:

* defining a `beakerlib` phase using `rlPhaseStart` and `rlPhaseEnd`
* explicitly calling `rstrnt-report-result` to submit a subresult

When run in `Beaker`, the `rlPhaseEnd` command called during a `beakerlib` test execution actually calls `rlReport` which uses `BEAKERLIB_COMMAND_REPORT_RESULT` command to report the result and, if I'm not mistaken, this defaults to using `rstrnt-report-result`. @sopos, please correct me if I'm wrong.

The effect is that each `beakerlib` phase and `rstrnt-report-result` call create a "subresult" which is shown in the `Beaker` web interface under the actual test. The proposal is to add a similar support for subresults to `tmt`.

Here are some more ideas caught from the brainstorm:

* We could inject a custom script using `BEAKERLIB_COMMAND_REPORT_RESULT` and perform necessary actions on the guest
* The result specification could be extended to include a list of subresults (which would be similar to the `check` attribute)
* It would be good to report detected `avc` failures for each test phase separately
* Report plugins would have to invent the proper mapping to e.g. `junit` format as the spec does not know subresults

All in all it seems this would bring a substantial level of complexity (currently the test granularity was simply defined by the test metadata definition for each test) so we should discuss this well before jumping into the implementation.

### Base Work

* spec — https://github.com/teemtee/tmt/pull/2933
* class — https://github.com/teemtee/tmt/pull/3094
* https://github.com/teemtee/tmt/pull/3106
* actual saving of subresults into `results.yaml` - https://github.com/teemtee/tmt/pull/3200

### Report Plugins

* display — https://github.com/teemtee/tmt/pull/3107
* html — https://github.com/teemtee/tmt/pull/3095
* junit
    *  https://github.com/teemtee/tmt/pull/3150
    * https://github.com/teemtee/tmt/pull/3307
    * https://github.com/teemtee/tmt/pull/3177
    * https://github.com/teemtee/tmt/pull/3365
    * https://github.com/teemtee/tmt/pull/3382
* polarion — https://github.com/teemtee/tmt/pull/3166 (not actually related to subresults)
* reportportal — https://github.com/teemtee/tmt/pull/3331
    * https://github.com/teemtee/tmt/pull/3356
    * https://github.com/teemtee/tmt/pull/3370
[...]

PR #3094, opened 2024-07-18
Propagate result phase into the report plugins

seberm:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

This is a prerequisite for report plugins to support "subresults" aka **phases**.

Related to:
- https://github.com/teemtee/tmt/pull/2933
- https://github.com/teemtee/tmt/pull/2977

### DIFF
--- a/tmt/result.py
+++ b/tmt/result.py
@@ -136,6 +136,29 @@ class CheckResult(BaseResult):
 
 
 @dataclasses.dataclass
+class PhaseCheckResult(CheckResult):
+    """
+    Describes what tmt knows about a single phase test check result.
+
+    It does not contain any additional fields; it simply defines a type to
+    easily differentiate between a :py:class:`tmt.result.CheckResult` and a
+    ``CheckResult`` located within a result phase.
+    """
+
+
+@dataclasses.dataclass
+class PhaseResult(BaseResult):
+    """ Describes what tmt knows about a single test phase result (sub-result) """
+
+    check: list[PhaseCheckResult] = field(
+        default_factory=cast(Callable[[], list[PhaseCheckResult]], list),
+        serialize=lambda results: [result.to_serialized() for result in results],
+        unserialize=lambda serialized: [
+            PhaseCheckResult.from_serialized(check) for check in serialized]
+        )
+
+
+@dataclasses.dataclass
 class Result(BaseResult):
     """ Describes what tmt knows about a single test result """
 
@@ -157,6 +180,13 @@ class Result(BaseResult):
         default_factory=ResultGuestData,
         serialize=lambda value: value.to_serialized(),
         unserialize=lambda serialized: ResultGuestData.from_serialized(serialized)
+        )
+
+    phase: list[PhaseResult] = field(
+        default_factory=cast(Callable[[], list[PhaseResult]], list),
+        serialize=lambda results: [result.to_serialized() for result in results],
+        unserialize=lambda serialized: [
+            PhaseResult.from_serialized(phase) for phase in serialized]
         )
 
     check: list[CheckResult] = field(


psss:
This one is really trivial, proposing to squeeze into `1.35`.

tmt/result.py
```diff
@@ -125,6 +125,11 @@ def show(self) -> str:
         return ' '.join(components)
 
 
+@dataclasses.dataclass
+class PhaseResult(BaseResult):
+    """ Describes what tmt knows about a single test phase result (sub-result) """
+
```
happz, line 131: A phase result should include also `check` field, see the specification at https://tmt.readthedocs.io/en/stable/spec/plans.html#results-format - AFAICT, it’s the only extra field beyond the base result.

psss: So... we will store check results for individual phases, right, makes sense, good point!

seberm: @happz thanks for the catch. Hopefully, it should be resolved now.

happz: Correct, per-phase check results are one of the requirements, the one that forces us into reworking checks to be able to materialize as scripts on guests between phases.

seberm:
Hello @happz @psss ,
I am still thinking about a tiny modification - I think we probably should create a new type for phase checks, e.g. `PhaseCheckResult(CheckResult)`. This would allow us to differentiate between the result checks and the result **phase checks**, especially in the report plugins.

Do you agree?

happz:
> Hello @happz @psss , I am still thinking about a tiny modification - I think we probably should create a new type for phase checks, e.g. `PhaseCheckResult(CheckResult)`.

What would be the difference between `CheckResult` and the new `PhaseCheckResult` class? The existing classes usually add new fields on top of their base class: `CheckResult` adds `event`, `Result` adds `context` and `serial_number` and so on, so what new fields would `PhaseCheckResult` carry that `Check` lacks?

> This would allow us to easily differentiate between the result checks and the result **phase checks**, especially in the report plugins.

Test check results live in the `check` list, and phase check results live in the `phases[N].check` list. The structure, the fields describing the check outcome are the same, the position provides the context.

> Do you agree?

Not as of now. Can you demonstrate in what way would different classes improve the code, e.g. making it simpler, easier to read and maintain, or easy to check for linters?

seberm:
> The existing classes usually add new fields on top of their base class: CheckResult adds event, Result adds context and serial_number and so on, so what new fields would PhaseCheckResult carry that Check lacks?

It would carry no new fields, just a type.

> Not as of now. Can you demonstrate in what way would different classes improve the code, e.g. making it simpler, easier to read and maintain, or easy to check for linters?

It would allow me to easily differentiate between the result types in `display_outcome(result: BaseResult) -> None` and use this method to show all kinds of results. It's just an early draft:
- https://github.com/teemtee/tmt/commit/d10ebe3c8b214fabcb1577b14f655d3e13205e60#diff-f9ad5f6605f10698f0208842ba0e42ee7bf1508151827baf78c3660b611e4457R84

If you still think this does not make sense, I will try to find a different solution.

happz:
> > The existing classes usually add new fields on top of their base class: CheckResult adds event, Result adds context and serial_number and so on, so what new fields would PhaseCheckResult carry that Check lacks?
> 
> It would carry no new fields, just a type.

Ack.

> 
> > Not as of now. Can you demonstrate in what way would different classes improve the code, e.g. making it simpler, easier to read and maintain, or easy to check for linters?
> 
> It would allow me to easily differentiate between the result types in `display_outcome(result: BaseResult) -> None` and use this method to show all kinds of results. It's just an early draft:
> 
> * [d10ebe3#diff-f9ad5f6605f10698f0208842ba0e42ee7bf1508151827baf78c3660b611e4457R84](https://github.com/teemtee/tmt/commit/d10ebe3c8b214fabcb1577b14f655d3e13205e60#diff-f9ad5f6605f10698f0208842ba0e42ee7bf1508151827baf78c3660b611e4457R84)

I see, makes sense. Let's try it then and see how it goes.

thrix (approved):
LGTM

KwisatzHaderach (approved):

happz:
Apparently unrelated failures in tests, going to merge this one and we'll see.

Issue #2392, opened 2023-10-11
hardware: provision system with iommu enabled


snits:
With beaker it is possible using the key_value element to find systems that have an iommu enabled (currently for x86, and in the near future aarch64 as well) by looking to for the VIRT_IOMMU key having a value of one.

Example of beaker xml to find a system with the setting, here looking for an Intel system with more than 1 disk and an iommu enabled:

```
<hostrequires>
  <and>
    <key_value op="=" key="VIRT_IOMMU" value="1"/>
    <key_value op=">" key="NR_DISKS" value="1"/>
    <cpu>
      <!-- GenuineIntel, AuthenticAMD -->
      <vendor op="==" value="GenuineIntel"/>
    </cpu>
    <system>
      <hypervisor op="=" value=""/>
      <has_console op="=" value="True"/>
    </system>
  </and>
</hostrequires>
```

Currently this key is created during system inventory by beaker-system-scan. beaker-system-scan, depending on the system architecture, looks for the relevant acpi table (DMAR for Intel, IVRS for AMD, and IORT for ARM) and does a sanity check to see that it has an expected entry. The main reason behind this is because by default the Intel IOMMU kernel module is not enabled, so by checking the acpi tables it can be seen that it is enabled in the bios. For AMD and ARM it would probably be simpler to look and see that the /sys/kernel/iommu_groups directly is populated. That would also work for checking an Intel based system, if it is booted with intel_iommu=on. 

Another thought though, is there currently a way for someone to provision a guest that has a device passed through from the host, or a host capable of that? If yes, then the key might not be of use to someone beyond me or needed. If there already is a way to find a host capable of providing that support to a guest that would probably work for me if I can provision the host.

happz:
Seems like a low-hanging fruit:

* `beaker` would transform this into `key_value` element
* `artemis` would do the same - we should check whether this is something AWS/OpenStack may deliver, but Beaker would
* `virtual` might be able to construct corresponding VM, if there's a support for such a parameter
* the rest - `container`, `local` - would simply emit the expected "sorry, no can do" warning and proceed.

qiankehan:
For virtual.libvirt, there is 3 types of [emulated iommu devices](https://libvirt.org/formatdomain.html#iommu-devices): 
[...]

PR #3100, opened 2024-07-21
Add the `iommu` specification for hardware

qiankehan:
Pull Request Checklist

* [x]  implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] mention the version
* [ ] include a release note

### DIFF
--- a/spec/hardware/iommu.fmf
+++ b/spec/hardware/iommu.fmf
@@ -1,0 +1,32 @@
+summary: |
+    Select or provision a guest with the `Input–output memory management unit`.
+
+description: |
+    .. code-block::
+
+       iommu:
+           # Boolean, whether the guest supports IOMMU
+           is-supported: true|false
+           # String, the model name for IOMMU
+           model-name: "virtio"|"intel"|"smmuv3"|...
+
+    .. versionadded:: 1.37
+
+    .. versionchanged:: 1.37
+       ``beaker`` plugin supports ``iommu``
+
+example:
+  - |
+    # Require a guest that supports IOMMU
+    iommu:
+        is-supported: true
+
+  - |
+    # Require a guest with the IOMMU of virtio model
+    iommu:
+        is-supported: true
+        model-name: virtio
+
+link:
+    - implemented-by: /tmt/steps/provision/mrack.py
+      note: "``iommu.is-supported`` only"
--- a/tests/unit/provision/mrack/test_hw.py
+++ b/tests/unit/provision/mrack/test_hw.py
@@ -10,6 +10,7 @@ from tmt.hardware import (
     _parse_cpu,
     _parse_disk,
     _parse_hostname,
+    _parse_iommu,
     _parse_location,
     _parse_memory,
     _parse_system,
@@ -783,6 +784,35 @@ def test_zcrypt_mode(root_logger: Logger) -> None:
                     '_op': 'like',
                     '_value': 'C%A'
                     }
+                }
+            }
+        }
+
+
+def test_iommu_is_supported(root_logger: Logger) -> None:
+
+    result = _CONSTRAINT_TRANSFORMERS['iommu.is_supported'](
+        _parse_iommu({"is-supported": True}), root_logger)
+
+    assert result.to_mrack() == {
+        'system': {
+            'key_value': {
+                '_key': 'VIRT_IOMMU',
+                '_op': '==',
+                '_value': '1'
+                }
+            }
+        }
+
+    result = _CONSTRAINT_TRANSFORMERS['iommu.is_supported'](
+        _parse_iommu({"is-supported": False}), root_logger)
+
+    assert result.to_mrack() == {
+        'system': {
+            'key_value': {
+                '_key': 'VIRT_IOMMU',
+                '_op': '==',
+                '_value': '0'
                 }
             }
         }
--- a/tests/unit/provision/mrack/test_hw.py
+++ b/tests/unit/provision/mrack/test_hw.py
@@ -791,31 +791,18 @@ def test_zcrypt_mode(root_logger: Logger) -> None:
 
 def test_iommu_is_supported(root_logger: Logger) -> None:
 
-    result = _CONSTRAINT_TRANSFORMERS['iommu.is_supported'](
-        _parse_iommu({"is-supported": True}), root_logger)
+    for value in True, False:
 
-    assert result.to_mrack() == {
-        'system': {
+        result = _CONSTRAINT_TRANSFORMERS['iommu.is_supported'](
+            _parse_iommu({"is-supported": value}), root_logger)
+
+        assert result.to_mrack() == {
             'key_value': {
                 '_key': 'VIRT_IOMMU',
                 '_op': '==',
-                '_value': '1'
+                '_value': str(int(value))
                 }
             }
-        }
-
-    result = _CONSTRAINT_TRANSFORMERS['iommu.is_supported'](
-        _parse_iommu({"is-supported": False}), root_logger)
-
-    assert result.to_mrack() == {
-        'system': {
-            'key_value': {
-                '_key': 'VIRT_IOMMU',
-                '_op': '==',
-                '_value': '0'
-                }
-            }
-        }
 
 
 def test_location_lab_controller(root_logger: Logger) -> None:
--- a/tmt/schemas/provision/hardware.yaml
+++ b/tmt/schemas/provision/hardware.yaml
@@ -355,6 +355,20 @@ definitions:
     # empty `boot`.
     minProperties: 1
 
+  # Hw requirements `iommu` block
+  iommu:
+    type: object
+
+    properties:
+      is-supported:
+        type: boolean
+
+      model-name:
+        type: string
+
+    additionalProperties: false
+    minProperties: 1
+
   # HW requirements: basic block
   block:
     type: object
@@ -404,6 +418,9 @@ definitions:
 
       virtualization:
         "$ref": "#/definitions/virtualization"
+
+      iommu:
+        "$ref": "#/definitions/iommu"
 
     additionalProperties: false
 
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -512,6 +512,21 @@ def _transform_zcrypt_mode(
         children=[MrackHWKeyValue('ZCRYPT_MODE', beaker_operator, actual_value)])
 
 
+def _transform_iommu_is_supported(
+        constraint: tmt.hardware.FlagConstraint,
+        logger: tmt.log.Logger) -> MrackBaseHWElement:
+
+    test = (constraint.operator, constraint.value)
+
+    if test in [(tmt.hardware.Operator.EQ, True), (tmt.hardware.Operator.NEQ, False)]:
+        return MrackHWKeyValue('VIRT_IOMMU', '==', '1')
+
+    if test in [(tmt.hardware.Operator.EQ, False), (tmt.hardware.Operator.NEQ, True)]:
+        return MrackHWKeyValue('VIRT_IOMMU', '==', '0')
+
+    return _transform_unsupported(constraint, logger)
+
+
 def _transform_location_lab_controller(
         constraint: tmt.hardware.TextConstraint,
         logger: tmt.log.Logger) -> MrackBaseHWElement:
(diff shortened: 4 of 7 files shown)


spec/hardware/iommu.fmf
```diff
@@ -0,0 +1,26 @@
+summary: |
+    Select or provision a guest with the `Input–output memory management unit`.
+
+description: |
+    .. code-block::
+
+       iommu:
+           # Boolean, whether the guest supports IOMMU
+           is-supported: "x"
```
happz, line 9: A boolean key should be documented with `true|false` value, see e.g. https://github.com/teemtee/tmt/blob/main/spec/hardware/virtualization.fmf#L18

spec/hardware/iommu.fmf
```diff
@@ -0,0 +1,26 @@
...
+           # String, the model name for virtualizated IOMMU
+           model: "virtio"|"intel"|"smmuv3"
```
happz, line 11: First, I wouldn't say "virtualized". The model is simply the one requested - it may be satisfied by an actual bare-metal machine, correct? `virtual` plugin may spawn a VM with the requested IOMMU model, but `beaker` may be able to find a machine with the specified model.

Second, I would probably drop the list from specification, or at least make it an open-ended list, see https://github.com/teemtee/tmt/blob/main/spec/hardware/virtualization.fmf#L15. There may be more models or aliases, let's not suggests these are the only ones possible.

spec/hardware/iommu.fmf
```diff
@@ -0,0 +1,26 @@
...
+
+    .. versionadded:: 1.35
+
+example:
+  - |
+    # Require a guest that supports IOMMU
+    iommu:
+        is-supported: true|false
```
happz, line 19: I would expect `true` instead of `true|false` - if the example is showing me "this is how to request a guest that supports IOMMU", `false` would tell me how to request one that does *not* support it.

spec/hardware/iommu.fmf
qiankehan: > First, I wouldn't say "virtualized". The model is simply the one requested - it may be satisfied by an actual bare-metal machine, correct? `virtual` plugin may spawn a VM with the requested IOMMU model, but `beaker` may be able to find a machine with the specified model.

Here I intended to say: the model is only for the plugins like `virtual` or `artemis.openstack`. It will not work for `beaker`. Is it OK to comment that here?
> 
> Second, I would probably drop the list from specification, or at least make it an open-ended list, see https://github.com/teemtee/tmt/blob/main/spec/hardware/virtualization.fmf#L15. There may be more models or aliases, let's not suggests these are the only ones possible.

Good idea~

happz: > > First, I wouldn't say "virtualized". The model is simply the one requested - it may be satisfied by an actual bare-metal machine, correct? `virtual` plugin may spawn a VM with the requested IOMMU model, but `beaker` may be able to find a machine with the specified model.
> 
> Here I intended to say: the model is only for the plugins like `virtual` or `artemis.openstack`. It will not work for `beaker`. Is it OK to comment that here?

I don't think so, because AFAICT, it's not limited to `virtual` or `openstack`, AWS and Azure can also deliver machines with IOMMU support, and I bet the list continues. And they may use the same model names, or even add new ones, so the `model` key is not that pinned to these plugins only.

In general, the specification should remain infrastructure-agnostic, especially when it comes to genuine HW requirements. It's slightly different when speaking about not-so-much-HW-but-still-affecting-provisioning requirements, like job delegation in Beaker, as no other infrastructure has this particular feature, but IOMMU is not that specific. It can be delivered by multiple infrastructures, at least in theory, we don't have support for it yet, but it's possible, and therefore it's not limited to just `virtual` and `beaker`.

We use `link` to point to plugins that actually implement and support the requirement, see e.g. https://github.com/teemtee/tmt/blob/main/spec/hardware/hostname.fmf#L23. For now, this list would be empty, the patch adding implementation to `beaker` plugin would also update this list with `implemented-by` link, and then other plugins and so on.

tmt/schemas/provision/hardware.yaml
```diff
@@ -333,6 +333,20 @@ definitions:
     # empty `boot`.
     minProperties: 1
 
+  # Hw requirements `iommu` block
```
idorax, line 336: Looks `Hw` right here should be `HW`.

spec/hardware/iommu.fmf
```diff
@@ -0,0 +1,26 @@
+summary: |
+    Select or provision a guest with the `Input–output memory management unit`.
+
+description: |
+    .. code-block::
+
+       iommu:
+           # Boolean, whether the guest supports IOMMU
+           is-supported: "x"
+           # String, the model name for virtualizated IOMMU
```
idorax, line 10: `virtualizated` is a typo? should be `virtualized`?

spec/hardware/iommu.fmf
qiankehan: It is a typo here. Updated

qiankehan:
Thanks for your reviews~ I have updated the PR~

qiankehan:
Hi, I have updated the PR and add the beaker plugin support. Please help review it. Thank you~

spec/hardware/iommu.fmf
```diff
@@ -0,0 +1,31 @@
...
+    Select or provision a guest with the `Input–output memory management unit`.
+
+description: |
+    .. code-block::
+
+       iommu:
+           # Boolean, whether the guest supports IOMMU
+           is-supported: true|false
+           # String, the model name for IOMMU
+           model: "virtio"|"intel"|"smmuv3"|...
```
happz, line 11: I just realized we did not align this field with the rest of the specifications, see https://tmt.readthedocs.io/en/stable/spec/hardware.html#names-and-ids. This field should be called `model-name` as it's a label rather than a numerical ID.

spec/hardware/iommu.fmf
```diff
@@ -0,0 +1,31 @@
...
+        is-supported: true
+
+  - |
+    # Require a guest with the IOMMU of virtio model
+    iommu:
+        is-supported: true
+        model: virtio
+
+link:
+    - implemented-by: /tmt/steps/provision/mrack.py
```
happz, line 31: IIUIC, the patch adds support for `is-supported` field only to Beaker plugin, please add a similar note here: https://github.com/teemtee/tmt/blob/main/spec/hardware/system.fmf#L37

tmt/steps/provision/mrack.py
```diff
@@ -512,6 +512,28 @@ def _transform_zcrypt_mode(
...
 
 
+def _transform_iommu_is_supported(
+        constraint: tmt.hardware.FlagConstraint,
+        logger: tmt.log.Logger) -> MrackBaseHWElement:
+
+    test = (constraint.operator, constraint.value)
+
+    if test in [(tmt.hardware.Operator.EQ, True), (tmt.hardware.Operator.NEQ, False)]:
+        return MrackHWKeyValue("VIRT_IOMMU", '==', '1')
```
idorax, line 522: For `VIRT_IOMMU`, double quotation marks are used. But for `==` and `1`, single quotation marks are used. Please make them consistent.

tmt/steps/provision/mrack.py
```diff
@@ -512,6 +512,28 @@ def _transform_zcrypt_mode(
...
+
+    if test in [(tmt.hardware.Operator.EQ, False), (tmt.hardware.Operator.NEQ, True)]:
+        return MrackHWKeyValue("VIRT_IOMMU", '==', '0')
```
idorax, line 525: The same comment for L522.

martinhoyer (commented):
I'm not sure about the 'model' part. In beaker, I don't see a way to filter by  (maybe I'm missing something).    

While other provisioning methods could theoretically support this going forward, I wouldn't expose it to users if there is no use for it now.  
We can filter by cpu to at least get the intel/amd/arm.

happz:
> I'm not sure about the 'model' part. In beaker, I don't see a way to filter by (maybe I'm missing something).
> 
> While other provisioning methods could theoretically support this going forward, I wouldn't expose it to users if there is no use for it now. We can filter by cpu to at least get the intel/amd/arm.

IIUIC, libvirt and qemu-based plugins, like `virtual`, can and will make use of this requirement. And I would like to avoid encoding one requirement into another, i.e. using `cpu.vendor-name` to pick a model of IOMMU. That complicates reasoning about requirements and makes the picture opaque: do they use it to select a CPU, or something else we don’t support yet? It’s happening with `hostname` quite often.

More discussion in https://github.com/teemtee/tmt/issues/2392

martinhoyer:
> > I'm not sure about the 'model' part. In beaker, I don't see a way to filter by (maybe I'm missing something).
> > While other provisioning methods could theoretically support this going forward, I wouldn't expose it to users if there is no use for it now. We can filter by cpu to at least get the intel/amd/arm.
> 
> IIUIC, libvirt and qemu-based plugins, like `virtual`, can and will make use of this requirement. And I would like to avoid encoding one requirement into another, i.e. using `cpu.vendor-name` to pick a model of IOMMU. That complicates reasoning about requirements and makes the picture opaque: do they use it to select a CPU, or something else we don’t support yet? It’s happening with `hostname` quite often.
> 
> More discussion in #2392

Ok cool, from the discussion I got the impression it'll be only used by beaker provisioning.

tmt/steps/provision/mrack.py
```diff
@@ -512,6 +512,28 @@ def _transform_zcrypt_mode(
...
+
+    return _transform_unsupported(constraint, logger)
+
+
+def _transform_iommu_model(
+        constraint: tmt.hardware.TextConstraint,
+        logger: tmt.log.Logger) -> MrackBaseHWElement:
+
+    return _transform_unsupported(constraint, logger)
```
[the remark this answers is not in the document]
qiankehan: I think there is no need to parse **model-name** for `beaker`. I will delete the function.

qiankehan:
I have updated it as the comments above~ Please help to review it again. Thanks~

thrix (approved):
LGTM

happz:
Added `discuss` label - we're still unsure about `model-name` being used to pick the IOMMU "model". We're considering `model`, `model-name`, but also `platform`, `driver`, and maybe other ideas. Leaving the PR open for now, we're going to get back to it next week.

martinhoyer (approved):

tests/unit/provision/mrack/test_hw.py
```diff
@@ -788,6 +789,35 @@ def test_zcrypt_mode(root_logger: Logger) -> None:
...
+
+    assert result.to_mrack() == {
+        'system': {
+            'key_value': {
+                '_key': 'VIRT_IOMMU',
+                '_op': '==',
+                '_value': '0'
+                }
+            }
+        }
```
KwisatzHaderach, line 818: how about this into a for cycle like `for value in True, False:` since that is the only thing different for these 2 sections? then at the `_value` field you can have `str(int(value))`

idorax (approved):

therazix (approved):

tests/unit/provision/mrack/test_hw.py
idorax: @qiankehan Looks the related test function could be refined according to the suggestion above, e.g.
```python
def test_iommu_is_supported(root_logger: Logger) -> None:

    for value in True, False:
        result = _CONSTRAINT_TRANSFORMERS['iommu.is_supported'](
            _parse_iommu({"is-supported": value}), root_logger)

        assert result.to_mrack() == {
            'system': {
                'key_value': {
                    '_key': 'VIRT_IOMMU',
                    '_op': '==',
                    '_value': str(int(value)) 
                    }
                }
            }
```

thrix: @qiankehan we would like to merge this, can you maybe improve this?

thrix:
This was discussed on the hacking meeting, and we will go with current version.

psss:
Just did a quick check with the following `provision` step config:

```yaml
provision:
    how: beaker
    hardware:
        iommu:
            is-supported: true
```

But don't see `VIRT_IOMMU` mentioned in the Beaker job xml. Have I missed anything? @happz, does it work for you?

psss (changes requested):
The change looks good, but I don't see any hardware requirement generated in the Beaker job. This would also deserve at least a short release note.

happz:
> Just did a quick check with the following `provision` step config:
> 
> ```yaml
> provision:
>     how: beaker
>     hardware:
>         iommu:
>             is-supported: true
> ```
> 
> But don't see `VIRT_IOMMU` mentioned in the Beaker job xml. Have I missed anything? @happz, does it work for you?

I believe that is cause by the branch being outdated, it's missing this patch: https://github.com/teemtee/tmt/commit/2bef9dd842a4d3971815f4aa912f23223afb0684

psss:
> I believe that is cause by the branch being outdated, it's missing this patch: [2bef9dd](https://github.com/teemtee/tmt/commit/2bef9dd842a4d3971815f4aa912f23223afb0684)

Right, now it works! :+1:

tests/unit/provision/mrack/test_hw.py
psss: Covered in 9deed7ef.

psss (approved):
Added a short release note, fixed & simplified the test in 9deed7ef. Now should be good for merge.

PR #3106, opened 2024-07-23
Rename the result phase to subresult

seberm:
In today's review meeting, we discussed that it might be a good idea to use a different name for "phase" results, such as "subresult".

The term "phase" was initially used in reference to beakerlib and the `rlPhase{Start,End}` functions. However, the name should be probably more general and not dependent on the used framework. Additionally, the phase keyword is used internally for different purposes.

This relates to:
- https://github.com/teemtee/tmt/pull/2933

CC: @happz, @thrix, @psss

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


seberm:
Do you have any additional ideas or suggestions? Are there any other places where the rename needs to be applied?

tmt/steps/prepare/shell.py
```diff
@@ -63,7 +63,7 @@ def go(
             *,
             guest: 'Guest',
             environment: Optional[tmt.utils.Environment] = None,
-            logger: tmt.log.Logger) -> list[PhaseResult]:
+            logger: tmt.log.Logger) -> list[SubResult]:
```
happz, line 66: Yeah, this is actually incorrect :) Here's what happened:

* I'm working on a patch series in which steps would be able to save results of the individual phases - e.g. `prepare` `ansible` would save "result" for each playbook applied, we would have `prepare/results.yaml`, and so on. Just like results from tests, but for things `prepare` and `finish` do.
* to represent these results, I was about to add new class, `PhaseResult` - we call these individual configurations of steps "phases". Now you can see the naming conflict, right? I had that class in my patch series.
* then your patch adding `PhaseResult` and phase check results landed and added the class of the same name but a very different meaning.
* then, **my** patch landed, **not** adding this class, because it already existed. And now, there is only one `PhaseResult` class, but different parts of the code expect it to be something else: code related to subresults need it to change to `SubResult` but `prepare` and `finish` code expects it to exists.

So, not all renames you did are valid, and your patch is stealing a class I need for my ultimate plan :) I would propose the following:

* your PR should probably not rename `PhaseResult` to `SubResult`, but clone it so we would have both, `SubResult` with the list of checks, and `PhaseResult`, derived from `BaseResult`, with no extra fields.
* all `go()` methods in `prepare` and `finish` plugins refer to `PhaseResult`, and that is correct and must stay: `-> list[PhaseResult]` in `prepare` and `finish` shall remain to refer to the class I hoped to add.

And that should be it, I guess. Your patch would be suddenly much smaller, without changing steps, `SubResult` would not be born by renaming, but by cloning `PhaseResult`, and `PhaseResult` would lose the `check` attribute, because that one belongs to the "subresult" result. WDYT?

seberm: Hello @happz , I am sorry I've misunderstood your patch. I've tried to modify my changes in the following way:
- Removed renames of `PhaseResult` -> `SubResult` from steps.
- `PhaseResult` inherits from `BaseResult` without adding new fields.
- `SubResult` also inherits from `BaseResult`, adding just a list of checks.

Could you please re-check that my changes in the schemas are correct?

Also, are the docstrings for `PhaseResult` and `SubResult` sufficiently descriptive?

happz: Well, you didn't misunderstand my patch, it was my mistake during rebase, I think, that let your code kidnap my precious class ;)

Yep, I think it all looks good, thanks.

seberm: Do you think we should try to squeeze this patch into v1.35?

happz: I wouldn't try to rock the boat so late...

happz (approved):

tmt/schemas/results.yaml
```diff
@@ -142,8 +142,8 @@ items:
     check:
       $ref: "/definitions/check_results"
 
-    phase:
-      $ref: "/definitions/phase_results"
+    subresult:
+      $ref: "/definitions/subresults"
```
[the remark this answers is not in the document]
seberm: Should be fixed now, thanks!

psss (approved):
Thanks!

thrix (approved):
LGTM, added Testing Farm tracking issue, as this will break Testing Farm's results.yaml processing.

PR #3110, opened 2024-07-24
Extend coverage of prepare/finish ansible plugin with more images

happz:
The coverage now matches what `prepare/install` is tested with. This change required refactoring of some bits of `prepare/install` test:

* moving generic helpers into `tests/install.sh`,
* changing tests for `ansible` plugin to use `tests/images.sh` instead of its original one-shot use of images.

This gives us:

* the same coverage for both plugins,
* one place where new environment needs to be added,
* less code duplication, more re-use of "image testing" helpers,
* and, the best, it actually discovers a genuine bug: `ansible` plugins did not report Python interpreter as their essential requirement, therefore playbooks were impossible to run in some environments.

Pull Request Checklist

* [x] implement the feature


KwisatzHaderach (approved):

happz:
Unrelated failures, merging.

PR #3127, opened 2024-08-02
User per-guest SSH master sockets to prevent re-use

happz:
It seems that under some conditions, using same socket path for SSH master processes may lead to SSH commands to operate over incorrect and unexpected guest without noticing. Using per-guest paths should prevent such a race condition.

Pull Request Checklist

* [x] implement the feature


martinhoyer (approved):
Thanks!

thrix (approved):
Lgtm

PR #3133, opened 2024-08-06
Improve version comparison in `--feeling-safe` check

happz:
Simple string comparion will fail for versions like `1.4.0.dev1595+ga35d7140.d20240806` vs `1.38` - the former is clearly older, but the comparison will yield false-ish result.

`packaging.version.Version` should provide smarter and correct comparison for these corner cases.

Pull Request Checklist

* [x] implement the feature


psss (approved):
Looks good, thanks for fixing this.

PR #3143, opened 2024-08-10
Support `virtiofs` in the `virtual` provision step

### LINKED ISSUE #2348: Add virtiofs to `tmt provision` local qemu

frantisekz:
Fixes https://github.com/teemtee/tmt/issues/2348

This wires up the virtiofs mounting support that testcloud gained. It now works even with user sessions with libvirt >= 10.0 . If I could get some pointers on how to make it (arg parsing wise), It could also support passing multiple desired mounts.

As a possible future enhancement, this could get wired up to be automatically used for runs which include file copying to the VM.

@cgwalters fyi, we finally have all the necessary pieces!

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -54,6 +54,7 @@ UserNetworkConfiguration: Any
 QCow2StorageDevice: Any
 RawStorageDevice: Any
 TPMConfiguration: Any
+VIRTIOFSConfiguration: Any
 
 
 def import_testcloud() -> None:
@@ -71,6 +72,7 @@ def import_testcloud() -> None:
     global QCow2StorageDevice
     global RawStorageDevice
     global TPMConfiguration
+    global VIRTIOFSConfiguration
     try:
         import libvirt
         import testcloud.image
@@ -86,6 +88,7 @@ def import_testcloud() -> None:
             SystemNetworkConfiguration,
             TPMConfiguration,
             UserNetworkConfiguration,
+            VIRTIOFSConfiguration,
             X86_64ArchitectureConfiguration,
             )
         from testcloud.workarounds import Workarounds
@@ -342,6 +345,11 @@ class TestcloudGuestData(tmt.steps.provision.GuestSshData):
         option=('-a', '--arch'),
         choices=['x86_64', 'aarch64', 's390x', 'ppc64le'],
         help="What architecture to virtualize, host arch by default.")
+    virtiofs: str = field(
+        default="",
+        option=('-s', '--virtiofs'),
+        help="What directory to mount where, eg."
+             "'<host path>:<guest path>'")
 
     list_local_images: bool = field(
         default=False,
@@ -615,6 +623,7 @@ class GuestTestcloud(tmt.GuestSsh):
         memory ..... memory size for vm
         disk ....... disk size for vm
         connection . either session (default) or system, to be passed to qemu
+        virtiofs ... What directory to mount where, eg. <host path>:<guest path>
         arch ....... architecture for the VM, host arch is the default
     """
 
@@ -626,6 +635,7 @@ class GuestTestcloud(tmt.GuestSsh):
     memory: Optional['Size']
     disk: Optional['Size']
     connection: str
+    virtiofs: str
     arch: str
 
     # Not to be saved, recreated from image_url/instance_name/... every
@@ -909,6 +919,7 @@ class GuestTestcloud(tmt.GuestSsh):
 
         # Initialize and prepare testcloud image
         assert testcloud is not None
+        assert libvirt is not None
         self._image = testcloud.image.Image(self.image_url)
         self.verbose('qcow', self._image.name, 'green')
         if not Path(self._image.local_path).exists():
@@ -971,6 +982,22 @@ class GuestTestcloud(tmt.GuestSsh):
         self._domain.coreos = self.is_coreos
 
         self._apply_hw_arch(self._domain, self.is_kvm, self.is_legacy_os)
+
+        # Do we have virtiofs to handle?
+        if self.virtiofs:
+            if (f"qemu:///{self.connection}" == "qemu:///session" and
+                    libvirt.getVersion() < 10000000):
+                raise tmt.utils.ProvisionError("Only system connection is supported"
+                                               " for virtiofs on libvirt < 10.0.")
+
+            virtiofs_split = self.virtiofs.split(":")
+            if len(virtiofs_split) != 2:
+                raise tmt.utils.ProvisionError("virtiofs specified in a bad format, use "
+                                               "<host path>:<guest path>")
+
+            self._domain.virtiofs_configuration = [
+                VIRTIOFSConfiguration(
+                    virtiofs_split[0], virtiofs_split[1], 0)]
 
         mac_address = testcloud.util.generate_mac_address()
         if f"qemu:///{self.connection}" == "qemu:///system":


tmt/steps/provision/testcloud.py
```diff
@@ -972,6 +983,22 @@ def start(self) -> None:
...
+                    libvirt.getVersion() < 10000000):
+                raise tmt.utils.ProvisionError("Only system connection is supported"
+                                               " for virtiofs on libvirt < 10.0.")
+
+            virtiofs_split = self.virtiofs.split(":")
+            if len(virtiofs_split) != 2:
+                raise tmt.utils.ProvisionError("virtiofs specified in a bad format, use "
+                                               "<host path>:<guest path>")
+
+            self._domain.virtiofs_configuration = [
```
happz, line 998: Right, this processing should go into a custom normalization method, and plugin should work with lists and tuples or dataclasses. Validate & normalize input from fmf and CLI into a single internal representation.

tmt/steps/provision/testcloud.py
```diff
@@ -342,6 +345,11 @@ class TestcloudGuestData(tmt.steps.provision.GuestSshData):
         option=('-a', '--arch'),
         choices=['x86_64', 'aarch64', 's390x', 'ppc64le'],
         help="What architecture to virtualize, host arch by default.")
+    virtiofs: str = field(
+        default="",
```
happz, line 349: Would it mae sense to make it `multiple=True` from the beginning? It would require `default_factory=list` and possibly custom validator, and maybe also a dataclass to hold paths. You mentioned multiple mounts would be doable, would this help you implement it? And are there any blockers? So far it looks very cheap to add, but I don’t know virtiofs, so no push.

How about different name? As a user, I’m not really interested in virtiofs as a technology, but, I would be very interested in *mounting* directories - and there are also other plugins, I bet `container` could support *mounts* too, so how about something more general, like, hm, `mount`? :)

Any  hance of needing some kind of flags? Docker and podman have some, `z` and `Z` at least, does virtiofs have something similar?

tmt/steps/provision/testcloud.py
```diff
@@ -972,6 +983,22 @@ def start(self) -> None:
...
+                raise tmt.utils.ProvisionError("Only system connection is supported"
+                                               " for virtiofs on libvirt < 10.0.")
+
+            virtiofs_split = self.virtiofs.split(":")
+            if len(virtiofs_split) != 2:
+                raise tmt.utils.ProvisionError("virtiofs specified in a bad format, use "
+                                               "<host path>:<guest path>")
+
+            self._domain.virtiofs_configuration = [
+                VIRTIOFSConfiguration(
```
qiankehan, line 999: Need to check if the <host path> and <guest path> are absolute paths. Relative host path is invalid to testcloud or libvirt.  Relative guest path doesn't make sense to guest.

tmt/steps/provision/testcloud.py
qiankehan: > Would it mae sense to make it `multiple=True` from the beginning? It would require `default_factory=list` and possibly custom validator, and maybe also a dataclass to hold paths. You mentioned multiple mounts would be doable, would this help you implement it? And are there any blockers? So far it looks very cheap to add, but I don’t know virtiofs, so no push.
> 
> How about different name? As a user, I’m not really interested in virtiofs as a technology, but, I would be very interested in _mounting_ directories - and there are also other plugins, I bet `container` could support _mounts_ too, so how about something more general, like, hm, `mount`? :)
> 
> Any hance of needing some kind of flags? Docker and podman have some, `z` and `Z` at least, does virtiofs have something similar?



> Would it mae sense to make it `multiple=True` from the beginning? It would require `default_factory=list` and possibly custom validator, and maybe also a dataclass to hold paths. You mentioned multiple mounts would be doable, would this help you implement it? And are there any blockers? So far it looks very cheap to add, but I don’t know virtiofs, so no push.
> 
> How about different name? As a user, I’m not really interested in virtiofs as a technology, but, I would be very interested in _mounting_ directories - and there are also other plugins, I bet `container` could support _mounts_ too, so how about something more general, like, hm, `mount`? :)
How about the name `shared-fs`?
> 
> Any hance of needing some kind of flags? Docker and podman have some, `z` and `Z` at least, does virtiofs have something similar?

tmt/steps/provision/testcloud.py
```diff
@@ -972,6 +983,22 @@ def start(self) -> None:
...
+                    virtiofs_split[0], virtiofs_split[1], 0)]
```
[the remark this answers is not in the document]
frantisekz: Yeah, it's intentional that testcloud's cli doesn't support this yet, only one share for now.

But the backend code there is ready for more shared mounts.

tmt/steps/provision/testcloud.py
frantisekz: Good point, what'd be the best way to tackle? Just take a look if host path starts with '/', and prepend current dir if it doesn't?

qiankehan: `os.path` module will help that checking.
For the host directory, I think user should prepare them. For the guest directory, because guest is provisioned, so testcloud or tmt should prepare it. The host path could be a file or non-exist, for the former, tmt can pre-create it and record that in log; for the latter, just raise an error.   
So the code will be like:
```py
if not os.path.isabs(host_path):
    raise ProvisionError(f"Cannot apply hardware requirement '{constraint}', the host path is not absolute.")
if os.path.isfile(host_path):
    raise ProvisionError(f"Cannot apply hardware requirement '{constraint}', the host path is an existing file.")
if not os.path.exists(host_path):
    os.mkdir(host_path)
    logger.debug(f"the host path {constraint.value} don't exist, create it by tmt ")
```
For the guest path, I think such checking will help. Can we check the 2nd and the 3rd conditions for guest by tmt? Otherwise, we can check them in testcloud.  
  
Now I test testcloud for the non-exist guest path and guest file path.   
For the non-exist guest path, testcloud will create it and mount the virtiofs to it. It is expected
version: testcloud-0.9.13-1.fc41.noarch
```
➜  ~ testcloud create fedora:40 --virtiofs /tmp/export:/mnts

[cloud-user@testcloud ~]$ mount|grep virtiofs
virtiofs-0 on /mnts type virtiofs (rw,relatime)
```
For the file path, the virtiofs will not mounted. And it causes the cloud-final.service failed:
```
➜  ~ ssh -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null cloud-user@192.168.124.215
Warning: Permanently added '192.168.124.215' (ED25519) to the list of known hosts.
cloud-user@192.168.124.215's password: 
[systemd]
Failed Units: 1
  cloud-final.service

[cloud-user@testcloud ~]$ mount|grep virtiofs
```
I think it is better to check the file path first and raise an error.

happz: Unrelated to the proposal, just one note: please, use `pathlib` in tmt code, `os.path` interface is considered legacy and should not be used.

PR #3144, opened 2024-08-12
Use concrete step data class when injecting requires

happz:
Besides raw data, `Step.delegate()` can accept data class instance too. Converting `prepare` step to use this method to provide better type safety.

Pull Request Checklist

* [x] implement the feature


thrix (approved):
LGTM

therazix (approved):

PR #3150, opened 2024-08-15
Remove support for junit-xml, use Jinja2 instead

seberm:
This PR removes support for the deprecated `junit-xml` library and uses the Jinja2 to generate JUnit XML instead.

It also prepares support for various junit flavors (`--flavor`) and adds the possibility to specify user-defined JUnit Jinja templates using the `--template-path` and `--flavor=custom` options. These options are mutually exclusive.

The current implementation requires that custom templates must be valid XML files. If we want to allow users use non-XML custom templates, we would probably go with something like [template report plugin](https://github.com/teemtee/tmt/pull/3150#discussion_r1724853623) for results.

```
/plans/example<br>
    report
        how: junit
        warn: The 'custom' JUnit flavor is used, you are solely responsible for the validity of the XML schema.
        warn: The pretty print is always disabled for 'custom' JUnit flavor
        warn: The generated XML output is not valid XML file or it is not valid against the XSD schema.
        warn: <string>:1:1:FATAL:PARSER:ERR_DOCUMENT_EMPTY: Start tag expected, '<' not found
        warn: Start tag expected, '<' not found, line 1, column 1 (<string>, line 1)

plan failed

Cause number 1:

    The generated XML output is not a valid XML file.
```

The implementation of the `polarion` flavor and its support in the Polarion report plugin will be addressed in a separate PR. In this PR, the Polarion report plugin just calls the `junit.make_junit_xml` function in nearly the same way as before, aiming to keep the Polarion `xunit.xml` file unchanged.

### TODOs:
- [x] Probably divide the changes into multiple PRs (polarion and junit separately)
- [x] Define XSD JUnit for `default` flavor and validate the XML output against this schema. Warn user if there are any problems.
- ~~[ ] Possibility to specify template variables directly from cmdline as key-values (e.g. `--var <key> <value>`) for a `custom` flavor / user-defined template?~~ - This could be a part of "template" report plugin in the future.
- [x] The difference between junit-xml generated files and the new junit files is that the quotes `"` are not escaped to `&quot;` inside of tag data (they **are** correctly escaped inside attributes). According to my knowledge, this should be OK and the XML is also valid.
- [x] Add tests
- [x] Add test which checks there is no schema/XML warning in the tmt logs when running non-custom flavor `The generated XML output is not a valid XML file or it is not valid against the XSD schema`.
- [x] The `duration` should be probably converted to `float` and propagated to XML `time="<val>"` attributes as float, not an integer. If this gets changed, is it possible to enforce float attributes in XSD schema? - tmt works with seconds, let's just convert them to floats using a jinja filter.

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -4,6 +4,22 @@
     Releases
 ======================
 
+tmt-1.37
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`/plugins/report/junit` report plugin now validates all the XML
+flavors against their respective XSD schemas and tries to prettify the final
+XML output. These functionalities are always disabled for ``custom`` flavors.
+The prettify functionality can be controlled for non-custom templates by
+``--prettify`` and ``--no-prettify`` arguments.
+
+The :ref:`/plugins/report/junit` report plugin now uses Jinja instead of
+``junit-xml`` library to generate the JUnit XMLs. It also adds support for a
+new ``--flavor`` argument. Using this argument the user can choose between a
+``default`` flavor, which keeps the current behavior untouched, and a
+``custom`` flavor where user must provide a custom template using a
+``--template-path`` argument.
+
 
 tmt-1.37.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,8 @@ provision-virtual = [
     ]
 provision-container = []
 report-junit = [
-    "junit_xml>=1.9",
+    # Required to support XML parsing and checking the XSD schemas.
+    "lxml>=4.6.5",
     ]
 report-polarion = [
     "tmt[report-junit]",
@@ -139,6 +140,7 @@ dependencies = [
     "types-jinja2",
     "types-babel",
     "types-docutils",
+    "types-lxml",
     ]
 features = ["all"]
 
@@ -216,7 +218,7 @@ module = [
     "guestfs.*",
     "html2text.*",
     "fmf.*",
-    "junit_xml.*",
+    "lxml.*",
     "libvirt.*",
     "nitrate.*",
     "pylero.*",
--- a/tests/report/junit/data/custom.xml.j2
+++ b/tests/report/junit/data/custom.xml.j2
@@ -1,0 +1,6 @@
+<?xml version="1.0" ?>
+<tests>
+    {% for result in RESULTS %}
+        <test name="{{ result.name | e }}" value="{{ result.result.value | e }}"/>
+    {% endfor %}
+</tests>
--- a/tests/unit/test_report_junit.py
+++ b/tests/unit/test_report_junit.py
@@ -19,7 +19,7 @@ def report_fix(tmppath: Path, root_logger):
     type(plan_mock).name = name_property
     type(step_mock).plan = plan_mock
 
-    out_file_path = str(tmppath / "out.xml")
+    out_file_path = Path(tmppath / "out.xml")
 
     report = ReportJUnit(
         logger=root_logger,
@@ -120,11 +120,14 @@ class TestStateMapping:
 
         report.go()
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="0" failures="0" tests="1" time="0.0">
-    <testsuite disabled="0" errors="0" failures="0" name="name" skipped="0" tests="1" time="0">
-        <testcase name="/pass"/>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="0" failures="0" skipped="0" tests="1" time="0.0">
+    <testcase name="/pass">
+
+
+                </testcase>
+  </testsuite>
 </testsuites>
 """)
 
@@ -133,13 +136,13 @@ class TestStateMapping:
         results.append(Result(result=ResultOutcome.INFO, name="/info", serial_number=1))
         report.go()
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="0" failures="0" tests="1" time="0.0">
-    <testsuite disabled="0" errors="0" failures="0" name="name" skipped="1" tests="1" time="0">
-        <testcase name="/info">
-            <skipped type="skipped" message="info"/>
-        </testcase>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="0" failures="0" skipped="1" tests="1" time="0.0">
+    <testcase name="/info">
+      <skipped type="skipped" message="info"/>
+    </testcase>
+  </testsuite>
 </testsuites>
 """)
 
@@ -148,13 +151,13 @@ class TestStateMapping:
         results.append(Result(result=ResultOutcome.WARN, name="/warn", serial_number=1))
         report.go()
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="1" failures="0" tests="1" time="0.0">
-    <testsuite disabled="0" errors="1" failures="0" name="name" skipped="0" tests="1" time="0">
-        <testcase name="/warn">
-            <error type="error" message="warn"/>
-        </testcase>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="1" failures="0" skipped="0" tests="1" time="0.0">
+    <testcase name="/warn">
+      <error type="error" message="warn"/>
+    </testcase>
+  </testsuite>
 </testsuites>
 """)
 
@@ -163,13 +166,13 @@ class TestStateMapping:
         results.append(Result(result=ResultOutcome.ERROR, name="/error", serial_number=1))
         report.go()
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="1" failures="0" tests="1" time="0.0">
-    <testsuite disabled="0" errors="1" failures="0" name="name" skipped="0" tests="1" time="0">
-        <testcase name="/error">
-            <error type="error" message="error"/>
-        </testcase>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="1" failures="0" skipped="0" tests="1" time="0.0">
+    <testcase name="/error">
+      <error type="error" message="error"/>
+    </testcase>
+  </testsuite>
 </testsuites>
 """)
 
@@ -178,12 +181,12 @@ class TestStateMapping:
         results.append(Result(result=ResultOutcome.FAIL, name="/fail", serial_number=1))
         report.go()
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="0" failures="1" tests="1" time="0.0">
-    <testsuite disabled="0" errors="0" failures="1" name="name" skipped="0" tests="1" time="0">
-        <testcase name="/fail">
-            <failure type="failure" message="fail"/>
-        </testcase>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="0" failures="1" skipped="0" tests="1" time="0.0">
+    <testcase name="/fail">
+      <failure type="failure" message="fail"/>
+    </testcase>
+  </testsuite>
 </testsuites>
 """)
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -1,5 +1,9 @@
 import dataclasses
-from typing import TYPE_CHECKING, Optional, overload
+import functools
+from collections.abc import Iterator
+from typing import TYPE_CHECKING, Any, Optional, TypedDict, cast, overload
+
+from jinja2 import FileSystemLoader, select_autoescape
 
 import tmt
 import tmt.base
@@ -10,16 +14,24 @@ import tmt.steps
 import tmt.steps.report
 import tmt.utils
 from tmt.plugins import ModuleImporter
+from tmt.result import ResultOutcome
 from tmt.utils import Path, field
+from tmt.utils.templates import default_template_environment, render_template_file
 
 if TYPE_CHECKING:
-    import junit_xml
+    import lxml
 
-    from tmt.steps.report import ReportPlugin
-    from tmt.steps.report.polarion import ReportPolarionData
+    from tmt._compat.typing import TypeAlias
 
-DEFAULT_NAME = "junit.xml"
+    XMLElement: TypeAlias = Any
 
+
+DEFAULT_NAME = 'junit.xml'
+DEFAULT_FLAVOR_NAME = 'default'
+CUSTOM_FLAVOR_NAME = 'custom'
+
+# Relative path to tmt junit template directory.
+DEFAULT_TEMPLATE_DIR = Path('steps/report/junit/templates/')
 
 # ignore[unused-ignore]: Pyright would report that "module cannot be
 # used as a type", and it would be correct. On the other hand, it works,
@@ -31,10 +43,10 @@ DEFAULT_NAME = "junit.xml"
 # is enough to suppress it. And then mypy complains about an unused
 # ignore, hence `unused-ignore` code, leading to apparently confusing
 # directive.
-import_junit_xml: ModuleImporter['junit_xml'] = ModuleImporter(  # type: ignore[unused-ignore]
-    'junit_xml',
+import_lxml: ModuleImporter['lxml'] = ModuleImporter(  # type: ignore[valid-type]
+    'lxml',
     tmt.utils.ReportError,
-    "Missing 'junit-xml', fixable by 'pip install tmt[report-junit]'.",
+    "Missing 'lxml', fixable by 'pip install tmt[report-junit]'.",
     tmt.log.Logger.get_bootstrap_logger())
 
 
@@ -54,44 +66,232 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
         h, m, s = duration.split(':')
         return int(h) * 3600 + int(m) * 60 + int(s)
     except Exception as error:
-        raise tmt.utils.ReportError(
-            f"Malformed duration '{duration}' ({error}).")
+        raise tmt.utils.ReportError(f"Malformed duration '{duration}'.") from error
+
+
+class ResultWrapper:
+    """
+    The context wrapper for :py:class:`tmt.Result`.
+
+    Adds possibility to wrap the :py:class:`tmt.Result` and dynamically add more attributes which
+    get available inside the template context.
+    """
+
+    class PropertyDict(TypedDict):
+        """ Defines a property dict, which gets propagated into the final template properties. """
+
+        name: str
+        value: str
+
+    def __init__(self, wrapped: tmt.Result) -> None:
+        self._wrapped = wrapped
+        self._properties: dict[str, str] = {}
+
+    @property
+    def properties(self) -> list[PropertyDict]:
+        return [{'name': k, 'value': v} for k, v in self._properties.items()]
+
+    @properties.setter
+    def properties(self, keyval: dict[str, str]) -> None:
+        self._properties = keyval
+
+    def __getattr__(self, name: str) -> Any:
+        """ Returns an attribute of a wrapped ``tmt.Result`` instance """
+        return getattr(self._wrapped, name)
+
+
+class ResultsContext:
+    """
+    The results context for Jinja templates.
+
+    A class which keeps the results context (especially the result summary) for JUnit template. It
+    wraps all the :py:class:`tmt.Result` instances into the :py:class:`ResultWrapper`.
+    """
+
+    def __init__(self, results: list[tmt.Result]) -> None:
+        # Decorate all the tmt.Results with more attributes
+        self._results: list[ResultWrapper] = [ResultWrapper(r) for r in results]
+
+    def __iter__(self) -> Iterator[ResultWrapper]:
+        """ Possibility to iterate over results by iterating an instance """
+        return iter(self._results)
+
+    def __len__(self) -> int:
+        """ Returns the number of results """
+        return len(self._results)
+
+    @functools.cached_property
+    def passed(self) -> list[ResultWrapper]:
+        """ Returns results of passed tests """
+        return [r for r in self._results if r.result == ResultOutcome.PASS]
+
+    @functools.cached_property
+    def skipped(self) -> list[ResultWrapper]:
+        """ Returns results of skipped tests """
+        return [r for r in self._results if r.result == ResultOutcome.INFO]
+
+    @functools.cached_property
+    def failed(self) -> list[ResultWrapper]:
+        """ Returns results of failed tests """
+        return [r for r in self._results if r.result == ResultOutcome.FAIL]
+
+    @functools.cached_property
+    def errored(self) -> list[ResultWrapper]:
+        """ Returns results of tests with error/warn outcome """
+        return [r for r in self._results if r.result in [
+                ResultOutcome.ERROR,
+                ResultOutcome.WARN]]
+
+    @functools.cached_property
+    def duration(self) -> int:
+        """ Returns the total duration of all tests in seconds """
+        # cast: mypy does not understand the proxy-ness of `ResultWrapper`. `r.duration`
+        # will exists, therefore adding a `cast` to convince mypy the list is pretty much
+        # nothing but the list of results.
+        return sum(
+            duration_to_seconds(r.duration) or 0 for r in cast(list[tmt.Result], self._results))
 
 
 def make_junit_xml(
-        report: 'ReportPlugin[ReportJUnitData]|ReportPlugin[ReportPolarionData]'
-        ) -> 'junit_xml.TestSuite':
-    """ Create junit xml object """
-    junit_xml = import_junit_xml()
+        phase: tmt.steps.report.ReportPlugin[Any],
+        flavor: str = DEFAULT_FLAVOR_NAME,
+        template_path: Optional[Path] = None,
+        include_output_log: bool = True,
+        prettify: bool = True,
+        results_context: Optional[ResultsContext] = None,
+        **extra_variables: Any
+        ) -> str:
+    """
+    Create JUnit XML file and return the data.
 
-    suite = junit_xml.TestSuite(report.step.plan.name)
+    :param phase: instance of a :py:class:`tmt.steps.report.ReportPlugin`.
+    :param flavor: name of a JUnit flavor to generate.
+    :param template_path: if set, the provided template will be used instead of a pre-defined
+        flavor template. In this case, the ``flavor`` must be set to ``custom`` value.
+    :param include_output_log: if enabled, the ``<system-out>`` tags are included in the final
+        template output.
+    :param prettify: allows to control the XML pretty print.
+    :param results_context: if set, the provided :py:class:`ResultsContext` is used in a template.
+    :param extra_variables: if set, these variables get propagated into the Jinja template.
+    """
 
-    for result in report.step.plan.execute.results():
+    # Get the template context for tmt results
+    results_context = results_context or ResultsContext(phase.step.plan.execute.results())
+
+    # Prepare the template environment
+    environment = default_template_environment()
+
+    template_path = template_path or tmt.utils.resource_files(
+        DEFAULT_TEMPLATE_DIR / Path(f'{flavor}.xml.j2'))
+
+    # Use a FileSystemLoader for a non-custom flavor
+    if flavor != CUSTOM_FLAVOR_NAME:
+        environment.loader = FileSystemLoader(
+            searchpath=tmt.utils.resource_files(DEFAULT_TEMPLATE_DIR))
+
+    def _read_log(log: Path) -> str:
+        """ Read the contents of a given result log """
+        if not log:
+            return ''
+
         try:
-            main_log = report.step.plan.execute.read(result.log[0])
-        except (IndexError, AttributeError):
-            main_log = None
-        case = junit_xml.TestCase(
-            result.name,
-            classname=None,
-            elapsed_sec=duration_to_seconds(result.duration))
+            return str(phase.step.plan.execute.read(log))
+        except tmt.utils.FileError:
+            return ''
 
-        if report.data.include_output_log:
-            case.stdout = main_log
+    environment.filters.update({
+        'read_log': _read_log,
+        'duration_to_seconds': duration_to_seconds,
+        'failures': tmt.result.Result.failures,
+        })
 
-        # Map tmt OUTCOME to JUnit states
-        if result.result == tmt.result.ResultOutcome.ERROR:
-            case.add_error_info(result.result.value, output=result.failures(main_log))
-        elif result.result == tmt.result.ResultOutcome.FAIL:
-            case.add_failure_info(result.result.value, output=result.failures(main_log))
-        elif result.result == tmt.result.ResultOutcome.INFO:
-            case.add_skipped_info(result.result.value, output=result.failures(main_log))
-        elif result.result == tmt.result.ResultOutcome.WARN:
-            case.add_error_info(result.result.value, output=result.failures(main_log))
-        # Passed state is the default
-        suite.test_cases.append(case)
+    # Explicitly enable the autoescape because it's disabled by default by tmt.
+    # See /teemtee/tmt/issues/2873 for more info.
+    environment.autoescape = select_autoescape(enabled_extensions=('xml.j2'))
 
-    return suite
+    xml_data = render_template_file(
+        template_path,
+        environment,
+        RESULTS=results_context,
+        PLAN=phase.step.plan,
+        INCLUDE_OUTPUT_LOG=include_output_log,
+        **extra_variables)
+
+    # Try to use lxml to check the flavor XML schema and prettify the final XML
+    # output.
+    try:
+        from lxml import etree
+    except ImportError:
+        phase.warn(
+            "Install 'tmt[report-junit]' to support neater JUnit XML output and the XML schema "
+            "validation against the XSD.")
+        return xml_data
+
+    xml_parser_kwargs: dict[str, Any] = {
+        'remove_blank_text': prettify,
+        'schema': None,
+        }
+
+    # The schema check must be done only for a non-custom JUnit flavors
+    if flavor != CUSTOM_FLAVOR_NAME:
+        xsd_schema_path = Path(tmt.utils.resource_files(
+            Path(f'steps/report/junit/schemas/{flavor}.xsd')))
+
+        schema_root: XMLElement = etree.XML(xsd_schema_path.read_bytes())
+        xml_parser_kwargs['schema'] = etree.XMLSchema(schema_root)
+    else:
+        phase.warn(
+            f"The '{CUSTOM_FLAVOR_NAME}' JUnit flavor is used, you are solely responsible "
+            "for the validity of the XML schema.")
+
+        if prettify:
+            phase.warn(f"The pretty print is always disabled for '{CUSTOM_FLAVOR_NAME}' JUnit "
+                       "flavor.")
+
+        xml_parser_kwargs['remove_blank_text'] = prettify = False
+
+    xml_parser = etree.XMLParser(**xml_parser_kwargs)
+    try:
+        # S320: Parsing of untrusted data is known to be vulnerable to XML
+        # attacks.
+        tree_root: XMLElement = etree.fromstring(xml_data, xml_parser)  # noqa: S320
+
+    except etree.XMLSyntaxError as e:
+        phase.warn(
+            'The generated XML output is not a valid XML file or it is not valid against the '
+            'XSD schema.')
+
+        if flavor != CUSTOM_FLAVOR_NAME:
+            phase.warn('Please, report this problem to project maintainers.')
+
+        for err in e.error_log:
+            phase.warn(str(err))
+
+        # Return the prettified XML without checking the XSD
+        del xml_parser_kwargs['schema']
+
+        try:
+            tree_root = etree.fromstring(  # noqa: S320
+                xml_data, etree.XMLParser(**xml_parser_kwargs))
+        except etree.XMLSyntaxError as error:
+            phase.verbose('rendered XML', xml_data, 'red')
+            raise tmt.utils.ReportError(
+                'The generated XML output is not a valid XML file. Use `--verbose` argument '
+                'to show the output.') from error
+
+    # Do not be fooled by the `encoding` parameter: even with `utf-8`,
+    # `tostring()` will still return bytes. `unicode`, on the other
+    # hand, would give us a string, but then, no XML declaration.
+    # So, we get bytes, and we need to apply `decode()` on our own.
+    xml_output: bytes = etree.tostring(
+        tree_root,
+        xml_declaration=True,
+        pretty_print=prettify,
+        # The 'utf-8' encoding must be used instead of 'unicode', otherwise
+        # the XML declaration is not included in the output.
+        encoding='utf-8')
+
+    return str(xml_output.decode('utf-8'))
 
 
 @dataclasses.dataclass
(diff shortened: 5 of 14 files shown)


tmt/steps/report/junit.py
```diff
@@ -103,6 +47,19 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
 
+    flavor: str = field(
+        default=None,
+        option='--flavor',
+        metavar='FLAVOR',
+        choices=[DEFAULT_FLAVOR_NAME],
+        help="Name of a JUnit flavor to generate. By default, the 'default' flavor is used.")
+
+    template_path: str = field(
+        default=None,
```
happz, line 58: When the default is `None`, type of the field cannot be just `str` - when left unset, `None` would be assigned to it & we'd get apparently string field with the value of `None`. `Optional[str]` would be the right type here and above.

tmt/steps/report/junit.py
```diff
@@ -111,36 +68,112 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
+    def duration(self):
+        return sum([duration_to_seconds(r.duration) for r in self._results])
+
+
 @tmt.steps.provides_method('junit')
 class ReportJUnit(tmt.steps.report.ReportPlugin[ReportJUnitData]):
     """
-    Save test results in JUnit format.
+    Save test results in chosen JUnit flavor format. When flavor is set to
+    custom, the ``template_path`` with a path to a custom template must be
```
happz, line 114: `template-path`, as keys in fmf shall use `-` when composed from multiple words.

tmt/steps/report/junit.py
```diff
@@ -111,36 +68,112 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
     _data_class = ReportJUnitData
 
+    def check_options(self) -> None:
+        """ Check the module options """
+
+        if self.data.flavor and self.data.template_path:
+            raise tmt.utils.ReportError("The options '--template-path' and "
+                                        "'--flavor' are mutually exclusive.")
+
+        self.data.flavor = self.data.flavor or DEFAULT_FLAVOR_NAME
```
happz, line 130: Please, add a `TODO` for me - we need a better way when looking for the source of option, I don't have the solution right now, but we already have some components. This `self.data.flavor or ...` is a task for `default` setting of the field itself, doing it in this imperative way is not going to be easy to maintain.

tmt/steps/report/junit.py
```diff
@@ -111,36 +68,112 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
-        junit_xml = import_junit_xml()
-        suite = make_junit_xml(self)
+        self.check_options()
 
         assert self.workdir is not None
         f_path = self.data.file or self.workdir / DEFAULT_NAME
+
+        # Get the template according to chosen flavor
+        if self.data.flavor == 'custom':
+            template_path = Path(self.data.template_path)
```
happz, line 146: Check `tmt.utils.normalize_path`, it should be used in the field definition, and the type should be `Optional[Path]`. The path-like input should be converted into a `Path` instance on the boundary, and that's the step data class.

tmt/steps/report/junit.py
```diff
@@ -111,36 +68,112 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
-                    junit_xml.TestSuite.to_file(fw, [suite])
+            self.write(
+                f_path,
+                data=tmt.utils.render_template_file(
+                    template_path,
+                    environment,
+                    results=results_context,
+                    base_dir=self.step.plan.execute.workdir,
+                    plan=self.step.plan,
+                    include_output_log=self.data.include_output_log))
```
happz, line 175: * I'd recommend using upper-cased names for our custom variables injected into the field, i.e. `RESULTS`, `BASE_DIR`, ... It makes them stand out in the template.
* how about `WORKDIR` instead of `base_dir`? tmt does not have many "base" directories, but "workdir" is a very common thing, sticking to the name would make the story more consistent.

seberm:
Hello @happz , thanks for your early hints. It's still very early for a review. Many code parts will probably change, especially due to the Polarion report plugin.

LecrisUT:
Is it possible to validate the `junit.xml` files created?

seberm:
> Is it possible to validate the `junit.xml` files created?

Currently, this functionality is not supported, and the XML files are not validated against any XSD schema, even when using `junit-xml`.

From python-junit-xml `README.rst`:

> As there is no definitive Jenkins JUnit XSD that I could find, the XML documents created by this module support a schema based on Google searches and the Jenkins JUnit XML reader source code. File a bug if something doesn't work like you expect it to. For Bamboo situation is the same.

The `junit-xml` repository currently includes only Python [unit tests](https://github.com/kyrus/python-junit-xml/tree/master/tests), while TMT tests have **very basic** XML checks ([polarion](https://github.com/teemtee/tmt/blob/main/tests/report/polarion/test.sh), [default junit](https://github.com/teemtee/tmt/blob/main/tests/report/junit/test.sh)). However, there is no existing code that validates the entire XML file against an XSD schema.

I plan to create simple XSD schemas for each JUnit flavor supported by TMT and add tests validating the generated JUnit files against these schemas.

Additionally, it would be a good idea to have TMT dynamically validate the schema of the generated JUnit files. If a file does not conform to its respective schema, the report plugin should fail, unless e.g. the `--force` argument is used, which would allow it to continue despite the schema mismatch.

happz:
> I plan to create simple XSD schemas for each JUnit flavor supported by TMT and add tests validating the generated JUnit files against these schemas.
> 
> Additionally, it would be a good idea to have TMT dynamically validate the schema of the generated JUnit files. If a file does not conform to its respective schema, the report plugin should fail, unless e.g. the `--force` argument is used, which would allow it to continue despite the schema mismatch.

+1. Check out https://github.com/teemtee/tmt/pull/3153, it does similar validation when consuming `results.yaml`, but it's not a blocker, only a warning sign, at least for now. That said, these files are not created by tmt - junit files would be. Failing this particular validation would mean the plugin is creating a broken output, which would be a bug. I'm not sure `--force` would be needed - if users wanted to overcome this bug temporarily, they should be able to modify either the schema or the template, and `--force` would sort of normalize the behavior in my eyes.

tmt/steps/report/junit.py
seberm: I've taken a different approach by creating a flavor called `custom`. When this flavor is selected, the `--template-path` option must also be provided. Additionally, `--template-path` cannot be used with any other flavor.

I haven't found a better solution :/

tmt/steps/report/junit.py
seberm: Is it correct now? Can I normalize the path the same way (with `tmt.utils.normalize_path`) also for the `--file` option?

tmt/steps/report/junit.py
```diff
@@ -58,40 +42,170 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+    xml_data = render_template_file(
+        template_path,
+        environment,
+        RESULTS=results_context,
+        PLAN=report.step.plan,
+        INCLUDE_OUTPUT_LOG=report.data.include_output_log,
+        **extra_variables)
+
+    # Try to use lxml to check the flavor XML schema and prettify the final XML
+    # output.
```
happz, line 146: FTR, I for one would not bother with prettifying the output - it's been rendered by a template whose owner had full and absolute control over indentation and line splitting - if they decided that their elements shall be indented by 13 tabs for every odd level, then they should have such indentation. Their template, their rules, who are we to mess up their team's perfectly selected formatting rules :) Validation, absolutely, but I would advise against reformatting and changing the output in any way.

tmt/steps/report/junit.py
happz: I believe so, `--file` looks like a relict of times when `normalize_path` did not exist yet.

tmt/steps/report/junit.py
```diff
@@ -58,40 +42,170 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
-    return suite
+            return str(report.step.plan.execute.read(log))
+        except AttributeError:
+            return ''
+
+    environment = default_template_environment()
+
+    # Use a FileSystemLoader for a non-custom flavor
+    if flavor != CUSTOM_FLAVOR_NAME:
+        environment.loader = FileSystemLoader(searchpath=Path('tmt/steps/report/junit/templates/'))
```
happz, line 124: I'm not 100 % sure, but I would expect some form of `resource_files` here as well, because the search path must also reflect the position of the installed tmt on users' systems.

tmt/steps/report/junit.py
```diff
@@ -58,40 +42,170 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+            Path(f'steps/report/junit/templates/{flavor}.xml.j2')))
+
+    environment.filters.update({
+        'read_log': _read_log,
+        'duration_to_seconds': duration_to_seconds,
+        })
+
+    # Explicitly enable the autoescape because it's disabled by default by TMT
+    # (see /teemtee/tmt/issues/2873 for more info.
+    environment.autoescape = select_autoescape(enabled_extensions=('xml'))
```
happz, line 135: Is `xml` correct, or does Jinja cut away the `.j2` suffix?

tmt/steps/report/junit.py
```diff
@@ -58,40 +42,170 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+        return sum([duration_to_seconds(r.duration) or 0 for r in self._results])
+
+
 def make_junit_xml(
-        report: 'ReportPlugin[ReportJUnitData]|ReportPlugin[ReportPolarionData]'
-        ) -> 'junit_xml.TestSuite':
-    """ Create junit xml object """
-    junit_xml = import_junit_xml()
+        report: 'ReportPlugin[ReportJUnitData]|ReportPlugin[ReportPolarionData]',
+        flavor: str,
```
happz, line 92: * most likely a nitpicking, but please, consider `phase` instead of `report`. I tried to use it throughout the code to represent plugin instances, which is the case here as well.
* and if `default` would be the default value of `flavor`, which would make sense, right, then the call in `polarion` plugin could be even simpler :)

On the other hand, given the note in `polarion`, it makes sense to keep the `flavor='default'` there even if `flavor` would have a default value, because this argument will change when `polarion` flavor arrives. But, I'd still argue that `flavor: str = DEFAULT_FLAVOR` would be very much due.

tmt/steps/report/junit.py
```diff
@@ -58,40 +42,170 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
-            case.add_skipped_info(result.result.value, output=result.failures(main_log))
-        elif result.result == tmt.result.ResultOutcome.WARN:
-            case.add_error_info(result.result.value, output=result.failures(main_log))
-        # Passed state is the default
-        suite.test_cases.append(case)
-
-    return suite
+            return str(report.step.plan.execute.read(log))
+        except AttributeError:
+            return ''
```
happz, line 118: Is it safe to let this slip silently? When can the `read()` end up with an exception?

tmt/steps/report/junit.py
seberm: I think you might be right:
- https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.select_autoescape

```
The select_autoescape() function returns a function that works roughly like this:

def autoescape(template_name):
    if template_name is None:
        return False
    if template_name.endswith(('.html', '.htm', '.xml'))
```

I will change it to `xml.j2`

tmt/steps/report/junit.py
seberm: Changed the `report -> phase` and I added the default value to `flavor` argument. Thanks.

tmt/steps/report/junit.py
seberm: I agree that we should probably disable the pretty printing for the `custom` flavor/template. However, I still believe it should be enabled for predefined TMT JUnit flavors. Without it, the XML output isn't very clean - jinja macros add a lot of whitespace and poor indentation.

I spent some time modifying the templates (playing with [Jinja whitespace control](https://jinja.palletsprojects.com/en/3.0.x/templates/#whitespace-control)) to improve the output, but this made the Jinja templates much harder to read.

This is an example output with `pretty_print=False` and `remove_blank_text=False`:
```
<?xml version='1.0' encoding='utf-8'?>
<testsuites disabled="0" errors="1" failures="1" tests="7" time="1">
    <testsuite name="/plans/example&lt;br&gt;" disabled="0" errors="1" failures="1" skipped="1" tests="7" time="1">

                <testcase name="/tests/call-dmesg&quot;&lt;br&gt;">

                        <system-out>test1
</system-out>
                </testcase>

                <testcase name="/tests/mine-rep-multi">
                        <failure type="failure" message="fail"/>

                </testcase>

                <testcase name="/tests/mine-rep-pass">

                </testcase>

                <testcase name="/tests/please-fail">
                        <error type="error" message="error"/>

                </testcase>

                <testcase name="/tests/please-skip">
                        <skipped type="skipped" message="info">skipped
</skipped>

                        <system-out>skipped
</system-out>
                </testcase>

                <testcase name="/tests/sleep" time="1">

                </testcase>

                <testcase name="/tests/special-output">

                        <system-out>&lt;h2&gt;ahoj"jak"se&amp;mas&lt;/h2&gt;
</system-out>
                </testcase>
    </testsuite>

</testsuites>
```

tmt/steps/report/junit.py
seberm: The old implementation was catching `(IndexError, AttributeError)` and it was failing silently setting the `main_log = None`, which is why I was handling it in a similar way:
- https://github.com/teemtee/tmt/blob/c5c214a88af4fb9bf03940386e6df90b0724cf40/tmt/steps/report/junit.py#L71

I'm not entirely sure how to address this. I think I'm fine with handling it more transparently by simply using `return str(phase.step.plan.execute.read(log))` and not catching any exceptions.

tmt/steps/report/junit.py
happz: Or, we can simply provide `--prettify` or `--pretty` option to enable this behavior, and let the user decide. Can be enabled by default for non-`custom` flavors, disabled by default for `custom` ones, and everyone would be happy :)

seberm: The custom template no longer even needs to be in XML format:

Example test input: https://github.com/teemtee/tmt/blob/7148b056697843c0707970b545c84004af7572d8/tests/report/junit/data/custom.xml.j2

happz: Yeah, that's a correct observation. It's similar to `export` where you can go with `--how=template`, and use whatever Jinja template you wish, or `--how=rst` which takes its own Jinja template and calls `template` plugin to render it.

Maybe we could do the same here - probably not here and now, but it would make sense to have a `report/template` plugin that would focus on rendering results, and change `junit` plugin to 1. ship templates for various flavors, 2. call `template` plugin, 3. perform validation & pretty-printing, as bells & whistles on top of the actual junit production. I wouldn't let `junit` render anything for now, it should be a XML outcome, otherwise, we risk `junit` would start rendering Asciidoc representation of results because users realize they can use any template - but we could give them `report/template` someday in the future.

pyproject.toml
```diff
@@ -60,7 +60,9 @@ provision-virtual = [
     ]
 provision-container = []
 report-junit = [
-    "junit_xml>=1.9",
+    # Required by to support XML parsing and checking the XSD schemas.
```
happz, line 63: "by to", I suppose one of these should go.

tests/report/junit/data/custom.xml.j2
```diff
@@ -0,0 +1,3 @@
+{% for result in RESULTS %}
+name='{{ result.name | e }}',value='{{ result.result.value | e }}'
+{% endfor %}
```
happz, line 3: I would rather try a simple XML - not emitting an XML could put a thought in users' heads that using this plugin to render anything but XML is a good idea :)

tmt/steps/report/junit.py
seberm: Ok, I agree. I'll change the code to disallow non-XML custom templates, just to be sure users will not misuse it. I also added info about it to the description of this PR.

tests/report/junit/data/custom.xml.j2
seberm: I modified the test to use the XML. Thanks.

tmt/steps/report/junit.py
```diff
@@ -58,40 +69,188 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+        """ Returns the total duration of all tests in seconds """
+        return sum([duration_to_seconds(r.duration) or 0 for r in self._results])
+
+
 def make_junit_xml(
-        report: 'ReportPlugin[ReportJUnitData]|ReportPlugin[ReportPolarionData]'
-        ) -> 'junit_xml.TestSuite':
-    """ Create junit xml object """
-    junit_xml = import_junit_xml()
+        phase: 'ReportPlugin[ReportJUnitData]|ReportPlugin[ReportPolarionData]',
```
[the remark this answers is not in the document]
happz: I'm afraid both are correct :) It's a plugin *instance*, and represents one *phase* of the `report` step.

One note though: `ReportPlugin[ReportStepData]` or even `ReportPlugin[Any]` would probably be better, because this list will never be complete, anyone can create an out-of-the-tree plugin & use `junit` plugin to render its templates. Limiting it to just two examples does not seem correct.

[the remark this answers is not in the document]
seberm: Ok, I've changed it to `ReportPlugin[Any]`. I was keeping it the same way as it was before.

tmt/steps/report/junit.py
```diff
@@ -58,40 +69,188 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+        return iter(self._results)
+
+    def __len__(self) -> int:
+        """ Returns the number of results """
+        return len(self._results)
+
+    @functools.cached_property
+    def executed(self) -> list[tmt.Result]:
+        """ Returns results of all executed tests """
+        return [r for r in self._results if r.result != ResultOutcome.INFO]
```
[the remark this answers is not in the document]
KwisatzHaderach: nevermind, I see it now :)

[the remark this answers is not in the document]
seberm: Actually, I do not remember why I created this property at all. It's not used anymore, so I've removed it completely.

tmt/steps/report/junit.py
```diff
@@ -58,40 +69,188 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+    # Prepare the template environment
+    environment = default_template_environment()
 
-    for result in report.step.plan.execute.results():
+    template_dir = Path('steps/report/junit/templates/')
+    template_path = template_path or tmt.utils.resource_files(
+        template_dir / Path(f'{flavor}.xml.j2'))
+
+    # Use a FileSystemLoader for a non-custom flavor
+    if flavor != CUSTOM_FLAVOR_NAME:
```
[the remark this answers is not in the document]
seberm: No, I am checking the `custom` flavor is not used. The **non-custom** flavors will use a Jinja2 `FileSystemLoader` to load a whole tmt folder with non-custom flavor jinja2 templates.

I want to be able to use jinja's template "includes" and template inheritance (blocks) for non-custom flavors and I want to disallow this behavior for `custom` flavors. I want to force the users to use self-contained custom templates (by this I mean the whole template defined in one file).

If we wanted to allow users to use template inheritance (`FileSystemLoader`), it would require getting a template directory (as a CLI parameter) and also pre-define some custom template entry point (which must always exist in this folder) which will be rendered by tmt (e.g. `custom.j2.xml`).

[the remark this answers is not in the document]
KwisatzHaderach: from the definition it only allows `default` and `custom` flavours, that's the issue here I was pointing out (and not specifying fully)

so maybe changing the definition is needed?

seberm:
I've moved the decorator which adds the `properties` to `tmt.Result` into the `junit` module (related PR https://github.com/teemtee/tmt/pull/3166). Hopefully, it's better now :).

tmt/steps/report/junit/templates/_base.xml.j2
```diff
@@ -0,0 +1,37 @@
...
+                    {% endif %}
+
+                    {% if INCLUDE_OUTPUT_LOG and main_log %}
+                        <system-out>{{ main_log | e }}</system-out>
+                    {% endif %}
+
+                    {# Optionally add the result properties #}
+                    {% if result.properties is defined %}
+                        {% with properties=result.properties %}
+                            {% include "includes/_properties.xml.j2" %}
```
KwisatzHaderach, line 28: since you already have it here why not add it for the testsuites as well?

also, properties are missing from the xsd so it would fail validation if someone actually used it before the polarion part gets merged

seberm:
Could someone help me fix the tests? I'm having trouble understanding why they are failing and whether the errors are related to my pull request. Is it possible that the `python3-lxml` package is not installed in some environments?

seberm:
> Could someone help me fix the tests? I'm having trouble understanding why they are failing and whether the errors are related to my pull request. Is it possible that the `python3-lxml` package is not installed in some environments?

```
Error: 
 Problem 1: conflicting requests
  - nothing provides python3.9dist(lxml) >= 4.9 needed by tmt+report-junit-1.36.dev888-1.20240828170910228891.pr3166.26.g07024055.el9.noarch from @commandline
  - nothing provides python3.9dist(types-lxml) needed by tmt+report-junit-1.36.dev888-1.20240828170910228891.pr3166.26.g07024055.el9.noarch from @commandline
 Problem 2: package tmt+all-1.36.dev888-1.20240828170910228891.pr3166.26.g07024055.el9.noarch from @commandline requires tmt+report-junit = 1.36.dev888-1.20240828170910228891.pr3166.26.g07024055.el9, but none of the providers can be installed
```

LecrisUT:
`pyrhon-lxml` doesn't seem to be available on el9. Maybe you'll have to package it to epel9 or if there are alternatives...

pyproject.toml
```diff
@@ -61,7 +61,9 @@ provision-virtual = [
     ]
 provision-container = []
 report-junit = [
-    "junit_xml>=1.9",
+    # Required to support XML parsing and checking the XSD schemas.
+    "lxml>=4.9.0",
+    "types-lxml",
```
LecrisUT, line 66:
```suggestion
```
Typesheds should not be included. Still pending on https://github.com/teemtee/tmt/pull/2841 for how these will be handled in the future, but for now these are only manually tracked in pre-commit config

tmt/steps/report/junit.py
seberm: I decided to just catch the `tmt.utils.FileError` here. Hopefully, this should be enough. Do you agree? Or should I for some reason catch also the `AttributeError`?

pyproject.toml
seberm: Ok, I've excluded the `types-lxml` here, thanks.

martinhoyer:
`time` is probably not in float? 
```
FAILED test_report_junit.py::TestStateMapping::test_pass - AssertionError: Attribute mismatch at #document.testsuites.testsuite: found time="0.0"
assert '0' == '0.0'
```

seberm:
> `time` is probably not in float?
> 
> ```
> FAILED test_report_junit.py::TestStateMapping::test_pass - AssertionError: Attribute mismatch at #document.testsuites.testsuite: found time="0.0"
> assert '0' == '0.0'
> ```

It was `int` in the `testsuite`:

```
<testsuite disabled="0" errors="0" failures="0" name="name" skipped="0" tests="1" time="0">
```

But it was a float in the `testsuites` tag :) :

```
<testsuites disabled="0" errors="0" failures="0" tests="1" time="0.0">
```

I am perfectly fine with a `time` as an integer. But I think the `time` types should be the same across all these tags.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,229 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+                Path(f'steps/report/junit/schemas/{flavor}.xsd')))
+
+            schema_root: XMLElement = etree.XML(xsd_schema_path.read_bytes())
+            xml_parser_kwargs['schema'] = etree.XMLSchema(schema_root)
+        else:
+            phase.warn(
+                f"The '{CUSTOM_FLAVOR_NAME}' JUnit flavor is used, you are solely responsible "
+                "for the validity of the XML schema.")
+
+            phase.warn(f"The pretty print is always disabled for '{CUSTOM_FLAVOR_NAME}' JUnit "
```
happz, line 240: A warning is emitted, but I was unable to find where `prettify` is set to `False` for custom flavors.

tmt/steps/report/junit.py
happz: Ok, let's see whether we can improve it someday.

tmt/steps/report/junit.py
seberm: Good catch. I've broken the prettify behavior for custom templates when I was fixing the typing problems, I think.

It should work again:
- https://github.com/teemtee/tmt/compare/0ac068fe0636e49ec95f34eeb5a0e1f1c66de6b7..81b54df4f1340833d21fc05d2457273f4aa90edb

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+
+    def set_properties(self, keyval: dict[str, str]) -> None:
+        self._properties = keyval
+
+    def __getattr__(self, name: str) -> Any:
+        """ Returns an attribute of a wrapped ``tmt.Result`` instance """
+        return getattr(self._wrapped, name)
+
+
+def _add_result_properties(result_instance: tmt.Result) -> ResultWrapper:
```
happz, line 98: * this function could be a class method of `ResultWrapper`, it's basically piece of its functionality, isolated outside of the class, and
* maybe it's just my impression, but is there any actuall difference between `_add_result_properties(result)` and `ResultWrapper(result)`? If not, I'd drop this function completely, it does not seem to have added value.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+
+    def __init__(self, wrapped: tmt.Result) -> None:
+        self._wrapped = wrapped
+        self._properties: dict[str, str] = {}
+
+    @property
+    def properties(self) -> list[PropertyDict]:
+        return [{'name': k, 'value': v} for k, v in self._properties.items()]
+
+    def set_properties(self, keyval: dict[str, str]) -> None:
```
happz, line 90: This method does not seem to be called from any place in this PR, is it needed? If it's still necessary, I'd recommend turning it into property setter with `@properties.setter`.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+
+def _add_result_properties(result_instance: tmt.Result) -> ResultWrapper:
+    """ Dynamically decorate the ``tmt.Result`` instance - add ``properties`` attribute """
+    return ResultWrapper(result_instance)
+
+
+class ResultsContext:
+    """
+    A class which keeps the results context (especially the result summary) for
+    JUnit template. It wraps all the ``tmt.Result`` instances into the ``ResultWrapper``.
```
happz, line 106: The first line should bear a short summary, more details should be given on following lines. Also, use `:py:class:` when referring to other classes, Sphinx will turn it into nice links.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+        """ Returns results of tests with error/warn outcome """
+        return [r for r in self._results if r.result in [
+                ResultOutcome.ERROR,
+                ResultOutcome.WARN]]
+
+    @functools.cached_property
+    def duration(self) -> int:
+        """ Returns the total duration of all tests in seconds """
+        durations: list[int] = [duration_to_seconds(r.duration) or 0 for r in self._results]
+        return sum(durations)
```
happz, line 147: `durations` does not seem to be needed as a standalone variable, and as a list, because `sum()` can take an iterable, including a generator: `sum(duration_to_seconds(r.duration) or 0 for r in self._results)`. The list of durations does not need to materialize, ever.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+        Jinja template.
+    """
+
+    # Get the template context for TMT results
+    results_context = results_context or ResultsContext(phase.step.plan.execute.results())
+
+    # Prepare the template environment
+    environment = default_template_environment()
+
+    template_dir = Path('steps/report/junit/templates/')
```
happz, line 181: I'd promote this to module-level "constant", something like `DEFAULT_TEMPLATE_PATH` or something similar. It's an important string constant, hidden in the code.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+                phase.warn(str(err))
+
+            # Return the prettified XML without checking the XSD
+            del xml_parser_kwargs['schema']
+
+            try:
+                tree_root = etree.fromstring(  # noqa: S320
+                    xml_data, etree.XMLParser(**xml_parser_kwargs))
+            except etree.XMLSyntaxError as e:
+                phase.warn(str(e))
```
happz, line 269: This should probably be dropped in favor of `from e` when `ReportError` is raised below. It's not like `e` is just something to mention and move on - `ReportError` will kill this reporting attempt, and `e` is the genuine cause - with the exception of chaining, the cause will be shown with proper significance.

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+                phase.verbose('data', xml_data, 'red')
```
happz, line 270: `data` as a label is often pretty useless because it's so overused. Would there be a different label we could use, maybe `rendered XML` or something along that line?

tmt/steps/report/junit.py
```diff
@@ -58,40 +67,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+            tree_root,
+            xml_declaration=True,
+            pretty_print=prettify,
+            # The 'utf-8' encoding must be used instead of 'unicode', otherwise
+            # the XML declaration is not included in the output.
+            encoding='utf-8')
+
+        return str(xml_output.decode('utf-8'))
+
+    except ImportError:
```
happz, line 289: Hmm, how about wrapping only the import into this `try-except` block? That way the rest of the code which imports nothing could save one level of pointless indentation.

tmt/steps/report/junit.py
```diff
@@ -101,7 +300,30 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
         option='--file',
         metavar='PATH',
         help='Path to the file to store JUnit to.',
-        normalize=lambda key_address, raw_value, logger: Path(raw_value) if raw_value else None)
+        normalize=tmt.utils.normalize_path)
+
+    flavor: str = field(
+        default=DEFAULT_FLAVOR_NAME,
+        option='--flavor',
+        metavar='FLAVOR',
```
happz, line 308: I think you can drop `metavar` - with `choices`, tmt core will turn this into `--flavor [default|custom]`, showing the actual choices.

tmt/steps/report/junit.py
```diff
@@ -101,7 +300,30 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
+        choices=[DEFAULT_FLAVOR_NAME, CUSTOM_FLAVOR_NAME],
+        help=f"Name of a JUnit flavor to generate. By default, the '{DEFAULT_FLAVOR_NAME}' flavor "
+        "is used.")
```
happz, line 311: A multiline string can be used, tmt will dedent it properly. Way better than spurious trailing whitespace :)

tmt/steps/report/junit.py
```diff
@@ -101,7 +300,30 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
+        help='Path to a custom template file to use for JUnit creation.',
+        normalize=tmt.utils.normalize_path)
+
+    prettify: bool = field(
+        default=True,
+        option=('--prettify / --no-prettify'),
+        is_flag=True,
+        show_default=True,
+        help="Enable the XML pretty print for generated JUnit file. This option is always "
+        f"disabled for '{CUSTOM_FLAVOR_NAME}' template flavor.")
```
happz, line 326: Multiline string should be used, see above.

tmt/steps/report/junit.py
```diff
@@ -114,34 +336,49 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
-                    # For older junit-xml
-                    junit_xml.TestSuite.to_file(fw, [suite])
-            self.info("output", f_path, 'yellow')
+                fw.write(xml_data)
+
+            self.info('output', f_path, 'yellow')
         except Exception as error:
-            raise tmt.utils.ReportError(
-                f"Failed to write the output '{f_path}' ({error}).")
+            raise tmt.utils.ReportError(f"Failed to write the output '{f_path}' ({error}).")
```
happz, line 384: Since you're changing this line, let's stop encoding `error` into message, and use proper `raise ... from error`.

tmt/steps/report/junit.py
seberm: This method will be used in the Polarion report plugin:
- https://github.com/teemtee/tmt/pull/3166/files#diff-3f066603b29f42721b98aab30f0d1afac08020970356069510803b03c9b4b234R312

It was originally defined as a `@properties.setter`, but again, I was getting some typing errors that are not showing anymore. I will change it to a property setter in this PR and we can solve possible typing problems in #3166.

tmt/steps/report/junit.py
seberm: Does it make sense to also use the `from e` construction when catching the global `Exception` on these lines?

- https://github.com/teemtee/tmt/blob/81b54df4f1340833d21fc05d2457273f4aa90edb/tmt/steps/report/junit.py#L383
- https://github.com/teemtee/tmt/blob/81b54df4f1340833d21fc05d2457273f4aa90edb/tmt/steps/report/junit.py#L65

tmt/steps/report/junit.py
happz: Sure, I believe we will be able to resolve typing issues.

tmt/steps/report/junit.py
happz: Hm, maybe `self._results` is mistyped, who knows. Let' see what comes out of it, I believe we will be able to sort it out.

tmt/steps/report/junit.py
happz: Yep, absolutely.

tmt/steps/report/junit.py
seberm: > If not, I'd drop this function completely, it does not seem to have added value.

Correct, It's not anymore. Removed, thanks.

tmt/steps/report/junit.py
seberm: I've added `:py:class:` and created short summaries. I hope it's resolved.

tmt/steps/report/junit.py
seberm: I've pushed the change. Could you please take a look and help me resolve the error?

tmt/steps/report/junit.py
seberm: Ok, thanks. It should be resolved now.

tmt/steps/report/junit.py
seberm: Good catch. Thanks.

tmt/steps/report/junit.py
seberm: Should be fixed now.

tmt/steps/report/junit.py
happz:
```diff
diff --git a/tmt/steps/report/junit.py b/tmt/steps/report/junit.py
index db09ab40..f6fe4a1c 100644
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -1,7 +1,7 @@
 import dataclasses
 import functools
 from collections.abc import Iterator
-from typing import TYPE_CHECKING, Any, Optional, TypedDict, overload
+from typing import TYPE_CHECKING, Any, Optional, TypedDict, cast, overload
 
 from jinja2 import FileSystemLoader, select_autoescape
 
@@ -145,7 +145,11 @@ class ResultsContext:
     @functools.cached_property
     def duration(self) -> int:
         """ Returns the total duration of all tests in seconds """
-        return sum(duration_to_seconds(r.duration) or 0 for r in self._results)
+        # cast: mypy does not understand the proxy-ness of `ResultWrapper`. `r.duration`
+        # will exists, therefore adding a `cast` to convince mypy the list is pretty much
+        # nothing but the list of results.
+        return sum(
+            duration_to_seconds(r.duration) or 0 for r in cast(list[tmt.Result], self._results))
 
 
 def make_junit_xml(
```

tmt/steps/report/junit/templates/_base.xml.j2
seberm: I think the `testsuites` properties should be probably part of the `ResultsContext` in a very similar way as for `ResultWrapper`. I would like to implement it probably as a part of #3166 so this PR will make it into v1.36.

tmt/steps/report/junit.py
```diff
@@ -101,7 +299,35 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
...
-        normalize=lambda key_address, raw_value, logger: Path(raw_value) if raw_value else None)
+        normalize=tmt.utils.normalize_path)
+
+    flavor: str = field(
+        default=DEFAULT_FLAVOR_NAME,
+        option='--flavor',
+        choices=[DEFAULT_FLAVOR_NAME, CUSTOM_FLAVOR_NAME],
+        help=f"""
+            Name of a JUnit flavor to generate. By default, the '{DEFAULT_FLAVOR_NAME}' flavor
+            is used.
```
martinhoyer, line 310: This looks weird in the generated documentation:
```
flavor: default|custom

    Name of a JUnit flavor to generate. By default, the ‘default’ flavor is used.

    Default: default
```
```suggestion
            Name of a JUnit flavor to generate.
```

martinhoyer (approved):

KwisatzHaderach (approved):

tmt/steps/report/junit.py
seberm: Thanks, modified.

tests/unit/test_report_junit.py
```diff
@@ -120,11 +120,14 @@ def test_pass(self, report_fix):
...
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="0" failures="0" tests="1" time="0.0">
-    <testsuite disabled="0" errors="0" failures="0" name="name" skipped="0" tests="1" time="0">
-        <testcase name="/pass"/>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="0" failures="0" skipped="0" tests="1" time="0.0">
+    <testcase name="/pass">
+
```
seberm, line 127: @thrix The data content inside the testcase tags cannot be easily removed - even the blank content still counts as data.

From the lxml [docs](https://lxml.de/tutorial.html#parser-objects):

By default, lxml.etree uses a standard parser with a default setup. If you want to configure the parser, you can create a new instance:

```
>>> parser = etree.XMLParser(remove_blank_text=True)  # lxml.etree only!
```

This creates a parser that removes empty text between tags while parsing, which can reduce the size of the tree and avoid dangling tail text if you know that whitespace-only content is not meaningful for your data. An example:

```
>>> root = etree.XML("<root>  <a/>   <b>  </b>     </root>", parser)

>>> etree.tostring(root)
b'<root><a/><b>  </b></root>'
```

Note that the whitespace content inside the `<b>` tag was not removed, as content at leaf elements tends to be data content (even if blank). You can easily remove it in an additional step by traversing the tree:

```
>>> for element in root.iter("*"):
...     if element.text is not None and not element.text.strip():
...         element.text = None

>>> etree.tostring(root)
b'<root><a/><b/></root>'
```

So - we can remove it as suggested in the docs, but I am not convinced we should do it.

happz: Be careful, I'm not sure about it but if we have elements that contain e.g. test output, like `<failure/>`, it may contain whitespace and it might be expected. A post-pocessing like the one mentioned above might harm elements like this.

seberm: >  it may contain whitespace and it might be expected

Yes, exactly. Let's keep it as it is now.

docs/releases.rst
```diff
@@ -14,6 +14,19 @@ file does not follow the :ref:`result specification</spec/plans/results>`.
...
+The :ref:`/plugins/report/junit` report plugin now validates all the XML
+flavors against their respective XSD schemas and tries to prettify the final
+XML output. These functionalities are always disabled for ``custom`` flavors.
+The prettify functionality can be controlled for non-custom templates by
+``--prettify`` and ``--no-prettify`` arguments.
+
+The :ref:`/plugins/report/junit` report plugin now uses Jinja instead of
+``junit-xml`` library to generate the JUnit XMLs. It also adds support for a
+new ``--flavor`` argument. Using this argument the user can choose between a
+``default`` flavor, which keeps the current behavior intouched, and a
```
lukaszachy, line 26:
```suggestion
``default`` flavor, which keeps the current behavior untouched, and a
```

tmt/steps/report/junit.py
```diff
@@ -54,44 +66,230 @@ def duration_to_seconds(duration: Optional[str]) -> Optional[int]:
...
+            Path(f'steps/report/junit/schemas/{flavor}.xsd')))
+
+        schema_root: XMLElement = etree.XML(xsd_schema_path.read_bytes())
+        xml_parser_kwargs['schema'] = etree.XMLSchema(schema_root)
+    else:
+        phase.warn(
+            f"The '{CUSTOM_FLAVOR_NAME}' JUnit flavor is used, you are solely responsible "
+            "for the validity of the XML schema.")
+
+        phase.warn(f"The pretty print is always disabled for '{CUSTOM_FLAVOR_NAME}' JUnit "
```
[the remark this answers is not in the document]
seberm: Sure, modified. Thanks.

tests/unit/test_report_junit.py
```diff
@@ -120,11 +120,14 @@ def test_pass(self, report_fix):
...
         report.go()
 
-        assert_xml(out_file_path, """<?xml version="1.0" ?>
+        assert_xml(out_file_path, """<?xml version='1.0' encoding='utf-8'?>
 <testsuites disabled="0" errors="0" failures="0" tests="1" time="0.0">
-    <testsuite disabled="0" errors="0" failures="0" name="name" skipped="0" tests="1" time="0">
-        <testcase name="/pass"/>
-    </testsuite>
+  <testsuite name="name" disabled="0" errors="0" failures="0" skipped="0" tests="1" time="0.0">
+    <testcase name="/pass">
```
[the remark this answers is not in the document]
seberm: Hello @thrix, please see the comments in https://github.com/teemtee/tmt/pull/3150#discussion_r1741955786, thanks.

tests/unit/test_report_junit.py
thrix: @seberm thanks, what can we do, let's continue with what we have, it is still valid xml :)

Thanks for chekcking

thrix (approved):
LGTM

Issue #3156, opened 2024-08-20
tmt: can't disable lint check G001


bgoncalv:
```
tmt --version
tmt version: 1.35.0
```

```
git clone https://gitlab.com/redhat/centos-stream/tests/kernel/kernel-tests
cd kernel-tests
git checkout 91734f29d9d1e5fc0a9e46d4527454c0c65963fe
tmt lint --outcome-only fail --disable-check G001
Lint checks on all
fail G001 duplicate id "3454c370-0d0f-494f-aef9-b842a53756d9" in "/kdump/analyse-crash-cmd/common_analyse"
fail G001 duplicate id "3454c370-0d0f-494f-aef9-b842a53756d9" in "/kdump/analyse-crash-cmd/simple_analyse"
fail G001 duplicate id "3454c370-0d0f-494f-aef9-b842a53756d9" in "/kdump/analyse-crash-cmd/simple_check"
fail G001 duplicate id "896b41ea-f80b-4aaf-9641-19e3b5b09627" in "/kdump/kexec-boot/basic"
fail G001 duplicate id "896b41ea-f80b-4aaf-9641-19e3b5b09627" in "/kdump/kexec-boot/exec_live"
fail G001 duplicate id "896b41ea-f80b-4aaf-9641-19e3b5b09627" in "/kdump/kexec-boot/memdebug"
fail G001 duplicate id "942f0232-906a-4644-9c5a-10a056845e58" in "/memory/config_debug_wx_test/config_debug_wx_test"
fail G001 duplicate id "942f0232-906a-4644-9c5a-10a056845e58" in "/memory/config_memtest/config_memtest"
fail G001 duplicate id "985f0229-d93c-44d3-8c67-f6b5472bf3d7" in "/security/crypto/17113"
fail G001 duplicate id "985f0229-d93c-44d3-8c67-f6b5472bf3d7" in "/security/crypto/1990040"
fail G001 duplicate id "c8c9ef7d-f618-4450-b379-452da8276732" in "/storage/ssd/nvme_suspend_disk_resume_io"
fail G001 duplicate id "c8c9ef7d-f618-4450-b379-452da8276732" in "/storage/ssd/nvme_suspend_mem_resume_io"
```

martinhoyer:
@happz you're too fast for me

PR #3158, opened 2024-08-20
Fix handling of empty set of allowed linters

happz:
Empty set is not the same as unset, and the test must respect that.

Fixes https://github.com/teemtee/tmt/issues/3156

Pull Request Checklist

* [x] implement the feature

PR #3159, opened 2024-08-20
Fix disabled LintableCollection lint checks being applied

martinhoyer:
Resolves [#3156](https://github.com/teemtee/tmt/issues/3156)  

The problem was that while `tmt.base.LintableCollection.resolve_enabled_linters()` correctly resolved disabled checks, it proceeds to do the diabled checks anyway. 

@engelmi Could you take a look please? If you have a better fix (and test), it would be appreciated ;)


martinhoyer:
oh well, duplicate of https://github.com/teemtee/tmt/pull/3158/

Issue #2884, opened 2024-04-24
tmt cannot run tests on Fedora CoreOS containers


happz:
The `/usr/local` in the container is a symlink to `/var/uslocal` which does not exist. Therefore tmt fails scripts like `/usr/local/bin/tmt-abort`.

```shell
(dev) [pts-3:2]: happz@multivac [main] ~/git/tmt/tests/prepare/install/data $ TMT_SHOW_TRACEBACK=1 tmt -vvv -c distro=fedora-coreos run -a provision --how container --image localhost/fedora-coreos:tmt-tests plan --name escape
/var/tmp/tmt/run-082
Found 1 plan.

/escape
    discover
        how: shell
        order: 50
        summary: 1 test selected
            /script-00
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: container
        order: 50
        image: localhost/fedora-coreos:tmt-tests
        primary address: tmt-082-xykfAYfY
        topology address: tmt-082-xykfAYfY
        Check for container image 'localhost/fedora-coreos:tmt-tests'.
        name: tmt-082-xykfAYfY
        Create network 'tmt-082-network'.
        multihost name: default-0
        arch: x86_64
        distro: Fedora CoreOS 39.20240322.3.0
        kernel: 6.6.13-100.fc38.x86_64
        package manager: dnf5
        selinux: yes
        is superuser: yes
    
        summary: 1 guest provisioned
    prepare
        queued push task #1: push to default-0
        
        push task #1: push to default-0
    
        queued prepare task #1: default-0 on default-0
        queued prepare task #2: requires on default-0
        
        prepare task #1: default-0 on default-0
        how: install
        order: 50
        package: 1 package requested
            dnf-command(copr)
            cmd: rpm -q --whatprovides 'dnf-command(copr)' || dnf5 install -y  'dnf-command(copr)'
            out: no package provides dnf-command(copr)
            out: Updating and loading repositories:
            out:  Fedora 39 - x86_64 - Updates           100% | 674.0 KiB/s |   8.7 MiB |  00m13s
            out:  Fedora 39 openh264 (From Cisco) - x86_ 100% |   4.1 KiB/s |   6.6 KiB |  00m02s
            out:  Fedora 39 - x86_64                     100% | 691.3 KiB/s |  35.6 MiB |  00m53s
            out:  Fedora 39 - x86_64 - Updates Archive   100% |   2.7 MiB/s |  31.3 MiB |  00m12s
[...]

PR #3161, opened 2024-08-21
Implement the `bootc` provision plugin

### LINKED ISSUE #3013: Create a podman-bootc provisioner
Seems now there is a simple way to run bootable linux containers based on bootc via podman.

ckyrouac:
This creates a new provision plugin that is built on top of the existing TestCloud (virtual) plugin. It adds new parameters to pass a Containerfile or container image. The plugin will then build a container image (if necessary) then build a bootc disk image from the container image using bootc image builder. Currently, bootc requires podman to be run as root when building a disk image. This is typically handled by running a podman machine as root.

An additional parameter "add-deps" toggles building a derived container image with the tmt test requirements.

Resolves #3013 

Pull Request Checklist

* [X] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] include a release note

---
This is a work in progress. I'm opening this PR early to get feedback on the high level design. I will add tests, docs, etc. after we solidify the higher level design.

If you want to try running the code here is an example fmf plan:

```
provision:
  how: bootc
  containerimage: quay.io/centos-bootc/centos-bootc:stream9
  disk: 20
summary: Testing bootc plugin
execute:
  how: tmt
  script: |
    echo "ok"
```


tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,157 @@
...
+            "podman", "build", f'{self.workdir}',
+            "-f", f'{self.workdir}/Containerfile',
+            "-t", image_tag
+            ).run(cwd=self.workdir, stream_output=True, logger=self._logger)
+
+        return image_tag
+
+    # build the "base" or user supplied container image
+    def _build_base_image(self, containerfile: str, workdir: str) -> str:
+        image_tag = "localhost/tmtbase"  # TODO: how do I get a unique ID here?
```
ckyrouac, line 104: I think using the tmt run # will work for this. Is there a good way to access it here? Or is there another unique ID I could use for the image tag?

ckyrouac:
@cgwalters fyi

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,157 @@
...
+            containerfile: file://Containerfile
+
+    """
+
+    _data_class = BootcData
+    _guest_class = GuestTestcloud
+    _guest = None
+
+    # build a "modified" container image from the base image with tmt dependencies added
+    def _build_derived_image(self, base_image: str) -> str:
```
ckyrouac, line 79: I went this route instead of using podman-bootc (to leverage smbios) primarily because podman-bootc is not yet released in the official repos (it's in [German's copr repo](https://github.com/containers/podman-bootc?tab=readme-ov-file#fedora)).

ckyrouac:
@cgwalters fyi

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,157 @@
...
+        self._logger.debug("Building modified container image with necessary tmt packages/config")
+        containerfile_template = Template('''
+            FROM $base_image
+
+            RUN \
+            dnf -y install cloud-init rsync && \
+            ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants && \
```
cgwalters, line 86: Minor nit, missing the `dnf clean all` so we're leaking unnecessary cruft into `/var`; to make this sustainable long term we probably need to go to a declarative input tool, cc https://gitlab.com/fedora/bootc/osbuild-cfg

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # build the bootc disk
+    def _build_bootc_disk(self, containerimage: str) -> None:
+        self._logger.debug("Building bootc disk image")
+        tmt.utils.Command(
+            "podman", "run", "--rm", "--privileged",
+            "-v", "/var/lib/containers/storage:/var/lib/containers/storage",
+            "--security-opt", "label=type:unconfined_t",
+            "-v", f"{self.workdir}:/output",
+            "quay.io/centos-bootc/bootc-image-builder:latest", "build",
```
cgwalters, line 121: I've been trying to avoid people hardcoding this because we already have the situation where there are two builders (one upstream, one for RHEL). Exact same comment as https://github.com/containers/podman-bootc/pull/58#discussion_r1668661289

It's actually somewhat likely that we move out of the quay.io/centos-bootc namespace at some point; while we'd probably maintain a redirect for "a while", with the "consume a label" approach it automatically derives from the base image which avoids hardcoding entirely.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,157 @@
...
+    # build a "modified" container image from the base image with tmt dependencies added
+    def _build_derived_image(self, base_image: str) -> str:
+        self._logger.debug("Building modified container image with necessary tmt packages/config")
+        containerfile_template = Template('''
+            FROM $base_image
+
+            RUN \
+            dnf -y install cloud-init rsync && \
+            ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants && \
+            rm /usr/local -rf && ln -sr /var/usrlocal /usr/local && mkdir -p /var/usrlocal/bin
```
cgwalters, line 87: We should also probably add a TODO somewhere for tmt to put its temporary binaries somewhere other than `/usr/local` too.

cgwalters (commented):
Mostly a skim (I don't know the tmt codebase either) but looks sane to me!

tmt/steps/provision/bootc.py
happz: How unique does the "unique ID" need to be? Do you wish it to be the same when the user re-runs the previous run, or should it be unique every time `tmt run` is invoked?

tmt/steps/provision/bootc.py
happz: I believe it might be this issue, spawned by slightly different conditions, but maybe hitting the same wall? https://github.com/teemtee/tmt/issues/2884

tmt/steps/provision/bootc.py
happz: Would it make sense to make the namespace configurable? In any case, having a default as a module-level constant should happen, but is it something the user might wish to override? E.g. `namespace: quay.io/acme/bootc`? I think it would be good to open this piece to users.

happz:
* LGTM, I like the reuse of existing plugins, that's very nice.
* Would it also make sense to support the `container` plugin? IIUIC, the image produced by `podman` can be run both with `podman run`, or, converted to qcow2, as a VM. I could imagine both ways being available to use, maybe through some `runtime-plugin: container|virtual` switch. Is it a bad idea?
* I will have more low-level comments related to the actual implementation, but I won't bother you now till you receive the high-level review. Just ping when you're ready for the boring stuff :)

martinhoyer (commented):
I'm still trying to speedrun reading/learning all things bootc (and tmt provisioning plugins), but fwiw, looks cool to me.  

@happz about the container plugin support - [Fedora docs](https://docs.fedoraproject.org/en-US/bootc/getting-started/#_local_development) says:  

> for fully-fledged tests it is not recommended to run a bootable container via, for instance, podman-run. One reason among others is that the filesystem is writable when being executed as an OCI container while most of the filesystem is mounted read-only on a deployed bootc system. That means the running container behaves differently than a deployed system. Yet, if you desire to run some quick tests it is recommended to run the container in detached mode.  

 From what I understand, `podman-bootc` could be pretty cool to use, once available.

happz:
> I'm still trying to speedrun reading/learning all things bootc (and tmt provisioning plugins), but fwiw, looks cool to me.
> 
> @happz about the container plugin support - [Fedora docs](https://docs.fedoraproject.org/en-US/bootc/getting-started/#_local_development) says:
> 
> > for fully-fledged tests it is not recommended to run a bootable container via, for instance, podman-run. One reason among others is that the filesystem is writable when being executed as an OCI container while most of the filesystem is mounted read-only on a deployed bootc system. That means the running container behaves differently than a deployed system. Yet, if you desire to run some quick tests it is recommended to run the container in detached mode.
> 
> From what I understand, `podman-bootc` could be pretty cool to use, once available.

I agree, it's not a perfect 1:1 substitution, but, exactly: for quick tests or basic test development, it may give me results faster than VM. I for one work on binutils and C/C++ toolchain in general, and my area of focus is fairly simple - compile this, run `objdump` on that, grep for expected section names, this kind of stuff. Learning about a typo in my reproducer quickly is very valuable, and once I'm done, I can always use full VM. I do it today already: I develop tests with `container`, then I switch to `beaker` or `1minutetip` to get a more real environment before committing the new test to git. So, `podman-run`, for my trivial component, would be very welcome, even with caveats like the one you mentioned :)

tmt/steps/provision/bootc.py
ckyrouac: I think keeping the ID the same when re-running the previous run will work. If I understand how tmt runs work, this will allow modifying the test code without rebuilding the bootc disk which can take a few minutes.

martinhoyer:
> I agree, it's not a perfect 1:1 substitution, but, exactly: for quick tests or basic test development, it may give me results faster than VM. I for one work on binutils and C/C++ toolchain in general, and my area of focus is fairly simple - compile this, run `objdump` on that, grep for expected section names, this kind of stuff. Learning about a typo in my reproducer quickly is very valuable, and once I'm done, I can always use full VM. I do it today already: I develop tests with `container`, then I switch to `beaker` or `1minutetip` to get a more real environment before committing the new test to git. So, `podman-run`, for my trivial component, would be very welcome, even with caveats like the one you mentioned :)

Thank you for the insight in your development process. I hope I can see it in more detail one day.

tmt/steps/provision/bootc.py
ckyrouac: good catch, updated in latest push

tmt/steps/provision/bootc.py
ckyrouac: added a parameter for the user to configure this in the latest push

ckyrouac:
> Would it also make sense to support the container plugin? IIUIC, the image produced by podman can be run both with podman run, or, converted to qcow2, as a VM. I could imagine both ways being available to use, maybe through some runtime-plugin: container|virtual switch. Is it a bad idea?

I think the existing container plugin will handle this case without any additional code. The bootc image is just another container that can be run like a typical image.

lukaszachy (commented):
`--add-deps` doesn't install require/recommend yet, right? 

Either way, we have chicken&egg problem in the case 'dist-git-source' is used as the list of packages is known _after_ provision :/ But IMO we can ignore that for now to have something working.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+
+            RUN \
+            dnf -y install cloud-init rsync && \
+            ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants && \
+            rm /usr/local -rf && ln -sr /var/usrlocal /usr/local && mkdir -p /var/usrlocal/bin && \
+            dnf clean all
+        ''')
+        containerfile = containerfile_template.substitute({"base_image": base_image})
+        with open(f'{self.workdir}/Containerfile', 'w') as file:
+            file.write(containerfile)
```
happz, line 164: Could be simplified a bit, we're slowly transitioning to how `pathlib` lets us work with files, `${path}.write_text()` should work like a charm:

```python
(self.workdir / 'Containerfile').write_text(containerfile_template.substitute({"base_image": base_image}))
```

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+                containerimage = self._build_derived_image(data.containerimage)
+            self._build_bootc_disk(containerimage, data.image_builder)
+        elif data.containerfile is not None:
+            containerimage = self._build_base_image(data.containerfile, data.containerfile_workdir)
+            if data.add_deps:
+                containerimage = self._build_derived_image(containerimage)
+            self._build_bootc_disk(containerimage, data.image_builder)
+        else:
+            self._logger.fail("Either containerfile or containerimage must be specified.")
+            raise SystemExit(1)
```
happz, line 220: This is definitely worth an exception, e.g. `raise tmt.utils.ProvisionError("Either containerfile or containerimage must be specified.")`. tmt has some basic exception rendering, please leave the reporting and exit strategy to tmt core, it will be fine :)

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
+        """ Provision the bootc instance """
+        super().go(logger=logger)
+
+        data = BootcData.from_plugin(self)
+        data.set_image(f"file://{self.workdir}/qcow2/disk.qcow2")
+        data.show(verbose=self.verbosity_level, logger=self._logger)
+
+        containerimage = ""
```
happz, line 207: Please, don't use an empty string to represent "unset" or "undefined" or "undecided" value. It's very hard for linters to distinguish from strings that have value, that are all the same to linters. `None` is a more easier to track and much better when it comes to report undefined/unset/undecided/unknown value is passed where defined/set/decided/known value is expected.

IIUIC, eventually this name will carry a valid value, either from `data.containerimage` or from `self._build_base_image()`, therefore I'd drop this declaration completely, `containerimage = data.containerimage` below should set it well-enough for both runtime and linters to understand it.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+
+    _data_class = BootcData
+    _guest_class = GuestTestcloud
+    _guest = None
+    _id = str(uuid.uuid4())[:8]
+
+    # build a "modified" container image from the base image with tmt dependencies added
+    def _build_derived_image(self, base_image: str) -> str:
+        self._logger.debug("Building modified container image with necessary tmt packages/config")
+        containerfile_template = Template('''
```
happz, line 153: Please, use Jinja templates, it's the solution used throughout the tmt code. See https://github.com/teemtee/tmt/blob/main/tmt/utils/templates.py#L314

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+    .. code-block:: shell
+
+        podman machine init --rootful --disk-size 200 --memory 8192 \
+        --cpus 8 -v /var/tmp/tmt:/var/tmp/tmt -v $HOME:$HOME
+    """
+
+    _data_class = BootcData
+    _guest_class = GuestTestcloud
+    _guest = None
+    _id = str(uuid.uuid4())[:8]
```
happz, line 148: What is the purpose of this id? Would it be possible to use `Run.unique_id`, https://github.com/teemtee/tmt/blob/main/tmt/base.py#L3317? It should be available somewhere from `self.parent.step.my_run.unique_id`.

We should think twice before adding new kind ID, especially when it's with no clear connection to run owning images being built, might be much harder to locate them when removing things.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+    image_builder: str = field(
+        default=DEFAULT_IMAGE_BUILDER,
+        option=('--image-builder'),
+        metavar='IMAGEBUILDER',
+        help="""
+             The full repo:tag url of the bootc image builder image to use for
+             building the bootc disk image.
+             """)
+
+    def set_image(self, image: str) -> None:
```
happz, line 92: This kind of setters is not necessary, `self.image = image` is perfectly acceptable.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+                ).run(cwd=self.workdir, stream_output=True, logger=self._logger)
+
+        super().remove()
+
+
+@dataclasses.dataclass
+class BootcData(tmt.steps.provision.testcloud.ProvisionTestcloudData):
+    containerfile: Optional[str] = field(
+        default=None,
+        option=('--containerfile'),
```
happz, line 47: `()` wrapping is not needed when there is only one option.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+                 parent: Optional[tmt.utils.Common] = None,
+                 logger: tmt.log.Logger,
+                 containerimage: Optional[str]) -> None:
+        super().__init__(data=data, logger=logger, parent=parent, name=name)
+
+        if containerimage:
+            self.containerimage = containerimage
+
+    def remove(self) -> None:
+        if self.containerimage:
```
happz, line 35: I'm not 100 % sure, but if `self.containerimage` was not assigned a value in `__init__()`, this test might fail because Python would report a missing attribute. It would be better to mark `containerimage` as `Optional[str]` and give it `None` as a default - it obviously can be irrelevant, `None` is easier for linters to track and prevent the use.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+class GuestBootc(GuestTestcloud):
+    containerimage: str
+
+    def __init__(self,
+                 *,
+                 data: tmt.steps.provision.GuestData,
+                 name: Optional[str] = None,
+                 parent: Optional[tmt.utils.Common] = None,
+                 logger: tmt.log.Logger,
+                 containerimage: Optional[str]) -> None:
```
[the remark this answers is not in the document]
ckyrouac: ah, yea that’s right. I had originally marked containerimage as optional because I had set `_guest_class = GuestBootc` which caused the `GuestBootc` class to be instantiated without a `containerimage`. Using `GuestBootc` as the `_guest_class` was causing some other odd issues (the tests would pass when running them individually, but when running multiple tests only the first would pass) so I flipped it back to GuestTestcloud which seems to work as expected.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,232 @@
...
+            "-t", image_tag
+            ).run(cwd=self.workdir, stream_output=True, logger=self._logger)
+        return image_tag
+
+    # build the bootc disk
+    def _build_bootc_disk(self, containerimage: str, image_builder: str) -> None:
+        self._logger.debug("Building bootc disk image")
+        tmt.utils.Command(
+            "podman", "run", "--rm", "--privileged",
+            "-v", "/var/lib/containers/storage:/var/lib/containers/storage",
```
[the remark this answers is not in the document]
ckyrouac: Moved the string to a constant at the top of the module. This is the default location for podman container storage. It can be changed in podman configs. I'm not sure it's worth adding an option to change it until we find it necessary. If you think we should add the option up front that works too.

tmt/steps/provision/bootc.py
ckyrouac: Thanks, I wasn't sure how to access the unique run id so I just generated one. It is appended to the container images tags to keep them unique for cleanup, etc. I switched to using `self.parent.step.my_run.unique_id`.

ckyrouac:
@happz thanks for the review! I believe I addressed all your suggestions.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+
+    _data_class = BootcData
+    _guest_class = GuestTestcloud
+    _guest = None
+    _id = str(uuid.uuid4())[:8]
+
+    def _get_id(self) -> str:
+        # FIXME: cast() - https://github.com/teemtee/tmt/issues/1372
+        parent = cast(tmt.steps.provision.Provision, self.parent)
```
ckyrouac, line 149: copied this from https://github.com/teemtee/tmt/blob/21c9e9793f5c81a6546eb2da5abeec87efcc7685/tmt/steps/provision/__init__.py#L789

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,241 @@
...
+import tmt.steps
+import tmt.steps.provision
+import tmt.steps.provision.testcloud
+import tmt.utils
+from tmt.steps.provision.testcloud import GuestTestcloud
+from tmt.utils import field
+from tmt.utils.templates import render_template
+
+DEFAULT_IMAGE_BUILDER = "quay.io/centos-bootc/bootc-image-builder:latest"
+CONTAINER_STORAGE_DIR = "/var/lib/containers/storage"
```
happz, line 17: This should be `Path` instance right here and now. We try to wrap path-like strings with `tmt.utils.Path` as soon as they enter tmt internals.

tmt/steps/provision/bootc.py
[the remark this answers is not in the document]
happz: I'm still convinced this can be improved: given the code in `go()`, `containerimage` will never be `None`. It will either be a string returned by `_build_derived_image()`, or a string returned by `_build_base_image()`, in any case `containerimage` variable in `go()` will always contain a valid string. Therefore this parameter does not need to be `Optional`, and the test below is also superfluous.

tmt/steps/provision/bootc.py
happz: The attribute is no longer used, and it should be removed.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,241 @@
...
+        assert parent.plan is not None
+        assert parent.plan.my_run is not None
+        assert parent.plan.my_run.unique_id is not None
+        return parent.plan.my_run.unique_id
+
+    def _build_derived_image(self, base_image: str) -> str:
+        """ Build a "derived" container image from the base image with tmt dependencies added """
+        if not self.workdir:
+            raise tmt.utils.ProvisionError(
+                "self.workdir must be defined")
```
happz, line 159: To an end user, this error message is not very helpful. IIUIC, `assert self.workdir is not None  # narrow type` should be good enough, it's a common pattern that `self.workdir` remains `Optional`.

tmt/steps/provision/bootc.py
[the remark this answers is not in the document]
happz: Hm, I'm not familiar enough with `podman` & co. to tell how often one might need to change this path. I suppose it's safe for now to be marked as a "constant", we can always add an option later if it becomes clear there's a use case for it.

martinhoyer:
@happz let's rebase and run the pipeline here?

thrix:
We agreed to focus on getting this in 1.38, the due date for finishing is 24th October.

ckyrouac:
thanks for looking at this and planning for 1.38. Is there anything I can do to help move it along?

thrix:
> thanks for looking at this and planning for 1.38. Is there anything I can do to help move it along?

@ckyrouac not really, the team is overhelmed with reviews.

I am getting validation errors, I assume the jsonschema needs an update for the new provision plugin:

```
❯ pwd
/var/home/mvadkert/git/github.com/teemtee/tmt/tests/provision/bootc/data

❯ tmt lint
    warn: /containerfile_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/includes_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20} is not valid under any of the given schemas
    warn: /containerfile_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/needs_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20} is not valid under any of the given schemas
    warn: /image_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerimage': 'localhost/tmt-bootc-includes-deps', 'disk': 20} is not valid under any of the given schemas
    warn: /image_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerimage': 'localhost/tmt-bootc-needs-deps', 'disk': 20} is not valid under any of the given schemas
/containerfile_includes_deps
warn C000 key "add-deps" not recognized by schema /schemas/provision/container
warn C000 key "containerfile" not recognized by schema /schemas/provision/container
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/container
warn C000 key "disk" not recognized by schema /schemas/provision/container
warn C000 value of "how" is not "container"
warn C000 key "add-deps" not recognized by schema /schemas/provision/mrack
warn C000 key "containerfile" not recognized by schema /schemas/provision/mrack
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/mrack
warn C000 key "disk" not recognized by schema /schemas/provision/mrack
warn C000 value of "how" is not "beaker"
warn C000 key "add-deps" not recognized by schema /schemas/provision/virtual
warn C000 key "containerfile" not recognized by schema /schemas/provision/virtual
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/virtual
warn C000 value of "how" is not "virtual"
warn C000 key "add-deps" not recognized by schema /schemas/provision/artemis
warn C000 key "containerfile" not recognized by schema /schemas/provision/artemis
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/artemis
warn C000 key "disk" not recognized by schema /schemas/provision/artemis
warn C000 value of "how" is not "artemis"
warn C000 key "add-deps" not recognized by schema /schemas/provision/minute
warn C000 key "containerfile" not recognized by schema /schemas/provision/minute
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/minute
warn C000 key "disk" not recognized by schema /schemas/provision/minute
warn C000 value of "how" is not "minute"
warn C000 key "add-deps" not recognized by schema /schemas/provision/connect
warn C000 key "containerfile" not recognized by schema /schemas/provision/connect
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/connect
warn C000 key "disk" not recognized by schema /schemas/provision/connect
warn C000 value of "how" is not "connect"
warn C000 key "add-deps" not recognized by schema /schemas/provision/local
warn C000 key "containerfile" not recognized by schema /schemas/provision/local
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/local
warn C000 key "disk" not recognized by schema /schemas/provision/local
warn C000 value of "how" is not "local"
warn C000 fmf node failed schema validation
pass P004 discover step methods are all known

/containerfile_needs_deps
warn C000 key "add-deps" not recognized by schema /schemas/provision/container
warn C000 key "containerfile" not recognized by schema /schemas/provision/container
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/container
warn C000 key "disk" not recognized by schema /schemas/provision/container
warn C000 value of "how" is not "container"
warn C000 key "add-deps" not recognized by schema /schemas/provision/mrack
warn C000 key "containerfile" not recognized by schema /schemas/provision/mrack
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/mrack
warn C000 key "disk" not recognized by schema /schemas/provision/mrack
warn C000 value of "how" is not "beaker"
warn C000 key "add-deps" not recognized by schema /schemas/provision/virtual
warn C000 key "containerfile" not recognized by schema /schemas/provision/virtual
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/virtual
warn C000 value of "how" is not "virtual"
warn C000 key "add-deps" not recognized by schema /schemas/provision/artemis
warn C000 key "containerfile" not recognized by schema /schemas/provision/artemis
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/artemis
warn C000 key "disk" not recognized by schema /schemas/provision/artemis
warn C000 value of "how" is not "artemis"
warn C000 key "add-deps" not recognized by schema /schemas/provision/minute
warn C000 key "containerfile" not recognized by schema /schemas/provision/minute
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/minute
warn C000 key "disk" not recognized by schema /schemas/provision/minute
warn C000 value of "how" is not "minute"
warn C000 key "add-deps" not recognized by schema /schemas/provision/connect
warn C000 key "containerfile" not recognized by schema /schemas/provision/connect
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/connect
warn C000 key "disk" not recognized by schema /schemas/provision/connect
warn C000 value of "how" is not "connect"
warn C000 key "add-deps" not recognized by schema /schemas/provision/local
warn C000 key "containerfile" not recognized by schema /schemas/provision/local
warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/local
warn C000 key "disk" not recognized by schema /schemas/provision/local
warn C000 value of "how" is not "local"
warn C000 fmf node failed schema validation
pass P004 discover step methods are all known

/image_includes_deps
warn C000 key "add-deps" not recognized by schema /schemas/provision/container
warn C000 key "containerimage" not recognized by schema /schemas/provision/container
warn C000 key "disk" not recognized by schema /schemas/provision/container
warn C000 value of "how" is not "container"
warn C000 key "add-deps" not recognized by schema /schemas/provision/mrack
warn C000 key "containerimage" not recognized by schema /schemas/provision/mrack
warn C000 key "disk" not recognized by schema /schemas/provision/mrack
warn C000 value of "how" is not "beaker"
warn C000 key "add-deps" not recognized by schema /schemas/provision/virtual
warn C000 key "containerimage" not recognized by schema /schemas/provision/virtual
warn C000 value of "how" is not "virtual"
warn C000 key "add-deps" not recognized by schema /schemas/provision/artemis
warn C000 key "containerimage" not recognized by schema /schemas/provision/artemis
warn C000 key "disk" not recognized by schema /schemas/provision/artemis
warn C000 value of "how" is not "artemis"
warn C000 key "add-deps" not recognized by schema /schemas/provision/minute
warn C000 key "containerimage" not recognized by schema /schemas/provision/minute
warn C000 key "disk" not recognized by schema /schemas/provision/minute
warn C000 value of "how" is not "minute"
warn C000 key "add-deps" not recognized by schema /schemas/provision/connect
warn C000 key "containerimage" not recognized by schema /schemas/provision/connect
warn C000 key "disk" not recognized by schema /schemas/provision/connect
warn C000 value of "how" is not "connect"
warn C000 key "add-deps" not recognized by schema /schemas/provision/local
warn C000 key "containerimage" not recognized by schema /schemas/provision/local
warn C000 key "disk" not recognized by schema /schemas/provision/local
warn C000 value of "how" is not "local"
warn C000 fmf node failed schema validation
pass P004 discover step methods are all known

/image_needs_deps
warn C000 key "add-deps" not recognized by schema /schemas/provision/container
warn C000 key "containerimage" not recognized by schema /schemas/provision/container
warn C000 key "disk" not recognized by schema /schemas/provision/container
warn C000 value of "how" is not "container"
warn C000 key "add-deps" not recognized by schema /schemas/provision/mrack
warn C000 key "containerimage" not recognized by schema /schemas/provision/mrack
warn C000 key "disk" not recognized by schema /schemas/provision/mrack
warn C000 value of "how" is not "beaker"
warn C000 key "add-deps" not recognized by schema /schemas/provision/virtual
warn C000 key "containerimage" not recognized by schema /schemas/provision/virtual
warn C000 value of "how" is not "virtual"
warn C000 key "add-deps" not recognized by schema /schemas/provision/artemis
warn C000 key "containerimage" not recognized by schema /schemas/provision/artemis
warn C000 key "disk" not recognized by schema /schemas/provision/artemis
warn C000 value of "how" is not "artemis"
warn C000 key "add-deps" not recognized by schema /schemas/provision/minute
warn C000 key "containerimage" not recognized by schema /schemas/provision/minute
warn C000 key "disk" not recognized by schema /schemas/provision/minute
warn C000 value of "how" is not "minute"
warn C000 key "add-deps" not recognized by schema /schemas/provision/connect
warn C000 key "containerimage" not recognized by schema /schemas/provision/connect
warn C000 key "disk" not recognized by schema /schemas/provision/connect
warn C000 value of "how" is not "connect"
warn C000 key "add-deps" not recognized by schema /schemas/provision/local
warn C000 key "containerimage" not recognized by schema /schemas/provision/local
warn C000 key "disk" not recognized by schema /schemas/provision/local
warn C000 value of "how" is not "local"
warn C000 fmf node failed schema validation
pass P004 discover step methods are all known

    warn: /containerfile_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/includes_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20, 'name': 'default-0'} is not valid under any of the given schemas
    warn: /containerfile_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/needs_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20, 'name': 'default-0'} is not valid under any of the given schemas
    warn: /image_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerimage': 'localhost/tmt-bootc-includes-deps', 'disk': 20, 'name': 'default-0'} is not valid under any of the given schemas
    warn: /image_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerimage': 'localhost/tmt-bootc-needs-deps', 'disk': 20, 'name': 'default-0'} is not valid under any of the given schemas
Lint checks on all
```

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,242 @@
...
+            "-f", self._expand_path(containerfile),
+            "-t", image_tag
+            ).run(cwd=self.workdir, stream_output=True, logger=self._logger)
+        return image_tag
+
+    def _build_bootc_disk(self, containerimage: str, image_builder: str) -> None:
+        """ Build the bootc disk from a container image using bootc image builder """
+        self._logger.debug("Building bootc disk image")
+        tmt.utils.Command(
+            "podman", "run", "--rm", "--privileged",
```
thrix, line 201: Whould this not run under root user for it to work? I get this error on my laptop when trying to run it:

```
❯ pwd
/var/home/mvadkert/git/github.com/teemtee/tmt/tests/provision/bootc/data
❯ tmt run -vvv -e TMT_BOOTC_CONTAINERFILE_RUNDIR=.
/var/home/mvadkert/.local/share/tmt/run-002
    warn: /containerfile_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/includes_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20} is not valid under any of the given schemas
    warn: /containerfile_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/needs_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20} is not valid under any of the given schemas
    warn: /image_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerimage': 'localhost/tmt-bootc-includes-deps', 'disk': 20} is not valid under any of the given schemas
    warn: /image_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerimage': 'localhost/tmt-bootc-needs-deps', 'disk': 20} is not valid under any of the given schemas
Found 4 plans.

/containerfile_includes_deps
    discover
        how: shell
        order: 50
        summary: 1 test selected
            /booted image
    provision
        
        how: bootc
        order: 50
        name: default-0
        how: bootc
        image: file:///var/home/mvadkert/.local/share/tmt/run-002/containerfile_includes_deps/provision/default-0/qcow2/disk.qcow2
        containerfile: ./includes_deps.containerfile
        add deps: False
        memory: 2048 MB
        disk: 20 GB
        fail: Command 'podman run --rm --privileged -v /var/lib/containers/storage:/var/lib/containers/storage --security-opt label=type:unconfined_t -v /var/home/mvadkert/.local/share/tmt/run-002/containerfile_includes_deps/provision/default-0:/output quay.io/centos-bootc/bootc-image-builder:latest build --type qcow2 --local localhost/tmtbase-1729527982' returned 1.
    finish

plan failed

Cause number 1:

    provision step failed

    Cause number 1:

        Command 'podman run --rm --privileged -v /var/lib/containers/storage:/var/lib/containers/storage --security-opt label=type:unconfined_t -v /var/home/mvadkert/.local/share/tmt/run-002/containerfile_includes_deps/provision/default-0:/output quay.io/centos-bootc/bootc-image-builder:latest build --type qcow2 --local localhost/tmtbase-1729527982' returned 1.

        stderr (8/8 lines)
        Trying to pull quay.io/centos-bootc/bootc-image-builder:latest...
        Getting image source signatures
        Copying blob sha256:311336741819859e92064b81199a47397788d736d764f561a89cba9583600387
        Copying blob sha256:35056b112ca2a75be2e91b1076785d9712b1790aea49ebfd61f1799d090d9057
        Copying blob sha256:2edf720f10afc7137e2387bdb3ba800eec2602ac16da698ec19642a4d6fbc4dd
        Copying config sha256:767a10b2f63a72d479d9b102c2016790c63b045f6b9533c639fed605c299ee14
        Writing manifest to image destination
        2024/10/21 16:27:47 error: cannot validate the setup: this command must be run in rootful (not rootless) podman
```

per:

https://github.com/osbuild/bootc-image-builder?tab=readme-ov-file#building

And we will need to put it under the `--feeling-safe` option.

Personally I feel this might a bit limit the usability here, I am wondering if it would not make more sense to spin a VM and run this command in it, but yeah, that will make this a lot harder to do, but avoid the `root` requirement of this feature.

thrix: We can do that later, any idea if this privileged root requirement will ever go away?

ckyrouac: currently, creating the bootc disk requires a loopback mount which is why root is required. I don't think that requirement will go away anytime soon, although it's a common pain point so it might happen sooner rather than later.

> Personally I feel this might a bit limit the usability here, I am wondering if it would not make more sense to spin a VM and run this command in it

A common pattern is to run podman-machine (a VM) to enable rootful podman. Then default all podman commands to use the machine, e.g.

```
podman machine init --rootful --disk-size 200 --memory 8192 --cpus 8 -v /var/tmp/tmt:/var/tmp/tmt -v $HOME:$HOME
export CONTAINER_CONNECTION=podman-machine-default-root
podman run ...
```

ckyrouac: this is documented in the PR here: https://github.com/teemtee/tmt/pull/3161/files#diff-f05fc41435a3de82fd67c56c9bb94d0b781ddaf4182fdf907b4d02ce3531476eR129

thrix: @ckyrouac can you please incorporate `podman machine init` so we do not need to care about it, i.e for the time of building the image, it would be done automatically, not requiring any user action.

thrix: @ckyrouac could we detect if running as root and have a default based on that? Seems like a good middle ground. And make an option to be able to force whatever if that is what the user wants?

thrix:
> I am getting validation errors, I assume the jsonschema needs an update for the new provision plugin:
> 
> ```
> ❯ pwd
> /var/home/mvadkert/git/github.com/teemtee/tmt/tests/provision/bootc/data
> 
> ❯ tmt lint
>     warn: /containerfile_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/includes_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20} is not valid under any of the given schemas
>     warn: /containerfile_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerfile': '$TMT_BOOTC_CONTAINERFILE_RUNDIR/needs_deps.containerfile', 'containerfile-workdir': '.', 'disk': 20} is not valid under any of the given schemas
>     warn: /image_includes_deps:provision - {'how': 'bootc', 'add-deps': False, 'containerimage': 'localhost/tmt-bootc-includes-deps', 'disk': 20} is not valid under any of the given schemas
>     warn: /image_needs_deps:provision - {'how': 'bootc', 'add-deps': True, 'containerimage': 'localhost/tmt-bootc-needs-deps', 'disk': 20} is not valid under any of the given schemas
> /containerfile_includes_deps
> warn C000 key "add-deps" not recognized by schema /schemas/provision/container
> warn C000 key "containerfile" not recognized by schema /schemas/provision/container
> warn C000 key "containerfile-workdir" not recognized by schema /schemas/provision/container
> warn C000 key "disk" not recognized by schema /schemas/provision/container
> warn C000 value of "how" is not "container"
> ... snip ....
> ```

@ckyrouac FTR the respective schemas are in https://github.com/teemtee/tmt/tree/bc64a2fd569c2d1179cb63bd50dbefff429d4f14/tmt/schemas/provision, you will need to add one for this new provision plugin

happz:
Bumping to 1.39: we decided to shorten the 1.38 release cycle and ship it tomorrow to deliver two important features. 1.39 should be also shortened, taking just 2 weeks.

thrix:
@ckyrouac Let us know if you would need any help and I promise I will be more timely on the reviews. Thanks for bringing this feature to `tmt` and sorry this took so long.

ckyrouac:
@thrix no worries. I pushed all the requested changes in a few commits to make it easier to review. I can squash after review if you want.

martinhoyer (approved):
Very nice, thanks!  
In my testing, I've hit https://github.com/containers/podman/issues/23127 and wondering if we could print a hint for users to install `podman-machine`. Perhaps we could also add it as "Recommends" to `tmt+provision-virtual` subpackage? Not sure.

Other thing I've noticed is that bootc provisioning does not 'require' `'tmt+provision-container' ` and `tmt+provision-virtual`. i.e. it should, afaik, fail with:  
`Install 'tmt+provision-container' to provision using this method.`  
while it instead fails on:
`[Errno 2] No such file or directory: 'podman'`

tmt/schemas/provision/bootc.yaml
```diff
@@ -0,0 +1,71 @@
...
+    type: string
+
+  containerfile_workdir: 
+    type: string
+
+  containerimage:
+    type: string
+
+  add_deps:
+    type: bool
```
happz, line 65:
```suggestion
    type: bool
    type: boolean
```

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,311 @@
...
+from tmt.utils.templates import render_template
+
+DEFAULT_IMAGE_BUILDER = "quay.io/centos-bootc/bootc-image-builder:latest"
+CONTAINER_STORAGE_DIR = tmt.utils.Path("/var/lib/containers/storage")
+
+PODMAN_MACHINE_NAME = 'podman-machine-tmt'
+PODMAN_ENV = tmt.utils.Environment.from_dict({"CONTAINER_CONNECTION": f'{PODMAN_MACHINE_NAME}-root'})
+PODMAN_MACHINE_CPU = os.getenv('TMT_BOOTC_PODMAN_MACHINE_CPU', '2')
+PODMAN_MACHINE_MEM = os.getenv('TMT_BOOTC_PODMAN_MACHINE_MEM', '2048')
+PODMAN_MACHINE_DISK_SIZE = os.getenv('TMT_BOOTC_PODMAN_MACHINE_DISK_SIZE', '50')
```
happz, line 23: `PODMAN_MACHINE_*` - at first I wanted to convert these into keys in `BootcData`, i.e. make them configurable via fmf and CLI. As they exist here and now it's just environment variable, and that's not consistent with other plugins and their configurables.

But then: it's just `hardware`, we already have ways how a HW requirement should be specified, https://tmt.readthedocs.io/en/stable/spec/hardware.html#cpu, and here we have pretty much the following:

```
hardware:
  cpu:
    processors: ...
  memory: ...
  disk:
    - size: ...
```

It's not really `hardware` key, rather `podman-machine-hwardware`, but the content, the way how HW requirement are given, should be the same. So, I have the following proposal:

* please, drop the environment variables from these: they do not follow the schema we established for plugin keys, https://tmt.readthedocs.io/en/stable/overview.html#plugin-variables, and their very existence suggests it should be possible to modify the number of `podman-machine` CPUs this way - it shouldn't we need to stick to HW specification we already have to remain consistent.
* we're using Pint to help us with units, so we could track the magnitude and unit together, therefore `2048` should be `2048 MB` or something similar: see how `virtual` does it, https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/testcloud.py#L180. The code below would then use `PODMAN_MACHINE_MEM.to('MB').magnitude` when passing the value to commands.
* together, we would end up with something like the following:
  ```python
  DEFAULT_PODMAN_MACHINE_CPU = 2
  DEFAULT_PODMAN_MACHINE_MEM: 'Size' = tmt.hardware.UNITS('2048 MB')
  DEFAULT_PODMAN_MACHINE_DISK_SIZE: 'Size' = tmt.hardware.UNITS('50 GB')
  ```

That would align your code with the rest of tmt, but it would also remove the configurability of these three values, which would be unfortunate, but I strongly believe we should allow it, but not this way. And I would submit a follow-up patch implementing `podman-machine-hardware` key, to open this to users - I don't want to block this PR on it, but I also don't want to let `TMT_BOOTC_PODMAN_MACHINE_MEM` being used, because once it's out there, it's much harder to get rid of it...

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,311 @@
...
+            chmod +x /opt/testcloud-guest.sh && \
+            echo "[Unit]" >> /etc/systemd/system/testcloud.service && \
+            echo "Description=Testcloud guest integration" >> /etc/systemd/system/testcloud.service && \
+            echo "After=cloud-init.service" >> /etc/systemd/system/testcloud.service && \
+            echo "[Service]" >> /etc/systemd/system/testcloud.service && \
+            echo "ExecStart=/bin/bash /opt/testcloud-guest.sh" >> /etc/systemd/system/testcloud.service && \
+            echo "[Install]" >> /etc/systemd/system/testcloud.service && \
+            echo "WantedBy=multi-user.target" >> /etc/systemd/system/testcloud.service && \
+            systemctl enable testcloud.service
+        '''
```
happz, line 201: I have a feeling this template (and the one with the HTTP server above) is starting to get out of hands - a complicated multiline string buried in a function. How about making it a module member, e.g. `CONTAINERFILE_TEMPLATE`, with some nice indentation of commands in it, maybe a comment or two? It's really hard for me to track what's happening there, it's very dense. Something like this would make it easier to decipher, at least for me:

```python
CONTAINER_TEMPLATE = """
FROM {{ base_image }}

RUN    dnf -y install cloud-init rsync \
    # Make sure cloud-init runs
    && ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants \
    # ...
    && rm /usr/local -rf \
    && ln -sr /var/usrlocal /usr/local \
    && mkdir -p /var/usrlocal/bin \
    && dnf clean all \
    # Workaround some testcloud problem??
    && echo "{{ testcloud_guest }}" >> /opt/testcloud-guest.sh \
    ...
```

Note I moved `&&` to the beginning of lines - I'm used to this way, but if you believe `&&` at the end is better, no problem. I'd just keep commands spread across multiple lines, with comments and better visibility into why they are there happening - as I reviewer, I can see some manipulation with `/usr/local`, but that's all... A hint or two would help future readers.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,311 @@
...
+        except:
+            self._logger.debug("Unable to remove existing podman machine (it might not exist)")
+
+        self._logger.debug("Initializing podman machine")
+        tmt.utils.Command(
+            "podman", "machine", "init", "--rootful",
+            "--disk-size", PODMAN_MACHINE_DISK_SIZE,
+            "--memory", PODMAN_MACHINE_MEM,
+            "--cpus", PODMAN_MACHINE_CPU,
+            "-v", "/var/tmp/tmt:/var/tmp/tmt",
```
happz, line 257: I'd make this path module-level member, some `DEFAULT_TMP_PATH` or something - the issue is, it's in the middle of code, it's not configurable here, it looks like a workdir root users **can** and often **do** override, so it will need to be configurable in the future, and linters complain it's an unsafe use of a temporary directory. It's not, it's perfectly fine, but moving it to module level, we could waive silence linter more easily.

ckyrouac:
force pushed updates addressing all the suggestions.

I had broken out the podman-machine cpu/mem parameters into environment variables with the idea that the podman-machine would be spun up once for multiple tests. With the current code, a new machine is created and destroyed for each test run. That's mostly because I'm not sure where/how to hook into the higher level run code. Although, the more I think about it, I guess allowing each test to specify the podman-machine cpu/mem requirements makes sense because some tests might require significantly more resources than other tests.

> please, install and enable pre-commit in your repo

hmm, I have it installed but it didn't run on one of those commits for some reason.

ckyrouac:
hmm, looking into the test failures

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,364 @@
...
+# install dependencies required by tmt
+dnf -y install cloud-init rsync
+ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants
+dnf clean all
+
+# add the scripts in /var/lib/tmt/scripts to the PATH
+touch /etc/environment && \
+echo "export PATH=$PATH:/var/lib/tmt/scripts" >> /etc/environment && \
+
+# create a simple webserver used by tmt to verify the VM is running
```
thrix, line 48: @frantisekz @ckyrouac I wonder if this is needed now that testcloud was fixed to be able to start this sever by itself:

https://github.com/teemtee/testcloud/pull/6/files

ckyrouac: nice! Removed it

ckyrouac:
not sure why the CI tests were failing. The tests are passing locally, in CI the setup was failing to create the /var/tmp/tmt/test-XXXX directory. I can't find the previous run results to get the exact error after I pushed the latest updates.

psss:
One of the old jobs is [here](https://artifacts.dev.testing-farm.io/3b72dcfa-de4e-4a1a-8323-02ebdcb7c88c/). There seems to be a problem with `tmt init` at the beginning:

```
:: [ 09:08:32 ] :: [  BEGIN   ] :: Running 'tmt init'
Path '/var/ARTIFACTS/work-bootcdy90p5cu/plans/provision/bootc/discover/default-0/tests/tests/provision/bootc' already has a parent fmf tree root '/var/ARTIFACTS/work-bootcdy90p5cu/plans/provision/bootc/discover/default-0/tests/.fmf'.
Do you really want to initialize a nested tree? [y/N]: Aborted!
:: [ 09:08:33 ] :: [   FAIL   ] :: Command 'tmt init' (Expected 0, got 1)
```

Using `tmt init --force` should help.

ckyrouac:
ah, thanks for the link. The latest commit should fix it. Not sure if you need to trigger another test run or not.

tmt.spec
```diff
@@ -90,6 +90,20 @@ Recommends:     qemu-system-x86-core
...
+Obsoletes:      tmt-provision-bootc < %{version}-%{release}
+%endif
+Requires:       tmt == %{version}-%{release}
+Requires:       tmt+provision-virtual == %{version}-%{release}
+Requires:       podman
+# virtiofsd is for podman-machine
+Recommends:     virtiofsd
+
+%description -n tmt+provision-bootc %_metapackage_description
+
```
psss, line 106: At least a minimal `%files` section is also needed to convince `rpm-build` to actually create the subpackage. See bottom of the spec.

tmt.spec
```diff
@@ -90,6 +90,20 @@ Recommends:     qemu-system-x86-core
 
 %description -n tmt+provision-virtual %_metapackage_description
 
+%package -n     tmt+provision-bootc
```
happz, line 93: RPM build fails with the following:

```
Processing files: tmt+provision-bootc-1.39.dev888-1.20241106153118869510.pr3161.23.g7d63c518.fc39.noarch

Error: The package name contains an extras name `provision-bootc` that was not found in the metadata.
Check if the extras were removed from the project. If so, consider removing the subpackage and obsoleting it from another.

error: Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_EXTRAS_NOT_FOUND_ERROR___SEE_STDERR ***

Error: The package name contains an extras name `provision-bootc` that was not found in the metadata.
Check if the extras were removed from the project. If so, consider removing the subpackage and obsoleting it from another.

error: Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_EXTRAS_NOT_FOUND_ERROR___SEE_STDERR ***
```

I'm not an expert on rpmbuild, but IIUIC, the problem is no `provision-bootc` extras in `pyproject.toml`. Even an empty one, like `provision-container`, could resolve the problem: https://github.com/teemtee/tmt/blob/main/pyproject.toml#L62

ckyrouac: thanks, pushed the change to pyproject.tml

ckyrouac:
The tests are failing with `Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory`. Any ideas why?

omertuc:
> The tests are failing with `Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory`. Any ideas why?

I ran into this before locally, I solved it with `systemctl enable virtqemud.socket`

happz:
> The tests are failing with `Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory`. Any ideas why?

Tests depend on additional setup, shouldn't we add a dedicated plan & let it run in the `provision` check? IIUIC, that's what we did with `provision-virtual` tests, they are now grouped under `/plans/provision/virtual` which tasks care of the necessary setup, starting libvirt is exactly one of the steps in `plans/provision/virtual.fmf`, and since bootc provisioning plugin depends on an extra setup as well, we should isolate them in the same manner.

@psss does it make sense? It does look like another provision-foo plan is needed, and I think we can add it to the same Github check as `virtual` tests.

ckyrouac:
@happz added a prepare step to the bootc plan. Could you trigger another test run?

psss:
> @psss does it make sense? It does look like another provision-foo plan is needed, and I think we can add it to the same Github check as `virtual` tests.

Yes, this definitely makes sense. Let's have this handled in a consistent way.

ckyrouac:
If I understand this right, I believe this was already in place. The [prepare step](https://github.com/teemtee/tmt/pull/3161/files#diff-93304e094886795c4b6e58cf3826294d1ba36e115721a851bd284cfdc31510bfR10) just wasn’t in place yet but is now.

Now the bootc tests pass but there is an [avc failure](https://artifacts.dev.testing-farm.io/f59d5356-f9a9-400a-b3c7-a9ef75602018/work-bootchri1stdf/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/checks/avc.txt) that I don’t understand. Any ideas?

happz:
> If I understand this right, I believe this was already in place. The [prepare step](https://github.com/teemtee/tmt/pull/3161/files#diff-93304e094886795c4b6e58cf3826294d1ba36e115721a851bd284cfdc31510bfR10) just wasn’t in place yet but is now.
> 
> Now the bootc tests pass but there is an [avc failure](https://artifacts.dev.testing-farm.io/f59d5356-f9a9-400a-b3c7-a9ef75602018/work-bootchri1stdf/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/checks/avc.txt) that I don’t understand. Any ideas?

A rebase should help, the problem has been addressed in https://github.com/teemtee/tmt/pull/3355

martinhoyer:
> If I understand this right, I believe this was already in place. The [prepare step](https://github.com/teemtee/tmt/pull/3161/files#diff-93304e094886795c4b6e58cf3826294d1ba36e115721a851bd284cfdc31510bfR10) just wasn’t in place yet but is now.
> 
> Now the bootc tests pass but there is an [avc failure](https://artifacts.dev.testing-farm.io/f59d5356-f9a9-400a-b3c7-a9ef75602018/work-bootchri1stdf/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/checks/avc.txt) that I don’t understand. Any ideas?

I'm guessing it's the usage of /var, nosuid transition:
https://danwalsh.livejournal.com/78312.html

I can see `container-selinux` package being installed in the log btw. Maybe a bug there on in podman-machine package? Not sure how to solve this. @psss @happz

happz:
> > If I understand this right, I believe this was already in place. The [prepare step](https://github.com/teemtee/tmt/pull/3161/files#diff-93304e094886795c4b6e58cf3826294d1ba36e115721a851bd284cfdc31510bfR10) just wasn’t in place yet but is now.
> > Now the bootc tests pass but there is an [avc failure](https://artifacts.dev.testing-farm.io/f59d5356-f9a9-400a-b3c7-a9ef75602018/work-bootchri1stdf/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/checks/avc.txt) that I don’t understand. Any ideas?
> 
> I'm guessing it's the usage of /var, nosuid transition: https://danwalsh.livejournal.com/78312.html
> 
> I can see `container-selinux` package being installed in the log btw. Maybe a bug there on in podman-machine package? Not sure how to solve this. @psss @happz

It has been reported for bootc image builder, see https://github.com/osbuild/bootc-image-builder/issues/645. It does look like something bootc either should not do, or ask selinux-policy maintainers to allow.

martinhoyer:
> Hmm that probably relates to our initialization of containers/storage at bootc install time. But I am confused as to why we may only be seeing this here. It may relate somehow to the host policy? We do have some coverage for this in other CI tests… Long story short if this theory is right then we need to suppress the domain transition or maybe just switch to skopeo which we probably want to do anyways.

Thanks,
@cgwalters @ckyrouac Would you prefer merging it "as is", perhaps like a tech preview or addressing the selinux issue first, most likely sliding to the next release cycle?

ckyrouac:
I will be afk on paternity leave soon (most likely in a week), so I won't be able to work on this much for a few months. So, unless someone else can pick up the work, I think merging as-is makes the most sense.

martinhoyer:
> I will be afk on paternity leave soon (most likely in a week)  

Congratulations in advance sir!

psss:
> So, unless someone else can pick up the work, I think merging as-is makes the most sense.

Ok, then I suggest to mark the `avc` failure as expected for now and include this in `1.39` as a tech preview. I've added a short release not in 44b0cd38. Does it look ok?

plans/provision/bootc.fmf
```diff
@@ -0,0 +1,39 @@
+summary: Bootc virtual machine via testcloud
```
psss, line 1: This plan is currently run under the `full` packit job. Seems to be working fine, but this means that it is executed for all targets. Should we move it under the `provision` packit job? @happz, I guess that's what you meant in your [comment](https://github.com/teemtee/tmt/pull/3161#issuecomment-2468435410)? Or are we good as it is now?

tests/provision/bootc/test.sh
```diff
@@ -0,0 +1,64 @@
...
+    rlPhaseStartTest "Image that needs dependencies"
+        rlRun "podman build . -f $CONTAINERFILE_NEEDS_DEPS -t $IMAGE_NEEDS_DEPS"
+        rlRun "cp $IMAGE_NEEDS_DEPS_PLAN ."
+        rlRun "tmt -vvvvv run -i $run"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Image that already includes dependencies"
+        rlRun "podman build . -f $CONTAINERFILE_INCLUDES_DEPS -t $IMAGE_INCLUDES_DEPS"
+        rlRun "cp $IMAGE_INCLUDES_DEPS_PLAN ."
+        rlRun "tmt -vvvvv run -i $run"
```
psss, line 39: If reusing the same run `--id` it's better to use `tmt run --scratch` to remove previous run content before executing a new run. Otherwise this results into following tmt runs to just report `done`.

tests/provision/bootc/test.sh
```diff
@@ -0,0 +1,64 @@
...
+CONTAINERFILE_NEEDS_DEPS="$(pwd)/data/needs_deps.containerfile"
+CONTAINERFILE_NEEDS_DEPS_PLAN="$(pwd)/data/containerfile_needs_deps.fmf"
+CONTAINERFILE_INCLUDES_DEPS="$(pwd)/data/includes_deps.containerfile"
+CONTAINERFILE_INCLUDES_DEPS_PLAN="$(pwd)/data/containerfile_includes_deps.fmf"
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        # cleanup previous runs
+        test -d /var/tmp/tmt/testcloud && rlRun "rm -rf /var/tmp/tmt/testcloud"
```
psss, line 18: If we need to do a complete cleanup, what about using `TMT_WORKDIR_ROOT` instead to choose a different workdir root? We're trying to write tests so that they can be easily/safely executed on user laptop as well. This would remove for example all fetched images.

tmt.spec
```diff
@@ -90,6 +90,20 @@ Recommends:     qemu-system-x86-core
...
+Summary:        Dependencies required for tmt bootc machine provisioner
+Provides:       tmt-provision-bootc == %{version}-%{release}
+%if 0%{?fedora} < 40
+Obsoletes:      tmt-provision-bootc < %{version}-%{release}
+%endif
```
psss, line 98: There was no `tmt-provision-bootc` package before, so I think we can drop the `Obsoletes` here.

tmt/steps/provision/bootc.py
```diff
@@ -0,0 +1,348 @@
...
+            logger=self._logger,
+            env=PODMAN_ENV if self._rootless else None)
+
+        try:
+            tmt.utils.Command(
+                "podman", "machine", "rm", "-f", PODMAN_MACHINE_NAME
+                ).run(cwd=self.workdir, stream_output=True, logger=self._logger)
+        except BaseException:
+            self._logger.debug(
+                "Unable to remove podman machine {PODMAN_MACHINE_NAME}, it might not exist")
```
psss, line 84:
```suggestion
                "Unable to remove podman machine '{PODMAN_MACHINE_NAME}', it might not exist.")
```

psss (changes requested):
Added a couple of comments. I tried the minimal example but ended up with the following error:

```
Command 'podman machine start podman-machine-tmt' returned 125.

# stdout (1/1 lines)
Starting machine "podman-machine-tmt"

# stderr (2/2 lines)
time="2024-11-15T14:33:33+01:00" level=error msg="process 419571 has not ended"
Error: failed to find virtiofsd: exec: "virtiofsd": executable file not found in $PATH
```

I tried this under a regular user. Do I understand it correctly that it is necessary to always run this under root?

ckyrouac:
@psss I addressed most of your suggestions in recent pushes. I probably won’t have time to refactor the tests per your suggestions though.

@happz I think I removed your commit via a force push, sorry. I’m done pushing now :-)

plans/provision/bootc.fmf
psss: Covered by f3772f6533398dcd93b28f6c3f88bed83597e9a4.

psss:
I've added a bunch of changes in f781e34b, mainly to fix & simplify tests:

* Make sure that all test cases are actually executed.
* Run tests directly in the `data` directory.
* Simplify plans using inheritance.
* Plus some minor adjustments.

Last remaining problem is the avc check. I guess we should just disable it if we cannot `xfail` it?

tests/provision/bootc/test.sh
psss: Should be covered by f781e34b.

martinhoyer (approved):

psss (approved):

Issue #3162, opened 2024-08-22
tmt will try to clean all guests when users run tmt clean guests with --id specified


skycastlelily:
(dev) [lnie@ceph-04 tmt]$ tmt clean guests --id run-060
clean
guests
        
            guest: stopped
            guest: removed
            summary: 0 tasks completed
        finish
        
            guest: stopped
            guest: removed
            summary: 0 tasks completed
        finish
        
            guest: stopped
            guest: removed
            summary: 0 tasks completed

PR #3163, opened 2024-08-22
Changing Clean.guests() to accept run IDs

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -3899,7 +3899,7 @@ class Clean(tmt.utils.Common):
                         self.cli_invocation.options['quiet'] = quiet
         return successful
 
-    def guests(self, id_: tuple[str, ...]) -> bool:
+    def guests(self, run_ids: tuple[str, ...]) -> bool:
         """ Clean guests of runs """
         self.info('guests', color='blue')
         root_path = Path(self.opt('workdir-root'))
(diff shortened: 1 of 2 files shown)


skycastlelily:
Fix this issue:https://github.com/teemtee/tmt/issues/3162:)

tmt/base.py
```diff
@@ -3899,7 +3899,7 @@ def _stop_running_guests(self, run: Run) -> bool:
                         self.cli_invocation.options['quiet'] = quiet
         return successful
 
-    def guests(self) -> bool:
+    def guests(self, id_: tuple[str, ...]) -> bool:
```
happz, line 3902: How about giving the parameter a different name? After the patches that landed in this area I for one feel we have way too many "id" mentions around, and it's easy to forget what they mean. `id_` is very generic, how about `run_ids`? It would be immediately clear, and it would not collide with the built-in `id()` function.

happz:
@skycastlelily the commit message and PR subject would deserve a bit more work. "Pass id_ to Clean.guests command" describes "what" is happening in the patch, but the patch is pretty simple and a reviewer (or any other reader of the code) can see that quite soon. But "why" is not addressed, why is it that `id_` needs to be passed to `guests()`?

The issue should be also mentioned in the commit message, so we get both linked together in git, not just in a github comment. And the issue actually provides the reason for "why", it would be build upon it when writing the commit message. For example, something like the following might help us understand why the commit happens to exist:

```
Changing Clean.guests() to accept run IDs

That way, tmt can honor both run IDs given to `tmt clean` and/or `tmt clean guests`.

Fixes #3162
```

skycastlelily:
sure, much better:)

skycastlelily:
Thanks for your mentor, learned:)

therazix (approved):

martinhoyer (approved):

PR #3166, opened 2024-08-26
Use Jinja2 and polarion JUnit flavor to generate XUnit in polarion report plugin

seberm:
Do not edit the default JUnit flavor generated by junit report plugin using an `ElementTree`, use the predefined Jinja2 Polarion template. This PR also adds the XSD schema for a Polarion XUnit file. The checking against this schema is done dynamically by lxml in the same way as for the `default` junit flavor.

Related to:

Depends on:
- https://github.com/teemtee/tmt/pull/3150

## TODOs:
- [x] `testuites` properties should be probably part of `ResultsContext` (as property)
- [x] Test the actual polarion import (done by @KwisatzHaderach )

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/tests/report/polarion/test.sh
+++ b/tests/report/polarion/test.sh
@@ -7,14 +7,143 @@ rlJournalStart
         rlRun "set -o pipefail"
     rlPhaseEnd
 
-    rlPhaseStartTest
-        rlRun "tmt run -avr execute report -h polarion --project-id RHELBASEOS --no-upload --planned-in RHEL-9.1.0 --file xunit.xml 2>&1 >/dev/null | tee output" 2
+    rlPhaseStartTest 'Test the properties gets propagated to testsuites correctly'
+        rlRun "tmt run -avr execute report -h polarion --no-upload --project-id RHELBASEOS --template mytemplate --planned-in RHEL-9.1.0 --arch x86_64 --description mydesc --assignee myassignee --pool-team mypoolteam --platform myplatform --build mybuild --sample-image mysampleimage --logs mylogslocation --compose-id mycomposeid --file xunit.xml 2>&1 >/dev/null | tee output" 2
         rlAssertGrep "1 test passed, 1 test failed and 1 error" "output"
-        rlAssertGrep '<testsuite name="/plan" disabled="0" errors="1" failures="1" skipped="0" tests="3"' "xunit.xml"
-        rlAssertGrep '<property name="polarion-project-id" value="RHELBASEOS" />' "xunit.xml"
-        rlAssertGrep '<property name="polarion-testcase-id" value="BASEOS-10914" />' "xunit.xml"
-        rlAssertGrep '<property name="polarion-custom-plannedin" value="RHEL-9.1.0" />' "xunit.xml"
         rlAssertGrep "Maximum test time '2s' exceeded." "xunit.xml"
+
+        # testsuites and testsuite tag attributes
+        rlAssertGrep '<testsuites disabled="0" errors="1" failures="1" tests="3"' "xunit.xml"
+        rlAssertGrep '<testsuite name="/plan" disabled="0" errors="1" failures="1" skipped="0" tests="3"' "xunit.xml"
+        # Main testsuite properties
+        rlAssertGrep '<property name="polarion-project-id" value="RHELBASEOS"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-project-span-ids" value="RHELBASEOS,RHELBASEOS"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-testrun-title" value="plan_[0-9]\+"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-testrun-template-id" value="mytemplate"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-user-id" value="' "xunit.xml"
+
+        # Custom testsuite properties
+        rlAssertGrep '<property name="polarion-custom-description" value="mydesc"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-plannedin" value="RHEL-9.1.0"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-assignee" value="myassignee"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-poolteam" value="mypoolteam"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-arch" value="x86_64"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-platform" value="myplatform"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-build" value="mybuild"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-sampleimage" value="mysampleimage"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-logs" value="mylogslocation"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-composeid" value="mycomposeid"/>' "xunit.xml"
+
+        # The testcase properties
+        rlAssertGrep '<property name="polarion-testcase-id" value="BASEOS-10913"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-testcase-id" value="BASEOS-10914"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-testcase-id" value="BASEOS-10915"/>' "xunit.xml"
+        rlAssertGrep '<property name="polarion-testcase-project-id" value="RHELBASEOS"/>' "xunit.xml"
+    rlPhaseEnd
+
+    rlPhaseStartTest 'Test the facts properties'
+        rlRun "tmt run -avr execute report -h polarion --no-upload --project-id RHELBASEOS --use-facts --file xunit.xml 2>&1 >/dev/null | tee output" 2
+        rlAssertGrep '<property name="polarion-custom-hostname" value="' "xunit.xml"
+        rlAssertGrep '<property name="polarion-custom-arch" value="' "xunit.xml"
+    rlPhaseEnd
+
+    rlPhaseStartTest 'The "None" string should never be in a property value'
+        rlRun "tmt run -avr execute report -h polarion --no-upload --project-id RHELBASEOS --template '' --planned-in '' --arch '' --description '' --assignee '' --pool-team '' --platform '' --build '' --sample-image '' --logs '' --compose-id '' --file xunit.xml 2>&1 >/dev/null | tee output" 2
+        rlAssertNotGrep 'value="None"' "xunit.xml"
+
+        rlRun "export \
+        TMT_PLUGIN_REPORT_POLARION_PROJECT_ID= \
+        TMT_PLUGIN_REPORT_POLARION_TITLE= \
+        TMT_PLUGIN_REPORT_POLARION_DESCRIPTION= \
+        TMT_PLUGIN_REPORT_POLARION_TEMPLATE= \
+        TMT_PLUGIN_REPORT_POLARION_PLANNED_IN= \
+        TMT_PLUGIN_REPORT_POLARION_ASSIGNEE= \
+        TMT_PLUGIN_REPORT_POLARION_POOL_TEAM= \
+        TMT_PLUGIN_REPORT_POLARION_ARCH= \
+        TMT_PLUGIN_REPORT_POLARION_PLATFORM= \
+        TMT_PLUGIN_REPORT_POLARION_BUILD= \
+        TMT_PLUGIN_REPORT_POLARION_SAMPLE_IMAGE= \
+        TMT_PLUGIN_REPORT_POLARION_LOGS= \
+        TMT_PLUGIN_REPORT_POLARION_COMPOSE_ID= \
+        "
+        rlRun "tmt run -avr execute report -h polarion --no-upload --project-id RHELBASEOS --file xunit.xml 2>&1 >/dev/null | tee output" 2
+        rlAssertNotGrep 'value="None"' "xunit.xml"
+    rlPhaseEnd
+
+    rlPhaseStartTest 'Check the plugin behavior based on setting ENV variables'
+        rlRun "export \
+        TMT_PLUGIN_REPORT_POLARION_PROJECT_ID=myprojectid \
+        TMT_PLUGIN_REPORT_POLARION_TITLE=mytitle \
+        TMT_PLUGIN_REPORT_POLARION_DESCRIPTION=mydesc \
+        TMT_PLUGIN_REPORT_POLARION_TEMPLATE=mytemplate \
+        TMT_PLUGIN_REPORT_POLARION_PLANNED_IN=myplannedin \
+        TMT_PLUGIN_REPORT_POLARION_ASSIGNEE=myassignee \
+        TMT_PLUGIN_REPORT_POLARION_POOL_TEAM=mypoolteam \
+        TMT_PLUGIN_REPORT_POLARION_ARCH=x86_64 \
+        TMT_PLUGIN_REPORT_POLARION_PLATFORM=myplatform \
+        TMT_PLUGIN_REPORT_POLARION_BUILD=mybuild \
+        TMT_PLUGIN_REPORT_POLARION_SAMPLE_IMAGE=mysampleimage \
+        TMT_PLUGIN_REPORT_POLARION_LOGS=mylogslocation \
+        TMT_PLUGIN_REPORT_POLARION_COMPOSE_ID=mycomposeid \
+        "
+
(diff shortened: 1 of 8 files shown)


tmt/steps/report/junit.py
```diff
@@ -266,7 +267,7 @@ class ReportJUnitData(tmt.steps.report.ReportStepData):
         default=DEFAULT_FLAVOR_NAME,
         option='--flavor',
         metavar='FLAVOR',
-        choices=[DEFAULT_FLAVOR_NAME, CUSTOM_FLAVOR_NAME],
+        choices=[DEFAULT_FLAVOR_NAME, CUSTOM_FLAVOR_NAME, 'polarion'],
```
happz, line 270: Hmm, can `junit` plugin really render the correct and meaningful `polarion` flavor? IIUIC, it cannot, because there is plenty of code in the `polarion` plugin precrunching information for the template. So, `--how junit --flavor polarion` is probably not going to work as one might expect.

In this sense, `polarion` is probably not a flavor owned by the `junit` plugin, and it can and should be isolated from the `junit` plugin as any other 3rd party junit-rendering plugin there might be. Both are shipped with tmt, sure, but on the implementation level, `polarion` brings this "flavor", by providing a template and calling `junit` functionality, all that while `junit` is not aware of `polarion` existence. At least not by its name :)

There is also `polarion` being mentioned by name in the template below, would it be possible to make that point generic enough for `polarion` template to take over control and render `properties`? What I'm worried about: let's say a few more flavors appear out there, in the wild world, `pytest` and `jest`, and these would also like to include `<properties/>` elements. Will we add their names to the `{% if (FLAVOR == 'polarion') and result.properties is defined %}` condition as well? I suspect we wouldn't, which makes me think this "leaking" of `polarion` into `_base.xml.j2` is not right - it's necessary to let `polarion` emit this element, but can we do it in way in which `_base.xml.j2` is not guarding it with a name-based check?

seberm: As discussed yesterday, I have moved the properties decorator into the `junit.py` module. I also removed the `flavor` condition for polarion from the `_base.xml.j2` template and the `polarion` from the supported arguments of the `--flavor` argument. The `polarion` flavor is now used only internally by polarion report plugin.

tmt/steps/report/polarion.py
```diff
@@ -224,53 +232,46 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
             'description', 'planned_in', 'assignee', 'pool_team', 'arch', 'platform', 'build',
             'sample_image', 'logs', 'compose_id', 'fips']
 
-        xml_data = make_junit_xml(
-            phase=self,
+        # TODO: The current behavior is that the None value is propagated into the final XML as a
```
seberm, line 235: @KwisatzHaderach
What is the expected behavior here? Should the `None` value always be propagated into the property value as a string `"None"? My expectation would be the `None` should propagate as an empty string.

KwisatzHaderach: project_id is required, I'm quite surprised I've forgot to add a check for it and raise in case it's missing

so my suggestion would be to add a last resort project_id taken from the login credentials `PolarionWorkItem._session.default_project` and then raising if still empty/None (AFAIK the argument is required for pylero so it shouldn't happen), only bad detail is this can lead to unexpected reports in whatever the user has set as their default project, but at least it would be whatever they use in authentication which should be fine

seberm: Thanks for the reply. I've added value from `PolarionWorkItem._session.default_project` as a last resort and I've tried to document it. If the `bool(project_id)` still evaluates to `False`, I am now raising a `ReportError` exception.

I also added the test which checks there is no `None` string in the property value:

```
rlAssertNotGrep 'value="None"' "xunit.xml"
```

tmt/steps/report/polarion.py
```diff
@@ -204,6 +213,8 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         from tmt.export.polarion import find_polarion_case_ids, import_polarion
         import_polarion()
         from tmt.export.polarion import PolarionWorkItem
+
+        # TODO: Is this assert really necessary? Isn't this always true?
```
seberm, line 217: @KwisatzHaderach is this assertion really necessary? Shouldn't we e.g. raise some ReportError exception if the `bool(PolarionWorkItem)` evaluates to `False`?

tmt/steps/report/polarion.py
```diff
@@ -213,93 +224,119 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
                     "please run 'tmt tests export --how polarion' on it.")
                 continue
+
             work_item_id, test_project_id = find_polarion_case_ids(result.ids)
 
             if test_project_id is None:
                 self.warn(f"Test case '{result.name}' missing or not found in Polarion.")
                 continue
 
+            # TODO: try to remove the asserts by raising a proper ReportError exception?
```
seberm, line 301: @KwisatzHaderach 
Same here - What about checking `work_item_id` and `project_span_id` for `None` value and raising a proper exception instead of failing completely without a proper error message for the user?

tmt/steps/report/polarion.py
KwisatzHaderach: yeah, that sounds like a good idea

tmt/steps/report/polarion.py
KwisatzHaderach: probably not, I've added a lot of these to make typing happy, so if no linting tool complains then feel free to remove

tests/report/polarion/test.sh
```diff
@@ -7,14 +7,97 @@ rlJournalStart
...
+        TMT_PLUGIN_REPORT_POLARION_POOL_TEAM=mypoolteam \
+        TMT_PLUGIN_REPORT_POLARION_ARCH=x86_64 \
+        TMT_PLUGIN_REPORT_POLARION_PLATFORM=myplatform \
+        TMT_PLUGIN_REPORT_POLARION_BUILD=mybuild \
+        TMT_PLUGIN_REPORT_POLARION_SAMPLE_IMAGE=mysampleimage \
+        TMT_PLUGIN_REPORT_POLARION_LOGS=mylogslocation \
+        TMT_PLUGIN_REPORT_POLARION_COMPOSE_ID=mycomposeid \
+        "
+
+        # TODO: how to check these?
```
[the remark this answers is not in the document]
seberm: I just wanted to test the "false" value of `TMT_PLUGIN_REPORT_POLARION_UPLOAD` (which probably means setting `TMT_PLUGIN_REPORT_POLARION_UPLOAD=''`, but I think we can skip this check.

For the `TMT_PLUGIN_REPORT_POLARION_USE_FACTS` I just wanted to check it is effective, which means the difference between setting `TMT_PLUGIN_REPORT_POLARION_USE_FACTS="yes"` and `TMT_PLUGIN_REPORT_POLARION_USE_FACTS=''`.

tmt/steps/report/polarion.py
```diff
@@ -213,93 +223,120 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
             if test_project_id is None:
                 self.warn(f"Test case '{result.name}' missing or not found in Polarion.")
                 continue
 
-            assert work_item_id is not None
-            assert project_span_ids is not None
+            if work_item_id is None:
+                raise tmt.utils.ReportError(
+                    f"Work item ID missing or not found in Polarion (Test case: '{result.name}', "
+                    "Test project ID: '{test_project_id}).")
```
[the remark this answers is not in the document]
seberm: Is really `work_item_id` unreachable? This would mean if `work_item_id==None` then always `test_project_id==None`.

Is this always true? `if test_project_id is None -> work_item_id is None`

In that case, we could just change the condition to `if None in [work_item_id, test_project_id]: ...` or just add `or work_item_id is None` as you recommended.

tests/report/polarion/test.sh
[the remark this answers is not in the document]
KwisatzHaderach: well, you can set it to true/false with almost anything resembling a true/false value https://click.palletsprojects.com/en/8.1.x/parameters/#parameter-types but the thing is since False is the default you have no way of knowing if it had any effect

the facts are easy, just check existence/nonexistence of the hostname property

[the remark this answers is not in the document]
happz: For flags, please use either `1` or `0`, see the last paragraph in https://tmt.readthedocs.io/en/stable/overview.html#plugin-variables:

> For setting flag-like option, 0 and 1 are the expected value. For example, an interactive mode would be enabled in this run:
> 
> ```
> TMT_PLUGIN_EXECUTE_TMT_INTERACTIVE=1 tmt run ... execute -h tmt ...
> ```

tmt/steps/report/polarion.py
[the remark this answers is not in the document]
KwisatzHaderach: yeah, either it found the item and filled both values or it didn't and neither of the values are set...assuming everything went fine in creating the instance in pylero

yeah, whichever condition is fine with me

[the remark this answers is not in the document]
seberm: Ok, changed. My construction of condition does not pass the typing checks. I think you can mark this resolved.

tests/report/polarion/test.sh
```diff
@@ -7,14 +7,125 @@ rlJournalStart
...
+        TMT_PLUGIN_REPORT_POLARION_BUILD \
+        TMT_PLUGIN_REPORT_POLARION_SAMPLE_IMAGE \
+        TMT_PLUGIN_REPORT_POLARION_LOGS \
+        TMT_PLUGIN_REPORT_POLARION_COMPOSE_ID \
+        TMT_PLUGIN_REPORT_POLARION_USE_FACTS \
+        "
+
+        # The facts must not be set
+        rlRun "export TMT_PLUGIN_REPORT_POLARION_USE_FACTS=0"
+        rlRun "tmt run -avr execute report -h polarion --no-upload --file xunit.xml 2>&1 >/dev/null | tee output" 2
```
KwisatzHaderach, line 120: I would expect this to fail since you have no project id set

KwisatzHaderach: was thinking that none of these can actually run and yeah, these are not run in the pipelines so they can be run locally only...and the project_id would be taken from your pylero config...it's fun returning to own code couple months/years later 😆

seberm: Yeah, I agree. They are not failing on my machine due to `PolarionWorkItem._session.default_project`. The tests will probably fail in the CI - I will fix them by providing `--project-id` and I will probably add one more test which will check the tmt fails if no `project_id` is found.

seberm: But firstly I need to resolve the lxml installation problems in the CI :/ (https://github.com/teemtee/tmt/pull/3150#issuecomment-2315937339).

KwisatzHaderach: thanks, IMO it's better to always have explicit project ID

but as mentioned in the comment CI is not running these, tests/report/polarion/main.fmf clearly states `enabled: false` since we don't have any credentials in there and also no access to a Polarion instance

may be moved into teemtee/tests and test with requre, but am not sure if it is doable, definitely not in scope of this patch

seberm: > `tests/report/polarion/main.fmf` clearly states `enabled: false`

Oh, I missed that :man_facepalming: . Ok never mind. Let's just for now keep the tests here for local testing? I will add an explicit `--project-id` to these tests.

tmt/steps/report/polarion.py
```diff
@@ -213,84 +227,115 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
+            results_context=results_context,
+            )
+
         f_path = self.data.file or self.workdir / DEFAULT_NAME
-        with open(f_path, 'wb') as fw:
-            ElementTree.ElementTree(xml_tree).write(fw)
+
+        try:
+            with open(f_path, 'w') as fw:
+                fw.write(xml_data)
```
martinhoyer, line 336:
```suggestion
            f_path.write_text(xml_data)
```
Just a nitpick, but if there is only one write operation, contextmanager and variable assignment looks redundant too me.  
(same in junit.py)

martinhoyer (approved):
Neat :rocket:

tmt/steps/report/polarion.py
seberm: Modified for both plugins. For junit I've added it as a part of https://github.com/teemtee/tmt/pull/3150.

Thanks.

KwisatzHaderach (approved):
looking good, export to polarion is working fine for me with an easy/small plan

tmt/steps/report/junit.py
```diff
@@ -83,24 +82,36 @@ class PropertyDict(TypedDict):
...
         self._properties: dict[str, str] = {}
 
     @property
     def properties(self) -> list[PropertyDict]:
         return [{'name': k, 'value': v} for k, v in self._properties.items()]
 
     @properties.setter
-    def properties(self, keyval: dict[str, str]) -> None:
-        self._properties = keyval
+    def properties(self, keyval: dict[str, Optional[str]]) -> None:
```
seberm, line 93: @happz If you remember we discussed incompatible mypy types if I use `@properties.setter` instead of defining a separate setter method (e.g. `set_properties()`).

The mypy doesn't like that the property setter accepts the `dict` and a property getter returns a `list` of properties.

The error is:
```
tmt/steps/report/polarion.py:313: error: Incompatible types in assignment (expression has type "dict[str, str]", variable has type "list[PropertyDict]")  [assignment]
tmt/steps/report/polarion.py:323: error: Incompatible types in assignment (expression has type "dict[str, Any]", variable has type "list[PropertyDict]")  [assignment]
```

Do you have any hints/recommendations on how to solve it?

happz: Ok, so there's an issue for mypy discussing this (started with "it's a perversion we don't want to support" :), so I ended up with this:

```diff
diff --git a/tmt/steps/report/polarion.py b/tmt/steps/report/polarion.py
index 1f7c4d53..5f92dbd3 100644
--- a/tmt/steps/report/polarion.py
+++ b/tmt/steps/report/polarion.py
@@ -259,7 +259,7 @@ class ReportPolarion(tmt.steps.report.ReportPlugin[ReportPolarionData]):
             'pool_team',
             'sample_image']
 
-        testsuites_properties = {}
+        testsuites_properties: dict[str, Optional[str]] = {}
 
         for tr_field in other_testrun_fields:
             param = self.get(tr_field, os.getenv(f'TMT_PLUGIN_REPORT_POLARION_{tr_field.upper()}'))
@@ -310,7 +310,10 @@ class ReportPolarion(tmt.steps.report.ReportPlugin[ReportPolarionData]):
                 'polarion-testcase-project-id': test_project_id,
                 }
 
-            result.properties = testcase_properties
+            # ignore[assignment]: mypy does not support different types for property getter
+            # and setter. The assignment is correct, but mypy cannot tell.
+            # See https://github.com/python/mypy/issues/3004 for getter/setter discussions
+            result.properties = testcase_properties  # type: ignore[assignment]
 
         assert self.workdir is not None
 
@@ -320,7 +323,10 @@ class ReportPolarion(tmt.steps.report.ReportPlugin[ReportPolarionData]):
             'polarion-testrun-title': title,
             'polarion-project-span-ids': ','.join([project_id, *project_span_ids])})
 
-        results_context.properties = testsuites_properties
+        # ignore[assignment]: mypy does not support different types for property getter
+        # and setter. The assignment is correct, but mypy cannot tell.
+        # See https://github.com/python/mypy/issues/3004 for getter/setter discussions
+        results_context.properties = testsuites_properties  # type: ignore[assignment]
 
         xml_data = make_junit_xml(
             phase=self,
```

seberm: :D, ok, thanks. I did the modifications, marking this as resolved.

tmt/steps/report/polarion.py
```diff
@@ -213,93 +227,121 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
-                self.step.plan.provision.guests()[0].primary_address
-            properties['polarion-custom-arch'] = self.step.plan.provision.guests()[0].facts.arch
+            guests = self.step.plan.provision.guests()
+            try:
+                testsuites_properties['polarion-custom-hostname'] = guests[0].primary_address
+                testsuites_properties['polarion-custom-arch'] = guests[0].facts.arch
+            except IndexError as error:
+                raise tmt.utils.ReportError(
+                    'Failed to get the Polarion facts from '
+                    '`step.plan.provision.guests()`') from error
```
happz, line 279: To me as a user, `step.plan.provision.guests()` is nothing but a dark magic I don't understand. Did I do something wrong? Is Polarion broken? I bet we could provide some helpful tips for user reading the error messages :)

seberm: Sure. Is it better now?

happz (approved):

PR #3178, opened 2024-08-29
Add new examples to overwrite the inheritance problem

mcasquer:
The examples code block from spec/plans/main.fmf
is being wrongly inherited by some of the specs. Adding
new examples blocks to avoid this problem.

Pull Request Checklist

### DIFF
--- a/spec/plans/report.fmf
+++ b/spec/plans/report.fmf
@@ -4,6 +4,7 @@ story:
     the testing is finished.
 description:
     Report test results according to user preferences.
+example:
 
 /display:
     summary: Show results in the terminal window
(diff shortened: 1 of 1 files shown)


mcasquer:
The same idea needs to be implemented for the `discover` specification where the same happens and perhaps in the `execute` as well. The point is that using the same approach **will erase the full spec** from the documentation. So my guess is that further modifications (still unknown to me) are needed for both cases.

spec/plans/report.fmf
```diff
@@ -202,33 +203,7 @@ description:
...
-                                - LOG_PATH
-                /plan/two:
-                    result: PLAN_RESULT
-                        /test/one:
-                            result: TEST_RESULT
-                            log:
-                              - LOG_PATH
+        Save the report into a ``report.yaml`` file.
 
         Where ``OVERALL_RESULT`` is the overall result of all plan
```
psss, line 208:
```suggestion
        The ``OVERALL_RESULT`` is the overall result of all plan
```

Moving the example down sounds ok, but `Where` is referencing to something which has been already mentioned above. What about using `The` instead? Same for the other values below.

spec/plans/report.fmf
```diff
@@ -262,3 +237,29 @@ description:
         to the execute step plan run directory. The ``log`` key
         will be a list of such paths, even if there is just a single
         log.
+
+    example: |
+        .. code-block:: yaml
```
psss, line 242: Should be removed, the story `example` key is automatically rendered as `yaml`. Now the directive is shown in the generated docs.

psss (changes requested):
Looks ok, just two comments.

spec/plans/report.fmf
mcasquer: @psss done!

spec/plans/report.fmf
mcasquer: @psss done !

spec/plans/report.fmf
psss: Thanks!

spec/plans/report.fmf
psss: Good!

psss (approved):
Thanks for fixing this! Looks good, I've only adjusted the `example` indentation.

psss:
Spec change only, merging.

PR #3184, opened 2024-09-03
Use `TMT_REPORT_ARTIFACTS_URL` envvar again

kkaarreell:
The former change accidentally switched the code to use TMT_PLUGIN_REPORT_REPORTPORTAL_ARTIFACTS_URL
environment variable. That broke the functionality until a workaround had been implemented in TF.
This change restores the original usage of TMT_REPORT_ARTIFACTS_URL due to the fact that this variable is not reportportal plugin specific.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -163,7 +163,8 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
     artifacts_url: Optional[str] = field(
         metavar="ARTIFACTS_URL",
         option="--artifacts-url",
-        default=os.getenv('TMT_REPORT_ARTIFACTS_URL'),
+        default=_str_env_to_default('artifacts_url',
+                                    os.getenv('TMT_REPORT_ARTIFACTS_URL')),
         help="Link to test artifacts provided for report plugins.")
 
     launch_url: Optional[str] = None
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -163,7 +163,7 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
     artifacts_url: Optional[str] = field(
         metavar="ARTIFACTS_URL",
         option="--artifacts-url",
-        default=_str_env_to_default('artifacts_url', None),
+        default=os.getenv('TMT_REPORT_ARTIFACTS_URL'),
         help="Link to test artifacts provided for report plugins.")
 
     launch_url: Optional[str] = None


happz:
Both variables should be accepted, it’s not a one-or-the-other situation. `TMT_PLUGIN_REPORT_REPORTPORTAL_ARTIFACTS_URL` because it’s what tmt tells its users is the envvar corresponding to `artifacts-url` key/option of `report/reportportal` plugin, according to general key-to-envvar schema, and `TMT_REPORT_ARTIFACTS_URL` because it’s what tmt agreed upon with its users should exist for the benefit of all (report) plugins. The former was not supported, incorrectly, and by fixing it we broke the latter.

kkaarreell:
> Both variables should be accepted, it’s not a one-or-the-other situation. `TMT_PLUGIN_REPORT_REPORTPORTAL_ARTIFACTS_URL` because it’s what tmt tells its users is the envvar corresponding to `artifacts-url` key/option of `report/reportportal` plugin, according to general key-to-envvar schema, and `TMT_REPORT_ARTIFACTS_URL` because it’s what tmt agreed upon with its users should exist for the benefit of all (report) plugins. The former was not supported, incorrectly, and by fixing it we broke the latter.

I have updated PR so it accepts both values. Still, I do not think this is the right approach. tmt exports many `TMT_` variables and if whatever tmt plugin decides to present that information to a user I don't believe the right approach is to introduce duplicate variables.

KwisatzHaderach (approved):

happz (approved):

psss (approved):

lukaszachy (approved):

4N0body5 (approved):

happz:
> > Both variables should be accepted, it’s not a one-or-the-other situation. `TMT_PLUGIN_REPORT_REPORTPORTAL_ARTIFACTS_URL` because it’s what tmt tells its users is the envvar corresponding to `artifacts-url` key/option of `report/reportportal` plugin, according to general key-to-envvar schema, and `TMT_REPORT_ARTIFACTS_URL` because it’s what tmt agreed upon with its users should exist for the benefit of all (report) plugins. The former was not supported, incorrectly, and by fixing it we broke the latter.
> 
> I have updated PR so it accepts both values. Still, I do not think this is the right approach. tmt exports many `TMT_` variables and if whatever tmt plugin decides to present that information to a user I don't believe the right approach is to introduce per-plugin variable duplicating the value.

The schema described in https://tmt.readthedocs.io/en/stable/overview.html#plugin-variables should work automagically for all plugins and all their keys. Once we start rejecting this mechanism for individual options, we should also start keeping a list of exceptions so users are aware it's not worth following the schema, because it no longer applies to all plugins and all their keys anymore. As I see it, accepting both envvars hurts nobody and keeps things consistent.

(The plugin-specific option existed before the plugin-agnostic one, so, technically, `TMT_REPORT_ARTIFACTS_URL` is the duplicate... :)

PR #3197, opened 2024-09-06
Pre commit bump

martinhoyer:
Another monthly version bump which could be automated, but not really because there are too many failures but we're getting there.


martinhoyer:
@happz this is me commiting e6ecfc19607117309835a51514e50eeffb163f6d   

, so if you can take a look, it would be much appreciated.

tmt/utils/__init__.py
```diff
@@ -5522,7 +5522,8 @@ def dataclass_normalize_field(
     value = raw_value
 
     if dataclasses.is_dataclass(container):
-        _, _, _, _, metadata = container_field(type(container), keyname)
+        tc = type(container) if not isinstance(container, type) else container
+        _, _, _, _, metadata = container_field(tc, keyname)
```
happz, line 5526: `tc` seems to have only purpose in life, to make the line shorter. Maybe some indentation could achieve the same, something like the following:

```suggestion
        _, _, _, _, metadata = container_field(
          type(container) if not isinstance(container, type) else container, keyname)
```

happz:
LGTM. BTW dropping `pre-commit` label, IIUIC, it's rather "tmt in pre-commit" than "pre-commit in tmt".

tmt/utils/__init__.py
martinhoyer: And readability, but yeah, sure.

martinhoyer:
> LGTM. BTW dropping `pre-commit` label, IIUIC, it's rather "tmt in pre-commit" than "pre-commit in tmt".

Hmm, make sense. Thanks!

happz (approved):

janhavlin (approved):

KwisatzHaderach (approved):

seberm (approved):

happz:
Unrelated failure, merging.

PR #3199, opened 2024-09-09
Add logfile size upload limit to the `reportportal` plugin

kkaarreell:
Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema
* [x] include a release note

### DIFF
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -9,6 +9,7 @@ import requests
 import tmt.hardware
 import tmt.log
 import tmt.steps.report
+import tmt.utils
 from tmt.result import ResultOutcome
 from tmt.utils import field, yaml_to_dict
 
@@ -46,10 +47,7 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
 
 
 def _size_env_to_default(option: str, default: 'Size') -> 'Size':
-    env_var = 'TMT_PLUGIN_REPORT_REPORTPORTAL_' + option.upper()
-    if env_var not in os.environ or os.getenv(env_var) is None:
-        return default
-    return tmt.hardware.UNITS(os.getenv(env_var, str(default)))
+    return tmt.hardware.UNITS(_str_env_to_default(option, str(default)))
 
 
 @dataclasses.dataclass
@@ -198,7 +196,10 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
         help=f"""
               Size limit in bytes for log upload to ReportPortal.
               The default limit is {DEFAULT_LOG_SIZE_LIMIT}.
-              """)
+              """,
+        normalize=tmt.utils.normalize_data_amount,
+        serialize=lambda limit: str(limit),
+        unserialize=lambda serialized: tmt.hardware.UNITS(serialized))
 
     traceback_size_limit: 'Size' = field(
         option="--traceback-size-limit",
@@ -207,7 +208,10 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
         help=f"""
               Size limit in bytes for traceback log upload to ReportPortal.
               The default limit is {DEFAULT_TRACEBACK_SIZE_LIMIT}.
-              """)
+              """,
+        normalize=tmt.utils.normalize_data_amount,
+        serialize=lambda limit: str(limit),
+        unserialize=lambda serialized: tmt.hardware.UNITS(serialized))
 
     exclude_variables: str = field(
         option="--exclude-variables",
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -15,6 +15,8 @@ if TYPE_CHECKING:
     from tmt._compat.typing import TypeAlias
 
 JSON: 'TypeAlias' = Any
+DEFAULT_LOG_SIZE_LIMIT: int = 1024 * 1024
+DEFAULT_TRACEBACK_SIZE_LIMIT: int = 50 * 1024
 
 
 def _flag_env_to_default(option: str, default: bool) -> bool:
@@ -41,11 +43,54 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
     return str(os.getenv(env_var))
 
 
-def _filter_invalid_chars(data: str) -> str:
+class LogFilterSettings:
+    size: int
+    is_traceback: bool
+
+    def __init__(self,
+                 size: int = DEFAULT_LOG_SIZE_LIMIT,
+                 is_traceback: bool = False):
+        self.size = size
+        self.is_traceback = is_traceback
+
+
+def _filter_invalid_chars(data: str,
+                          settings: LogFilterSettings) -> str:
     return re.sub(
         '[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+',
         '',
         data)
+
+
+def _filter_log_per_size(data: str,
+                         settings: LogFilterSettings) -> str:
+    size = len(data)
+    if settings.is_traceback:
+        variable = "TMT_PLUGIN_REPORT_REPORTPORTAL_LOG_SIZE_LIMIT"
+        option = "--traceback-size-limit"
+    else:
+        variable = "TMT_PLUGIN_REPORT_REPORTPORTAL_LOG_SIZE_LIMIT"
+        option = "--log-size-limit"
+    header = (f"WARNING: Uploaded log has been truncated because its size {size} bytes "
+              f"exceeds tmt reportportal plugin limit of {settings.size} bytes."
+              f"The limit is controlled with {option} plugin option or"
+              f"{variable} environment variable.\n\n")
+    if size > settings.size:
+        return f"{header}{data[:settings.size]}"
+    return data
+
+
+_LOG_FILTERS = [
+    _filter_log_per_size,
+    _filter_invalid_chars,
+    ]
+
+
+def _filter_log(log: str, settings: Optional[LogFilterSettings] = None) -> str:
+    settings = settings or LogFilterSettings()
+    for log_filter in _LOG_FILTERS:
+        log = log_filter(log, settings=settings)
+    return log
 
 
 @dataclasses.dataclass
@@ -141,6 +186,28 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
              Pass the defect type to be used for failed test, which is defined in the project
              (e.g. 'Idle'). 'To Investigate' is used by default.
              """)
+
+    log_size_limit: int = field(
+        option="--log-size-limit",
+        metavar="LOG_SIZE_LIMIT",
+        default=int(
+            _str_env_to_default('log_size_limit', str(DEFAULT_LOG_SIZE_LIMIT))),
+        help=f"""
+              Size limit in bytes for log upload to ReportPortal.
+              The default limit is {DEFAULT_LOG_SIZE_LIMIT} bytes
+              ({DEFAULT_LOG_SIZE_LIMIT / 1024 / 1024} MB).
+              """)
+
+    traceback_size_limit: int = field(
+        option="--traceback-size-limit",
+        metavar="TRACEBACK_SIZE_LIMIT",
+        default=int(
+            _str_env_to_default('traceback_size_limit', str(DEFAULT_TRACEBACK_SIZE_LIMIT))),
+        help=f"""
+              Size limit in bytes for traceback log upload to ReportPortal.
+              The default limit is {DEFAULT_TRACEBACK_SIZE_LIMIT} bytes
+              ({DEFAULT_TRACEBACK_SIZE_LIMIT / 1024} kB).
+              """)
 
     exclude_variables: str = field(
         option="--exclude-variables",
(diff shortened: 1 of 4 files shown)


tmt/steps/report/reportportal.py
```diff
@@ -142,6 +150,15 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
              (e.g. 'Idle'). 'To Investigate' is used by default.
              """)
 
+    log_size_limit: str = field(
+        option="--log-size-limit",
+        metavar="LOG_SIZE_LIMIT",
+        default=_str_env_to_default('log_size_limit', '1048576'),
+        help="""
+             Size limit in bytes for log upload to ReportPortal.
+             The default limit is 1048576 bytes (1 MB).
```
happz, line 159: The default value should not be encoded in two places, in a string. Please, make it a module-level "constant", e.g. `DEFAULT_LOG_SIZE_LIMIT: int = 1048576`, then use it in calls and texts: `_str_env_to_default('log_size_limit', str(DEFAULT_LOG_SIZE_LIMIT))`, `f"""... The default limit is {DEFAULT_LOG_SIZE_LIMIT} bytes ({DEFAULT_LOG_SIZE_LIMIT / 1024 / 1024} MB)."""`

Yes, it's more work, but the magic number is now defied at one place and one place only, all three mentions are derived from this single value, and note that I defined it as an integer, because it is an integer, it's a size limit - the fact that `_str_env_to_default()` needs a string should not dictate and bend the type of this value, that's a problem of `_str_env_to_default()`.

Also, the type of this field should also be `int`, I think `normalize_int(_str_env_to_default(...))` should work, and it should always either return an integer or raise an error e.g. when a user tries `--log-size-limit=foo`.

tmt/steps/report/reportportal.py
```diff
@@ -590,15 +607,19 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
                             response = self.rp_api_post(
                                 session=session,
                                 path="log/entry",
-                                json={"message": _filter_invalid_chars(log),
+                                json={"message": _filter_invalid_chars(
+                                    _filter_log_per_size(log,
+                                                         int(self.data.log_size_limit))),
```
happz, line 612: It was one filter at first, but now we have two, which means there can easily be seven next month :) How about detaching them from the code that needs to call them? With some simple "registry":

```python
def _filter_invalid_chars(data: str, step_data: 'ReportReportPortalData') -> str:
    return re.sub(
        '[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+',
        '',
        data)


def _filter_log_per_size(data: str, step_data: 'ReportReportPortalData') -> str:
    size = len(log)
    if size > step_data.log_size_limit:
        return (f"Log file was not uploaded to ReportPortal because its size {size} bytes"
                f"exceeds {step_data.log_size_limit} bytes.")
    return log


_LOG_FILTERS = [
  _filter_invalid_chars,
  _filter_log_per_size
]


def _filter_log(self, log: str) -> str:
  for log_filter in _LOG_FILTERS:
    log = log_filter(log, self.data)
  return log

message = self._filter_log(log)
```

happz:
Would it be useful to submit at least an abridged version of the log, with the "too long, had to cut" header? With the proposed code, user gets a note and nothing more, while even a shortened log might bring at least some value.

tmt/steps/report/reportportal.py
kkaarreell: Yes, that makes sense and thank you for such an extensive reasoning. :-)

tmt/steps/report/reportportal.py
```diff
@@ -607,19 +622,15 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
                             response = self.rp_api_post(
                                 session=session,
                                 path="log/entry",
-                                json={"message": _filter_invalid_chars(
-                                    _filter_log_per_size(log,
-                                                         int(self.data.log_size_limit))),
+                                json={"message": self.data._filter_log(log),
```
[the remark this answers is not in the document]
kkaarreell: Hm, probably I got confused by your proposal to use `ReportReportPortalData` type in an argument. I though that by bonding it with this specific class including it inside of it would just simplify data access. But overall I agree it is not that much related, but I would personally avoid using ReportReportPortalData type argument completely and revert back to the `limit` argument. OTOH, that was probably a way how to overcome the varying number of arguments in filter functions, correct?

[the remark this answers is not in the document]
happz: Exactly, it was a way how to reduce both (and future) filter callable to the same denominator while giving them access to phase parameters.

tmt/steps/report/reportportal.py
kkaarreell: Btw, for now I didn't use `normalize_int` as it requires `logger` as a 2nd argument.

tmt/steps/report/reportportal.py
kkaarreell: Updated.

tmt/steps/report/reportportal.py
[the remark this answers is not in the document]
kkaarreell: Updated.

tmt/steps/report/reportportal.py
```diff
@@ -41,13 +42,34 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
...
         '',
         data)
 
 
+def _filter_log_per_size(data: str, step_data: 'ReportReportPortalData') -> str:
+    size = len(data)
+    if size > step_data.log_size_limit:
+        return (f"WARNING: Log file was truncated because its size {size} bytes "
+                f"exceeds {step_data.log_size_limit} bytes.\n\n"
+                f"{data[:step_data.log_size_limit]}")
```
martinhoyer, line 57: Wouldn't adding the warning message make the returned str exceed the size limit again?  

```suggestion
                f"{data}"[:step_data.log_size_limit])
```

kkaarreell: Well, I guess this is a semantics problem. Does the limit apply to the original log or the uploaded log. The primary motivation was the former, though I can definitely modify it.

happz: I understood the limit is imposed by RP, therefore it was applied to test logs so they would not be rejected by RP. That would mean the warning needs to be squeezed into the limit as well, trimming the original log a bit more.

Maybe we should keep this filter at the last place, to trim whatever previous filters prepared.

kkaarreell: The limit is not imposed by RP, it is our decision to introduce the limit as uploading large logs increases RP load and makes its reaction very slow or freeze.. or fail upload eventually. We already encountered few tests producing huge logs (tens of MBs). Moreover, we want to save disk space.

martinhoyer: I also misunderstood. Make sense then. No objections from my side, except perhaps adding a comment it is not RP imposed limit to prevent future misunderstandings.

happz: I see, mea culpa then.

kkaarreell: I have modified the warning message, please check.

martinhoyer: Apologies for the late reply. LGTM.

martinhoyer (approved):

happz (approved):

tmt/steps/report/reportportal.py
```diff
@@ -142,6 +166,17 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
              (e.g. 'Idle'). 'To Investigate' is used by default.
              """)
 
+    log_size_limit: int = field(
+        option="--log-size-limit",
+        metavar="LOG_SIZE_LIMIT",
```
psss, line 171: Shouldn't this be just `SIZE` or `LIMIT` or `BYTES` or something similar?

psss (changes requested):
Thanks for adding this! Looks good, only we should also update the schema to add the newly supported option. Also, it might be worth at least a short release note?

tmt/steps/report/reportportal.py
happz: I didn't comment on this because `reportportal` docstring & options help texts are in a queue for complete review, including metavars. So, getting one right didn't look important to me when the rest needed to be polished too.

kkaarreell: In fact, I am considering introducing one more (smaller) limit specifically for traceback file.

therazix (approved):

tmt/steps/report/reportportal.py
psss: @kkaarreell, so do you still plan some changes here? Will you add the release note / schema adjustment? Or shall I?

psss:
Missing schema update, release note, open question, moving to `1.38`.

tmt/steps/report/reportportal.py
kkaarreell: Hi @psss , I have just pushed one more commit as I have decided to separate ordinary test logs and error/traceback logs that are being used for auto-analysis. I believe such a log should be even smaller.

martinhoyer: I belive that would be `tmt/schemas/report/reportportal.yaml` and `docs/releases.rst`

kkaarreell:
@psss @happz @martinhoyer Hi, I did some updates in the script so please review it whether it makes sense. I have decided to introduce separate limit for traceback log as this is the input for RP AutoAnalysis and I believe that it doesn't make sense to match 1MB logs.

tmt/steps/report/reportportal.py
```diff
@@ -41,13 +43,56 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
...
     return re.sub(
         '[^\u0020-\uD7FF\u0009\u000A\u000D\uE000-\uFFFD\U00010000-\U0010FFFF]+',
         '',
         data)
 
 
+def _filter_log_per_size(data: str,
+                         settings: LogFilterSettings) -> str:
+    size = len(data)
+    if settings.is_traceback:
```
happz, line 68: It seems to me the following lines are needed only when `if size > settings.size`, we should move them under that condition then, and construct these variables only when they are really needed.

tmt/steps/report/reportportal.py
```diff
@@ -41,13 +43,56 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
     return str(os.getenv(env_var))
 
 
-def _filter_invalid_chars(data: str) -> str:
+class LogFilterSettings:
```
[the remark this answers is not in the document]
kkaarreell: DOne

tmt/steps/report/reportportal.py
kkaarreell: Done

tmt/steps/report/reportportal.py
```diff
@@ -41,13 +43,51 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
...
         '',
         data)
 
 
+def _filter_log_per_size(data: str,
+                         settings: LogFilterSettings) -> str:
+    size = len(data)
+    if size > settings.size:
+        if settings.is_traceback:
+            variable = "TMT_PLUGIN_REPORT_REPORTPORTAL_LOG_SIZE_LIMIT"
```
therazix, line 65: Shouldn't the variable be named `TRACEBACK_SIZE_LIMIT` here?
```suggestion
            variable = "TMT_PLUGIN_REPORT_REPORTPORTAL_TRACEBACK_SIZE_LIMIT"
```

kkaarreell: Fixed.

tmt/steps/report/reportportal.py
```diff
@@ -41,13 +43,51 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
...
+    if size > settings.size:
+        if settings.is_traceback:
+            variable = "TMT_PLUGIN_REPORT_REPORTPORTAL_TRACEBACK_SIZE_LIMIT"
+            option = "--traceback-size-limit"
+        else:
+            variable = "TMT_PLUGIN_REPORT_REPORTPORTAL_LOG_SIZE_LIMIT"
+            option = "--log-size-limit"
+        header = (f"WARNING: Uploaded log has been truncated because its size {size} bytes "
+                  f"exceeds tmt reportportal plugin limit of {settings.size} bytes."
+                  f"The limit is controlled with {option} plugin option or"
```
therazix, line 72: If I am not mistaken, there will be no space between these sentences. Shouldn't it be like this?
```suggestion
                  f"exceeds tmt reportportal plugin limit of {settings.size} bytes. "
                  f"The limit is controlled with {option} plugin option or "
```

happz:
Rebased & moved release note where it belongs.

In f617cb31, I tried to switch from `int` to `tmt.hardware.Size` which should be the language to use when speaking about amounts of data. Out of the box, it should allow users to use whatever input they wish, as long as it specifies an amount of data (`--log-size-limit 1GB`, `log-size-limit: 7 MiB`, `log-size-limit: 1024 bytes`, ...) & plugin does not need to worry about conversions and printing, Pint takes care of everything as long as we 1. convert to same base when needed (`.to('bytes')`) and 2. use `.magnitude` when needed.

kkaarreell:
> Rebased & moved release note where it belongs.
> 
> In [f617cb3](https://github.com/teemtee/tmt/commit/f617cb3108db0aade145776fa1cf1f52cfe014ea), I tried to switch from `int` to `tmt.hardware.Size` which should be the language to use when speaking about amounts of data. Out of the box, it should allow users to use whatever input they wish, as long as it specifies an amount of data (`--log-size-limit 1GB`, `log-size-limit: 7 MiB`, `log-size-limit: 1024 bytes`, ...) & plugin does not need to worry about conversions and printing, Pint takes care of everything as long as we 1. convert to same base when needed (`.to('bytes')`) and 2. use `.magnitude` when needed.
> 
> @kkaarreell I tried to test it a bit, let me know what do you think.
> 
> Edit: still testing, it wasn't ready.

Sounds great. I will try to test it as well later this week.

PR #3200, opened 2024-09-09
Save the subresults for tmt-report-result

seberm:
All the results generated by `tmt-report-results` become objects of `tmt.result.SubResult` and they get assigned under a parent `tmt.Result`. Also, set the `TESTID` and `BEAKERLIB_COMMAND_REPORT_RESULT` test environment variables, so every call of beakerlib's `rlPhaseEnd` macro becomes a subresult.

Related to:

Blocked by:
- https://github.com/teemtee/tmt/pull/3270

### TODOs:
- [x] Try to find all the possible documentation places where this needs to be modified/mentioned - please provide tips.
- Add tests
    - [x] For beakerlib subresults
    - [x] For shell subresults
    - [x] Add tests for subresults logs reported by `tmt-report-result -o <file>`
    - [x] Any more tests? Any ideas?
- [x] Fix existing tests / fix the code to pass tests
- [x] Any special handling for `result: restraint`? -> No. The subresults for test marked with `result: restraint` will be handled the same way as before (they get converted to "main" tmt custom results). The new thing here is that beakerlib phases are now also interpreted as tmt subresults. This means that in combination with `result: restraint`, the phases get also converted as main tmt custom results.
- [x] Missing subresults logs when `tmt-report-result` gets called and log file is provided as a third argument (`rstrnt-report-result: TESTNAME TESTRESULT LOGFILE [METRIC]` supports that). With `-o` everything works as expected. Is this a bug?
    - (answer from @happz ): `tmt-report-result` is our interface, we implement whatever we need to be there in a form we prefer and think is the best suited for the task of reporting results to tmt, and that comes with a set of options - it's not a 1:1 replacement for `rstrnt-report-result`.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note


tmt/schemas/results.yaml
```diff
@@ -135,6 +139,12 @@ items:
     duration:
       $ref: "#/definitions/duration"
 
+    # TODO: Fixme? e.g. using ref. otherwise I am getting the error:
```
seberm, line 142: @happz Aren't we missing the `data-path` in the `result` and `subresult` schemas? Otherwise, I'm still getting the "result format violation" errors.

happz: Result schema is definitely missing `data-path` key, subresults shouldn't have one, IIUIC, data path is a path in which tmt stores test data and info and stuff, so there is no dedicated subpath for a subresult.

seberm: Are you sure about the `data-path` in the subresult? I do not have much insight, but it seems the subresults generated by `tmt-report-result` are using this field when the `tmt-report-results.yaml` gets created:

- https://github.com/teemtee/tmt/blob/7b16db66907e9b2c62cc0643abd27a3a120e4a26/tmt/steps/execute/scripts/tmt-report-result#L20

happz: Hmmm, that does not look right. `TMT_TEST_EXE` is `${TMT_TREE/tree/execute}`, `TMT_TEST_DATA` is said to be "Path to the directory where test can store logs and other artifacts generated during its execution." and I can't imagine that `${TMT_TEST_DATA/$TMT_TREE/tree/execute}` could be anything sane.

@psss WDYT? There is "test data" for the test, `TMT_TEST_DATA` points to it, but data paths for Beakerlib phases? tmt does not manage those, in custom results we do not even have any idea whether the test used any subresult data paths, so we don't touch them, this seems like a mistake to me.

tests/execute/result/subresults/test.fmf
```diff
@@ -0,0 +1,32 @@
...
+    summary: Beakerlib rlPhaseEnd as tmt subresult
+
+    # Explicitly set the TESTID to non-empty value. Also, set the
+    # BEAKERLIB_COMMAND_REPORT_RESULT to `tmt-report-result` explicitly. The
+    # command in this variable gets called with every `rlPhaseEnd`.
+    #
+    # Refs:
+    # - https://github.com/teemtee/tmt/issues/2826#issue-2225993479
+    # - https://github.com/teemtee/tmt/issues/2826#issuecomment-2085014960
+    environment:
```
seberm, line 25: We should discuss the handling of the `TESTID` and `BEAKERLIB_COMMAND_REPORT_RESULT` env variables. These variables must be set to generate the subresults from `rlPhaseEnd` macros. Without them, only a "parent" result is generated. The `TESTID` for example, could be maybe set to the parent result's `serial-number`.

Do you think it's a good idea to somehow set/override these variables directly in the tmt or should we just keep them unset, create documentation, and let the user decide?

For more info, please see:
- https://github.com/teemtee/tmt/issues/2826#issuecomment-2085014960
- https://github.com/beakerlib/beakerlib/blob/4321c54e2e99f74b91c9376aa4324dafd0300c3d/src/testing.sh#L1075

CC: @psss @happz @lukaszachy

seberm (commented):
I've tried to change the method docstrings, `releases.rst` and `report-result.fmf` story to reflect the subresults feature. Are there any other places where I should document this feature?

Let me know if you need any further adjustments / tests. Thanks.

spec/plans/results.fmf
```diff
@@ -141,6 +141,9 @@ description: |
           check:
             ...
 
+          # String, path to /data directory storing possible test artifacts
+          data-path: path/to/test/data
```
happz, line 145: We do have `data-path` for the whole test, `$TMT_TEST_DATA`, but should we have one for subresults as well?

tests/execute/result/subresults/test.fmf
seberm: As discussed today, the env variables are set internally by tmt the same way as for `BEAKERLIB_COMMAND_SUBMIT_LOG`. I also tried adding a test that checks whether these values are actually set.

tmt/schemas/results.yaml
seberm: Also, do you think we should somehow limit the `data-path` using a `pattern` in the schema?

For a parent result, the value is always a relative path (`data/guest/default-0/tests/example-1/data`). For subresults, the `data-path` seems to always start with `pattern: "^/"` - it's not relative, but absolute where root starts at 'execute' directory, e.g. `/data/guest/default-0/tests/example-1/data`.

Isn't it a bug? Also, the subresult's `data-path` seems to be always the same as the parent `data-path` (without `/` prefix).

tmt/steps/execute/__init__.py
```diff
@@ -854,7 +786,20 @@ def extract_tmt_report_results(self, invocation: TestInvocation) -> list["tmt.Re
...
 
-        return self._process_results_reduce(invocation, collection.results)
+        # TODO: Can we use the Result.from_invocation(...) here? Should I somehow map the results
+        # to the test invocation here or not?
+        results = [tmt.Result.from_serialized(r) for r in collection.results]
+
+        # TODO: Should I somehow map the Result to test invocation?
+        for result in results:
+            # Ensure the result name is in test namespace
+            # TODO: Do we want to move it under main test namespace or not, I suppose we do.
```
seberm, line 796: I suppose we want to add a parent test name as a namespace/prefix to all subresults, correct?

tmt/steps/execute/__init__.py
```diff
@@ -854,7 +786,20 @@ def extract_tmt_report_results(self, invocation: TestInvocation) -> list["tmt.Re
 
         collection.validate()
 
-        return self._process_results_reduce(invocation, collection.results)
+        # TODO: Can we use the Result.from_invocation(...) here? Should I somehow map the results
+        # to the test invocation here or not?
```
seberm, line 790: Do we want to somehow map the subresults to the `TestInvocation`? I somehow feel there is still a space for the improvement in the code below. Any tips/recommendations?

tmt/schemas/results.yaml
seberm: @psss What is your opinion about `data-path` for subresult? Thanks.

seberm (commented):
Hello @psss @happz , I've tried to modify the implementation of the subresults saving according to your recommendations. Hopefully, I got it right. Could you please take a look?

I've also added more comments, improved docstrings and the release notes.

The tests seem to pass. I've only encountered errors in fedora-40 checks which seem unrelated, but I plan to recheck them again just to be sure.

Thanks!

seberm:
Also, do you think the `result: restraint` requires special handling for subresults? From my understanding, I don't believe it does. Please, correct me If I am wrong.

tmt/frameworks/__init__.py
```diff
@@ -108,15 +108,31 @@ def get_pull_options(
 
         return []
 
+    @classmethod
+    def results_to_subresults(cls, results: list[tmt.result.Result]) -> list[tmt.result.SubResult]:
+        """ Convert given results to tmt subresults """
+
+        return [
```
seberm, line 115: TODO: Try to find a better way to convert `tmt.Result` to `tmt.result.SubResult`, e.g. using their `to_spec/from_spec`. CC: @happz

happz: I'd probably move this method into `Result` class, on the assumption that `Result` knows best how it should be "lowered" into `Subresult`. Its presence here looks like an implementation leak, suddenly frameworks own the secret sauce, instead of `tmt.result`.

tmt/steps/execute/__init__.py
happz: I *think* we should do that, but in a safe way - `from_test_invocation()` would change some things it shouldn't touch, like `name`, but also some very important things that need to be in order, like `serial-number`. Maybe we need is some kind of "sanitize" method that would do what was (is?) in `_process_results_partials()`:

```python
            # For the result representing the test itself, set the important
            # attributes to reflect the reality.
            if partial_result.name == test.name:
                partial_result.start_time = invocation.start_time
                partial_result.end_time = invocation.end_time
                partial_result.duration = invocation.real_duration
                partial_result.context = self.step.plan._fmf_context
```

I feel like I can't really focus at this moment, but this looks like an important step - we shouldn't probably mess with all names, but we should make sure timestamps and context and serial number are reported correctly, if nothing else. But then again, `from_test_invocation()` does exactly that, and if frameworks use it to construct their overall tests, we don't need to worry about it *here*.

tmt/steps/execute/__init__.py
seberm: I've moved the namespacing into the `to_subresult(invocation)` method of the `tmt.Result`. If think it makes sense to have it there. I still have a feeling the namespacing should be probably enforced if the Result **gets converted** to a SubResult.

The framework can now pass the results unmodified (it will not call the `to_subresult` method).

- Or we can modify the behavior of the `to_subresult` method with some flags, e.g. `to_subresult(add_test_namespace=False)`.
- Or we can optionally add namespace only if the `invocation` argument of `to_subresult(invocation)` is set.

tmt/frameworks/__init__.py
seberm: Great tip. Makes sense - moved to `tmt.Result.to_subresult`.

psss:
This is one of the most important blockers on the way to the full subresults support, proposing as a `must` have.

docs/releases.rst
```diff
@@ -4,6 +4,27 @@
...
+still has an impact on the parent test result. See also
+:ref:`/stories/features/report-result`.
+
+The beakerlib test framework tests now generate tmt subresults. The behavior is
+very similar to the shell test framework with ``tmt-report-result`` command
+calls (see above). The ``tmt-report-result`` now gets called with every
+``rlPhaseEnd`` macro and the tmt subresult gets created. The
+difference is that the subresults outcomes do not influence the parent result
+outcome in any way - the parent result outcome is always set to the outcome
+reported by the Beaker. The subresults are always assigned under the parent tmt
```
happz, line 23: This is and isn't true - subresults do not influence the parent result *as far as tmt is interested*, but beakerlib *does* count them in. A failed phase - subresult - will affect the whole test outcome as reported by beakerlib framework. How about something like this?

> The difference is that the subresults outcomes are not evaluated by tmt, tmt only captures them, and tmt relies on Beakerlib and its result reporting - which does take outcomes of phases into account - to determine the final test outcome. The subresults are always assigned under the parent tmt result...

tmt/schemas/results.yaml
happz: I see the attribute is still part of the patch - @psss, please, chime in.

tmt/frameworks/__init__.py
```diff
@@ -112,11 +112,14 @@ def get_pull_options(
...
             cls,
             invocation: 'TestInvocation',
+            results: list[tmt.result.Result],
             logger: tmt.log.Logger) -> list[tmt.result.Result]:
         """
         Extract test results.
 
         :param invocation: test invocation to which the check belongs to.
+        :param results: list of results which will be optionally saved as subresults by a
+            framework.
```
happz, line 122: "which will be optionally saved as subresults by a framework" - or not, who knows what framework's gonna do? Maybe they are good enough. I'd recommend sticking to what they are, not what is happening to them in the two frameworks we have today, i.e. something like "current list of results as reported by test via ``results.yaml``."

tmt/frameworks/beakerlib.py
```diff
@@ -53,8 +62,15 @@ def get_pull_options(
     def extract_results(
             cls,
             invocation: 'TestInvocation',
+            results: list[tmt.result.Result],
             logger: tmt.log.Logger) -> list[tmt.result.Result]:
         """ Check result of a beakerlib test """
+
+        # The outcome of a parent tmt result must be never modified based on subresults outcomes.
+        # The parent result outcome will be always set to outcome reported by a Beaker. The
```
happz, line 70: "beakerlib" instead of "Beaker". Beaker is a provisioning service and/or test runner, beakerlib is the helper library and test framwork.

tmt/frameworks/beakerlib.py
```diff
@@ -66,7 +82,7 @@ def extract_results(
         beakerlib_results_filepath = invocation.path / 'TestResults'
 
         try:
-            results = invocation.phase.read(beakerlib_results_filepath, level=3)
+            bkr_results = invocation.phase.read(beakerlib_results_filepath, level=3)
```
happz, line 85: I'd recommend using `beakerlib_` prefix, if one is needed, `bkr_` does smell like "Beaker" which is something completely different.

tmt/frameworks/shell.py
```diff
@@ -23,30 +21,118 @@ def get_test_command(
...
+
+            return [tmt.Result.from_test_invocation(
+                invocation=invocation,
+                result=result,
+                log=[invocation.relative_path / tmt.steps.execute.TEST_OUTPUT_FILENAME],
+                note=note)]
+
+        # Handle the `tmt-report-result` command results as a single test with assigned tmt
+        # subresults.
+        return cls._process_results_reduce(invocation, results)
```
happz, line 138: If you switch the condition above and switch the order, `if results:`, you can `return cls._process_results_reduce(invocation, results)` right from that branch above, and the whole block under `if not results:` could be indented one level to the left, which could make a slightly better diff - the code probably did not change that much, it's the extra indentation that inflates it.

tmt/steps/execute/__init__.py
```diff
@@ -854,7 +786,7 @@ def extract_tmt_report_results(self, invocation: TestInvocation) -> list["tmt.Re
 
         collection.validate()
 
-        return self._process_results_reduce(invocation, collection.results)
+        return [tmt.Result.from_serialized(r) for r in collection.results]
```
happz, line 789: Check the docstring of this method, there's one line that no longer applies.

tmt/steps/execute/__init__.py
```diff
@@ -904,11 +836,17 @@ def extract_results(
...
-            return self.extract_tmt_report_results(invocation)
+            tmt_report_results = self.extract_tmt_report_results(invocation)
 
-        return invocation.test.test_framework.extract_results(invocation, logger)
+        # Propagate loaded `tmt_report_results` to test framework, which will handle these results
+        # accordingly (e.g. saves them as a tmt subresults).
+        return invocation.test.test_framework.extract_results(
+            invocation=invocation,
+            results=tmt_report_results,
+            logger=logger)
```
happz, line 849: Are the parameter names necessary? The call might fit on one line without them (and with a small change of `tmt_report_results` becoming `reported_results`, or even just `results`? :)

tmt/frameworks/__init__.py
seberm: Ok, changed to just `current list of results as reported by a test`.

stories/features/report-result.fmf
```diff
@@ -12,14 +12,13 @@ description: |
...
-    Note, that the only value currently processed by ``tmt`` is
-    the test result, other options are silently ignored.
+    all these calls will be internally saved as tmt subresults. The final
+    result will be the most severe rating. Available values ordered by severity
+    are SKIP, PASS, WARN and FAIL.
 
-    When ``tmt-report-result`` is called, the return value of the
-    test itself is ignored, and only result saved by
-    ``tmt-report-result`` is consumed by tmt.
+    When ``tmt-report-result`` is called, the return value of the test itself
```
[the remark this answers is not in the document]
seberm: Changed, marking as resolved.

tmt/frameworks/shell.py
seberm: Yes, of course! It’s strange that I didn’t notice this when I wrote the code. Thanks for pointing it out :).

tmt/steps/execute/__init__.py
seberm: Thanks, I've just removed the line.

tmt/steps/execute/__init__.py
seberm: Sure, modified. Thanks :)

tmt/schemas/results.yaml
seberm: @happz @psss Do you think we should remove `data-path` also from `tmt-report-results.yaml` generated by `tmt-report-result`?
- https://github.com/teemtee/tmt/blob/3803a4414365045a582ab13861580a0c25f1b642/tmt/steps/execute/scripts/tmt-report-result#L23

Or this one should be kept there for some reason (e.g. for `custom`/`restraint` results)?

happz: I'd probably try not emitting the key, there's no sane way to "guess" the data path anyway on this level, and tmt knows it and sets it for results when they are constructed (`data_path=invocation.relative_test_data_path` in `from_test_invocation()`).

seberm: - [ ] AI: Ok. So I will create a separate PR for these changes which I think should be merged before this PR:
- Remove `data-path` for subresults (scheme and SubResult class)
- Remove `data-path` from `tmt-report-result` script & run the full test, see if something breaks + fix the tests

seberm: I've pushed the changes, hopefully, this comment is now resolved. The removal of `data-path` from `tmt-report-results` will be covered in https://github.com/teemtee/tmt/pull/3270

tmt/result.py
```diff
@@ -363,6 +374,24 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
         return self
 
+    def to_subresult(self, invocation: 'tmt.steps.execute.TestInvocation') -> 'SubResult':
+        """ Convert result to tmt subresult """
+
+        # TODO: Try to find a better way for conversion (e.g. using
+        # from_spec/to_spec/from_serialized/to_serialized).
+        return SubResult(
+            # Add subresult name under the test namespace
+            name=f"{invocation.test.name}/{self.name.lstrip('/')}",
```
psss, line 384: Do we want to prefix the subtest name with the main test name? On the level of subtets there should be no collision and showing shorter names in the report seems as a better option from user perspective, I'd say. Currently it looks like this:

```
report
    how: display
        fail /test/beakerlib/bad
            fail /test/beakerlib/bad/Test
        pass /test/beakerlib/good
            pass /test/beakerlib/good/Test
        warn /test/beakerlib/need
            warn /test/beakerlib/need/Cleanup
        errr /test/beakerlib/weird (beakerlib: TestResults FileError)
    summary: 1 test passed, 1 test failed, 1 warn and 1 error
```

It could look like this:

```
report
    how: display
        fail /test/beakerlib/bad
            fail Test
        pass /test/beakerlib/good
            pass Test
        warn /test/beakerlib/need
            warn Cleanup
        errr /test/beakerlib/weird (beakerlib: TestResults FileError)
    summary: 1 test passed, 1 test failed, 1 warn and 1 error
```

psss (changes requested):
Looks great! Thanks a lot for working on this. Added some questions and suggestions, but nothing major.

martinhoyer (commented):
The bits I could understand looks good to me.

tmt/result.py
psss: @juk, any preference or expectations here?

tmt/frameworks/beakerlib.py
```diff
@@ -36,7 +36,16 @@ def get_environment_variables(
         return Environment({
             'BEAKERLIB_DIR': EnvVarValue(invocation.path),
             'BEAKERLIB_COMMAND_SUBMIT_LOG': EnvVarValue(
-                f'bash {tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.path}')
+                f'bash {tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.path}'),
+
+            # The command in this variable gets called with every `rlPhaseEnd` call in beakerlib.
+            'BEAKERLIB_COMMAND_REPORT_RESULT': EnvVarValue(
+                f'bash {tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.path}'),
```
happz, line 43: Should we use `tmt.utils.DEFAULT_SHELL` and maybe even `tmt.utils.SHELL_OPTIONS`? Or, isn't the `TMT_REPORT_RESULT_SCRIPT.path` already executable, and does not need any `bash` at all?

tmt/frameworks/shell.py
```diff
@@ -23,15 +21,103 @@ def get_test_command(
...
+
     @classmethod
     def extract_results(
             cls,
             invocation: 'TestInvocation',
+            results: list[tmt.result.Result],
             logger: tmt.log.Logger) -> list[tmt.result.Result]:
-        """ Check result of a shell test """
+        """
+
```
happz, line 98: Spurious empty line.

tmt/result.py
```diff
@@ -189,6 +189,14 @@ class CheckResult(BaseResult):
         serialize=lambda event: event.value,
         unserialize=CheckEvent.from_spec)
 
+    def to_subcheck(self) -> 'SubCheckResult':
+        """ Convert check to a tmt SubCheckResult """
+
+        # TODO: Try to find a better way for conversion?
+        # TODO: Some attributes will be probably modified for SubCheckResult e.g. name (namespace)
+        # TODO: Still need to check if this actually works
```
happz, line 197: So, does it work? :)

tmt/result.py
```diff
@@ -269,7 +277,9 @@ def from_test_invocation(
             result: ResultOutcome,
             note: Optional[str] = None,
             ids: Optional[ResultIds] = None,
-            log: Optional[list[Path]] = None) -> 'Result':
+            log: Optional[list[Path]] = None,
+            subresult: Optional[list[SubResult]] = None,
+            ) -> 'Result':
```
happz, line 282: (Formatting nitpicking `ruff` might make obsolete): `) -> 'Result':` belongs on the same line as `subresult`.

tmt/result.py
happz: +1 for *not* prefixing - I'd keep the phase/subtest name as provided by the test and/or framework.

tmt/result.py
```diff
@@ -363,6 +374,24 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
+            name=f"{invocation.test.name}/{self.name.lstrip('/')}",
+            result=self.result,
+            note=self.note,
+            log=self.log,
+            start_time=self.start_time,
+            end_time=self.end_time,
+            duration=self.duration,
+
+            # Also convert all CheckResult instances to SubCheckResult instances
+            check=[c.to_subcheck() for c in self.check])
```
happz, line 393: A single-letter variable that's not `i` or `j` :) See https://github.com/teemtee/tmt/pull/3200/files#r1799649164

tmt/result.py
seberm: If prefixing is not needed, that's great - it will simplify to code.

Just to be sure - Do we also want to remove the `/` prefix from the **subresult** name? I suppose yes.

```
fail /test/beakerlib/bad
            fail Test
```

vs.

```
fail /test/beakerlib/bad
            fail /Test
```

seberm: Ok, I've removed the code for prefixing. But the subresult names (both beakerlib and `tmt-report-result`) are **still** prefixed with `/`. This is done by `tmt-report-result` script for "shell" tests and for beakerlib this seems to be an automatic behavior. So, I suppose this answers my question :) - let's keep it on the framework.

```
report
    how: display
        fail /test/beakerlib
            pass /phase-setup
            pass /Internal-test-of-environment-variable-values
            pass /phase-test-pass
            fail /phase-test-fail
            pass /extra-tmt-report-result/good
            fail /extra-tmt-report-result/bad
            warn /extra-tmt-report-result/weird
            skip /extra-tmt-report-result/skip
            pass /phase-test-multiple-tmt-report-result
            pass /phase-cleanup
        fail /test/shell/fail
            pass /fail-subtest/good
            fail /fail-subtest/fail
            warn /fail-subtest/weird
            skip /fail-subtest/skip
        pass /test/shell/pass
            pass /pass-subtest/good0
            pass /pass-subtest/good1
            pass /pass-subtest/good2
        pass /test/shell/skip
            pass /skip-subtest/extra-pass
            skip /skip-subtest/extra-skip
    summary: 2 tests passed and 2 tests failed
```

docs/releases.rst
```diff
@@ -4,6 +4,31 @@
     Releases
 ======================
 
+tmt-1.38.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Each execution of ``tmt-report-result`` command inside a shell
+test will now create a tmt subresult. The parent result outcome is
```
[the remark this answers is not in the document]
seberm: Changed to `main result`, marking as resolved, thanks!

tests/execute/result/subresults/beaker-phases-subresults.sh
```diff
@@ -0,0 +1,47 @@
...
+    rlPhaseStartTest "Internal test of environment variable values"
+        rlRun "test -n \"\$TMT_TEST_SERIAL_NUMBER\" -o -n \"\$TESTID\"" 0 "Check the variables are not empty"
+        rlAssertEquals "TESTID must be set to TMT_TEST_SERIAL_NUMBER" "$TESTID" "$TMT_TEST_SERIAL_NUMBER"
+
+        rlRun "[[ \$BEAKERLIB_COMMAND_REPORT_RESULT =~ tmt-report-result$ ]]" 0 "Check the variable contains path to a tmt-report-result script"
+    rlPhaseEnd
+
+    rlPhaseStartTest "phase-test pass"
+        rlRun "echo mytest-pass | tee output" 0 "Check output"
+        rlAssertGrep "mytest-pass" "output"
```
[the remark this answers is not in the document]
seberm: Thanks for the tip. Modified.

tests/execute/result/subresults.sh
```diff
@@ -0,0 +1,126 @@
...
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "run_dir=\$(mktemp -d)" 0 "Create run directory"
+        rlRun "pushd subresults"
+        rlRun "set -o pipefail"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test the subresults were generated into the results.yaml"
+        rlRun "tmt run --id $run_dir --scratch -v 2>&1 >/dev/null | tee output" 1
```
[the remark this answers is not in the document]
seberm: Marking as resolved.

tests/execute/result/subresults/test.fmf
```diff
@@ -0,0 +1,28 @@
+/pass:
```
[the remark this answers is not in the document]
seberm: Added, thanks.

tmt/steps/execute/__init__.py
```diff
@@ -851,7 +779,7 @@ def extract_tmt_report_results(self, invocation: TestInvocation) -> list["tmt.Re
 
         collection.validate()
 
-        return self._process_results_reduce(invocation, collection.results)
+        return [tmt.Result.from_serialized(r) for r in collection.results]
```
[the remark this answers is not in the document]
seberm: Changed here and also in the code in frameworks :).

tests/execute/result/subresults/beaker-phases-subresults.sh
```diff
@@ -0,0 +1,47 @@
...
+    rlPhaseEnd
+
+    rlPhaseStartTest "phase-test fail"
+        rlRun "echo mytest-fail | tee output" 0 "Check output"
+        rlAssertGrep "asdf-asdf" "output"
+    rlPhaseEnd
+
+    rlPhaseStartTest "phase-test multiple tmt-report-result"
+        rlRun "touch /tmp/bkr_good_log"
```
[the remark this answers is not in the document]
seberm: Heh, good catch. It should be fixed now. Feel free to mark this as resolved.

tmt/frameworks/beakerlib.py
seberm: Thanks for the hint. Just `tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.path` seems to be working fine.

tmt/result.py
seberm: Modified, thanks.

tmt/result.py
seberm: :D, fixed.

What about adding some automatic lint checks for these single-letter variable names in ruff/pylint?
CC: @martinhoyer

tests/execute/result/subresults/main.fmf
```diff
@@ -0,0 +1,7 @@
+/plan:
```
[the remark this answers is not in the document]
seberm: Resolved, thanks.

tmt/steps/execute/__init__.py
seberm: Marking as resolved. We decided (in the comments below) to remove the main test prefix and keep the subresult name on the test framework.

tmt/result.py
seberm: Yes, it seems to work as expected:

```
ipdb> results[0]
CheckResult(name='dmesg', result=<ResultOutcome.SKIP: 'skip'>, original_result=<ResultOutcome.SKIP: 'skip'>, note=None, log=[], start_time='2024-10-15T20:53:46.202625+00:00', end_time='2024-10-15T20:53:46.202653+00:00', duration='00:00:00', event=<CheckEvent.BEFORE_TEST: 'before-test'>)
ipdb> results[0].to_subcheck()
SubCheckResult(name='dmesg', result=<ResultOutcome.SKIP: 'skip'>, original_result=<ResultOutcome.SKIP: 'skip'>, note=None, log=[], start_time='2024-10-15T20:53:46.202625+00:00', end_time='2024-10-15T20:53:46.202653+00:00', duration='00:00:00', event=<CheckEvent.BEFORE_TEST: 'before-test'>)
```

I tried to convert `Result` to `SubResult` the same way, but it's necessary to filter in only some of the `Result` fields. Would you mind to take a look?
- https://github.com/teemtee/tmt/pull/3200/commits/978fe968cb26d976061a803e8d54e0ad5376b448

Or if you think the direct mapping of the Result fields to SubResult ones is better, just let me know and I'll revert the commit.

Thanks :)

seberm: - [ ] TODO: Try to get the SubResult fields from the class in `to_subresult` method, do not define them explicitly. Try to use "container_*" methods from `tmt.utils`.

tests/execute/result/subresults/beaker-phases-subresults.sh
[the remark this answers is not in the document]
psss: Thanks. And we can drop the `pipefail` line.

tests/execute/result/subresults/beaker-phases-subresults.sh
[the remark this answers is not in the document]
psss: Thanks!

tmt/result.py
psss: Yeah, looks good. Both main result and subresult names look consistent. And we find this does not please user eye we can always remove later.

tests/execute/result/subresults/beaker-phases-subresults.sh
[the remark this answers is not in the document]
psss: Done in 1bba15d7.

tmt/steps/execute/__init__.py
psss: @seberm, @happz, I guess this one has been already covered?

seberm: I would say yes, it's resolved (at least from my perspective). I don't think any further changes are needed and we can mark this as resolved.

psss (approved):
Now looks good, thanks!

tmt/frameworks/beakerlib.py
psss: I guess the same applies for the `tmt-file-submit` script, proposing in #3310.

seberm:
Just a side note for users using the `result: restraint` in combination with `framework: beakerlib` - the tmt will now generate subresult for each beakerlib phase which gets converted to normal/main custom tmt result. This will probably generate a lot of new results for users using this feature.

@psss I've added note into release notes, please see and review: https://github.com/teemtee/tmt/pull/3200/commits/2ee252ecf7a28bc45dfbfa74581a254e1e4e63a6

Thanks

psss:
> @psss I've added note into release notes, please see and review: [2ee252e](https://github.com/teemtee/tmt/commit/2ee252ecf7a28bc45dfbfa74581a254e1e4e63a6)

Yeah, makes sense! :+1:

happz (approved):

PR #3201, opened 2024-09-09
Add a new `Test Runner` section to the guide

psss:
Let users know which collections are available on the test runner.

Pull Request Checklist


psss:
An old docs todo item. Blocked on Testing Farm having the docs published. Or are we ok with pointing directly to git?

happz:
I’m not sure we should add TF-motivated caveats and notes into tmt metadata specification. It’s basically spoiling the information space with info good for a single special user of tmt. In other words, if - when… - MS picks tmt for their CI system, will we amend tmt specification to include pecularities of that system as well?

What would you say of we had a page dedicated to “Running tmt in TF”, collecting this kind of information? BTW, this is exactly the sort of info I would look for on TF website - what tmt version, what plugins, what collections can I use, and so on. Metadata specification does not feel like a good place for hosting this info.

psss:
> What would you say of we had a page dedicated to “Running tmt in TF”, collecting this kind of information?

Yeah, makes sense. In d11db6d8 I've added a new `Testing Farm` section to the guide. @happz, @thrix, @martinhoyer, what do you think?

docs/guide.rst
```diff
@@ -1019,3 +1019,6 @@ Current Limits
 
 __ https://github.com/teemtee/tmt/labels/multihost
 __ https://github.com/teemtee/tmt/issues/2047
+
+
+.. include:: guide/testing-farm.inc.rst
```
happz, line 1024: How about adding a "Test runners" header, plus a short paragraph, something like "tmt can run in many environments, your mileage may vary, and here's what we know about various runners", then followed by the very nice text about TF you already have? Then the note should point to this "test runners" header rather than directly to the TF section.

I would like to avoid TF seeming as the only option. The note says "yep, things may be different" and points to TF page directly, leaving no space for any other test runner. And I'd say there's at least one other already, our workstations when we develop tests, for example. I think it would be very helpful and very tidy if we could collect what we know about that runner as well as TF, put together a collection of quirks, things to look for, special setup to be aware of when running tmt (locally|in TF|in Github|...). And that's why "test runners" leading to TF directly feel like we would neglect any other test runner already, and I'd like to keep that space open for the future additions.

psss: Yeah, makes sense, done in 3f53cb1c.

therazix (approved):

happz (approved):

KwisatzHaderach (approved):

happz:
Documentation-only PR, merging.

Issue #2872, opened 2024-04-18
Report `fail` in `tmt lint` for empty environment files


bgoncalv:
In our repo we have a lot of tests that when running tmt lint shows `warn: Empty environment file`

It would be nice if we could force tmt lint to report failure in this cases, afaik it is not handled as a warning class.

My suggestion would be running a command like below would force the lint to report failure.

```
tmt lint --outcome-only fail --enforce-check Cxxx .
```

psss:
For this specific case, when `environment-file` is pointing to an empty file, I would say we should even give a failure by default. Could be a good first issue, here's an example plan check:

https://github.com/teemtee/tmt/blob/22731b9761b62177de00ee7818c2f0ee4efcfe03/tmt/base.py#L2027-L2039

mcasquer:
@psss @happz I guess we can close this issue as the PR has been merged right?

PR #3204, opened 2024-09-11
Add a new `lint` check for empty environment files

mcasquer:
New function that raises a failure if the size of
the environment file, if this one exists, is zero.

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -2348,7 +2348,7 @@ class Plan(
         yield from _lint_step('finish')
 
     def lint_empty_env_files(self) -> LinterReturn:
-        """ P008: env files are not empty """
+        """ P008: environment files are not empty """
 
         env_files = self.node.get("environment-file") or []
 
@@ -2359,7 +2359,7 @@ class Plan(
         for env_file in env_files:
             env_file = (self.anchor_path / Path(env_file)).resolve()
             if not env_file.stat().st_size:
-                yield LinterOutcome.FAIL, f'the file "{env_file}" is empty'
+                yield LinterOutcome.FAIL, f"the environment file '{env_file}' is empty"
                 return
 
         yield LinterOutcome.PASS, 'no empty environment files'
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -2347,6 +2347,23 @@ class Plan(
         yield from _lint_step('execute')
         yield from _lint_step('finish')
 
+    def lint_empty_env_files(self) -> LinterReturn:
+        """ P008: env files are not empty """
+
+        env_files = self.node.get("environment-file") or []
+
+        if not env_files:
+            yield LinterOutcome.SKIP, 'no environment files found'
+            return
+
+        for env_file in env_files:
+            env_file = (self.anchor_path / Path(env_file)).resolve()
+            if not env_file.stat().st_size:
+                yield LinterOutcome.FAIL, f'the file "{env_file}" is empty'
+                return
+
+        yield LinterOutcome.PASS, 'no empty environment files'
+
     def wake(self) -> None:
         """ Wake up all steps """
 
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -12,7 +12,8 @@ The :ref:`/plugins/provision/beaker` provision plugin gains
 support for :ref:`system.model-name</spec/hardware/system>` and
 :ref:`system.vendor-name</spec/hardware/system>` hardware requirements.
 
-``tmt lint`` now reports a failure if empty environment files are found.
+The ``tmt lint`` command now reports a failure if empty
+environment files are found.
 
 
 tmt-1.38.0
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -12,6 +12,8 @@ The :ref:`/plugins/provision/beaker` provision plugin gains
 support for :ref:`system.model-name</spec/hardware/system>` and
 :ref:`system.vendor-name</spec/hardware/system>` hardware requirements.
 
+``tmt lint`` now reports a failure if empty environment files are found.
+
 
 tmt-1.38.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/lint/plan/data/empty_env_file.fmf
+++ b/tests/lint/plan/data/empty_env_file.fmf
@@ -1,0 +1,5 @@
+summary: Empty environment file plan
+environment-file:
+    - empty_env
+execute:
+    script: tmt --help
--- a/tests/lint/plan/data/empty_env_file.fmf
+++ b/tests/lint/plan/data/empty_env_file.fmf
@@ -1,5 +1,5 @@
 summary: Empty environment file plan
 environment-file:
-    - empty_env
+  - empty_env
 execute:
     script: tmt --help
--- a/tests/lint/plan/data/env
+++ b/tests/lint/plan/data/env
@@ -1,0 +1,1 @@
+test=test
--- a/tests/lint/plan/data/good.fmf
+++ b/tests/lint/plan/data/good.fmf
@@ -2,3 +2,5 @@ summary: Basic smoke test
 execute:
     script: tmt --help
 id: 9c73e336-983e-4349-9586-b63c20ea2b89
+environment-file:
+    - env
--- a/tests/lint/plan/test.sh
+++ b/tests/lint/plan/test.sh
@@ -74,7 +74,7 @@ rlJournalStart
         rlAssertGrep 'warn C000 key "wrong" not recognized by schema /schemas/prepare/feature' $rlRun_LOG
 
         rlRun -s "$tmt plan lint empty_env_file" 1
-        rlAssertGrep "fail P008 the environment file '$(pwd)/empty_env' is empty" $rlRun_LOG
+        rlAssertGrep "fail P008 the environment file '.*/tests/lint/plan/data/empty_env' is empty" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "P007: step phases require existing guests and roles"
--- a/tests/lint/plan/test.sh
+++ b/tests/lint/plan/test.sh
@@ -50,6 +50,7 @@ rlJournalStart
 
         rlRun -s "$tmt plan lint invalid_url" 1
         rlAssertGrep "fail P005 remote fmf id in \"default-0\" is invalid, repo 'http://invalid-url' cannot be cloned" $rlRun_LOG
+        rlAssertGrep "skip P008 no environment files found" $rlRun_LOG
 
         rlRun -s "$tmt plan lint invalid_ref" 1
         rlAssertGrep "fail P005 remote fmf id in \"default-0\" is invalid, git ref 'invalid-ref-123456' is invalid" $rlRun_LOG
@@ -71,6 +72,9 @@ rlJournalStart
         rlRun -s "$tmt plan lint invalid-plugin-key" 0
         rlAssertGrep 'warn C000 key "wrong" not recognized by schema$' $rlRun_LOG
         rlAssertGrep 'warn C000 key "wrong" not recognized by schema /schemas/prepare/feature' $rlRun_LOG
+
+        rlRun -s "$tmt plan lint empty_env_file" 1
+        rlAssertGrep "fail P008 the file \"$(pwd)/empty_env\" is empty" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "P007: step phases require existing guests and roles"
--- a/tests/lint/plan/test.sh
+++ b/tests/lint/plan/test.sh
@@ -74,7 +74,7 @@ rlJournalStart
         rlAssertGrep 'warn C000 key "wrong" not recognized by schema /schemas/prepare/feature' $rlRun_LOG
 
         rlRun -s "$tmt plan lint empty_env_file" 1
-        rlAssertGrep "fail P008 the file \"$(pwd)/empty_env\" is empty" $rlRun_LOG
+        rlAssertGrep "fail P008 the environment file '$(pwd)/empty_env' is empty" $rlRun_LOG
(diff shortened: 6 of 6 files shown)


tmt/base.py
```diff
@@ -2130,6 +2130,18 @@ def lint_unknown_keys(self) -> LinterReturn:
 
         yield LinterOutcome.PASS, 'correct keys are used'
 
+    def lint_empty_env_files(self) -> LinterReturn:
+        """ P001: all files are not empty """
+
+        env_file = self.plan_environment_file
+
+        if (self.my_run and env_file.exists() and env_file.stat().st_size == 0):
```
happz, line 2138: I'm not sure this would work: by design, `lint_*` callbacks are called not in runtime, `tmt run ...`, but e.g. before committing changes, via `tmt * lint`. Therefore `self.my_run` should always be unset when linters run.

I think there is also a misunderstanding in your way: `plan_environment_file` is a per-plan file, provided by tmt to plans and tests, with path exposed via `TMT_PLAN_ENVIRONMENT_FILE` envvar, and as such does exist only in the context of `tmt run`. Which makes it non-lintable entity.

If you wanted to check whether files referenced by plan's `environment-file`, https://tmt.readthedocs.io/en/stable/spec/plans.html#environment-file, you'd need to check `self.node`, `self.node.get("environment-file") or []` should give you the list of files referenced by the plan.

tmt/base.py
```diff
@@ -2130,6 +2130,22 @@ def lint_unknown_keys(self) -> LinterReturn:
 
         yield LinterOutcome.PASS, 'correct keys are used'
 
+    def lint_empty_env_files(self) -> LinterReturn:
+        """ P001: env files are not empty """
+
+        env_files = self.node.get("environment-file") or []
+        if len(env_files) > 0:
```
happz, line 2137: `if env_files:` is the Python way of checking list emptiness.

tmt/base.py
mcasquer: @happz thanks for the feedback I am still a newbie here, just a couple of comments. The idea of this patch is to cover the issue #2872 which mentions to report the fail in the `tmt lint` so that should be the idea right?, or maybe you want to rethink the intention of the issue. Regarding the plan environment file I changed the code according to your review, I hope now it's appropiate
cc @psss

tmt/base.py
```diff
@@ -2130,6 +2130,22 @@ def lint_unknown_keys(self) -> LinterReturn:
...
+            for env_file in env_files:
+                path = os.path.normpath(env_file)
+                size = os.path.getsize(path)
```
happz, line 2140: Use `pathlib` instead of `os.path`, please. It will eventually make the condition simpler as well:

```python
for _env_file in env_files:
  env_file = Path(env_file).resolve()

  if not env_file.exists() or not env_file.stat().st_size:
    ..
```

Catches both cases when the file does not exists - deserves its own message - and the file being empty.

tmt/base.py
happz: I think the idea matches the issue, you just focused on the wrong "environment file", that's all :)

tmt/base.py
mcasquer: Updated, thanks !

tmt/base.py
mcasquer: Updated, thanks !

tmt/base.py
```diff
@@ -2130,6 +2130,21 @@ def lint_unknown_keys(self) -> LinterReturn:
 
         yield LinterOutcome.PASS, 'correct keys are used'
 
+    def lint_empty_env_files(self) -> LinterReturn:
+        """ P001: env files are not empty """
```
psss, line 2134: This is adding a new check with the same `P001` identifier as above. Let's use the first free slot which seems to be `P008`.

tmt/base.py
```diff
@@ -2130,6 +2130,21 @@ def lint_unknown_keys(self) -> LinterReturn:
...
+
+        env_files = self.node.get("environment-file") or []
+        if env_files:
```
psss, line 2137: Is the condition needed? What about dropping it altogether? The `for` cycle would handle an empty `list` itself.

tmt/base.py
```diff
@@ -2130,6 +2130,21 @@ def lint_unknown_keys(self) -> LinterReturn:
...
+            for env_file in env_files:
+                env_file = Path(env_file).resolve()
+
+                if not env_file.exists() or not env_file.stat().st_size:
+                    yield LinterOutcome.FAIL, f'the file "{env_file}" does not exists or is empty'
```
psss, line 2142: The non-existent file is actually not covered:

```
tmt lint

File '/tmp/empty/environment' doesn't exist.
```

We do not get the full `lint` output as expected.

psss (changes requested):
Thanks for improving this! Added a couple comments. Would be good to handle the non-existent file as part of the `lint` report instead of bailing out with a catched exception.

Would you mind adding a simple test coverage? You could just extend the [existing test](https://github.com/teemtee/tmt/blob/main/tests/lint/plan/test.sh). Also, what about adding a short release note?

tmt/base.py
happz: It gives a better message: when the list is empty, you get an immediate `PASS`, while with a loop it would be hard to report such a pass as the loop would not indicate the number of iterations it took.

tmt/base.py
happz: Meh, incorrectly indented `return`, should be on the level of `for`.

tmt/base.py
psss: Not sure I understand. If the list is empty, the `for` loop would be skipped and you get right to the `pass` statement. But yes, if the `return` statement is misplaced, then it's a completely different story.

tmt/base.py
psss: I see, moving `return` makes sense. It does not fix the problem mentioned above though.

tmt/base.py
```diff
@@ -2130,6 +2130,21 @@ def lint_unknown_keys(self) -> LinterReturn:
...
+
+                    return
+
+        yield LinterOutcome.PASS, 'no empty environment files found'
```
psss, line 2146: This sounds rather like a `SKIP` doesn't it? If there are no environment files there's nothing to check. What about handling this before the loop?

```python
if not env_files:
    yield LinterOutcome.SKIP, 'no environment files found'
    return
```

tmt/base.py
happz: Yes, `return` is misplaced. I suppose it's structured as other lint checks, e.g.:

```python
    def lint_unknown_keys(self) -> LinterReturn:
        """ T001: all keys are known """

        # We don't want adjust in show/export so it is not yet in Test._keys
        invalid_keys = self._lint_keys(EXTRA_TEST_KEYS + OBSOLETED_TEST_KEYS + ['adjust'])

        if invalid_keys:
            for key in invalid_keys:
                yield LinterOutcome.FAIL, f'unknown key "{key}" is used'

            return

        yield LinterOutcome.PASS, 'correct keys are used'
```

Without the `if invalid_keys`, code would run the loop, finish the loop, and continue to `yield PASS`, which would be fine when `invalid_keys` is empty, but incorrect when it's not empty. Hence the test & hard `return` in that branch.

tmt/base.py
happz: Makes sense, it is more like a `SKIP`. And yep, I wrote current checks with `SKIP` being sorted out ASAP, e.g.:

```python
    def lint_manual_test_path_exists(self) -> LinterReturn:
        """ T007: manual test path is not an actual path """

        if not self.manual:
            yield LinterOutcome.SKIP, 'not a manual test'
            return

        if not self.path:
            yield LinterOutcome.FAIL, 'directory path is not set'
            return

        if not self.manual_test_path.exists():
            yield LinterOutcome.FAIL, f'manual test path "{self.manual_test_path}" does not exist'
            return

        yield LinterOutcome.PASS, f'manual test path "{self.manual_test_path}" does exist'
```

So it would be consistent with rest of the checks.

tmt/base.py

tmt/base.py

tmt/base.py
psss: Thanks!

tmt/base.py
psss: The `PASS` statement, as suggested above, is missing. Thus you would never get ok:

```
> tmt lint
/plans/example
pass C001 summary key is set and is reasonably long
skip P004 discover step is not defined
```

tmt/base.py
psss: The non-existent case is still not fixed:

```
> tmt lint

File '/tmp/empty/environment' doesn't exist.
```

tmt/base.py
```diff
@@ -2305,6 +2305,23 @@ def _lint_step(step: str) -> LinterReturn:
         yield from _lint_step('execute')
         yield from _lint_step('finish')
 
+    def lint_empty_env_files(self) -> LinterReturn:
+        """ P008: env files are not empty """
+
+        env_files = self.node.get("environment-file") or []
+
+        if not env_files:
+            yield LinterOutcome.SKIP, 'no empty environment files found'
```
psss, line 2314:
```suggestion
            yield LinterOutcome.SKIP, 'no environment files found'
```

Perhaps like this?

psss:
> Would you mind adding a simple test coverage? You could just extend the [existing test](https://github.com/teemtee/tmt/blob/main/tests/lint/plan/test.sh). Also, what about adding a short release note?

@mcasquer, what do you think about adding a simple test and a short release note?

mcasquer:
> > Would you mind adding a simple test coverage? You could just extend the [existing test](https://github.com/teemtee/tmt/blob/main/tests/lint/plan/test.sh). Also, what about adding a short release note?
> 
> @mcasquer, what do you think about adding a simple test and a short release note?

Perfect, I will add them along with the above suggestions !

tmt/base.py

tmt/base.py

psss (approved):
Now looks good, thanks! Added just a few minor adjustments in 2ec03046.

psss:
Test `/tests/lint/plan/explicit-root` is failing. Seems that your newly added test coverage has revealing a bug. Congratulations! :) Here are simple steps to reproduce:

```
> cd tests/lint/plan/data
> tmt plan lint good
/good
pass C001 summary key is set and is reasonably long
skip P004 discover step is not defined
pass P008 no empty environment files

Lint checks on all

> cd ..
> tmt --root data plan lint good

[Errno 2] No such file or directory: '/home/psss/git/tmt/tests/lint/plan/env'
```

Both commands should give the same result as the `--root` option is instructing `tmt` to use provided directory as the `fmf` metadata tree root. Seems that path to the `env` file is not correctly constructed when `--root` is used. @mcasquer, could you please have a look?

mcasquer:
> Both commands should give the same result as the `--root` option is instructing `tmt` to use provided directory as the `fmf` metadata tree root. Seems that path to the `env` file is not correctly constructed when `--root` is used. @mcasquer, could you please have a look?

@psss Ok I will take a look :)

mcasquer:
> Both commands should give the same result as the `--root` option is instructing `tmt` to use provided directory as the `fmf` metadata tree root. Seems that path to the `env` file is not correctly constructed when `--root` is used. @mcasquer, could you please have a look?

@psss Is this the only code that implements this `--root` feature?
https://github.com/teemtee/tmt/blob/b058850af4947c4dc2152a428d5f04a66dc1cfaa/tmt/cli.py#L290-L292

happz:
@mcasquer correct. Once an fmf tree is loaded, objects derived from it - tests, plans, stories - should have `fmf_root` attribute storing the path. And I'd say that's what's missing: `env_file = Path(env_file).resolve()` should be `env_file = (self.fmf_root / Path(env_file)).resolve()` - or, better, once https://github.com/teemtee/tmt/pull/3212 gets merged, `env_file = (self.anchor_path / Path(env_file)).resolve()`.

mcasquer:
> @mcasquer correct. Once an fmf tree is loaded, objects derived from it - tests, plans, stories - should have `fmf_root` attribute storing the path. And I'd say that's what's missing: `env_file = Path(env_file).resolve()` should be `env_file = (self.fmf_root / Path(env_file)).resolve()` - or, better, once #3212 gets merged, `env_file = (self.anchor_path / Path(env_file)).resolve()`.

@happz that's indeed the point, thanks a bunch! I will change it then to the final version as I see at this moment the #3212 is planned for 1.38 milestone too

mcasquer:
@psss @happz everything should be fine now :grinning:

psss:
> @psss @happz everything should be fine now 😀

Great thanks! In 10177f3b I've just added back the minor adjustments which were lost in the rebase.

mcasquer:
@psss it seems the data directory is not included in the path when doing the lint
```bash
Test result
fail P008 the environment file '/var/ARTIFACTS/work-coreu1ik6621/plans/features/core/discover/default-0/tests/tests/lint/plan/data/empty_env' is empty`
```
vs what the test expects
```bash
:: [ 10:33:43 ] :: [   FAIL   ] :: File '/var/tmp/rlRun_LOG.rrmwXJFk' should contain 'fail P008 the environment file '/var/ARTIFACTS/work-coreu1ik6621/plans/features/core/discover/default-0/tests/tests/lint/plan/empty_env' is empty' 
```

psss:
Yeah, the problem was in the fact, that the test is run twice, each time in a different directory:

https://github.com/teemtee/tmt/blob/7c08e96170470eb6e3cb99fbf8b705862a1e0a58/tests/lint/plan/test.sh#L6-L11

So we cannot use `$(pwd)` there. I think we're good with a basic check. What about fba14f24?

mcasquer:
> So we cannot use `$(pwd)` there. I think we're good with a basic check. What about [fba14f2](https://github.com/teemtee/tmt/commit/fba14f24b24bac888ff22aa367c76268f3bcb00c)?

Ok I see, looks fine for me now,  thanks ! :)

psss:
Everything green, the `epel-9` job actually finished, looks just like some packit-github sync issue, merging.

Issue #3130, opened 2024-08-05
Please print from where tmt is loading beakerlib libraries


yarda:
It's very confusing, because tmt uses some magic github URL to checkout beakerlib library, e.g. the following in the fmf file:
```
require:
  - library(tuned/basic)
```
causes tmt to silently checkout repository from the https://github.com/beakerlib/tuned/tree/master/basic. I think it should log the URL it's checking out in the default log verbosity (I was able to get this information only be enabling verbose debug logging).

psss:
Should be easy to implement. What about including the library repo in the verbose output? Perhaps level 2? Or maybe even 3? I don't think it would be a good idea to show the information by default as there can be many libraries identified in the discovered tests.

The "magic github url" is documented in the [require](https://tmt.readthedocs.io/en/stable/spec/tests.html#require) attribute.

yarda:
How much libraries are usually used simultaneously? Up to five? Just guessing. In such case it's 5 more lines, e.g.:
```
Including tuned/basic (https://github.com/beakerlib/tuned/tree/master/basic)
...
```
Or whatever source/URL is used to get the library. It can be even more compressed output.

Taking into account tmt is currently extremely verbose, e.g.:
```
$ tmt try rhel-9@minute
/var/tmp/tmt/run-032
Let's try /Sanity/RFE-specific-tuning-according-to-proc-cpuinfo with /default/plan on rhel-9@minute.

/default/plan
    discover
        how: fmf
        order: 50
        directory: /home/yarda/git-rhel/tests/tuned
        hash: e9853a5
        summary: 1 test selected
            /Sanity/RFE-specific-tuning-according-to-proc-cpuinfo
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: minute
        order: 50
        image: rhel-9
        flavor: m1.small
        auto-select-network: false
        skip-prereserve: false
        1mt-image: 1MT-RHEL-9.5.0-20240805.2
            cmd: whoami
            err: ssh: connect to host 10.0.186.176 port 22: Connection refused
            cmd: whoami
            err: ssh: connect to host 10.0.186.176 port 22: Connection refused
            cmd: whoami
            err: Warning: Permanently added '10.0.186.176' (ED25519) to the list of known hosts.
            out: root
        multihost name: default-0
        arch: x86_64
        distro: Red Hat Enterprise Linux 9.5 Beta (Plow)
        kernel: 5.14.0-490.el9.x86_64
[...]

PR #3208, opened 2024-09-12
Log the information about used beakerlib libraries

seberm:
Print the "calculated" `url`, `ref` and `path` of the used beakerlib libraries which will be cloned by tmt. Print the information in verbose mode of `level=3`.

Fixes:
- https://github.com/teemtee/tmt/issues/3130

```
$ tmt -r ~/tmt-learning/03-beakertest run --until discover -vvv report --how display
...
/plans/example
summary: Basic smoke test
    discover
        how: fmf
        order: 50
        directory: /home/user/tmt-learning/03-beakertest
        using remote git library:
            url: https://github.com/beakerlib/tuned
            ref: master
            path: /basic
        using remote git library:
            url: https://github.com/beakerlib/openssl
            ref: devel
            path: /certgen
        summary: 1 test selected
            /tests/example
    report
        how: display
        order: 50
        summary: no results found
...
```

CC: @psss , @yarda

Pull Request Checklist

* [x] implement the feature
* [x] include a release note


tmt/libraries/beakerlib.py
```diff
@@ -330,6 +330,13 @@ def fetch(self) -> None:
                     self.parent.debug(f"Library {self} is copied into {directory}")
                     shutil.copytree(library_path, local_library_path, dirs_exist_ok=True)
 
+                    self.parent.verbose(
+                        'library',
+                        f"using remote repository: url='{self.url}', ref='{self.ref}'"
+                        f", path='{self.name}'",
```
happz, line 336: `'using remote repository', {'url': self.url, 'ref': self.ref, 'path': self.path}` might be better formatted, given it's a structured information.

seberm: Yeah, that's better :), thanks.

Btw I just want to check this part: `'path': self.path`. I think we want to use the `'path': self.name` in this case. The `self.path` is empty.

```
/plans/example
summary: Basic smoke test
    discover
        how: fmf
        order: 50
        directory: /home/user/tmt-learning/03-beakertest
        using remote git library:
            url: https://github.com/beakerlib/tuned
            ref: master
            path: /basic
        using remote git library:
            url: https://github.com/beakerlib/openssl
            ref: devel
            path: /certgen
        summary: 1 test selected
            /tests/example
    report
        how: display
        order: 50
        summary: no results found
```

happz: Hm, maybe we want both `path` and `name`? It feels like an fmf ID, maybe it's already there somewhere or it could be constructed (https://github.com/teemtee/tmt/blob/main/tmt/base.py#L140) and its `to_minimal_spec()` logged, instead of composing it and checking for empty fields "manually".

seberm: We can use the `self.identifier` which is just a `DependencySimple(str)` in the case of a "backward-compatible syntax" library name and the fields like `self.name`, `self.path`, etc. are parsed from it.

If I use a [full fmf identifier beaker library notation](https://tmt.readthedocs.io/en/stable/spec/tests.html#require), the type of `self.identifier` is a `DependencyFmfId` and it works as expected.

The output with your suggested changes is the following:

```
/plans/example
summary: Basic smoke test
    discover
        how: fmf
        order: 50
        directory: /home/user/tmt-learning/03-beakertest
        using remote git library: library(tuned/basic)
        using remote git library:
            url: https://github.com/beakerlib/openssl
            ref: devel
            name: /certgen
            type: library
        summary: 1 test selected
            /tests/example
```

I think we unfortunately must compose the fields manually, at least for "backward-compatible" library syntax.

seberm: I agree, let's use the `name` as a name and `path` as a path :).

We can add a condition to print all the non-empty fields from `self.identifier` fmf id like the following, but I still think it would probably be better just to construct the dict "manually":

```
                    # By default, handle the backward-compatible library syntax
                    library_fields = {
                        'url': self.url,
                        'ref': self.ref,
                        'name': self.name}

                    if isinstance(self.identifier, DependencyFmfId):
                        library_fields = cast(dict[str, Optional[str]], self.identifier.to_minimal_dict())

                    self.parent.verbose('using remote git library', library_fields, 'green',
                                        level=3)

```

You choose.

happz: I can't shake off the feeling we're instantiating an fmf id :) It has the same fields, same meanings, how about this?

```python
fake_library_id = self.identifier if isinstance(self.identifier, DependencyFmfId) \
    else FmfId(url=self.url, ref=self.ref, path=self.path, name=self.name)

self.parent.verbose(
    'using remote git library',
    fake_library_id.to_minimal_dict(),
    'green',
    level=3)
```

therazix (approved):

psss (approved):
Thanks for implementing this! Looks good. Added a simple test coverage in 3a76959c and a short mention to release notes in bfe8a77b.

seberm:
> Thanks for implementing this! Looks good. Added a simple test coverage in 3a76959c and a short mention to release notes in bfe8a77b.

Great, thanks!

PR #3209, opened 2024-09-12
Reportportal should note guest hostname for investigation

happz:
It may happen the guest name or role are not set, yet there are multiple guests reported into the same launch, when reporting multiple tmt runs into the same RP launch, e.g. when those runs run the same tests but on different hosts. This is a common case for those running tests on multiple architectures, e.g. when testing a Koji build.

Pull Request Checklist

* [x] implement the feature
* [ ] include a release note


happz:
@kkaarreell @4N0body5 what do you think about this addition?

KwisatzHaderach (approved):

kkaarreell:
> @kkaarreell @4N0body5 what do you think about this addition?

Ack

tmt/steps/execute/__init__.py
```diff
@@ -751,7 +751,7 @@ def _process_results_partials(
         :returns: list of results.
         """
 
-        test, guest = invocation.test, invocation.guest
+        test, _ = invocation.test, invocation.guest
```
therazix, line 754: Is there a reason to keep `invocation.guest` here?
```suggestion
        test = invocation.test
```

happz: Addressed in 0985617b

therazix (approved):

happz:
Unrelated failures, merging.

PR #3210, opened 2024-09-12
Forbid use of os.path.join, point user to tmt._compat.Path

happz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/discover/data/distgit_test_plugin.py
+++ b/tests/discover/data/distgit_test_plugin.py
@@ -1,6 +1,5 @@
-import os
 
-from tmt.utils import DistGitHandler
+from tmt.utils import DistGitHandler, Path
 
 MOCK_SOURCES_FILENAME = 'mock_sources'
 SERVER_PORT = 9000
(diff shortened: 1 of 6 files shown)


tests/discover/data/distgit_test_plugin.py
```diff
@@ -17,14 +16,14 @@ class TestDistGit(DistGitHandler):
     usage_name = "TESTING"
     server = f"http://localhost:{SERVER_PORT}"
 
-    def url_and_name(self, cwd='.'):
-        with open(os.path.join(cwd, MOCK_SOURCES_FILENAME)) as f:
-            data = f.read()
+    def url_and_name(self, cwd=None):
```
martinhoyer, line 19: Was there ever a consensus about the use of `https://docs.python.org/3/library/os.html#os.PathLike` or something to annotate 'pathlike'?

martinhoyer (approved):
Lovely

tests/discover/data/distgit_test_plugin.py
happz: Nope, never discussed there, we use `tmt.utils.Path` everywhere.

therazix (approved):

happz:
Unrelated failure, merging.

PR #3212, opened 2024-09-12
Use the current directory as `anchor_path` if there is no `fmf_root`

falconizmi:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -775,7 +775,7 @@ class Core(
     # TODO: cached_property candidates
     @property
     def fmf_root(self) -> Optional[Path]:
-        # Check if fmf root exists
+        # Check if fmf root is defined
         if self.node.root is not None:
             return Path(self.node.root)
         return None
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -773,7 +773,12 @@ class Core(
     # TODO: cached_property candidates
     @property
     def fmf_root(self) -> Path:
-        return Path(self.node.root)
+        # Check if fmf root exists
+        try:
+            return Path(self.node.root)
+        except TypeError:
+            raise tmt.utils.GeneralError(
+                "No fmf root found. Directory is not initialized with `tmt init`")
 
     @property
     def git_root(self) -> Optional[Path]:
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -147,8 +147,7 @@ class FmfId(
     VALID_KEYS: ClassVar[list[str]] = ['url', 'ref', 'path', 'name']
 
     #: Keys that are present, might be set, but shall not be exported.
-    NONEXPORTABLE_KEYS: ClassVar[list[str]] = [
-        'fmf_root', 'git_root', 'default_branch']
+    NONEXPORTABLE_KEYS: ClassVar[list[str]] = ['fmf_root', 'git_root', 'default_branch']
 
     # Save context of the ID for later - there are places where it matters,
     # e.g. to not display `ref` under some conditions.
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -776,10 +776,9 @@ class Core(
     @property
     def fmf_root(self) -> Optional[Path]:
         # Check if fmf root exists
-        try:
+        if self.node.root is not None:
             return Path(self.node.root)
-        except TypeError:
-            return None
+        return None
 
     @property
     def anchor_path(self) -> Path:
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -148,12 +148,11 @@ class FmfId(
 
     #: Keys that are present, might be set, but shall not be exported.
     NONEXPORTABLE_KEYS: ClassVar[list[str]] = [
-        'fmf_root', 'anchor_path', 'git_root', 'default_branch']
+        'fmf_root', 'git_root', 'default_branch']
 
     # Save context of the ID for later - there are places where it matters,
     # e.g. to not display `ref` under some conditions.
     fmf_root: Optional[Path] = None
-    anchor_path: Optional[Path] = None
     git_root: Optional[Path] = None
     default_branch: Optional[str] = None
 
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -147,11 +147,13 @@ class FmfId(
     VALID_KEYS: ClassVar[list[str]] = ['url', 'ref', 'path', 'name']
 
     #: Keys that are present, might be set, but shall not be exported.
-    NONEXPORTABLE_KEYS: ClassVar[list[str]] = ['fmf_root', 'git_root', 'default_branch']
+    NONEXPORTABLE_KEYS: ClassVar[list[str]] = [
+        'fmf_root', 'anchor_path', 'git_root', 'default_branch']
 
     # Save context of the ID for later - there are places where it matters,
     # e.g. to not display `ref` under some conditions.
     fmf_root: Optional[Path] = None
+    anchor_path: Optional[Path] = None
     git_root: Optional[Path] = None
     default_branch: Optional[str] = None
 
@@ -772,17 +774,20 @@ class Core(
 
     # TODO: cached_property candidates
     @property
-    def fmf_root(self) -> Path:
+    def fmf_root(self) -> Optional[Path]:
         # Check if fmf root exists
         try:
             return Path(self.node.root)
         except TypeError:
-            raise tmt.utils.GeneralError(
-                "No fmf root found. Directory is not initialized with `tmt init`")
+            return None
+
+    @property
+    def anchor_path(self) -> Path:
+        return self.fmf_root or Path.cwd()
 
     @property
     def git_root(self) -> Optional[Path]:
-        return tmt.utils.git.git_root(fmf_root=self.fmf_root, logger=self._logger)
+        return tmt.utils.git.git_root(fmf_root=self.anchor_path, logger=self._logger)
 
     # Caching properties does not play nicely with mypy and annotations,
     # and constructing a workaround would be hard because of support of
@@ -800,7 +805,7 @@ class Core(
 
         return tmt.utils.fmf_id(
             name=self.name,
-            fmf_root=self.fmf_root,
+            fmf_root=self.anchor_path,
             logger=self._logger)
 
     @functools.cached_property
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -167,16 +167,9 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
 
                 logger.info('playbook-path', playbook_path, 'green')
 
-            # Check if fmf root exists
-            try:
-                playbook_root = self.step.plan.fmf_root
-            except TypeError:
-                raise PrepareError(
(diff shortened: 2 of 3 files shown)


tmt/steps/prepare/ansible.py
```diff
@@ -166,9 +166,16 @@ def go(
 
                 logger.info('playbook-path', playbook_path, 'green')
 
+            # Check if fmf root exists
+            try:
+                playbook_root = self.step.plan.fmf_root
+            except TypeError:
+                raise PrepareError(
+                    "No fmf root found. Directory is not initialized with `tmt init`")
```
happz, line 174: Seems needlessly complicated :)

```python
if not self.step.plan.fmf_root:
    raise PrepareError("No fmf root found. Directory is not initialized with `tmt init`")
```

falconizmi: I already tried to do it like that but its not working, because fmf_root is not value but a getter, its failing before it can go to inside the if statement.

happz: I see. But that's very interesting information: it fails because the getter, declared as `-> Path`, runs `return Path(self.node.root)`, and we realized an fmf node may not always be present. So what you do is scratching the surface, your change will turn "give me fmf root" into an exception in one place only, while everything else trying to access fmf root will be left unsanitized. We should go deeper :)

So, it feels like this check & exception should be raised by `fmf_root` itself, to prevent anyone from accessing `fmf_root` without a punishment. I'm not sure how exactly the exception raised *now* by getter looks like, could you share it? Or feel free to design the test right away.

BTW there's enough blame for more pieces of code, e.g. all these classes, `Core`, `Test`, `Plan`, `Story` - they take fmf node as an input parameter: how can they take an fmf node when we agreed that without an fmf node, there is no fmf tree and therefore no fmf nodes representing plans and tests?? How can we ever get to run a playbook, as directed by a `prepare` step of a plan when there is no fmf root and therefore no plans to recognize?? So there is more to review, basically, all constructs based on fmf are now suspicious when they can get constructed without an fmf root and get as far as to run a playbook :) I would start by checking whether we could put the check & exception into `fmf_root` for now, but eventually, we might check how can we even have `Plan` instance running playbooks when we have no fmf root to start with, and push this check to trigger way sooner.

falconizmi: Is the new change enough for now? I tested it and it returns me the exception message
`~/.local/bin/tmt run -vvv prepare --how ansible --playbook packages.yml provision --how virtual` (in directory without tmt init)

lukaszachy: > , we might check how can we even have Plan instance running playbooks when we have no fmf root to start with, and push this check to trigger way sooner.

But we (technically) do allow to specify the plan via CLI, right? IMO in that case fmf root doesn't need to exist.

happz: I'm not sure I follow, how do we specify plan via CLI? `plan --name ...` comes to mind, but that plan must exist first in fmf, and that's all I have.

psss: I think, a little test would help a lot. With reproducing, with general understanding. I guess the `extend the test coverage` item shouldn't have been removed from the checklist :grin:

So, I believe, @lukaszachy is right: We can still do quick action from the command line with the default plan. The example provided by @falconizmi reproduces. My quick check confirmed:

```
cd
tmt run provision -h container prepare -h ansible --playbook something.yaml login finish
...
fail: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'
```

So yes, it's true that we are able to run outside of an `fmf` root because when no `fmf` metadata is detected the default plan is used instead.

https://github.com/teemtee/tmt/blob/fb19fb0863104fa0d79499fc57929cbf675c5fed/tmt/base.py#L3361-L3363

psss: Hm, the `--playbook` options states:

```
Path or URL of an ansible playbook to run.
The playbook path must be relative to the metadata tree root.
```

We need to decide what to do when the default plan is used. I can see a use case for quickly setting up a guest for experimenting when no `fmf` root is around. So... what about something like this:

```
When no `fmf` root is around, the path should be relative to the current working directory.
```

happz: Then we also need to change the type of `self.fmf_root` to be `Optional`, and fix the fallout. The assumption that fmf root always exists is widely spread. Maybe we should have something like `self.anchor_path` that would provide this particular path (`self.fmf_root or Path.cwd()`). Many places may need to perform this “is it there or not?” dance.

psss: > Maybe we should have something like `self.anchor_path` that would provide this particular path

Yeah, that sounds good.

falconizmi: If I understand correctly, instead of having property `fmt_root`  and its getter in base.py, we will have `anchor_path` property and its getter (that would try to return `fmf_root` and if it fails, it would return `Path.cwd()` ) and I would change only those which used `fmf_root` getter from `base.py` and the rest of `fmf_root` usage we would keep them as they are?

happz: Yes, that's what I would do. If objects like `Plan` can exist without an fmf root, then it shouldn't be a sin to access `Plan.fmf_root`, instead an "undefined" value like `None` should be returned.

tmt/steps/prepare/ansible.py
```diff
@@ -70,8 +70,11 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
 
     .. warning::
 
-        When specifying playbooks with paths, all paths must be
-        relative to the metadata tree root.
+        When specifying playbooks with paths:
+        if metadata tree root exist:
+            all paths must be relative to the metadata tree root.
+        if metadata tree root NOT exist:
+            all paths must be relative to the current working directory.
```
falconizmi, line 77: Is it readable and well structered enough?

falconizmi: Hm, and for some reason, after I commited the pre-commit didnt start the tmt lint check
`tmt lint.............................................(no files to check)Skipped`

tmt/base.py
```diff
@@ -771,12 +773,20 @@ def _fmf_id(self) -> None:
...
     # TODO: cached_property candidates
     @property
-    def fmf_root(self) -> Path:
-        return Path(self.node.root)
+    def fmf_root(self) -> Optional[Path]:
+        # Check if fmf root exists
+        try:
+            return Path(self.node.root)
+        except TypeError:
+            return None
```
happz, line 781: I'd recommend checking whether `self.node.root` is `None` or not. It's a simple check, more readable than `try/except`.

tmt/steps/prepare/ansible.py
happz: Not bad at all, although there are small changes I'd make: some a bit of rewording, and we can use ReST formatting for better rendering (try it out with `tmt run prepare -h ansible --help` and/or `make docs`, `docs/_build/plugins/prepare.html#ansible`)

```rest
        When specifying playbooks with paths:

        * If a metadata tree root exists, all paths must be relative to the metadata tree root.
        * If the metadata tree root does not exist, all paths must be relative to the current working directory.
```

therazix (approved):

tmt/steps/prepare/ansible.py
```diff
@@ -70,8 +70,11 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
 
     .. warning::
 
-        When specifying playbooks with paths, all paths must be
-        relative to the metadata tree root.
+        When specifying playbooks with paths:
+
+        * If a metadata tree root exists, all paths must be relative to the metadata tree root.
+        * If the metadata tree root does not exist, \
+all paths must be relative to the current working directory.
```
happz, line 77: ReST supports breaking lines, add new-line & align the new line with the beginning of the line above:

```
        * If the metadata tree root does not exist,
          all paths must be relative to the current working directory.
```

tmt/base.py
```diff
@@ -772,12 +774,19 @@ def _fmf_id(self) -> None:
 
     # TODO: cached_property candidates
     @property
-    def fmf_root(self) -> Path:
-        return Path(self.node.root)
+    def fmf_root(self) -> Optional[Path]:
+        # Check if fmf root exists
```
happz, line 778: Probably "if fmf root is defined", because we don't really check whether it exists, i.e. `Path(self.node.root).exists()`.

tmt/base.py
```diff
@@ -147,11 +147,13 @@ class FmfId(
...
 
     #: Keys that are present, might be set, but shall not be exported.
-    NONEXPORTABLE_KEYS: ClassVar[list[str]] = ['fmf_root', 'git_root', 'default_branch']
+    NONEXPORTABLE_KEYS: ClassVar[list[str]] = [
+        'fmf_root', 'anchor_path', 'git_root', 'default_branch']
 
     # Save context of the ID for later - there are places where it matters,
     # e.g. to not display `ref` under some conditions.
     fmf_root: Optional[Path] = None
+    anchor_path: Optional[Path] = None
```
happz, line 156: @falconizmi are these two changes needed? Why does `anchor_path` attribute sneak into an fmf ID?

tmt/steps/prepare/ansible.py
psss: Seems to be covered.

tmt/base.py
falconizmi: I mistook it as necessary thing for anchor_path getter. They are removed now.

psss: Seems to be solved.

tmt/utils/git.py
```diff
@@ -448,7 +448,7 @@ def validate_git_status(test: 'tmt.base.Test') -> tuple[bool, str]:
     """
     sources = [
         *test.fmf_sources,
-        test.fmf_root / '.fmf' / 'version'
+        test.anchor_path / '.fmf' / 'version'
```
psss, line 451: I think, here we should leave `fmf_root`. The `validate_git_status` function is used in `tmt teste export --how nitrate` for checking the test metadata are synced with the remote repo, so there definitely should be an `fmf` root where the test belongs.

psss (changes requested):
Thanks for working on this. Looks good, just one comment. I've also pushed some minor changes in ba39b471.

tmt/utils/git.py
happz: What if `fmf_root` is `None`? Which, as we know, would happen when running tmt outside of an fmf root. I suppose `test.fmf_root` should be tested, and if it's set, then added to the list. Its automatic presence is incorrect.

psss: Here, for this use case, it just cannot be `None`. Tests which are exported need to be stored on disk and have their `fmf` root defined (this is not like default plan or test metadata created under the `shell` discover step). So, if we want to double check it here, we can raise an exception at the beginning of the function.

psss: What about 6f41bc72? Does it look ok, @happz?

happz: Yep, or that. My point was, it can be `None` and that value must be handled somehow - at this moment, it was blindly added to the list, which we realized is not the right choice. A test & an exception are perfectly valid option.

psss (approved):

PR #3213, opened 2024-09-12
Forbid use of `os.path`, point user to `tmt._compat.Path`

happz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/unit/test_schemas.py
+++ b/tests/unit/test_schemas.py
@@ -85,8 +85,8 @@ Detailed validation error:
         pytest.fail(f"{label} {name} fails validation")
 
 
-def _tree_path(tree):
-    return os.path.relpath(os.path.abspath(tree._path))
+def _tree_path(tree: tmt.Tree):
+    return tree._path.absolute().relative_to(Path.cwd())
 
 
 def extract_testcase_id(arg):
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -1606,9 +1606,12 @@ def expand_node_data(data: T, fmf_context: FmfContext) -> T:
         # don't expand it at all, let's put the original value back
         # into the stream.
         # See https://github.com/teemtee/tmt/issues/2654.
+        #
+        # TID251: `pathlib` does not provide `os.patch.expandvars`, its
+        # use is allowed.
         expanded_env: list[tuple[bool, str, str]] = [
             (False, item[1:], item) if item.startswith('@')
-            else (True, os.path.expandvars(f'${item}'), item)
+            else (True, os.path.expandvars(f'${item}'), item)  # noqa: TID251
             for item in split_data]
 
         # Expand unexpanded items, this time with fmf context providing
@@ -1621,8 +1624,9 @@ def expand_node_data(data: T, fmf_context: FmfContext) -> T:
                 if was_expanded:
                     expanded_ctx.append(item)
                     continue
-
-                expanded = os.path.expandvars(f'${item}')
+                # TID251: `pathlib` does not provide `os.patch.expandvars`, its
+                # use is allowed.
+                expanded = os.path.expandvars(f'${item}')  # noqa: TID251
                 expanded_ctx.append(f'${original_item}' if expanded.startswith('$') else expanded)
 
         # This cast is tricky: we get a string, and we return a
@@ -2006,15 +2010,15 @@ class Plan(
                 'link': links.to_spec()
                 })
 
-        for name in names:
-            (directory, plan) = os.path.split(name)
-            directory_path = path / directory.lstrip('/')
-            has_fmf_ext = os.path.splitext(plan)[1] == '.fmf'
-            plan_path = directory_path / (plan + ('' if has_fmf_ext else '.fmf'))
+        for plan_name in names:
+            plan_path = Path(plan_name)
+
+            if plan_path.suffix != '.fmf':
+                plan_path = plan_path.parent / f'{plan_path.name}.fmf'
 
             # Create directory & plan
             tmt.utils.create_directory(
-                path=directory_path,
+                path=plan_path.parent,
                 name='plan directory',
                 dry=dry,
                 logger=logger)
@@ -2031,7 +2035,7 @@ class Plan(
                 plans = Tree(
                     path=path,
                     logger=logger).plans(
-                    names=[f"^{name}$"],
+                    names=[f"^{plan_name}$"],
                     apply_command_line=False)
                 tmt.utils.jira.link(tmt_objects=plans, links=links, logger=logger)
 
@@ -2727,22 +2731,21 @@ class Story(
                 'link': links.to_spec()
                 })
 
-        for name in names:
-            # Prepare paths
-            (directory, story) = os.path.split(name)
-            directory_path = path / directory.lstrip('/')
-            has_fmf_ext = os.path.splitext(story)[1] == '.fmf'
-            story_path = directory_path / (story + ('' if has_fmf_ext else '.fmf'))
+        for story_name in names:
+            story_path = Path(story_name)
+
+            if story_path.suffix != '.fmf':
+                story_path = story_path.parent / f'{story_path.name}.fmf'
 
             # Create directory & story
             tmt.utils.create_directory(
-                path=directory_path,
+                path=path / story_path.parent,
                 name='story directory',
                 dry=dry,
                 logger=logger)
 
             tmt.utils.create_file(
-                path=story_path,
+                path=path / story_path,
                 name='story',
                 content=story_content,
                 dry=dry,
@@ -2753,7 +2756,7 @@ class Story(
                 stories = Tree(
                     path=path,
                     logger=logger).stories(
-                    names=[f"^{name}$"],
+                    names=[f"^{story_name}$"],
                     apply_command_line=False)
                 tmt.utils.jira.link(tmt_objects=stories, links=links, logger=logger)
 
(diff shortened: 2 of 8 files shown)


tmt/base.py
```diff
@@ -1991,15 +1995,15 @@ def create(
...
+        for plan_name in names:
+            plan_path = Path(plan_name)
+
+            if plan_path.suffix != '.fmf':
+                plan_path = plan_path.parent / f'{plan_path.name}.fmf'
 
             # Create directory & plan
             tmt.utils.create_directory(
-                path=directory_path,
+                path=plan_path.parent,
```
falconizmi, line 2006: Maybe here is missing `path/` as you did in story

happz: Addressed in b70de6ef

therazix (approved):

tmt/utils/__init__.py
```diff
@@ -576,7 +576,7 @@ def from_file(
             content = environment_filepath.read_text()
 
         # Parse yaml file
-        if os.path.splitext(filename)[1].lower() in ('.yaml', '.yml'):
+        if filename.rsplit('.')[-1].lower() in ('yaml', 'yml'):
```
psss, line 579: What about using `Path(filename).suffix` here?

happz: Sure, fixed in c23a9b97

tmt/base.py
```diff
@@ -2006,15 +2010,15 @@ def create(
...
-        for name in names:
-            (directory, plan) = os.path.split(name)
-            directory_path = path / directory.lstrip('/')
-            has_fmf_ext = os.path.splitext(plan)[1] == '.fmf'
-            plan_path = directory_path / (plan + ('' if has_fmf_ext else '.fmf'))
+        for plan_name in names:
+            plan_path = path / Path(plan_name).unrooted()
+
+            if plan_path.suffix != '.fmf':
+                plan_path = plan_path.parent / f'{plan_path.name}.fmf'
```
martinhoyer, line 2017: could use `with_suffix('.fmf') instead of f-string. Not sure if it has any benefit, just saying :)

tmt/base.py
```diff
@@ -4007,7 +4010,7 @@ def runs(self, id_: tuple[str, ...]) -> bool:
         if keep is not None:
             # Sort by modify time of the workdirs and keep the newest workdirs
             all_workdirs.sort(
-                key=lambda workdir: os.path.getmtime(workdir / 'run.yaml'), reverse=True)
+                key=lambda workdir: (workdir / 'run.yaml').stat().st_mtime, reverse=True)
```
martinhoyer, line 4013: same as above: stat() vs lstat(), assuming we don't need to handle symlinks?

tmt/export/nitrate.py
```diff
@@ -2,6 +2,7 @@
 import os
 import re
 import types
+import urllib.parse
```
martinhoyer, line 5: can be just `from urllib.parse import urlparse`?

tmt/steps/discover/fmf.py
```diff
@@ -544,8 +544,9 @@ def do_the_discovery(self, path: Optional[Path] = None) -> None:
                     'git', 'log', '--format=', '--stat', '--name-only', f"{modified_ref}..HEAD"
                     ), cwd=self.testdir)
             if output.stdout:
-                directories = [os.path.dirname(name) for name in output.stdout.split('\n')]
-                modified = {f"^/{re.escape(name)}" for name in directories if name}
+                directories = [Path(name).parent for name in output.stdout.split('\n')]
```
martinhoyer, line 547: `.splitlines()`, while we're at it? :)

martinhoyer (approved):
Just nitpicks and notes.  

> TID251: `pathlib` does not provide `os.patch.expandvars`, its use is allowed

I'll open an issue to Ruff, if it make sense to have this check more configurable, wdyt?

tests/unit/test_schemas.py
```diff
@@ -85,8 +85,8 @@ def validate_node(tree, node, schema, label, name):
         pytest.fail(f"{label} {name} fails validation")
 
 
-def _tree_path(tree):
-    return os.path.relpath(os.path.abspath(tree._path))
+def _tree_path(tree: tmt.Tree):
+    return tree._path.absolute().relative_to(Path.cwd())
```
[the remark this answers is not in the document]
happz: I bet it can be simplified. Symlinks probably are not the biggest thread, path that's relative to some other directory when CWD is assumed by the code might be a problem, hence this "normalization".

tmt/base.py
happz: `plan_path = (plan_path.parent / plan_path.name).with_suffix('.fmf')` certainly looks less readable to me :/

tmt/base.py
happz: `stat()` seems correct to me, we do want to follow symlinks - getting mtime of a symlink itself does not look useful.

tmt/export/nitrate.py
happz: It can, but I'd first agree on some policy. I for one am not a big friend of `import foo from bar` as I think it may easily pollute the namespace with dozens of names that have no place on the module level, but it does make sense for some common, well-known and well-understood names - `Enum`, `Guest`, `format`, `Result` and so on. If we could set up a linter check that prevents us from importing everything everywhere, I'd be much happier.

tmt/steps/discover/fmf.py
happz: In a separate patch? Because :)

```
$ grep -r "split(['\"]\\\n" tmt
tmt/export/nitrate.py:        for line in cast(str, field.get('environment')).split('\n'):
tmt/steps/discover/fmf.py:                directories = [os.path.dirname(name) for name in output.stdout.split('\n')]
tmt/utils/yaml.py.bak:        for paragraph in value.rstrip().split('\n\n'):
tmt/utils/yaml.py.bak:        yield from value.rstrip().split('\n')
tmt/utils/yaml.py.bak:    for line in result.stdout.split('\n'):
tmt/utils/yaml.py.bak:    for line in result.stdout.split('\n'):
tmt/utils/yaml.py.bak:        for line in content.split("\n"):
tmt/utils/yaml.py.bak:        remotes = output.stdout.split('\n')
tmt/utils/git.py:    for line in result.stdout.split('\n'):
tmt/utils/git.py:    for line in result.stdout.split('\n'):
tmt/utils/git.py:        remotes = output.stdout.split('\n')
tmt/utils/structured_field.py:        for line in content.split("\n"):
tmt/utils/__init__.py:        output_lines = output.strip().split('\n')
tmt/utils/__init__.py:        for paragraph in value.rstrip().split('\n\n'):
tmt/utils/__init__.py:        yield from value.rstrip().split('\n')
tmt/utils/__init__.py:        remotes = output.stdout.split('\n')
tmt/result.py:            iterator = iter(reversed(log.split("\n")))
=
```

tmt/utils/__init__.py
psss: The suffix includes dots as well, so it should be `.yaml` and `.yml`.
Failing `/tests/core/environment-file` confirms this.

tests/unit/test_schemas.py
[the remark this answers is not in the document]
martinhoyer: right, I was thinking adding the `.absolute()` to the tmt.Tree:
```
- self._path = path or Path.cwd()
+ self._path.absolute() = path or Path.cwd() 
```
but that's just useless overthinking from my part.

tmt/utils/__init__.py
happz: Fixed in 0306d54e

tmt/base.py
martinhoyer: you're right.

tmt/steps/discover/fmf.py
martinhoyer: As you wish, although it's "just" 9 other places, you have some `.bak`s and `\n\n` there. :)

tmt/export/nitrate.py
martinhoyer: Sure, I have no horse in the usage of `from` in external imports.     
Still, it could be `import urllib.parse.urlparse` though.

martinhoyer:
@happz FTR, most of my PR review comments are usually questions mainly for myself to learn/understand things. Please feel free to "Resolve" them in case they are blocking something somewhere. (not just this PR) :)

tmt/export/nitrate.py
happz: Hmm, isn't `urlparse` a callable, not a module?

```
$ python
Python 3.12.6 (main, Sep  9 2024, 00:00:00) [GCC 13.3.1 20240522 (Red Hat 13.3.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib.parse.urlparse
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'urllib.parse.urlparse'; 'urllib.parse' is not a package
>>> 
```

martinhoyer: ha, indeed. ok so I do have a horse :D

tests/unit/test_schemas.py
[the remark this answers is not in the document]
psss: Let's move on, we can always improve in the future :)

psss (approved):
Thanks for improving this!

PR #3220, opened 2024-09-17
Document the `known issue` label on the contribute page

psss:
The label is supposed to be used for tracking important issues affecting many users so that those can be identified easily without wasting time and repeated questions on the chat.

Pull Request Checklist

* [x] write the documentation


happz (approved):

psss:
Hm, just realized that the current link shows only open issues. And, if I remember correctly, we want to show closed ones as well, because these important issues shoud get off the list only when they are deployed in production, not already when the fix is merged. So this one would be better, right?

 * https://github.com/teemtee/tmt/issues?q=label%3A%22known+issue%22

Issue #3008, opened 2024-06-12
Plugin schema often lacks shared fields - name, where, order


happz:
```
/plans/cut-8to9 
warn C000 key "upgrade-path" not recognized by schema /schemas/execute/tmt 
warn C000 key "url" not recognized by schema /schemas/execute/tmt 
warn C000 value of "how" is not "tmt" 
warn C000 key "name" not recognized by schema /schemas/execute/upgrade 
warn C000 fmf node failed schema validation
```

```
(dev) [pts-1:0]: happz@multivac [template-raise-an-error] ~/git/tmt $ l tmt/schemas/prepare/
total 20
drwxr-xr-x. 1 happz happz  114 Jun  3 11:31 .
drwxr-xr-x. 1 happz happz  206 Jun 10 21:30 ..
-rw-r--r--. 1 happz happz  273 Aug  7  2023 ansible.yaml
-rw-r--r--. 1 happz happz  664 Aug  7  2023 errata.yaml
-rw-r--r--. 1 happz happz  421 Apr 22 17:42 feature.yaml
-rw-r--r--. 1 happz happz 1019 Aug  7  2023 install.yaml
-rw-r--r--. 1 happz happz  265 Aug  7  2023 shell.yaml
(dev) [pts-1:0]: happz@multivac [template-raise-an-error] ~/git/tmt $ grep 'order' tmt/schemas/prepare/*
tmt/schemas/prepare/install.yaml:  order:
tmt/schemas/prepare/install.yaml:    $ref: "/schemas/core#/definitions/order"
(dev) [pts-1:0]: happz@multivac [template-raise-an-error] ~/git/tmt $ 
```

psss:
Looks like a duplicate of #2984.

PR #3225, opened 2024-09-18
Add missing `where`, `name` and `order` keywords into schema

bajertom:
Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema

Fixes issue #2984 and #3008.
Regarding #3008 - `where` (and  `order` and `name`) is present in all other prepare schemas (except errata). Not sure whether to include it in this one as well?

### DIFF
--- a/tmt/schemas/prepare/errata.yaml
+++ b/tmt/schemas/prepare/errata.yaml
@@ -42,5 +42,15 @@ properties:
   no-filter:
     type: boolean
 
+  name:
+    type: string
+
+  order:
+    $ref: "/schemas/core#/definitions/order"
+
+  where:
+    $ref: "/schemas/common#/definitions/where"
+
+
 required:
   - how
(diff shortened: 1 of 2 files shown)


happz:
> Regarding #3008 - `where` (and `order` and `name`) is present in all other prepare schemas (except errata).

`where` should be also accepted by all `prepare` plugins, including `errata`.

> Not sure whether to include it in this one as well?

I'd go for it, you can include `where` in the subject, yet another key that should be accepted, it's just nobody pointed it out in the issues.

bajertom:
I've included `where`, `name` and `order` into errata and feature schemas. In shell and ansible schemas it is included via reference to common.yaml schema.

bajertom:
Extended unit tests failing due to change in dnf stdout x stderr, fixed here https://github.com/teemtee/tmt/pull/3227.

psss (approved):
Thanks!

PR #3227, opened 2024-09-19
Check both `stdout` and `stderr` in package manager tests

psss:
We don't care that much where `dnf` puts the info about which packages were installed. Plus in different `dnf` versions the output goes into different output streams. See also:

* https://github.com/rpm-software-management/dnf5/pull/1641

Pull Request Checklist


thrix (approved):
LGTM

happz (approved):

PR #3228, opened 2024-09-19
Support container provisioner in toolbox

### LINKED ISSUE #1020: Improve `tmt` user experience on SilverBlue system
When you are on SilverBlue system you are trying to avoid adding packages as layers to the system. So, it's great that you have [podman example in the documentation](https://tmt.readthedocs.io/en/stable/overview.html?highlight=podman#install) but you can't use this for running the tests if they using container or VM. Could you please improve this experince?

Pain points I saw now:
- The install example will not mount existing project directory.
- The container tests won't work out of the box.
- The VM tests won't work out of the box. (not tested but without access to /dev/kvm it won't work)

The best option would be to have a working [Toolbx](https://github.com/containers/toolbox) image, however, I guess that is not doable because it's always rootless and you probably need privilege and rootfull container.

thrix:
For Fedora Silverblue users it is common to run podman via `flatpak-spawn --host` which runs podman on the host system itself. This requires to pass the toolbox container name when running `podman cp` to correctly copy stuff from the toolbox container, where `tmt` is installed to the provisioned container.

Fixes #1020

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tests/provision/container/toolbox/test.sh
+++ b/tests/provision/container/toolbox/test.sh
@@ -39,7 +39,7 @@ rlJournalStart
         rlRun "type toolbox_run"
 
         # Install make and hatch
-        rlRun "toolbox_run sudo dnf -y make hatch"
+        rlRun "toolbox_run sudo dnf -y install make hatch"
 
         # Create a copy of the tmt tree, to mitigate possible permission issues
         rlRun "cp -Rf ${TMT_TREE} ${TOOLBOX_TREE}"
--- a/tests/provision/container/toolbox/test.sh
+++ b/tests/provision/container/toolbox/test.sh
@@ -23,7 +23,7 @@ rlJournalStart
         rlRun "echo export XDG_RUNTIME_DIR=/run/user/$toolbox_user_id >> /home/$toolbox_user/.bashrc"
         rlRun "echo export DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$toolbox_user_id/bus >> /home/$toolbox_user/.bashrc"
 
-        rlRun "sudo -iu $toolbox_user toolbox create -y $toolbox_container_name"
+        rlRun "sudo -iu $toolbox_user toolbox create -i ${TOOLBOX_IMAGE} -y $toolbox_container_name"
     rlPhaseEnd
 
     toolbox_run() {
--- a/tests/provision/container/toolbox/test.sh
+++ b/tests/provision/container/toolbox/test.sh
@@ -1,130 +1,62 @@
 #!/bin/bash
 . /usr/share/beakerlib/beakerlib.sh || exit 1
 
-#
-# The test works in 3 auto-detected modes:
-#
-# 1. Run via `tmt` in Packit environment
-#
-# 2. Run locally, e.g. via `tmt try`
-#
-# 3. Running directly the test script `./test.sh` as a non-root user
-#
-# Running toolbox under root user does not work well, so in case of 1. and 2. a separate
-# account `toolbox` is created. The user needs to have a working dbus session available,
-# so additional steps are taken to activate systemd user session and set required environment
-# variables.
-#
-# In case of 1. the tmt is installed from the copr repository created by Packit.
-#
-# In case of 2. and 3. we need to install tmt from the sources in the toolbox container.
-# For 2. the sources are copied `TMT_TREE` and in case of 3. from the current git repository.
-#
+# Use `tmt try` to run this test locally, running directly the script will not work.
 
 rlJournalStart
     rlPhaseStartSetup
         rlRun "toolbox_container_name=\$(uuidgen)" 0 "Generate toolbox container name"
-
-        # When running via tmt, we will use a new user to run toolbox
-        if env | grep -Eq "(PACKIT_COPR_PROJECT|TMT_TREE)"; then
-          rlRun "toolbox_user=toolbox" 0 "Generate toolbox container name"
-        fi
+        rlRun "toolbox_user=toolbox" 0 "Set user for running toolbox"
     rlPhaseEnd
 
     rlPhaseStartTest "Create toolbox container"
-        # When run via tmt, we will use a test user, toolbox under root does not work well
-        if env | grep -Eq "(PACKIT_COPR_PROJECT|TMT_TREE)"; then
+        # Add a toolbox user. Running toolbox under root user does not work well,
+        # so a separate user account is created.
+        rlRun "useradd $toolbox_user"
+        rlRun "toolbox_user_id=$(id -u $toolbox_user)"
 
-          # Add a toolbox user
-          rlRun "useradd $toolbox_user"
-          rlRun "toolbox_user_id=$(id -u $toolbox_user)"
+        # Make sure systemd user session runs for the new user. The user session
+        # hosts a dbus session, which is required for toolbox.
+        rlRun "loginctl enable-linger $toolbox_user"
 
-          # Make sure systemd user session runs for the new user
-          rlRun "loginctl enable-linger $toolbox_user"
+        # Add required environment variables for toolbox to the user's environment.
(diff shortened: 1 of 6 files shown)


thrix:
@jkonecny12 this is the change I talked about, with it tmt container provisioner works well :)

tmt/steps/provision/podman.py
```diff
@@ -283,6 +284,33 @@ def _run_ansible(
...
+                "Unable to detect toolbox container name, '/run/.containerenv' not found."
+                )
+            return None
+
+        with open('/run/.containerenv') as envfile:
+            for line in envfile.readline():
+                if line.startswith('name="'):
+                    return line[6:-1]
+
+        return None
```
happz, line 312: These two look like they belong to guest facts, as they are properties of a machine similar to `has_selinux` or `arch`, for example. Once moved, you would reach them via  `self.parent.plan.my_run.runner.facts.*`.

tmt/steps/provision/podman.py
```diff
@@ -283,6 +284,33 @@ def _run_ansible(
             log=log,
             silent=silent)
 
+    @functools.cached_property
+    def _is_toolbox(self) -> bool:
+        """ Return ``True`` if running in toolbox, ``False`` otherwise. """
+        if os.path.exists('/run/.toolboxenv'):
```
happz, line 290: `os.path` should be replaced with `Path().exists()`

tmt/steps/provision/podman.py
thrix: thanks for the pointers

tmt/steps/provision/podman.py
thrix: need to remember we use pathlib here :)

janhavlin:
I'm getting this error, any idea?
```
❯ tmt --version
tmt version: 1.37.0.dev20+gce8a3c6e

❯ tmt run -a provision -h container -i fedora:40 plan -n /testing-farm/sanity
/var/tmp/tmt/run-135

/testing-farm/sanity
    discover
        how: fmf
        directory: /var/home/jhavlin/work/tests
        tests: /testing-farm/script
        summary: 1 test selected
    provision
        
        how: container
        image: fedora:40
        fail: Command 'podman pull -q fedora:40' returned 1.
        fail: Command 'podman pull -q fedora:40' returned 1.
        fail: Command 'podman pull -q fedora:40' returned 1.
        fail: Command 'podman pull -q fedora:40' returned 1.
        fail: Command 'podman pull -q fedora:40' returned 1.
        fail: Retries of 'Pulling 'fedora:40' image' unsuccessful.
    finish

plan failed

Cause number 1:

    provision step failed

    Cause number 1:

        Retries of 'Pulling 'fedora:40' image' unsuccessful.

        The exception was caused by 5 earlier exceptions

        Cause number 1:

            Command 'podman pull -q fedora:40' returned 1.

            stderr (1/1 lines)
            Portal call failed: Failed to start command: Failed to change to directory “/var/tmp/tmt/run-135/testing-farm/sanity/provision/default-0” (No such file or directory)
...
```

thrix:
@janhavlin this is something we should document, add this to your environment (I have it in .bashrc):

```
export TMT_WORKDIR_ROOT=$HOME/.local/share/tmt
```

janhavlin:
Thanks, I got it working!

jkonecny12:
> @janhavlin this is something we should document, add this to your environment (I have it in .bashrc):
> 
> ```
> export TMT_WORKDIR_ROOT=$HOME/.local/share/tmt
> ```

Couldn't we have this resolved by default correctly? Requesting users to set somewhere something to make things work seems like an issue generator and unhappy users.

thrix:
> > @janhavlin this is something we should document, add this to your environment (I have it in .bashrc):
> > ```
> > export TMT_WORKDIR_ROOT=$HOME/.local/share/tmt
> > ```
> 
> Couldn't we have this resolved by default correctly? Requesting users to set somewhere something to make things work seems like an issue generator and unhappy users.

@jkonecny12 I agree it is sad, but we will need to have a discussion with `tmt` folks :)

tmt/steps/provision/podman.py
thrix: moved to facts

tmt/steps/provision/podman.py
thrix: moved to facts, where we run shell commands, so this does not apply anymore, resolving.

thrix:
FTR the workdir autodetection will be another patch but it won't be in 1.40.

psss:
The newly added test failed because of a missing `toolbox` package. Should be fixed by e58860d0.

happz (approved):

skycastlelily (approved):

tests/provision/container/toolbox/test.sh
```diff
@@ -0,0 +1,159 @@
...
+    # Execution locally via tmt
+    elif env | grep -q TMT_TREE; then
+        rlPhaseStartTest "Local execution via tmt: Install tmt from TMT_TREE"
+            TOOLBOX_TREE="/var/tmp/tree"
+            TMT_COMMAND="env -C ${TOOLBOX_TREE} hatch -e dev run env -C /tmp tmt"
+
+            rlRun "type toolbox_run"
+
+            # Install all dependencies needed for hatch installation
+            rlRun "toolbox_run sudo dnf -y install git gcc hatch krb5-devel libpq-devel libvirt-devel python-devel"
```
psss, line 90: What about using `make develop` instead of duplicating the list here?

tests/provision/container/toolbox/test.sh
```diff
@@ -0,0 +1,159 @@
...
+# For 2. the sources are copied `TMT_TREE` and in case of 3. from the current git repository.
+#
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "toolbox_container_name=\$(uuidgen)" 0 "Generate toolbox container name"
+
+        # When running via tmt, we will use a new user to run toolbox
+        if env | grep -Eq "(PACKIT_COPR_PROJECT|TMT_TREE)"; then
+          rlRun "toolbox_user=toolbox" 0 "Generate toolbox container name"
```
psss, line 30: Copy & paste comment?

psss (commented):
The change itself looks good. The test setup looks really super-complex and thus, I'm afraid, a bit fragile. Added a few comments though.

tests/provision/container/toolbox/test.sh
thrix: @psss yeah, I can try, is a problem to add to `make develop` also `hatch`?

happz: `make develop` calls `make _deps` which installs `hatch python3-devel python3-hatch-vcs rpm-build`.

thrix: @psss used `make develop`

tests/provision/container/toolbox/test.sh
thrix: yes, fixed, thank you

thrix:
> The newly added test failed because of a missing `toolbox` package. Should be fixed by [e58860d](https://github.com/teemtee/tmt/commit/e58860d0c402d6f528420d942e4ec3a9d1031a44).

I simplified, so I hope this concern is now addressed

psss (approved):
Looks much better now, thanks! Though it seems that tests are red. You've mentioned that `el9` should be disabled, right? Plus there's seems to be an [avc failure](https://artifacts.dev.testing-farm.io/b3d58468-a649-4881-8f8b-a5a12856a957/work-the-restd_1p16_m/plans/provision/container/the-rest/execute/data/guest/default-0/tests/provision/container/toolbox-15/checks/avc.txt) on `fedora-41+`.

thrix:
Issues here are unrelated, merging

PR #3229, opened 2024-09-19
Support Fedora Image Mode

thrix:
To support Fedora Image mode the following changes were required:
    
* Fix `rsync` installation. The tool is not included in the distribution and
   the installation method was completely broken, i.e. `rpm-ostree` has
   no `--install-root` option. We did not hit it as there is no good
   testing environment available that would cover this.
    
* Introduce `TMT_SCRIPTS_DEST_DIR` environment variable set by default
   to `/usr/local/bin` except distributions using `rpm-ostree`  where `/var/lib/tmt/scripts`
   is used. The variable specified the directory hosting the `tmt` scripts. The `PATH` is set
  system-wide via a new `/etc/profile.d/tmt.sh` script for `rpm-ostree` distributions.
    
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] mention the version
* [x] include a release note


tmt/steps/execute/__init__.py
```diff
@@ -58,6 +59,9 @@
 # Scripts source directory
 SCRIPTS_SRC_DIR = tmt.utils.resource_files('steps/execute/scripts')
 
+# Scripts destination directory, hardcoded in the tmt.sh script
+SCRIPTS_DEST_DIR = "/var/tmp/tmt/bin"
```
happz, line 63: It's a path, it shall be a `Path` instance.

tmt/steps/execute/__init__.py
```diff
@@ -77,10 +81,10 @@ class ScriptCreatingFile(Script):
 
 # Script handling reboots, in restraint compatible fashion
 TMT_REBOOT_SCRIPT = ScriptCreatingFile(
-    path=Path("/usr/local/bin/tmt-reboot"),
+    path=Path(f"{SCRIPTS_DEST_DIR}/tmt-reboot"),
     aliases=[
-        Path("/usr/local/bin/rstrnt-reboot"),
-        Path("/usr/local/bin/rhts-reboot")],
+        Path(f"{SCRIPTS_DEST_DIR}/rstrnt-reboot"),
```
happz, line 86: Once `SCRIPTS_DEST_DIR` becomes `Path`, this would turn into `SCRIPTS_DEST_DIR / 'rstrnt-reboot'`

tmt/steps/execute/__init__.py
```diff
@@ -89,43 +93,52 @@ class ScriptCreatingFile(Script):
...
-        Path("/usr/local/bin/rhts-abort")],
+        Path(f"{SCRIPTS_DEST_DIR}/rstrnt-abort"),
+        Path(f"{SCRIPTS_DEST_DIR}/rhts-abort")],
     related_variables=[],
     created_file="abort"
     )
 
+# Profile script for adding SCRIPTS_DEST_DIR to executable pats system-wide
+TMT_ETC_PROFILE_D = Script(
+    path=Path(f"/etc/profile.d/tmt.sh"),
```
happz, line 132: No interpolation, `f` is not needed.

tmt/steps/execute/__init__.py
```diff
@@ -597,6 +610,9 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
 
     def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
         """ Prepare additional scripts for testing """
+        # Create scripts directory
+        guest.execute(Command("mkdir", "-p", f"{SCRIPTS_DEST_DIR}"))
```
happz, line 614: `str(SCRIPTS_DEST_DIR)` might be more readable.

tmt/steps/execute/__init__.py
happz: And I'd add a way to override this path right from the beginning, check out `effective_pidfile_root()`, but that would conflict with easy hard-coding the path into `tmt.sh`... Still, hard-coded paths turned out to not be generally usable, sooner or later, might be a good investment to get this one configurable while we're not [ushed by a looming regression...

tmt/steps/execute/__init__.py
thrix: changed.

tmt/steps/execute/__init__.py
thrix: changed.

tmt/steps/execute/__init__.py
thrix: moved all occurrences using `Path`

tmt/steps/execute/__init__.py
thrix: Makes sense. Let's make it configurable.

thrix: should be done.

tmt/steps/execute/__init__.py
```diff
@@ -595,19 +636,41 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
 
         return invocations
 
+    def _render_script_template(self, source: Path, context: dict[str, str]) -> Path:
```
happz, line 639: This looks like the implementation of a script class leaked into the plugin.

I'd suggest extending `Script` to become a context manager, i.e. adding `__enter__` and `__exit__` methods. The default implementation of `__enter__` would return `self.path`, but `ScriptTemplate` class would do the magic we have here, render the template and write down the path to temporary file. Then do `with script as path` below, and `ScriptTemplate`'s `__exit__` would remove the temporary file when leaving the context. Plugin does not need to be concerned about itnernals of various script implementations, all it needs is a local path to a file, and the rest shouldn't be its responsibility.

tmt/steps/execute/__init__.py
```diff
@@ -89,43 +119,54 @@ class ScriptCreatingFile(Script):
...
     created_file="abort"
     )
 
+# Profile script for adding SCRIPTS_DEST_DIR to executable pats system-wide
+TMT_ETC_PROFILE_D = ScriptTemplate(
+    path=Path("/etc/profile.d/tmt.sh"),
+    aliases=[],
+    related_variables=[],
+    context={
+        'scripts_dest_dir': str(effective_scripts_dest_dir())
```
happz, line 162: I'd recommend using upper case for variable names, which seems to be common in the Jinja world in general, and definitely true in our own templates.

tmt/steps/execute/__init__.py
thrix: ack. nice idea, working it out

tmt/steps/execute/__init__.py
thrix: no objections

martinhoyer (commented):
Do I understand it correctly, that the intention is to have default directory which is writable on all platforms, as opposed to having a default directory, which would be changed if 'default' is not writable?

tmt/steps/execute/__init__.py
thrix: @happz should be resolved. First time I used context manager in the code I wrote, it really makes the code a lot nicer.

.github/workflows/shellcheck.yml
```diff
@@ -28,6 +28,7 @@ jobs:
           exclude-path: |
             tests/**
             examples/**/test.sh
+            tmt/steps/execute/scripts/tmt.sh.j2
```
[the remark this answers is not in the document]
thrix: wanted to be conservative, but it is true that I do not think it makes sense for any templated jinja's so:

```suggestion
            tmt/steps/execute/scripts/*.sh.j2
```

tmt/steps/execute/__init__.py
thrix: done, it is now upper case.

tmt/steps/execute/__init__.py
```diff
@@ -5,6 +5,7 @@
 import os
 import signal as _signal
 import subprocess
+import tempfile
```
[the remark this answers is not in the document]
thrix: @martinhoyer resolved.

tmt/steps/execute/__init__.py
```diff
@@ -55,32 +58,101 @@
...
+@dataclass
+class ScriptTemplate(Script):
+    """
+    Represents a Jinja2 templated script.
+
+    Must be used as a context manager. The context manager returns
+    the source file path.
+
+    The source file name is constructed from the name of the file specified
+    via the ``path` attribute that is extended with the ``.j2`` suffix.
```
thrix, line 112:
```suggestion
    via the ``path`` attribute, with the ``.j2`` suffix appended.
```

tmt/steps/execute/__init__.py
```diff
@@ -55,32 +58,101 @@
...
 class Script:
-    """ Represents a script provided by the internal executor """
+    """
+    Represents a script provided by the internal executor.
+
+    Must be used as a context manager. The context manager returns
+    the source file path.
+
+    The source file name matches the name of the file specified via
+    the ``path` attribute and must be located in the directory specified
```
thrix, line 77:
```suggestion
    the ``path`` attribute and must be located in the directory specified
```

tmt/steps/execute/__init__.py
```diff
@@ -55,32 +58,101 @@
...
+        os.unlink(self._rendered_script_path)
+
+
+def effective_scripts_dest_dir() -> Path:
+    """
+    Find out what the actual scripts destination directory is.
+
+    If the ``TMT_SCRIPTS_DEST_DIR`` variable is set, it is used as the scripts
+    destination directory. Otherwise, the default of :py:data:`SCRIPTS_DEST_DIR`
+    is used.
```
thrix, line 141:
```suggestion
    If the ``TMT_SCRIPTS_DEST_DIR`` environment variable is set, it is used
    as the scripts destination directory. Otherwise, the default
    of :py:data:`SCRIPTS_DEST_DIR` is used.
```

tmt/steps/execute/__init__.py
```diff
@@ -55,32 +58,101 @@
...
+    the ``path`` attribute and must be located in the directory specified
+    via :py:data:`SCRIPTS_SRC_DIR` variable.
+    """
 
     path: Path
     aliases: list[Path]
     related_variables: list[str]
 
+    def __enter__(self) -> Path:
+        return SCRIPTS_SRC_DIR / self.path.name
```
happz, line 86: This should probably return `effective_scripts_dest_dir()` instead of the default value.

happz: Nevermind, I mistook it for `DEST` variable.

tmt/steps/execute/scripts/tmt.sh.j2
```diff
@@ -0,0 +1,4 @@
+# shellcheck shell=bash
+
+# tmt provides executable scripts under this path
+export PATH={{ SCRIPTS_DEST_DIR }}:$PATH
```
happz, line 4: Shouldn't the value be wrapped in quotes, to prevent nasty and ugly effects of spaces and other magic characters? `PATH="{{ ... }}:$PATH"`?

tmt/steps/execute/__init__.py
thrix: hmm, how did I miss that

tmt/steps/execute/scripts/tmt.sh.j2
thrix:
```suggestion
export PATH="{{ SCRIPTS_DEST_DIR }}:$PATH"
```

tmt/steps/execute/__init__.py
thrix:
```suggestion
        return  effective_scripts_dest_dir() / self.path.name
```

thrix: oh, I was not blind, ok :) did nothing

thrix:
@happz @psss added docs, all comments should be resolved

thrix:
Btw we agreed on getting this into Testing Farm sooner than 1.38, once we verify it does not break anything.

thrix:
Please vote for the new location in this discussion poll:

https://github.com/teemtee/tmt/discussions/3260

kkaarreell:
What about SELinux? Given this dir hosts executables, it should be assigned with the respective selinux context.

kkaarreell:
> Do I understand it correctly, that the intention is to have default directory which is writable on all platforms, as opposed to having a default directory, which would be changed if 'default' is not writable?

I am also interested to know. It looks like that alternative location would require some system "bending" and I am not sure if doing this by default Is a good approach. I expect that most tested systems won't be bootc.

thrix:
> What about SELinux? Given this dir hosts executables, it should be assigned with the respective selinux context.

Thanks, depending on the discussion we can adjust the SELinux context. I did not yet see a problem with this with the current path `/var/tmp/tmt/bin` but might be worth checking :)

thrix:
> > Do I understand it correctly, that the intention is to have default directory which is writable on all platforms, as opposed to having a default directory, which would be changed if 'default' is not writable?
> 
> I am also interested to know. It looks like that alternative location would require some system "bending" and I am not sure if doing this by default Is a good approach. I expect that most tested systems won't be bootc.

Bending is already in place and we talked about proper documentation. Feel free to read the MR docs being added.

tmt/steps/execute/__init__.py
```diff
@@ -597,16 +680,18 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
 
     def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
         """ Prepare additional scripts for testing """
+        # Create scripts directory
+        guest.execute(Command("mkdir", "-p", str(effective_scripts_dest_dir())))
```
martinhoyer, line 684: How about using rsync with `--mkpath` ?  
Actually, there is also `--executability` or `--perms`/`--chmod` to get rid of the other ssh command.  
(again, can be a separate PR)

tmt/steps/execute/__init__.py
```diff
@@ -597,16 +680,18 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
-            source = SCRIPTS_SRC_DIR / script.path.name
-
-            for dest in [script.path, *script.aliases]:
-                guest.push(
-                    source=source,
-                    destination=dest,
-                    options=["-p", "--chmod=755"],
-                    superuser=guest.facts.is_superuser is not True)
+            with script as source:
+                for dest in [script.path, *script.aliases]:
```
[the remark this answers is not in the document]
thrix: @martinhoyer yes please, not definitely something for this patch :)

tmt/steps/execute/__init__.py
thrix: @martinhoyer rsync --mkpath was added in rsync 3.2.3 (6 Aug 2020). I wonder if it will work on all supported OSes

martinhoyer: I think I did check that before writing the comment.

lukaszachy: That option has to exist for rsync on the guest as well, right? If so we definitely _cannot_ use that as rhel-7 has just rsync-3.1.2

martinhoyer: hmm, yeah I also see el8 is too old as well. Alas.

tmt/steps/execute/__init__.py
```diff
@@ -55,77 +58,199 @@
...
+    def __exit__(self, *args: object) -> None:
+        pass
+
+    def guest_path(self, guest: Guest, alias: Optional[str] = None) -> Path:
+        """ Return path of the script or the script alias on a guest """
+        filename = alias or self.source_filename
+        scripts_dest_dir = effective_scripts_dest_dir(
+            default=SCRIPTS_DEST_DIR_OSTREE if guest.facts.is_ostree else SCRIPTS_DEST_DIR
+            )
+        return self.destination_path if self.destination_path else scripts_dest_dir / filename
```
happz, line 117: * Is `alias` ever optional? In `script.guest_path(guest, dest)`, `dest` - which I proposed to rename to `filename` - comes from `[script.source_filename, *script.aliases]`, `source_filename` is `str, `script.aliases` is `list[str]`, all in all, `alias` will never be `None`. And `source_filename` is one of the inputs, therefore `filename = alias or self.source_filename` is not needed.
* I'd recommend changing `alias` to `filename` or something, sometimes it's alias, sometimes it's `source_filename`.
* if `self.destination_path` is set, you can return the final path immediately, and only after that you can construct `scripts_dest_dir`.
* `prepare_scripts()` already got hold of the actual `scripts_dest_dir`, hm, do we need this method at all? If the script has `destination_path`, `prepare_scripts` can use it, if it's not set, `prepare_scripts` can do `scripts_dest_dir / filename` because it already has `scripts_dest_dir`.

tmt/steps/execute/__init__.py
```diff
@@ -596,16 +721,25 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
-            source = SCRIPTS_SRC_DIR / script.path.name
-
-            for dest in [script.path, *script.aliases]:
-                guest.push(
-                    source=source,
-                    destination=dest,
-                    options=["-p", "--chmod=755"],
-                    superuser=guest.facts.is_superuser is not True)
+            with script as source:
+                for dest in [script.source_filename, *script.aliases]:
```
happz, line 734: `dest` is now just a filename, correct? `filename` would be better here, there's way too many "dest"-ish names already.

tmt/steps/execute/__init__.py
```diff
@@ -55,77 +58,199 @@
...
 
-# Scripts source directory
+#: Scripts source directory
 SCRIPTS_SRC_DIR = tmt.utils.resource_files('steps/execute/scripts')
 
+#: The default scripts destination directory
+SCRIPTS_DEST_DIR = Path("/usr/local/bin")
+
+#: The default scripts destination directory for rpm-ostree based distributions, https://github.com/teemtee/tmt/discussions/3260
+SCRIPTS_DEST_DIR_OSTREE = Path("/var/lib/tmt/scripts")
```
happz, line 68: Maybe we should call these two `DEFAULT_*`, because the actual path will come out of `effective_scripts_dest_dir()`.

tmt/steps/provision/__init__.py
```diff
@@ -437,6 +438,20 @@ def _query_is_superuser(self, guest: 'Guest') -> Optional[bool]:
 
         return output.stdout.strip() == 'root'
 
+    def _query_is_ostree(self, guest: 'Guest') -> Optional[bool]:
+        # https://github.com/vrothberg/chkconfig/commit/538dc7edf0da387169d83599fe0774ea080b4a37#diff-562b9b19cb1cd12a7343ce5c739745ebc8f363a195276ca58e926f22927238a5R1334
+        output = self._execute(
+            guest,
+            Command(
+                'bash',
```
happz, line 446: `DEFAULT_SHELL`?

tmt/steps/execute/__init__.py
thrix: I do believe I addressed all, except the last one. I added this to mitigate duplication in:

tmt/frameworks/beakerlib.py

where I needed ti for:

```
   36         return Environment(data: {
   37             'BEAKERLIB_DIR': EnvVarValue(raw_value: invocation.path),
   38             'BEAKERLIB_COMMAND_SUBMIT_LOG': EnvVarValue(
   39                 raw_value: f'bash {tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.guest_path(guest: invocation.guest)}')
   40             })
```

tmt/steps/execute/__init__.py
thrix: added as DEFAULT

tmt/steps/provision/__init__.py
thrix: forgot about it, thx

tmt/steps/execute/__init__.py
```diff
@@ -55,77 +58,205 @@
...
     related_variables: list[str]
+    enabled: Callable[[Guest], bool]
+
+    def __enter__(self) -> Path:
+        return SCRIPTS_SRC_DIR / self.source_filename
+
+    def __exit__(self, *args: object) -> None:
+        pass
+
+    def guest_path(self, guest: Guest, filename: Optional[str] = None) -> Path:
```
happz, line 111:
```suggestion
    # TODO: (by happz) consider converting this method to `Guest` (cached) property.
    # It returns a value that doesn't really change per script... Or guest. 
    def guest_path(self, guest: Guest, filename: Optional[str] = None) -> Path:
```

tmt/steps/execute/__init__.py
thrix: yeah, it is, fixed

happz:
@thrix that script path was bugging me, I wasn't sure why, but eventually, I decided to give it a try. It turned out it was the multiple places calling it, and pretty much the independence on guest or script - it didn't really change, no matter what the script filename was, it was called and always gave the same answer... So I played a bit with the code, it's not perfect, but the itching is gone :) Please, consider the following changes:

```diff
diff --git a/tmt/frameworks/beakerlib.py b/tmt/frameworks/beakerlib.py
index 507dd0f5..4a70c05b 100644
--- a/tmt/frameworks/beakerlib.py
+++ b/tmt/frameworks/beakerlib.py
@@ -36,7 +36,7 @@ class Beakerlib(TestFramework):
         return Environment({
             'BEAKERLIB_DIR': EnvVarValue(invocation.path),
             'BEAKERLIB_COMMAND_SUBMIT_LOG': EnvVarValue(
-                f'bash {tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.guest_path(invocation.guest)}')
+                f'bash {invocation.guest.scripts_path / tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.source_filename}')  # noqa: E501
             })
 
     @classmethod
diff --git a/tmt/steps/execute/__init__.py b/tmt/steps/execute/__init__.py
index 4ea9703c..564d3364 100644
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -727,13 +727,9 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
 
     def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
         """ Prepare additional scripts for testing """
-        # For rpm-ostree based distributions use a different default destination directory
-        scripts_dest_dir = effective_scripts_dest_dir(
-            default=DEFAULT_SCRIPTS_DEST_DIR_OSTREE
-            if guest.facts.is_ostree else DEFAULT_SCRIPTS_DEST_DIR)
 
         # Make sure scripts directory exists
-        guest.execute(Command("mkdir", "-p", str(scripts_dest_dir)))
+        guest.execute(Command("mkdir", "-p", str(guest.scripts_path)))
 
         # Install all scripts on guest
         for script in self.scripts:
@@ -742,7 +738,7 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
                     if script.enabled(guest):
                         guest.push(
                             source=source,
-                            destination=script.guest_path(guest, filename),
+                            destination=(script.destination_path or guest.scripts_path) / filename,
                             options=[
                                 "-p",
                                 "--chmod=755"],
diff --git a/tmt/steps/provision/__init__.py b/tmt/steps/provision/__init__.py
index fe5ddf76..de43937e 100644
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -844,6 +844,17 @@ class Guest(tmt.utils.Common):
         return tmt.package_managers.find_package_manager(
             self.facts.package_manager)(guest=self, logger=self._logger)
 
+    @functools.cached_property
+    def scripts_path(self) -> Path:
+        """ Absolute path to tmt scripts directory """
+
+        import tmt.steps.execute
+
+        # For rpm-ostree based distributions use a different default destination directory
+        return tmt.steps.execute.effective_scripts_dest_dir(
+            default=tmt.steps.execute.DEFAULT_SCRIPTS_DEST_DIR_OSTREE
+            if self.facts.is_ostree else tmt.steps.execute.DEFAULT_SCRIPTS_DEST_DIR)
+
     @classmethod
     def options(cls, how: Optional[str] = None) -> list[tmt.options.ClickOptionDecoratorType]:
         """ Prepare command line options related to guests """
```

thrix:
> @thrix that script path was bugging me, I wasn't sure why, but eventually, I decided to give it a try. It turned out it was the multiple places calling it, and pretty much the independence on guest or script - it didn't really change, no matter what the script filename was, it was called and always gave the same answer... So I played a bit with the code, it's not perfect, but the itching is gone :) Please, consider the following changes:
> 
> ```diff
> diff --git a/tmt/frameworks/beakerlib.py b/tmt/frameworks/beakerlib.py
> index 507dd0f5..4a70c05b 100644
> --- a/tmt/frameworks/beakerlib.py
> +++ b/tmt/frameworks/beakerlib.py
> @@ -36,7 +36,7 @@ class Beakerlib(TestFramework):
>          return Environment({
>              'BEAKERLIB_DIR': EnvVarValue(invocation.path),
>              'BEAKERLIB_COMMAND_SUBMIT_LOG': EnvVarValue(
> -                f'bash {tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.guest_path(invocation.guest)}')
> +                f'bash {invocation.guest.scripts_path / tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.source_filename}')  # noqa: E501
>              })
>  
>      @classmethod
> diff --git a/tmt/steps/execute/__init__.py b/tmt/steps/execute/__init__.py
> index 4ea9703c..564d3364 100644
> --- a/tmt/steps/execute/__init__.py
> +++ b/tmt/steps/execute/__init__.py
> @@ -727,13 +727,9 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
>  
>      def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
>          """ Prepare additional scripts for testing """
> -        # For rpm-ostree based distributions use a different default destination directory
> -        scripts_dest_dir = effective_scripts_dest_dir(
> -            default=DEFAULT_SCRIPTS_DEST_DIR_OSTREE
> -            if guest.facts.is_ostree else DEFAULT_SCRIPTS_DEST_DIR)
>  
>          # Make sure scripts directory exists
> -        guest.execute(Command("mkdir", "-p", str(scripts_dest_dir)))
> +        guest.execute(Command("mkdir", "-p", str(guest.scripts_path)))
>  
>          # Install all scripts on guest
>          for script in self.scripts:
> @@ -742,7 +738,7 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
>                      if script.enabled(guest):
>                          guest.push(
>                              source=source,
> -                            destination=script.guest_path(guest, filename),
> +                            destination=(script.destination_path or guest.scripts_path) / filename,
>                              options=[
>                                  "-p",
>                                  "--chmod=755"],
> diff --git a/tmt/steps/provision/__init__.py b/tmt/steps/provision/__init__.py
> index fe5ddf76..de43937e 100644
> --- a/tmt/steps/provision/__init__.py
> +++ b/tmt/steps/provision/__init__.py
> @@ -844,6 +844,17 @@ class Guest(tmt.utils.Common):
>          return tmt.package_managers.find_package_manager(
>              self.facts.package_manager)(guest=self, logger=self._logger)
>  
> +    @functools.cached_property
> +    def scripts_path(self) -> Path:
> +        """ Absolute path to tmt scripts directory """
> +
> +        import tmt.steps.execute
> +
> +        # For rpm-ostree based distributions use a different default destination directory
> +        return tmt.steps.execute.effective_scripts_dest_dir(
> +            default=tmt.steps.execute.DEFAULT_SCRIPTS_DEST_DIR_OSTREE
> +            if self.facts.is_ostree else tmt.steps.execute.DEFAULT_SCRIPTS_DEST_DIR)
> +
>      @classmethod
>      def options(cls, how: Optional[str] = None) -> list[tmt.options.ClickOptionDecoratorType]:
>          """ Prepare command line options related to guests """
> ```



> @thrix that script path was bugging me, I wasn't sure why, but eventually, I decided to give it a try. It turned out it was the multiple places calling it, and pretty much the independence on guest or script - it didn't really change, no matter what the script filename was, it was called and always gave the same answer... So I played a bit with the code, it's not perfect, but the itching is gone :) Please, consider the following changes:
> 
> ```diff
> diff --git a/tmt/frameworks/beakerlib.py b/tmt/frameworks/beakerlib.py
> index 507dd0f5..4a70c05b 100644
> --- a/tmt/frameworks/beakerlib.py
> +++ b/tmt/frameworks/beakerlib.py
> @@ -36,7 +36,7 @@ class Beakerlib(TestFramework):
>          return Environment({
>              'BEAKERLIB_DIR': EnvVarValue(invocation.path),
>              'BEAKERLIB_COMMAND_SUBMIT_LOG': EnvVarValue(
> -                f'bash {tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.guest_path(invocation.guest)}')
> +                f'bash {invocation.guest.scripts_path / tmt.steps.execute.TMT_FILE_SUBMIT_SCRIPT.source_filename}')  # noqa: E501
>              })
>  
>      @classmethod
> diff --git a/tmt/steps/execute/__init__.py b/tmt/steps/execute/__init__.py
> index 4ea9703c..564d3364 100644
> --- a/tmt/steps/execute/__init__.py
> +++ b/tmt/steps/execute/__init__.py
> @@ -727,13 +727,9 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
>  
>      def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
>          """ Prepare additional scripts for testing """
> -        # For rpm-ostree based distributions use a different default destination directory
> -        scripts_dest_dir = effective_scripts_dest_dir(
> -            default=DEFAULT_SCRIPTS_DEST_DIR_OSTREE
> -            if guest.facts.is_ostree else DEFAULT_SCRIPTS_DEST_DIR)
>  
>          # Make sure scripts directory exists
> -        guest.execute(Command("mkdir", "-p", str(scripts_dest_dir)))
> +        guest.execute(Command("mkdir", "-p", str(guest.scripts_path)))
>  
>          # Install all scripts on guest
>          for script in self.scripts:
> @@ -742,7 +738,7 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
>                      if script.enabled(guest):
>                          guest.push(
>                              source=source,
> -                            destination=script.guest_path(guest, filename),
> +                            destination=(script.destination_path or guest.scripts_path) / filename,
>                              options=[
>                                  "-p",
>                                  "--chmod=755"],
> diff --git a/tmt/steps/provision/__init__.py b/tmt/steps/provision/__init__.py
> index fe5ddf76..de43937e 100644
> --- a/tmt/steps/provision/__init__.py
> +++ b/tmt/steps/provision/__init__.py
> @@ -844,6 +844,17 @@ class Guest(tmt.utils.Common):
>          return tmt.package_managers.find_package_manager(
>              self.facts.package_manager)(guest=self, logger=self._logger)
>  
> +    @functools.cached_property
> +    def scripts_path(self) -> Path:
> +        """ Absolute path to tmt scripts directory """
> +
> +        import tmt.steps.execute
> +
> +        # For rpm-ostree based distributions use a different default destination directory
> +        return tmt.steps.execute.effective_scripts_dest_dir(
> +            default=tmt.steps.execute.DEFAULT_SCRIPTS_DEST_DIR_OSTREE
> +            if self.facts.is_ostree else tmt.steps.execute.DEFAULT_SCRIPTS_DEST_DIR)
> +
>      @classmethod
>      def options(cls, how: Optional[str] = None) -> list[tmt.options.ClickOptionDecoratorType]:
>          """ Prepare command line options related to guests """
> ```

@happz thanks, applied, plus removed the unneeded `guest_path` function from `Script`.

tmt/steps/execute/__init__.py
```diff
@@ -596,16 +714,23 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
 
     def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
         """ Prepare additional scripts for testing """
+
+        # Make sure scripts directory exists
+        guest.execute(Command("sudo" if guest.facts.is_superuser else "",
```
thrix, line 719:
```suggestion
        guest.execute(Command("sudo" if not guest.facts.is_superuser else "",
```

tmt/steps/provision/__init__.py
```diff
@@ -437,6 +438,20 @@ def _query_is_superuser(self, guest: 'Guest') -> Optional[bool]:
 
         return output.stdout.strip() == 'root'
 
+    def _query_is_ostree(self, guest: 'Guest') -> Optional[bool]:
+        # https://github.com/vrothberg/chkconfig/commit/538dc7edf0da387169d83599fe0774ea080b4a37#diff-562b9b19cb1cd12a7343ce5c739745ebc8f363a195276ca58e926f22927238a5R1334
+        output = self._execute(
+            guest,
+            Command(
```
martinhoyer, line 445: Why `Command` instead of Path.exists()?

martinhoyer (approved):

tmt/steps/provision/__init__.py
happz: `Path.exists()` would tell you whether the path exists on the machine running tmt.

tmt/steps/execute/__init__.py
```diff
@@ -55,77 +58,192 @@
...
+
+    If the ``TMT_SCRIPTS_DEST_DIR`` environment variable is set, it is used
+    as the scripts destination directory. Otherwise, the ``default``
+    is returned.
+    """
+
+    if SCRIPTS_DEST_DIR_VARIABLE in os.environ:
+        return Path(os.environ[SCRIPTS_DEST_DIR_VARIABLE])
+
+    return default
```
KwisatzHaderach, line 167: can be just
```suggestion
    return Path(os.environ.get(SCRIPTS_DEST_DIR_VARIABLE, default))
```
seems to be working fine even for Path(Path())
but either way is fine with me

KwisatzHaderach (approved):

tmt/steps/provision/__init__.py
thrix: Yeah, we need it for the guest.

tmt/steps/execute/__init__.py
thrix: save 3 lines, good point

happz:
@thrix do you plan to include https://github.com/teemtee/tmt/pull/3229/commits/e2db1d03ec5b9eeb0cc1bc307561833d03c32a86 in this PR? It would be lost between squashes, and such a change deserves its own spot in the sun :)

thrix:
> @thrix do you plan to include [e2db1d0](https://github.com/teemtee/tmt/commit/e2db1d03ec5b9eeb0cc1bc307561833d03c32a86) in this PR? It would be lost between squashes, and such a change deserves its own spot in the sun :)

@happz of course not trying to squeeze it in, had to use it to test this MR in my environment :D
It is dropped with other changes. Ready for your review.

tmt/steps/provision/__init__.py
```diff
@@ -1306,23 +1333,7 @@ def _check_rsync(self) -> CheckRsyncOutcome:
...
 
-        # Install under '/root/pkg' for read-only distros
-        # (for now the check is based on 'rpm-ostree' presence)
-        # FIXME: Find a better way how to detect read-only distros
-        # self.debug("Check for a read-only distro.")
-        if self.facts.package_manager == 'rpm-ostree':
-            self.package_manager.install(
-                Package('rsync'),
-                options=tmt.package_managers.Options(
-                    install_root=Path('/root/pkg'),
```
psss, line 1317: Is this really not needed? It seems there is a `--sysroot=SYSROOT` option supported by `rpm-ostree`. What should happen for ostree images where `rsync` is not installed? There definitely was a use case for that before.

docs/overview.rst
```diff
@@ -466,6 +466,18 @@ TMT_REBOOT_TIMEOUT
     How many seconds to wait for a connection to succeed after
     guest reboot. By default, it is 10 minutes.
 
+
+TMT_SCRIPTS_DEST_DIR
```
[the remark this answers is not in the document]
thrix: ok, changing thanks

docs/releases.rst
```diff
@@ -4,7 +4,6 @@
     Releases
 ======================
 
-
```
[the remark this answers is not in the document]
thrix: ack, bad rebase

tmt/steps/provision/__init__.py
thrix: @psss no it is not needed:

1. first of all, this code never got called in tests, because in all environments `rsync` was available (even RHIVOS), if it would be called, it started to be failing and we would know sooner this part of the code was total bogus
2. special handling is not needed at all, `rsync` gets installed as on normal distros, see [this part of the code](https://github.com/teemtee/tmt/blob/d2b99968e0c07af67ae888ec2c45c9528c531933/tmt/steps/provision/__init__.py#L1336) where it is installed if not available
3. while `sysroot` is available, it is not needed to use it, this is a leftover from the time we were porting from `yum` to `rpm-ostree`, where the leftover options come from

docs/overview.rst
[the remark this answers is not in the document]
psss: Good, thanks!

tmt/steps/provision/__init__.py
psss: Ah, I see. Thanks for the clarification.

psss (approved):
Now looks good. Thanks much for implementing this!

PR #3230, opened 2024-09-20
Fix missing `f-string` in the `tmt lint` check

psss:
Instead of the key name `key "{bad_property}" not recognized` was printed to the user when an invalid key was present in the step phase configuration.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/lint/plan/data/invalid-plugin-key.fmf
+++ b/tests/lint/plan/data/invalid-plugin-key.fmf
@@ -1,0 +1,5 @@
+prepare:
+    how: feature
+    wrong: key
+execute:
+    how: tmt
--- a/tests/lint/plan/test.sh
+++ b/tests/lint/plan/test.sh
@@ -67,6 +67,10 @@ rlJournalStart
         rlAssertGrep "fail P001 unknown key \"discove\" is used" $rlRun_LOG
         rlAssertGrep "fail P001 unknown key \"environmen\" is used" $rlRun_LOG
         rlAssertGrep "fail P001 unknown key \"summaryABCDEF\" is used" $rlRun_LOG
+
+        rlRun -s "$tmt plan lint invalid-plugin-key" 0
+        rlAssertGrep 'warn C000 key "wrong" not recognized by schema$' $rlRun_LOG
+        rlAssertGrep 'warn C000 key "wrong" not recognized by schema /schemas/prepare/feature' $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "P007: step phases require existing guests and roles"
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -939,7 +939,7 @@ class Core(
                             f'key "{bad_property}" not recognized by schema {error.schema["$id"]}'
                     else:
                         yield LinterOutcome.WARN, \
-                            'key "{bad_property}" not recognized by schema'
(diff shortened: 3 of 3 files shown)


psss:
After adding a simple test coverage I've noticed that the invalid key does not result in a non-zero exit code:

```
:: [ 15:26:41 ] :: [  BEGIN   ] :: Running 'tmt plan lint invalid-plugin-key'
    warn: /invalid-plugin-key:prepare - {'how': 'feature', 'wrong': 'key'} is not valid under any of the given schemas
/invalid-plugin-key
warn C000 key "wrong" not recognized by schema
warn C000 value of "how" is not "ansible"
warn C000 key "wrong" not recognized by schema /schemas/prepare/install
warn C000 value of "how" is not "install"
warn C000 key "wrong" not recognized by schema
warn C000 value of "how" is not "shell"
warn C000 key "wrong" not recognized by schema /schemas/prepare/errata
warn C000 value of "how" is not "errata"
warn C000 key "wrong" not recognized by schema /schemas/prepare/feature
warn C000 fmf node failed schema validation
skip P004 discover step is not defined
pass P007 prepare phase 'default-0' does not require specific guest

    warn: /invalid-plugin-key:prepare - {'how': 'feature', 'wrong': 'key', 'name': 'default-0'} is not valid under any of the given schemas
Lint checks on all

:: [ 15:26:42 ] :: [   FAIL   ] :: Command 'tmt plan lint invalid-plugin-key' (Expected 1, got 0)
```

@happz, any idea why these invalid keys only give `warn` instead of `fail`?

https://github.com/teemtee/tmt/blob/94f25736a5bb04e46d57e847e8901f5dd03e19b6/tmt/base.py#L936-L942

happz:
> @happz, any idea why these invalid keys only give `warn` instead of `fail`?
> 
> https://github.com/teemtee/tmt/blob/94f25736a5bb04e46d57e847e8901f5dd03e19b6/tmt/base.py#L936-L942

Mostly because schema-based validation is still not good enough to be enforced, reporting is clumsy, there are out-of-tree schemas that are not recognized by tmt, and I never felt it's ready because it's detached from plugin keys and maintained separately (recent patch adding `when` & co...). On top of that, tmt did not stop its users from using invalid keys, and making the validation mandatory would be a huge step we have not discussed yet.

psss:
Ok, makes sense. I was a bit confused by other cases which actually `fail` when an uknown key is present:

https://github.com/teemtee/tmt/blob/fb19fb0863104fa0d79499fc57929cbf675c5fed/tests/lint/plan/test.sh#L64-L69

Will adjust the exit code expectations for now for the new test.

happz:
> Ok, makes sense. I was a bit confused by other cases which actually `fail` when an uknown key is present:
> 
> https://github.com/teemtee/tmt/blob/fb19fb0863104fa0d79499fc57929cbf675c5fed/tests/lint/plan/test.sh#L64-L69
> 
> Will adjust the exit code expectations for now for the new test.

I believe these were present in some form, I merely converted them to a linter, and they were more restrictive, hence the fail.

psss:
Understod. Ok to merge this one-liner?

happz (approved):

seberm (approved):

happz:
Unrelated failure, merging.

Issue #3226, opened 2024-09-19
tmt-report-result: may need to escape special symbols like colon


juk:
It seems like tmt-report-result script adds testcase name  to log file name. https://github.com/teemtee/tmt/blob/8245c4797d69b6a1574ae9e4445d6eb571080389/tmt/steps/execute/scripts/tmt-report-result#L26 
One of my tests has colons in it's testcase names which appear is not expected by some parser.
```
01:32:49         Command returned '0' (success).
01:32:49         Extract results of '/livepatch/selftests'.
01:32:49         fail:
            mapping values are not allowed in this context
              in "<unicode string>", line 1, column 26
01:32:49     
01:32:49         summary: 0 tests executed
```
tmt-report-results.yaml
```
  - name: /1..7 selftests: livepatch: test-callbacks.sh Pass
    result: pass
    log:
      - 1..7_selftests:_livepatch:_test-callbacks.sh_Pass_test-callbacks_result.log
    end-time: "2024-09-14T01:32:26.859726+00:00"
    data-path: /data/guest/default-0/livepatch/selftests-1/data
  - name: /2..7 selftests: livepatch: test-ftrace.sh Pass
    result: pass
    log:
      - 2..7_selftests:_livepatch:_test-ftrace.sh_Pass_test-ftrace_result.log
    end-time: "2024-09-14T01:32:28.841241+00:00"
    data-path: /data/guest/default-0/livepatch/selftests-1/data
  - name: /3..7 selftests: livepatch: test-livepatch.sh Pass
    result: pass
    log:
      - 3..7_selftests:_livepatch:_test-livepatch.sh_Pass_test-livepatch_result.log
    end-time: "2024-09-14T01:32:36.808714+00:00"
    data-path: /data/guest/default-0/livepatch/selftests-1/data
  - name: /4..7 selftests: livepatch: test-shadow-vars.sh Pass
    result: pass
    log:
      - 4..7_selftests:_livepatch:_test-shadow-vars.sh_Pass_test-shadow-vars_result.log
    end-time: "2024-09-14T01:32:36.868439+00:00"
    data-path: /data/guest/default-0/livepatch/selftests-1/data
  - name: /5..7 selftests: livepatch: test-state.sh Pass
    result: pass
    log:
      - 5..7_selftests:_livepatch:_test-state.sh_Pass_test-state_result.log
    end-time: "2024-09-14T01:32:43.781280+00:00"
[...]

PR #3231, opened 2024-09-22
Quote values in the `tmt-report-results.yaml` file

seberm:
Put all the generated values inside quotation marks in the `tmt-report-results.yaml` YAML file.

Try to fix:
- https://github.com/teemtee/tmt/issues/3226

Pull Request Checklist

* [x] implement the feature
* [x] include a release note


tmt/steps/execute/scripts/tmt-report-result
```diff
@@ -10,14 +10,14 @@ die() { echo "$*" >&2; exit 2; }  # complain to STDERR and exit with error
 needs_arg() { if [ -z "$OPTARG" ]; then die "No arg for --$OPT option"; fi; }
 
 write_report_file () {
-    echo "  - name: ${TESTNAME}" >> "$REPORT_RESULT_OUTPUTFILE"
-    echo "    result: ${TESTRESULT}" >> "$REPORT_RESULT_OUTPUTFILE"
+    echo "  - name: \"${TESTNAME}\"" >> "$REPORT_RESULT_OUTPUTFILE"
```
psss, line 13: And now we have to hope that user does not include a `"` character in the test name :grin: 

    tmt-report-result "first:\" test" pass

Which then results into:

    Invalid yaml syntax: while parsing a block mapping
      in "<unicode string>", line 1, column 5
    did not find expected key
      in "<unicode string>", line 1, column 21

Shall we escape quotes as well?

seberm: > Shall we escape quotes as well?

I think yes, we should escape them. Please, take a look again, the double quotes should be handled correctly now.

I also added a release note.

seberm: Hm, and if we want to be completely precise, we should probably escape double quotes in the log file name (`$outputFile`) as well, because on Unix filesystems, quotes can appear in filenames :man_facepalming: .

There is also a `data-path`, but this seems to be always constructed correctly from TMT variables:

```
echo "    data-path: \"${TMT_TEST_DATA/$TMT_TEST_EXE}\"" >> "$REPORT_RESULT_OUTPUTFILE"
```

What do you think?

tests/execute/result/special.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
psss, line 1: Seems that `fmf` test metadata are missing, thus the test script is not executed.

tests/execute/result/special/special-chars.fmf
```diff
@@ -0,0 +1,9 @@
+execute:
+    how: tmt
+discover:
+    how: shell
+    tests:
+      - name: '0..7 special: characters: *$@|&>< in: the: name'
+        test: "tmt-report-result '0..7 \"special\": \" characters: *$@|&>< in: the: name' PASS"
```
psss, line 7: I think `result: restraint` needs to be set as well to actually parse the generated results file.

tmt/steps/execute/scripts/tmt-report-result
psss: I'd say, for now the fix is sufficient. Let's get rid of the regression and improve things in the future if needed.

psss (changes requested):
Thanks much for fixing this! Looks good, added a few comments and adjusted release not width in 9f78c356.

psss (approved):
Now looks good, thanks!

thrix (approved):
LGTM

therazix (approved):

PR #3233, opened 2024-09-23
Do not traceback upon an empty docstring

psss:
Providing no docstring for a `click` command would result into a confusing `list index out of range` error. Check for the rendered paragraphs first to prevent the traceback.


psss:
@tkoscieln ran into this when implementing the new `tmt link` command. Trivial change.

happz (approved):

Issue #3185, opened 2024-09-03
Implement the check `result` key


psss:
Similarly to the `result` key it would be good to have a `check-result` key which would decide how the check results should affect the final result of a test, e.g. `respect` or `ignore`. The original result of the test execution itself should be stored in the `results.yaml` as well so that it can be inspected if needed.

There is an expectation that check results should by default affect the final result so that possible problems are revealed for example in Testing Farm viewer, so the default values should probably be `respect`.

dennisbrendel:
General comment on the way check works: It calls `dmesg -c` (or `dmesg --read-clear`), so it also messes with the buffer, which can be unexpected for users. In addition it is not always reliable, because some tests also clear the buffer during runtime.

Are there concerns switching over to reading the journal instead? And fallback to `dmesg` when no systemd-journald is available?

martinhoyer:
> General comment on the way check works: It calls `dmesg -c` (or `dmesg --read-clear`), so it also messes with the buffer, which can be unexpected for users. In addition it is not always reliable, because some tests also clear the buffer during runtime.
> 
> Are there concerns switching over to reading the journal instead? And fallback to `dmesg` when no systemd-journald is available?

Good point, thanks! [Check](https://tmt.readthedocs.io/en/stable/spec/tests.html#check) is a optional step and not necessarily 'dmesg'. For example, there can be "journal" check plugin implemented.

That said, it was discussed before. I'd like to look into incorporating `coredumpctl` tool as a "test check". 
In any case, this PR is not the right place for this discussion :)

martinhoyer:
@psss I'm thinking whether it wouldn't make sense to have `ignore` option per each check instead. It could look like:
```
check:
  - how: avc
    enable: false
  - how: dmesg
    ignore: true
```
wdyt?

happz:
@martinhoyer Would it make sense to use the `result` key, similarly to how is used in test metadata? I.e. if I wish to (temporarily) ignore a check, I'd set its `result` to `pass`, and so on. The benefit would be fewer new keys and concepts, and this `ignore` seems like a special case of `result`, https://tmt.readthedocs.io/en/stable/spec/tests.html#result

martinhoyer:
@happz hmm not sure. Afaik we don't want to change the result of the check itself.  
Other idea is to allow check.enabled to be true/false/ignore, which could be nice from user perspective, but I assume quite funky under the hood.

[...]

PR #3239, opened 2024-09-25
Have failure in check result propagate to the test result

martinhoyer:
Trying to address https://github.com/teemtee/tmt/issues/3185  

This is a bit more complex than I thought, so I thought I'd open a (rough initial) draft and ask(beg) smarter people for a direction to go.
Should it be processed in execute step? 
Any pointers appreciated @happz @psss. :pray: 

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


tmt/result.py
```diff
@@ -302,31 +322,54 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
+        # Check for failed checks
+        checks_failed = any(check.result == ResultOutcome.FAIL for check in self.check)
+
+        if interpret == ResultInterpret.RESPECT:
+            if self.respect_checks and checks_failed:
+                self.result = ResultOutcome.FAIL
+                if self.note:
+                    self.note += ', check failed'
+                else:
+                    self.note = 'check failed'
```
happz, line 337: @psss @thrix @lukaszachy what do you think about changing results’ `note` field to a list of strings? It is clearly used like a list, various places add their notes, they always have to chech whether there's any note already, add a comma… If it looks like a list of notes, and it quacks like a list of notes…

Will it break anything, e.g. in TF?

martinhoyer:
Rebased to @happz's [result-store-original-result](https://github.com/teemtee/tmt/tree/result-store-original-result) https://github.com/teemtee/tmt/issues/3185 and tried to use beakerlib for the first time. Feels werid to do all the assertgreps, but I've been merely following the existing examples.

Still proof of concept, but at least I finally figured that the "after-test" check were not available without the change in execute/internal.py.

happz:
@martinhoyer if you change the base branch to the one from https://github.com/teemtee/tmt/pull/3147, the diff should reduce a bit.

tmt/result.py
```diff
@@ -302,31 +322,54 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
+                if self.note:
+                    self.note += ', unexpected pass'
+                else:
+                    self.note = 'unexpected pass'
             return self
 
         # Extend existing note or set a new one
         if self.note:
             self.note += f', original result: {self.result.value}'
-
```
happz, line 311: Please, keep the whitespace for those of us who don’t enjoy dense code :)

martinhoyer: Don't remember doing this. autopep8?  
In any case, I'll clean things up before being ready for review.

tmt/result.py
```diff
@@ -302,31 +322,54 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
 
-        if interpret in (ResultInterpret.RESPECT, ResultInterpret.CUSTOM):
+        if interpret == ResultInterpret.CUSTOM:
+            return self
+
+        # Check for failed checks
+        checks_failed = any(check.result == ResultOutcome.FAIL for check in self.check)
+
+        if interpret == ResultInterpret.RESPECT:
+            if self.respect_checks and checks_failed:
```
happz, line 332: `respect_check` feels more like an input, IIUIC it's basically `interpret`, but in respect to checks. So, they should enter the method through the same door, so to speak.

martinhoyer: I see, so you want me to create
```
class CheckResults(enum.Enum):
    RESPECT = 'respect'
    IGNORE = 'ignore'
...
```
?

I'm so lost in all the schemas, specs, et al.

happz: I suppose we could re-use `ResultInterpret`, it has the same purpose, and `skip` can serve as `ignore`.

happz: Would this help?

```python
return _result.interpret_result(
    ResultInterpret(invocation.test.result) if invocation.test.result else ResultInterpret.RESPECT,
    ResultInterpret(invocation.test.check_result) if invocation.test.check_result else ResultInterpret.RESPECT
)
```

martinhoyer:
>  tmt.utils.GeneralError: Test check 'dmesg' was not found in check registry.

perplexing

tmt/result.py
happz: > I'm so lost in all the schemas, specs, et al.

It shouldn't be that messy. If I can help somehow, let me know, maybe the whole area needs more documentation.

tmt/result.py
```diff
@@ -286,7 +301,12 @@ def from_test_invocation(
             ids=ids,
             log=log or [],
             guest=ResultGuestData.from_test_invocation(invocation=invocation),
-            data_path=invocation.relative_test_data_path)
+            data_path=invocation.relative_test_data_path,
+            check_result=check_result,
+            check=invocation.check_results)
+
+        for check in _result.check:
```
martinhoyer, line 308: debugging leftovers, will remove.

tmt/result.py
martinhoyer: > ```python
> return _result.interpret_result(
>     ResultInterpret(invocation.test.result) if invocation.test.result else ResultInterpret.RESPECT,
>     ResultInterpret(invocation.test.check_result) if invocation.test.check_result else ResultInterpret.RESPECT
> )
> ```

Trying to wrap my head around it. Something like this: 
```
diff --git a/tmt/result.py b/tmt/result.py
index 7e172a47..c8ab2238 100644
--- a/tmt/result.py
+++ b/tmt/result.py
@@ -237,10 +237,10 @@ class Result(BaseResult):
         serialize=lambda path: None if path is None else str(path),
         unserialize=lambda value: None if value is None else Path(value)
         )
-    check_result: str = field(
-        default='respect',
-        serialize=lambda value: value,
-        unserialize=lambda value: value
+    check_result: ResultInterpret = field(
+        default=ResultInterpret.RESPECT,
+        serialize=lambda value: value.value,
+        unserialize=lambda value: ResultInterpret(value)
         )
 
     @classmethod
@@ -286,7 +286,8 @@ class Result(BaseResult):
         default_ids.update(ids)
         ids = default_ids
 
-        check_result = cast(str, invocation.test.node.get('check-result', 'respect'))
+        check_result = ResultInterpret(invocation.test.node.get(
+            'check-result', ResultInterpret.RESPECT.value))
 
         _result = Result(
             name=invocation.test.name,
@@ -319,6 +320,7 @@ class Result(BaseResult):
         attributes, following the ``interpret`` value.
 
         :param interpret: how to interpret current result.
+        :param check_interpret: how to interpret check results.
         :returns: :py:class:`Result` instance containing the updated result.
         """
 
@@ -329,7 +331,7 @@ class Result(BaseResult):
         checks_failed = any(check.result == ResultOutcome.FAIL for check in self.check)
 
         if interpret == ResultInterpret.RESPECT:
-            if self.check_result == 'respect' and checks_failed:
+            if self.check_result == ResultInterpret.RESPECT and checks_failed:
                 self.result = ResultOutcome.FAIL
                 if self.note:
                     self.note += ', check failed'
```
And rename "ignore" to "skip"?

tests/execute/result/check_results/main.fmf
```diff
@@ -0,0 +1,49 @@
+summary: Tests for check results behavior
```
martinhoyer, line 1: > behavior

This is bad and I like to change to `behaviour` if allowed :gb:

tmt/result.py
happz: Yes, something like this, with one exception: I don't understand why `Result.check_result` exists. How checks should be interpreted is a test-level metadata key, i.e. it should come from the test, not the result. Isn't it the same story as `Test.result`?

martinhoyer: Thanks. Honestly that's a leftover from the previous attempts to make things work that I missed and kept.

happz: In the light of https://github.com/teemtee/tmt/issues/3185#issuecomment-2383398847, no `check-result` on test level, I suppose...

This means new `result` check-level field would reside in `tmt.checks.Check` class (and `_RawCheck`), and interpretation of check results need to match a result with corresponding check in `test.check` list...

martinhoyer: Yep

spec/tests/check.fmf
```diff
@@ -37,6 +44,22 @@ example:
...
+        result: xfail
+
+  - |
+    # Enable multiple checks with different result interpretations
+    check:
+      - how: avc
+        result: respect
+      - how: dmesg
+        result: skip
+      - how: kernel-panic
```
martinhoyer, line 60: making up test check plugins was not a best idea. Use better example.

spec/tests/result.fmf
```diff
@@ -42,6 +41,7 @@ example:
     # Look for $TMT_TEST_DATA/results.yaml (or results.json) with custom results
     result: custom
 
+
```
martinhoyer, line 44: I know, I know.

tmt/result.py
```diff
@@ -310,9 +318,82 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
+                self.result = ResultOutcome.FAIL
+                if self.note:
+                    self.note += ', unexpected pass'
+                else:
+                    self.note = 'unexpected pass'
+            return self
+
+        # Handle individual check results
+        for check in self.check:
+            check_result = ResultInterpret(check.result)
```
happz, line 360: The field should already be `ResultInterpret`, as mentioned in a comment above WRT its definition. And I just noticed that `Test.result` isn't an enum, which is definitely incorrect, and a bad example :/

tmt/checks/__init__.py
```diff
@@ -100,6 +101,11 @@ class Check(
         default=True,
         is_flag=True,
         help='Whether the check is enabled or not.')
+    result: Any = field(
+        default='respect',
+        serialize=lambda result: result.value if isinstance(result, enum.Enum) else str(result),
+        unserialize=lambda spec: spec,
+        help='How to interpret the check result.')
```
happz, line 108: I think we can switch from `Any` to `ResultInterpret`, if we add the `from_spec()` method to it, `unserialize` could use it. Might need also `normalize` callback, but I'd pay that price for getting rid of `Any`. With `Any`, no user of this field can't be sure what the value is.

tmt/result.py
```diff
@@ -310,9 +318,82 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
+                if self.note:
+                    self.note += ', check failed'
+                else:
+                    self.note = 'check failed'
+            return self
+
+        # Handle XFAIL for the overall test result
+        if interpret == ResultInterpret.XFAIL:
+            if self.result == ResultOutcome.PASS and checks_failed:
+                self.result = ResultOutcome.FAIL
```
martinhoyer, line 339: @psss I got a little entangled in the xfail logic and the `# Handle individual check results` part should probably be done before this.   

I wonder if it will be digestible enough for the users, I mean:
 xfail test fails, check with default 'respect' fails = FAIL
 i.e. users would have to add xfail also to check, if expected. Could be breaking change for some?

spec/tests/check.fmf
```diff
@@ -16,6 +16,13 @@ description: |
 
     See :ref:`/plugins/test-checks` for the list of available checks.
 
+    The `result` key can be used to specify how the check result should be
+    interpreted. It accepts three options:
+    - 'respect' (default): The check result affects the overall test result.
+    - 'skip': The check result does not affect the overall test result.
```
psss, line 22: So here we're introducing a new value `skip` which is currently not present in the `Test.result` key specification. What about keeping the consistency here as well and support `pass, info, warn, error, fail` in the very same way?

happz: Hmmm,`skip` is definitely one of the values of https://github.com/teemtee/tmt/blob/main/tmt/result.py#L43, and it's also allowed by schema, https://github.com/teemtee/tmt/blob/main/tmt/schemas/common.yaml#L481. If it's not supposed to be supported as a `result` value, we should remove it from these places as well.

happz: Although, I'd argue the specification should be updated instead, because `skip` is a perfectly valid test outcome, like `pass` or `fail`, and it makes sense that one might want to enforce it as a test outcome via `result`, just like one might want to enforce `pass` or `xfail`.

happz: And +1 to accepting all keys allowed by `ResultInterpret`, i.e. why allowing `skip` and not `xfail` or `pass`. Except `custom`, that one does not make much sense for a test check.

happz: And `restraint`, `restraint` is also useless at this point of code :)

happz: Hm, looks more and more like we need `ChecResultInterpretat` enum & schema, too many exceptions.

tmt/result.py
sbertramrh: Thanks for doing this. I would also like to petition for the following scenario.
if the before test portion is the only part that fails for the check, the test should be `self.result = ResultOutcome.WARN`.

spec/tests/check.fmf
martinhoyer: To be honest, I got pretty frustrated with this code. I guess I'm too dumb for mixing custom mixins, uninheritable classes, specifications, schemas, et al. I'm sure there are good reasons for all of this, but idk, somehow I think doing a relatively small change should be easier.

Anyway, should I create some BaseResultInterpretMixin class w/ SerializableContainer?

I tried to went back to using 'ignore', but now we're mixing two things together, results and how to interpret it. :(

happz: > To be honest, I got pretty frustrated with this code. I guess I'm too dumb for mixing custom mixins, uninheritable classes, specifications, schemas, et al. I'm sure there are good reasons for all of this, but idk, somehow I think doing a relatively small change should be easier.

That's sad to hear. I believe that while it might be a relatively small change, it's not a small change given the interaction of test and test check results. Schema, specification, the fact that enum inheritance is limited, well, that's a fact we can't bypass.

> Anyway, should I create some BaseResultInterpretMixin class w/ SerializableContainer?

I for one don't think so. Where do you think it would be needed?

> I tried to went back to using 'ignore', but now we're mixing two things together, results and how to interpret it. :(

We can get on a call and try to get through the parts that cause you trouble, if you want to.

martinhoyer: > I for one don't think so. Where do you think it would be needed?

It's just that the ResultsInterpret is not exactly [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  
Only thing that came to mind, except for using str literals instead of enums.

> We can get on a call and try to get through the parts that cause you trouble, if you want to.

Thanks, much appreciated.
It seems like we strayed a bit too far from the original goal and while we have check result keys, it's used the same way as test result and I'm not sure if it is the best way to let users define that the check result should be "ignored".

martinhoyer: > I don't want to start rewriting the patch for you, but you obviously do not enjoy the coding at this moment, if it would help, we can sit on it.

Thanks @happz, I'll get to it with a pair of fresh eyes later. Sorry for letting my depression leak to the comments here.

happz: Ok, as you wish. Just a couple of notes:

* something like this could work for propagating how check results should be interpreted - `check_result = CheckResultInterpret(check.result.value)` is definitely not the right way how to get this info.
  ```diff
  -        return _result.interpret_result(invocation.test.result)
  +        return _result.interpret_result(
  +            invocation.test.result,
  +            {
  +                check.how: check.result
  +                for check in invocation.test.check
  +            }
  +        )
 
       def interpret_result(
               self,
  -            interpret: ResultInterpret) -> 'Result':
  +            interpret_test: ResultInterpret,
  +            interpret_checks: dict[str, CheckResultInterpret]) -> 'Result':
  ```
* use a helper function for setting a note, the repetition of `if`/`else` consumes 4 lines for every single note, a helper function would take care of `', ...'` and condition, greatly improving the readability of code.
* this looks like a mistake, check result propagated into result - is that correct? We probably don't want check result to be promoted to main result, instead failed results would turn the main result into a failed one - a skipped check does not mean main result should turn to `SKIP`
   ```python
              elif interpret_check in [CheckResultInterpret.PASS, CheckResultInterpret.FAIL,
                                    CheckResultInterpret.INFO, CheckResultInterpret.WARN,
                                    CheckResultInterpret.ERROR, CheckResultInterpret.SKIP]:
                  self.result = ResultOutcome(check.value)
  ```
* the order should be probably reversed: first we need to sort out check results, because they might be `PASS`, but their `result` key might be set to `fail`, turning them to failed checks - `any(check.result == ResultOutcome.FAIL for check in self.check)` will miss checks that would turn into failed checks because of their interpretation.

tmt/result.py
```diff
@@ -356,119 +356,95 @@ def from_test_invocation(
...
+            check_result = interpret_checks.get(check.name, CheckResultInterpret.RESPECT)
 
             if check_result == CheckResultInterpret.RESPECT:
                 if check.result == ResultOutcome.FAIL:
                     self.result = ResultOutcome.FAIL
-                    if self.note:
-                        self.note += ', check failed'
-                    else:
-                        self.note = 'check failed'
+                    self._set_note('check failed')
```
happz, line 403: This branch does look superfluous to me: `check.result` should be changed, not `self.result` - at this point, we're deciding the fate of a single check result, and the test result is beyond the horizon. The effect of this check result on the test result will be evaluated later. Which means, the outcome of the test did not change in any way, therefore no need for a note -> no need to check for `RESPECT` + `FAIL`.

tmt/result.py
```diff
@@ -356,119 +356,95 @@ def from_test_invocation(
...
             elif check_result in [CheckResultInterpret.PASS, CheckResultInterpret.FAIL,
                                   CheckResultInterpret.INFO, CheckResultInterpret.WARN,
                                   CheckResultInterpret.ERROR, CheckResultInterpret.SKIP]:
-                self.result = ResultOutcome(check_result.value)
-                if self.note:
-                    self.note += f', check result: {check.result.value}'
-                else:
-                    self.note = f'check result: {check.result.value}'
+                # We don't promote check result to main result, but failed checks turn the
+                # main result to FAIL
```
happz, line 419: We don't know that yet: if the test has `result: info`, it does not matter what is the outcome of checks, they should never override the main test result, because the user clearly said "skip" is the desired and final value. We still need to resolve them, to record their final outcomes though, therefore we can't skip check results when `result: skip` is set.

I have a feeling you are getting too entangled in the problem. My advice would be to ignore the other part of the problem: focus on the interpretation of one single check result, and completely ignore what it might do to the test result. Go over all check results, and interpret every single one of them as needed. Ignore whatever their effect might be on the main result. Only after that, take `self.result` and `interpret_test`, collect the final check results, and resolve their interaction. If `test.result` is `SKIP`, `PASS`, etc., set `self.result` and `exit`. If it's `RESPECT`, compute the effect of check results, switch `self.result` to `FAIL` or `WARN`, exit. And so on. We are not in the business of saving CPU cycles, make it correct first, and then we can make it fast by introducing shortcuts. We can afford crude and readable code :)

tmt/result.py
martinhoyer: make sense

tmt/result.py
happz: Maybe even one more: how about ignoring this method for now, and instead constructing a simplet method that would focus on interpretation. Feed it with ResultOutcome, ResultInterpret, map of check name/check result outome and a map of check name/check interpretation, and let it decide the final result, final check reults, and add a parametrized unit test for it for various combinations. Do not feed it with Result or CheckResuly or any structures, just enums and strings. Get that right, get it working, then we can discuss just how this process works instead of worrying sbout other structures. Once we arr happy, translate it to this method.

martinhoyer: Thanks again. I'll get to it tomorrow.  
> Once we arr  

🏴‍☠️

martinhoyer: >  I'll get to it tomorrow.

hmm, looks like a busy weekend. Monday it is.

martinhoyer:
Trying to implement the check results as discussed yesterday. Started from scratch. Still don't understand specs, schemas and docs generation. Why can't we just have a one place to define things at?

tmt/checks/enums.py
```diff
@@ -0,0 +1,28 @@
+import enum
```
happz, line 1: Why keep them in a standalone file? `__init__.py` has about 250+ lines, isn't that small enough?

tmt/result.py
```diff
@@ -189,6 +189,27 @@ class CheckResult(BaseResult):
         default=CheckEvent.BEFORE_TEST,
         serialize=lambda event: event.value,
         unserialize=CheckEvent.from_spec)
+    check_result: CheckResultInterpret = field(
```
happz, line 192: The result class cannot contain a field that governs its own interpretation. That is owned by a different entity: `Test.result` defines how `Result` should be interpreted, similarly `Check.result` defines how `CheckResult` should be interpreted. This field should not exist, check result simply cannot decide itself, it **must** be **told** how to interpret the value.

```python
    def interpret_check_result(self, interpret: CheckResultInterpret) -> 'CheckResult':
        """Interpret check result according to the check_result interpretation."""

        if interpret == CheckResultInterpret.RESPECT:
            return self

        if interpret == CheckResultInterpret.INFO:
            self.result = ResultOutcome.INFO
        elif interpret == CheckResultInterpret.XFAIL:
            self.result = {
                ResultOutcome.FAIL: ResultOutcome.PASS,
                ResultOutcome.PASS: ResultOutcome.FAIL
                }.get(self.result, self.result)

        return self
```

Take a `CheckResult` instance, find the corresponding `Check`, most likely from the `invocation.test.check` field, and pass its `result` to `CheckResult.interpret_check_result()`. Recall the mapping I proposed last week to be put into `Result.interpret_result()`:

```python
{
    check.how: check.result
    for check in invocation.test.check
}
```

It's the same solution, the idea is there: the interpretation **does not** lie in the result itself, it's owned by test (for test results) or test check (for test check results), and must be extracted from them (which is exactly what `Result.from_test_invocation()` does, `return _result.interpret_result(invocation.test.result)`).

tmt/result.py
```diff
@@ -347,6 +368,21 @@ def interpret_result(self, interpret: ResultInterpret) -> 'Result':
...
 
+        # Interpret check results
+        failed_checks: list[CheckResult] = []
+        for check in self.check:
+            interpreted_check = check.interpret_check_result()
+            if (
+                    interpreted_check.result == ResultOutcome.FAIL and
+                    interpreted_check.check_result != CheckResultInterpret.INFO
+                    ):
+                failed_checks.append(interpreted_check)
```
happz, line 379: This will need a bit more work, because `self.check` is left untouched: interpretation happens, but `interpreted_check` is then thrown away, it's not stored anywhere. It's basically swapping one list for another, simple list comprehension. Together with the necessity of accessing checks to get their `result` key, it might be something like the following:

```python
# This is constructed in `from_test_invocation()` and fed to `interpret_result()` - just like the result interpretation:
{
    check.how: check.result
    for check in invocation.test.check
}


def interpret_result(self, interpret: ResultInterpret, interpret_checks: dict[str, CheckResultInterpret]) -> 'Result':
...

self.check = [
    check_result.interpret_check_result(interpret_checks[check_result.name])
    for check_result in self.check
]

failed_checks = [
    check_result
    for check_result in self.check
    if check_result.outcome == ResultOutcome.FAIL]
```

happz:
> Trying to implement the check results as discussed yesterday. Started from scratch. Still don't understand specs, schemas and docs generation. Why can't we just have a one place to define things at?

We can. It somehow started separately, in an organic way, as projects grow, and the amount of work necessary to turn the tide and introduce a single source of truth is not trivial. The work is ongoing, and one possible answer might be that I, who also dislike the currently fractured state of things, am unable to change it all in one night.

tmt/checks/enums.py
martinhoyer: I just needed to deal with that circular import somehow. Would gladly do a less bad solution.

tmt/result.py
martinhoyer: Thanks for the review. Needs work, yes. Though I might need someone to go through the code with me, because most of the time I'm just guessing what is supposed to do what and how and then trying a blind trial-error. Not enjoyable and I'm back to feeling like an useless idiot.

happz: Well, I did offer to have an online session, and the offer still stands. Add something to my calendar and we can try it out.

martinhoyer:
> > Trying to implement the check results as discussed yesterday. Started from scratch. Still don't understand specs, schemas and docs generation. Why can't we just have a one place to define things at?
> 
> We can. It somehow started separately, in an organic way, as projects grow, and the amount of work necessary to turn the tide and introduce a single source of truth is not trivial. The work is ongoing, and one possible answer might be that I, who also dislike the currently fractured state of things, am unable to change it all in one night.

Right, but isn't that a sunk cost fallacy? I'm not saying you should do it alone in one night, but perhaps have a defined "state" to strive for and think about whether it wouldn't be worth working towards it in a long run.

tmt/result.py
martinhoyer: awesome, much appreciated.

happz:
> > > Trying to implement the check results as discussed yesterday. Started from scratch. Still don't understand specs, schemas and docs generation. Why can't we just have a one place to define things at?
> > 
> > 
> > We can. It somehow started separately, in an organic way, as projects grow, and the amount of work necessary to turn the tide and introduce a single source of truth is not trivial. The work is ongoing, and one possible answer might be that I, who also dislike the currently fractured state of things, am unable to change it all in one night.
> 
> Right, but isn't that a sunk cost fallacy? I'm not saying you should do it alone in one night, but perhaps have a defined "state" to strive for and think about whether it wouldn't be worth working towards it in a long run.

Yeah, that was just an example. There is a long-term goal, it just takes so many small steps to get there. Just recently plugin docs began to be generated from plugin sources. It's useful on its own, but it also served as an experiment on what would we need to render plugin schema from plugin sources. And so on, we're moving toward a much simpler state of things, with fewer sources of truth, it's just very slow.

martinhoyer:
Not at all sure about what's happening in execute/internal

tmt/schemas/common.yaml
```diff
@@ -493,15 +493,3 @@ definitions:
     type: string
     # yamllint disable-line rule:line-length
     pattern: "^\\d{2,}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+\\+\\d{2}:\\d{2}$"
-
```
martinhoyer, line 496: Does it make sense to move it to test.yaml, or is check expected to be used elsewhere in future?

sbertramrh: I am hoping we can move the triggering of check to the plan, with the expectation that the check will operate on each test in the plan. But I am not sure if that work has been green lighted yet.

docs/templates/plugins.rst.j2
```diff
@@ -30,8 +30,10 @@
             {% else %}
     Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor %}
             {% endif %}
+        {% elif actual_default.__class__.__name__ == 'CheckResultInterpret' %}
```
martinhoyer, line 33: Ugly workaround.  
@happz Do you have any cleaner solution to point me to please? I spent way more time than I'd like to admit trying to solve this.

happz: Yeah, this is not very nice :/ What fields forced you to take this road? `result` is the new one, but it's a string, like `how`, so I don't immediately recall what might be a problem. What's logged by the script?

martinhoyer:
@psss @happz I've marked the 'outdated' discussions as resolved, as it got quite messy with all the refactoring - sorry. I believe all the points have been incorporated though.

docs/templates/plugins.rst.j2
happz: Nevermind, I can try it out.

martinhoyer: It says `CheckResultInterpret.RESPECT` is the value. I've tried to add `exporter=lambda result: result.value` to Check.result field, but to no avail.

happz: Yep, got it, it's an enum, I suppose it's the first one in plugins, and we can solve it once and for all :)

```diff
diff --git a/docs/scripts/generate-plugins.py b/docs/scripts/generate-plugins.py
index a37d83c1..9368f731 100755
--- a/docs/scripts/generate-plugins.py
+++ b/docs/scripts/generate-plugins.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import dataclasses
+import enum
 import sys
 import textwrap
 from typing import Any
@@ -106,6 +107,12 @@ def container_intrinsic_fields(container: ContainerClass) -> list[str]:
     return field_names
 
 
+def is_enum(value: Any) -> bool:
+    """ Find out whether a given value is an enum member """
+
+    return isinstance(value, enum.Enum)
+
+
 def _create_step_plugin_iterator(registry: tmt.plugins.PluginRegistry[tmt.steps.Method]):
     """ Create iterator over plugins of a given registry """
 
@@ -184,6 +191,7 @@ def main() -> None:
         STEP=step_name,
         PLUGINS=plugin_generator,
         REVIEWED_PLUGINS=REVIEWED_PLUGINS,
+        is_enum=is_enum,
         container_fields=tmt.utils.container_fields,
         container_field=tmt.utils.container_field,
         container_ignored_fields=container_ignored_fields,
diff --git a/docs/templates/plugins.rst.j2 b/docs/templates/plugins.rst.j2
index 5b99149c..5969cc4c 100644
--- a/docs/templates/plugins.rst.j2
+++ b/docs/templates/plugins.rst.j2
@@ -30,7 +30,7 @@
             {% else %}
     Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor %}
             {% endif %}
-        {% elif actual_default.__class__.__name__ == 'CheckResultInterpret' %}
+        {% elif is_enum(actual_default) %}
     Default: ``{{ actual_default.value }}``
         {% else %}
             {% set _ = LOGGER.warn("%s/%s.%s: could not render default value, '%s'" | format(STEP, plugin_id, field_name, actual_default), shift=0) %}
```

martinhoyer: Nice, thanks.

KwisatzHaderach (approved):

tests/execute/result/basic.sh
```diff
@@ -38,7 +38,7 @@ rlJournalStart
         run   "errr"   "/test/error"          ""         2
         run   "pass"   "/test/xfail-fail"     "fail"     0
         run   "fail"   "/test/xfail-pass"     "pass"     1
```
martinhoyer, line 40: Are we ok with this? I see no reason to add note when original and actual results are the same, no?

tmt/result.py
```diff
@@ -317,11 +334,18 @@ def from_test_invocation(
             ids=ids,
             log=log or [],
             guest=ResultGuestData.from_test_invocation(invocation=invocation),
-            data_path=invocation.relative_test_data_path)
+            data_path=invocation.relative_test_data_path,
+            check=invocation.check_results or [])
 
-        return _result.interpret_result(invocation.test.result)
+        interpret_checks = {check.how: check.result for check in invocation.test.check}
```
martinhoyer, line 340: This way doesn't really work when multiple check.how/check.name are the same. There surely has to be a better way to add/match `CheckResultInterpret` to each `Check`?

tmt/result.py
```diff
@@ -187,7 +187,24 @@ class CheckResult(BaseResult):
...
+
+    def interpret_check_result(self, interpret: CheckResultInterpret) -> 'CheckResult':
+        """Interpret check result according to the check_result interpretation."""
+
+        if interpret == CheckResultInterpret.RESPECT:
+            return self
+        if interpret == CheckResultInterpret.INFO:
+            self.result = ResultOutcome.INFO
+
+        elif interpret == CheckResultInterpret.XFAIL and self.event != CheckEvent.BEFORE_TEST:
```
martinhoyer, line 201: I was thinking about adding a logic that would check for both before and after and make the xfail 'pass' only if after or both are failed, but wasn't sure if it's needed. ?

martinhoyer: On second thought, this does need to be improved.  
So, what are the expectations?
before fails > pass
before pass > pass
after fails > pass
after fails > fail
?

tmt/result.py
happz: I believe we spoke about a way: there is a trivial mapping between a result and its parent test, `TestInvocation` stores results & points to the test. We did not needed to transition from check result to a check, therefore there is no such mapping yet, we need to establish one. My proposal was to add a dedicated mapping to `TestInvocation` class which would map from `CheckResult` to `Check` instances, and it would be populated in `_run_checks_for_test()` (most likely), as that is the place where check results are born. Something like this:

```diff
diff --git a/tmt/steps/execute/__init__.py b/tmt/steps/execute/__init__.py
index 407b4ac0..7f77d179 100644
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -16,6 +16,7 @@ import fmf.utils
 
 import tmt
 import tmt.base
+import tmt.checks
 import tmt.log
 import tmt.steps
 import tmt.utils
@@ -175,6 +176,9 @@ class TestInvocation:
     results: list[Result] = dataclasses.field(default_factory=list)
     check_results: list[CheckResult] = dataclasses.field(default_factory=list)
 
+    check_result_to_check: dict[CheckResult, tmt.checks.Check] = dataclasses.field(
+        default_factory=dict)
+
     check_data: dict[str, Any] = field(default_factory=dict)
 
     return_code: Optional[int] = None
@@ -954,6 +958,8 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
                 result.end_time = format_timestamp(timer.end_time)
                 result.duration = format_duration(timer.duration)
 
+                invocation.check_result_to_check[result] = check
+
             results += check_results
 
         return results
```

It's not clean, it's not nice, I'd like to make it look and feel like the test invocation, but I don't have that ready and I promise to polish it later. With this, the following should work: use this mapping to collect check result/interpretation pairs, and use the right one for the given check result when calling the interpretation method:

```python
interpret_checks = {
    check_result: check.result for check_result, check in invocation.check_result_to_check.items()
}

...

check_result.interpret_check_result(interpret_checks[check_result]))
```

tmt/checks/__init__.py
```diff
@@ -83,6 +84,19 @@ def from_spec(cls, spec: str) -> 'CheckEvent':
...
+    INFO = 'info'
+    RESPECT = 'respect'
+    XFAIL = 'xfail'
+
+    @classmethod
+    def from_spec(cls, spec: str) -> 'CheckResultInterpret':
+        try:
+            return CheckResultInterpret(spec)
+        except ValueError:
+            raise ValueError(f"Invalid check result interpretation '{spec}'.")
```
seberm, line 97: Does it make sense to use `raise ... from ...` here?

```
except ValueError as error:
    raise ValueError(f"Invalid check result interpretation '{spec}'.") from error
```

tmt/checks/__init__.py
```diff
@@ -100,6 +114,12 @@ class Check(
         default=True,
         is_flag=True,
         help='Whether the check is enabled or not.')
+    result: CheckResultInterpret = field(
+        default=CheckResultInterpret.RESPECT,
+        help='How to interpret the check result.',
+        serialize=lambda result: result.value,
+        unserialize=CheckResultInterpret.from_spec
+        )
```
seberm, line 122: Just a small nitpick, move the `)` right after `unserialize=CheckResultInterpret.from_spec` to avoid an unnecessary line.

tmt/result.py
```diff
@@ -187,7 +187,24 @@ class CheckResult(BaseResult):
     event: CheckEvent = field(
         default=CheckEvent.BEFORE_TEST,
         serialize=lambda event: event.value,
-        unserialize=CheckEvent.from_spec)
+        unserialize=CheckEvent.from_spec
+        )
+
+    def interpret_check_result(self, interpret: CheckResultInterpret) -> 'CheckResult':
+        """Interpret check result according to the check_result interpretation."""
```
seberm, line 194: The docstring format should probably be changed to:

```
""" Interpret check result according to the check_result interpretation """
```

or even to:

```
""" Interpret check result according to the ``check_result`` interpretation """
```

seberm (commented):
The parts I understood seem fine to me. Great work!

seberm (approved):

tmt/result.py
happz: There's no ``check_result`` in the docstring or signature :) Maybe "check result" would be better?

tmt/checks/__init__.py
happz: Not necessarily, since `spec` is shown, it should be fairly clear why it failed the validation.

tmt/result.py
happz: No idea, TBH. Should there be a difference between before/after treatment? Current checks do some preparation in their `before` execution, it would be hard for them to fail in `before` but pass in `after`. It seems to me the `after` should have more effect on the result outcome, it's usually where the test happens, although I wouldn't let `fail` in `before` slip as `pass`. Maybe `warning`?

@psss ^^

tmt/checks/__init__.py
martinhoyer: Good discussion point though. It's not generally being used here, the [B904](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/#raise-without-from-inside-except-b904) is marked in 'ignore'.  

So, if we want to adhere to this rule, there could either be a pull request with enabled check and modified all the existing violations ... or the existing exceptions could get `noqa` so the rule can be removed from ignore list and any new code would comply.

tmt/result.py
martinhoyer: oh, now I remembered I tried to go this way, but `CheckResult` is unhashable. And I can't really add `frozen=True` to it to make it so. Am I missing something @happz ?

happz: Ouch, I forgot about that, it's not immutable :/

Hm, how about going with what you have already, leaving the "multiple checks with the same `how`" case for later? Because if I'm not mistaken, this case is broken in more places than here. For example, checks fill `CheckResult.name` with their name, e.g. https://github.com/teemtee/tmt/blob/main/tmt/checks/dmesg.py#L199, therefore one check running twice will produce two results with the same attributes. So, more work to make this feature correctly supported.

For now, I'd say "running one check multiple times for the same test is not supported yet" and move on to finish the PR, and we would make the code right later.

martinhoyer: Sounds good.

tmt/result.py
martinhoyer: changed to "check_result" in https://github.com/teemtee/tmt/pull/3239/commits/345bde830384b610d0d4bccd395b7f0a00aa5b26

tmt/checks/__init__.py
martinhoyer: addressed in https://github.com/teemtee/tmt/pull/3239/commits/345bde830384b610d0d4bccd395b7f0a00aa5b26

tmt/checks/__init__.py
martinhoyer: Added in https://github.com/teemtee/tmt/pull/3239/commits/345bde830384b610d0d4bccd395b7f0a00aa5b26 
Created https://github.com/teemtee/tmt/issues/3308 to not forget about this.

tests/execute/result/check_results/test.fmf
```diff
@@ -0,0 +1,84 @@
...
+    framework: shell
+    duration: 1m
+    check:
+        - how: dmesg
+          result: respect
+    # Expected outcome: PASS (test passes, check passes)
+
+/check-fail-respect:
+    summary: Test with failing dmesg check (respect)
+    test: echo "Fail Test Check Pattern" | sudo tee /dev/kmsg
```
psss, line 16: If the test needs `root` for their execution I would suggest to mark `/tests/execute/result/check` with `provision-only` and `provision-local` tags (optionally with `provision-virtual` as well) to make sure it gets the permissions and is not executed as part of the regular plans.

tests/execute/result/check_results/test.fmf
```diff
@@ -0,0 +1,84 @@
...
+
+/check-pass:
+    summary: Test with passing checks
+    test: echo "Test passed"
+    framework: shell
+    duration: 1m
+    check:
+        - how: dmesg
+          result: respect
+    # Expected outcome: PASS (test passes, check passes)
```
psss, line 12: I would say this content is an excellent candidate for `description`. With `tmt show` in the test directory you would quickly learn all necessary info.

psss (commented):
Added a couple of comments.

tmt/result.py
martinhoyer: my bad, I forgot the underscore.

docs/templates/plugins.rst.j2
```diff
@@ -30,8 +30,10 @@
             {% else %}
     Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor %}
             {% endif %}
+        {% elif is_enum(actual_default) %}
+    Default: ``{{ actual_default.value }}``
```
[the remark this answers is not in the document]
martinhoyer: I've tried to add 
```
    {% elif is_enum(metadata.type) %}
{{ option }}: ``{{ metadata.type.__members__.keys() | join('|') }}``
```
Not sure what's up.

Btw, I notice a typo below, `suboptional` -> `suboptimal` @happz ?

tests/execute/result/check_results/test.fmf
martinhoyer: Thanks. I was gonna ask about the root thing - Isn't there a better way to trigger a check failure that doesn't require root?

docs/templates/plugins.rst.j2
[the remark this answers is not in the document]
happz: Oh yeah, that looks like a typo.

The template addition looks good to me, it didn't work? `member.value for member in ResultOutcome.__members__.values()` would be probably better, in the `PASS = 'pass'` we want the `'pass'` string, which is the `value` of a member.

spec/tests/check.fmf
```diff
@@ -13,6 +13,13 @@ description: |
     the user can run into. Another useful checks would be kernel
     panic detection, core dump collection or collection of system
     logs.
+    By default, the check results affect the overall test outcome.
```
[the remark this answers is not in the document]
martinhoyer: Done in https://github.com/teemtee/tmt/pull/3239/commits/c7ba24e419cab584e7c3af6869b878efb4e82b23

tests/execute/result/check_results.sh
```diff
@@ -0,0 +1,53 @@
...
+    rlAssertGrep "${res} ${tn}" $rlRun_LOG
+
+    echo
+}
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        rlRun "pushd check_results"
+        rlRun "set -o pipefail"
```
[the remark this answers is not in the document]
martinhoyer: Thanks, copy-paste leftover, changed in https://github.com/teemtee/tmt/pull/3239/commits/c7ba24e419cab584e7c3af6869b878efb4e82b23

tests/execute/result/check_results.sh
```diff
@@ -0,0 +1,53 @@
...
+        run "pass" "/test/check-pass" 0
+        run "fail" "/test/check-fail-respect" 1
+        run "pass" "/test/check-fail-info" 0
+        run "fail" "/test/check-xfail-pass" 1
+        run "pass" "/test/check-xfail-fail" 0
+        run "pass" "/test/check-override" 0
+    rlPhaseEnd
+
+    rlPhaseStartTest "Verbose execute prints result"
+        rlRun -s "tmt run --id \${run} --scratch --until execute tests --filter tag:-cherry_pick provision --how local execute -v 2>&1 >/dev/null" "1"
```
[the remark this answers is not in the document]
martinhoyer: Thanks, copy-paste leftover, changed in https://github.com/teemtee/tmt/pull/3239/commits/c7ba24e419cab584e7c3af6869b878efb4e82b23

tests/execute/result/check_results/test.fmf
```diff
@@ -0,0 +1,84 @@
+summary: Tests for check results behaviour
+description: Verify that check results are correctly interpreted and affect test results
+
+/check-pass:
+    summary: Test with passing checks
+    test: echo "Test passed"
+    framework: shell
+    duration: 1m
```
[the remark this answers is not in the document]
martinhoyer: Thanks, done in https://github.com/teemtee/tmt/pull/3239/commits/c7ba24e419cab584e7c3af6869b878efb4e82b23

tests/execute/result/check_results/test.fmf
martinhoyer: Done in https://github.com/teemtee/tmt/pull/3239/commits/e873871d7fd0bfd9294e3f7ce1daa31a1e232491  
I've also used `enabled` instead of commenting-out the 'multiple' test. Make sense?

tmt/result.py
martinhoyer: I've tried to find a way to aggregate the before and after check results, but it feels a bit ugly and wrong:
```
diff --git a/tmt/result.py b/tmt/result.py
index 94c43443..19d67c2d 100644
--- a/tmt/result.py
+++ b/tmt/result.py
@@ -197,8 +197,8 @@ class CheckResult(BaseResult):
             return self
         if interpret == CheckResultInterpret.INFO:
             self.result = ResultOutcome.INFO
-
-        elif interpret == CheckResultInterpret.XFAIL and self.event != CheckEvent.BEFORE_TEST:
+        # Note that xfail applies to both BEFORE_TEST and AFTER_TEST
+        elif interpret == CheckResultInterpret.XFAIL:
             self.result = {
                 ResultOutcome.FAIL: ResultOutcome.PASS,
                 ResultOutcome.PASS: ResultOutcome.FAIL
@@ -367,10 +367,38 @@ class Result(BaseResult):
         original_result = self.result
 
         # Interpret check results
-        self.check = [
-            check_result.interpret_check_result(CheckResultInterpret(interpret_checks[check_result.name]))
-            for check_result in self.check
-            ]
+        interpreted_results: list[CheckResult] = []
+        xfail_failures = {
+            CheckEvent.BEFORE_TEST: False,
+            CheckEvent.AFTER_TEST: False
+        }
+
+        # First pass: interpret results and track XFAIL failures
+        for check_result in self.check:
+            interpretation_type = interpret_checks[check_result.name]
+            interpreter = CheckResultInterpret(interpretation_type)
+            interpreted_result = check_result.interpret_check_result(interpreter)
+
+            # Track if BEFORE/AFTER test XFAILs have failed
+            if (interpreter == CheckResultInterpret.XFAIL and
+                interpreted_result.result == ResultOutcome.FAIL and
+                check_result.event in xfail_failures):
+                xfail_failures[check_result.event] = True
+
+            interpreted_results.append(interpreted_result)
+
+        # Second pass: We want to make both xfail before and after check results pass
+        # if either or both failed
+        if any(xfail_failures.values()):
+            for _, (check_result, interpreted_result) in enumerate(
+                zip(self.check, interpreted_results)
+            ):
+                if (interpret_checks[check_result.name] == CheckResultInterpret.XFAIL and
+                    check_result.event in xfail_failures and
+                    interpreted_result.result == ResultOutcome.FAIL):
+                    interpreted_result.result = ResultOutcome.PASS
+
+        self.check = interpreted_results
 
         if interpret not in (ResultInterpret.RESPECT, ResultInterpret.XFAIL):
             self.result = ResultOutcome(interpret.value)
```

Would it make sense to only allow "xfail" for after test check? i.e. what's in my previous comment.
@happz @psss

happz: * `interpretation_type` should already be `CheckResultInterpret`, it should not be necessary to perform the `CheckResultInterpret(interpretation_type)` call
* I don't understand `enumerate()` when the index is thrown away via `_`

The rest, well, it's not pretty, but I'm not sure if it even *can be* pretty. It's complicated stuff, needs to trickle results through several layers. Maybe we could simplify with some functional magic, `reduce`, `map`, `any` & `all`, but I would prefer being more verbose here and now, so we all understand what's happening. We can always simplify things later if needed. LGTM as it is.

martinhoyer: > yeah, and I wouldn't ignore the before event, i.e. if it fails, the check fails in general.  

It wouldn't be ignored. Only when interpret is xfail and original result of before_test check is not "FAIL", we could convert it to warn, or info.

tests/execute/result/check_results.sh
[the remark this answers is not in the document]
martinhoyer: hmm, so it was actually needed, took me some time to realize.

tests/execute/result/main.fmf
```diff
@@ -10,3 +10,6 @@
 /special:
     summary: Test special characters generated to tmt-report-results.yaml
     test: ./special.sh
+/check_results:
```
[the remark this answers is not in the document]
martinhoyer: done

docs/templates/plugins.rst.j2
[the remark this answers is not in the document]
martinhoyer: @happz I thought it would be good to use `choices` here, but wasn't sure how to make it scalable for all enums:  https://github.com/teemtee/tmt/pull/3239/commits/f197ee035b395a74e60f60195bc1d26714931b0d

docs/releases.rst
```diff
@@ -25,6 +24,23 @@ tests from remote repositories where the user does not have write
...
+result of each check affects the overall test result. The following values
+are supported:
+
+respect (default)
+    The check result is respected and affects the overall test result.
+xfail
+    The check result is expected to fail (pass becomes fail and vice-versa).
+info
+    The check result is always treated as "INFO" and does not affect the
+    overall test result.
```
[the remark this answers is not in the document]
martinhoyer: done

tmt/result.py
martinhoyer: oh, after actually thinking about it - I should also drop the quotes, eh? :)

tmt/result.py
martinhoyer: @happz I've really tried to make the logic work, but in the end, I'd very much prefer to have any potential smarter way to do this once we have support for multiple checks per test, as it eventually breaks down having only `check.how` to go with.

docs/releases.rst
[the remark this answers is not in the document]
psss: Thanks! Proposing just a minor wording adjustment in 6d85a26e.

tests/execute/result/main.fmf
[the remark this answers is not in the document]
psss: Thanks!

tests/execute/result/check_results.sh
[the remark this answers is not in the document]
psss: Was it? Why? This is not clear to me.

tests/execute/result/check_results/test.fmf
psss: I'm not aware of any. Plus, in order to actually execute the check (either avc or dmesg) root is still necessary, so I guess we do not have any other choice. And if we're under root we can drop the `sudo`, right?

tests/execute/result/check_results/test.fmf
psss: Yeah, `enabled` definitely makes sense here. Thanks!

docs/templates/plugins.rst.j2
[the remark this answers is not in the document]
psss: Works nice, thanks!

psss (changes requested):
Overall looks good! Thanks for working on this. Added a few more comments.

psss:
Seems there are several tests failing with the following error:

    cannot represent an object: CheckResultInterpret.RESPECT

tmt/result.py
happz: Yep :)

tmt/result.py
happz: Agreed.

tests/execute/result/check.sh
```diff
@@ -0,0 +1,59 @@
...
+00:00:00 pass /test/check-override (on default-0) [3/6]
+00:00:00 pass /test/check-pass (on default-0) [4/6]
+00:00:00 pass /test/check-xfail-fail (on default-0) [5/6]
+00:00:00 fail /test/check-xfail-pass (on default-0) (Check 'dmesg' failed, original result: pass) [6/6]
+EOF
+    rlPhaseEnd
+
+    rlPhaseStartTest "Verify before test check xfail fail is interpreted as pass"
+        rlRun "echo 'Fail Test Check Pattern' | sudo tee /dev/kmsg"
+        rlRun -s "tmt run -a --scratch test --name /test/check-xfail-fail provision --how local report -v 2>&1 >/dev/null | grep report -A3"
```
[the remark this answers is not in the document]
martinhoyer: Yes, just the before test, as the previous tests cover the "after". But since you brought it up, I guess we can also expand it to both and make sure the tests returns 0.  
I haven't yet specifically checked for "WARN", as I didn't know if you guys are ok with the approach.

tests/execute/result/check.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+run()
+{
+    res=$1  # expected final result of test
+    tn=$2   # test name
+    ret=$3  # tmt return code
+
+    rlRun -s "tmt run -a --scratch --id \${run} test --name ${tn} provision --how local report -v 2>&1 >/dev/null | grep report -A2 | tail -n 1" \
```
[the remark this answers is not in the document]
martinhoyer: I see, I used execute/result/basic as a 'template'.  Should it be changed there as well?

tests/execute/result/check/test.fmf
```diff
@@ -0,0 +1,77 @@
...
+    description: "Expected outcome: PASS (test passes, check fails but xfail expects it to fail)"
+    test: echo "Fail Test Check Pattern" | sudo tee /dev/kmsg
+    check:
+        - how: dmesg
+          failure-pattern: Fail Test Check Pattern
+          result: xfail
+
+/check-multiple:
+    summary: Test with multiple checks with different result interpretations
+    description: "Expected outcome: FAIL (first dmesg check fails and is respected, second dmesg check passes but xfail expects it to fail, third failing dmesg check is just info)"
```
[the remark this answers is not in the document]
martinhoyer: Just in this long one, or all? By dropping `:` char, you mean after the "Expected outcome"?

tests/execute/result/check_results.sh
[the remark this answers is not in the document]
martinhoyer: It doesn't use tmt exit codes, so the runs that should fail..."fail" :)
```
[   FAIL   ] :: Result: fail, Test name: /test/check-fail-respect, tmt return code: 1 (Expected 1, got 0)
```

psss:
I believe the problem was caused by missing conversion of the new field to and from the spec. Should be fixed by b2cd608b. @happz, could you please have a look as well?

tmt/result.py
```diff
@@ -187,7 +187,27 @@ class CheckResult(BaseResult):
...
+        """Interpret check result according to the "check result" interpretation."""
+
+        if interpret == CheckResultInterpret.RESPECT:
+            return self
+        if interpret == CheckResultInterpret.INFO:
+            self.result = ResultOutcome.INFO
+        elif interpret == CheckResultInterpret.XFAIL:
+            mapping = (
+                # if 'xfail' BEFORE_TEST check passed, use WARN, otherwise switch PASS<->FAIL
+                {ResultOutcome.FAIL: ResultOutcome.PASS, ResultOutcome.PASS: ResultOutcome.WARN}
```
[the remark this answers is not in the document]
martinhoyer: It's this conversation: https://github.com/teemtee/tmt/pull/3239#discussion_r1812519058. Could be INFO as well, your call.

tests/execute/result/check/test.fmf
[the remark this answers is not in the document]
psss: Ah, I thought that `:` always makes `yaml` to think it's defining a new key. But with `|` it just works. So I'd suggest to wrap all descriptions.

tmt/result.py
```diff
@@ -317,11 +337,18 @@ def from_test_invocation(
...
-        return _result.interpret_result(invocation.test.result)
+        interpret_checks = {check.how: check.result for check in invocation.test.check}
 
-    def interpret_result(self, interpret: ResultInterpret) -> 'Result':
+        return _result.interpret_result(invocation.test.result, interpret_checks)
+
+    def interpret_result(
+            self,
+            interpret: ResultInterpret,
+            interpret_checks: dict[str, CheckResultInterpret]
```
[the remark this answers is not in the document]
psss: Thanks!

tmt/result.py
psss: Seems to be covered.

tests/execute/result/check/test.fmf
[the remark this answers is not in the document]
martinhoyer: done

tmt/result.py
```diff
@@ -317,49 +337,77 @@ def from_test_invocation(
...
             return self
 
-        # Extend existing note or set a new one
-        if self.note:
-            self.note += f', original result: {self.result.value}'
+        original_result = self.result
+
+        # Interpret check results
+        self.check = [
+            check_result.interpret_check_result(CheckResultInterpret(interpret_checks[check_result.name]))
```
psss, line 375:
```suggestion
            check_result.interpret_check_result(interpret_checks[check_result.name])
```

The `interpret_checks` values are already `CheckResultInterpret` objects:

```python
(Pdb) pp interpret_checks
{'avc': <CheckResultInterpret.XFAIL: 'xfail'>,
 'dmesg': <CheckResultInterpret.RESPECT: 'respect'>}

(Pdb) interpret_checks[self.check[0].name]
<CheckResultInterpret.RESPECT: 'respect'>
```

So we don't need to instantiate them again, right?

tmt/result.py
```diff
@@ -317,49 +337,77 @@ def from_test_invocation(
...
+
+        failed_checks = [
+            check_result
+            for check_result in self.check
+            if check_result.result == ResultOutcome.FAIL
+            ]
+
+        if failed_checks:
+            self.result = ResultOutcome.FAIL
+            check_note = ", ".join([f"Check '{check.name}' failed" for check in failed_checks])
```
psss, line 397:
```suggestion
            check_note = ", ".join([f"check '{check.name}' failed" for check in failed_checks])
```

A nitpick to make it look good when multiple checks fail. Now we get:

    Check 'avc' failed, Check 'dmesg' failed

tests/execute/result/check.sh
```diff
@@ -0,0 +1,44 @@
...
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-virtual}"
+        rlRun "pushd check"
+        rlRun "set -o pipefail"
+        # Write pattern for tests that need pre-existing dmesg content
+        rlRun "echo 'Fail Test Check Pattern' | sudo tee /dev/kmsg"
```
psss, line 11: Actual check tests are run on a guest provisioned on line 15, by default `virtual` provision is used, so this line will have no effect.

tmt/result.py
psss: Let's explicitly mention this in the docstring as well. It was a bit confusing to understand.

tmt/result.py
```diff
@@ -317,49 +337,77 @@ def from_test_invocation(
...
+                f"Invalid result '{interpret.value}' in test '{self.name}'."
+                )
+
+        if interpret == ResultInterpret.CUSTOM:
             return self
 
-        # Extend existing note or set a new one
-        if self.note:
-            self.note += f', original result: {self.result.value}'
+        original_result = self.result
```
psss, line 371: Seems that `self.original_result` already contains what we need, so it should not be necessary to introduce a new variable here.

tmt/result.py
happz: > The interpret_checks values are already CheckResultInterpret objects:

Correct, it's exactly as declared in the method signature.

> Let's explicitly mention this in the docstring as well. It was a bit confusing to understand.

I wouldn't :( I know it sounds tempting, but I feel it a sign of resignation :) The variable is a simple mapping, IDEs, and Sphinx should be able to extract and display this information from the type of the parameter, and, more importantly, at least to me, there should be not even the slightest shadow of a doubt about what type of values in `interpret_checks` would be.

It's the plague of Python projects, this endless second-guessing. That's why it's so important to establish such a type once and for all, on boundaries between tmt and the outside world, and to fight with sword and flame against any ambiguity, to never ever allow the "check result interpretation" values to be anything else but that enum. There must be no question about it on our minds, as soon as you see a name like `interpret_checks`, never should the question "hmm, what are the values, strings maybe?" appear. I for one am not very fond of explicitly pointing out a singular type in docstrings, because it always feels like resignation on this core principle and return of doubt and second-guessing :(

tmt/result.py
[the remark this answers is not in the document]
psss: A different approach discussed, closing this one.

tmt/result.py
psss: Ah! Right you are, in your fight, of course! For some reason I had my blind spot on exactly this line:

```python
            interpret_checks: dict[str, CheckResultInterpret]
```

Don't know why, don't know how it happened. Now I see it, and the world is in order again :-)

tests/unit/test_results.py
```diff
@@ -76,3 +86,272 @@
 def test_result_to_exit_code(outcomes: list[ResultOutcome], expected_exit_code: int) -> None:
     assert results_to_exit_code([MagicMock(result=outcome) for outcome in outcomes]) \
         == expected_exit_code
+
+
+@pytest.mark.parametrize(
+    ('check_results', 'interpret', 'expected_outcome'),
+    [
+        # Empty results default to PASS
```
martinhoyer, line 94: Is this expected btw?

psss: Yeah, this sounds good to me.

spec/tests/check.fmf
```diff
@@ -14,6 +14,27 @@ description: |
...
+
+        xfail
+            The check result is expected to fail (pass becomes
+            fail and vice-versa).
+
+        info
+            The check result is treated as an informational
+            message and does not affect the overall test result.
+
+    Note that running one check multiple times for the same test is
```
thrix, line 33: @psss Could we use make this more visible via a warning or note pls?

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

tmt/result.py
```diff
@@ -324,49 +349,116 @@ def from_test_invocation(
...
+
+    def interpret_check_result(
+            self,
+            check_name: str,
+            interpret_checks: dict[str, CheckResultInterpret]) -> ResultOutcome:
+        """
+        Aggregate all checks of given name and interpret the outcome
+
+        :param check_name: name of the check to be aggregated
+        :param interpret_checks: mapping of check:how and it's result interpret
```
thrix, line 374:
```suggestion
        :param interpret_checks: mapping of check:how and its interpreted result

KwisatzHaderach (approved):

thrix (approved):
LGTM

spec/tests/check.fmf
martinhoyer: Right, we discussed this yesterday @psss - would it be hard to add the detection and warning during discovery?

psss: Ack, fixed in cd0527d4.

tmt/result.py
psss: Fixed in cd0527d4.

seberm (approved):
LGTM. Great work Martin!

PR #2922, opened 2024-05-09
Add `link` command and linking to jira on new issues

tkoscieln:
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] extend the test coverage
* [x] mention the version
* [x] include a release note

PR #3240, opened 2024-09-26
Drop bootstrap logger requirement from module importer

happz:
It is not needed during import time, it is needed in runtime and importing code can provide logger when needed.

Pull Request Checklist

* [x] implement the feature


psss (approved):
Thanks!

psss:
Failures cause by some irrelevant network errors. Merging to unblock #2922.

PR #3248, opened 2024-09-30
Avoid using raw step data when injecting phases

happz:
Actual step data containers come with type annotations and better checking.

Pull Request Checklist

* [x] implement the feature


therazix (approved):

psss (approved):
Thanks for simplifying this!

PR #3252, opened 2024-10-01
Make the pidfile accessible by everyone

### LINKED ISSUE #2877: Only first user can run provision -h local. Other lack lock file permissions.
As a root create two users and execute tmt run for each of them in sequence:
```
# useradd A
# su -l A
$ tmt init -t mini
$ tmt run -a provision -h local
....
total: 1 test passed

$ logout

# useradd B
# su -l B
$ tmt init -t mini
$ tmt run -a provision -h local
...
warn: Test failed to manage its pidfile.
...
total: 1 error
```

Debug output shows
```
            errr /script-00 (pidfile locking)
                output.txt: /var/tmp/tmt/run-007/plans/example/execute/data/guest/default-0/script-00-1/output.txt
        Workdir '/var/tmp/tmt/run-007/plans/example/report/default-0' created.
        Read file '/var/tmp/tmt/run-007/plans/example/execute/data/guest/default-0/script-00-1/output.txt'.
                content: flock: cannot open lock file /var/tmp/tmt-test.pid.lock: Permission denied
```

mcasquer:
Creates the pidfile at an accessible location
for every user, this way the manage pidfile
warning is avoided.

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] mention the version
* [ ] include a release note


mcasquer:
To discuss here which of the added `# noqa` could be avoided

tmt/steps/execute/internal.py
```diff
@@ -34,7 +34,12 @@
 TEST_PIDFILE_LOCK_FILENAME = f'{TEST_PIDFILE_FILENAME}.lock'
 
 #: The default directory for storing test pid file.
-TEST_PIDFILE_ROOT = Path('/var/tmp')  # noqa: S108 insecure usage of temporary dir
+pid_directory = '/var/tmp/pid'  # noqa: S108 insecure usage of temporary dir
+Path(pid_directory).mkdir(parents=True, exist_ok=True)
+user_id = os.system("id -u")  # noqa: S605 Starting a process with a shell
+if user_id == os.stat(pid_directory).st_uid:
+    os.chmod(pid_directory, 0o777)  # noqa: S103 setting a permissive mask on file or directory
+TEST_PIDFILE_ROOT = Path(pid_directory)
```
happz, line 42: First, all this code needs to happen outside of the global, module-level scope - we do not want `tmt.steps.execute.internaluser_id` to exist, for example. We try to avoid import-time side effects, stuff happening when only importing tmt.

But there is a bigger problem, all this creates the directory on the machine running tmt, but that's the wrong place, it must exist (and have the right permissions) on the guest on which tests are running. Check the `setup()` method of `tmt.steps.provision.Guest` - this one runs to prepare the guest for testing, and I dare to say this would be the right place to extend, to create the directory on the guest.

happz: Pidfile is definitely created on the guest, see https://github.com/teemtee/tmt/blob/main/tmt/steps/execute/internal.py#L71. It is created by a shell wrapper `execute/internal` runs over SSH on the guest, and the wrapper creates the pidfile and then runs the test. It exists so scripts like `tmt-reboot` - which do run on the guest - could interrupt running test.

mcasquer: @happz so what about adding something similar to:

```python
pid_directory = '/var/tmp/pid'
Path(pid_directory).mkdir(parents=True, exist_ok=True)
self.execute(ShellScript(f"chmod ugo+rwx {pid_directory}"))
```
At the end of the setup function?
https://github.com/teemtee/tmt/blob/f9642cf4374c9d2cea1334c777db37c0d6a6ecfa/tmt/steps/provision/__init__.py#L1689
Ps: I still see the warn message when trying with the second user :confused:

happz: Yes, I would try putting it into the `Guest.setup()` method.

mcasquer: @happz with the following suggestion I still see the `pid` directory is created in the host running tmt

happz: Correct, because `Path(pid_directory).mkdir(parents=True, exist_ok=True)` is still running locally, on the machine running tmt. `self.execute(ShellScript(f"mkdir -p {pid_directory}"))` would translate this to the guest.

mcasquer: Sorry for my obvious question... so now what happens with the TEST_PIDFILE_ROOT defined int the internal.py?

mcasquer: Maybe I am misunderstanding something but that directory is being created in the TMT host

happz: Well, nothing happens to `TEST_PIDFILE_ROOT` :) It stays where it is, and what happens *to your patch* is that instead of having your own path in `pid_directory`, your code should use `tmt.steps.execute.internal.TEST_PIDFILE_ROOT` instead. That "constant" is correct, and should be used.

tmt/steps/provision/__init__.py
```diff
@@ -885,6 +885,9 @@ def setup(self) -> None:
 
         Setup the guest after it has been started. It is called after :py:meth:`Guest.start`.
         """
+        pid_directory = '/var/tmp/pid'  # noqa: S108 insecure usage of temporary dir
```
happz, line 888: This line shouldn't exist, it duplicates the already defined `TEST_PIDFILE_ROOT` constant and `effective_pidfile_root()` function that includes possible user change. Both commands below should use `effective_pidfile_root()` instead.

tmt/steps/provision/__init__.py
```diff
@@ -45,6 +45,7 @@
 from tmt.package_managers import FileSystemPath, Package, PackageManagerClass
 from tmt.plugins import PluginRegistry
 from tmt.steps import Action, ActionTask, PhaseQueue
+from tmt.steps.execute.internal import effective_pidfile_root
```
happz, line 48: This seems to create a circular imports, I recommend moving it into the `setup()` method instead.

mcasquer: Done, thanks !

tmt/steps/provision/__init__.py
```diff
@@ -912,6 +912,10 @@ def setup(self) -> None:
 
         Setup the guest after it has been started. It is called after :py:meth:`Guest.start`.
         """
+        from tmt.steps.execute.internal import effective_pidfile_root
+        pid_directory = effective_pidfile_root()
+        self.execute(ShellScript(f"mkdir -p {pid_directory}"))
+        self.execute(ShellScript(f"chmod ugo+rwx {pid_directory}"))
```
happz, line 918: Ha, one more thing I just realized: we can run both commands in one step, safe one needless SSH connection: `self.execute(ShellScript(f"mkdir -p {pid_directory} && chmod ugo+rwx {pid_directory}"))`

mcasquer: Updated, thanks !

happz (approved):

tmt/steps/execute/internal.py
```diff
@@ -34,7 +34,7 @@
 TEST_PIDFILE_LOCK_FILENAME = f'{TEST_PIDFILE_FILENAME}.lock'
 
 #: The default directory for storing test pid file.
-TEST_PIDFILE_ROOT = Path('/var/tmp')  # noqa: S108 insecure usage of temporary dir
+TEST_PIDFILE_ROOT = Path('/var/tmp/pid')  # noqa: S108 insecure usage of temporary dir
```
psss, line 37: What about using `/var/tmp/tmt/pid` instead to have all `tmt` releated files under one directory?

happz:
> @happz one question, in the [related issue ](https://github.com/teemtee/tmt/issues/2877)with the patch it's used `tmt -a provision -h local` which IIUC it's the local host already executing tmt, in this case the code raises again permissions issues when doing the chmod

Is your change here causing a regression? Or is your change here just making no difference WRT #2877?

psss:
Yeah, the thing is that for the `local` provision use case there is no improvement. The second user ends up with the following:

```
Command 'mkdir -p /var/tmp/pid && chmod ugo+rwx /var/tmp/pid' returned 1.

stderr (1/1 lines)
chmod: changing permissions of '/var/tmp/pid': Operation not permitted
```

What about this approach: The `pid` directory is created only if it does not exist with the correct permissions. If the directory is present we would suppose that it is already prepared with the right permissions. To prevent possible collisions with other projects I would suggest to use `/var/tmp/tmt/pid` as the default path as suggested above.

Btw, any security concern about creating world-writable directory under `/var/tmp`?

mcasquer:
> What about this approach: The `pid` directory is created only if it does not exist with the correct permissions. If the directory is present we would suppose that it is already prepared with the right permissions. To prevent possible collisions with other projects I would suggest to use `/var/tmp/tmt/pid` as the default path as suggested above.

Updated and working now, thanks!! @psss @happz

tmt/steps/execute/internal.py
thrix: addressed

tmt/steps/provision/__init__.py
thrix: addressed

psss (approved):

psss:
@happz, @mcasquer, seems we have a bunch of tests failing:

    /tests/execute/reboot/out-of-sessionwith
    /tests/provision/becomewith
    /tests/provision/container/alpinewith

And seems all of those seems to be related to the change. Unsuccessful reboot out of the session, creating the pid directory in a container under a regular user and no `bash` on `alpine`.

happz:
> @happz, @mcasquer, seems we have a bunch of tests failing:
> 
> ```
> /tests/execute/reboot/out-of-sessionwith
> /tests/provision/becomewith
> /tests/provision/container/alpinewith
> ```
> 
> And seems all of those seems to be related to the change. Unsuccessful reboot out of the session, creating the pid directory in a container under a regular user and no `bash` on `alpine`.

Checking. I think I know what's going in in the `reboot/out-of-session`, but others are a mystery to me at this point.

happz:
> /tests/provision/become

This one's easy: as an unprivileged user, you cannot create `/var/tmp/tmt/pid` directory:

```
/var/tmp/tmt/run-011
warn: User is feeling safe.
Found 1 plan.

/test/root
    discover
        how: shell
        order: 50
        summary: 1 test selected
            /user-is-root
    provision
        
        how: container
        order: 50
        become: True
        image: localhost/tmt/tests/container/fedora/rawhide/unprivileged:latest
        user: fedora
        primary address: tmt-011-CmLIngWk
        topology address: tmt-011-CmLIngWk
        Check for container image 'localhost/tmt/tests/container/fedora/rawhide/unprivileged:latest'.
        name: tmt-011-CmLIngWk
        Create network 'tmt-011-network'.
            cmd: if [ ! -d /var/tmp/tmt/pid ]; then mkdir -p /var/tmp/tmt/pid             && chmod ugo+rwx /var/tmp/tmt/pid; fi
            err: mkdir: cannot create directory '/var/tmp/tmt/pid': Permission denied
        fail: Command 'if [ ! -d /var/tmp/tmt/pid ]; then mkdir -p /var/tmp/tmt/pid             && chmod ugo+rwx /var/tmp/tmt/pid; fi' returned 1.
    finish

plan failed

Cause number 1:

    provision step failed

    Cause number 1:

        Command 'if [ ! -d /var/tmp/tmt/pid ]; then mkdir -p /var/tmp/tmt/pid             && chmod ugo+rwx /var/tmp/tmt/pid; fi' returned 1.

        # stderr (1 lines)
        mkdir: cannot create directory '/var/tmp/tmt/pid': Permission denied
```

happz:
> /tests/provision/container/alpine

Test checks for `bash` not being installed on guest. `GuestFacts._execute()` looks for a substring in command stderr, `f'executable file `{tmt.utils.DEFAULT_SHELL}` not found' in exc.stderr:`, but this shell script creating pidfile root ends with something else:

```
        # stderr (1 lines)
        Error: crun: executable file `/bin/bash` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

Looks like a simple matter, we need to make sure the check in `_execute()` detects this form as well as the original one.

happz:
> /tests/execute/reboot/out-of-sessionwith

This one fails because of incorrect pidfile root in `tmt-reboot` and `tmt-reboot-core`. Pidfile root was not exposed, so they were using the old one. Can we apply the following patch?

```diff
diff --git a/tests/execute/reboot/out-of-session.sh b/tests/execute/reboot/out-of-session.sh
index 119b0302..16d14cb9 100755
--- a/tests/execute/reboot/out-of-session.sh
+++ b/tests/execute/reboot/out-of-session.sh
@@ -4,6 +4,12 @@
 rlJournalStart
     rlPhaseStartSetup
         rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
+
+        # To allow running the test outside of tmt.
+        rlRun "TMT_TEST_PIDFILE_ROOT=${TMT_TEST_PIDFILE_ROOT:-/var/tmp/tmt/pid}"
+        rlRun "TMT_TEST_PIDFILE=${TMT_TEST_PIDFILE:-$TMT_TEST_PIDFILE_ROOT/tmt-test.pid}"
+        rlRun "TMT_TEST_PIDFILE_LOCK=${TMT_TEST_PIDFILE_LOCK:-$TMT_TEST_PIDFILE.lock}"
+
         rlRun "run=\$(mktemp -d -p /var/tmp)" 0 "Create run directory"
         rlRun "set -o pipefail"
         rlRun "pushd out-of-session"
@@ -39,7 +45,7 @@ rlJournalStart
         # and use it to issue a `tmt-reboot` from outside the direct process tree of the test.
         set -x
 
-        tmt_reboot_command="export TMT_TEST_PIDFILE=/var/tmp/tmt-test.pid; export TMT_TEST_PIDFILE_LOCK=/var/tmp/tmt-test.pid.lock; export TMT_DEBUG=1; tmt-reboot"
+        tmt_reboot_command="export TMT_TEST_PIDFILE=$TMT_TEST_PIDFILE; export TMT_TEST_PIDFILE_LOCK=$TMT_TEST_PIDFILE_LOCK; export TMT_DEBUG=1; tmt-reboot"
 
         if [ "$PROVISION_HOW" = "container" ]; then
             podman_exec="$(sed -nr 's/\s*Run command: (podman exec .*) \/bin\/bash.*cd.*/\1/p' tmt.output | head -1)"
diff --git a/tmt/steps/execute/internal.py b/tmt/steps/execute/internal.py
index 393d3ef4..6507852d 100644
--- a/tmt/steps/execute/internal.py
+++ b/tmt/steps/execute/internal.py
@@ -30,6 +30,8 @@ from tmt.utils import (
     format_timestamp,
     )
 
+# Note: if modified, pidfile root, filename, and lock filename must be
+# also changed in `tmt-reboot` and `tmt-reboot-core` scripts.
 TEST_PIDFILE_FILENAME = 'tmt-test.pid'
 TEST_PIDFILE_LOCK_FILENAME = f'{TEST_PIDFILE_FILENAME}.lock'
 
@@ -254,6 +256,8 @@ class ExecuteInternal(tmt.steps.execute.ExecutePlugin[ExecuteInternalData]):
         assert isinstance(self.parent, tmt.steps.execute.Execute)
         assert self.parent.plan.my_run is not None
 
+        environment['TMT_TEST_PIDFILE_ROOT'] = EnvVarValue(
+            effective_pidfile_root())
         environment['TMT_TEST_PIDFILE'] = EnvVarValue(
             effective_pidfile_root() / TEST_PIDFILE_FILENAME)
         environment['TMT_TEST_PIDFILE_LOCK'] = EnvVarValue(
diff --git a/tmt/steps/execute/scripts/tmt-reboot b/tmt/steps/execute/scripts/tmt-reboot
index 259dbec0..3851170a 100755
--- a/tmt/steps/execute/scripts/tmt-reboot
+++ b/tmt/steps/execute/scripts/tmt-reboot
@@ -6,7 +6,9 @@
 
 [ -n "$TMT_DEBUG" ] && set -x
 
-TMT_TEST_PIDFILE_LOCK="${TMT_TEST_PIDFILE_LOCK:-/var/tmp/tmt-test.pid}"
+TMT_TEST_PIDFILE_ROOT="${TMT_TEST_PIDFILE_ROOT:-/var/tmp/tmt/pid}"
+TMT_TEST_PIDFILE="${TMT_TEST_PIDFILE:-$TMT_TEST_PIDFILE_ROOT/tmt-test.pid}"
+TMT_TEST_PIDFILE_LOCK="${TMT_TEST_PIDFILE_LOCK:-$TMT_TEST_PIDFILE.lock}"
 
 PATH=/sbin:/usr/sbin:$PATH
 
diff --git a/tmt/steps/execute/scripts/tmt-reboot-core b/tmt/steps/execute/scripts/tmt-reboot-core
index 96ed3b8a..3057cfa2 100755
--- a/tmt/steps/execute/scripts/tmt-reboot-core
+++ b/tmt/steps/execute/scripts/tmt-reboot-core
@@ -2,7 +2,9 @@
 
 [ -n "$TMT_DEBUG" ] && set -x
 
-TMT_TEST_PIDFILE="${TMT_TEST_PIDFILE:-/var/tmp/tmt-test.pid}"
+TMT_TEST_PIDFILE_ROOT="${TMT_TEST_PIDFILE_ROOT:-/var/tmp/tmt/pid}"
+TMT_TEST_PIDFILE="${TMT_TEST_PIDFILE:-$TMT_TEST_PIDFILE_ROOT/tmt-test.pid}"
+TMT_TEST_PIDFILE_LOCK="${TMT_TEST_PIDFILE_LOCK:-$TMT_TEST_PIDFILE.lock}"
 
 if [ ! -e "$TMT_TEST_PIDFILE" ]; then
     echo "There is no running test to signal the reboot to!"
diff --git a/tmt/steps/provision/__init__.py b/tmt/steps/provision/__init__.py
index 89a90b38..1b41cfd2 100644
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -925,9 +925,14 @@ class Guest(tmt.utils.Common):
         """
         from tmt.steps.execute.internal import effective_pidfile_root
         pid_directory = effective_pidfile_root()
-        command = f"if [ ! -d {pid_directory} ]; then mkdir -p {pid_directory} \
-            && chmod ugo+rwx {pid_directory}; fi"
-        self.execute(ShellScript(command))
+        self.execute(ShellScript(
+            f"""
+            if [ ! -d {pid_directory} ]; then \
+                   mkdir -p {pid_directory} \
+                && chmod ugo+rwx {pid_directory}; \
+            fi
+            """
+        ))
 
     # A couple of requiremens for this field:
     #
```

mcasquer:
> This one fails because of incorrect pidfile root in `tmt-reboot` and `tmt-reboot-core`. Pidfile root was not exposed, so they were using the old one. Can we apply the following patch?

Patch applied !

happz:
> > This one fails because of incorrect pidfile root in `tmt-reboot` and `tmt-reboot-core`. Pidfile root was not exposed, so they were using the old one. Can we apply the following patch?
> 
> Patch applied !

It does not look applied, I'm looking at diff in this PR and I see no changes I proposed.

mcasquer:
> It does not look applied, I'm looking at diff in this PR and I see no changes I proposed.

My fault :confused: :disappointed_relieved: included now

mcasquer:
> 
> ```
>         # stderr (1 lines)
>         Error: crun: executable file `/bin/bash` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
> ```
> 
> Looks like a simple matter, we need to make sure the check in `_execute()` detects this form as well as the original one.

@happz 
But that's the message triggered now when creating the pidfile? I don't get why it doesn't match the substring

happz:
> > ```
> >         # stderr (1 lines)
> >         # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> >         Error: crun: executable file `/bin/bash` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
> >         # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Looks like a simple matter, we need to make sure the check in `_execute()` detects this form as well as the original one.
> 
> @happz But that's the message triggered now when creating the pidfile? I don't get why it doesn't match the substring

No idea, TBH. Looks similar, but it's different phrasing. I did not have much time to investigate, for now I'd propose to detect this version as well.

mcasquer:
> This one's easy: as an unprivileged user, you cannot create `/var/tmp/tmt/pid` directory:

Correct me if I'm wrong, but adding sudo to the tmt commands would be fine in this case?

happz:
> > This one's easy: as an unprivileged user, you cannot create `/var/tmp/tmt/pid` directory:
> 
> Correct me if I'm wrong, but adding sudo to the tmt commands would be fine in this case?

I believe so. Check guest fact whether superuser is enabled, from the `setup()` method it should be something like `self.facts.is_superuser`, and if it's false, we will need to run the script with `sudo`. It should be possible to incorporate it into the `ShellScript` content.

tests/execute/reboot/out-of-session.sh
```diff
@@ -4,6 +4,12 @@
 rlJournalStart
     rlPhaseStartSetup
         rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
+
+        # To allow running the test outside of tmt.
+        rlRun "TMT_TEST_PIDFILE_ROOT=${TMT_TEST_PIDFILE_ROOT:-/var/tmp/tmt/pid}"
+        rlRun "TMT_TEST_PIDFILE=${TMT_TEST_PIDFILE:-$TMT_TEST_PIDFILE_ROOT/tmt-test.pid}"
+        rlRun "TMT_TEST_PIDFILE_LOCK=${TMT_TEST_PIDFILE_LOCK:-$TMT_TEST_PIDFILE.lock}"
```
happz, line 11: Hmmm, I think we need to be more strict: I believe tmt running in TF sets these for our tests, to the old value of path and directory, and since we allow the variable to exist, the old ones sneak in and ruin the test. SO we need the following change, at least:

```
        rlRun "TMT_TEST_PIDFILE_ROOT=/var/tmp/tmt/pid"
        rlRun "TMT_TEST_PIDFILE=$TMT_TEST_PIDFILE_ROOT/tmt-test.pid"
        rlRun "TMT_TEST_PIDFILE_LOCK=$TMT_TEST_PIDFILE.lock"
```

See the log:

```
:: [ 13:02:45 ] :: [  BEGIN   ] :: Running 'TMT_TEST_PIDFILE_ROOT=/var/tmp/tmt/pid'
:: [ 13:02:45 ] :: [   PASS   ] :: Command 'TMT_TEST_PIDFILE_ROOT=/var/tmp/tmt/pid' (Expected 0, got 0)
:: [ 13:02:45 ] :: [  BEGIN   ] :: Running 'TMT_TEST_PIDFILE=/var/tmp/tmt-test.pid'
:: [ 13:02:45 ] :: [   PASS   ] :: Command 'TMT_TEST_PIDFILE=/var/tmp/tmt-test.pid' (Expected 0, got 0)
:: [ 13:02:45 ] :: [  BEGIN   ] :: Running 'TMT_TEST_PIDFILE_LOCK=/var/tmp/tmt-test.pid.lock'
:: [ 13:02:45 ] :: [   PASS   ] :: Command 'TMT_TEST_PIDFILE_LOCK=/var/tmp/tmt-test.pid.lock' (Expected 0, got 0)
```

File locations are obviously incorrect, old.

happz:
@mcasquer I'm afraid running test cases - `tmt --context provisiontest=$PROVISION_HOW run -rvvv plan --name /test/root` - with `sudo` will hide the problem: tmt, when told to use a less-privileged account to run tests on a guest, will fail to create pidfile directory. Adding `sudo` to the test will not remove the problem, the test may pass, but the problem will remain. IIUIC, the solutions should be adding `sudo` to the command actually creating the pidfile directory, i.e. https://github.com/teemtee/tmt/pull/3252/files#diff-f684593942660856c3bdc810b597c3956b4107716a554eece2c2bf08344a9503R928.

> Check guest fact whether superuser is enabled, from the setup() method it should be something like self.facts.is_superuser, and if it's false, we will need to run the script with sudo. It should be possible to incorporate it into the ShellScript content.

```python
        sudo = 'sudo' if not self.facts.is_superuser and self.become else ''

        self.execute(ShellScript(
            f"""
            if [ ! -d {pid_directory} ]; then \
                   {sudo} mkdir -p {pid_directory} \
                && {sudo} chmod ugo+rwx {pid_directory}; \
            fi
            """
            ))
```

See a similar problem solved with the same approach: https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/__init__.py#L1714. That should teach tmt to use sudo when needed for privileged operations.

tmt/steps/provision/__init__.py
```diff
@@ -923,6 +923,16 @@ def setup(self) -> None:
 
         Setup the guest after it has been started. It is called after :py:meth:`Guest.start`.
         """
+        from tmt.steps.execute.internal import effective_pidfile_root
```
happz, line 926: One thing I noticed: there's `Guest.setup()` and `GuestSsh.setup()`. The former is pretty empty, and you are adding first code into it; the latter already contains some code.

Byw `GuestSsh.setup()` does not call its base class implementation, it lacks `super().setup()` call. Looks like an omission which was fine, `Guest.setup()` did nothing so it didn't hurt. Now it does, and we need to call it. Please, add the `super()` call to `GuestSsh` so we have the chain complete.

mcasquer:
> IIUIC, the solutions should be adding `sudo` to the command actually creating the pidfile directory,

Oh, sorry for that I misunderstood your comment above, thanks a lot for so detailed explanation

tmt/steps/provision/__init__.py
mcasquer: So with this you mean it'd better if I move the code from `Guest.setup()` to the `GuestSsh` one and add the super().setup() ? or only do the last part in the Guest.setup() function?

happz: Nono, I think `Guest` is the right place. We just need one extra call at the beginning of `GuestSsh.setup`, and that would be calling its base class implementation of the method it overrides.

mcasquer: Got it, code updated, thanks!

happz:
@mcasquer my mistake: drop the `self.become` part of the condition, it must apply every time user is not a superuser:

```diff
diff --git a/tmt/steps/provision/__init__.py b/tmt/steps/provision/__init__.py
index 907d17dc..7dcda9d3 100644
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -924,7 +924,7 @@ class Guest(tmt.utils.Common):
         Setup the guest after it has been started. It is called after :py:meth:`Guest.start`.
         """
         from tmt.steps.execute.internal import effective_pidfile_root
-        sudo = 'sudo' if not self.facts.is_superuser and self.become else ''
+        sudo = 'sudo' if not self.facts.is_superuser else ''
         pid_directory = effective_pidfile_root()
         self.execute(ShellScript(
             f"""
```

That should resolve the test finally :)

mcasquer:
> That should resolve the test finally :)

Updated !

martinhoyer (approved):

tests/provision/facts/test-guest.sh
```diff
@@ -40,7 +42,7 @@ rlJournalStart
             fi
 
         elif [ "$PROVISION_HOW" = "container" ]; then
-            provision_options="--image fedora:39"
+            provision_options="--image $TEST_IMAGE_PREFIX/fedora/39/unprivileged:latest"
             bfu_provision_options="$provision_options --user=nobody"
```
happz, line 46: `--user` would need to change as well, to `fedora`, and the main image shouldn't be the unprivileged one. Together, I think something like this should be the outcome:

```
provision_options="--image $TEST_IMAGE_PREFIX/fedora/39:latest"
bfu_provision_options="--image $TEST_IMAGE_PREFIX/fedora/39/unprivileged:latest --user=fedora"
```

mcasquer:
It seems now there's an issue when setting the ACLs, the user doesn't have permissions
```
cmd:
            mkdir -p /var/tmp/tmt;
            setfacl -d -m o:rX /var/tmp/tmt
            err: setfacl: /var/tmp/tmt: Operation not permitted
```
Also I see some extra error about an rsync command along with the following message
`fail: Failed to push workdir to the guest. This usually means that login as 'fedora' to the guest does not work.`

tmt/steps/provision/__init__.py
```diff
@@ -923,6 +923,17 @@ def setup(self) -> None:
...
+        sudo = 'sudo' if not self.facts.is_superuser else ''
+        pid_directory = effective_pidfile_root()
+        self.execute(ShellScript(
+            f"""
+            if [ ! -d {pid_directory} ]; then \
```
martinhoyer, line 931: perhaps the test for pid_directory could be done before the shell?
```
if not pid_directory.exists() and not pid_directory.is_dir:
    self.execute(ShellScript(
    ...
```

martinhoyer (approved):
lgtm, thanks!

tmt/steps/provision/__init__.py
happz: We cannot use `pid_directory.exists()`, which runs on tmt runner, but we need to test the existence on the guest.

mcasquer:
I see now the failed tests are
`/tests/core/environment-file`
```
:: [ 10:30:11 ] :: [   FAIL   ] :: Command 'tmt run --environment STR=bad_str -rvvvddd plan --name /plan/good 2>&1             | tee output' (Expected 1, got 2)
:: [ 10:30:11 ] :: [   FAIL   ] :: File 'output' should contain 'AssertionError: assert 'bad_str' == 'O'' 
```
`/tests/prepare/ansible`
```
:: [ 10:54:20 ] :: [   FAIL   ] :: Command 'tmt run -i /tmp/tmp.7zaDNljweN --scratch -av provision -h container -i localhost/tmt/tests/container/alpine:latest plan -n /remote' (Expected 0, got 2)
:: [ 10:54:20 ] :: [   FAIL   ] :: File '/tmp/tmp.7zaDNljweN/log.txt' should contain 'ansible-playbook -vvv' 
```
Sorry I don't see how is this related with the patch... :confused:  @happz @psss could you shed some light here?

psss:
Hmmm, these seem to be caused by GitHub randomly refusing to serve the raw content:

    fail: Failed to fetch remote playbook 'https://raw.githubusercontent.com/teemtee/tmt/main/tests/prepare/ansible/data/playbook.yml'.
    403 Client Error: Forbidden for url: https://raw.githubusercontent.com/teemtee/tmt/22a46a4a6760517e3eadbbff0c9bebdb95442760/tests/core/env/data/vars.yaml

Sounds like they enabled some rate limiting, or we hit the limit:

* https://stackoverflow.com/questions/66522261/does-github-rate-limit-access-to-public-raw-files

Any suggestions?

mcasquer:
I see this test case is also failing:
```
Command '
mkdir -p /var/tmp/tmt;
setfacl -d -m o:rX /var/tmp/tmt
' returned 1.

# stderr (1 lines)
setfacl: /var/tmp/tmt: Operation not permitted
```
I think the same approach for the pidfile can be used?, something like: 
```python
sudo = 'sudo' if not self.facts.is_superuser else ''
...
    {sudo} setfacl -d -m o:rX {workdir_root}
    """))
```
what do you think? @psss @happz

psss:
> I think the same approach for the pidfile can be used?, something like:
> 
> ```python
> sudo = 'sudo' if not self.facts.is_superuser else ''
> ...
>     {sudo} setfacl -d -m o:rX {workdir_root}
>     """))
> ```
> 
> what do you think? @psss @happz

Yes, I think `sudo` should be used here as well. This part of code is under the following condition:

```python
if not self.facts.is_superuser and self.become
```

Meaning that we should probably use it here always. @happz, any thoughts?

@carlosrodfern, as this `setfacl` code was submitted by you, do you have any additional recommendations here?

psss:
Anyway, this is too late for 1.39. Let's merge this as one of the first changes in 1.40 in order to give it some testing.

KwisatzHaderach (approved):

carlosrodfern:
> `/tests/provision/become` failed again:
> 
> ```
> rsync: [generator] recv_generator: mkdir "/var/tmp/tmt/run-001" failed: Permission denied (13)
> *** Skipping any contents from this failed directory ***
> rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.3.0]
> ```
> 
> Seems that the approach suggested by @carlosrodfern could help here as well:
> 
> > Wouldn't it better to create `/var/tmp/tmt` first with the corresponding acl to avoid issues with umask 0027, and then do the nested dirs?
> 
> @carlosrodfern, do you have any specific suggestions for the acl setup? Seems there's nobody around who would be closely familiar with the `setfacl` possibilities. Should it be something like this?
> 
> ```
> sudo setfacl -d -m u::rwx,g::rwx,o::rwx /var/tmp/tmt
> ```
> 
> Or is there possibly some more fine-grained approach for this use case?

With an upper case `X`: `rwX`, and the most important here really is `others`: (`o`). This will need to run _before_ anything is created inside the `/var/tmp/tmt` directory.

I can try to pull this work and try a few things.

psss:
> With an upper case `X`: `rwX`, and the most important here really is `others`: (`o`). This will need to run _before_ anything is created inside the `/var/tmp/tmt` directory.

@mcasquer, could you try this? The relevant code for the local workdir creation should be here:

https://github.com/teemtee/tmt/blob/94951abdf36e72da28bc984e79172ab6b47a9a3f/tmt/utils/__init__.py#L2006-L2013

But we should ensure the needed permissions are set on the guest as well. Perhaps we should always (not only for the become use case) use the `Guest.setup()` method to create the workdir on the guest or check that it has the right permissions before any `guest.push()` is called.

> I can try to pull this work and try a few things.

If you could help to complete this, that would be great!

mcasquer:
> @mcasquer, could you try this? The relevant code for the local workdir creation should be here:

@psss setfacl added right after `pid_directory` folder creation, just in case I updated the existing one too.

tmt/steps/provision/__init__.py
```diff
@@ -923,6 +923,18 @@ def setup(self) -> None:
...
+                   {sudo} mkdir -p {pid_directory} \
+                && {sudo} setfacl -d -m u::rwX,g::rwX,o::rwX \
```
carlosrodfern, line 933: The directory is missing in here, right?

@psss @happz @mcasquer,

But also, this logic in general won't work for a system with `umask 0027`, if you want subsequent files and directories in the created directories to be world-writable.

The logic should be something like this:
1. Find out if the `work_dir` directory is prefix of `pid_directory`
2. Make `work_dir` world writable if it is not
3. Set the `o` acl to `rwX` in `work_dir`
4. Create the rest of directory with `mkdir -p {pid_directory}`.


If `work_dir` is not a prefix of `pid_directory`, then you would still need to create `pid_directory` and do chmod recursively, but keep into account that anything else created aftewards in that structure won't be world-writable. Only your pid path.

Here is a test:

Fail scenario:
```sh
# $ podman run -it quay.io/fedora/fedora:latest

dnf install -y /usr/bin/setfacl

umask 0027 # by default, files and directory will be created with no permission for others, even if you do chmod +rwx

mkdir -p /var/tmp/tmt/pid # create the structure
chmod -R o+rwX /var/tmp/tmt # even make it all world writable underneath the work directory
ls -lash /var/tmp/tmt # verify it
ls -lash /var/tmp/tmt/pid # verify it
touch /var/tmp/tmt/pid/pid.lock # however, the file is not, because umask doesn't allow it.
ls -lash /var/tmp/tmt/pid/pid.lock # check it.

# In fact, other directory you create `/var/tmp/tmt` won't have the world-writable permissions:
mkdir /var/tmp/tmt/test
ls -lash /var/tmp/tmt

# A test users:
useradd test
sudo -u test touch /var/tmp/tmt/pid/pid.lock
touch: cannot touch '/var/tmp/tmt/pid/pid.lock': Permission denied
sudo -u test ls /var/tmp/tmt/test
ls: cannot open directory '/var/tmp/tmt/test': Permission denied
```

Success scenario:

```sh
# $ podman run -it quay.io/fedora/fedora:latest

dnf install -y /usr/bin/setfacl
umask 0027

mkdir /var/tmp/tmt
chmod o+rwX /var/tmp/tmt
setfacl -d -m o:rwX /var/tmp/tmt
# Now everything that it is created, directories or files, within /var/tmp/tmt and its subdirectories are created world-writable

mkdir -p /var/tmp/tmt/pid
touch /var/tmp/tmt/pid/pid.lock
s -lash /var/tmp/tmt
# total 0
# 0 drwxr-xrwx+ 1 root root  6 Dec 10 14:41 .
# 0 drwxrwxrwt. 1 root root  6 Dec 10 14:40 ..
# 0 drwxr-xrwx+ 1 root root 16 Dec 10 14:41 pid

ls -lash /var/tmp/tmt/pid
# total 0
# 0 drwxr-xrwx+ 1 root root 16 Dec 10 14:41 .
# 0 drwxr-xrwx+ 1 root root  6 Dec 10 14:41 ..
# 0 -rw-r--rw-. 1 root root  0 Dec 10 14:41 pid.lock

sudo -u test touch /var/tmp/tmt/pid/pid.lock # exit 0
mkdir -p /var/tmp/tmt/test
sudo -u test ls /var/tmp/tmt/test # exit 0
```

mcasquer: Assuming it uses the `-p` flag no error is reported and the `workdir_root` directory is created during this process

mcasquer:
@carlosrodfern @psss @happz after applying the `setfacl` into the local (where tmt runs too) the pidfile directory cannot be created:
```
out: mkdir: cannot create directory '/var/tmp/tmt/pid': Permission denied
out: flock: cannot open lock file /var/tmp/tmt/pid/tmt-test.pid.lock: No such file or directory
```
What am I missing here?

carlosrodfern:
> > @carlosrodfern @psss @happz after applying the `setfacl` into the local (where tmt runs too) the pidfile directory cannot be created:
> > ```
> > out: mkdir: cannot create directory '/var/tmp/tmt/pid': Permission denied
> > out: flock: cannot open lock file /var/tmp/tmt/pid/tmt-test.pid.lock: No such file or directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What am I missing here?
> 
> Is it with `container` or `virtual` in the test?
> 
> During the test, can you add `ls -lash /var/tmp/tmt/pid`, and `getfacl /var/tmp/tmt` and get the output?

Hmmm, doesn't the container provisioner create the `/var/tmp/tmt` locally and mounts it in the container? when you mount it, the permissions are not passed to the container. Would it be happening that in the case of the container, the code that sets up the workdir in `_check_or_create_workdir_root_with_perms` is not being run because the workdir root exists since it is mounted, so the pid file is being created within the `/var/tmp/tmt` before the `setup` code that sets the acl?

You could test getting the setfacl out of the `if not self.workdir_root.is_dir()`:

```python
        def _check_or_create_workdir_root_with_perms() -> None:
            """ If created workdir_root has to be 1777 for multi-user"""
            if not self.workdir_root.is_dir():
                try:
                    self.workdir_root.mkdir(exist_ok=True, parents=True)
                    self.workdir_root.chmod(0o1777)
                    command = Command('setfacl', '-d', '-m', 'o:rwX', str(self.workdir_root)) # I don't think this 
                    self.run(command)
                except OSError as error:
                    raise FileError(f"Failed to prepare workdir '{self.workdir_root}': {error}")
```

thrix:
@carlosrodfern howdy, I am the 1.42 release lead, can you please re-review if ok with you?

mcasquer:
> You could test getting the setfacl out of the `if not self.workdir_root.is_dir()`, or perhaps remove the condition all together:

Sorry for not replying, I removed the condition but it seems tests are still failing

carlosrodfern:
I will take a look.

carlosrodfern:
@mcasquer, I created a pull request with a fix. I tested it on your changes and they pass: https://github.com/teemtee/tmt/pull/3483

happz:
I'd argue https://github.com/teemtee/tmt/pull/3518 should go in first, because of tests/provision/facts/test-guest.sh changes, and there is also https://github.com/teemtee/tmt/pull/3483 which is very related and maybe should be a prerequisite of this PR, as mentioned in https://github.com/teemtee/tmt/pull/3252#issuecomment-2614146199

psss:
Current fix breaks the `virtual` provision because the default acl on `/var/tmp/tmt` causes the generated private key to be readable by group and others:

```
> ls -l /var/tmp/tmt/run-106/user/plan/fresh/provision/default-0/id_ecdsa
-rw-r--r--. 1 psss psss 505 26. bře 08.21 /var/tmp/tmt/run-106/user/plan/fresh/provision/default-0/id_ecdsa
```

Which results into `ssh` to refuse to connect to the guest:

```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0644 for '/var/tmp/tmt/run-106/user/plan/fresh/provision/default-0/id_ecdsa' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/var/tmp/tmt/run-106/user/plan/fresh/provision/default-0/id_ecdsa": bad permissions
```

Unfortunately this will need some more work, moving to backlog.

PR #3263, opened 2024-10-05
Replace open+read/write with pathlib's read/write_text/bytes

happz:
Pull Request Checklist

* [x] implement the feature


tmt/base.py
```diff
@@ -4326,8 +4326,8 @@ def resolve_dynamic_ref(
 
     # Read it, process it and get the value of the attribute 'ref'
     try:
-        with open(ref_filepath, encoding='utf-8') as datafile:
-            data = tmt.utils.yaml_to_dict(datafile.read())
+        data = tmt.utils.yaml_to_dict(ref_filepath.read_text(encoding='utf-8'))
+
```
martinhoyer, line 4330: I believe utf8 is default, so could be dropped since we're "modernizing"?

happz: I’d love to, but I’m not sure about the default:

> … the contents of the file are returned as [str](https://docs.python.org/3/library/stdtypes.html#str), the bytes having been first decoded using a platform-dependent encoding or using the specified encoding if given.

It may not necessarily be `utf-8`, IIUIC. I’m not sure what’s better, whether we should leave it to OS and Python to pick a sane default, or whether we should enforce the default as we see it, or whether we should always enforce UTF-8, without any exception. I also think we still ignore the problem of locales when parsing output of various commands, too, so far we were able to get away with it. Seems like we have several topics to discuss :)

thrix: Yeah, I believe another patch rather?

therazix (approved):

tmt/utils/__init__.py
```diff
@@ -2005,10 +2008,14 @@ def write(
...
-                file.write(data)
+            if mode == 'a':
+                path.append_text(data, encoding='utf-8', errors='replace')
+
+            else:
+                path.write_text(data, encoding='utf-8', errors='replace')
+
         except OSError as error:
-            raise FileError(f"Failed to write '{path}'.\n{error}")
+            raise FileError("Failed to write into '{path}' file.") from error
```
thrix, line 2018: @happz why is f-string not needed here?

tmt/utils/__init__.py
```diff
@@ -4325,20 +4330,19 @@ def url_and_name(self, cwd: Optional[Path] = None) -> list[tuple[str, str]]:
...
-                            f"Couldn't match '{self.sources_file_name}' "
-                            f"content with '{self.re_source.pattern}'.")
-                    used_hash, source_name, hash_value = match.groups()
-                    ret_values.append((self.lookaside_server + self.uri.format(
-                        name=package,
-                        filename=source_name,
-                        hash=hash_value,
-                        hashtype=used_hash.lower()
-                        ), source_name))
+            for line in (cwd / self.sources_file_name).read_text().splitlines():
```
thrix, line 4333: Seems there is a pattern here, could we add

```
def read_lines(path: Path) -> list[str]:
    return path.read_text().splitlines()
```

To our utils? So it would become:

```
            for line in read_path_lines(cwd / self.sources_file_name):
```

happz: Added in 187f79de. Iterator is not possible because both `read_text()` and `splitlines()` have their own options.

tmt/utils/__init__.py
happz: Fixed in 9068dc45.

tmt/utils/__init__.py
thrix: ack, looks great to me.

thrix (approved):
LGTM

tmt/utils/__init__.py
```diff
@@ -1985,16 +1988,16 @@ def read(self, path: Path, level: int = 2) -> str:
             path = self.workdir / path
         self.debug(f"Read file '{path}'.", level=level)
         try:
-            with open(path, encoding='utf-8', errors='replace') as data:
-                return data.read()
+            return path.read_text(encoding='utf-8', errors='replace')
+
         except OSError as error:
-            raise FileError(f"Failed to read '{path}'.\n{error}")
+            raise FileError("Failed to read from '{path}'.") from error
```
psss, line 1994: Missing f-string here?

psss (changes requested):
Nice! Just one nitpick?

tmt/utils/__init__.py
happz: Yeah, it's missing :/ I fixed the same problem in `write()`, and forgot to check `read()`. Fixed in 90500c70.

psss (approved):

happz:
Unrelated failures, merging.

Issue #3138, opened 2024-08-07
bug: tmt try multihost plan breaks terminal


spetrosi:
After running a plan with multi-host provisioning with 3+ hosts, terminal "breaks", it prints many spaces in the beginning of the lines, and then doesn't work. With two nodes, the output is fine. With >3 nodes, it starts shifting.

Plan from plans/test_plan.fmf:

```
summary: A general test for a system role
tag: general_test
provision:
  - name: control_node
    role: control_node
    connection: system
  - name: managed_node1
    role: managed_node
    connection: system
  - name: managed_node2
    role: managed_node
    connection: system
  - name: managed_node3
    role: managed_node
    connection: system
environment:
    ANSIBLE_VER: 2.17
    REPO_NAME: "mssql"
    PYTHON_VERSION: 3.12
prepare:
  # Replace with feature: epel: enabled once https://github.com/teemtee/tmt/pull/3128 is merged
  - name: Enable epel to install beakerlib
    script: |
      # CS 10 and Fedora doesn't require epel
      if grep -q -e 'CentOS Stream release 10' -e 'Fedora release' /etc/redhat-release; then
        exit 0
      fi
      yum install epel-release yum-utils -y
      yum-config-manager --enable epel epel-debuginfo epel-source
discover:
  - name: Prepare managed nodes
    how: fmf
    url: https://github.com/linux-system-roles/tft-tests
    ref: main
    where: managed_node
    filter: tag:prep_managed_node
execute:
    how: tmt
```

I am running it with 

`tmt try -p plans/test_plan centos-stream-9`

What my terminal outputs: https://pastebin.com/Ae18humU

Terminal breaks after `Running 'rlImport library'` line

spetrosi:
```
$ tmt --version
tmt version: 1.34.0
```

spetrosi:
I am providing a simpler reproducer, a plan that provisions 3 nodes, and a simple test. It enough to create two files like below and run `tmt try -p plans/reproducer`.

`plans/reproducer.fmf`:
```
summary: Reproduce bug with extra indentation
provision:
  - name: control_node
    role: control_node
  - name: managed_node1
    role: managed_node
  - name: managed_node2
    role: managed_node
discover:
  - name: Run on 3 managed nodes
    how: fmf
    where: managed_node
    filter: tag:reproducer
  - name: Run on control node
    how: fmf
    where: control_node
    filter: tag:reproducer
execute:
    how: tmt

```

[...]

PR #3265, opened 2024-10-07
Run plans in parallel

happz:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/base/plan.py
+++ b/tmt/base/plan.py
@@ -29,6 +29,7 @@ import tmt.steps.provision
 import tmt.steps.report
 import tmt.templates
 import tmt.utils
+import tmt.utils.filesystem
 import tmt.utils.git
 import tmt.utils.jira
 from tmt._compat.pathlib import Path
@@ -284,6 +285,9 @@ class Plan(
         'ansible',
     ]
 
+    def step_logger(self, step_name: str) -> tmt.log.Logger:
+        return self._logger.descend(logger_name=step_name)
+
     def __init__(
         self,
         *,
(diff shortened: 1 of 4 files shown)


lukaszachy:
Nice PR, it works well for execution runs!

 The usage of 'login' might need some better user experience, IMO.
I have 3 plans executed with `tmt run -avvv login`. Until first login all output is nice but then this happens:
```
[/plan/z]         finish task #1: login
[/plan/y]         login: Starting interactive shell
[/plan/z]         login: Starting interactive shell
[/plan/y]             cmd: /bin/true
[/plan/z]             cmd: /bin/true
                                    [/plan/x]                 duration limit: 300 seconds
                                                                                         [/plan/x]                 cmd: echo
                                                                                                                            [/plan/y]             cmd: bash
[/plan/z]             cmd: bash
                               bash-5.3# bash-5.3# [/plan/x]                 out: 
                                                                                  [/plan/x]                 00:00:00 pass /test/d (on default-0) [4/4]

 [/plan/x]     
               [/plan/x]         summary: 4 tests executed 
<etc, omitted>
```
Once I try to type something, it seems that I have input to all three guests, each of them receiving some parts:
```
bash-5.3# echo
bash: co: command not found
bash-5.3# 
bash-5.3# 
bash-5.3# 
bash: ho: command not found
bash-5.3#
```

A bit related UX problem might be with finding why plan failed (eg when require are not present) from webui (without grep to limit log to particular plan).

happz:
> Nice PR, it works well for execution runs!
> 
> The usage of 'login' might need some better user experience, IMO. I have 3 plans executed with `tmt run -avvv login`. Until first login all output is nice but then this happens:
> 
> ```
> [/plan/z]         finish task #1: login
> [/plan/y]         login: Starting interactive shell
> [/plan/z]         login: Starting interactive shell
> [/plan/y]             cmd: /bin/true
> [/plan/z]             cmd: /bin/true
>                                     [/plan/x]                 duration limit: 300 seconds
>                                                                                          [/plan/x]                 cmd: echo
>                                                                                                                             [/plan/y]             cmd: bash
> [/plan/z]             cmd: bash
>                                bash-5.3# bash-5.3# [/plan/x]                 out: 
>                                                                                   [/plan/x]                 00:00:00 pass /test/d (on default-0) [4/4]
> 
>  [/plan/x]     
>                [/plan/x]         summary: 4 tests executed 
> <etc, omitted>
> ```
> 
> Once I try to type something, it seems that I have input to all three guests, each of them receiving some parts:
> 
> ```
> bash-5.3# echo
> bash: co: command not found
> bash-5.3# 
> bash-5.3# 
> bash-5.3# 
> bash: ho: command not found
> bash-5.3#
> ```
> 
> A bit related UX problem might be with finding why plan failed (eg when require are not present) from webui (without grep to limit log to particular plan).

This will be a problem, as long as you have N interactive commands sharing not just stdout and stderr, but also stdin. It will never work reliably, not possible. See https://github.com/teemtee/tmt/issues/3138#issuecomment-2294166854 and https://github.com/teemtee/tmt/issues/3138#issuecomment-2294171408.

I think this has no simple solution, not without introducing some kind of screen or tmux-like terminal multiplexing, or ncruses-like windowing to split the screen into windows, one for each "stream of execution". Without this, N interactive objects competing over the single terminal will never lead to a usable output.

lukaszachy:
> This will be a problem, as long as you have N interactive commands sharing not just stdout and stderr, but also stdin. It will never work reliably, not possible.

How about enforcing exclusive access to stdin for single connection? 
Possibly as opt-in with default being "tmt is sorry, but login (interactive) cannot be used with plans running in parallel")

In a simplest fix I'd be happy just to see the warning to know what mess I got myself into.

happz:
> > This will be a problem, as long as you have N interactive commands sharing not just stdout and stderr, but also stdin. It will never work reliably, not possible.
> 
> How about enforcing exclusive access to stdin for single connection? Possibly as opt-in with default being "tmt is sorry, but login (interactive) cannot be used with plans running in parallel")

Not just plans, even a single plan with multiple guests and `login` would hit this. Or an `interactive` test on multiple guests.

Yes, this kind of input from the user would be needed. Out of the box, not going to work, so something or someone needs to step in and turn it into a sequence. Either tmt, by refusing to run `login` in parallel (then it needs to guess or be told the order), or by introducing some kind of separate windows (which would be lovely!), or the user by telling tmt which "stream of execution" is supposed to own the terminal, or... We can collect the ideas.

happz:
Wouldn't it be awesome if tmt spawned virtual terminals for these "streams of execution", and let user attach `tmux` or `screen`?

Issue #3235, opened 2024-09-23
Race condition in the `testcloud` plugin


psss:
Based on the recent results of the `/plans/provision/virtual/*` plans it seems the race condition is still there. Multihost tests quite frequently fail with:

    Failed to connect in 300s.

@frantisekz, in #3036 you've mentioned that the change opens way to address the problem in an atomic way. Would you have some time to look into this?

psss:
Here's a [recent example job](https://artifacts.osci.redhat.com/testing-farm/f8652173-3a93-4d75-87af-f2c612c1aae4/) with the [full log details](https://artifacts.osci.redhat.com/testing-farm/f8652173-3a93-4d75-87af-f2c612c1aae4/work-preparelzoi94gk/plans/provision/virtual/prepare/execute/data/guest/default-0/tests/prepare/multihost-3/data/tmp-tmp.sm38eMweLF-log.txt).

frantisekz:
Yeah, I had it upstream for some time: https://pagure.io/testcloud/c/6da8ed8c96e3f748f8c242dc168b45017be251f5?branch=master

I can bake out copr build at least if that'd help with testing? Release is currently held by https://pagure.io/testcloud/issue/191 from rewritten image code, I'll poke some more there to get that resolved.

PR #3244, opened 2024-09-26
Require the latest `testcloud` with the race fix

psss:
Let's try if it's better now!

PR #2078, opened 2023-05-12
Use `testcloud` domain API v2

frantisekz:
Refactors domain configuration interface between tmt and testcloud, to allow greater flexibility and superior api stability when making changes and additions in the future.

Apart from that, doesn't expose support for (will come as a separate PR):
- tpm emulation
- uefi
- cpu core count setting (more advanced topologies to follow https://tmt.readthedocs.io/en/stable/spec/hardware.html#cpu TBD)

testcloud counterparts: 
- https://pagure.io/testcloud/c/6f0942cb3ba31ff0927ee8b151218e8dc8b34d7d?branch=master
- https://pagure.io/testcloud/pull-request/155
- https://pagure.io/testcloud/c/7ee6dc839a46c72db6d23bcd2e057c2de6fe5ea8?branch=master
- ...

(compatible testcloud built at): https://copr.fedorainfracloud.org/coprs/frantisekz/testcloud-wip/

PR #3266, opened 2024-10-07
Pass `{USER, COREOS}_DATA` separately in `testcloud`

frantisekz:
Together with now staged testcloud-0.11.0, should resolve https://github.com/teemtee/tmt/issues/3235 once and for all.

Required testcloud is built at: https://copr.fedorainfracloud.org/coprs/frantisekz/testcloud-temp/

The {USER, COREOS}_DATA variables were affected by threads overwriting them directly in testcloud's config object as observed in https://github.com/teemtee/tmt/pull/3244#issuecomment-2392236256 and following comments. Testcloud-side change can be seen here: https://pagure.io/testcloud/c/ee1202bcf57cd475f4586f3693e2127f9a12e92e?branch=master . 

Only these two variables can differ between various instances in tmt's use of testcloud, the others are kept separately since the domain api switchover last year: https://github.com/teemtee/tmt/pull/2078 .

Co-authored by @jskladan 

@happz @psss 

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -24,6 +24,11 @@ discovered tests. This can be useful especially when fetching
 tests from remote repositories where the user does not have write
 access.
 
+A race condition in the
+:ref:`/plugins/provision/virtual.testcloud` plugin has been fixed,
+thus multihost tests using this provision method should now work
+reliably without unexpected connection failures.
+
 
 tmt-1.37.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ provision-beaker = [
     "mrack>=1.15.1",
     ]
 provision-virtual = [
-    "testcloud>=0.11.2",
+    "testcloud>=0.11.3",
     ]
 provision-container = []
 report-junit = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ provision-beaker = [
     "mrack>=1.15.1",
     ]
 provision-virtual = [
-    "testcloud>=0.9.13",
+    "testcloud>=0.11.2",
     ]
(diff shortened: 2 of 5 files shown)


frantisekz:
So, the CoreOS tests failed previously, the upcoming run has https://pagure.io/testcloud/pull-request/204 included which should resolve that too.

tldr. previously, we could use sock.connect_ex(("127.0.0.1", self.get_instance_port())) in testcloud CoreOS/user sessions to verify that ssh port is open, but something had changed in how this works in libvirt/qemu, and the call now instantly returns OK.

frantisekz:
@psss Reading through the results, if I am not missing anything, the issues are mainly:

`fail: Command 'rpm -q --whatprovides tree-but-spelled-wrong diffutils || dnf install -y  tree-but-spelled-wrong diffutils' returned 1.`

which wouldn't seem caused by testcloud/this PR, and fedora-coreos failures, which I have problems reproducing locally. All in all, this seems to resolve the mysterious failures at least for normal Fedora Cloud :/

.packit.yaml
```diff
@@ -150,19 +153,8 @@ jobs:
...
-    allowed_pr_authors: [ "packit", "psss", "lzachar" ]
-    allowed_committers: [ "packit", "psss", "lzachar" ]
-    dist_git_branches: *all-targets
-
   - job: bodhi_update
     trigger: commit
     dist_git_branches:
       - fedora-branched
-      - epel-9
+      - epel-9
```
happz, line 160:
```suggestion
      - epel-9

```

psss:
Bumped the require again in 8f4d1db3 to get the regression fix.

psss:
Fixed timeout and adjusted the test to exercise the custom command path only: 06d5618f
Filed a new issue to improve the timeout handling: #3305 
Included a short release note in: 31e9b73d

psss (approved):
Now seems to be working nice! :tada:

psss:
@happz, and it's green! \o/

happz (approved):

PR #3269, opened 2024-10-08
Remove `skip` from the `ResultInterpret` enumeration

psss:
As discussed today, the `skip` value was never documented in the test `result` key specification and was added by an accident.

There seems to be a use case to support skipping selected tests but that functionality is substantially different from the result interpretation configured by the `result` key which happens only after test execution is completed so most probably a new test key should be introduced for it.

Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema


martinhoyer (approved):

happz:
Unrelated failures, merging.

Issue #2346, opened 2023-09-22
Provisioning should support Beaker pool enforcement/exclusion


happz:
In Beaker jobs, this is possible to request/exclude a specific **Beaker** pool (not to be confused with **Artemis** pools):

```
system-pool: !virtkvm-gating
system-pool: virtkvm-gating
```

First, it's a feature sought by current Beaker users, and it probably shouldn't be hard to extend the Beaker plugin/mrack to support this.

Second, I'm not sure how to support this with Artemis, and to what extent this should be decoupled from `how: beaker`. As Artemis can and does wrap Beaker service too, it can certainly accept constraints of this kind and propagate them to Beaker behind the curtain. Feels like something generic enough for `hardware`, but it's not actual HW requirements, more like an organizational aspect... Anyway, adding HW labels, open for discussion.

psss:
@skycastlelily, would you be possibly interested in this? We consider this for 1.38 or 1.39.

skycastlelily:
Sure,will look into and work on it tomorrow:)

psss:
That's great, proposing for `1.38` then. Thanks!

skycastlelily:
Here are the merge requests :)
https://github.com/teemtee/tmt/pull/3271 
https://gitlab.com/testing-farm/artemis/-/merge_requests/1283

thrix:
Seems nothing else to do here on tmt side, closing.

PR #3074, opened 2024-07-02
Support `beaker.pool` hardware requirement for mrack

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [x] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

PR #3271, opened 2024-10-09
Translate `beaker.pool` hardware requirement properly for `mrack`

skycastlelily:
Related to https://github.com/teemtee/tmt/issues/2346

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/spec/hardware/beaker.fmf
+++ b/spec/hardware/beaker.fmf
@@ -25,3 +25,4 @@ example:
 
 link:
   - implemented-by: /tmt/steps/provision/mrack.py
+    note: "``==`` and ``!=`` operators only"
--- a/tmt/hardware.py
+++ b/tmt/hardware.py
@@ -1379,7 +1379,8 @@ def _parse_beaker(spec: Spec) -> BaseConstraint:
 
     group = And()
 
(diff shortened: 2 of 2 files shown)


happz:
@skycastlelily I think that with the last push, you overwritten the original changes adding the implementation to `tmt/steps/provision/mrack.py`. All that's left now is the operator update (pre-commit & mypy complain about the list vs tuple, BTW).

It's really better to avoid force pushes, one can easily lose changes. We recommend adding new changes to PRs with new commits, e.g. named "squash: ...", eventually they would be squashed before merging, when all is done.

happz:
>> I think that with the last push, you overwritten the original changes
>> adding the implementation
>
> That's on purpose, I already implemented beaker.pool in merged #3074 , after thinking twice, I prefer to keep the implementation way as it is.

I see, makes sense.

BTW pre-commit is still failing, that's not related to this discussion, and should be fixed.

skycastlelily:
Updated^^

happz (approved):

therazix (approved):

psss:
All failures are irrelevant, merging.

PR #3282, opened 2024-10-11
The `--feeling-safe` option required for `local` provision

psss:
Prepare users for the change in the `local` provision plugin which requires the `--feeling-safe` option to be used when using the plugin to prevent dangerous operations.

* Document the `TMT_FEELING_SAFE` variable
* Mention the change in the release notes
* Update the plugin docs accordingly

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] include a release note

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -366,6 +366,13 @@ TMT_PLUGINS
     Path to a directory with additional plugins. Multiple paths
     separated with the ``:`` character can be provided as well.
 
+TMT_FEELING_SAFE
+    Set this variable to ``True`` to enable potentially dangerous
+    operations such as executing tests directly on the test runner
+    using the ``local`` provision method. Use with caution, only
+    when you can fully trust the ``tmt`` metadata or if you know
+    what you are doing.
+
 TMT_CONFIG_DIR
     Path to an alternative directory with config files. By default
     ``~/.config/tmt`` is used.
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -367,7 +367,7 @@ TMT_PLUGINS
     separated with the ``:`` character can be provided as well.
 
 TMT_FEELING_SAFE
-    Set this variable to ``True`` to enable potentially dangerous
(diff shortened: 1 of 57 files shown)


psss:
I've temporarily lowered the minimal version number to `1.37` to trigger test failures we would get only after `1.38` is released.

docs/overview.rst
```diff
@@ -366,6 +366,13 @@ TMT_PLUGINS
     Path to a directory with additional plugins. Multiple paths
     separated with the ``:`` character can be provided as well.
 
+TMT_FEELING_SAFE
+    Set this variable to ``True`` to enable potentially dangerous
```
happz, line 370: Hmmm, I think this might be incorrect, as an `envvar` key of a Click option, the allowed values would be `1` and `0`. `True` might work, because it's as a non-empty string it's true-ish, but `TMT_FEELING_SAFE=False` shouldn't.

psss: Seems that `click` is happy with any of these:

For enabling the flag:

* `1`
* `true`
* `t`
* `yes`
* `y`
* `on`

For disabling the flag:

* `0`
* `false`
* `f`
* `no`
* `n`
* `off`

We should probably choose one and use it consistently across the docs. Any preference?

happz: Good to know it supports more, I recall having trouble with at least some of them. Nevermind then.

The selection probably happened already, see the last paragraph in https://tmt.readthedocs.io/en/stable/overview.html#plugin-variables :)

psss: Ack, updated in 753c0cf8.

martinhoyer (approved):

KwisatzHaderach (approved):

psss:
Just realized we should mention the option in the guide as well: f154e101
Removing the temporary adjustments in ab987e03 (we should see some failures now).

psss:
Just irrelevant `testcloud` failures, merging.

PR #3286, opened 2024-10-14
Add a helper for saving command output in a file

happz:
This will be a very needed functionality: the plugin runs a command, and its output needs to be saved in a dedicated file, usually because the output represents some interesting, standalone piece of information. For example, an AVC denial report generated by `ausearch`.

If it were the only such actor, it would be fine to keep the implementation in the `avc` plugin. But, with the advent of saving results of the `prepare` and `finish` phases, the question is, if I get `results.yaml` for a `prepare` step, would it even contain anything besides `pass` or `error`? Well, it can contain logs of commands like `ansible-playbook` or shell scripts. And suddenly we have several plugins that may need to save some walls of text, usually produced by a command, in files, and then link those files to results and present them to a user.

And because of that, it makes sense to provide a helper function that would produce unified output for all of them. Why should `avc` separate stdout and stderr by two lines and `prepare/shell` with just one? Not on my watch.

Pull Request Checklist

* [x] implement the feature


KwisatzHaderach (approved):

psss:
Some problem on TF side?

    Guest couldn't be provisioned: Artemis resource ended

Let's reschedule?

psss (approved):
Nice!

tmt/utils/__init__.py
```diff
@@ -2446,29 +2461,104 @@ def from_env(cls) -> 'TracebackVerbosity':
...
-            output_lines = output_lines[-OUTPUT_LINES:]
+            line_summary = f"{min(len(content_lines), OUTPUT_LINES)}/{len(content_lines)}"
+            content_lines = content_lines[-OUTPUT_LINES:]
 
-        yield f'{name} ({line_summary} lines)'
-        yield OUTPUT_WIDTH * '~'
-        yield from output_lines
-        yield OUTPUT_WIDTH * '~'
+        yield f'{comment_sign} {name} ({line_summary} lines)'
+        yield comment_sign + (OUTPUT_WIDTH - 1) * '~'
```
psss, line 2481:
```suggestion
        yield f'{comment_sign} ' + (OUTPUT_WIDTH - 2) * '~'
```

Would you be ok with this as well? I find it a little bit more pleasing to the eye...

happz: I don't mind, just need to fix tests after such a change.

psss: That would be nice, thanks!

happz: Addressed in 21f18965

tmt/utils/__init__.py
```diff
@@ -2446,28 +2461,105 @@ def from_env(cls) -> 'TracebackVerbosity':
...
-        yield f'{name} ({line_summary} lines)'
-        yield OUTPUT_WIDTH * '~'
-        yield from output_lines
-        yield OUTPUT_WIDTH * '~'
+        ## ${label}
+
+        # ${command}
+
+        # stdout (N lines)
+        #~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
psss, line 2530:
```suggestion
        # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Here's a couple more in the docstring.

happz: Addressed in 7fa0e284

tmt/utils/__init__.py
psss: Thanks!

PR #3289, opened 2024-10-15
Add an extra escape wrapper for illegal XML chars in junit report plugin

seberm:
The default Jinja `escape` (`e`) filter does not escape the XML illegal characters (like control characters). The python-xml-junit library was replacing the characters with an empty string:
- https://github.com/kyrus/python-junit-xml/blob/4bd08a272f059998cedf9b7779f944d49eba13a6/junit_xml/__init__.py#L325

For more info, please see:
- http://stackoverflow.com/questions/1707890/fast-way-to-filter-illegal-xml-unicode-chars-in-python
- https://www.w3.org/TR/xml11/#charsets

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -4,6 +4,14 @@
     Releases
 ======================
 
+tmt-1.38.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`/plugins/report/junit` report plugin now removes all
+invalid XML characters from the final JUnit XML.
+
+
+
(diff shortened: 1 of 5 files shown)


psss (approved):
Thanks for fixing this! Looks good and works as expected.

seberm:
> Thanks for fixing this! Looks good and works as expected.

Thanks for the review!

Have you also tested the changes against the internal "base-os" tmt tests? Thanks.

happz (approved):

psss:
> Have you also tested the changes against the internal "base-os" tmt tests? Thanks.

Yes, before the change there's the traceback, with the change everything's green.

Issue #3284, opened 2024-10-14
test failed with error after triggering a kernel panic


coiby:
After triggering a kernel panic, the system can be rebooted but the test just failed with error.  I notice a workaround is to execute the kernel panic trigger command by `tmt-reboot`. 

Not if the test is written with beakerlib, a similar error will occur and something like  `# the errr could also be 00:00:28 errr /client-test/tests/client (on client) (beakerlib: State 'imcomplete') [1/1]` will also be printed.

Here are the logs and the reproducer.

## Logs
```sh
# tmt run -vvv
/var/tmp/tmt/run-020
Found 1 plan.

/example
summary: Basic kernel panic test
    discover
        how: shell
        order: 50
        summary: 1 test selected
            /script-00
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: virtual
        order: 50
        memory: 2048 MB
        disk: 40 GB
        qcow: Fedora-Cloud-Base-Generic.x86_64-40-1.14.qcow2
        effective hardware: {}
        name: tmt-020-PuwHCrwG
        key: /var/tmp/tmt/run-020/example/provision/default-0/id_ecdsa
        progress: booting...
        primary address: 127.0.0.1
        topology address: 127.0.0.1
        port: 10022
        multihost name: default-0
        arch: x86_64
        distro: Fedora Linux 40 (Cloud Edition)
        kernel: 6.8.5-301.fc40.x86_64
        package manager: dnf
        selinux: yes
        is superuser: yes
    
        summary: 1 guest provisioned
    prepare
        queued push task #1: push to default-0
        
        push task #1: push to default-0
    
        queued prepare task #1: requires on default-0
        
        prepare task #1: requires on default-0
        how: install
        summary: Install required packages
        name: requires
        order: 70
        where: default-0
        package: 1 package requested
            /usr/bin/flock
            cmd: rpm -q --whatprovides /usr/bin/flock || dnf install -y  /usr/bin/flock
            out: util-linux-core-2.40-0.9.rc1.fc40.x86_64
    
        queued pull task #1: pull from default-0
        
        pull task #1: pull from default-0
    
        summary: 1 preparation applied
    execute
        queued execute task #1: default-0 on default-0
        
        execute task #1: default-0 on default-0
        how: tmt
        order: 50
        exit-first: false
            test: /script-00
                cmd:
                    echo 2 > /proc/sys/kernel/panic
                    sync
[...]

PR #3291, opened 2024-10-15
Wait for guest connection before test restart

happz:
Even without a reboot, tmt still needs to verify the guest is up and running. The reboot might be triggered beyond the control of tmt, and that is fine, we just need to be sure we restart the test on guest that's alive.

Related to https://github.com/teemtee/tmt/issues/3284

Pull Request Checklist

* [x] implement the feature


skycastlelily (approved):
LBTM^^

therazix (approved):

psss (approved):

PR #3292, opened 2024-10-16
Add support to set number of CPUs for virtual run

jkonecny12:
Without this some tests could be pretty slow because standard is 2 CPUs which is not much these days.

Hi, I'm fighting with mypy and other things and I have hard time to work on this. So I'm creating a PR if someone wants to help me with debugging to finish this if you see the benefit. 

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tests/provision/virtual.testcloud/data/plan.fmf
+++ b/tests/provision/virtual.testcloud/data/plan.fmf
@@ -10,6 +10,7 @@ provision:
     image: fedora
     disk: 10
     memory: 2048
+    cpu: 3
     #user: custom_user
     connection: system
     arch: x86_64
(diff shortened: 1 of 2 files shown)


happz:
@jkonecny12 can you try adding `cpu.processors` HW requirement to your plans? That should do the trick. See https://tmt.readthedocs.io/en/stable/spec/hardware.html#cpu for the specification.

```yaml
hardware:
  cpu:
    processors: 79
```

Or, via command line:

```shell
--hardware 'cpu.processors=79'
```

AFAICT this is not documented correctly, I'll fix that, but that should be all that's needed. And the best thing, the same should work for Beaker and Testing Farm too ;)

jkonecny12:
Sounds great. I wouldn't expect that this is hidden under `hardware` when the `memory` is just top class citizen.

happz:
> Sounds great. I wouldn't expect that this is hidden under `hardware` when the `memory` is just top class citizen.

Yeah, `memory` and `disk` happened before I could force `hardware` into the system :) Internally, they just add `memory: ..` and `disk: ...` HW requirements, and if I could, I would remove them in a heartbeat.

PR #3293, opened 2024-10-16
Document cpu.processors is supported by virtual plugin

happz:
Pull Request Checklist

* [x] implement the feature


jkonecny12:
Hi, could you please add it also here:
https://tmt.readthedocs.io/en/stable/spec/plans.html#virtual

to make it easy to discover. To me it is not obvious that `hardware` is something usable for `virtual` provider.

therazix (approved):

seberm (approved):

happz:
Documentation-only, merging.

psss:
> Maybe it's just me, but when I'm looking for something in the documentation I'm not reading it from start of the section but looking on the details of the given specific part. But I'm fine even with this solution.

Perhaps, we could also provide some more links to the hardware support overview?
https://tmt.readthedocs.io/en/latest/plugins/provision.html#hardware-requirement-support
It seems useful but is probably not linked from all relevant places.

PR #3295, opened 2024-10-17
Translate `cpu.stepping` hardware requirement for `mrack`

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] extend the test coverage
* [x] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [x] mention the version
* [x] include a release note

PR #3296, opened 2024-10-17
Translate ``cpu.frequency`` hardware requirement

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/spec/hardware/cpu.fmf
+++ b/spec/hardware/cpu.fmf
@@ -41,8 +41,8 @@ description: |
           # Number or string, an ID of CPU vendor.
           vendor: 1234|"> 1234"
 
-          # Integer or string, CPU speed.
-          speed: 2300|">= 2300"
+          # Float or string, CPU frequency.
+          frequency: 2300.0|">= 2300.0"
 
           # Integer or string, CPU stepping.
           stepping: 10|">= 10"
--- a/spec/hardware/cpu.fmf
+++ b/spec/hardware/cpu.fmf
@@ -42,6 +42,7 @@ description: |
           vendor: 1234|"> 1234"
 
           # Float or string, CPU frequency.
+          # MHz are assumed when no unit is specified.
           frequency: 2300.0|">= 2300.0"
 
           # Integer or string, CPU stepping.
--- a/spec/hardware/cpu.fmf
+++ b/spec/hardware/cpu.fmf
@@ -41,6 +41,9 @@ description: |
           # Number or string, an ID of CPU vendor.
           vendor: 1234|"> 1234"
 
+          # Integer or string, CPU speed.
+          speed: 2300|">= 2300"
+
           # Integer or string, CPU stepping.
           stepping: 10|">= 10"
 
--- a/tmt/hardware.py
+++ b/tmt/hardware.py
@@ -1097,7 +1097,8 @@ def _parse_cpu(spec: Spec) -> BaseConstraint:
             'model',
             'family',
             'vendor',
-            'stepping'
+            'stepping',
+            'speed'
             )
         )
 
--- a/tmt/hardware.py
+++ b/tmt/hardware.py
@@ -747,7 +747,7 @@ class NumberConstraint(Constraint[float]):
             raw_value: str,
             original_constraint: Optional['Constraint[Any]'] = None,
             allowed_operators: Optional[list[Operator]] = None,
-            default_unit: Optional[Any] = 'MHz'
+            default_unit: Optional[Any] = None
             ) -> T:
 
         def _cast_number(raw_value: Any) -> float:
@@ -756,10 +756,6 @@ class NumberConstraint(Constraint[float]):
 
             if isinstance(raw_value, str):
                 raw_value = raw_value.strip()
-
-                if raw_value.startswith('0x'):
-                    return float(raw_value, base=16)
-
                 return float(raw_value)
 
             raise SpecificationError(f"Could not convert '{raw_value}' to a number.")
@@ -959,10 +955,11 @@ def _parse_int_constraints(
         ]
 
 
-def _parse_float_constraints(
+def _parse_number_constraints(
         spec: Spec,
         prefix: str,
-        constraint_keys: tuple[str, ...]) -> list[BaseConstraint]:
+        constraint_keys: tuple[str, ...],
+        default_unit: Optional[Any] = None) -> list[BaseConstraint]:
     """ Parse number-like constraints defined by a given set of keys, to float """
 
     return [
@@ -970,7 +967,8 @@ def _parse_float_constraints(
             f'{prefix}.{constraint_name.replace("-", "_")}',
             str(spec[constraint_name]),
             allowed_operators=[
-                Operator.EQ, Operator.NEQ, Operator.LT, Operator.LTE, Operator.GT, Operator.GTE])
+                Operator.EQ, Operator.NEQ, Operator.LT, Operator.LTE, Operator.GT, Operator.GTE],
+            default_unit=default_unit)
         for constraint_name in constraint_keys
         if constraint_name in spec
         ]
@@ -1156,12 +1154,13 @@ def _parse_cpu(spec: Spec) -> BaseConstraint:
             )
         )
 
-    group.constraints += _parse_float_constraints(
+    group.constraints += _parse_number_constraints(
         spec,
         'cpu',
         (
             'frequency',
-            )
+            ),
+        default_unit='MHz'
         )
 
     group.constraints += _parse_text_constraints(
--- a/tmt/hardware.py
+++ b/tmt/hardware.py
@@ -701,8 +701,8 @@ class FlagConstraint(Constraint[bool]):
             )
 
 
-class NumberConstraint(Constraint[int]):
-    """ A constraint representing a dimension-less number """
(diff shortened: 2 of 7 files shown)


spec/hardware/cpu.fmf
```diff
@@ -41,6 +41,9 @@ description: |
           # Number or string, an ID of CPU vendor.
           vendor: 1234|"> 1234"
 
+          # Integer or string, CPU speed.
+          speed: 2300|">= 2300"
```
happz, line 45: What is "CPU speed"?

skycastlelily:
> What is "CPU speed"?

How quickly a cpu can process data, measured in HZ,ah,I should put this
information into spec?:)

martinhoyer:
> How quickly a cpu can process data, measured in HZ,ah,I should put this information into spec?:)

I believe in this context, we are using MHz.

skycastlelily:
Yeah, Mhz, Ghz, TBO, I don't use it for my team's testing, I'm just trying
to fill gaps.
My goal is: when beaker  users come , they find almost all they need are
already implemented☺️
Or, you think I should set a default unit for it?:)
Btw, for cpu.speed ,float is better,though int will fit,I guess, most of
users need, do you think I should add or extend
NumberConstraint(Constraint[int]) to have float support?

happz:
> > How quickly a cpu can process data, measured in HZ,ah,I should put this information into spec?:)
> 
> I believe in this context, we are using MHz. Also, I think it's best to do just Mhz to keep it simple? Or is there any use-case to do conversions et al, @happz ?

The good thing is, that we do not need to worry about conversions :) Thanks to Pint, `2.6 GHz` or `2400 MHz` should be easy to convert to the right unit when needed. Let the user input anything, make it clear the field expects hertz but GHz or kHz are perfectly fine, and convert it to whatever Beaker/AWS/Azure/GCP/whatever desires when constructing a recipe for that infrastructure (`value.to('GHz').magnitude`, for example).

It's like the `memory` key - user is not limited to one unit, as long as they stick to units describing a size of memory. And we say "Bytes are assumed when no unit is specified.". Let's do the same but with a unit one might use for CPU frequency, GHz, MHz, whatever we pick, as long as we let user know and use it accordingly in the code.

> Oh and it might be better to use "frequency", instead of "speed".

+1 for `frequency`, it's also clearer to me what it means. "speed" is pretty ambiguous, although it's widely used.

skycastlelily:
Updated^^

docs/releases.rst
```diff
@@ -18,7 +20,6 @@ option or with the environment variable ``TMT_FEELING_SAFE`` set
 to ``True``. See the :ref:`/stories/features/feeling-safe` section
 for more details and motivation behind this change.
 
-
```
happz, line 21: Unrelated change.

docs/releases.rst
```diff
@@ -4,10 +4,12 @@
     Releases
 ======================
 
-
 tmt-1.38.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+The :ref:`/plugins/provision/beaker` provision plugin gains support
+for :ref:`cpu.stepping</spec/hardware/cpu>` hardware requirements.
```
happz, line 11: Singular requirement, "... hardware requirement.".

tmt/hardware.py
```diff
@@ -731,12 +731,50 @@ def _cast_number(raw_value: Any) -> int:
...
+    """ A constraint representing a dimension-less float number """
+
+    @classmethod
+    def from_specification(
+            cls: type[T],
+            name: str,
+            raw_value: str,
+            original_constraint: Optional['Constraint[Any]'] = None,
+            allowed_operators: Optional[list[Operator]] = None,
+            default_unit: Optional[Any] = 'MHz'
```
happz, line 750: Most likely `None` would be better, `MHz` would make sense for just some of the requirements. Note that the original `NumberConstraint` did not have a default unit at all, it's just a dimensionless number most of the time. This one wouldn't be that different, and only *some* would need to specify this field.

tmt/hardware.py
```diff
@@ -904,11 +942,28 @@ def wrapper(spec: Spec, index: int) -> BaseConstraint:
...
+            f'{prefix}.{constraint_name.replace("-", "_")}',
+            str(spec[constraint_name]),
+            allowed_operators=[
+                Operator.EQ, Operator.NEQ, Operator.LT, Operator.LTE, Operator.GT, Operator.GTE])
+        for constraint_name in constraint_keys
+        if constraint_name in spec
+        ]
+
+
+def _parse_float_constraints(
```
happz, line 962: Should be `_parse_number_constraints()`, to match the `NumberConstraint` class name.

tmt/hardware.py
```diff
@@ -1097,7 +1152,15 @@ def _parse_cpu(spec: Spec) -> BaseConstraint:
...
-            'stepping'
+            'stepping',
+            )
+        )
+
+    group.constraints += _parse_float_constraints(
+        spec,
+        'cpu',
+        (
+            'frequency',
```
happz, line 1163: Here we would definitely specify the default unit, `MHz`, `_parse_number_constraints()` would then propagate it to `NumberConstraint` class. Note that this is similar to `memory`, but `memory` uses `SizeConstraint` where we do have a default unit already predefined, always `bytes`. For more generic "number" constraint, we can't predefined one, so we need to specify it when parsing a constraint.

tmt/hardware.py
```diff
@@ -731,12 +731,50 @@ def _cast_number(raw_value: Any) -> int:
...
+
+        def _cast_number(raw_value: Any) -> float:
+            if isinstance(raw_value, float):
+                return raw_value
+
+            if isinstance(raw_value, str):
+                raw_value = raw_value.strip()
+
+                if raw_value.startswith('0x'):
+                    return float(raw_value, base=16)
```
happz, line 761: Relict from `int`, right? There's no `float` written in hexadecimal form.

spec/hardware/cpu.fmf
```diff
@@ -108,8 +114,13 @@ example:
     cpu:
         hyper-threading: true
 
+  - |
+    # Request a CPU with specified stepping.
+    cpu:
+        stepping: 9
```
happz, line 120: Unrelated changes, `stepping`?

spec/hardware/cpu.fmf
```diff
@@ -108,8 +114,13 @@ example:
...
+
 link:
   - implemented-by: /tmt/steps/provision/artemis.py
     note: "``cpu.vendor``, ``cpu.vendor-name`` and ``cpu.hyper-threading`` not implemented yet"
   - implemented-by: /tmt/steps/provision/mrack.py
-    note: "``cpu.flag``, ``cpu.processors``, ``cpu.model``, ``cpu.cores``, ``cpu.model-name``, ``cpu.hyper-threading`` and ``cpu.vendor-name`` only"
+    note: "``cpu.flag``, ``cpu.processors``, ``cpu.model``, ``cpu.cores``, ``cpu.model-name``, ``cpu.hyper-threading``, ``cpu.stepping`` and ``cpu.vendor-name`` only"
```
happz, line 126: Doesn't mention `frequency`.

spec/hardware/cpu.fmf
```diff
@@ -41,6 +41,9 @@ description: |
           # Number or string, an ID of CPU vendor.
           vendor: 1234|"> 1234"
 
+          # Float or string, CPU frequency.
```
happz, line 44: Doesn't mention the default unit, see how `memory` is documented.

spec/hardware/cpu.fmf
happz: I see, stepping commits are included in this PR. You can edit this PR to change its base branch to the one providing stepping changes, that should remove stepping changes from this PR's diff.

skycastlelily:
That's because this is cpu-speed branch, cpu.frequency for mrack is
implemented on cpu-speed-mrack branch, which is based on this branch:)

skycastlelily:
>  You can edit this PR to change its base branch to the one providing
stepping changes

Updated and changed^^

happz:
Blocked on https://github.com/teemtee/tmt/pull/3295

tmt/hardware.py
```diff
@@ -731,12 +731,46 @@ def _cast_number(raw_value: Any) -> int:
...
             as_quantity=False,
-            as_cast=_cast_number,
+            as_cast=_cast_int,
             original_constraint=original_constraint,
             allowed_operators=allowed_operators
             )
 
 
+class NumberConstraint(Constraint[float]):
+    """ A constraint representing a dimension-less float number """
```
happz, line 741: `IntegerConstraint` remains dimension-less, but `NumberConstraint` isn't, we already use it for a constraint with a default unit :)

skycastlelily:
Ah,yes,and updated^^

happz (approved):

therazix (approved):

happz:
Unrelated Fedora 41 failures, merging.

PR #3309, opened 2024-10-22
Support multiple links in the `tmt link` command

psss:
In order to make the syntax consistent with the `tmt test create` command and to allow providing multiple links in a single command let's use a ``--link`` option instead of the first argument for provide the `relation:target` pairs.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] include a release note


janhavlin (approved):

therazix (approved):

thrix (approved):
LGTM

psss:
> Maybe try to add some basic tests? E.g at least check the error is raised when no link is provided?

Makes sense, added in 499d1632. Does it look ok, @seberm?

PR #3320, opened 2024-10-25
Fix missing Testing Farm link in release notes

thrix:
Some bad rebase caused the sentence to be not finished. Not sure where that happened, but let's fix it :)


psss (approved):
Originally there was a link to the `Testing Farm` section. Added back in 45a0995d.

PR #3325, opened 2024-10-29
Disable credentials persistence in Github checkout action

happz:
According to a couple of articles, the default should be `false`, but it's not, which makes the token exposed to actions that do not need it. According to a linter I tried just for fun, we should enforce it to close this hole.

[1] https://github.com/actions/checkout/issues/485
[2] https://github.com/woodruffw/zizmor

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,6 +12,8 @@ jobs:
       SKIP: no-commit-to-branch
     steps:
     - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
     - uses: actions/setup-python@v5
       with:
         python-version: |
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - name: Repository checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
           persist-credentials: false
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -20,6 +20,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - id: ShellCheck
         name: Differential ShellCheck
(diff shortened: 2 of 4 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -20,6 +20,7 @@ jobs:
         uses: actions/checkout@v3
```
martinhoyer, line 20: perhaps we can bump this to v4 as part of this?

martinhoyer (approved):

KwisatzHaderach (approved):

.github/workflows/shellcheck.yml
happz: I know very little about Gihub Actions - `persist-credentials` felt simple and understandable to me, but I have much worse idea about changes related to bumping actions to v4. I don't mind, but someone else need to decide.

lukaszachy (approved):

psss (commented):
Looks ok, but is this needed if we're not using any token to clone git repositories?

.github/workflows/pre-commit.yml
```diff
@@ -12,6 +12,8 @@ jobs:
       SKIP: no-commit-to-branch
     steps:
     - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
```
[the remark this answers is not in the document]
happz: Good point, I don't know much about this part. I suppose both answers can be argued: no, we don't use the Github token, therefore we don't need this setting. But also yes, we should use it, as it closes a potential security hole: we don't use the token now, but we may in the future, and we may easily forget to enable this feature, plus it speaks about "credentials", there may be other tokens, not just the GH one.

Despite us not using a Github token today, I for one would add the setting to prevent any issue in the future.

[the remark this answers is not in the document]
psss: Ok, sounds reasonable.

psss (approved):

.github/workflows/shellcheck.yml
happz: Aha! Now I get it: it's the last `actions/checkout` using `v3`. Fixing in e61ff666.

psss:
Checks, which are affected by this change are green, merging.

PR #3328, opened 2024-10-30
Mark AVC check as expected when running AVC check test on localhost

happz:
This is an unexpected consequence of checks being interpreted in 1.38:
tmt in TF runs a test, test runs another tmt which triggers an AVC
denial, on purpose, but thanks to all this happening on localhost, TF
tmt can notice the AVC denial, and report it as a failed check. Adding
`xfail` to prevent this.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/test/check/main.fmf
+++ b/tests/test/check/main.fmf
@@ -16,6 +16,12 @@ tier: 2
       - provision-local
       - provision-virtual
 
+    adjust+:
+      - when: provision_how == local
+        check:
+          - how: avc
+            result: xfail
+
 /watchdog:
     test: ./test-watchdog.sh
     duration: 15m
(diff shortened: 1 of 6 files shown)


tests/test/check/main.fmf
```diff
@@ -16,6 +16,12 @@ tier: 2
       - provision-local
       - provision-virtual
 
+    adjust+:
+      - when: provision_how == local
```
psss, line 20:
```suggestion
      - when: provision-how == local
```

Would you mind of we call the new `context` adjust dimension `provision-how`. The naming would be consistent with other lower-case keys we're using across the config.

psss (approved):
Thanks!

seberm (approved):

martinhoyer (approved):
Thanks!

psss:
Checked the failures, all are related to the `fedora-41` changes, merging.

PR #3329, opened 2024-10-30
Add support for the `--arch` option in `tmt try`

falconizmi:
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] mention the version
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -16,7 +16,8 @@ support for :ref:`system.model-name</spec/hardware/system>`,
 The ``tmt lint`` command now reports a failure if empty
 environment files are found.
 
-The ``tmt try`` command now supports the new :ref:`/stories/cli/try/option/arch` option.
+The ``tmt try`` command now supports the new
+:ref:`/stories/cli/try/option/arch` option.
 
 
 tmt-1.38.0
--- a/tmt/cli.py
+++ b/tmt/cli.py
@@ -1796,6 +1796,7 @@ def try_command(context: Context, **kwargs: Any) -> None:
     """
 
     tmt.trying.Try.store_cli_invocation(context)
+
     # Inject custom image and provision method to the Provision options
     options = _construct_trying_provision_options(context.params)
     if options:
@@ -1810,8 +1811,8 @@ def try_command(context: Context, **kwargs: Any) -> None:
     trying.go()
 
 
-def _construct_trying_provision_options(
-        params: Any) -> dict[str, Any]:
+def _construct_trying_provision_options(params: Any) -> dict[str, Any]:
+    """ Convert try-specific options into generic option format """
     options: dict[str, Any] = {}
 
     if params['image_and_how']:
--- a/tests/unit/test_try.py
+++ b/tests/unit/test_try.py
@@ -12,6 +12,8 @@ import tmt.cli
          {'image': 'fedora', 'how': 'virtual'}),
         ({'image_and_how': ('fedora@virtual',), 'arch': 'aarch64'},
          {'image': 'fedora', 'how': 'virtual', 'arch': 'aarch64'}),
+        ({'image_and_how': (), 'arch': 'aarch64'},
+         {'arch': 'aarch64'}),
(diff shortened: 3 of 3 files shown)


psss (changes requested):
Thanks for implementing this! Looks good. Just one nitpick.

tests/try/basic/test.sh
```diff
@@ -42,6 +42,13 @@ rlJournalStart
         rlAssertGrep "Run .* successfully finished. Bye for now!" $rlRun_LOG
     rlPhaseEnd
 
+    rlPhaseStartTest "Arch"
+        rlRun -s "./arch.exp"
+        rlAssertGrep "Let's try.*/plans/basic" $rlRun_LOG
+        rlAssertGrep "arch: aarch64" $rlRun_LOG
+        rlAssertGrep "Run .* successfully finished. Bye for now!" $rlRun_LOG
+    rlPhaseEnd
+
```
psss, line 51: The test is failing because it needs support for nested virtualization. One option could be to create a new `/tests/try/virtual` test, move this part and `exp` file there and tag it as `provision-only`. But I'm afraid, when executed under nested virtualization in Testing Farm it would be very slow. @thrix, @frantisekz, thoughts?

KwisatzHaderach (approved):

tests/try/basic/test.sh
thrix: It is slow, and as we talked about it, it should be rather checked via a unit test, which might be out of scope of this change. @happz promised to look at it afaik. For now I would just disble this test.

therazix (approved):

tests/try/basic/test.sh
falconizmi: As @happz promised and discussed in slack, I applied his suggestion and removed exp test

docs/releases.rst
```diff
@@ -12,6 +12,9 @@ The :ref:`/plugins/provision/beaker` provision plugin gains
 support for :ref:`system.model-name</spec/hardware/system>`
 hardware requirement.
 
+The ``tmt try`` command now supports the new :ref:`/stories/cli/try/option/arch` option
+backed by :ref:`hardware/arch</spec/hardware/arch>`.
```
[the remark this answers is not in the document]
psss: Thanks!

tmt/cli.py
```diff
@@ -1796,30 +1796,10 @@ def try_command(context: Context, image_and_how: str, arch: str, **kwargs: Any)
...
-        # Add guest architecture if provided
-        if arch:
-            options['arch'] = arch
-
-        # For 'connect' rename 'image' to 'guest'
-        if options.get('how') == 'connect':
-            options['guest'] = options.pop('image')
-
-        tmt.steps.provision.Provision.store_cli_invocation(context=None, options=options)
+    if context.params['image_and_how']:
```
psss, line 1800: Why are the options stored only when `image_and_how` is defined? What is the point of this condition?

happz: Good question :) https://github.com/teemtee/tmt/commit/5b7f89205#diff-2bcd3056539a32d0370d652cfc6c04bd19b674cae84c341c3be7fc51cc3c60e6R1615

psss: Yeah, the thing is that you can do `tmt try --arch aarch64` and this would not work. And I think this is a perfectly valid use case. You are ok with the default provision method and image but just want to specify the arch.

psss: It would be good to check options first and then store them in the invocation if there's anything. Also dont' you want to actually use the `kwargs`?

```python
options = _construct_trying_provision_options(kwargs)
if options:
    ...
```

falconizmi: Okay I edit it so we can specify just `arch` and I did use `context.params`, as for me it looks more readable

psss: Thanks, added a simple test case for this in 7e173580.

psss (approved):

PR #3331, opened 2024-10-30
Add subresults support for reportportal report plugin

seberm:
This PR adds support for tmt subresults for tmt reportportal plugin. It tries to map the subresults as child items of parent test result in a specific launch.

Please don't review this PR while it is marked as a draft.

More info:
- https://reportportal.io/docs/api/service-api/start-child-item-using-post-1

## TODOs:
- [x] Core functionality (import of tmt subresults into RP)
- [x] Find a way to map the subresults to parent test/step reportportal launch items
- [x] Should the import of subresults be optional via argument? Or via fmf test metadata proposed by @kkaarreell ? What should be the default behavior? 
- [x] Try to resolve `status:interrupted` issues for subresults in RP, why is this happening?
- [ ] There are still some problems in local reportportal tests (subresults need special handling) - will be handled by @4N0body5 
- [ ] Missing tests for beakerlib subresults (subresults reported by shell tests using the `tmt-report-result` command should be already covered) - Nice to have (can be added in follow-up PR)

Blocked by:
- https://github.com/teemtee/tmt/pull/3356
- https://github.com/teemtee/tmt/pull/3370
- https://github.com/teemtee/tmt/pull/3372

Related:
- https://issues.redhat.com/browse/TT-306

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -4,11 +4,17 @@
     Releases
 ======================
 
+
 tmt-1.43
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Add the ``--workdir-root`` option for the ``tmt clean images``
 command so that users can specify the directory they want.
+
+A new ``upload-subresults`` key has been introduced for the
+:ref:`/plugins/report/reportportal` plugin, allowing the import of
+tmt subresults as child test items into ReportPortal. This
+behavior is optional and is disabled by default.
 
 
 tmt-1.42.1
--- a/tests/report/reportportal/data/plan.fmf
+++ b/tests/report/reportportal/data/plan.fmf
@@ -3,12 +3,13 @@ summary: Testing the integration of tmt and Report Portal via its API
 discover:
     how: fmf
 provision:
-    how: local
+    how: container
 execute:
     how: tmt
 report:
     how: reportportal
     project: test_tmt
+    upload-subresults: true
 
 context:
     component: tmt
--- a/tests/report/reportportal/data/test.fmf
+++ b/tests/report/reportportal/data/test.fmf
@@ -14,3 +14,23 @@
 /weird:
     summary: An error encountered
     test: this-is-a-weird-command
+
+/subresults:
+    summary: Test the tmt subresults (ReportPortal child items)
+    test: |
+        tmt-report-result /subtest/good PASS
+        tmt-report-result /subtest/fail FAIL
+        tmt-report-result /subtest/weird WARN
+
+/subresults-restraint:
+    result: restraint
+    test: |
+        tmt-report-result /subtest-restraint/good PASS
+        tmt-report-result /subtest-restraint/fail FAIL
+        tmt-report-result /subtest-restraint/weird WARN
+
+# TODO: Enable test of subresults for beakerlib phases
+#/subresults-beakerlib:
+#    summary: Test the Beakerlib phases are propagated as tmt subresults
+#    framework: beakerlib
+#    test: ./beaker-phases-subresults.sh
--- a/tests/report/reportportal/test.sh
+++ b/tests/report/reportportal/test.sh
@@ -8,10 +8,19 @@ ARTIFACTS=$TMT_REPORT_ARTIFACTS_URL
 
 PLAN_PREFIX='/plan'
 PLAN_STATUS='FAILED'
-TEST_PREFIX='/test'
-declare -A test=([1,'uuid']="" [1,'name']='/bad'   [1,'status']='FAILED'
-                 [2,'uuid']="" [2,'name']='/good'  [2,'status']='PASSED'
-                 [3,'uuid']="" [3,'name']='/weird' [3,'status']='FAILED')
+TEST_PREFIX=''
+
+# TODO: Subresults for beakerlib
+declare -A test=([1,'uuid']="" [1,'name']='/test/bad'        [1,'status']='FAILED'
+                 [2,'uuid']="" [2,'name']='/test/good'       [2,'status']='PASSED'
+                 [3,'uuid']="" [3,'name']='/test/subresults' [3,'status']='FAILED'
+                 [4,'uuid']="" [4,'name']='/subtest/good' [4,'status']='PASSED'
+                 [5,'uuid']="" [5,'name']='/subtest/fail' [5,'status']='FAILED'
+                 [6,'uuid']="" [6,'name']='/subtest/weird' [6,'status']='FAILED'
+                 [7,'uuid']="" [7,'name']='/test/subresults-restraint/subtest-restraint/good' [7,'status']='PASSED'
+                 [8,'uuid']="" [8,'name']='/test/subresults-restraint/subtest-restraint/fail' [8,'status']='FAILED'
+                 [9,'uuid']="" [9,'name']='/test/subresults-restraint/subtest-restraint/weird' [9,'status']='FAILED'
+                 [10,'uuid']="" [10,'name']='/test/weird'      [10,'status']='FAILED')
 DIV="|"
 
 
@@ -56,15 +65,16 @@ function identify_suite(){
 # Read and verify reported test names and uuids from $rlRun_LOG
 #
 # GLOBALS:
-# >> $test_uuid[1..3], $test_fullname[1..3]
+# >> $test_uuid[1..10], $test_fullname[1..10]
 function identify_tests(){
     rlLog "Verify and get test data"
 
-    for i in {1..3}; do
+    for i in {1..10}; do
         test_fullname[$i]=${TEST_PREFIX}${test[$i,'name']}
 
+        rlAssertGrep " \(sub-\)\?test: ${test_fullname[$i]}" $rlRun_LOG
         rlAssertGrep "test: ${test_fullname[$i]}" $rlRun_LOG
-        test_uuid[$i]=$(rlRun "grep -m$i -A1 'test:' $rlRun_LOG | tail -n1 | awk '{print \$NF}' ")
+        test_uuid[$i]=$(rlRun "grep -m$i -A1 ' \(sub-\)\?test:' $rlRun_LOG | tail -n1 | awk '{print \$NF}' ")
         rlAssertNotEquals "Assert the test$i UUID is not empty" "${test_uuid[$i]}" ""
         test[$i,'uuid']=${test_uuid[$i]}
     done
@@ -77,9 +87,8 @@ function identify_tests(){
 # ARGUMENT:
 #   request_url
 function rest_api(){
-
     rlLog "REST API request (GET $1)"
-    response=$(curl --write-out "$DIV%{http_code}" --silent -X GET "$1" -H  "accept: */*" -H  "Authorization: bearer $TOKEN")
+    response=$(curl --write-out "$DIV%{http_code}" --silent -X GET "$1" -H  "Accept: */*" -H  "Authorization: Bearer $TOKEN")
 
     response_code=${response##*"$DIV"}
     response=${response%"$DIV"*}
@@ -102,7 +111,6 @@ rlJournalStart
         fi
     rlPhaseEnd
 
-
     echo -e "\n\n\n::   PART 1\n"
 
     rlPhaseStartTest "Core Functionality"
@@ -112,9 +120,9 @@ rlJournalStart
         # TMT RUN
         rlLogInfo "A run with default setup"
         rlRun -s "tmt run --id $run --verbose --all" 2
-        rlAssertGrep "url: http.*redhat.com.ui/#${PROJECT}/launches/all/[0-9]{4}" $rlRun_LOG -Eq
+        rlAssertGrep "url: https?://.*\.redhat\.com/ui/#${PROJECT}/launches/all/[0-9]+" $rlRun_LOG -Eq
         identify_launch  # >> $launch_uuid, $launch_id
-        identify_tests   # >> $test_uuid[1..3], $test_fullname[1..3]
+        identify_tests   # >> $test_uuid[1..10], $test_fullname[1..10]
     rlPhaseEnd
 
 
@@ -150,10 +158,15 @@ rlJournalStart
         done
         rm tmp_attributes*
 
-        for i in {1..3}; do
+        for i in {1..10}; do
             echo ""
             test_name[$i]=${test[$i,'name']}
             test_name=${test_name[$i]}
+
+            # Strip the '/test' prefix from the name cause the tests are saved
+            # in `test.fmf` without this prefix.
+            test_name=${test_name#/test}
+
             test_fullname=${test_fullname[$i]}
             test_uuid=${test_uuid[$i]}
             test_status[$i]=${test[$i,'status']}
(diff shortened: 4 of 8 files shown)


kkaarreell:
I think it might be handy to control subresults import on a per test level through test fmf metadata.

seberm:
These are just some screenshots from my testing.

BTW: @kkaarreell Do you have any idea why the subresult items have the `status:interrupted` tag assigned to them? I still can't figure out why is this happening.

### Launch overview (parent tmt results):
01-Screenshot 2024-11-19 at 12-56-28 Report Portal

### Beakerlib test detail (subresults / child test items):
02-Screenshot 2024-11-19 at 12-57-08 Report Portal-bkrlib-subresults

### Details of shell test subresults when calling `tmt-report-result` command multiple times:
03-Screenshot 2024-11-19 at 12-57-37 Report Portal-tmt-report-result-subresults

tmt/steps/report/reportportal.py
```diff
@@ -649,21 +651,52 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
+                                "type": "step",
+                                "startTime": subresult.start_time or subresult_start_time})
+
+                        child_item_uuid = yaml_to_dict(response.text).get("id")
+                        assert child_item_uuid is not None
+
+                        # Finish the child item
+                        response = self.rp_api_put(
+                            session=session,
+                            path=f"item/{item_uuid}",
```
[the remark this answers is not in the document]
seberm: Oh, great catch. How could I miss it? Fixed in https://github.com/teemtee/tmt/pull/3331/commits/5aa6562dac46bca4798996598e32f18b055b03e8, thanks!

kkaarreell:
I have tested the current code but it doesn't work (at least) for beakerlib tests. There are no logs for individual test phases (I am not even sure if beakerlib even produces them at all) and therefore there is nothing RP AA could work with or a reviewer look at when reviewing a particular failed phase. One can access logs for the whole test but because subresults are imported one can't assign a test result for a whole test as it needs to be done for individual phases. As a result, this feature is not usable ATM. As a precaution, phase reporting should be enabled only for tests that provide individual phase logs.

kkaarreell:
On the last screenshot there is a test phase failure but the error log is missing.

docs/releases.rst
```diff
@@ -5,6 +5,17 @@
...
 
+The :ref:`/plugins/report/reportportal` plugin now supports
+importing of tmt subresults as child test items.
+
+The :ref:`/plugins/report/reportportal` plugin now handles the
+timestamps for ``custom`` and ``restraint`` results correctly. It
+should never happen that the result's ``end-time`` will be higher
+then the ``start-time``. It should be also ensured that the start
+time of all launch items is the same or higher then the start time
+of a parent item/launch.
```
KwisatzHaderach, line 16: is this correct? you have `should never happen...end time...be higher than the start time`?

IMO it should be
```suggestion
should never happen that the result's ``start-time`` will be higher
than the ``end-time``. It should be also ensured that the end
time of all launch items is the same or higher than the start time
of a parent item/launch.
```

KwisatzHaderach (approved):
just one tiny nitpick

docs/releases.rst
seberm: Thanks! Fixed in "child" PR here: https://github.com/teemtee/tmt/pull/3356/commits/eb7b34aba30d749ac91ddf0c97ce2557452c46c5

This PR is based on top of `feature/reportportal-improvements` (https://github.com/teemtee/tmt/pull/3356) and still needs a rebase. I plan to do it as soon as we resolve all the issues there.

seberm:
Hello @kkaarreell,
> On the last screenshot there is a test phase failure but the error log is missing.

This should be fixed now.  I am now handling the failures for `SubResult` the same way as for `tmt.Result`. Would you find some time to review this PR and help to check the subresult data are imported into the ReportPortal as expected?

Just FYI: Some "local" tests are still broken, but I've reached out to @4N0body5 for help as my available time to work on TMT is now becoming limited :/.

I am also marking this PR as ready for review so everyone is aware that the core functionality is now ready to be tested.

subresults-failures-1-Screenshot 2025-01-21 at 11-58-37 Report Portal
subresults-failures-0-Screenshot 2025-01-21 at 11-57-33 Report Portal

seberm:
Hello @KwisatzHaderach , could you please take another look at this PR and help me with the review? There have been multiple rebases and fixes since your last review, which is why I'm re-requesting your input. Thanks.

KwisatzHaderach (approved):

tmt/steps/report/reportportal.py
```diff
@@ -436,13 +426,57 @@ def append_description(self, curr_description: str) -> str:
...
+            write_out_failures: bool = True) -> None:
+        """ Upload all result log files into the ReportPortal instance """
+
+        for index, log_path in enumerate(result.log):
+            try:
+                log = self.step.plan.execute.read(log_path)
+            except tmt.utils.FileError:
+                continue
+
+            level = "INFO" if log_path == result.log[0] else "TRACE"
```
happz, line 445: Or `index == 0`?

tmt/steps/report/reportportal.py
```diff
@@ -408,22 +403,17 @@ def get_defect_type_locator(self, session: requests.Session,
...
-                                json=json)
+        response = session.post(url=f"{self.url}/{path}", headers=self.headers, json=json)
         self.handle_response(response)
         return response
 
     def rp_api_put(self, session: requests.Session, path: str, json: JSON) -> requests.Response:
-        response = session.put(url=f"{self.url}/{path}",
-                               headers=self.headers,
-                               json=json)
+        response = session.put(url=f"{self.url}/{path}", headers=self.headers, json=json)
```
happz, line 416: Several unrelated changes, just reformatting the source code. Is that doing of pre-commit, can it be undone?

kkaarreell:
I have tested the current code and it seems to import test phase results to RP properly. Test phases are represented as individual sub-items and failing test phases have separate log which can be processed by RP auto-analysis.
IMHO it works as it should. I would just highly recommend to implement a switch for this functionality, ideally both on the plugin and test plan or test case (metadata) level and we should not enable this functionality by default as it may unneccessarily prolong test results review.

tmt/steps/report/reportportal.py
seberm: It was done by my editor and I commited it. Reverted: https://github.com/teemtee/tmt/pull/3331/commits/24c266a5288b0b59e765084a12697174e496511b

tmt/steps/report/reportportal.py
seberm: I think checking the index should also be fine. Like this?
- https://github.com/teemtee/tmt/pull/3331/commits/a65541e9cce8ff09b65e262762e035fb87deb4b8

Thanks

4N0body5:
So far I did only regression testing, but..
I failed to run the tmt reportportal tests, as the seem to be broken. There were errors before logging each use case and therefore it did not report nor finished the launches in the RP. Here are example errors writen out in the terminal:
```
/usr/share/beakerlib/testing.sh: line 1076: /usr/local/bin/rhts-report-result: No such file or directory
./test.sh: line 74: {1..10}: syntax error: operand expected (error token is “{1..10}”)
./test.sh: line 241: ((: content_index<: syntax error: operand expected (error token is "<") 
basename: missing operand
```
And when testing plugin manually, from [the original functionalities](https://docs.google.com/document/d/1FyxC17WBBdn-rhGWq6Pg-oBVUB0oU0gEJjF-JPFhhYE), there was problem with IDLE use case:
`cannot access local variable 'test_end_time' where it is not associated with a value`
I tested it either with local or guest tmt provisioning.

seberm:
Hello @4N0body5 ,

> `/usr/share/beakerlib/testing.sh: line 1076: /usr/local/bin/rhts-report-result: No such file or directory`

Could you please re-check that the `beakerlib` is installed on your system and in the guest? I am unable to reproduce this error. Also, I think you should make sure you are running the `-h local` provision method with the newly introduced `--feeling-safe` option:

```
The 'local' provision plugin requires '--feeling-safe' option.
```

I've tried testing with `-h container` and this works as expected:

```
cd tests/report/reportportal/data
python -m tmt run -a provision -h container report -h reportportal --no-ssl-verify --project osabart_personal --suite-per-plan tests -n 'test/subresults$' plans --default
```

Regarding the `cannot access local variable 'test_end_time' where it is not associated with a value` error - very good catch. I've tried to make sure the `test_end_time` is always defined. Should be fixed by: https://github.com/teemtee/tmt/pull/3331/commits/585e69738e35f1d2a6f59a3bb2dd7bc19f6121a0

Thanks for helping me with the testing!

tests/report/reportportal/data/test.fmf
```diff
@@ -14,3 +14,23 @@
...
+        tmt-report-result /subtest/weird WARN
+
+/subresults-restraint:
+    result: restraint
+    test: |
+        tmt-report-result /subtest-restraint/good PASS
+        tmt-report-result /subtest-restraint/fail FAIL
+        tmt-report-result /subtest-restraint/weird WARN
+
+# TODO: Enable test of subresults for beakerlib phases
```
thrix, line 32: @seberm a TODO item?

tests/report/reportportal/data/plan.fmf
```diff
@@ -3,7 +3,7 @@ summary: Testing the integration of tmt and Report Portal via its API
 discover:
     how: fmf
 provision:
-    how: local
+    how: container
```
thrix, line 6: What is the reason for this change?

tests/report/reportportal/test.sh
```diff
@@ -376,7 +410,12 @@ rlJournalStart
             test_case_id=$(yq -r ".\"$test_name\".id" test.fmf)
             [[ $test_case_id != null ]] && rlAssertEquals "Assert the test ${test_name} has a correct testCaseId" "$(echo $response | jq -r '.testCaseId')" "$test_case_id"
             echo "$response" | jq -r ".$jq_element" > tmp_attributes.json || rlFail "$jq_element listing into tmp_attributes.json"
-            rlAssertGrep "TMT_TREE" tmp_attributes.json
+
+            # FIXME: Ignore the check for subtests
```
thrix, line 414: a fixme item

tmt/steps/report/reportportal.py
```diff
@@ -658,68 +697,84 @@ def execute_rp_import(self) -> None:
...
+                item_status = "SKIPPED"
                 if result:
-                    # Shift the timestamp to the end of a test
-                    test_time = result.end_time or test_time
+                    # Shift the timestamp to the end of a test.
+                    test_end_time = result.end_time or test_end_time
+
+                    item_status = self.TMT_TO_RP_RESULT_STATUS[result.result]
 
                     # For each log
```
thrix, line 712: a leftover helper string?

tests/report/reportportal/data/plan.fmf
seberm: The reportportal tests are not designed to run in the GH CI. Instead, they should be executed locally by a developer against a specific ReportPortal instance. That's why I think it's preferable to run them in a container or using the '-h virtual' option. I am not "--feeling-safe" enough to run them locally :).

tmt/steps/report/reportportal.py
seberm: Removed, thanks.

thrix:
Discussion from the hacking session
* moved to 1.43 as it seems to need a bit more polishing, @4N0body5 could you help to finish this up in a mergable state for 1.43, in case @seberm would not have enought time please?
* we decided we want to make this opt-in, i.e. there should be a boolean field so users opt-in on reporting with subresults, default will be without subresults

4N0body5:
After discussion with @seberm, I successfully tested and approved the main functionality of subresult report.
However, we identified the following points for further action:
* Using 'provision -h local' on my PC caused major problems, preventing the report of subresults. It is to be determined whether this issue is reproducible or specific to my device.
*  Reportportal test requires significant attention, including fixing the phase `uuid-based rerun` stuck in an endless loop, addressing minor failures caused by additional subresult tests, rewriting magic constants and hard-coded test commands for scalability, and adding a new phase to support additional Beakerlib test. Here I can offer my help throughout following month.
* In addition to the option enabling report of subresults, the implementation requires additional improvement, specifically in the propagation of test item  attributes into subresult item's Item Details. It is mainly environment variables (parameters), tmt test ID (test case ID), test web link (code ref), which are uploaded in the test item's background but they do not appear in its UI due to how the RP works for root test items. Additionally, description and test attributes can be propagated as well, although these are displayed at least on the higher level. They can be either simply copied from the test level (description, test id, etc.) or updated based on the particular subresult/phase (environment variables, ie.). These improvements should be straightforward and can eventually be done in a follow-up PR, which I can manage.

seberm:
@thrix , thanks for the summary.

> we decided we want to make this opt-in, i.e. there should be a boolean field so users opt-in on reporting with subresults, default will be without subresults

I've added `--upload-subresults` option which makes the subresults import optional. By default, the subresults will not be imported. Please take a look at the change. Is it sufficient? Thanks.
- https://github.com/teemtee/tmt/pull/3331/commits/1e522228b8f213b8acb300b409006d494c9128b2

docs/releases.rst
```diff
@@ -4,6 +4,13 @@
     Releases
 ======================
 
+A new ``--upload-subresults`` option was introduced for the
```
happz, line 7: "A new `upload-subresults` key" - metadata view should be the primary one, i.e. we should speak about keys and omit the `--` prefix.

kkaarreell: Personally, I can't imagine using this feature myself without being able to enable/disable it on a test level. Let me explain. Most of my tests are "single test scenario" kind of tests, i.e. they are suppose to test one thing and despite I use Setup/Test/Cleanup structure typical for Beakerlib tests I do not expect to fail it due to multiple issues at once.
Having test phase reporting enabled would mean that any failure/bug in "Setup" would most likely result also in a failure in Test and maybe even in clean up. For me, it means that I need to review 2 or 3 failures instead of 1. Morever, there would be 2 or 3 different error logs, despite there is a single root cause.
The only situation where I would like to have this feature enabled would be for tests that iterate through multiple configurations, testing single configuration in a dedidated phase. Here, I can imagine that listing each phase individually may make more sense as it could allow to better target errors to phases.
However, a global switch won't let me to enable this feature of one type of tests and keep it disable for the rest. As a result, I won't be using it and I will hope that nobody will enable it for our Tier jobs.

kkaarreell: With that, I am not saying this PR can't be merged as it is. I am just trying to explain why additional development is more than appropriate since the benefit of this current implementation is IMHO questionable.

seberm: @happz Release not improved: https://github.com/teemtee/tmt/pull/3331/commits/7ca8451496ab9fb45042f12a7a923d582f26a258

seberm:
Hello @4N0body5 
to sync up. Hopefully, I've fixed the test for multiple test contacts. I probably reintroduced this fail by a bad rebase when fixing the conflicts against the `main` branch: https://github.com/teemtee/tmt/pull/3331/commits/258460e5eda8dada3b901e50148de10b0de7bb1a

The current status of the reportportal tests running inside of a `-h container` is (at least in my environment) the following. Hopefully, you will get the same results :).

My env:
```
provision
        how: container
        primary address: tmt-5wE-ifdrWDGm
        topology address: tmt-5wE-ifdrWDGm
        name: tmt-5wE-ifdrWDGm
        multihost name: default-0
        arch: x86_64
        distro: Fedora Linux 41 (Container Image)
        kernel: 6.9.7-1.qubes.fc32.x86_64
        package manager: dnf5
        selinux: no
        is superuser: yes
```

### Extended Functionality - HISTORY AGGREGATION

I am getting three fails for subresults.

```
...
:: [ 15:10:37 ] :: [   FAIL   ] :: Assert the previous item is not in history (Assert: "944762" should not equal "944762")
...
:: [ 15:10:38 ] :: [   FAIL   ] :: Assert the previous item is not in history (Assert: "944763" should not equal "944763")
...
:: [ 15:10:39 ] :: [   FAIL   ] :: Assert the previous item is not in history (Assert: "944764" should not equal "944764")
...
```


### Extended Functionality - NAME-BASED RERUN

Six fails, also caused by subresults.

```
...
:: [ 15:11:25 ] :: [   FAIL   ] :: Assert the test item has correct UUID (Assert: 'f34081dd-7e5f-46f6-95fb-2070242e8585' should equal '9fb03147-45ee-4d63-a25e-3bc023ae9ea1')
:: [ 15:11:25 ] :: [   FAIL   ] :: Assert the test item retry item has a correct UUID (Assert: 'null' should equal 'b2b7eff8-950c-43a0-a356-499c84145df0')
:: [ 15:11:25 ] :: [   FAIL   ] :: Assert the test item has correct UUID (Assert: '9fb03147-45ee-4d63-a25e-3bc023ae9ea1' should equal 'd9b1df1e-75ff-4ad4-bd4c-9bb71b9ec959')
:: [ 15:11:25 ] :: [   FAIL   ] :: Assert the test item retry item has a correct UUID (Assert: 'b2b7eff8-950c-43a0-a356-499c84145df0' should equal '384fa5f1-a44b-4672-84f2-f2fe31382e8b')
:: [ 15:11:26 ] :: [   FAIL   ] :: Assert the test item has correct UUID (Assert: 'd9b1df1e-75ff-4ad4-bd4c-9bb71b9ec959' should equal 'f34081dd-7e5f-46f6-95fb-2070242e8585')
:: [ 15:11:26 ] :: [   FAIL   ] :: Assert the test item retry item has a correct UUID (Assert: '384fa5f1-a44b-4672-84f2-f2fe31382e8b' should equal 'f34081dd-7e5f-46f6-95fb-2070242e8585')
...
```

### Extended Functionality - UUID-BASED RERUN

Two failures occur here, and then it enters an endless loop.

```
...
:: [ 15:11:53 ] :: [   FAIL   ] :: Assert the message of the info log is correct (Assert: 'Something bad happened!' should equal '')
:: [ 15:11:53 ] :: [   PASS   ] :: Assert the level of the info log is correct (Assert: 'ERROR' should equal 'ERROR')
:: [ 15:11:53 ] :: [   FAIL   ] :: Assert the message of the info log is correct (Assert: 'Something bad happened!' should equal '')
...
```

docs/releases.rst
seberm: Hello @kkaarreell ,
Do you have any ideas on how to implement this? I assume it would require defining an extra key for the test to control the upload of subresults into RP. Is that correct?

I'm concerned that this would require changes beyond this PR. However, I agree that having more granular control at the test level would be beneficial.

In the context of this PR, I would suggest keeping the global option/key that controls the subresults upload globally, with it being disabled by default. We can then introduce test-level control functionality in a follow-up PR

seberm: Just FYI:

We quickly discussed possible solutions with @happz. His idea was to add an `extra-reportportal.<something>` key to the test metadata to influence the subresults import behavior. These `extra-*` keys are [already supported by a test schema](https://github.com/teemtee/tmt/blob/5bf41986ed7ce0f886735c2cbcd543093c8c2c2c/tmt/schemas/test.yaml#L146), but is an undocumented feature used e.g. by nitrate (https://github.com/search?q=repo:teemtee/tmt%20extra-summary&type=code).

Also, the implementation/design of these `extra-` keys in the tmt should be improved (e.g. abstracted into a module) and documented.

happz: As discussed on hacking session, keys defined by plugins do not seem to be the desired solution. Instead, we agreed a key to control reporting of test subresults, plugin-agnostic, would be more fitting. `reportportal` and other plugins would then use it.

Filed https://github.com/teemtee/tmt/issues/3515 to track the addition of L1 key.

tests/report/reportportal/test.sh
happz: @seberm @thrix is there anything to address?

happz:
@seberm please, rebase when you find some spare time (`ruff format` has been added, most conflicts will be probably just formatting changes). Also, there are some unmarked checkboxes in the PR description, are there lose ends that need to be tackled before re-reviewing the PR & eventual merge?

seberm:
> @seberm please, rebase when you find some spare time (`ruff format` has been added, most conflicts will be probably just formatting changes). Also, there are some unmarked checkboxes in the PR description, are there lose ends that need to be tackled before re-reviewing the PR & eventual merge?

Hello @happz,
I've rebased the changes on top of the `main`. Hopefully, I haven't introduced any bugs :). I will test it as soon as I finish the system configuration on my new laptop.

About unmarked checkboxes - I've added more info into the PR description. The tests for beakerlib reportportal subresults are nice to have. The rest of the issues in the local reportportal tests will be handled by @4N0body5 (see https://github.com/teemtee/tmt/pull/3331#issuecomment-2621811524).

psss (approved):
Looks good and works are expected. Thanks much for implementing this! After discussion we've agreed that the necessary test adjustments will be done in a separate pull request once @4N0body5 has capacity. Added just one release note nitpick.

docs/releases.rst
```diff
@@ -4,6 +4,13 @@
     Releases
 ======================
 
+
+A new ``upload-subresults`` key has been introduced for the
+:ref:`/plugins/report/reportportal` plugin, allowing the import of
+tmt subresults as child test items into ReportPortal. This
+behavior is optional and is disabled by default.
```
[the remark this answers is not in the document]
psss: Moved in 2c03006e.

tests/report/reportportal/test.sh
psss: There are a couple more. Let's fix those during the proper fix of the test in the separate pull request as agreed.

lukaszachy:
rebased and squashed locally, push ---force.

If someone else will be doing the merge, please do it manually so commit hash is not modified. 
I'll rebase #3547 to this commit so there is no waste of time by running another round of CI jobs after this PR is merged.

lukaszachy:
[testing-farm:fedora-41-x86_64:provision](https://github.com/teemtee/tmt/pull/3331/checks?check_run_id=37805197153) fails only in tests using centos-stream-10 -> expected

lukaszachy:
[testing-farm:fedora-41-x86_64:full](https://github.com/teemtee/tmt/pull/3331/checks?check_run_id=37796084835) has passed in #3547, let's not waste evening time  -> merging

psss:
> After discussion we've agreed that the necessary test adjustments will be done in a separate pull request once @4N0body5 has capacity.

Filed a follow-up issue: #3590

PR #3332, opened 2024-10-31
Add warning for multiple same checks per test

martinhoyer:
Pull Request Checklist

* [X] implement the feature

### DIFF
--- a/tmt/checks/__init__.py
+++ b/tmt/checks/__init__.py
@@ -1,6 +1,7 @@
 import dataclasses
 import enum
 import functools
+from collections import Counter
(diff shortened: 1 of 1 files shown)


happz:
@martinhoyer what about making it a linter check? In the normalization callback, the context it can provide is much smaller: yes, it reports multiple checks of the same `how`, but for which test? The callback does not see its "parent" object, therefore it cannot share this info. A linter check on test-level could.

martinhoyer:
> @martinhoyer what about making it a linter check? In the normalization callback, the context it can provide is much smaller: yes, it reports multiple checks of the same `how`, but for which test? The callback does not see its "parent" object, therefore it cannot share this info. A linter check on test-level could.

Make sense, will look into it. Thanks!

PR #3334, opened 2024-10-31
Install essential requirements with its own order

happz:
It turns out essential requirements must be installed before regular phases start running, because these phases may need their requirements to even begin. In particular, we noticed `prepare/ansible` requires a Python interpreter, but the default `order` puts it before installation of (collected) requirements.

Pull Request Checklist

* [x] implement the feature


happz:
@psss attempt to address the essential requirement conundrum. I left out the documentation on purpose: I think various special `order` values deserve their own section, hopefully in 1.39, and this would fit in.

lukaszachy:
` /tests/try/basic` is failing:
- no /usr/bin/python for  './epel.exp'  (shouldn't that be solved by this PR?)
- changed messages for './install.exp' (f-41 got released, is that it?)

tmt/steps/prepare/__init__.py
```diff
@@ -100,6 +100,34 @@ def go(
...
 
 
+# Required & recommended packages
+#
+# Structures and code for collecting requirements for different guests
+# or their groups to avoid installing all test requirements on all
+# guests. For example, a test running on the "server" guest might
+# require package `foo` while the test running on the "client" might
+# require package `bar`, and `foo` and `bar` cannot be installed at the
+# sametime.
```
psss, line 110:
```suggestion
# same time.
```

psss (approved):
Thanks for fixing this so fast! Looks good, just one typo, not sure it's worth the respin.

tmt/steps/prepare/__init__.py
psss: Let's not block on that. Fixed in #3337.

psss:
Reviewed failures, all are related to the `fedora-41` changes.

PR #3336, opened 2024-10-31
Fix null example in report specs

happz:
`example` is a list of strings, as such it cannot be left empty, `null`. To reset the list to "no examples" state, use an empty list.

Pull Request Checklist

* [x] implement the feature


psss (approved):

psss:
This one is super-trivial, merging right away.

PR #2773, opened 2024-03-20
Add FIPS support to the prepare plugin

idorax:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

The second PR for #1062

Issue #1062, opened 2022-02-22
Introduce a new `prepare` plugin for common features


psss:
As the first step towards a consistent way to prepare guests for testing a new `feature` plugin for the `prepare` step was discussed. The expected config syntax looks like this:

```yaml
prepare:
    how: feature
    epel: enabled
    crb: disabled
```

Similarly there should be a L1 key `feature` which could be used
to enabled features for individual tests:

```yaml
test: ./test.sh
feature:
    epel: enabled
    fips: enabled
```

For the implementation an ansible role with playbooks covering individual features should be used.

psss:
> For the implementation an ansible role with playbooks covering individual features should be used.

There are some limitations for `ansible` implementation on `rhel` (e.g. only some modules available). Double-check this before implementing as we need this to be working reliable across all supported distros.

idorax:
@psss, I'll look into this issue.

psss:
> @psss, I'll look into this issue.

Great! As the first proof of concept you can start with creating just the new `feature` prepare plugin (we can keep the L1 functionality for later). The [install](https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/install.py) plugin can serve you as a good template for start.

idorax:
Before implementing this plugin, let's explain the main abbreviations listed above,
* `epel`: **Extra Packages for Enterprise Linux**. For details, please refer to doc [What is Extra Packages for Enterprise Linux \(or EPEL\)?](https://docs.fedoraproject.org/en-US/epel/#what_is_extra_packages_for_enterprise_linux_or_epel)
   > Extra Packages for Enterprise Linux (or EPEL) is a Fedora Special Interest Group that creates, maintains, and manages a high quality set of additional packages for Enterprise Linux, including, but not limited to, [Red Hat Enterprise Linux](https://docs.fedoraproject.org/en-US/quick-docs/fedora-and-red-hat-enterprise-linux/index.html) (RHEL), CentOS, Scientific Linux (SL), Oracle Linux (OL), AlmaLinux (AL) and Rocky Linux (RL) ......
* `crb`: **CodeReady Linux Builder**. For details, please refer to doc [Chapter 3. The CodeReady Linux Builder repository](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository)
[...]

PR #3344, opened 2024-11-06
Add FIPS prepare feature enabling FIPS mode on RHEL and CentOS

The-Mule:
This adds prepare feature for enabling FIPS mode. Notice that disabling FIPS mode is unsupported and hence not implemented. Moreover, keep in mind that the proper way of enabling FIPS mode is to install the system in FIPS mode. The procedure depicted in the Ansible playbook is the only way of enabling it after installation. Since the procedure requires modification of kernel command line, it only supports Guests that allows that (i.e. not containers). Last but not least, the procedure is only valid for RHEL/CentOS-Stream 8, 9 and 10.

This PR is based on #2773 since I am not allowed to continue the work there. 

Relates to #1062

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] modify the json schema
* [x] mention the version
* [x] include a release note


The-Mule:
FYI: Is there a way to inhibit feature on distributions other than RHEL and CentOS in feature.py?

happz:
> FYI: Is there a way to inhibit feature on distributions other than RHEL and CentOS in feature.py?

Via `ansible_distribution == "RedHat"` checks in `when`.

tmt/steps/prepare/feature.py
```diff
@@ -79,6 +80,23 @@ def disable(self) -> None:
     }
 
 
+class FIPS(ToggleableFeature):
+    NAME = 'fips'
+
+    def is_unsupported(self) -> bool:
+        return isinstance(self.guest, (tmt.steps.provision.local.GuestLocal,
+                          tmt.steps.provision.podman.GuestContainer))
```
happz, line 88: Class enumeration will fail as soon as someone comes up with their own plugin that would not support this operation. I'd recommend adding something like `localhost` attribute to `Guest` class, see https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/local.py#L21 - it's set to `False` by default, `localhost` plugins sets it to `True`, and code checks this flag - not the class - when it's deciding about localhost-ish features. So I guess we need something like `fips_support` or something similar, set it to `True` by default & `False` in these two classes.

The-Mule:
> > FYI: Is there a way to inhibit feature on distributions other than RHEL and CentOS in feature.py?
> 
> Via `ansible_distribution == "RedHat"` checks in `when`.
> 
> https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/feature/epel-enable.yaml#L23

Right, I understand that. I would prefer to abort sooner that in the playbook (throwing unsupported exception) but now thinking about that that I guess I will add a task to the playbook that will fail on other distros.

tmt/steps/prepare/feature.py
The-Mule: Let me think about it. Adding a new flag might not be needed in the end since there is basically an equivalence between fips_support and "is_not_container" and I can most likely detect that in the playbook as well by inspecting init process control groups. That might be more tmt-friendly solution.

happz: Just to be sure, will such autodetection work with Image mode, a container-ish runtime?

We might add the `is_container` or similar flag to `Guest` classes if it turns out it's needed for more than just FIPS. I'd like all involved plugins to see the same picture, i.e. one plugin detecting a fact one way while another one would use completely different steps to discover the same answer would be something I'd wish to address. Not a blocker for now though, just a thought.

The-Mule: So I actually add the support for enabling FIPS in containers too. The way how it "unofficially" works is that in a container you need the host kernel to be running in FIPS mode so the playbook only handles the userspace part. There are checks in the end of the playbook that the FIPS mode is correctly enabled and if the host is running in FIPS mode already it will pass and fail otherwise.

The way how it works is that the /proc/1/environ drives the detection of a container environment. By container I mean anything that does not have a control over kernelspace. I tried it in actual containers but I did not tried it anywhere else. I am not familiar with Image mode, is there some reasonably easy way to get a shell there to check /proc/1/environ content?

The-Mule:
> Out of curiosity: does it need to fail? I think we shouldn't dive into deep waters, but some day in the future, is FIPS applicable to other distros?

Actually, yes, at least for now. I suppose there are other distribution being FIPS certified but from the Red Hat bubble it is currently only RHEL (having actual validation) and CentOS (by inheritance). Even for Fedora FIPS mode is not really relevant, it might or might not work and if it does not then it is not a big deal. We don't really expect anybody running Fedora in FIPS mode.

In the future if someone would like to add support for other distros the currently playbook content can be easily moved into a standalone block and new block can be added for other distribution.

I actually have a test plan for the playbook (not sure where it would fit), it behaves right for container, minute and artemis provisioning backends when tried on Fedora-41, RHEL-7, RHEL-8, RHEL-9, RHEL-10-Beta, CentOS-7, CentOS-Stream-8 and Centos-Stream-9.

happz:
@The-Mule hello, please rebase when you find some spare time. There were changes in the `prepare/feature` area, namely features are now plugins, see how https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/feature/epel.py looks today.

The-Mule:
@happz Re-worked to match the current prepare features. Not tested yet, I'll try to do more testing tomorrow but would you take a look at the current version? I dropped the "disabled" method (since disabling fips is not supported) and hence it now tracebacks with NotImplemented, maybe the better approach would be the original one - to log warning?

The-Mule:
Moreover, I think it might be worth adding --fips to trying.py too...

tmt/steps/prepare/feature/fips-enable.yaml
```diff
@@ -0,0 +1,196 @@
+---
+- name: Enable FIPS mode on RHEL-7, RHEL-8, RHEL-9 and RHEL-10
```
thrix, line 2:
```suggestion
- name: Enable FIPS mode on RHEL / CentOS Stream
```

tmt/steps/prepare/feature/fips-enable.yaml
```diff
@@ -0,0 +1,196 @@
...
+  hosts: all
+
+  tasks:
+    - name: Fail on distribution other than RHEL or CentOS
```
thrix, line 6: CentOS is EOL, let's call it `CentOS Stream`

tmt/steps/prepare/feature/fips.py
```diff
@@ -0,0 +1,32 @@
...
+class FipsStepData(PrepareFeatureData):
+    fips: Optional[str] = field(
+        default='enabled',
+        option='--fips',
+        metavar='enabled',
+        help='Whether FIPS mode should be enabled')
+
+
+@provides_feature('fips')
+class Fips(Feature):
```
thrix, line 22: @happz is there a way to inject there more user facing documentation?

tmt/steps/prepare/feature/fips-enable.yaml
```diff
@@ -0,0 +1,196 @@
+---
```
thrix, line 1: Can we add a comment here to some official docs, where one could get more information about FIPS?

tmt/steps/prepare/feature/fips-enable.yaml
```diff
@@ -0,0 +1,196 @@
...
+
+    - name: Install crypto-policies-scripts and dracut-fips
+      ansible.builtin.dnf:
+        name:
+          - crypto-policies-scripts
+          - dracut-fips
+        state: present
+      when: ansible_facts['distribution_major_version'] | int > 7
+
+    # skip_ansible_lint is present int the following two tasks to prevent
```
thrix, line 71:
```suggestion
    # skip_ansible_lint is present in the following two tasks to prevent
```

thrix: @The-Mule couldn't we use the `package` module? Or does that also have issues?

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html

thrix (dismissed):
@The-Mule some comments from me

thrix:
@The-Mule I believe we will need a test for this rather?

tmt/steps/prepare/feature/fips.py
happz: Yes, plugin can have a docstring, and that would be rendered in docs (the PR adding `prepare/feature` plugin docs to Sphinx). We don't have a docstring in the EPEL plugin, but we should - maybe we could leave it be here, merge this PR and the one rendering docs, and tackle both plugins at the same time.

The-Mule:
Thanks for the feedback, I'll incorporate most of it. Though I have a question about testing, I can't really test on images as epel feature does because by default playbook is expected to fail there (unless the system where tests are running is not in FIPS mode). What other options I have in the testsuite? Am I able to run my test plan using a different provisioning than container one?

happz:
> Thanks for the feedback, I'll incorporate most of it. Though I have a question about testing, I can't really test on images as epel feature does because by default playbook is expected to fail there (unless the system where tests are running is not in FIPS mode). What other options I have in the testsuite? Am I able to run my test plan using a different provisioning than container one?

Absolutely, you can use `virtual` provisioning. Check e.g. `/tests/prepare/install`, basically you add `provision-*` tags to `main.fmf` of the test - `provision-only` tells tmt to run it only with those provisioning methods, not as part of general runs - and then use `PROVISION_HOW` variable in the test. It will be set by tmt and you can run something else when it's `virtual`, or expect different behavior. If thins should fail with containers, you may have a branch for this plugin and expect the failure.

thrix:
Moved to 1.43, more changes needed here

The-Mule:
Yes, sorry for the delay, tests are taking me a bit more time than I anticipated. Moving to the next release sounds fine to me.

psss:
Let's try to finish by March 11? @The-Mule will you have some time?

The-Mule:
Yes, I really want to finish this one, I promise to get back to it this week.

The-Mule:
Here's the summary of changes I made:

1. Enabling FIPS is not supported for some environments such as container or distributions other than RHEL or CentosStream. Previously the logic of rejecting unsupported environment was done in the playbook but it made it too big and hard to navigate and understand. I move the logic to tmt instead. Now it fips prepare feature who is checking if distribution is eligible for enabling FIPS and checking whether the provisioned system is a container or not. I had to create one additional guest fact called 'container' that is set to 'no' if a system is not a container (or perhaps one not running systemd), if a system is a container (running systemd) the fact contains the value of `container` environment variable set by systemd (e.g. oci or podman).
2. I added both negative and positive tests.
3. Previously the playbook supported also RHEL-7, I reduced the support for RHEL/CentosStream 8, 9 or 10. We can add support for RHEL-7 later if need be. Focusing on 8, 9 and 10 made the playbook much simpler.

The-Mule:
Rebased to main and fixed issues in fips.py caused by the rebase. Documentation might need to be added too but it requires to add version. One more noteworthy thing is that there is a requirement for FIPS-compatible ssh keys when using this prepare feature (you might start with keys that are fine for non-FIPS but once you pass the reboot point in the playbook you need FIPS-compatible ssh key so that ansible can continue).

tmt/steps/provision/__init__.py
```diff
@@ -356,6 +356,7 @@ class GuestFacts(SerializableContainer):
     is_ostree: Optional[bool] = None
     is_toolbox: Optional[bool] = None
     toolbox_container_name: Optional[str] = None
+    container: Optional[str] = None
```
happz, line 359: At first I thought "`is_container` would be better", but then I realized it's actually a container *name*. I'd recommend changing the attribute then, to make it clear what it carries, `container_name`.

And we already have `toolbox_container_name`...

tmt/steps/prepare/feature/fips.py
```diff
@@ -0,0 +1,52 @@
...
+            raise tmt.utils.GeneralError(
+                'FIPS prepare feature is not supported on ostree systems.'
+            )
+        if guest.facts.container:
+            raise tmt.utils.GeneralError(
+                'FIPS prepare feature is not supported on container systems.'
+            )
+        if not guest.facts.distro or (
+            not re.compile('Red Hat Enterprise Linux (8|9|10)\\.').match(guest.facts.distro)
+            and not re.compile('Centos Stream (8|9|10)').match(guest.facts.distro)
```
happz, line 47: I really, really wanted to avoid this. It's a road to a maintenance hell and a pile of exceptions :/

Can the patterns be made global values, and possibly even one pattern instead of two, i.e. something like `SUPPORT_DISTRO_PATTERN = re.compile(...)`? Not hidden in the code and simplified to just pattern it would be easier to maintain.

Or a list of patterns, I guess it would be more readable if not squashed into one expression, so something like (and use `r`aw string for regular expressions):

```python
SUPPORTED_DISTRO_PATTERNS = (
    re.compile(pattern)
    for pattern in (
        r'Red Hat Enterprise Linux (8|9|10)\.',
        r'Centos Stream (8|9|10)'
    )
)

# and in the code:
if not any(pattern.match(guest.facts.distro) for pattern in SUPPORTED_DISTRO_PATTERNS):
    raise ...
```

tmt/steps/provision/__init__.py
The-Mule: Good point, I actually started with is_container and eventually get to container (name) but I am a little skeptical about the value of keeping the name (it does not really matter  and I'll get back to is_container again. If there will ever be a use case for actual container implementation, it can be added eventually.

The-Mule: Resolved in [7a3a09b](https://github.com/teemtee/tmt/pull/3344/commits/7a3a09bd9f5c367a92cbd4d0e895c6572f8d4aaf).

The-Mule:
Some of my tests are using centos-stream-9 and centos-stream-10 virtual images. Unfortunately they seem to be missing beakerlib that is part of virtual prepare apparently:

```
        prepare task #6: prepare-image on default-0
        how: shell
        summary: Fetch the image, refresh dnf cache, install beakerlib
        name: prepare-image
        order: 50
        overview: 1 script found
        script:
            tmt run --remove plan --default provision --how virtual finish
            for image in /var/tmp/tmt/testcloud/images/*qcow2; do
                virt-customize --add $image --run-command 'dnf --refresh install -y beakerlib'
            done
            cmd: /var/tmp/tmt/run-045/plans/provision/virtual/prepare/tree/tmt-prepare-wrapper.sh-prepare-image-default-0
            err: /var/tmp/tmt/run-048
            err:
            err: /default/plan
            err:     provision
            err:         queued provision.provision task #1: default-0
            err:
            err:         provision.provision task #1: default-0
            err:         how: virtual
            err:         memory: 2048 MB
            err:         disk: 40 GB
            err:         progress: booting...
            err:         multihost name: default-0
            err:         arch: x86_64
            err:         distro: Fedora Linux 41 (Cloud Edition)
            err:
            err:         summary: 1 guest provisioned
            err:     finish
            err:
            err:         summary: 0 tasks completed
            err:         guest: stopped
            err:         guest: removed
            out: [   0.0] Examining the guest ...
            out: [  11.8] Setting a random seed
            out: [  11.8] Running: dnf --refresh install -y beakerlib
            err: CentOS Stream 9 - BaseOS                        8.1 MB/s | 8.7 MB     00:01
            err: CentOS Stream 9 - AppStream                     6.5 MB/s |  23 MB     00:03
            err: CentOS Stream 9 - Extras packages                50 kB/s |  20 kB     00:00
            err: No match for argument: beakerlib
            err: Error: Unable to find a match: beakerlib
            err: CentOS Stream 9 - BaseOS                         60 kB/s | 6.3 kB     00:00
            err: CentOS Stream 9 - AppStream                      78 kB/s | 6.4 kB     00:00
            err: CentOS Stream 9 - Extras packages                88 kB/s | 7.6 kB     00:00
            err: No match for argument: beakerlib
            err: Error: Unable to find a match: beakerlib
            err: virt-customize: error: dnf --refresh install -y beakerlib: command exited with an error
            err:
            err: If reporting bugs, run virt-customize with debugging enabled and include the complete output:
            err:
            err:   virt-customize -v -x [...]
        fail: Command '/var/tmp/tmt/run-045/plans/provision/virtual/prepare/tree/tmt-prepare-wrapper.sh-prepare-image-default-0' returned 1.
```

This issue is not related to this PR but it blocks its testing (I don't even need beakerlib there). Clearly beaker-harness repository is missing on the image. Any idea how to fix it?

tmt/steps/prepare/feature/fips.py
```diff
@@ -0,0 +1,53 @@
...
+    fips: Optional[str] = field(
+        default=None,
+        option='--fips',
+        metavar='enabled',
+        help='Whether FIPS mode should be enabled',
+    )
+
+
+@provides_feature('fips')
+class Epel(ToggleableFeature):
```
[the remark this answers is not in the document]
The-Mule: Uh oh, fixed in [50fd879](https://github.com/teemtee/tmt/pull/3344/commits/50fd8796f798e0e266d29779450345832ff1b6af), thanks!

tests/prepare/feature/epel/data/plans.fmf
```diff
@@ -5,6 +5,7 @@ execute:
     how: tmt
 
 /epel:
+
```
happz, line 8: Unrelated change.

tmt/steps/prepare/feature/fips-enable.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+  tasks:
+
+    - name: Install crypto-policies-scripts and dracut-fips
+      ansible.builtin.dnf:
+        name:
+          - crypto-policies-scripts
+          - dracut-fips
+        state: present
+      when: ansible_facts['distribution_major_version'] | int < 10
```
happz, line 30: As already established by EPEL playbooks, we prefer using prepared variables rather than `ansible_facts`, e.g. https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/feature/epel-enable.yaml#L9 - `ansible_distribution_major_version | int < 10` would be aligned with existing code base.

The same would apply to other occurrences, like `ansible_facts['architecture']`.

happz (approved):

happz:
@The-Mule a couple of comments, and add a short release note to `docs/releases.rst`, please.

tests/prepare/feature/epel/data/plans.fmf
The-Mule: Fixed in e895bef.

tmt/steps/prepare/feature/fips-enable.yaml
The-Mule: Fixed in [4a5bf44](https://github.com/teemtee/tmt/pull/3344/commits/4a5bf443ef34cea7cab9185f85479a951a78a91c).

docs/releases.rst
```diff
@@ -29,6 +29,11 @@ demonstrating how one plan can be turned into many plans.
 The ``deployment-mode`` context dimension is now included in test run
 exports to Polarion.
 
+FIPS mode can now be enabled for RHEL or CentosStream 8, 9 or 10 by
```
therazix, line 32: The release note should be under `tmt-1.45.0`.

The-Mule: Fixed.

The-Mule:
All issues are hopefully resolved, all commit are organized (squashed into more coherent commits), rebased to main. Release note is updated.

Please check release note and please notice that I added --fips into tmt try today, the change is super-trivial but it would be good if you can check it out (I tested it briefly). It is completely contained in https://github.com/teemtee/tmt/pull/3344/commits/ad03758cbff05bc259935c2ef6f4597796ecec92.

docs/releases.rst
The-Mule: oh, mea culpa, thanks for a thorough check, fixing...

The-Mule: Done. Sorry again.

tmt/steps/prepare/feature/fips-enable.yaml
The-Mule: Resolved, all occurrences updated.

The-Mule:
@thrix, can you cancel [the change request](https://github.com/teemtee/tmt/pull/3344#pullrequestreview-2563331767) (was addressed)?

psss (approved):
Looks great! Thanks much for implementing this!

Issue #3277, opened 2024-10-10
tmt should handle error due to outdated dnf cache


kkaarreell:
From time to time I see issues like
https://artifacts.dev.testing-farm.io/9ad46c7d-ce18-4ca5-9b3f-49c26170d935/
```
       [101/124] rpm-sign-0:4.19.94-1.fc42.x86 100% | 112.0   B/s | 196.0   B |  00m02s
        >>> Status code: 404 for https://d2lzkl7pfhq30w.cloudfront.net/pub/fedora/linux/
        >>> Status code: 404 for http://mirror.us.leaseweb.net/fedora/linux/development/
        >>> Status code: 404 for http://mirrors.rit.edu/fedora/fedora/linux/development/
        >>> Status code: 404 for https://mirrors.rit.edu/fedora/fedora/linux/development
        >>> Status code: 404 for https://us.mirrors.cicku.me/fedora/linux/development/ra
        >>> Status code: 404 for http://us.mirrors.cicku.me/fedora/linux/development/raw
        >>> Status code: 404 for https://dl.fedoraproject.org/pub/fedora/linux/developme
        >>> Status code: 404 for http://dl.fedoraproject.org/pub/fedora/linux/developmen
        >>> No more mirrors to try - All mirrors were already tried without success
        [102/124] gdb-minimal-0:15.1-2.fc42.x86   0% | 378.9 KiB/s |  31.5 KiB |  00m00s
        >>> Not finished - interrupted by error: Cannot download Packages/r/rpm-sign-4.1
        [103/124] libverto-libev-0:0.3.2-9.fc41   0% |   0.0   B/s |   0.0   B |  00m00s
        >>> Not finished - interrupted by error: Cannot download Packages/r/rpm-sign-4.1Failed to download packages
         Librepo error: Cannot download Packages/r/rpm-sign-4.19.94-1.fc42.x86_64.rpm: All mirrors were tried
```
which are happening (I believe) due to a recent but not updated repodata cache and running dnf makecache would most likely fix it. 

`tmt` prepare.install plugin should be handle this situation, e.g. update dnf cache and rerun the command.

psss:
Should be relatively simple to implement, let's try in 1.38.

PR #3348, opened 2024-11-08
Refresh metadata cache after failed package installation

bajertom:
Fixes [#3277](https://github.com/teemtee/tmt/issues/3277)

When testing I found out that there is different behavior in dnf5 compared to dnf4, where a mere `dnf makecache` would be enough, but dnf5 needs the --refresh option as well (or `dnf clean metadata` or `dnf clean expire-cache`), see https://bugzilla.redhat.com/show_bug.cgi?id=2324177

Not sure about the Exception being too generic. Should I create a new one, something like InstallationError? And notify the user through logger as to why `metadata --refresh` is happening?

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -186,3 +186,6 @@ class PackageManager(tmt.utils.Common):
             *installables: Installable,
             options: Optional[Options] = None) -> CommandOutput:
         raise NotImplementedError
+
+    def refresh_metadata(self) -> CommandOutput:
+        raise NotImplementedError
--- a/tmt/package_managers/rpm_ostree.py
+++ b/tmt/package_managers/rpm_ostree.py
@@ -100,6 +100,22 @@ class RpmOstree(tmt.package_managers.PackageManager):
 
         return extra_options
 
+    def refresh_metadata(self) -> CommandOutput:
+        self.guest.warn("Metadata refresh is not supported with rpm-ostree.")
+
+        return CommandOutput(stdout=None, stderr=None)
+
+        # The following should work, but it hits some ostree issue:
+        #
+        #   System has not been booted with systemd as init system (PID 1). Can't operate.
+        #   Failed to connect to bus: Host is down
+        #   System has not been booted with systemd as init system (PID 1). Can't operate.
+        #   Failed to connect to bus: Host is down
+        #   error: Loading sysroot: exit status: 1
+        #
+        # script = ShellScript(f'{self.command.to_script()} refresh-md --force')
+        # return self.guest.execute(script)
+
     def install(
             self,
             *installables: Installable,
--- a/tmt/steps/prepare/install.py
+++ b/tmt/steps/prepare/install.py
@@ -165,6 +165,24 @@ class InstallBase(tmt.utils.Common):
 
     def install(self) -> None:
         """ Perform the actual installation """
+        try:
+            self._install()
+
+        except Exception as exc1:
+            # We do not have any special handling for exceptions raised by the following code.
+            # Wrapping them with try/except gives us a chance to attach the original exception
+            # to whatever the code may raise, and therefore preserve the information attached
+            # to the original exception.
+            try:
+                # Refresh cache in case of recent but not updated change do repodata
+                self.guest.package_manager.refresh_metadata()
+                self._install()
+
+            except Exception as exc2:
+                raise exc2 from exc1
+
+    def _install(self) -> None:
+        """ Helper method to perform the actual installation steps """
         if self.local_packages:
             self.prepare_install_local()
             self.install_local()
--- a/tests/unit/test_package_managers.py
+++ b/tests/unit/test_package_managers.py
@@ -440,6 +440,87 @@ def test_install(
     assert_output(expected_output, output.stdout, output.stderr)
 
 
+def _parametrize_test_refresh_metadata() -> \
+        Iterator[tuple[
+            Container,
+            PackageManagerClass,
+            str,
+            Optional[str]]]:
+
+    for container, package_manager_class in CONTAINER_BASE_MATRIX:
+        if package_manager_class is tmt.package_managers.dnf.Yum:
+            yield container, \
+                package_manager_class, \
+                r"yum makecache", \
+                'Metadata'
+
+        elif package_manager_class is tmt.package_managers.dnf.Dnf:
+            yield container, \
+                package_manager_class, \
+                r"dnf makecache -y --refresh", \
+                'Metadata cache created'
+
+        elif package_manager_class is tmt.package_managers.dnf.Dnf5:
+            yield container, \
+                package_manager_class, \
+                r"dnf5 makecache -y --refresh", \
+                'Metadata cache created'
+
+        elif package_manager_class is tmt.package_managers.apt.Apt:
+            yield container, \
+                package_manager_class, \
+                r"export DEBIAN_FRONTEND=noninteractive; apt update", \
+                'packages'
+
+        elif package_manager_class is tmt.package_managers.rpm_ostree.RpmOstree:
+            yield pytest.param(
+                container,
+                package_manager_class,
+                r"rpm-ostree refresh-md --force",
+                'Available',
+                marks=pytest.mark.skip(reason="refresh-md does not work with how tmt runs ostree container")  # noqa: E501
+                )
+
+        elif package_manager_class is tmt.package_managers.apk.Apk:
+            yield container, \
+                package_manager_class, \
+                r"apk update", \
+                'OK:'
+
+        else:
+            pytest.fail(f"Unhandled package manager class '{package_manager_class}'.")
+
+
+@pytest.mark.containers
+@pytest.mark.parametrize(('container_per_test',
+                          'package_manager_class',
+                          'expected_command',
+                          'expected_output'),
+                         list(_parametrize_test_refresh_metadata()),
+                         indirect=["container_per_test"],
+                         ids=CONTAINER_MATRIX_IDS)
+def test_refresh_metadata(
+        container_per_test: ContainerData,
+        guest_per_test: GuestContainer,
+        package_manager_class: PackageManagerClass,
+        expected_command: str,
+        expected_output: Optional[str],
+        root_logger: tmt.log.Logger,
+        caplog: _pytest.logging.LogCaptureFixture) -> None:
+    package_manager = create_package_manager(
+        container_per_test,
+        guest_per_test,
+        package_manager_class,
+        root_logger)
+
+    output = package_manager.refresh_metadata()
+
+    assert_log(caplog, message=MATCH(
+        rf"Run command: podman exec .+? /bin/bash -c '{expected_command}'"))
+
+    assert_output(expected_output, output.stdout, output.stderr)
+
+
 def _parametrize_test_install_nonexistent() -> \
         Iterator[tuple[Container, PackageManagerClass, str, Optional[str]]]:
 
--- a/tmt/package_managers/apk.py
+++ b/tmt/package_managers/apk.py
@@ -122,6 +122,12 @@ class Apk(tmt.package_managers.PackageManager):
 
         return results
 
+    def refresh_metadata(self) -> CommandOutput:
+        script = ShellScript(
+            f'{self.command.to_script()} update')
+
+        return self.guest.execute(script)
+
     def install(
             self,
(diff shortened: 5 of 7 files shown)


tmt/steps/prepare/install.py
```diff
@@ -321,6 +335,11 @@ def install_debuginfo(self) -> None:
             self.guest.package_manager.check_presence(
                 *[Package(f'{package}-debuginfo') for package in self.debuginfo_packages])
 
+    def make_cache(self) -> None:
+        """ Generate metadata cache """
+
+        self.guest.package_manager.makecache()
```
happz, line 341: * can this be the default implementation of `make_cache()` method?
* `make_cache()` calling `makecache()` - please, pick one naming scheme :)
* hm, and does `make_cache()` even have to exist? It does not seem to bring added value, `install()` can call `self.guest.package_manager.makecache()` directly.

tmt/steps/prepare/install.py
```diff
@@ -163,8 +163,22 @@ def install_from_url(self) -> None:
...
+        try:
+            self.perform_installation()
+        except Exception:
+            # Refresh cache in case of recent but not updated change to repodata
+            if str(self.guest.package_manager) in ["dnf", "dnf5", "yum"]:
+                self.make_cache()
+            # Retry the installation
+            self.perform_installation()
+
+    def perform_installation(self) -> None:
```
happz, line 180: I'd recommend a different name - `_install`, or `_do_install` or something similar. It should be less public, i.e. starting with `_` would be wise, and it should immediately resemble its "parent", `install`. `perform_installation` is way too public API-ish, and does not look like it's related to `install()`, quite contrary, it looks like a standalone operation.

tmt/steps/prepare/install.py
```diff
@@ -163,8 +163,22 @@ def install_from_url(self) -> None:
...
+    def make_cache(self) -> None:
+        """ Default base method to generate metadata cache"""
+
     def install(self) -> None:
         """ Perform the actual installation """
+        try:
+            self.perform_installation()
+        except Exception:
+            # Refresh cache in case of recent but not updated change to repodata
+            if str(self.guest.package_manager) in ["dnf", "dnf5", "yum"]:
```
happz, line 175: * I'd run `makecache()` in every case. We already know `apt` would benefit from this, Ubuntu and Debian images come without cache, `apt-get install tree` will fail without running `apt-get update` first. I believe we can find implementations for all supported package managers, even if it'd be a no-op.

tmt/package_managers/__init__.py
```diff
@@ -186,3 +186,6 @@ def install_debuginfo(
             *installables: Installable,
             options: Optional[Options] = None) -> CommandOutput:
         raise NotImplementedError
+
+    def makecache(self) -> CommandOutput:
```
happz, line 190: `makecache()` looks very yum-ish. `install` is easy, installs stuff, but what does `makecache()` do? We could be more verbose, e.g. `refresh_metadata` or something similar. There are many caches, often it's nothing but a buzzword, let's be more developer-friendly and announce what we do here :)

tmt/steps/prepare/install.py
bajertom: Now it is calling it directly. I remember mypy complaining about not recognizing dynamically added methods that are defined only in specific subclasses. `makecache` was declared only in Dnf class in dnf.py, so I wanted to make sure it is called only by `make_cache` method of InstallDnf class from install.py and not by any other package manager, which would throw an error I believe. 

Now there is refresh_metadata method in every package manager class, so this is no longer necessary and removed. 

(make_cache x makecache - it felt wrong to me to have two methods named the same in two .py files, one calling the other. But my solution was probably even worse :-) )

tmt/steps/prepare/install.py
bajertom: Fixed.

tmt/steps/prepare/install.py
bajertom: I've added all the other package managers.

tmt/steps/prepare/install.py
```diff
@@ -165,6 +165,15 @@ def install_debuginfo(self) -> None:
 
     def install(self) -> None:
         """ Perform the actual installation """
+        try:
+            self._install()
+        except Exception:
+            # Refresh cache in case of recent but not updated change do repodata
+            self.guest.package_manager.refresh_metadata()
+            self._install()
```
happz, line 173: I wonder whether should capture the first exception as well - this way, the original one is lost, with all its data. How about something like this?

```python
try:
    self._install()

except Exception as exc1:
    # We do not have any special handling for exceptions raised by the following code.
    # Wrapping them with try/except gives us a chance to attach the original exception
    # to whatever the code may raise, and therefore preserve the information attached
    # to the original exception.
    try:
        # Refresh cache in case of recent but not updated change do repodata
        self.guest.package_manager.refresh_metadata()
        self._install()

    except Exception as exc2:
        raise exc2 from ex1
```

bajertom: You're right, that seems like an unnecessary loss of data - better to catch them both. I used your code, if that's alright:)

tmt/package_managers/rpm_ostree.py
```diff
@@ -100,6 +100,12 @@ def _extra_options(
 
         return extra_options
 
+    def refresh_metadata(self) -> CommandOutput:
+        script = ShellScript(
+            f'{self.command.to_script()} update ')
```
KwisatzHaderach, line 105: I don't think there is a command for this in rpm-ostree, you can use `dnf update`, but `rpm-ostree update` does not exist

actually you can do `rpm-ostree upgrade --check`

bajertom: I just checked on Fedora Silverblue and it does exist:
```
tom@fedora:~$ rpm -q rpm-ostree
rpm-ostree-2024.8-1.fc41.x86_64
tom@fedora:~$ rpm-ostree update
Receiving metadata objects: 1611/(estimating) 7.2 MB/s 447.6 MB               1614 metadata, 6089 content objects fetched; 446119 KiB transferred in 65 seconds; 883.1 MB content written
Receiving metadata objects: 1611/(estimating) 7.2 MB/s 447.6 MB... done
Staging deployment... done
```
However it made changes to the system:
```
Upgraded:
  alsa-lib 1.2.12-2.fc41 -> 1.2.13-3.fc41
  alsa-sof-firmware 2024.09-1.fc41 -> 2024.09.1-1.fc41
....
  vim-data 2:9.1.825-1.fc41 -> 2:9.1.866-1.fc41
  vim-minimal 2:9.1.825-1.fc41 -> 2:9.1.866-1.fc41
Removed:
  perl-Mozilla-CA-20240730-1.fc41.noarch
Added:
  opensc-libs-0.26.0-1.fc41.x86_64
Run "systemctl reboot" to start a reboot
```

So you're right, `rpm-ostree upgrade --check` is the way to go.

```
$ rpm-ostree upgrade --check
1 metadata, 0 content objects fetched; 592 B transferred in 1 seconds; 0 bytes content written
Note: --check and --preview may be unreliable.  See https://github.com/coreos/rpm-ostree/issues/1579
AvailableUpdate:
        Version: 41.20241121.0 (2024-11-21T03:10:35Z)
         Commit: dc7f3ffdfe61b3ff6df6e0f0a69c16a8a783851c1e07cc6d37a25bbb23aae4a7
   GPGSignature: Valid signature by 466CF2D8B60BC3057AA9453ED0622462E99D6AD1
           Diff: 327 upgraded, 10 removed, 1 added
```

KwisatzHaderach (approved):

happz:
@bajertom please, add a unit test for this new operation. It can be fairly trivial, and it would verify those commands work without crashing. See e.g. https://github.com/teemtee/tmt/blob/main/tests/unit/test_package_managers.py#L420 or other unit tests.

To run new unit tests, the following command should work (with dev environment activated, and let's say you'd call the test `test_refresh_metadata`):

```shell
$ hatch -e dev shell
$ pytest -vv --showlocals tests/unit/test_package_managers.py::test_refresh_metadata
```

bajertom:
I've added unit tests covering the command. Since there was difference between cases in "Metadata C/cache C/created" in different distros but for the same package manager (yum) I've decided to just stick with "Metadata". Is this enough or should I catch a more specific part of the stdout in this case?

Same goes for apt - it depends on whether the system is up to date or not. It's either (debian test)
```
Hit:1 http://repolink reponame
Hit:2 http://repolink reponame
50 packages can be upgraded. Run 'apt list --upgradable' to see them.
```

before `apt update` or (ubuntu test)
```
Hit:1 http://repolink reponame
Hit:2 http://repolink reponame
All packages are up to date.
```

if the system is up to date, so I decided to go for `packages` as a keyword here. 

However I am lost with rpm-ostree packager. On my VM I can run `rpm-ostree upgrade --check`, but the unit test throws this error 
```
DEBUG    tmt:log.py:742 Run command: podman exec ec41e25298a91ed3dd829df46b889f4605f5b33862696297c3146063dd030f5a /bin/bash -c 'rpm-ostree upgrade --check'
INFO     tmt:log.py:742     cmd: rpm-ostree upgrade --check
DEBUG    tmt:log.py:742 environment
DEBUG    tmt:log.py:742 Command event: 2.91e-06 waiting for process to finish
INFO     tmt:log.py:742     err: System has not been booted with systemd as init system (PID 1). Can't operate.
INFO     tmt:log.py:742     err: Failed to connect to bus: Host is down
INFO     tmt:log.py:742     err: System has not been booted with systemd as init system (PID 1). Can't operate.
INFO     tmt:log.py:742     err: Failed to connect to bus: Host is down
INFO     tmt:log.py:742     err: error: Loading sysroot: exit status: 1
DEBUG    tmt:log.py:742 Command event: 0.1897 waiting for process completed
DEBUG    tmt:log.py:742 Command event: 0.19 waiting for stream readers
DEBUG    tmt:log.py:742 Command event: 0.1902 stdout reader done
DEBUG    tmt:log.py:742 Command event: 0.1903 stderr reader done
DEBUG    tmt:log.py:742 Command returned '1' (failure).
```

happz:
> I've added unit tests covering the command. Since there was difference between cases in "Metadata C/cache C/created" in different distros but for the same package manager (yum) I've decided to just stick with "Metadata". Is this enough or should I catch a more specific part of the stdout in this case?

LGTM. It's not perfect, which is true for other test cases in that file, we're fine. Success of the command itself is a good indication things went well.

> However I am lost with rpm-ostree packager. On my VM I can run `rpm-ostree upgrade --check`, but the unit test throws this error
> 
> ```
> DEBUG    tmt:log.py:742 Run command: podman exec ec41e25298a91ed3dd829df46b889f4605f5b33862696297c3146063dd030f5a /bin/bash -c 'rpm-ostree upgrade --check'
> INFO     tmt:log.py:742     cmd: rpm-ostree upgrade --check
> DEBUG    tmt:log.py:742 environment
> DEBUG    tmt:log.py:742 Command event: 2.91e-06 waiting for process to finish
> INFO     tmt:log.py:742     err: System has not been booted with systemd as init system (PID 1). Can't operate.
> INFO     tmt:log.py:742     err: Failed to connect to bus: Host is down
> INFO     tmt:log.py:742     err: System has not been booted with systemd as init system (PID 1). Can't operate.
> INFO     tmt:log.py:742     err: Failed to connect to bus: Host is down
> INFO     tmt:log.py:742     err: error: Loading sysroot: exit status: 1
> DEBUG    tmt:log.py:742 Command event: 0.1897 waiting for process completed
> DEBUG    tmt:log.py:742 Command event: 0.19 waiting for stream readers
> DEBUG    tmt:log.py:742 Command event: 0.1902 stdout reader done
> DEBUG    tmt:log.py:742 Command event: 0.1903 stderr reader done
> DEBUG    tmt:log.py:742 Command returned '1' (failure).
> ```

Never seen this one before, and there are other rpm-ostree tests where I'd expect this to appear. No idea so far.

Can you rebase the PR? That should make the patch easier to review and play with.

BeeGrech:
LGTM, no other ideas aside from what has been mentioned

happz:
Thanks for the input. Let's make it a no-op for now, with a warning, and we can get back to it later. I guess the way how tmt runs ostree containers could be improved, enabling the feature eventually.

happz (approved):

psss (approved):
Thanks for improving this!

psss:
Failing test is a network glitch, merging.

PR #3354, opened 2024-11-12
Support passing pubkeys to authorized_keys for `mrack`

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -918,7 +918,7 @@ class BeakerGuestData(tmt.steps.provision.GuestSshData):
         option='--public-key',
         metavar='PUBLICKEY',
         help="""
-             If set, the public keys will be put under ssh authorized_keys.
+             Public keys to add among authorized SSH keys.
              """,
         multiple=True,
         normalize=tmt.utils.normalize_string_list)
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -913,6 +913,16 @@ class BeakerGuestData(tmt.steps.provision.GuestSshData):
              Submitting user must be a submission delegate for the ``USERNAME``.
              """)
 
+    public_keys: Optional[list[str]] = field(
+        default=None,
+        option='--public-keys',
+        metavar='PUBLICKEYS',
+        help="""
+             If set, Beaker jobs will be submitted on behalf of ``USERNAME``.
+             Submitting user must be a submission delegate for the ``USERNAME``.
+             """,
+        normalize=tmt.utils.normalize_string_list)
+
 
 @dataclasses.dataclass
 class ProvisionBeakerData(BeakerGuestData, tmt.steps.provision.ProvisionStepData):
@@ -948,6 +958,7 @@ class CreateJobParameters:
     kickstart: dict[str, str]
     whiteboard: Optional[str]
     beaker_job_owner: Optional[str]
+    public_keys: Optional[list[str]]
     group: str = 'linux'
 
     def to_mrack(self) -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


tmt/utils/__init__.py
```diff
@@ -5611,7 +5611,8 @@ def normalize_string_dict(
 
     if isinstance(raw_value, dict):
         return {
-            str(key).strip(): str(value).strip() for key, value in raw_value.items()
+            str(key).strip(): (value if isinstance(value, (list, dict, tuple))
+                               else str(value).strip()) for key, value in raw_value.items()
```
happz, line 5615: Why this change? Well, why the changes to `normalize_dict_string` and kickstart fields?

happz: I see, I get it now. But that's now a bit of a problem with regard to the name of this function, right? It's no longer "string dict" as values are no longer just strings. It looks like we should have more specialized version of this function, to normalize kickstart, rather than mangling the existing one.

skycastlelily: Yeah, I know we should rename the function,and I was waiting for your comment^^, just to be clear, you think we should have a new function,named normalize_kickstart,right?^^

happz: Yep, `normalize_kickstart` now looks more fitting the situation.

skycastlelily: Updated^^
https://github.com/teemtee/tmt/pull/3354/commits/e01d3c78d65d25b842ca275d2e2b0766c0dbf80a

tmt/steps/provision/mrack.py
```diff
@@ -866,6 +866,16 @@ class BeakerGuestData(tmt.steps.provision.GuestSshData):
...
 
+    public_keys: Optional[list[str]] = field(
+        default=None,
+        option='--public-keys',
+        metavar='PUBLICKEYS',
+        help="""
+             If set, Beaker jobs will be submitted on behalf of ``USERNAME``.
+             Submitting user must be a submission delegate for the ``USERNAME``.
+             """,
+        normalize=tmt.utils.normalize_string_list)
```
happz, line 877: * `public_key` + `public-key` + `multiple=True`. There is already `key`/`--key` key, a singular form, and on the command line I can use it multiple times (`--key foo --key bar --public-key baz ...`)
* `default_factory=list` will be better: it's a list, and an empty list is as good indication of "no keys to install" as `None`, but it's simpler because it's still just a list. So, `list[str]` + `default_factory=list` should work perfectly.
* `help` seems to be copy-pasted & deserves an update.

skycastlelily: Updated^^ https://github.com/teemtee/tmt/pull/3354/commits/83d8781edb23209858ca8613802b30879941ca5b

happz: One more note, I'd recommend tweaking the help text a bit. How about the following?

```
Public keys to add among authorized SSH keys.
```

skycastlelily: Updated^^:https://github.com/teemtee/tmt/pull/3354/commits/29abf1631e29ae545142901d0d4b3bd6ff488101

KwisatzHaderach (approved):

tmt/steps/provision/mrack.py
```diff
@@ -914,6 +914,16 @@ class BeakerGuestData(tmt.steps.provision.GuestSshData):
              Submitting user must be a submission delegate for the ``USERNAME``.
              """)
 
+    public_key: list[str] = field(
+        default_factory=list,
+        option='--public-key',
+        metavar='PUBLICKEY',
```
martinhoyer, line 920:
```suggestion
        metavar='PUBKEY',
```

martinhoyer (approved):

happz (approved):

PR #3355, opened 2024-11-12
Get rid of an irrelevant avc failure

psss:
Does not make sense to investigate as `fedora-39` is going away in a couple of weeks.


martinhoyer (approved):
If it's going away, wouldn't it be better to use "INFO" or something instead of xfail?

psss:
> If it's going away, wouldn't it be better to use "INFO" or something instead of xfail?

Makes sense, updated in 890f1169.

PR #3345, opened 2024-11-06
Fix timestamp processing

kkaarreell:
Fixes timestamp processing, especially in cases when the timestamp is missing in results.yaml.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

PR #3356, opened 2024-11-12
Multiple reportportal plugin improvements and fixes

seberm:
This PR should fix [the problem](https://github.com/teemtee/tmt/issues/3360) with a possible reportportal error when using `result: restraint` or `result: custom`:

```
Received non-ok status code 406 from ReportPortal: {"errorCode":40025,"message":"Start time of child ['Tue Nov 12 11:10:19 UTC 2024'] item should be same or later than start time ['2024-11-12T11:36:35.007'] of the parent item/launch '
11810'"}
```

The PR fixes multiple bugs, for more info see:
- https://github.com/teemtee/tmt/issues/3360
- https://github.com/teemtee/tmt/pull/3356#issuecomment-2503345143

This error is probably caused by the [badly sorted result start times](https://github.com/teemtee/tmt/blob/1933c03437006ce7d72881ca8c3706d6d81a8f38/tmt/steps/report/reportportal.py#L387). It can happen the datetime formats in the list are not in the same format. The `Result.start_time` is in ISO format but the `self.time()` returns `str(int(time() * 1000))`.

### The basic reproducer

```
$ cat main.fmf
test: |
    tmt-report-result /test/good PASS
    tmt-report-result /test/fail FAIL
    tmt-report-result /test/weird WARN

result: restraint

duration: 5m
```

```
$ python -m tmt -r ~/tmt-learning/20 run -a report --how reportportal -v --force
```

### Other improvements
There are also small code improvements like the possibility to ignore the SSL verification, adding the usage of Python properties instead of class methods, or adding handling of `ResultOutcome.SKIP` outcome which I believe is missing from the mapping.

### TODOs:
- [x] Link the issue with the problem description as soon as it's created.

Pull Request Checklist

* [x] implement the feature
* [x] update reportportal schema for `ssl-verify` option
* [ ] extend the test coverage - Tests will be extended and fixed (only for local testing) as a part of https://github.com/teemtee/tmt/pull/3331
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -30,6 +30,13 @@ current working directory is set to the path of the last
 executed test, so that users can easily investigate the test
 code there and experiment with it directly on the guest.
 
+The :ref:`/plugins/report/reportportal` plugin now handles the
+timestamps for ``custom`` and ``restraint`` results correctly. It
+should never happen that the result's ``end-time`` will be higher
+then the ``start-time``. It should be also ensured that the start
+time of all launch items is the same or higher then the start time
+of a parent item/launch.
+
 
 tmt-1.39.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -32,9 +32,9 @@ code there and experiment with it directly on the guest.
 
 The :ref:`/plugins/report/reportportal` plugin now handles the
 timestamps for ``custom`` and ``restraint`` results correctly. It
-should never happen that the result's ``end-time`` will be higher
-then the ``start-time``. It should be also ensured that the start
-time of all launch items is the same or higher then the start time
+should never happen that the result's ``start-time`` will be higher
+than the ``end-time``. It should be also ensured that the end
+time of all launch items is the same or higher than the start time
 of a parent item/launch.
 
 
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -32,10 +32,10 @@ code there and experiment with it directly on the guest.
 
 The :ref:`/plugins/report/reportportal` plugin now handles the
 timestamps for ``custom`` and ``restraint`` results correctly. It
-should never happen that the result's ``start-time`` will be higher
-than the ``end-time``. It should be also ensured that the end
-time of all launch items is the same or higher than the start time
-of a parent item/launch.
+should never happen that the result's ``start-time`` will be
+higher than the ``end-time``. It should be also ensured that the
+end time of all launch items is the same or higher than the start
+time of a parent item/launch.
 
 
 tmt-1.39.0
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -32,10 +32,10 @@ code there and experiment with it directly on the guest.
 
 The :ref:`/plugins/report/reportportal` plugin now handles the
 timestamps for ``custom`` and ``restraint`` results correctly. It
-should never happen that the result's ``start-time`` will be
-higher than the ``end-time``. It should be also ensured that the
-end time of all launch items is the same or higher than the start
-time of a parent item/launch.
+should prevent the ``start-time`` of a result being higher than
+the ``end-time``. It should be also ensured that the end time of
+all launch items is the same or higher than the start time of a
+parent item/launch.
 
 
 tmt-1.39.0
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -12,7 +12,7 @@ import tmt.log
 import tmt.steps.report
 import tmt.utils
 from tmt.result import ResultOutcome
-from tmt.utils import field, yaml_to_dict
+from tmt.utils import field, format_timestamp, yaml_to_dict
 
 if TYPE_CHECKING:
     from tmt._compat.typing import TypeAlias
@@ -352,7 +352,8 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
 
     @property
     def datetime(self) -> str:
-        return datetime.datetime.now(datetime.timezone.utc).isoformat()
+        # Use the same format of timestramp as tmt does
+        return format_timestamp(datetime.datetime.now(datetime.timezone.utc))
 
     @property
     def headers(self) -> dict[str, str]:
@@ -466,14 +467,17 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
             self.warn("SSL verification is disabled for all requests being made to ReportPortal "
                       f"instance ({self.data.url}).")
 
+        # Use the current datetime as a default, but this is the worst case scenario
+        # and we should use timestamps from results log as much as possible.
         launch_time = self.datetime
 
         # Support for idle tests
         executed = bool(self.step.plan.execute.results())
         if executed:
-            # Launch time should be the earliest start time of all plans. The minimum over datetime
-            # strings will work, because the datetime in ISO format is designed to be
-            # lexicographically sortable.
+            # Launch time should be the earliest start time of all plans.
+            #
+            # The datetime *strings* are in fact sorted here, but finding the minimum will work,
+            # because the datetime in ISO format is designed to be lexicographically sortable.
             launch_time = min([r.start_time or self.datetime
                                for r in self.step.plan.execute.results()])
 
@@ -603,9 +607,11 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
                 self.verbose("uuid", suite_uuid, "yellow", shift=1)
                 self.data.suite_uuid = suite_uuid
 
+            # The first test starts with the launch (at the worst case)
+            test_time = launch_time
+
             for result, test in self.step.plan.execute.results_for_tests(
                     self.step.plan.discover.tests()):
-                test_time = self.datetime
                 test_name = None
                 test_description = ''
                 test_link = None
@@ -616,7 +622,10 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
                 if result:
                     serial_number = result.serial_number
                     test_name = result.name
-                    test_time = result.start_time or self.datetime
+
+                    # Use the actual timestamp or reuse the old one if missing
+                    test_time = result.start_time or test_time
+
                     # for guests, save their primary address
                     if result.guest.primary_address:
                         item_attributes.append({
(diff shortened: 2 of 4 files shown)


seberm:
The problem with `start-time`/`end-time` timestamps should be also resolved by existing PR by @kkaarreell :
- https://github.com/teemtee/tmt/pull/3345

seberm:
Hello @mildas, could you please test if the changes in this PR resolve your issue (#3360) and if the data are imported to ReportPortal as expected? Thanks!

kkaarreell:
I have just tested a plain import of a test plan with passing and failing tests and it worked well.
I also tried to import test results that originally revealed the timestamp issue and it went well too.
I do not plan any particular testing ATM, from my POV it works well.

seberm:
@kkaarreell 

> I have just tested a plain import of a test plan with passing and failing tests and it worked well.
I also tried to import test results that originally revealed the timestamp issue and it went well too.
I do not plan any particular testing ATM, from my POV it works well.

Thanks for helping me with the testing! Could you please add a check to this PR that you did a review?

Also, do you know anyone else who uses the ReportPortal and could us help with the testing of these changes? What about @KwisatzHaderach?

KwisatzHaderach:
> Also, do you know anyone else who uses the ReportPortal and could us help with the testing of these changes? What about @KwisatzHaderach?

not actually using the rp plugin, we do all our exports through data router using polarion xml

KwisatzHaderach (approved):

The-Mule:
@seberm Hi Otto, we found one more issues in the RP plugin, in method def _filter_log_per_size there is the following line:

```
return f"{header}{data[:settings.size.to('bytes').magnitude]}"
```

But settings.size.to('bytes').magnitude is float actually (e.g. 1000000.0) and it causes:

```

Cause number 1:

    slice indices must be integers or None or have an __index__ method

        File /usr/lib/python3.12/site-packages/tmt/base.py, line 3749, in go
          plan.go()
        File /usr/lib/python3.12/site-packages/tmt/base.py, line 2439, in go
          step.go()
        File /usr/lib/python3.12/site-packages/tmt/steps/report/__init__.py, line 121, in go
          cast(Union[Action, ReportPlugin[ReportStepData]], phase).go()
        File /usr/lib/python3.12/site-packages/tmt/steps/report/reportportal.py, line 670, in go
          message = _filter_log(log,
        File /usr/lib/python3.12/site-packages/tmt/steps/report/reportportal.py, line 94, in _filter_log
          log = log_filter(log, settings=settings)
        File /usr/lib/python3.12/site-packages/tmt/steps/report/reportportal.py, line 81, in _filter_log_per_size
          return f"{header}{data[:settings.size.to('bytes').magnitude]}"
```

The fix is easy - int that, ie. replace that with:

```
return f"{header}{data[:int(settings.size.to('bytes').magnitude)]}"
```

(consulted with Karel and Milos).

Since this is rather generic PR for RP plugin fixes, would it be possible to squeeze this in?

tmt/steps/report/reportportal.py
```diff
@@ -449,13 +461,24 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
 
         self.check_options()
 
-        launch_time = self.time()
+        # If SSL verification is disabled, do not print warnings with urllib3
+        if not self.data.ssl_verify:
+            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+            self.warn("SSL verification is disabled for all requests being made to ReportPortal "
+                      f"instance ({self.data.url}).")
```
happz, line 468: Not just for ReportPortal: warnings will remain disabled for all code from this point forward, i.e. all `report` plugins invoked after this instance, and all `finish` and cleanup code of tmt itself. Can it be undone when this plugin finishes its work? The safest would be wrapping it all in a `try`/`finally` block and undo the act before leaving the function, but it's already so indented and so spaghettified that this way may lead to an extremely ugly diff...

seberm: Thanks for the hint. You are right we should re-enable the urllib2 warnings.

I agree the diff would be huge due to another indent.

What about using a e.g. context manager? We could move most of the code from the `go()` method into the new method `f(..)` and optionally disable the warnings, something like:

```python
@contextlib.contextmanager
def supress_ssl_warnings(silent=True):
    if silent:
        try:
            # Disable SSL warnings here
            yield
        finally:
            # Enable SSL warnings
            ...
    else:
        yield

def go(...):
    ...
    with supress_ssl_warnings(silent=not self.data.ssl_verify):
        self.f(...)
```

The diff would be still big, but this should not require another indent.

seberm: Something like this https://github.com/teemtee/tmt/pull/3356/commits/83e4e7889af01ee0712b1e344ffb6e02be8fa2b4 ? What do you think?

happz: > Thanks for the hint. You are right we should re-enable the urllib2 warnings.
> 
> I agree the diff would be huge due to another indent.
> 
> What about using a e.g. context manager? We could move most of the code from the `go()` method into the new method `f(..)` and optionally disable the warnings

A context manager would be my choice as well, and yes, splitting the method into two should help manage the indentation. It's an ugly workaround, but the method itself deserves a cleanup, otherwise *everything* will result in an ugly indentation mess :/

tmt/utils/__init__.py
```diff
@@ -6347,3 +6348,30 @@ def is_url(url: str) -> bool:
     """ Check if the given string is a valid URL """
     parsed = urllib.parse.urlparse(url)
     return bool(parsed.scheme and parsed.netloc)
+
+
+@contextlib.contextmanager
+def suppress_warning(category: Optional[type[Warning]]) -> Iterator[None]:
```
happz, line 6354: I think we need to handle synchronization, I'm not 100 % sure but I suspect this filtering will not be thread-safe. 
```python
_suppress_warning_lock = threading.RLock()

def suppress_warning(...):
    if category is not None:
        with _suppress_warning_lock:
            try:
                warnings.simplefilter('ignore', category)
                yield
            finally:
                # Reset the warning to its default for a specific category
                warnings.simplefilter('default', category)
    else:
        yield
```

The reentrant lock should allow called code to disable its own warning if necessary, as long as it runs in the same thread.

seberm: After reading the [docs](https://docs.python.org/3.11/library/warnings.html), I think you are right. It seems that the functions in the `warnings` module are not thread-safe, and we should handle this.

Modified in: https://github.com/teemtee/tmt/pull/3356/commits/029ca37f6e579891fdf6d61b33f3a95afe5ba37f

tmt/steps/report/reportportal.py
seberm: Yes, I agree.

If you think the diff in this PR is becoming too large, I can separate the patches for `--no-ssl-verify/--ssl-verify` functionality into a separate PR (so we can merge the bug fixes more quickly). Otherwise, I would keep them here.

Thanks!

docs/releases.rst
```diff
@@ -20,6 +20,13 @@ calling the ``tmt-report-result`` or by calling beakerlib's
 ``rlPhaseEnd`` saved in ``results.yaml`` are now relative to the
 ``execute`` directory.
 
+The :ref:`/plugins/report/reportportal` plugin now handles the
+timestamps for ``custom`` and ``restraint`` results correctly. It
+should never happen that the result's ``start-time`` will be
```
martinhoyer, line 25:
```suggestion
should prevent the ``start-time`` of a result being higher 
```

tmt/utils/__init__.py
```diff
@@ -6347,3 +6351,31 @@ def is_url(url: str) -> bool:
...
+    .. code-block:: python
+
+        with suppress_warning(urllib3.exceptions.InsecureRequestWarning):
+            ...
+
+    """
+    if category is not None:
+        with _suppress_warning_lock:
+            try:
+                warnings.simplefilter('ignore', category)
```
martinhoyer, line 6375: Not saying this is wrong, just wonder if it wouldn't be possible/better to shoehorn the catch_warnings context manager https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings ?

martinhoyer (approved):

docs/releases.rst
seberm: Fixed in https://github.com/teemtee/tmt/pull/3356/commits/5aa14401a4ca5aa143e4b976b7719b73ef3d6e96

Thanks

tmt/utils/__init__.py
seberm: The `warnings.catch_warnings` approach seems cleaner and more universal. What do you think? Could you please review?

- https://github.com/teemtee/tmt/pull/3356/commits/13eb123d9a762d8128f4cddd232059496c3aacf2

@happz Could you please also take a look? Thanks!

martinhoyer: @seberm awesome :)  

Wouldn't it make sense to use a different name? 

fwiw looking at warnings code: `if action == "all": return "always" # Alias` .. I'd probably remove the 'all' from `ActionType`
They also mention: "action: error, ignore, always, default, module, or once"

Out of curiosity, what would `ruff format` prefer:
```
with _catch_warning_lock, warnings.catch_warnings():
```
or
```
with _catch_warning_lock:
    with warnings.catch_warnings():
```
? Or is it ok with both? (can check myself later since I'm the one curious 🐕‍🦺)

tmt/utils/__init__.py
```diff
@@ -6280,3 +6281,33 @@ def is_url(url: str) -> bool:
...
     return bool(parsed.scheme and parsed.netloc)
+
+
+# Handle the thread synchronization for the `catch_warnings(...)` context manager
+_catch_warning_lock = RLock()
+ActionType = Literal['default', 'error', 'ignore', 'always', 'all', 'module', 'once']
+
+
+@contextlib.contextmanager
+def catch_warnings(
```
happz, line 6292: `suppress_warnings` seemed better for me, as we don't "catch" those warnings, as in "try -> except -> do something about the error", we suppress them.

tmt/utils/__init__.py
happz: LGTM, one comment WRT naming, but that's all I have.

tmt/utils/__init__.py
happz: Ok, makes sense. I missed we're allowing more actions.

happz (approved):

tmt/utils/__init__.py
seberm: Hello,
- Removed the `all` alias from the `ActionType` (https://github.com/teemtee/tmt/pull/3356/commits/417272be4973ae4a84127497ca2c5132461c8166).
- Renamed the `utils.catch_warnings` to `utils.catch_warnings_safe` (https://github.com/teemtee/tmt/pull/3356/commits/61b34f3fe363f1089512d8972c4d58942dd5ca42). Other names could be the following (no strong opinion there):
    - `catch_warnings_threadsafe`
    - `thread_safe_catch_warnings`
    - `catch_warnings_mt` ("multithreaded")
    - `catch_warnings_mt_safe`

> Out of curiosity, what would ruff format prefer:
> ...

Ruff prefers one `with` statement with multiple context managers separated by a comma:
- https://docs.astral.sh/ruff/rules/multiple-with-statements/

psss (approved):
Nice! Thanks much for improving this.

psss:
Red jobs are irrelevant failures. I've reported the `epel-next-release` download problem here:

* https://pagure.io/fedora-infrastructure/issue/12326

The remaining failure seems to be a network glitch. Merging.

Issue #803, opened 2021-06-10
Enter the test directory after login to the guest


brakogh:
Hi,

I'm using following command to run test and login to test machine:
_tmt run --all -vvv provision --how minute --image rhel8 test --name . login_

Test run ok. After login, I'm in following directory:
/var/tmp/tmt/run-018/plans/all/tree/

Naturally, I'm trying to run test from "tree" directory, in my case /var/tmp/tmt/run-018/plans/all/tree/integration/php-mariadb, by running runtest.sh. This is failing because test can't find requested libraries. After dicussion with @lukaszachy we found out that I have to be using /var/tmp/tmt/run-018/plans/all/**discover/default/tests**/integration/php-mariadb instead of tree/integration/php-mariadb directory. Running ./runtest.sh from discover directory works as expected.
So can you do _cd /var/tmp/tmt/run-018/plans/all/discover/default/tests/_ to be in correct directory right after login?

lukaszachy:
I think we miss option to login immediately after failed test and cd into tests' workdir automatically.

Athwale:
We are still hitting this problem in 2024.

happz:
It seems this might be related to `tmt try` also not entering test path after logging in. Possible ideas: with a single test, jumping to its directory; with multiple `discover` phases, jumping to the last one's workdir? @falconizmi volunteered to take a look.

PR #3357, opened 2024-11-12
Go to discover tests directory after login

falconizmi:
Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -1959,17 +1959,20 @@ class Login(Action):
             try:
                 guest.push()
                 if not cwd:
+                    # Use path of the last executed test as the default
+                    # current working directory
                     worktree = self.parent.plan.worktree
                     tests = self.parent.plan.discover.tests()
                     test_path = tests[-1].path if tests else None
-                    try:
+                    if test_path is None or worktree is None:
                         cwd = worktree
-                        if worktree and test_path:
+                    else:
+                        try:
                             cwd = worktree.parent / "discover" / test_path.unrooted()
-                        guest.execute(tmt.utils.ShellScript("/bin/true"),
-                                      interactive=True, cwd=cwd, env=env)
-                    except RunError:
-                        cwd = worktree
+                            guest.execute(tmt.utils.ShellScript("/bin/true"),
+                                          interactive=True, cwd=cwd, env=env)
+                        except RunError:
+                            cwd = worktree
 
             except tmt.utils.GeneralError:
                 self.warn("Failed to push workdir to the guest.")
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -1958,16 +1958,16 @@ class Login(Action):
             # Attempt to push the workdir to the guest
             try:
                 guest.push()
-                worktree = self.parent.plan.worktree
-                test_path = self.parent.plan.discover.tests(
-                    )[-1].path if self.parent.plan.discover.tests() else None
+                if not cwd:
+                    worktree = self.parent.plan.worktree
+                    tests = self.parent.plan.discover.tests()
+                    test_path = tests[-1].path if tests else None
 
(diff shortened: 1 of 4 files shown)


tmt/steps/__init__.py
```diff
@@ -1958,7 +1958,16 @@ def _login(
...
+                worktree = self.parent.plan.worktree
+                test_path = self.parent.plan.discover.tests(
+                    )[-1].path if self.parent.plan.discover.tests() else None
+
+                if worktree and test_path:
+                    test_path = worktree.parent / "discover" / test_path.unrooted()
+                else:
+                    test_path = worktree
+
+                cwd = cwd or test_path
```
happz, line 1970: All these new lines above would be wasted if `cwd` is set. How about putting them under `if not cwd`?

tmt/steps/__init__.py
```diff
@@ -1958,7 +1958,16 @@ def _login(
             # Attempt to push the workdir to the guest
             try:
                 guest.push()
-                cwd = cwd or self.parent.plan.worktree
+                worktree = self.parent.plan.worktree
+                test_path = self.parent.plan.discover.tests(
+                    )[-1].path if self.parent.plan.discover.tests() else None
```
happz, line 1963: I'd recommend calling `self.parent.plan.discover.tests()` just once, it may be costly to collect all tests just to pick the last one.

falconizmi:
Solves issue #803 
for `tmt try`, if user run `tmt try --login`, it wont run any tests as its how login was implemented:
`-l, --login                     Log into the guest only, do not run any`
and he will be logged in to `worktree` path 

So the user needs to run `tmt run -tests NAME` then, when he will be asked what to do next, choose `login` option and in that case, he will be logged in test path

tmt/steps/__init__.py
```diff
@@ -1958,7 +1958,16 @@ def _login(
             # Attempt to push the workdir to the guest
             try:
                 guest.push()
-                cwd = cwd or self.parent.plan.worktree
+                if not cwd:
+                    worktree = self.parent.plan.worktree
+                    tests = self.parent.plan.discover.tests()
+                    test_path = tests[-1].path if tests else None
+
+                    if worktree and test_path:
```
KwisatzHaderach, line 1966: I wouldn't even check for the worktree variable, or alternatively you should include that condition on the else branch too

falconizmi: I needed to check it as I am trying to get `.parent` content from it and without that I would get mypy error
So in case if worktree is None, then I would raise exception.

KwisatzHaderach: yeah, I thought it would be like that, but don't really like not checking it again on the else branch

KwisatzHaderach (approved):

falconizmi:
Okay, I have here some failures because the test path, that I expect, is `/var/tmp/tmt/run-XXX/something/discover/default-0/`, but instead when we have this test case:
plan
```
summary: Basic smoke test
execute:
    how: tmt
    script: tmt --help
```
and run it with this tmt command:
`tmt run -a provision -h local --feeling-safe`
and after the run take a look in the directory  `/var/tmp/tmt/run-XXX/something/discover/`, there is no `default-0/` directery.
And in the check, there is after that command `tmt run -l login -c true`, which will fail, because it executes the last run that doesnt have folder `default-0/`
(or at least that what I think now)

martinhoyer (approved):
I don't have much experience with tmt try code, but tried it and this /bin/true approach seems to work :)

psss (changes requested):
Thanks for implementing this. Looks good. I just think that we do not have to exercise the `/bin/true` check when there are no tests. Adjusted in b67e1545. Please have a look if the change is ok. It would be also good to add at least a basic test coverage and mention this in release notes as this feature was requested by users and they should learn about this good news :-)

PR #3358, opened 2024-11-12
Handle invalid option in the `tmt-reboot` script

psss:
Fix a minor issues [detected](https://openscanhub.fedoraproject.org/task/21963/log/tmt-1.39.dev888-1.20241112125004934013.pr3342.26.gffd43d85/scan-results.html) by the shell-check.

### DIFF
--- a/tmt/steps/execute/scripts/tmt-reboot
+++ b/tmt/steps/execute/scripts/tmt-reboot
@@ -18,6 +18,7 @@ while getopts "c:t:e" flag; do
         c) command="${OPTARG}";;
         t) timeout="${OPTARG}";;
         e) efi=False;;
+        *) exit 1;;
     esac
 done
 


tmt/steps/execute/scripts/tmt-reboot
```diff
@@ -18,6 +18,7 @@ while getopts "c:t:e" flag; do
         c) command="${OPTARG}";;
         t) timeout="${OPTARG}";;
         e) efi=False;;
+        *) exit 1;;
```
siteshwar, line 21: You may want to put a message there to the user about invalid argument.

siteshwar:
@psss Also, you may want to [disable checks](https://www.shellcheck.net/wiki/Ignore) on the warnings that you do not plan to fix. Thanks!

tmt/steps/execute/scripts/tmt-reboot
psss: This is already handled by `getopt`:

    tmt-reboot: illegal option -- x

So I guess we don't have to duplicate the error message.

psss:
> @psss Also, you may want to [disable checks](https://www.shellcheck.net/wiki/Ignore) on the warnings that you do not plan to fix. Thanks!

We already exclude certain paths in the github workflow:

https://github.com/teemtee/tmt/blob/938d8bc767ee77845c70fa8db6b3bce4dd0b8b47/.github/workflows/shellcheck.yml#L29-L33

Can we do a similar thing for `/usr/share/doc/tmt/examples/*` for openscanhub jobs?

siteshwar:
> @siteshwar, any hint here? Is there a config file for openscanhub where we could mark example files to be ignored?

@kdudka I could not find any related flags in `csmock` for this. Shall this be taken as a feature request?

@lachmanfrantisek @lachmanfrantisek This may need an extra flag on Packit side too.

kdudka:
Directories with tests to be suppressed in scan results can be specified via `--known-false-positives` but the feature is almost undocumented and works only locally.  I think we should migrate `csmock` to use `csfilter-kfp` from `csdiff` and make `csmock` take the `--kfp-git-url` option.

seberm (approved):

therazix (approved):

KwisatzHaderach (approved):

siteshwar:
> Directories with tests to be suppressed in scan results can be specified via `--known-false-positives` but the feature is almost undocumented and works only locally. I think we should migrate `csmock` to use `csfilter-kfp` from `csdiff` and make `csmock` take the `--kfp-git-url` option.

@psss You can add `<packagename>/exclude-paths.txt` file in https://github.com/openscanhub/known-false-positives to skip scanning certain paths.

PR #3361, opened 2024-11-13
Switch to the new `tmt` logo

martinhoyer:
Dependent on https://github.com/teemtee/docs/pull/9 See for additional context.
Cleaned up some (I presume) obsolete logo-related-steps from conf.py

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -29,12 +29,6 @@ STEPS = discover provision prepare execute finish report
 # A list of `plugins/*.rst` files to generate
 PLUGIN_TARGETS = $(addsuffix .rst,$(addprefix plugins/,$(STEPS))) plugins/test-checks.rst
 
-# A source of logo picture for Sphinx docs favicon
-LOGO_SRC = https://raw.githubusercontent.com/teemtee/docs/main/logo/tmt-small.png
-
-# Local filepath to fetched logo
-LOGO_DST = _static/tmt-small.png
-
 .DEFAULT_GOAL := help
 
 .PHONY: help \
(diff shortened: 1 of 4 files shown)


docs/Makefile
```diff
@@ -26,12 +26,6 @@ STEPS = discover provision prepare execute finish report
 # A list of `plugins/*.rst` files to generate
 PLUGIN_TARGETS = $(addsuffix .rst,$(addprefix plugins/,$(STEPS))) plugins/test-checks.rst
 
-# A source of logo picture for Sphinx docs favicon
-LOGO_SRC = https://raw.githubusercontent.com/teemtee/docs/main/logo/tmt-small.png
-
-# Local filepath to fetched logo
-LOGO_DST = _static/tmt-small.png
-
```
psss, line 34: Seems ok to drop the logo download. Unless github refuses to serve the raw content to the regular users as well ;-) @happz, any concerns here?

psss (commented):
Thanks, looks good, two questions though.

docs/Makefile
happz: I believe the downloaded file was not used at all. It was replaced with the remote one in https://github.com/teemtee/tmt/commit/9d19728b084d184e5727ab443d222402f46e7bd9

docs/conf.py
```diff
@@ -217,12 +217,11 @@ def _load_theme(
...
 # of the sidebar.
-html_logo = 'https://raw.githubusercontent.com/teemtee/docs/main/logo/tmt-transparent-175x175.png'
+html_logo = 'https://raw.githubusercontent.com/teemtee/docs/main/logo/tmt-logo-dark-background.png'
 
 # The name of an image file (within the static path) to use as favicon of the
-# docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
-# pixels large.
-html_favicon = 'https://raw.githubusercontent.com/teemtee/docs/main/logo/tmt-favicon.ico'
+# docs.
+html_favicon = 'https://raw.githubusercontent.com/teemtee/docs/main/logo/tmt-logo-dark-background.svg'
```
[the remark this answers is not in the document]
psss: And I've found the answer on the chat:
> ha, it looks like all browsers now support .svg favicons. Much better than .ico

docs/Makefile
psss: Ack, and if github decides to stop serving the raw files, we'll fix it. Hope they will not.

Issue #3363, opened 2024-11-15
Polarion XML fails to generate due to xmlSAX2Characters: huge text node


KwisatzHaderach:
I have a somewhat large plan, taking some time and writing a lot of information into the logs. The plan finishes fine with an all pass, but then xml generation fails on
```    The generated XML output is not a valid XML file. Use `--verbose` argument to show the output.

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        xmlSAX2Characters: huge text node, line 172178, column 10 (<string>, line 172178)
```
This is using Polarion report plugin, but will very likely end the same way for junit.

seberm:
Hello @KwisatzHaderach ,
the tmt polarion report plugin uses the junit report code internally which uses the Jinja2 and LXML to generate the final JUnit/XUnit XML file. This means the junit plugin is also affected.

This problem seems to be related to `lxml.etree.XMLParser` which appears to have a limit on the size of text nodes it can handle.

I've quickly looked at [XMLParser options](https://lxml.de/api/lxml.etree.XMLParser-class.html) and there is an option `huge_tree` which could hopefully help:

> **huge_tree** - disable security restrictions and support very deep trees and very long text content (only affects libxml2 2.7+)

I've tried to reproduce the problem locally (for now without tmt) and the `huge_tree` option is effective:

Create a large text node

```
#!/usr/bin/env python3

import xml.etree.ElementTree as ET

large_text = 'a' * (10 * 1024 * 1024 + 1)  # 10MB + 1 byte
root = ET.Element('root')
root.text = large_text
tree = ET.ElementTree(root)
tree.write('large_xml_file.xml')
```

Try to parse the file with `huge_tree=False`:

```
$ cat parse-xml.py
#!/usr/bin/env python3

import lxml.etree as ET

parser = ET.XMLParser(huge_tree=False)
try:
    tree = ET.parse('large_xml_file.xml', parser)
except ET.ParserError as e:
    print(e)


$ ./parse-xml.py
Traceback (most recent call last):
  File "/home/user/Repos/tmt/./parse-xml.py", line 9, in <module>
    tree = ET.parse('large_xml_file.xml', parser)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/lxml/etree.pyx", line 3541, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1879, in lxml.etree._parseDocument
[...]

PR #3365, opened 2024-11-18
Enable the `huge_tree` option for the `lxml` parser

seberm:
This PR should resolve the problem with the LXML limit on the size of text nodes it can handle:
- #3363 

For more info about a `huge_tree` option, please see:
- https://lxml.de/api/lxml.etree.XMLParser-class.html

> `huge_tree` - disable security restrictions and support very deep trees and very long text content (only affects libxml2 2.7+)

Especially this LXML FAQ section about **security concerns**:
- https://lxml.de/FAQ.html#is-lxml-vulnerable-to-xml-bombs

> Is lxml vulnerable to XML bombs?
> 
> This has nothing to do with lxml itself, only with the parser of libxml2. Since libxml2 version 2.7, the parser imposes hard security limits on input documents to prevent DoS attacks with forged input data. Since lxml 2.2.1, you can disable these limits with the `huge_tree` parser option if you need to parse really large, trusted documents. All lxml versions will leave these restrictions enabled by default.
> 
> Note that libxml2 versions of the 2.6 series do not restrict their parser and are therefore vulnerable to DoS attacks.

### TODOs:
- [x] Try to test the test with huge output without lxml (with just jinja) to check if bottleneck is really lxml (schema or pretty print)
- [x] Enable the "huge output" test in the CI? -> yes
- [x] Add a test for XML deep trees (using a custom junit flavor)

Pull Request Checklist:
* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -358,8 +358,8 @@ def make_junit_xml(
         except etree.XMLSyntaxError as error:
             phase.verbose('rendered XML', xml_data, 'red')
             raise tmt.utils.ReportError(
-                'The generated XML output is not a valid XML file (SyntaxError). Use `--verbose` '
-                'argument to show the output.') from error
+                'The generated XML output is not a valid XML file. Use `--verbose` argument '
+                'to show the output.') from error
 
     # Do not be fooled by the `encoding` parameter: even with `utf-8`,
     # `tostring()` will still return bytes. `unicode`, on the other
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -312,8 +312,7 @@ def make_junit_xml(
     xml_parser_kwargs: dict[str, Any] = {
         'remove_blank_text': prettify,
         'huge_tree': True,
-        'schema': None,
-        }
+        'schema': None}
 
     # The schema check must be done only for a non-custom JUnit flavors
     if flavor != CUSTOM_FLAVOR_NAME:
@@ -335,10 +334,8 @@ def make_junit_xml(
 
     xml_parser = etree.XMLParser(**xml_parser_kwargs)
     try:
-        # S320: Parsing of untrusted data is known to be vulnerable to XML
-        # attacks.
+        # S320: Parsing of untrusted data is known to be vulnerable to XML attacks.
         tree_root: XMLElement = etree.fromstring(xml_data, xml_parser)  # noqa: S320
-
     except etree.XMLSyntaxError as e:
         phase.warn(
             'The generated XML output is not a valid XML file or it is not valid against the '
@@ -363,16 +360,16 @@ def make_junit_xml(
                 'The generated XML output is not a valid XML file. Use `--verbose` argument '
                 'to show the output.') from error
 
-    # Do not be fooled by the `encoding` parameter: even with `utf-8`,
-    # `tostring()` will still return bytes. `unicode`, on the other
-    # hand, would give us a string, but then, no XML declaration.
-    # So, we get bytes, and we need to apply `decode()` on our own.
+    # Do not be fooled by the `encoding` parameter: even with `utf-8`, `tostring()` will still
+    # return bytes. `unicode`, on the other hand, would give us a string, but then, no XML
+    # declaration.  So, we get bytes, and we need to apply `decode()` on our own.
     xml_output: bytes = etree.tostring(
         tree_root,
         xml_declaration=True,
         pretty_print=prettify,
-        # The 'utf-8' encoding must be used instead of 'unicode', otherwise
-        # the XML declaration is not included in the output.
+
+        # The 'utf-8' encoding must be used instead of 'unicode', otherwise the XML declaration is
+        # not included in the output.
         encoding='utf-8')
 
     return str(xml_output.decode('utf-8'))
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -311,6 +311,7 @@ def make_junit_xml(
 
     xml_parser_kwargs: dict[str, Any] = {
         'remove_blank_text': prettify,
+        'huge_tree': True,
         'schema': None,
         }
 
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -336,6 +336,7 @@ def make_junit_xml(
     try:
         # S320: Parsing of untrusted data is known to be vulnerable to XML attacks.
         tree_root: XMLElement = etree.fromstring(xml_data, xml_parser)  # noqa: S320
+
     except etree.XMLSyntaxError as e:
         phase.warn(
             'The generated XML output is not a valid XML file or it is not valid against the '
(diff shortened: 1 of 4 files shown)


seberm:
Hello @thrix, @happz, @psss ,
Do you have any objections or concerns about enabling the `huge_tree` option, especially from the TestingFarm perspective?

I think we should be perfectly fine with enabling it.

Thanks!

seberm:
@KwisatzHaderach Are you able to somehow test this change and let me know if it resolves your problem?

The processing of huge files using LXML can take a lot of resources (esp. CPU). We can always add an option/condition to completely bypass the LXML processing and use just Jinja2. The LXML is used to just check the XML schema for non-custom JUnit flavors and for prettifying the XML output.

KwisatzHaderach (approved):
didn't get a chance to test it yet, but approving since it looks good and we need this sooner rather than later

seberm:
So I gave it an another round of tests. Let's take following test as an example:

```
require:
  - python
result: fail
test: python -c "print((('a' * 1023) + '\n') * 1024 * 10)"
```

As you can see from the results below, enabling the `huge_tree=True` option makes sense, and I think it will help solve the #3363 issue.

### 1) LXML completely bypassed (uninstalled), using just Jinja to generate the JUnit
Takes around 43s on my machine to generate, but it works.

```
$ python -m tmt -r ~/tmt-learning/huge-out run -a report --how junit --file x    43.01s user 2.08s system 71% cpu 1:03.41 total
```

### 2) LXML enabled (schema validation + pretty print), `huge_tree=False`
Takes around 44s on my machine, it crashes on `huge text node` error.

```
$ python -m tmt -r ~/tmt-learning/huge-out run -a report --how junit --file x    44.30s user 2.47s system 69% cpu 1:06.87 total
```

### 3) LXML enabled (schema validation + pretty print), `huge_tree=True`
Takes also around 44s on my machine, it works without error.

```
$ python -m tmt -r ~/tmt-learning/huge-out run -a report --how junit --file x    44.22s user 2.20s system 71% cpu 1:04.93 total
```

## (another?) problem I was facing
If you change the test to the following command (similar amount of data as in previous test but no line ending):

```
test: python -c "print('a' * (10 * 1024 * 1024 + 1))"
```

The tmt is still "working" on generating the report and it seems like it will never finish. This behavior is the same with or without the LXML installed.

I would appreciate it if someone could help me to profile this test in the tmt, so we could decide if this is actually a problem that needs to be solved. Perhaps not? Any ideas?

seberm:
@psss @happz Do you think it's a good idea to enable the test mentioned above (`test: python -c "print((('a' * 1023) + '\n') * 1024 * 10)"`) in the CI? I am not sure if it's a good idea to slow down the CI in this way. Thanks.

martinhoyer (commented):
If I understand it correctly, we want large depth (above 256) to require huge_tree.

We played with it locally with:
```
DEPTH=257
OUTPUT_FILE="deep.xml"

# Start the XML file with a root tag
echo "<root>" > "$OUTPUT_FILE"

# Generate nested tags
for i in $(seq 1 $DEPTH); do
    printf "%0.s " >> "$OUTPUT_FILE" # Indent for readability (optional)
    echo "<tag$i>" >> "$OUTPUT_FILE"
done

# Close the nested tags in reverse order
for i in $(seq $DEPTH -1 1); do
    printf "%0.s " >> "$OUTPUT_FILE" # Indent for readability (optional)
    echo "</tag$i>" >> "$OUTPUT_FILE"
done

# Close the root tag
echo "</root>" >> "$OUTPUT_FILE"
```

And then I though we can catch the exception and only use it when it's necessary?
```
from lxml import etree

def parse_with_huge_tree(file_path):
    try:
        tree = etree.parse(file_path)
        print("XML parsed successfully (default settings).")
        return tree
    except etree.LxmlError as e:
        print(f"LxmlError encountered: {e}")
        print("Retrying with huge_tree=True...")
        try:
            parser = etree.XMLParser(huge_tree=True)
            tree = etree.parse(file_path, parser)
            print("XML parsed successfully with huge_tree=True.")
            return tree
        except etree.LxmlError as retry_error:
            print(f"Retry failed with huge_tree=True: {retry_error}")
        except Exception as retry_exception:
            print(f"Unexpected error during retry: {retry_exception}")
    except Exception as e:
        print(f"Unexpected error: {e}")

# Usage
parse_with_huge_tree('./deep.xml')
```
Make sense?

(and maybe print a warning that it's quite large and we are being forced to use huge_tree, security vulnerability, et cetera)

seberm:
Hi @mhoyer,
it's eighter "large tag depth" or/and "big text content in tag data".

Thanks for testing the deep trees. The deep XML trees in the tmt junit plugin can happen only with `custom` flavors. I will add a test for it. In the `default` flavor or when using the `polarion` jinja template for generating XUnit, this should never happen.

Regarding trying with `huge_tree=False` and retrying with `huge_tree=True`, considering that the `huge_tree` option will **still be effectively enabled**, I don't see a reason to complicate the code other than informing the user about potential security/DoS concerns or that it may consume more system resources and take longer. Or is there any more?

seberm:
@martinhoyer FYI, I've modified the code to enable the `huge_tree` option only if necessary and print the warning to the user:
- https://github.com/teemtee/tmt/pull/3365/commits/21cfe4d58fd559affbcd4549d46a71652a8e88e4

I did more testing and discovered this should not prolong the time spent on LXML operations. Most of time seems to be spent in the Jinja2 `render_template_file` method, but it's a different issue.

seberm:
Hello @psss,

>  you've intentionally removed -r from the tmt run command [here](https://github.com/teemtee/tmt/pull/3365/files?diff=unified&w=0#diff-76f4c6cc642bd53f7744efaa0fca41ce533ea57e6dcc6d47383c4e21604fafa0L12)?

So, I've re-checked the code and I think the removal of `-r` is exactly what I wanted.

I intend to execute the tests only once (and not to remove data in `--id $run_dir`) and then run the report step multiple times with various junit arguments to check the report plugin arguments and functionality. Or am I still getting it wrong?

Right now, it could still happen that the subsequent runs will start in in-progress workdirs due to the for loop over tmt execute methods:
- https://github.com/teemtee/tmt/pull/3365/files#diff-76f4c6cc642bd53f7744efaa0fca41ce533ea57e6dcc6d47383c4e21604fafa0R12
 
But because the test uses only a tmt execute method, everything should work just fine. I think we could remove the for loop over the execute methods because it currently seems unnecessary.

seberm:
> I did more testing and discovered this should not prolong the time spent on LXML operations. Most of time seems to be spent in the Jinja2 render_template_file method, but it's a different issue.

It seems I've found where the bottleneck is:
- https://github.com/teemtee/tmt/pull/3382

seberm:
Hello @martinhoyer ,
just FYI, after multiple discussions and testing with @KwisatzHaderach,
I have reverted the changes for retrying with `huge_tree=True` and printing the warning to the user. 

The reason is it's hard to detect the "huge text node" or "deep tree" errors between **various** lxml and libxml2 versions. The error messages and message types are a little bit different. Sometimes the error message is only in `error.msg`, sometimes it's in `error.error_log` list. And the exception type is still the same - `etree.XMLSyntaxError`.

To sum up, I don't think it's worth complicating the code just to print a warning. Let's just keep the `huge_tree` option always enabled.

KwisatzHaderach:
Absolutely agree, simpler is usually better, tested the current implementation, along with #3382, and it is working fine on data from TF run where I encountered the error previously.

martinhoyer (approved):

tmt/steps/report/junit.py
```diff
@@ -359,19 +357,19 @@ def _read_log_filter(log: Path) -> str:
...
+    # declaration.  So, we get bytes, and we need to apply `decode()` on our own.
     xml_output: bytes = etree.tostring(
         tree_root,
         xml_declaration=True,
         pretty_print=prettify,
-        # The 'utf-8' encoding must be used instead of 'unicode', otherwise
-        # the XML declaration is not included in the output.
+
+        # The 'utf-8' encoding must be used instead of 'unicode', otherwise the XML declaration is
+        # not included in the output.
```
happz, line 372: I see several changes to comments, just reformatting, no actual modifications of their text. Is that on purpose? The patch could be a bit smaller. Here, then lines 363 and 337.

seberm: Sure. One of the comments was actually changed - I added "SyntaxError" to the exception message to be more explicit. But right now I think this is unnecessary because we are raising `ReportError` from `XMLSyntaxError`. 

The last comment modification with `S320` message - I just think it's better to put it on one line. It's a very small change, but I can also revert it if you want.

I've modified my changes, thanks.

happz (approved):

happz:
Failures on rawhide are not related to the change, merging.

PR #3366, opened 2024-11-18
Handle GitHub API rate limits

martinhoyer:
As per discussion on chat. Trying to handle rate limits for GitHub api.
https://stackoverflow.com/questions/66522261/does-github-rate-limit-access-to-public-raw-files
```
fail: Failed to fetch remote playbook 'https://raw.githubusercontent.com/teemtee/tmt/main/tests/prepare/ansible/data/playbook.yml'.
403 Client Error: Forbidden for url: https://raw.githubusercontent.com/teemtee/tmt/22a46a4a6760517e3eadbbff0c9bebdb95442760/tests/core/env/data/vars.yaml
```
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/core/environment-file/test.sh
+++ b/tests/core/environment-file/test.sh
@@ -5,6 +5,11 @@ rlJournalStart
     rlPhaseStartSetup
         rlRun 'pushd data'
         rlRun 'set -o pipefail'
+        # Copy vars.yaml from env/data
+        rlRun 'cp ../../env/data/vars.yaml .'
+        # Start local server
+        python3 ../../../utils/http_server.py &
+        server_pid=$!
     rlPhaseEnd
 
     good="plan --name /plan/good"
@@ -46,15 +51,18 @@ rlJournalStart
         # Good
         rlRun -s "tmt plan show fetch/good"
         rlAssertGrep "STR: O" $rlRun_LOG
-        rlAssertGrep "INT: 0"  $rlRun_LOG
+        rlAssertGrep "INT: 0" $rlRun_LOG
         # Bad
         rlRun -s "tmt plan show fetch/bad" 2
-        rlAssertGrep "Failed to extract variables from URL 'http://localhost:8000/wrong.yaml'."  $rlRun_LOG
+        rlAssertGrep "Failed to extract variables from URL 'http://localhost:8000/wrong.yaml'." $rlRun_LOG
         rlAssertGrep "404 Client Error: Not Found for url: http://localhost:8000/wrong.yaml" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartCleanup
-        rlRun 'rm output'
+        # Kill the server
+        kill $server_pid
+        # Clean up copied files
+        rlRun 'rm vars.yaml output'
         rlRun 'popd'
     rlPhaseEnd
 rlJournalEnd
--- a/tests/core/environment-file/test.sh
+++ b/tests/core/environment-file/test.sh
@@ -49,8 +49,8 @@ rlJournalStart
         rlAssertGrep "INT: 0"  $rlRun_LOG
         # Bad
         rlRun -s "tmt plan show fetch/bad" 2
-        rlAssertGrep "Failed to extract variables from URL '.*/tests/core/env/data/wrong.yaml'."  $rlRun_LOG -E
-        rlAssertGrep "404 Client Error: Not Found for url: .*/tests/core/env/data/wrong.yaml" $rlRun_LOG -E
+        rlAssertGrep "Failed to extract variables from URL 'http://localhost:8000/wrong.yaml'."  $rlRun_LOG
+        rlAssertGrep "404 Client Error: Not Found for url: http://localhost:8000/wrong.yaml" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/prepare/ansible/test.sh
+++ b/tests/prepare/ansible/test.sh
@@ -20,6 +20,9 @@ rlJournalStart
 
         rlRun "pushd data"
         rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        # Start local server in current directory (data)
+        python3 ../../../utils/http_server.py &
+        server_pid=$!
     rlPhaseEnd
 
     while IFS= read -r image; do
@@ -55,6 +58,8 @@ rlJournalStart
     done <<< "$IMAGES"
 
     rlPhaseStartCleanup
+        # Kill the server
+        kill $server_pid
         rlRun "rm -rf $run" 0 "Removing run directory"
         rlRun "popd"
     rlPhaseEnd
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -14,7 +14,15 @@ import tmt.steps.provision
 import tmt.utils
 from tmt.result import PhaseResult
 from tmt.steps.provision import Guest
-from tmt.utils import Path, PrepareError, field, normalize_string_list, retry_session
+from tmt.utils import (
+    DEFAULT_RETRIABLE_HTTP_CODES,
+    ENVFILE_RETRY_SESSION_RETRIES,
+    Path,
+    PrepareError,
+    field,
+    normalize_string_list,
+    retry_session,
+    )
 
 
 class _RawAnsibleStepData(tmt.steps._RawStepData, total=False):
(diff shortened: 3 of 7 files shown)


tmt/utils/__init__.py
```diff
@@ -4358,6 +4358,33 @@ def increment(
 
                 raise GeneralError(message) from error
 
+        # Handle GitHub-specific responses
+        # https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+        response = kwargs.get('response')
```
happz, line 4363: is it possible to use `cast()` to introduce some types? See `error` above, it should be possible to do something like `response = cast(Optional[requests.Response], kwargs.get('response'))`. The same applies to `headers` below, `headers: dict[str, str]` should help linters & limit the amount of `Any` running around.

happz: And maybe `headers` we would get for free, `response.headers` should exist all the time, even if it'd be empty, i.e. no need for `getattr`?

martinhoyer: @happz Thanks, adding cast already found and issue :)  
I'm not sure about the headers, as that already has `CaseInsensitiveDict[str]` type.

happz: Yes, once linters become aware of the `response` type, I suppose they would get a better view of its attributes, and we wouldn't need to bother with `cast` calls.

psss (approved):
Looks good, thanks!

tmt/utils/__init__.py
```diff
@@ -4358,6 +4358,30 @@ def increment(
 
                 raise GeneralError(message) from error
 
+        # Handle GitHub-specific responses
+        # https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+        response = cast(requests.Response, kwargs.get('response'))
```
happz, line 4363: Must be `Optional`, `get()` may return `None`.

tmt/utils/__init__.py
happz: I would expect `kwargs.get('response')` returns `Any`, because of `**kwargs: Any`. As soon as `cast(requests.Response, ...)` is applied, mypy should be able to infer the rest, but without this hint, it'd all be just `Any`.

tmt/utils/__init__.py
```diff
@@ -4358,6 +4358,30 @@ def increment(
...
+        if response is not None and 'X-GitHub-Request-Id' in response.headers:
```
thrix, line 4364: @martinhoyer by negating this condition you could save an indendation below I believe

```
        if response is None or 'X-GitHub-Request-Id' not in response.headers:
                  return super().increment(*args, **kwargs)
                  
        headers = response.headers
        ....
```

tmt/utils/__init__.py
```diff
@@ -4358,6 +4358,30 @@ def increment(
...
+            headers = response.headers
+
+            # Handle rate limiting
+            if response.status_code in (403, 429):
```
thrix, line 4368: maybe also here ... 

```
            if response.status_code not in (403, 429):
                return super().increment(*args, **kwargs)
```

tmt/utils/__init__.py
```diff
@@ -4358,6 +4358,30 @@ def increment(
...
+                    wait_time = reset_time - int(time.time())
+                    if wait_time > 0:
+                        time.sleep(wait_time)
+                        return super().increment(*args, **kwargs)
+
+                # Secondary rate limit - respect Retry-After if provided
+                if 'Retry-After' in headers:
+                    retry_after = int(headers['Retry-After'])
+                    time.sleep(retry_after)
+                    return super().increment(*args, **kwargs)
```
thrix, line 4383: this line does not seem to be needed, 4385 will be executed next?

martinhoyer:
Thanks for the reviews, I'm trying to make it work first and understand what's happening, then I'll polish it, add tests and re-request reviews.

tmt/steps/prepare/ansible.py
```diff
@@ -147,7 +147,14 @@ def go(
                 root_path = self.step.plan.my_run.tree.root
 
                 try:
-                    with retry_session() as session:
+                    with retry_session(status_forcelist=(
+                            403,  # Frobidden (but Github uses it for rate limiting)
```
psss, line 151:
```suggestion
                            403,  # Forbidden (but Github uses it for rate limiting)
```

tmt/utils/__init__.py
martinhoyer: smart

tmt/utils/__init__.py
```diff
@@ -4328,11 +4335,42 @@ def send(  # type: ignore[override]
...
+        self.logger.debug("Response headers", dict(response.headers))
+        self.logger.debug("Response text", response.data.decode('utf-8'))
+
     def increment(
             self,
             *args: Any,
             **kwargs: Any
             ) -> urllib3.util.retry.Retry:
+        if self.logger:
+            self.logger.debug(f'increment: {args=} {kwargs=}')
```
happz, line 4372: This call should be dropped, it was in my patch just to trace those calls, but it's not needed.

tests/core/environment-file/test.sh
```diff
@@ -44,11 +44,11 @@ rlJournalStart
 
     rlPhaseStartTest "Fetch a remote file"
         # Good
-        rlRun -s "tmt plan show fetch/good"
+        rlRun -s "tmt plan show fetch/good -ddddvvvv"
```
happz, line 47: This should no longer be needed.

tmt/utils/__init__.py
happz: This one is still present.

tests/core/environment-file/test.sh
```diff
@@ -48,13 +53,16 @@ rlJournalStart
         rlAssertGrep "STR: O" $rlRun_LOG
         rlAssertGrep "INT: 0"  $rlRun_LOG
         # Bad
-        rlRun -s "tmt plan show fetch/bad" 2
-        rlAssertGrep "Failed to extract variables from URL '.*/tests/core/env/data/wrong.yaml'."  $rlRun_LOG -E
-        rlAssertGrep "404 Client Error: Not Found for url: .*/tests/core/env/data/wrong.yaml" $rlRun_LOG -E
+        rlRun -s "tmt plan show fetch/bad -ddddvvvv" 2
```
psss, line 56: Probably remove the debugging options?

tests/prepare/ansible/test.sh
```diff
@@ -38,9 +41,9 @@ rlJournalStart
 
                 # Run given method
                 if [ "$PROVISION_HOW" = "local" ]; then
-                    rlRun "tmt run -i $run --scratch -av provision -h $PROVISION_HOW           plan -n /$plan"
+                    rlRun "tmt run -i $run --scratch -av provision -h $PROVISION_HOW           plan -n /$plan -d"
```
psss, line 44: Remove here as well? Btw, I think, `-d` does not have any effect here anyway.

tmt/utils/__init__.py
```diff
@@ -4358,8 +4393,90 @@ def increment(
 
                 raise GeneralError(message) from error
 
+        # Handle GitHub-specific responses
+        # https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit
+        response = cast(urllib3.response.HTTPResponse, kwargs.get('response'))
+        if response is None and 'X-GitHub-Request-Id' not in response.headers:
```
psss, line 4399:
```suggestion
        if response is None or 'X-GitHub-Request-Id' not in response.headers:
```

psss (changes requested):
Just three minor issues, otherwise look good.

tmt/utils/__init__.py
martinhoyer: done

tests/prepare/ansible/test.sh
martinhoyer: done

tests/core/environment-file/test.sh
martinhoyer: done

tmt/utils/__init__.py
martinhoyer: done

tmt/utils/__init__.py
martinhoyer: done

tmt/utils/__init__.py
martinhoyer: done

tmt/utils/__init__.py
martinhoyer: aaand done

martinhoyer:
@CodiumAI-Agent /improve

psss (approved):

happz (approved):

PR #3369, opened 2024-11-19
Add missing colon to the `tmt-try` tldr page

martinhoyer:
@psss Sorry about missing this during the updates. I've hit this now with tldr-lint. The other two files are ok.


psss (approved):
Thanks!

PR #3370, opened 2024-11-19
Fix log paths for subresults loaded from tmt-report-results.yaml

seberm:
The result/subresult log file path loaded from the `tmt-report-results.yaml` file must be modified to be relative to `execute` directory the same way as for `custom` / `restraint` results (see [ExecutePlugin._process_results_partials(...)](https://github.com/teemtee/tmt/blob/72ca8602d49ae95c8ed2346964011773cb15b10f/tmt/steps/execute/__init__.py#L836)).

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


KwisatzHaderach (approved):

seberm:
> I quickly checked the generated `results.yaml` file and it does not seem to be correct. I used the following test:
> 
> ```
> cd tests/execute/restraint/report-result/data
> tmt run --all test --name output report --how html --open
> ```
> 
> Seems that while the subresult logs have been fixed and are working fine the main result logs are now broken.

Great catch @psss ! The problem is that the log path fix is applied two times - once by my changes in `_load_tmt_report_results_file()` and secondly in `_process_results_paritals()` which is used by `extract_tmt_report_results_restraint()`. That's why the `restraint` results had a bad path.

I've modified my changes to apply the fix only once in `_load_tmt_report_results_file()`. Now it seems it's working correctly for all our cases.

Could you (@psss @happz) please check again and help me with the testing? I do not want to break anything for the users :). Thanks!

KwisatzHaderach (approved):

therazix (approved):

martinhoyer (approved):

happz:
Unrelated rawhide failures, merging.

PR #3372, opened 2024-11-20
Use the `rhts-report-result` alias for reporting `beakerlib` subresults

seberm:
I was troubleshooting why the **logs** for beakerlib phases (aka tmt subresults) are not saved by `tmt-report-result` script and the reason is the following.

The beakerlib calls the command in `BEAKERLIB_COMMAND_REPORT_RESULT` variable in the following way:

```
$BEAKERLIB_COMMAND_REPORT_RESULT "$testname" "$result" "$logfile" "$score"
```

See the implementation of `rlReport` here:
- https://github.com/beakerlib/beakerlib/blob/5a85937f557b735f32996eb55cd6b9a33f3fe653/src/testing.sh#L1076

Suppose we use `tmt-report-result` as the value for `BEAKERLIB_COMMAND_REPORT_RESULT` (as a script name). In that case, the script will not be compatible with the third `$logfile` positional argument - it will just ignore it because it accepts the logfile provided only by the `-o/--outputFile` option by default.

We want to use the `rhts-report-result` alias because of a compatibility layer implemented by the `tmt-report-result` script itself. If the script gets called with `rhts-report-result` name, it will **accept** the third positional argument as a `logfile`:

- https://github.com/teemtee/tmt/blob/e7cf41d1fe5a4dcbb3270758586f41313e9462ec/tmt/steps/execute/scripts/tmt-report-result#L101

Related to:
- https://github.com/teemtee/tmt/pull/3200
- https://github.com/teemtee/tmt/pull/3331

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/frameworks/beakerlib.py
+++ b/tmt/frameworks/beakerlib.py
@@ -8,11 +8,14 @@ import tmt.steps.execute
 import tmt.utils
 from tmt.frameworks import TestFramework, provides_framework
 from tmt.result import ResultOutcome
-from tmt.utils import Environment, EnvVarValue, Path
+from tmt.utils import Environment, EnvVarValue, GeneralError, Path
 
 if TYPE_CHECKING:
     from tmt.base import DependencySimple, Test
     from tmt.steps.execute import TestInvocation
+
+
+BEAKERLIB_REPORT_RESULT_COMMAND = 'rhts-report-result'
 
 
 @provides_framework('beakerlib')
@@ -45,8 +48,12 @@ class Beakerlib(TestFramework):
         # implemented by the script itself. If the script gets called with this name, it
         # will *accept* the third positional argument as a `logfile`.
         # - https://github.com/teemtee/tmt/blob/e7cf41d1fe5a4dcbb3270758586f41313e9462ec/tmt/steps/execute/scripts/tmt-report-result#L101
-        bkrlib_report_result_cmd = tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.aliases[1]
-        assert bkrlib_report_result_cmd == "rhts-report-result"
+        if BEAKERLIB_REPORT_RESULT_COMMAND not in [
+                tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.source_filename,
+                *tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.aliases]:
+            raise GeneralError("Beakerlib framework requires the "
+                               f"'{BEAKERLIB_REPORT_RESULT_COMMAND}' script to be available "
+                               "on a guest.")
 
         return Environment({
             'BEAKERLIB_DIR': EnvVarValue(invocation.path),
@@ -56,7 +63,7 @@ class Beakerlib(TestFramework):
 
             # The command in this variable gets called with every `rlPhaseEnd` call in beakerlib.
             'BEAKERLIB_COMMAND_REPORT_RESULT': EnvVarValue(
-                invocation.guest.scripts_path / bkrlib_report_result_cmd),
+                invocation.guest.scripts_path / BEAKERLIB_REPORT_RESULT_COMMAND),
 
             # This variables must be set explicitly, otherwise the beakerlib `rlPhaseEnd` macro
             # will not call the the command in `BEAKERLIB_COMMAND_REPORT_RESULT`.
--- a/tmt/frameworks/beakerlib.py
+++ b/tmt/frameworks/beakerlib.py
@@ -33,6 +33,21 @@ class Beakerlib(TestFramework):
             invocation: 'TestInvocation',
             logger: tmt.log.Logger) -> tmt.utils.Environment:
 
+        # The beakerlib calls the command in this variable in the following way:
+        # $BEAKERLIB_COMMAND_REPORT_RESULT "$testname" "$result" "$logfile" "$score"
+        # - https://github.com/beakerlib/beakerlib/blob/5a85937f557b735f32996eb55cd6b9a33f3fe653/src/testing.sh#L1076
+        #
+        # If we use the `tmt-report-result` value here, the script will not be compatible
+        # with the third `$logfile` positional argument - it will just ignore it because it
+        # accepts the logfile provided only by `-o/--outputFile` option be default.
+        #
+        # The reason the `rhts-report-result` alias is used here is a compatibility layer
+        # implemented by the script itself. If the script gets called with this name, it
+        # will *accept* the third positional argument as a `logfile`.
+        # - https://github.com/teemtee/tmt/blob/e7cf41d1fe5a4dcbb3270758586f41313e9462ec/tmt/steps/execute/scripts/tmt-report-result#L101
+        bkrlib_report_result_cmd = tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.aliases[1]
+        assert bkrlib_report_result_cmd == "rhts-report-result"
+
         return Environment({
             'BEAKERLIB_DIR': EnvVarValue(invocation.path),
             'BEAKERLIB_COMMAND_SUBMIT_LOG': EnvVarValue(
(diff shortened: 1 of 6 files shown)


tmt/frameworks/beakerlib.py
```diff
@@ -33,6 +33,21 @@ def get_environment_variables(
...
+        # If we use the `tmt-report-result` value here, the script will not be compatible
+        # with the third `$logfile` positional argument - it will just ignore it because it
+        # accepts the logfile provided only by `-o/--outputFile` option be default.
+        #
+        # The reason the `rhts-report-result` alias is used here is a compatibility layer
+        # implemented by the script itself. If the script gets called with this name, it
+        # will *accept* the third positional argument as a `logfile`.
+        # - https://github.com/teemtee/tmt/blob/e7cf41d1fe5a4dcbb3270758586f41313e9462ec/tmt/steps/execute/scripts/tmt-report-result#L101
+        bkrlib_report_result_cmd = tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.aliases[1]
+        assert bkrlib_report_result_cmd == "rhts-report-result"
```
happz, line 49: Changing the order of aliases, for whatever reason, will cause this to explode, and nothing will stop us from doing so.

How about putting `"rhts-report-result"` into a module-level "constant", and just making sure it's one of the aliases?

```
BEAKERLIB_REPORT_RESULT_COMMAND = 'rhts-report-result'

...

if BEAKERLIB_REPORT_RESULT_COMMAND not in tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.aliases:
    raise GeneralError("Something like Beakerlib framework requires rhts-report-result script to be installed")
```

The dependency on order isn't stable enough, and it'd be nice if frameworks could reserve their own aliases, like this one, but that would be a more complicated patch.

seberm: Hello @happz , thanks for the hint.

I've changed the code in https://github.com/teemtee/tmt/pull/3372/commits/0f0f860e717f1cb6ec058a1095f0e11387f1c3fc and also added `tmt.steps.execute.TMT_REPORT_RESULT_SCRIPT.source_filename` into the condition. It seems to me it should be probably there. Do you agree?

happz: Agreed, unlikely but possible.

seberm: Ok, marking this as resolved. Thanks!

martinhoyer (approved):
Don't know much about rhts stuff, but lgtm.

skycastlelily (approved):
LGTM:)

psss (approved):
Looks good, thanks!

psss:
Seems that `/tests/provision/become` fails after the change. The problem is that guest pull is not able to fetch the newly created logs:

```
rsync: [sender] send_files failed to open "/var/tmp/tmt/run-001/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/Cleanup_tmp.3KTaRskP64": Permission denied (13)
rsync: [sender] send_files failed to open "/var/tmp/tmt/run-001/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/Setup_tmp.7HHUBr6cCD": Permission denied (13)
rsync: [sender] send_files failed to open "/var/tmp/tmt/run-001/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/Test_tmp.vkwk0eUY8N": Permission denied (13)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1852) [generator=3.3.0]
```

Easy steps to reproduce locally:

```
cd tests/provision/become/data
tmt --context provisiontest=virtual run -rvvv plan --name /umask
```

seberm:
> Seems that `/tests/provision/become` fails after the change. The problem is that guest pull is not able to fetch the newly created logs:
> 
> ```
> rsync: [sender] send_files failed to open "/var/tmp/tmt/run-001/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/Cleanup_tmp.3KTaRskP64": Permission denied (13)
> rsync: [sender] send_files failed to open "/var/tmp/tmt/run-001/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/Setup_tmp.7HHUBr6cCD": Permission denied (13)
> rsync: [sender] send_files failed to open "/var/tmp/tmt/run-001/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/Test_tmp.vkwk0eUY8N": Permission denied (13)
> rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1852) [generator=3.3.0]
> ```
> 
> Easy steps to reproduce locally:
> 
> ```
> cd tests/provision/become/data
> tmt --context provisiontest=virtual run -rvvv plan --name /umask
> ```

Hello @psss @happz ,

So, if I understand this correctly, the `tmt` together with `guest.pull()` (the rsync) is running under the `fedora` user and that's why the rsync cannot read phase logs (they are readable only by root user):

```
[fedora@default-0 data]$ pwd
/var/tmp/tmt/run-054/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data
[fedora@default-0 data]$ ls -alh
total 20K
drwxr-xr-x+ 1 fedora fedora  190 Dec  9 11:56 .
drwxr-xr-x+ 1 fedora fedora  376 Dec  9 11:29 ..
-rw-------. 1 root   root    482 Dec  9 11:29 Cleanup_tmp.0xUJHxkmKJ
-rw-r--r--. 1 root   root   2.7K Dec  9 11:30 journal.xml
-rw-------. 1 root   root    580 Dec  9 11:29 Setup_tmp.LRb5CQS4dv
-rw-------. 1 root   root    472 Dec  9 11:29 Test_tmp.lRXdeF2tWd
-rw-r--r--. 1 root   root    383 Dec  9 11:29 tmt-report-results.yaml
```

The strange thing is that files created by tmt are always owned by `fedora:fedora` but files created by `beakerlib` (e.g. `journal.xml`, `TestResults`, ...) are owned by `root:root`. Is this expected? All these beakerlib files are **readable** by `group` and `others`, **except the phase logs**.

It seems like the ACLs are not applied when the `tmt-report-result` script uses `cp -f "$outputFile" "${TMT_TEST_DATA}/${fileprefix}_${filename}"` to copy phase log to `TMT_TEST_DATA` directory. What do you think?

```
[root@default-0 ]# getfacl /var/tmp/tmt/run-054/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data
getfacl: Removing leading '/' from absolute path names
# file: var/tmp/tmt/run-054/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data
# owner: fedora
# group: fedora
user::rwx
group::r-x
other::r-x
default:user::rwx
default:group::r-x
default:other::r-x
```

I tried to reproduce it like the following:

```
[root@default-0 tmp]# pwd
/tmp
[root@default-0 tmp]# touch mytest
[root@default-0 tmp]# chmod 600 mytest
[root@default-0 tmp]# ls -alh mytest
-rw-------. 1 root root 0 Dec  9 12:20 mytest
[root@default-0 tmp]# cp mytest /var/tmp/tmt/run-054/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/
[root@default-0 tmp]# ls -alh /var/tmp/tmt/run-054/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/mytest
-rw-------. 1 root root 0 Dec  9 12:20 /var/tmp/tmt/run-054/umask/execute/data/guest/default-0/Beakerlib-test-to-generate-report-files-on-guest-1/data/mytest
```

psss:
I believe that f157ae7 should fix the problem.

psss:
Trying with a less intrusive approach proposed by @happz: d4d1694f.

psss:
Important tests passed in at least one of the recent `provision` jobs, merging.

seberm:
@happz @psss Thanks for helping me finish this up!

PR #3378, opened 2024-11-25
Readme badges and pyproject cleanup

martinhoyer:
I'm not a friend with rst and have more experience with markdown. Will this render ok on PyPI page and other places? We also need to deal with man generation I presume?
PyPA should be ok based on: 
```
$ twine check dist/tmt-1.40.0.dev2+ga31e8651-py3-none-any.whl --strict
Checking dist/tmt-1.40.0.dev2+ga31e8651-py3-none-any.whl: PASSED
```

github would look like this:
image

In pyproject, I've replaced Python 3.11 with 3.13 and removed the unmaintainable comments of what versions are available where.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
           - "pygments>=2.7.4"
           - "requests>=2.25.1"
           - "ruamel.yaml>=0.16.6"
-          - "urllib3>=1.26.5, <2.0"
+          - "urllib3>=1.26.5, <3.0"
           - "jira>=3.5.0, <4"
 
           # report-junit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,23 +27,23 @@ repos:
         #
         # For now, we simply copy & paste from pyproject.toml :(
         additional_dependencies:
-          - "click>=8.0.3,!=8.1.4"   # 8.1.3 / 8.1.6 TODO type annotations tmt.cli.Context -> click.core.Context click/issues/2558
-          - "docutils>=0.16"         # 0.16 is the current one available for RHEL9
-          - "fmf>=1.3.0"
-          - "jinja2>=2.11.3"         # 3.1.2 / 3.1.2
-          - "packaging>=20"          # 20 seems to be available with RHEL8
-          - "pint>=0.16.1"           # 0.16.1
+          - "click>=8.0.3"
+          - "docutils>=0.16"
+          - "fmf>=1.7.0"
+          - "jinja2>=2.11.3"
+          - "packaging>=20"
+          - "pint>=0.16.1"
           - "pydantic>=1.10.14"
-          - "pygments>=2.7.4"        # 2.7.4 is the current one available for RHEL9
-          - "requests>=2.25.1"       # 2.28.2 / 2.31.0
-          - "ruamel.yaml>=0.16.6"    # 0.17.32 / 0.17.32
-          - "urllib3>=1.26.5, <2.0"  # 1.26.16 / 2.0.4
+          - "pygments>=2.7.4"
+          - "requests>=2.25.1"
+          - "ruamel.yaml>=0.16.6"
+          - "urllib3>=1.26.5, <2.0"
           - "jira>=3.5.0, <4"
 
           # report-junit
           - "lxml>=4.6.5"
 
-          - "typing-extensions>=4.9.0; python_version < '3.13'"
+          - "typing-extensions>=4.12.2; python_version < '3.13'"
           - "pytest"
           - "requre"
           # Do not install *types-click* - it's not recommended with Click 8 & newer,
@@ -75,17 +75,17 @@ repos:
         #
         # For now, we simply copy & paste from pyproject.toml :(
         additional_dependencies:
-          - "click>=8.0.3,!=8.1.4"   # 8.1.3 / 8.1.6 TODO type annotations tmt.cli.Context -> click.core.Context click/issues/2558
-          - "docutils>=0.16"         # 0.16 is the current one available for RHEL9
-          - "fmf>=1.3.0"
-          - "jinja2>=2.11.3"         # 3.1.2 / 3.1.2
-          - "packaging>=20"          # 20 seems to be available with RHEL8
-          - "pint>=0.16.1"           # 0.16.1 / 0.19.x  TODO: Pint 0.20 requires larger changes to tmt.hardware
+          - "click>=8.0.3"
+          - "docutils>=0.16"
+          - "fmf>=1.7.0"
+          - "jinja2>=2.11.3"
+          - "packaging>=20"
+          - "pint>=0.16.1"
           - "pydantic>=1.10.14"
-          - "pygments>=2.7.4"        # 2.7.4 is the current one available for RHEL9
-          - "requests>=2.25.1"       # 2.28.2 / 2.31.0
-          - "ruamel.yaml>=0.16.6"    # 0.17.32 / 0.17.32
-          - "urllib3>=1.26.5, <2.0"  # 1.26.16 / 2.0.4
+          - "pygments>=2.7.4"
+          - "requests>=2.25.1"
+          - "ruamel.yaml>=0.16.6"
+          - "urllib3>=1.26.5, <2.0"
           - "jira>=3.5.0, <4"
 
           # report-junit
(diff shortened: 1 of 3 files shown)


martinhoyer:
image

Needs consensus on what badges are actually useful (which once to pick) and which of those make sense for them to be dynamic (versions, build successful).

I would say that Copr can be renamed to "Development/nighly builds" and just point to teemtee/latest, for example.

happz:
> image
> 
> Needs consensus on what badges are actually useful (which once to pick) and which of those make sense for them to be dynamic (versions, build successful).
> 
> I would say that Copr can be renamed to "Development/nighly builds" and just point to teemtee/latest, for example.

I'd probably drop "ruff" and "pre-commit", the rest will lead a user to some artifacts or docs, while these two are more like a statement, "yup, we test it", and we don't expect these to ever be red for `main` branch, correct?

psss:
> I'd probably drop "ruff" and "pre-commit", the rest will lead a user to some artifacts or docs, while these two are more like a statement, "yup, we test it", and we don't expect these to ever be red for `main` branch, correct?

Agreed, I feel it the same way.

psss (changes requested):
Let's use the fresh copr link.

thrix (changes requested):

pyproject.toml
```diff
@@ -24,23 +24,24 @@ classifiers = [
...
-    "packaging>=20",          # 20 seems to be available with RHEL8
-    "pint>=0.16.1",           # 0.16.1
+    "jinja2>=2.11.3",
+    "packaging>=20",
+    "pint>=0.16.1",
     "pydantic>=1.10.14",
-    "pygments>=2.7.4",        # 2.7.4 is the current one available for RHEL9
-    "requests>=2.25.1",       # 2.28.2 / 2.31.0
-    "ruamel.yaml>=0.16.6",    # 0.17.32 / 0.17.32
-    "urllib3>=1.26.5, <3.0",  # 1.26.16 / 2.0.4
```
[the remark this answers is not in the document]
martinhoyer: @thrix Well, it seems unmaintainable, definitely was a mistake to add Fedora versions there. Mea culpa.  

I presume we can add "explanation" for the lock, like "version xyz available on epel9 [04/2015]" or something", but not sure if there is any value as that is the case for all of them: "Min versions are dictated by minimal versions of RHEL/EPEL9 because we don't have something like https://github.com/teemtee/tmt/pull/3605, this is also why we have Python 3.9, even though it will be EOL in 6 months"

martinhoyer:
> I'd probably drop "ruff" and "pre-commit", the rest will lead a user to some artifacts or docs, while these two are more like a statement, "yup, we test it", and we don't expect these to ever be red for `main` branch, correct?

> Agreed, I feel it the same way.

@psss @happz I don't see what's wrong with that. It just tells people that this project uses pre-commit and ruff, which can increase the respect for the project, making them more likely to get involved. I mean, it's literally called a "badge".  

Alas, removed as requested.

README.rst
```diff
@@ -3,6 +3,8 @@
                     tmt - Test Management Tool
 ==================================================================
 
+|docs| |matrix| |fedora-pkg| |copr-build| |pypi-version| |python-versions| |ruff| |pre-commit| |license|
```
mcasquer, line 6:
```suggestion
|docs| |matrix| |fedora-pkg| |copr-build| |pypi-version| |python-versions| |license|
```
@martinhoyer these should be removed from here as well

martinhoyer: haha, nice one martin.

mcasquer:
@martinhoyer LGTM, just a thought: do you think we can point the `python-versions` badge to a different place than pypi? IMHO some place in the docs would be better although I haven't found anything apart from a [brief mention in the Unit Tests](https://tmt.readthedocs.io/en/stable/contribute.html#unit-tests) section

happz:
Seems to be rendering well in both HTML docs and man page: neither of these two shows any sign of badges :/ Which I guess is fine, yet a bit surprising.

pyproject.toml
```diff
@@ -24,44 +24,46 @@ classifiers = [
     "License :: OSI Approved :: MIT License",
     "Natural Language :: English",
     "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
happz, line 27: I understand the explanation of switching to 3.13 in `tool.hatch.envs.test.matrix` below, but why dropping 3.11 "badge of compatibility" completely? tmt should be still quite usable with Python 3.11, and Python 3.11 EOL is in 2027 or something. Adding 3.12 and 3.13, sure, removing 3.11 does not seem right.

martinhoyer: Sure, but then we probably want to add all 9-13?

README.rst
```diff
@@ -21,3 +23,31 @@ inheritance and elasticity metadata are organized in the structure
...
+.. |matrix| image:: https://img.shields.io/badge/Matrix-000?logo=matrix&logoColor=fff
+    :target: https://matrix.to/#/#tmt:fedoraproject.org
+    :alt: Matrix chat
+
+.. |fedora-pkg| image:: https://img.shields.io/fedora/v/tmt?logo=fedora&logoColor=fff&color=fff&labelColor=51a2da
+    :target: https://src.fedoraproject.org/rpms/tmt
+    :alt: Fedora package
+
+.. |copr-build| image:: https://img.shields.io/badge/dynamic/json?color=blue&label=Copr&query=builds.latest.state&url=https%3A%2F%2Fcopr.fedorainfracloud.org%2Fapi_3%2Fpackage%3Fownername%3D%40teemtee%26projectname%3Dlatest%26packagename%3Dtmt%26with_latest_build%3DTrue
+    :target: https://copr.fedorainfracloud.org/coprs/g/teemtee/stable/
```
martinhoyer, line 40: Do we really want `/stable/`? I would say `/latest/` make more sense...

pyproject.toml
martinhoyer: added

happz: I agree, 3.10 should be there too. No idea why it's missing.

README.rst
thrix: @martinhoyer can you update it pls?

mcasquer (approved):
LGTM

.pre-commit-config.yaml
```diff
@@ -27,23 +27,23 @@ repos:
...
+          - "pint>=0.16.1"
           - "pydantic>=1.10.14"
-          - "pygments>=2.7.4"        # 2.7.4 is the current one available for RHEL9
-          - "requests>=2.25.1"       # 2.28.2 / 2.31.0
-          - "ruamel.yaml>=0.16.6"    # 0.17.32 / 0.17.32
-          - "urllib3>=1.26.5, <2.0"  # 1.26.16 / 2.0.4
+          - "pygments>=2.7.4"
+          - "requests>=2.25.1"
+          - "ruamel.yaml>=0.16.6"
+          - "urllib3>=1.26.5, <2.0"
```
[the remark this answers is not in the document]
happz: Indeed. @martinhoyer please, check this out, and the rest of requirements if possible too, just it case there another typo like this one.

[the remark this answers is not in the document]
martinhoyer: no this is correct. there is urllib 1x on EL10 and urllib 2x on Fedora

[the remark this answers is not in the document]
martinhoyer: oh! I'm blind. Thanks. Too bad graphite added extra `+` symbol btw

happz (approved):

.pre-commit-config.yaml
```diff
@@ -75,17 +75,17 @@ repos:
...
```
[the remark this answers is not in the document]
happz: Yeah, correct... 

@martinhoyer ^^

pyproject.toml
```diff
@@ -24,41 +24,45 @@ classifiers = [
...
+    "nitrate>=1.9.0",
     "python-bugzilla>=3.2.0",
     ]
 export-polarion = [
-    "pylero>=0.0.8",
+    "pylero>=0.1.0",
     ]
 provision-beaker = [
-    "mrack>=1.23.1",
+    "mrack>=1.23.3",
```
happz, line 65:
```suggestion
    "mrack>=1.23.2",
```

This is not going to work, according to PyPI, the most recent version is     `1.23.2`.

PR #3379, opened 2024-11-26
Document overview of supported `restraint` scripts

psss:
Seems we don't have an overview of the `restraint` scripts which are supported by `tmt`. Let's add a new section to `Questions`.

Pull Request Checklist

* [x] write the documentation

### DIFF
--- a/docs/questions.rst
+++ b/docs/questions.rst
@@ -158,6 +158,25 @@ Metadata Specification and are supposed to be synced temporarily
 to keep backward compatibility.
 
 
+Restraint Compatibility
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For backward-compatibility ``tmt`` provides selected commands
+of the `restraint`__ framework so that existing tests can be more
+easily migrated. Currently the following scripts are supported:
+
+* ``rhts-abort`` and ``rhts-abort`` — :ref:`/stories/features/abort`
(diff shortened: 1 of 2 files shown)


martinhoyer (approved):
Perhaps the storries/deferred restraint section could be updated as part of this PR?  

Other thing is that for some reason I don't see this section when searching the docs for "Restraint". 

One last thought would be that it's another thing to maintain, but I guess it's worth it?

LGTM otherwise :)

seberm (approved):

psss:
> Perhaps the storries/deferred restraint section could be updated as part of this PR?

Good point, should be addressed in c6e0077a.

docs/questions.rst
```diff
@@ -158,6 +158,27 @@ Metadata Specification and are supposed to be synced temporarily
...
+.. _restraint-compatibility:
+
+Restraint Compatibility
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For backward-compatibility ``tmt`` provides selected commands
+of the `restraint`__ framework so that existing tests can be more
+easily migrated. Currently the following scripts are supported:
+
+* ``rhts-abort`` and ``rhts-abort`` — :ref:`/stories/features/abort`
```
therazix, line 170: Maybe I misunderstood, but why is each command listed twice? Shouldn't one of them have the `rstrnt` prefix instead of `rhts`?

psss: Ah! Good catch! Apparently a bunch of us have a blind spot there ;-) Should be fixed in bf58e1f1.

therazix (approved):

PR #3380, opened 2024-11-26
Uninstall bootc in Fedora CoreOS testing image

happz:
bootc depends on podman, podman depends on diffutils, diffutils are being removed for the purpose of tmt tests. And rpm-ostree does not remove dependants automatically.

Pull Request Checklist

* [x] implement the feature


psss (approved):

psss:
Hm, seems the problem is still there...

happz:
> Hm, seems the problem is still there...

Hm, yeah, because I'm an idiot... 0c077a7d to the rescue.

Issue #3371, opened 2024-11-20
Broken 'reconnect'  usage between tmt run invocations  


lukaszachy:
It should possible to keep the rundir and use already provisioned machine in subsequent `tmt run` commands.  But doing that hits the traceback for 'ssh' guests.

Reproducer:
```
$ cat plan.fmf
execute:
    how: tmt
    script: echo

$ tmt run --id BUG provision -h virtual
...
summary: 1 guest provisioned

$ tmt run --id BUG discover prepare
plan failed

The exception was caused by 1 earlier exceptions

Cause number 1:

    prepare step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:
```

Full traceback (TMT_SHOW_TRACEBACK=full) shows that assert fails:
```
            File /usr/lib/python3.13/site-packages/tmt/steps/provision/__init__.py, line 1469, in _ssh_master_socket_path
              assert isinstance(self.parent, tmt.steps.provision.Provision)

              self = <tmt.steps.provision.testcloud.GuestTestcloud object at 0x7f7a26529590>
```

https://github.com/teemtee/tmt/blob/cc75eb6bf245b3bcba1da68732f252b95a3f0970/tmt/steps/provision/__init__.py#L1469

As in the `tmt run --id BUG discover prepare` the ssh connection is established during the prepare and thus the `self.parent` is `tmt.steps.prepare.Prepare` making the assert fail.

psss:
It would be good to get this fixed soon, proposing for 1.40.

PR #3381, opened 2024-11-26
Fix reuse of already provisioned machine

lukaszachy:
When provision is already done and one uses same rundir to reuse this machine, the ssh connection might be initialized in other step than provision. Previously this failed assertion, making this usage impossible.

Fix: #3371

Pull Request Checklist

* [x] implement the feature


thrix (approved):
LGTM

PR #3384, opened 2024-11-26
Add fallback method for generating shorter SSH master socket

happz:
It turns out that even when using a hashing function, the final filepath may still be too long.

Changing the process slightly: still using a hashing function as a fallback - different one - and trying to use short substrings of the hash. Just like git forges do, showing and accepting short commit IDs. This also offers a nice way to prevent possible hash collisions that, while extremely unlikely, cannot be ruled out and would be terribly hard to debug.

Pull Request Checklist

* [x] implement the feature


happz:
@KwisatzHaderach FYI

I need to add a test or two, as soon as I find out how to force tmt into creating some ugly socket paths (probably via guest names).

KwisatzHaderach (approved):

psss (approved):
Looks good!

psss:
Failures seem to be irrelevant:

    Timeout, server ... not responding.

Merging.

PR #3386, opened 2024-11-28
Support `--workdir-root` in the `tmt run` command

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note


KwisatzHaderach (approved):

psss:
Many tests are failing, it seems that the workdir is not created if it does not exist before. Reproducible easily by removing the `/var/tmp/tmt` directory or setting a different workdir:

    TMT_WORKDIR_ROOT=/tmp/new tmt run discover -v

skycastlelily:
> Many tests are failing, it seems that the workdir is not created if it does not exist before. Reproducible easily by removing the `/var/tmp/tmt` directory or setting a different workdir:
> 
> ```
> TMT_WORKDIR_ROOT=/tmp/new tmt run discover -v
> ```

yes, I was too confident about the mr,  I failed to realize one difference compared to workdir-root-clean mr : workdir-root may not exist for run case:sweat_smile: 

Updated,should be ready now:)

psss (approved):
Thanks for improving this, looks good!

thrix (approved):
Works for me, LGTM

martinhoyer (approved):

PR #3388, opened 2024-12-02
Document runner vs guest compatibility

happz:
Adds a YAML file to collect the info, and a script & template to render into a nice ReST file as part of docs.

A preview: https://tmt--3388.org.readthedocs.build/en/3388/guide.html#guest-compatibility

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation


happz:
@psss as discussed, my shot at "runner vs guest" compatibility. There are some loose ends though:

* where should I put the page? I added to the very top of the top-level menu, which of course is not the final place. Can you propose the proper location?
* the name might need a change. We were discussing issues between a runner and a guest, hence the "runner-guest compatibility", and I tried to stick with it. If you have a better idea, I'm all ears.
* I couldn't convince Sphinx to re-use notes. I imagined multiple cells would refer to the same note, it should be possible, but Sphinx always renumbered them. So, for now, a note may be repeated multiple times... Still looking for a better solution.
* updating tests to avoid the problematic combination, as agreed it would be part of this PR too.

docs/runner-guest-compatibility.yaml
```diff
@@ -0,0 +1,49 @@
...
+    title: Fedora Rawhide and Ansible 2.17+
+    text: >
+      :ref:`prepare/ansible</plugins/prepare/ansible>`,
+      :ref:`prepare/feature</plugins/prepare/feature>`
+      and :ref:`finish/ansible</plugins/finish/ansible>` will not work
+      out of the box. See their documentation for details and possible
+      workarounds.
+
+  - pattern: '.*? \+ (Alpine|Debian|Ubuntu)'
+    title: Beakerlib not being packaged for non-RHEL OS
```
martinhoyer, line 45: > non-RHEL OS  

non Fedora-based / rpm-based / ...?

Perhaps it would be more accurate to mention that beakerlib framework requires beakerlib being installed or being available in enabled repos?

psss: Seems to be addressed by 4080e1d805085338d01adbc4dcb93c7c38aa189a.

docs/runner-guest-compatibility.yaml
```diff
@@ -0,0 +1,52 @@
...
+  # but we do not know for sure.
+  - 'Alpine .*'
+  - 'Ubuntu .*'
+  - 'Debian .*'
+
+notes:
+  - pattern: 'Fedora Rawhide \+ (CentOS 7|Centos Stream 8)'
+    title: Fedora Rawhide and Ansible 2.17+
+    text: >
+      :ref:`prepare/ansible</plugins/prepare/ansible>`,
```
psss, line 38:
```suggestion
      The :ref:`prepare/ansible</plugins/prepare/ansible>`,
```

docs/runner-guest-compatibility.yaml
```diff
@@ -0,0 +1,52 @@
...
+      :ref:`prepare/feature</plugins/prepare/feature>`
+      and :ref:`finish/ansible</plugins/finish/ansible>` will not work
```
psss, line 40:
```suggestion
      and :ref:`finish/ansible</plugins/finish/ansible>` plugins will not work
```

docs/runner-guest-compatibility.yaml
```diff
@@ -0,0 +1,52 @@
...
+      out of the box. See their documentation for details and possible
+      workarounds.
+
+  - pattern: '.*? \+ (Alpine|Debian|Ubuntu)'
+    title: Beakerlib not packaged for Linux distributions outside of Fedora family
+    text: >
+      :ref:`beakerlib</spec/tests/framework>` will not work because it
```
psss, line 47:
```suggestion
      Tests using the :ref:`beakerlib</spec/tests/framework>` framework will not work because it
```

docs/templates/runner-guest-compatibility-matrix.rst.j2
```diff
@@ -0,0 +1,48 @@
+{#
+    A Jinja2 template for rendering HW requirement support matrix in tmt's own docs.
```
psss, line 2: A copy & paste leftover?

psss (changes requested):
Thanks for putting this together! Looks good. Added a few comments.

psss:
> * where should I put the page? I added to the very top of the top-level menu, which of course is not the final place. Can you propose the proper location?

What about adding this info under the [Test Runner](https://tmt.readthedocs.io/en/stable/guide.html#test-runner) section?

> * the name might need a change. We were discussing issues between a runner and a guest, hence the "runner-guest compatibility", and I tried to stick with it. If you have a better idea, I'm all ears.

Sounds good to me, or perhaps `Runner vs Guest Compatibility`?

> * I couldn't convince Sphinx to re-use notes. I imagined multiple cells would refer to the same note, it should be possible, but Sphinx always renumbered them. So, for now, a note may be repeated multiple times... Still looking for a better solution.

Sure, can be improved later.

> * updating tests to avoid the problematic combination, as agreed it would be part of this PR too.

Ack, this seems to be already covered, right?

psss:
Hm, the `avc` seems to be unrelated:

```
type=AVC msg=audit(12/05/2024 15:59:25.949:2611) : avc:  denied  { getattr } for  pid=24982 comm=systemd-hostnam path=cgroup:[4026531835]
dev="nsfs" ino=4026531835 scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:nsfs_t:s0 tclass=file permissive=0
----
type=AVC msg=audit(12/05/2024 15:59:25.950:2612) : avc:  denied  { getattr } for  pid=24982 comm=systemd-hostnam path=pid:[4026531836]
dev="nsfs" ino=4026531836 scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:nsfs_t:s0 tclass=file permissive=0
```

Perhaps some random hiccup? Or does it appear reliably/repeatedly?

happz:
> Hm, the `avc` seems to be unrelated:
> 
> ```
> type=AVC msg=audit(12/05/2024 15:59:25.949:2611) : avc:  denied  { getattr } for  pid=24982 comm=systemd-hostnam path=cgroup:[4026531835]
> dev="nsfs" ino=4026531835 scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:nsfs_t:s0 tclass=file permissive=0
> ----
> type=AVC msg=audit(12/05/2024 15:59:25.950:2612) : avc:  denied  { getattr } for  pid=24982 comm=systemd-hostnam path=pid:[4026531836]
> dev="nsfs" ino=4026531836 scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:nsfs_t:s0 tclass=file permissive=0
> ```
> 
> Perhaps some random hiccup? Or does it appear reliably/repeatedly?

Reliably. These are the pesky AVCs @martinhoyer most likely spoke about when discussing Rawhide and `xfail`.

psss:
> Sounds good to me, or perhaps `Runner vs Guest Compatibility`?

Or, maybe, if it's hosted under the `Test Runner` section it could be called `Guest Compatibility` as the test runner part is kind of implied in that whole section. Just brainstorming.

milosmalik:
The SELinux denials shown above are new in Fedora rawhide and they are not yet fixed. However, they are already reported:
 * https://bugzilla.redhat.com/show_bug.cgi?id=2330722
 * https://bugzilla.redhat.com/show_bug.cgi?id=2330476
 * https://bugzilla.redhat.com/show_bug.cgi?id=2329752

psss:
Thanks for the pointers, @milosmalik. So, I guess, we can `xfail` those `avc` checks for now. Or, if we don't want to catch similar fresh avc issues on `rawhide`, we could perhaps make all avc checks `info` for `rawhide`?

docs/templates/runner-guest-compatibility-matrix.rst.j2
happz: Addressed in 5a7833a7

tests/prepare/feature/test.sh
```diff
@@ -16,6 +16,13 @@ rlJournalStart
 
     # EPEL
     for image in $images; do
+        if rlIsFedora ">=42" && (is_centos_7 "$image" || is_ubi_8 "$image"); then
+                rlLogInfo "Skipping because Ansible shipped with Fedora does not support Python 3.6"
+            rlPhaseEnd
```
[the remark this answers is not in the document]
happz: Addressed in 5a7833a7

happz:
> > Sounds good to me, or perhaps `Runner vs Guest Compatibility`?
> 
> Or, maybe, if it's hosted under the `Test Runner` section it could be called `Guest Compatibility` as the test runner part is kind of implied in that whole section. Just brainstorming.

Ok, moved, renamed, done.

psss (approved):

Issue #3387, opened 2024-11-28
bootc provision needs `--rootfs` when using fedora-bootc image


HuijingHei:
Provision failed when using `quay.io/fedora/fedora-bootc:41` image, seems it needs `--rootfs`
```
Cause number 1:

    provision step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        Command 'podman run --rm --privileged -v /var/lib/containers/storage:/var/lib/containers/storage --security-opt label=type:unconfined_t -v /var/tmp/tmt/run-002/plans/basic/containerfile/butane/provision/default-0:/output quay.io/centos-bootc/bootc-image-builder:latest build --type qcow2 --local localhost/tmtmodified-1732783970' returned 1.

        # stdout (1/1 lines)
        # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        Generating manifest manifest-qcow2.json
        # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

        # stderr (8/8 lines)
        # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        Trying to pull quay.io/centos-bootc/bootc-image-builder:latest...
        2024/11/28 08:56:54 error: cannot build manifest: no default root filesystem type specified in container, please use "--rootfs" to set manually
        # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Run manually and add for example `--rootfs xfs`, it works.
```
$ sudo podman run --rm --privileged -v /var/lib/containers/storage:/var/lib/containers/storage -v /home/fedora/fedora-bootc:/output --security-opt label=type:unconfined_t quay.io/centos-bootc/bootc-image-builder:latest build --type qcow2 --rootfs xfs --local quay.io/fedora/fedora-bootc:41
```

HuijingHei:
> Can you copy/paste here the tmt plan / config that you used? Thanks

[...]

PR #3390, opened 2024-12-03
Add the `--rootfs` option for `bootc` image building

henrywang:
bootc image builder needs `--rootfs` for image which does not have fs type included. Use `xfs` as rootfs to work with RHEL, CentOS and Fedora.

Resolve: https://github.com/teemtee/tmt/issues/3387

### DIFF
--- a/tmt/steps/provision/bootc.py
+++ b/tmt/steps/provision/bootc.py
@@ -138,10 +138,10 @@ class BootcData(tmt.steps.provision.testcloud.ProvisionTestcloudData):
     rootfs: str = field(
         default="xfs",
         option=('--rootfs'),
-        metavar='ROOTFS',
+        choices=['ext4', 'xfs', 'btrfs'],
         help="""
-             Select root filesystem type. Overrides the default from the source 
-             container. Supported value: ext4, xfs, and btrfs
+             Select root filesystem type. Overrides the default from the source
+             container.
              """)
 
 
@@ -340,7 +340,7 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
             containerimage = data.container_image
             if data.add_tmt_dependencies:
                 containerimage = self._build_derived_image(data.container_image)
-            self._build_bootc_disk(containerimage, data.image_builder)
+            self._build_bootc_disk(containerimage, data.image_builder, data.rootfs)
 
         # Build image according to the container file
         elif data.container_file is not None:
--- a/tmt/steps/provision/bootc.py
+++ b/tmt/steps/provision/bootc.py
@@ -135,6 +135,15 @@ class BootcData(tmt.steps.provision.testcloud.ProvisionTestcloudData):
              building the bootc disk image.
              """)
 
+    rootfs: str = field(
+        default="xfs",
+        option=('--rootfs'),
+        metavar='ROOTFS',
+        help="""
+             Select root filesystem type. Overrides the default from the source 
+             container. Supported value: ext4, xfs, and btrfs
+             """)
+
 
 @tmt.steps.provides_method('bootc')
 class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
@@ -148,6 +157,7 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
         provision:
             how: bootc
             container-image: quay.io/centos-bootc/centos-bootc:stream9
+            rootfs: xfs
 
     Here's a config example using a Containerfile:
 
@@ -158,6 +168,7 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
             container-file: "./my-custom-image.containerfile"
             container-file-workdir: .
             image-builder: quay.io/centos-bootc/bootc-image-builder:stream9
+            rootfs: ext4
             disk: 100
 
     Another config example using an image that already includes tmt
@@ -169,6 +180,7 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
             how: bootc
             add-tmt-dependencies: false
             container-image: localhost/my-image-with-deps
+            rootfs: btrfs
 
     This plugin is an extension of the virtual.testcloud plugin.
     Essentially, it takes a container image as input, builds a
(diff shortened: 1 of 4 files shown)


tmt/steps/provision/bootc.py
```diff
@@ -271,6 +271,8 @@ def _build_bootc_disk(self, containerimage: str, image_builder: str) -> None:
             "build",
             "--type",
             "qcow2",
+            "--rootfs",
+            "xfs",
```
HuijingHei, line 275: Thanks for working on this! I think it would fix the problem, but not sure if `centos-bootc` needs this.

happz: Would it make sense to make the addition and/or the FS type configurable?

travier: yes, it would be good to make the filesystem type a configuration option

travier: we can probably also duplicate the test to make sure that we test both fedora & centos bootc images here.

tmt/steps/provision/bootc.py
```diff
@@ -135,6 +135,15 @@ class BootcData(tmt.steps.provision.testcloud.ProvisionTestcloudData):
...
              """)
 
+    rootfs: str = field(
+        default="xfs",
+        option=('--rootfs'),
+        metavar='ROOTFS',
+        help="""
+             Select root filesystem type. Overrides the default from the source 
+             container. Supported value: ext4, xfs, and btrfs
+             """)
```
happz, line 145: Two usability hints:

* if there is a set of choices, please use `choices=[...]` to list them, see e.g. https://github.com/teemtee/tmt/blob/main/tmt/cli/_root.py#L221
* do not mention the choices in the help text, adding `choices=` will make them appear as metavar, i.e. `--rootfs=xfs|ext4|btrfs`, which means you should also drop `metavar`.

henrywang:
@happz Could you please help enable `testing-farm:fedora-41-x86_64:provision` test? Looks I can't trigger tests. Thanks.

tmt/steps/provision/bootc.py
```diff
@@ -271,6 +283,8 @@ def _build_bootc_disk(self, containerimage: str, image_builder: str) -> None:
             "build",
             "--type",
             "qcow2",
+            "--rootfs",
+            rootfs,
```
HuijingHei, line 287: It would be better if add condition for `fedora-bootc` as `centos-bootc` does not need it.

henrywang: Hi @HuijingHei ,

The `centos-bootc` has default rootfs configured in container image https://gitlab.com/redhat/centos-stream/containers/bootc/-/blob/c9s/centos-stream-common.yaml?ref_type=heads#L17. But `fedora-bootc` does not.

The `--rootfs` in bib will overwrite rootfs configured in container image. https://github.com/osbuild/bootc-image-builder/blob/d42ad7b6d06cb9cffff36a7ccd6754c337303dfd/bib/cmd/bootc-image-builder/main.go#L256. So do not need additional code for that and it's not the checking point of this plugin.

HuijingHei: Makes sense to me, thanks!

tmt/steps/provision/bootc.py
skycastlelily: I guess,  to address @travier's  request you need:
Extend https://github.com/teemtee/tmt/blob/main/tests/provision/bootc/test.sh and add two more files,
say, includes-deps-fedora.containerfile  needs-deps-fedora.containerfile.

Otherwise,looks good to me.

henrywang: Hi @skycastlelily, in this case `includes-deps.containerfile` will cover `fedora-bootc:41` and `needs-deps.containerfile` covers `centos-bootc:stream9`. We have all of them covered. So do not need extra test added.

skycastlelily: TBO, I got your thoughts before I commented, and I'm not quite familiar with bootc stuff, but I believe the original contributor @ckyrouac add both includes-deps and needs-deps tests  for some reason, and your commit would will not test needs-deps for fedora and includes-deps for centos.

henrywang: Right. The `includes-deps` is for Containerfile which has `dnf install rsync cloud-init` (tmt needs) installed. The `needs-deps` is for Containerfile which does not have `dnf install rsync cloud-init` installed and to test `add-tmt-dependencies` property.

So one for rhel-bootc and another for fedora-bootc don't have any impact for the test, but add one more image covered.

skycastlelily:
>   to test add-tmt-dependencies property

And, whether add-tmt-dependencies related tmt code works correctly has
nothing to do with the image being used.
Makes sense to me now, thanks for your time.

henrywang:
@skycastlelily I make a rebase to sync with latest main. Can we land this PR? Thanks.

skycastlelily:
Sorry, I don't have the authority to merge, you should ask @psss  or @happz ,I was just trying to help them to do some review work,as the list waiting to be reviewed is way too long.

henrywang:
Hi @psss @happz, could you please help triggering test and merge this PR? Thanks.

KwisatzHaderach (approved):

thrix (approved):
LGTM

martinhoyer (approved):

tmt/steps/provision/bootc.py
psss, line 287: So we always set the `rootfs` value. This means that we always override by our default. What if the user wants to keep the default `rootfs` value from the source container? Shall we add a special value `default` for that use case?

psss (changes requested):
I think we should also support keeping / respecting the default `rootfs` value from the source container. See my comment.

tmt/steps/provision/bootc.py
happz: Probably some kind of `”original”` or `”keep”` default?

psss: Yes, something like this. I thought `default` would cover well the `default from the source container`. Maybe it could be confusing. Your `original` sounds good as well. Or we could use `source` if `source container` is the term commonly used.

henrywang:
> I think we should also support keeping / respecting the default `rootfs` value from the source container. See my comment.

I have my thought on this.

I think that's not TMT bootc plugin scope. User build their own container image, so they can take care of it.

The bib and boot image on rootfs:

- The `centos-bootc` has default rootfs configured in container image https://gitlab.com/redhat/centos-stream/containers/bootc/-/blob/c9s/centos-stream-common.yaml?ref_type=heads#L17. But `fedora-bootc` does not.

- The `--rootfs` in bib will overwrite rootfs configured in container image. https://github.com/osbuild/bootc-image-builder/blob/d42ad7b6d06cb9cffff36a7ccd6754c337303dfd/bib/cmd/bootc-image-builder/main.go#L256. So do not need additional code for that and it's not the checking point of this plugin.

psss:
Summary from the chat discussion: The fedora-bootc image is not going to add the default `rootfs` so it seems the best way for now is to always provide the default by us so that the image building works for the user regardless if the source container defines `rootf` or not.

psss:
The `provision` job failure is irrelevant, merging.

PR #3393, opened 2024-12-03
Store multiple invocations of dmesg check for the same test in one file

### LINKED ISSUE #3317: check dmesg overwrites itself with tests that reboot
tmt-reboots dmesg-after will run when the reboot is called. Not an issue in general but the problem occurs when the test completes or has multiple reboots, the name of the file used to save it is the same.
/var/tmp/tmt/run-022/kernel-automotive/plans/sst_kernel_ft/memory/plan/execute/data/guest/default-0/external_er/memory/config_memtest/config_memtest-1/checks/dmesg-after-test.txt
So it will be overwritten each time and only the very last one will remain. I think the same is happening with the dmesg-before-test.txt as well. Can we index this during a single test run so that we have the contents of all the captures?

Can we create the log file capture from dmesg with the epoch date appended in the name, so something like:
`dmesg-after-test-$(date +%s).txt`?
Or equivalent for python syntax.

happz:
This should prevent loosing dmesg reports in the case of tests that restart or reboot: all reports except the last one are lost.

The filename remains the same, but it's no longer overwritten. `dmesg` check invocations append to it, with proper headers:

```
## Acquired at 2024-12-03T13:12:23.536502+00:00

# stdout (812 lines)
[    0.000000] Linux version 6.11.4-301.fc41.x86_64 (mockbuild@9b6b61418589428cb880a7020233b56f) (gcc (GCC) 14.2.1 20240912 (Red Hat 14.2.1-3), GNU ld version 2.43.1-2.fc41) #1 SMP PREEMPT_DYNAMIC Sun Oct 20 15:02:33 UTC 2024
[    0.000000] Command line: BOOT_IMAGE=(hd0,gpt3)/vmlinuz-6.11.4-301.fc41.x86_64 no_timer_check console=tty1 console=ttyS0,115200n8 systemd.firstboot=off root=UUID=815e66c2-6a8a-4984-a890-1a3c710bf933 rootflags=subvol=root
[    0.000000] BIOS-provided physical RAM map:
[    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009fbff] usable
...
## Acquired at 2024-12-03T13:12:23.596179+00:00
...
## Acquired at 2024-12-03T13:12:23.622283+00:00
...
```

Fixes #3317

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/test/check/data/main.fmf
+++ b/tests/test/check/data/main.fmf
@@ -13,6 +13,15 @@
             failure-pattern:
               - 'Hypervisor detected'
 
+    /multiple-reports:
+        check:
+          - how: dmesg
+            result: respect
+          - how: dmesg
+            result: respect
+          - how: dmesg
+            result: respect
+
 /avc:
   check:
     - how: avc
--- a/tests/test/check/test-dmesg.sh
+++ b/tests/test/check/test-dmesg.sh
@@ -20,6 +20,7 @@ rlJournalStart
         rlRun "harmless=$run/plan/execute/data/guest/default-0/dmesg/harmless-1"
         rlRun "segfault=$run/plan/execute/data/guest/default-0/dmesg/segfault-1"
         rlRun "custom_patterns=$run/plan/execute/data/guest/default-0/dmesg/custom-patterns-1"
+        rlRun "multiple_reports=$run/plan/execute/data/guest/default-0/dmesg/multiple-reports-1"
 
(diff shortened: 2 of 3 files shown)


skycastlelily (approved):
LGTM^^

martinhoyer (approved):
Neat

PR #3396, opened 2024-12-03
Implement the `coredump` test check plugin

martinhoyer:
Leaving kdump and abrt to subsequent, separate PRs.

This PR is adding the initial implementation of a `coredump` Test Check.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] mention the version
* [x] include a release note

### DIFF
--- a/spec/tests/check.fmf
+++ b/spec/tests/check.fmf
@@ -10,9 +10,9 @@ description: |
     In some cases we want to run additional checks while running a
     test. A nice example is a check for unexpected SELinux AVCs
     produced during the test, this can point to additional issues
-    the user can run into. Another useful checks would be kernel
-    panic detection, core dump collection or collection of system
-    logs.
+    the user can run into. Other useful checks include kernel panic
+    detection, system crash detection via coredump, or collection of
+    system logs.
 
     By default, the check results affect the overall test outcome.
     To change this behaviour, use the ``result`` key, which accepts
@@ -74,9 +74,18 @@ example:
       - how: dmesg
         result: info
 
+  - |
+    # Configure the coredump check with ignore patterns
+    check:
+      - how: coredump
+        ignore-pattern:
+          - 'Process.*\\(sleep\\).*dumped core'  # Ignore sleep crashes
+          - 'Signal: .*\\(SIGSEGV\\)'            # Ignore segmentation faults
+
 link:
   - implemented-by: /tmt/checks
   - implemented-by: /tmt/result.py
   - verified-by: /tests/test/check/avc
   - verified-by: /tests/test/check/dmesg
+  - verified-by: /tests/test/check/coredump
   - verified-by: /tests/execute/result/check
--- a/tests/test/check/data/main.fmf
+++ b/tests/test/check/data/main.fmf
@@ -61,6 +61,28 @@
                       ls -alZ /root/passwd.log; \
                       rm -f /root/passwd.log" || /bin/true
 
+/coredump:
+    check:
+      - how: coredump
+
+    /harmless:
+        test: /bin/true
+        summary: Test coredump check passes with no crashes
+
+    /segfault:
+        test: |
+            sleep 10 & kill -SIGSEGV %1
+        summary: Test coredump check fails with segfault
+
+    /ignore-pattern:
+        check:
+          - how: coredump
+            ignore-patterns:
+              - 'Signal:\s+\d+\s+\((SIG)?SEGV\)'
+        test: |
+            sleep 10 & kill -SIGSEGV %1
+        summary: Test coredump check passes with ignored crash pattern
+
 /watchdog/ping:
   require:
     - /usr/bin/uptime
--- a/tests/test/check/data/main.fmf
+++ b/tests/test/check/data/main.fmf
@@ -77,7 +77,7 @@
     /ignore-pattern:
         check:
           - how: coredump
-            ignore-patterns:
+            ignore-pattern:
               - 'Signal:\s+\d+\s+\((SIG)?SEGV\)'
         test: |
             sleep 10 & kill -SIGSEGV %1
--- a/tests/test/check/test-coredump.sh
+++ b/tests/test/check/test-coredump.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+function assert_check_result () {
+    rlAssertEquals "$1" "coredump:$2" "$(yq -r ".[] | select(.name == \"$4\") | .check | .[] | select(.event == \"$3\") | \"\\(.name):\\(.result)\"" $results)"
+}
+
+function check_systemd_availability () {
+    has_systemd=true
+    if [ "$PROVISION_HOW" = "container" ] || { [ "$PROVISION_HOW" = "local" ] && ! systemctl --version &>/dev/null; }; then
+        has_systemd=false
+    fi
+    echo $has_systemd
+}
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-virtual}"
+        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        rlRun "results=$run/plan/execute/results.yaml"
+        rlRun "pushd data"
+        rlRun "set -o pipefail"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test harmless run with $PROVISION_HOW"
+        rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /coredump/harmless"
+        rlRun "cat $results"
+
+        # Check if systemd is available in the provisioner
+        has_systemd=$(check_systemd_availability)
+
+        if [ "$has_systemd" = "false" ]; then
+            # Container or local without systemd won't have required capabilities
+            assert_check_result "coredump as a before-test should skip without systemd" "skip" "before-test" "/coredump/harmless"
+            assert_check_result "coredump as an after-test should skip without systemd" "skip" "after-test" "/coredump/harmless"
+        else
+            # Other provisioners should pass with no crashes
+            assert_check_result "coredump as a before-test should pass" "pass" "before-test" "/coredump/harmless"
+            assert_check_result "coredump as an after-test should pass" "pass" "after-test" "/coredump/harmless"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test segfault with $PROVISION_HOW"
+        # Check if systemd is available in the provisioner
+        has_systemd=$(check_systemd_availability)
+
+        if [ "$has_systemd" = "false" ]; then
+            # No systemd means no coredump capabilities, expect success
+            rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /coredump/segfault"
+        else
+            # Other provisioners should detect the crash, expect failure
+            rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /coredump/segfault" "1"
+        fi
+        rlRun "cat $results"
+
+        if [ "$has_systemd" = "false" ]; then
+            # No systemd means no coredump capabilities
+            assert_check_result "coredump as a before-test should skip without systemd" "skip" "before-test" "/coredump/segfault"
+            assert_check_result "coredump as an after-test should skip without systemd" "skip" "after-test" "/coredump/segfault"
+        else
+            # Other provisioners should detect the crash
+            assert_check_result "coredump as a before-test should pass" "pass" "before-test" "/coredump/segfault"
+            assert_check_result "coredump as an after-test should fail" "fail" "after-test" "/coredump/segfault"
+
+            # Verify coredump was captured
+            rlRun "ls -l $run/plan/execute/data/guest/default-0/coredump/segfault-1/checks/dump._usr_bin_bash_SIGSEGV_*.core"
+            rlLogInfo "$(ls -l $run/plan/execute/data/guest/default-0/coredump/segfault-1/checks/)"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test ignore pattern with $PROVISION_HOW"
+        # This test uses the same segfault code but with a pattern to ignore it
+        # Check if systemd is available in the provisioner
+        has_systemd=$(check_systemd_availability)
+
+        # This test should pass regardless of systemd availability
+        rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /coredump/ignore-pattern"
+        rlRun "cat $results"
+
+        if [ "$has_systemd" = "false" ]; then
+            # No systemd means no coredump capabilities
+            assert_check_result "coredump as a before-test should skip without systemd" "skip" "before-test" "/coredump/ignore-pattern"
+            assert_check_result "coredump as an after-test should skip without systemd" "skip" "after-test" "/coredump/ignore-pattern"
+        else
+            # Even though there's a crash, it should pass because of the ignore pattern
+            assert_check_result "coredump as a before-test should pass" "pass" "before-test" "/coredump/ignore-pattern"
+            assert_check_result "coredump as an after-test should pass" "pass" "after-test" "/coredump/ignore-pattern"
+
+            # Verify the crash was still logged even though it was ignored
+            rlRun "ls -l $run/plan/execute/data/guest/default-0/coredump/ignore-pattern-1/checks/"
+            rlLogInfo "$(ls -l $run/plan/execute/data/guest/default-0/coredump/ignore-pattern-1/checks/)"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartCleanup
+        rlRun "popd"
+        rlRun "rm -r $run" 0 "Remove run directory"
+    rlPhaseEnd
+rlJournalEnd
--- a/tmt/checks/coredump.py
+++ b/tmt/checks/coredump.py
@@ -60,7 +60,7 @@ class CoredumpCheck(Check):
         """Convert to raw specification."""
         spec = super().to_spec()
 
-        spec["ignore-patterns"] = [  # type: ignore[reportGeneralTypeIssues,typeddict-unknown-key,unused-ignore]
+        spec["ignore-pattern"] = [  # type: ignore[reportGeneralTypeIssues,typeddict-unknown-key,unused-ignore]
             pattern.pattern for pattern in self.ignore_patterns
         ]
 
@@ -494,7 +494,7 @@ class Coredump(CheckPlugin[CoredumpCheck]):
 
         check:
           - how: coredump
-            ignore-patterns:
+            ignore-pattern:
               - 'Process.*\\(sleep\\).*dumped core'  # Ignore sleep crashes
               - 'Package: ddcutil/2.1.2-2.fc41'      # Ignore dumps of a specific package
 
--- a/tmt/checks/coredump.py
+++ b/tmt/checks/coredump.py
@@ -33,7 +33,7 @@ class CoredumpCheck(Check):
     """Configuration for the coredump check."""
 
     # Patterns to ignore in crash reports
-    ignore_patterns: list[Pattern[str]] = field(
+    ignore_pattern: list[Pattern[str]] = field(
         default_factory=list,
         help="""
              Optional list of regular expressions to ignore in crash reports.
@@ -61,7 +61,7 @@ class CoredumpCheck(Check):
         spec = super().to_spec()
 
         spec["ignore-pattern"] = [  # type: ignore[reportGeneralTypeIssues,typeddict-unknown-key,unused-ignore]
-            pattern.pattern for pattern in self.ignore_patterns
+            pattern.pattern for pattern in self.ignore_pattern
         ]
 
         return spec
@@ -341,13 +341,9 @@ class CoredumpCheck(Check):
                     continue
 
                 # Skip if this crash matches any ignore pattern
-                if self.ignore_patterns:
+                if self.ignore_pattern:
                     matching_pattern = next(
-                        (
-                            pattern
-                            for pattern in self.ignore_patterns
-                            if pattern.search(crash_info)
-                        ),
+                        (pattern for pattern in self.ignore_pattern if pattern.search(crash_info)),
                         None,
                     )
                     if matching_pattern:
--- a/tmt/checks/coredump.py
+++ b/tmt/checks/coredump.py
@@ -94,7 +94,7 @@ class CoredumpCheck(Check):
                 guest.execute(
                     ShellScript("sudo mkdir -p /etc/systemd/coredump.conf.d")
                     & ShellScript(
-                        f"echo '{COREDUMP_CONFIG}' | sudo tee"
+                        f"echo '{COREDUMP_CONFIG}' | sudo tee "
                         "/etc/systemd/coredump.conf.d/50-tmt.conf > /dev/null"
                     )
                 )
@@ -591,7 +591,6 @@ class Coredump(CheckPlugin[CoredumpCheck]):
         # from tmt.utils.hints import get_hints
 
         # Skip if coredumpctl is not available, as detected in before-test
-        sleep(1)
         if not check.is_available:
             return [
                 CheckResult(
--- a/tmt/checks/coredump.py
+++ b/tmt/checks/coredump.py
@@ -355,14 +355,13 @@ class CoredumpCheck(Check):
                         continue
 
                 # Save the crash info (with --all to ensure we can access it)
-                info_filename = f"dump.{exe}_{sig}_{pid}.txt"
-                info_path = check_files_path / info_filename
+                info_filepath = check_files_path / f"dump.{exe}_{sig}_{pid}.txt"
                 guest.execute(
                     ShellScript(
-                        f"sh -c {sudo_prefix}coredumpctl info --all --no-pager {pid} > {info_path!s}"  # noqa: E501
+                        f"sh -c {sudo_prefix}coredumpctl info --all --no-pager {pid} > {info_filepath!s}"  # noqa: E501
                     )
                 )
-                logger.debug(f"Saved crash info to {info_path}")
+                logger.debug(f"Saved crash info to {info_filepath}")
 
                 # Try to save the coredump if available
                 if corefile not in ("none", "missing"):
--- a/tmt/checks/coredump.py
+++ b/tmt/checks/coredump.py
@@ -1,0 +1,614 @@
+import re
+from re import Pattern
+from time import sleep
+from typing import TYPE_CHECKING, Optional
+
+import tmt.log
+import tmt.steps.execute
+import tmt.steps.provision
+import tmt.utils
+from tmt.checks import Check, CheckEvent, CheckPlugin, _RawCheck, provides_check
+from tmt.container import container, field
+from tmt.result import CheckResult, ResultOutcome
+from tmt.utils import Command, Path, ShellScript
+
+if TYPE_CHECKING:
+    import tmt.base
+    from tmt.steps.execute import TestInvocation
+    from tmt.steps.provision import Guest
+
+
+COREDUMP_LAST_DUMP_FILENAME = "coredump-latest"
+
+# Can be set in /etc/coredumpct.conf.d/
+# See `man coredump.conf`
+COREDUMP_CONFIG = """[Coredump]
+Storage=external
+Compress=yes
+"""
+
+
+@container
+class CoredumpCheck(Check):
+    """Configuration for the coredump check."""
+
+    # Patterns to ignore in crash reports
+    ignore_patterns: list[Pattern[str]] = field(
+        default_factory=list,
+        help="""
+             Optional list of regular expressions to ignore in crash reports.
+             If a crash report matches any of these patterns, it will be ignored
+             and not cause a failure. Any other crashes will still cause the test
+             to fail. If no patterns are specified, any crash will cause a failure.
+             """,
+        metavar="PATTERN",
+        normalize=tmt.utils.normalize_pattern_list,
+        exporter=lambda patterns: [pattern.pattern for pattern in patterns],
+        serialize=lambda patterns: [pattern.pattern for pattern in patterns],
+        unserialize=lambda serialized: [re.compile(pattern) for pattern in serialized],
+    )
+
+    # Internal flag to track if we can run coredumpctl on the host
+    is_available: bool = True
+    is_availability_reason: Optional[str] = None
+
+    # Path to the file storing information about coredumps before test execution
+    # Default is current directory, will be properly set in _save_existing_coredumps
+    coredump_last_dumps_filepath: Path = field(default_factory=Path, internal=True)
+
+    def to_spec(self) -> _RawCheck:
+        """Convert to raw specification."""
+        spec = super().to_spec()
+
+        spec["ignore-patterns"] = [  # type: ignore[reportGeneralTypeIssues,typeddict-unknown-key,unused-ignore]
+            pattern.pattern for pattern in self.ignore_patterns
+        ]
+
+        return spec
+
+    def to_minimal_spec(self) -> _RawCheck:
+        """Convert to minimal raw specification."""
+        return self.to_spec()
+
+    def _configure_coredump(self, guest: "Guest", logger: tmt.log.Logger) -> None:
+        """
+        Try configure coredump storage.
+
+        Non-privileged users might not have permission to change the config,
+        while being able to use coredumpctl
+        """
+        try:
+            guest.execute(
+                ShellScript("mkdir -p /etc/systemd/coredump.conf.d")
+                & ShellScript(
+                    f"echo '{COREDUMP_CONFIG}' > /etc/systemd/coredump.conf.d/50-tmt.conf"
+                )
+            )
+            return
+        except tmt.utils.RunError:
+            logger.debug("Unable to configure coredump directly, trying with sudo")
+
+        # If failed and not root, try with sudo
+        if guest.facts.is_superuser is False:
+            try:
+                guest.execute(
+                    ShellScript("sudo mkdir -p /etc/systemd/coredump.conf.d")
+                    & ShellScript(
+                        f"echo '{COREDUMP_CONFIG}' | sudo tee"
+                        "/etc/systemd/coredump.conf.d/50-tmt.conf > /dev/null"
+                    )
+                )
+                logger.debug("Configured coredump with sudo")
+                return
+            except tmt.utils.RunError:
+                logger.debug(
+                    "Unable to configure coredump even with sudo, continuing with default settings"
+                )
+        else:
+            logger.debug("Unable to configure coredump, continuing with default settings")
+
+    def _has_coredumpctl(self, guest: "Guest", logger: tmt.log.Logger) -> bool:
+        """Check if coredumpctl command is available."""
+        try:
+            guest.execute(ShellScript("coredumpctl --version"), silent=True)
+            return True
+        except tmt.utils.RunError:
+            logger.debug("coredumpctl command not found")
+            return False
+
+    def _has_active_socket(self, guest: "Guest", logger: tmt.log.Logger) -> bool:
+        """Check if systemd-coredump.socket is active or can be activated."""
+        try:
+            # Try activating the socket if it's not already active
+            guest.execute(
+                ShellScript(
+                    "systemctl is-active systemd-coredump.socket || "
+                    "systemctl start systemd-coredump.socket"
+                ),
+                silent=True,
+            )
+            return True
+        except tmt.utils.RunError:
+            # If the user isn't root, try with sudo
+            if guest.facts.is_superuser is False:
+                try:
+                    guest.execute(
+                        ShellScript(
+                            "sudo systemctl is-active systemd-coredump.socket || "
+                            "sudo systemctl start systemd-coredump.socket"
+                        ),
+                        silent=True,
+                    )
+                    logger.debug("systemd-coredump.socket activated with sudo")
+                    return True
+                except tmt.utils.RunError:
+                    logger.debug(
+                        "Unable to access or start systemd-coredump.socket, even with sudo"
+                    )
+                    return False
+            else:
+                logger.debug("Unable to access or start systemd-coredump.socket")
+                return False
+
+    def _has_required_permissions(self, guest: "Guest", logger: tmt.log.Logger) -> bool:
+        """
+        Check if we have sufficient permissions to access coredump data.
+
+        Note: coredumpctl returns 1 when no dumps are found, which is a valid
+        case that indicates we have proper permissions.
+        """
+        # Try without sudo first
+        try:
+            # Permission errors shouldn't return 1, while "no dumps found" does.
+            # See https://github.com/systemd/systemd/blob/main/src/basic/errno-util.h
+            guest.execute(ShellScript("coredumpctl info &>/dev/null || [ $? -eq 1 ]"), silent=True)
+            logger.debug("Has permission to access coredump data")
+            return True
+        except tmt.utils.RunError:
+            # If failed, try with sudo if the user is not already root
+            if guest.facts.is_superuser is False:
+                try:
+                    guest.execute(
+                        ShellScript("sudo coredumpctl info &>/dev/null || [ $? -eq 1 ]"),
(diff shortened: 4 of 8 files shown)


martinhoyer:
Updated description, rebased to main. Marking as ready for review as I'd really appreciate some feedback.  

@happz continuing the discussion from chat - this is the only solution I found for `GuestCapabilities` and `_query_capabilities`, with which I can sleep at night...I hope it's ok for you too, once you check it *after* your PTO :)

tmt/steps/provision/__init__.py
```diff
@@ -285,14 +285,62 @@ class CheckRsyncOutcome(enum.Enum):
...
+    Linux capabilities as defined in capabilities.h.
+
+    Each enum member represents a Linux capability that can be granted to processes.
+    The integer value of each member corresponds to its bit position in the kernel's
+    capability bitmask, as defined in the Linux kernel's capabilities.h.
+
+    These values are stable across kernel versions as they are part of the kernel's
+    ABI guarantee. New capabilities can only be added with new numbers at the end.
+
+    See: include/uapi/linux/capability.h in the Linux kernel source.
```
happz, line 299: We have an actual, official link: https://github.com/torvalds/linux/blob/master/include/uapi/linux/capability.h

tmt/steps/provision/__init__.py
```diff
@@ -317,31 +365,43 @@ class GuestFacts(SerializableContainer):
...
-        unserialize=lambda raw_value: {
-            GuestCapability(raw_capability): enabled
-            for raw_capability, enabled in raw_value.items()
-            }
-        )
+    #: Permitted Linux capabilities on this guest.
+    capabilities: Optional[list[GuestCapability]] = field(
+        default=None,
+        serialize=lambda caps: [cap.value for cap in caps] if caps else None,
+        unserialize=lambda serialized: [GuestCapability(value) for value in serialized]
```
happz, line 376: Just a nitpicking: IIUIC the code here, it'd emit a list of integers, which wouldn't be nice to read by humans. Maybe we should emit the names instead of values here, it would be a bit more work when unserializing, but inspecting the on-disk files would be easier for puny humans.

tmt/steps/provision/__init__.py
```diff
@@ -317,31 +365,43 @@ class GuestFacts(SerializableContainer):
...
+
+        :param cap: The capability to check for.
+        :returns: True if the guest has the specified capability, False otherwise.
+        """
+        if not self.capabilities or cap not in self.capabilities:
+            return False
+
+        return True
+
+    def has_capabilities(self, caps: list[GuestCapability]) -> bool:
```
happz, line 394: All we need is an iterable, so `*caps: GuestCapability` would give us a tuple, and we could call this method without constructing a list, `has_capability(GuestCapability.CAP_SYSLOG, GuestCapability.CAP_PERFMON)`.

And then you can drop the `has_capability` method because it would be the same as `has_capabilities(a_single_capability)` :)

tmt/steps/provision/__init__.py
```diff
@@ -579,13 +651,38 @@ def _query_is_ostree(self, guest: 'Guest') -> Optional[bool]:
...
+        bitmask_hex = output.stdout.split()[1]
+        # TODO If hex = 0000000000000000, we could try again to read /proc/self/status with
+        # sudo to get capabilities available for superuser
+        bitmask = int(bitmask_hex, 16)
+
+        # Convert the bitmask to a list of permitted capabilities
+        capabilities: list[GuestCapability] = []
+        for cap in GuestCapability:
+            if bitmask & (1 << cap.value):
+                capabilities.append(cap)
```
happz, line 683: List comprehension to the rescue! Or at least to a slightly shorter & more elegant oneliner without an accumulator variable:

```python
return [cap for cap in GuestCapability if bitmask & (1 << cap.value)]
```

tmt/steps/provision/podman.py
```diff
@@ -475,11 +474,6 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         self._guest.start()
         self._guest.setup()
 
-        # TODO: this might be allowed with `--privileged`...
-        self._guest.facts.capabilities[GuestCapability.SYSLOG_ACTION_READ_ALL] = False
-        # ... while this seems to be forbidden completely.
-        self._guest.facts.capabilities[GuestCapability.SYSLOG_ACTION_READ_CLEAR] = False
```
happz, line 481: I'm not sure we should give up the distinction between these two "capabilities". Yes, it's a single capability, `CAP_SYSLOG`, but what I tried to emulate by hijacking the capabilities enum is the different level of functionality allowed: a privileged container seems to have a limited console ring buffer access, and it is possible to read the messages, but not remove them. That is something I would like to preserve, it would give users with container if not a clear picture of "this is what happened while the test was running", but at least a "this is all that has happened, sorry, can't do better".

I think it's no longer practical to squeeze this into `GuestCapabilities`, but having a way of tracking the "can read" vs "can read and erase" flavor of `CAP_SYSLOG` would be nice. I'll try to find a test that would demonstrate this behavior.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,350 @@
...
+            # This should be default anyway
+            config = """[Coredump]
+Storage=external
+Compress=yes
+ProcessSizeMax=32G
+ExternalSizeMax=32G
+JournalSizeMax=767M
+MaxUse=0
+KeepFree=0
+"""
```
happz, line 74: I would make this a module-level string, with some nice name. It's pretty much a multiline constant, method indentation doesn't suit it.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,350 @@
...
+            guest.execute(
+                Command(
+                    "sh",
+                    "-c",
+                    f"mkdir -p /etc/systemd/coredump.conf.d && "
+                    f"echo '{config}' > /etc/systemd/coredump.conf.d/50-tmt.conf",
```
happz, line 80: Check out `DEFAULT_SHELL`, I believe it's the one we should use for running shell commands.

BTW `ShellScript` might be more readable, `&&` plus a missing comma seems like a trap. How about this?

```python
guest.execute(ShellScript(
    f"""
    mkdir -p /etc/systemd/coredump.conf.d && "echo '{config}' > /etc/systemd/coredump.conf.d/50-tmt.conf
    """)

# This should also work:
guest.execute(
    ShellScript('mkdir -p /etc/systemd/coredump.conf.d')
    and ShellScript(f"echo '{config}' > /etc/systemd/coredump.conf.d/50-tmt.conf"))
```

`guest.execute()` should transparently convert shell script into command, with`DEFAULT_SHELL -c ...`

tmt/checks/coredump.py
```diff
@@ -0,0 +1,350 @@
...
+                    invocation.guest, logger, str(invocation.check_files_path)):
+                return ResultOutcome.ERROR
+
+            # Create timestamp directory and store current time for --since filtering
+            invocation.guest.execute(
+                Command(
+                    "sh",
+                    "-c",
+                    f"mkdir -p {DEFAULT_TMP_PATH} && "
+                    f"date '+%Y-%m-%d %H:%M:%S' > {DEFAULT_TMP_PATH}/coredump-timestamp"
```
happz, line 230: A single, shared, mutable, global state. All tmt invocations sharing the guest - e.g. with `provision -h local` - will change this location without any kind of synchronization. It's better to use a directory that cannot conflict with anyone else, e.g. AVC check uses `ausearch_timestamp_filepath = invocation.check_files_path / AUSEARCH_TIMESTAMP_FILENAME`.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,350 @@
...
+            environment: Optional[tmt.utils.Environment] = None,
+            logger: tmt.log.Logger,
+            ) -> list[CheckResult]:
+        """Check for crashes before the test starts."""
+        # Required capabilities for full coredump functionality
+        required_capabilities = [
+            GuestCapability.CAP_SYS_ADMIN,    # Configure coredump
+            GuestCapability.CAP_SYSLOG,       # Read journal
+            GuestCapability.CAP_DAC_READ_SEARCH,  # Access coredump files
+            GuestCapability.CAP_DAC_OVERRIDE  # Access protected files
```
happz, line 316: Why aren't these two capabilities tested in `essential_requires()`?

tmt/checks/coredump.py
```diff
@@ -0,0 +1,350 @@
...
+
+        except tmt.utils.RunError as exc:
+            logger.debug(f"Failed to check for crashes: {exc}")
+            return False
+
+    def _check_coredump(
+            self, invocation: "TestInvocation", event: CheckEvent, logger: tmt.log.Logger
+            ) -> ResultOutcome:
+        """Check coredump status and return appropriate result."""
+        # Only configure coredump in before_test
```
happz, line 217: Maybe the branch below should be in its own method, e.g. `create_coredump_timestamp()`, which would be then called only before the test. Written like this, the method depends on external state and decides on its own what to do, and if there's a condition like this the outside world knows, why even giving this method a chance to not obey? `before_test` will call it, `after_test` will not.

tmt/checks/coredump.py
happz: Plus the `sh -c`, seems to be present in many commands. `ShellScript` will serve us better, `guest.execute(ShellScript(f"mkdir -p {DEFAULT_TMP_PATH} && date '+%Y-%m-%d %H:%M:%S' > {DEFAULT_TMP_PATH}/coredump-timestamp"))`

happz:
> @happz continuing the discussion from chat - this is the only solution I found for `GuestCapabilities` and `_query_capabilities`, with which I can sleep at night...I hope it's ok for you too,

As expected, it's a lot of work to get the list of allowed capabilities, but not as much as I was worried. We have our own enum with a lot of entries we don't need, but these new checks start using entries we didn't use before, I guess we may easily find new plugins checking yet another entries in the future. And the list is pretty stable, so LGTM.

> once you check it _after_ your PTO :)

Yeah, yeah, I know, but Geralt needs a break anyway, Ciri is safe, so I have some spare time on my hands.

tmt/steps/provision/__init__.py
martinhoyer: I've initially had a `Final[list[str]]`, but creating a type, `Literal[*<the final list>]` is not possible on Python <3.10 and while `Literal[tuple(<final list>)] does work, mypy can't deal with it.  Using `eumerate` didn't feel right to me, as it's a bit of black box, but I'd be happy to implement whatever you say is the best way @happz :)

tmt/steps/provision/__init__.py
martinhoyer: Weird that wasn't caught by ruff.

tmt/steps/provision/podman.py
martinhoyer: ok, looks like I misunderstood your chat message earlier. 
So if I understand it correctly, we could deduce this based on is-container, is-rootful, is-root / has-capability? (or something like that)

tmt/checks/coredump.py
martinhoyer: Good question - basically this is all a guesswork, I don't actually know if we need these, or any for that matter. Also, I don't think you need `CAP_SYSLOG`, to read journal, at least on my laptop, I have cap_syslog 'not permitted', but can do journalctl, as well as run coredumpctl.

I was experimenting with some ways to figure out if want to skip the coredump check, like
 - doesn't have systemd -> skip
 - doesn't have coredumpctl
   - doesn't have permission to modify coredumpctl config -> skip
   - install coredumpctl
- etc...

But I think I'm overthinking it, imho it can be just has_systemd and can modify the config. Skip everything else, wdyt? (removing essential_requires alltogether)

martinhoyer:
> > once you check it _after_ your PTO :)
> 
> Yeah, yeah, I know, but Geralt needs a break anyway, Ciri is safe, so I have some spare time on my hands.

Have you just finished the books or the games? :) 
Thanks for the review! Will try to address most of the comments asap.

tmt/steps/provision/__init__.py
martinhoyer: Addressed in 62ad828b667fc6680072f375e6bb30edff9725a9

tmt/steps/provision/__init__.py
martinhoyer: Addressed in 62ad828b667fc6680072f375e6bb30edff9725a9

tmt/steps/provision/__init__.py
martinhoyer: Addressed in 62ad828b667fc6680072f375e6bb30edff9725a9

tmt/checks/coredump.py
martinhoyer: Addressed in c0c05df176604cf7fa3ec96d92f952e03b886726

tmt/checks/coredump.py
martinhoyer: I've used ShellScript everywhere, except where it would require shell=True, at least as far as I understand.

tmt/checks/coredump.py
martinhoyer: simplified in c0c05df176604cf7fa3ec96d92f952e03b886726

tmt/checks/coredump.py
martinhoyer: Addressed in c0c05df176604cf7fa3ec96d92f952e03b886726, though at some point I became quite confused about what should be where, so reviews needed :)

tmt/steps/provision/__init__.py
happz: I'm not sure I can catch up with the last comment. My point was to emit `cap.name` instead of `cap.value`: in files, we would see nice `CAP_FOO` instead of `79`. Unsrialize would need to do a bit more work, `GuestCapability.__members__[value]`.

tmt/steps/provision/podman.py
happz: > So if I understand it correctly, we could deduce this based on is-container, is-rootful, is-root / has-capability? (or something like that)

Maybe. I kind of liked having it as a "capability", even an invented one, since `SYSLOG_ACTION_READ_ALL` and `SYSLOG_ACTION_READ_CLEAR` are genuine kernel buffer commands (https://linux.die.net/man/3/klogctl), so it's something that exists and is documented. Plus, the container plugin is aware of the situation already, it would be in better position to change these flags, the rest of tmt would just use them instead of inferring from other flags what to do.

So, as of now, the best solution to me seems to be adding them back to `GuestCapability` enum, with some high value. Basically reverting this bit here.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,288 @@
...
+
+# Can be set in /etc/coredumpct.conf.d/
+# See `man coredump.conf`
+COREDUMP_CONFIG = """[Coredump]
+Storage=external
+Compress=yes
+"""
+
+# Can be set in /proc/sys/kernel/core_pattern
+# CORE_PATTERN = '|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h'
```
happz, line 31: I'm not sure to which this comment belongs.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,288 @@
...
+                    ignored = False
+                    for pattern in self.ignore_patterns:
+                        if pattern.search(crash_info):
+                            logger.debug(
+                                f"Ignoring crash due to pattern '{pattern.pattern}': {crash_info}"
+                                )
+                            ignored = True
+                            break
+                    if ignored:
+                        continue
```
happz, line 155: If we could live without knowing which pattern matched, this could be a one-liner with `any()`.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,288 @@
...
+        """Check for crashes before the test starts."""
+
+        if not invocation.guest.facts.has_systemd or not check._check_coredump_available(
+                invocation.guest, logger):
+            logger.debug("coredump not available, skipping..")
+            check.is_available = False
+            return [CheckResult(name="coredump", result=ResultOutcome.SKIP)]
+
+        if not check._create_coredump_timestamp(invocation):
+            outcome = ResultOutcome.ERROR
```
happz, line 266: It seems to me this particular change of `outcome` value is ignored, as it can be overwritten by the future call to `check._check_coredump()`.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,288 @@
...
+
+            return has_crashes
+
+        except tmt.utils.RunError as exc:
+            logger.debug(f"Failed to check for crashes: {exc}")
+            return False
+
+    def _check_coredump(
+            self, invocation: "TestInvocation", event: CheckEvent, logger: tmt.log.Logger
+            ) -> ResultOutcome:
```
happz, line 188: In the name of connecting dots and letting users in, would it make sense to return not just an outcome, but also the list of logs, i.e. info and coredump files created by `_check_for_crashes()`? I would imagine `_check_for_crashes()` and `_check_coredump()` would basically collect the paths, and add them to `logs` field of the eventual check result, together with the outcome.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,288 @@
...
+            check: "CoredumpCheck",
+            invocation: "TestInvocation",
+            environment: Optional[tmt.utils.Environment] = None,
+            logger: tmt.log.Logger,
+            ) -> list[CheckResult]:
+        """Check for crashes after the test finishes."""
+        # Skip if coredumpctl is not available, as detected in before-test
+        sleep(1)
+        if not check.is_available:
+            return [CheckResult(name="coredump", result=ResultOutcome.SKIP)]
```
happz, line 285: I forgot to push it, but, I'm trying to expose as much as possible to end users and CI systems, please, add a note to SKIP outcomes: https://github.com/teemtee/tmt/pull/3498

tmt/checks/coredump.py
```diff
@@ -0,0 +1,288 @@
...
+                invocation.guest, logger):
+            logger.debug("coredump not available, skipping..")
+            check.is_available = False
+            return [CheckResult(name="coredump", result=ResultOutcome.SKIP)]
+
+        if not check._create_coredump_timestamp(invocation):
+            outcome = ResultOutcome.ERROR
+
+        check._configure_coredump(invocation.guest, logger)
+        outcome = check._check_coredump(invocation, CheckEvent.BEFORE_TEST, logger)
```
happz, line 269: What is expected from checking for coredumps that appear between the timestamp and the beginning of the test?

tmt/steps/provision/__init__.py
```diff
@@ -349,34 +394,47 @@ class GuestFacts(SerializableContainer):
...
+        unserialize=lambda serialized: [GuestCapability(value) for value in serialized]
+        if serialized
+        else None,
     )
 
     os_release_content: dict[str, str] = field(default_factory=dict)
     lsb_release_content: dict[str, str] = field(default_factory=dict)
 
-    def has_capability(self, cap: GuestCapability) -> bool:
+    def has_capability(self, capability: GuestCapability) -> bool:
```
happz, line 415: Seems like https://github.com/teemtee/tmt/commit/62ad828b667fc6680072f375e6bb30edff9725a9 got lost somewhere during the rebase.

martinhoyer: Indeed. I got tangled in mypy errors. Will get it back.

martinhoyer: ok it's back

tmt/steps/provision/podman.py
```diff
@@ -626,7 +625,7 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
+    @property
```
happz, line 628: This property should not be needed anymore, the one provided by the base class should be doing the very same thing.

tmt/checks/coredump.py
martinhoyer: also addressed

tmt/checks/coredump.py
martinhoyer: Should be gone.

tmt/checks/coredump.py
martinhoyer: Soo, I've rebased it on top of https://github.com/teemtee/tmt/pull/3498 to avoid dealing with it afterwards

tmt/steps/provision/podman.py
martinhoyer: should be gone.

tmt/checks/coredump.py
martinhoyer: hopefully addressed

martinhoyer:
Added some polish and rebased on main. @happz Can you spot something very bad straight away? fwiw, I'm not super confident about the code, as it's been quite some time and I don't remember everything, on the other hand, there shouldn't be breaking changes, so if @psss, et all wants to have it in the release, maybe as a preview, I can add the docs, etc.

psss:
The two red tests seem to be irrelevant failures (failed to boot instance) but the new coredump test seems to be [failing](https://artifacts.osci.redhat.com/testing-farm/811e9932-e9d6-42d5-b295-fa3853c2042d/work-the-restj1xlc9yw/plans/provision/virtual/the-rest/execute/data/guest/default-0/tests/test/check/coredump-15/output.txt) as well:

```
00:00:00     skip coredump (Coredump detection was skipped because coredumpctl is not available or has insufficient privileges.
The coredump check requires systemd-coredump to be installed and the systemd-coredump.socket to be active. Additionally, the user must have sufficient permissions to access coredump data.) (before-test check)
```

tmt/checks/coredump.py
```diff
@@ -0,0 +1,412 @@
...
+        # from tmt.utils.hints import get_hints
+
+        # Skip if coredumpctl is not available, as detected in before-test
+        sleep(1)
+        if not check.is_available:
+            return [
+                CheckResult(
+                    name="coredump",
+                    result=ResultOutcome.SKIP,
+                    note=[cls._get_not_available_message()],
```
martinhoyer, line 398: This needs to be improved
```
        note:
          - |-
            Coredump detection was skipped because coredumpctl is not available or has insufficient privileges.

            The coredump check requires systemd-coredump to be installed and the systemd-coredump.socket to be active. Additionally, the user must have sufficient permissions to access coredump data.
            ```

Comment (happz) [reply tmt/checks/coredump.py:398]: Yes, it does, and maybe it will: https://github.com/teemtee/tmt/pull/3596/files#diff-f9ad5f6605f10698f0208842ba0e42ee7bf1508151827baf78c3660b611e4457R49

Comment (martinhoyer) [reply tmt/checks/coredump.py:398]: @happz I meant the way I'm adding note here, not the display. Are you saying it's ok here?

Comment (happz) [reply tmt/checks/coredump.py:398]: I see. I'm not sure what you need to improve then, didn't catch my eye, it will just look ugly, I'm afraid, being longer than just a couple of words.

File: tmt/checks/coredump.py [hunk 20]
```diff
@@ -0,0 +1,412 @@
...
+
+        # Then check for the required components independently
+        return (
+            self._has_coredumpctl(guest, logger)
+            and self._has_active_socket(guest, logger)
+            and self._has_required_permissions(guest, logger)
+        )
+
+    def _check_for_crashes(
+        self, guest: "Guest", logger: tmt.log.Logger, check_files_path: str, start_time: str
```
happz, line 146: A path shall be a `Path`. `invocation.guest, logger, str(invocation.check_files_path), start_time` makes it a string, and than it's used as a string instead of much nicer `pathlib` API: `info_file = check_files_path / f"dump.{exe}_{sig}_{pid}.txt"`, `dump_file = check_files_path / f"dump.{exe}_{sig}_{pid}.core"`

tmt/checks/coredump.py
```diff
@@ -0,0 +1,412 @@
...
+
+        # Check for crashes
+        has_crashes = self._check_for_crashes(
+            invocation.guest, logger, str(invocation.check_files_path), start_time
+        )
+
+        # Get list of generated files
+        try:
+            files_output = invocation.guest.execute(
+                Command("find", str(invocation.check_files_path), "-type", "f")
```
happz, line 244: I believe `Command` accepts `Path` instances as well as strings. `Command("find", invocation.check_files_path, "-type", "f")` should work like a charm.

tmt/checks/coredump.py
martinhoyer: My bad, thanks! How about https://github.com/teemtee/tmt/pull/3396/commits/f834f2957d640db207ba9a68c5fedfb45093dfc2 ?

tmt/checks/coredump.py
martinhoyer: Trying this https://github.com/teemtee/tmt/pull/3396/commits/0d506d979a2eb90ef53bfccb3a46719081144134

tmt/checks/coredump.py
martinhoyer: Thought so, but didn't know if it's a good practice, if you say so, I'll do the changes :)

martinhoyer: https://github.com/teemtee/tmt/pull/3396/commits/b03bb6b1725dc88446223529243905558ec1c22a

martinhoyer:
> ```
> 00:00:00     skip coredump (Coredump detection was skipped because coredumpctl is not available or has insufficient privileges.
> The coredump check requires systemd-coredump to be installed and the systemd-coredump.socket to be active. Additionally, the user must have sufficient permissions to access coredump data.) (before-test check)
> ```

I presume we don't have permissions to run coredumpctl also on virtual machines - should we?
Adding VMs to "expected skip" and trying to improve the messages: https://github.com/teemtee/tmt/pull/3396/commits/c69e2657b474a787f56355f97da862fb2c7a05cc https://github.com/teemtee/tmt/pull/3396/commits/0d506d979a2eb90ef53bfccb3a46719081144134

martinhoyer:
> I presume we don't have permissions to run coredumpctl also on virtual machines - should we?  

hmm, maybe that not a good presumption. Will investigate...

happz:
> > I presume we don't have permissions to run coredumpctl also on virtual machines - should we?
> 
> hmm, maybe that not a good presumption. Will investigate...

Yeah, running `coredumpctl` in a full-fledged VM, as `root`, that does sound like something we should be able to do. If I'm not mistaken, there is no issue with AVC check in the same environment.

martinhoyer:
```
[root@default-0 tree]# systemctl status systemd-coredump.socket
● systemd-coredump.socket - Process Core Dump Socket
     Loaded: loaded (/usr/lib/systemd/system/systemd-coredump.socket; static)
     Active: active (listening) since Wed 2025-03-26 15:34:30 UTC; 1min 38s ago
 Invocation: 7dbf694668af40efab524431d094d022
       Docs: man:systemd-coredump(8)
     Listen: /run/systemd/coredump (SequentialPacket)
   Accepted: 0; Connected: 0;
     CGroup: /system.slice/systemd-coredump.socket
 ```
 coredump seems to be up on a vanilla `Fedora-Cloud-Base-Generic` VM for me. Maybe it was caused by the unrelated failures you mentioned @psss ?

**Comment by martinhoyer** [comment]
oh lord:
```
# coredumpctl list --no-pager
No coredumps found.
[root@vm-10-0-184-99 ~]# echo $?
1
```
```
def _has_required_permissions(self, guest: "Guest", logger: tmt.log.Logger) -> bool:
        """Check if we have sufficient permissions to access coredump data."""
        try:
            guest.execute(ShellScript("coredumpctl list --no-pager")
```

**Comment by martinhoyer** [comment]
F40 run does not seem to be related:
```
[   FAIL   ] :: File '/var/tmp/rlRun_LOG.bTPO2w8p' should contain 'PLAY \[Enable EPEL repositories\]' 
[   FAIL   ] :: File '/var/tmp/rlRun_LOG.bTPO2w8p' should contain 'Run .* successfully finished. Bye for now!' 
::   RESULT: FAIL (Epel)
```
`tests/try/basic/Epel`

File: tmt/checks/coredump.py (hunk 20 above)
Comment (happz) [reply tmt/checks/coredump.py]: WRT

```
                 info_filename = f"dump.{exe}_{sig}_{pid}.txt"
                 info_path = check_files_path / info_filename
```

Would `info_filepath = check_files_path / f"dump.{exe}_{sig}_{pid}.txt"` be too long? I'm not a fan of variables existing just to be used once unless there's some issue with readability or formatting, and `dump_filename` & `info_filename` do not seem to have any other purpose. Also, I'd recommend `*_filepath` naming scheme, I try to use it consistently throughout the tmt codebase, because "file" is quite overloaded, and "filepath" & "dirpath" are cleaner while being just slightly longer.

File: tmt/checks/coredump.py (hunk 19 above)
Comment (happz) [reply tmt/checks/coredump.py]: I wouldn't invest too much work into this particular area; if I get my way with hints and how they are rendered on the console and elsewhere, these messages will get polished anyway to be informative and nice to read. On the other hand, I have no idea when said changes can land in tmt, so maybe it's not a bad idea to have at least something nicer to read.

File: tmt/checks/coredump.py (hunk 20 above)
Comment (martinhoyer) [reply tmt/checks/coredump.py]: Agree, thanks! Will change when near a keyboard

**Comment by psss** [comment]
Seems the problem is still [there](https://artifacts.osci.redhat.com/testing-farm/02723c7e-1839-4520-93c2-d19e68a68196/work-the-restl4nwus93/plans/provision/virtual/the-rest/execute/data/guest/default-0/tests/test/check/coredump-15/output.txt):

```
pass /coredump/harmless
    output.txt: /tmp/tmp.LysbJDVUpn/plan/execute/data/guest/default-0/coredump/harmless-1/output.txt
    skip coredump (Coredump detection was skipped: insufficient permissions to access coredump data.

    The coredump check requires systemd-coredump to be installed and the systemd-coredump.socket to be active. Additionally, the user must have sufficient permissions to access coredump data.) (before-test check)
    skip coredump (Coredump detection was skipped: insufficient permissions to access coredump data.

    The coredump check requires systemd-coredump to be installed and the systemd-coredump.socket to be active. Additionally, the user must have sufficient permissions to access coredump data.) (after-test check)
```

**Comment by martinhoyer** [comment]
Hopefully complete now.  

I've added `match-patterns` feature to a different branch - https://github.com/martinhoyer/tmt/compare/coredump...teemtee:tmt:coredumo-match-patterns 
To be implemented once we have a good way to run multiple checks with the same name.

File: tests/test/check/test-coredump.sh [hunk 22]
```diff
@@ -74,7 +74,7 @@ rlJournalStart
         has_systemd=$(check_systemd_availability)
 
         # This test should pass regardless of systemd availability
-        rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /coredump/ignore-pattern"
+        rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /coredump/ignore-pattern -ddddvvvv"
```
martinhoyer, line 77: argh, a leftover from the endless pain and suffering.

tmt/checks/coredump.py
```diff
@@ -147,32 +147,28 @@ def _has_active_socket(self, guest: "Guest", logger: tmt.log.Logger) -> bool:
...
+        """
+        Check if we have sufficient permissions to access coredump data.
+
+        Note: coredumpctl returns 1 when no dumps are found, which is a valid
+        case that indicates we have proper permissions.
+        """
         # Try without sudo first
         try:
-            guest.execute(ShellScript("coredumpctl list --no-pager"), silent=True)
+            guest.execute(ShellScript("coredumpctl info &>/dev/null || [ $? -eq 1 ]"), silent=True)
```
martinhoyer, line 158: I probably should have added extra comment. It is my understanding that returncode 1 is safe to use as any permission or other wierd problems would return a different code.  

That said, I gave up trying to decipher https://github.com/systemd/systemd/blob/main/src/basic/errno-util.h after 5 seconds.

tmt/checks/coredump.py
```diff
@@ -218,6 +214,49 @@ def _check_coredump_available(
 
         return True, ""
 
+    def _wait_for_coredump_processes(self, guest: "Guest", logger: tmt.log.Logger) -> None:
```
martinhoyer, line 217: Would be cool to have a test for this, but I'm definitely gonna be able to find better things to do in near future.

spec/tests/check.fmf
```diff
@@ -10,9 +10,9 @@ description: |
     In some cases we want to run additional checks while running a
     test. A nice example is a check for unexpected SELinux AVCs
     produced during the test, this can point to additional issues
-    the user can run into. Another useful checks would be kernel
-    panic detection, core dump collection or collection of system
-    logs.
+    the user can run into. Other useful checks include kernel panic
```
martinhoyer, line 13: Well, this might not fly, but it technically detects kernel panic :lab_coat:

tmt/checks/coredump.py
```diff
@@ -0,0 +1,601 @@
...
+      `'Package: package-name/.*'`
+
+    - Ignore crashes during a specific command:
+      `'Command Line: .*specific-command-pattern.*'`
+
+    .. versionadded:: 1.46
+    """
+
+    @staticmethod
+    def _get_not_available_message(reason: str = "") -> str:
```
happz, line 510: I'd recommend using `None` instead of `""` - it is much easier for type checkers to follow `None` around, while `""` is a valid string. For signaling "unset" or "unknown", `None` is much better choice. Applies to `is_availability_reason` too, it should be `Optional[str]`.

spec/tests/check.fmf
```diff
@@ -74,9 +74,18 @@ example:
       - how: dmesg
         result: info
 
+  - |
+    # Configure the coredump check with ignore patterns
+    check:
+      - how: coredump
+        ignore-patterns:
```
psss, line 81:
```suggestion
        ignore-pattern:
```

Let's use singular to be consistent with other keys.

tests/test/check/data/main.fmf
```diff
@@ -61,6 +61,32 @@
                       ls -alZ /root/passwd.log; \
                       rm -f /root/passwd.log" || /bin/true
 
+/coredump:
+    check:
+      - how: coredump
+    tag+:
+      - provision-only
+      - provision-container
+      - provision-virtual
```
psss, line 70: This tags are only used for the top-level test selection, not used under the `data` directories.

tmt/checks/coredump.py
```diff
@@ -0,0 +1,601 @@
...
+                crash_info = guest.execute(ShellScript(cmd)).stdout
+                if not crash_info:
+                    logger.debug(f"No crash info available for PID {pid}")
+                    continue
+
+                # Skip if this crash matches any ignore pattern
+                if self.ignore_patterns and any(
+                    pattern.search(crash_info) for pattern in self.ignore_patterns
+                ):
+                    logger.debug(f"Ignoring crash due to pattern match in: {crash_info}")
```
psss, line 345: Would be nice to include a `note` that crash happened but was ignored based on the `ignore-pattern` key.

martinhoyer, line 345: Add a note to CheckResult saying it was matched and switch from .debug to .info

happz: Please, add a subissue to https://github.com/teemtee/tmt/issues/3553, that's right in the area.

psss: Is this supposed to be a hint? This is just a regular check result note, something similar to `test raised avc as expected` here we get `crash detected but ignored as instructed`.

happz: Right, I wouldn't make it a hint, it's just the area I attempt to poke, i.e. make very trivial messages like "timeout" more verbose and informative. This covers not just hints, but also result notes and exceptions.

Feel free to file a standalone issue, if you do so, please refer to #3553 so I can keep track of what's left.

psss: Ok, filed as #3653.

tmt/checks/coredump.py
martinhoyer: Addressed

tmt/checks/coredump.py
martinhoyer: As discussed, adding a note wouldn't be trivial here. Can be done in other PR, alongside the other TODOs waiting for #3498.
I've added there at least: `logger.info( f"Ignoring crash in PID {pid} ({exe}) due to pattern match: '{matching_pattern.pattern}'"`

tmt/checks/coredump.py
```diff
@@ -0,0 +1,614 @@
...
+        except tmt.utils.RunError:
+            logger.debug("Unable to configure coredump directly, trying with sudo")
+
+        # If failed and not root, try with sudo
+        if guest.facts.is_superuser is False:
+            try:
+                guest.execute(
+                    ShellScript("sudo mkdir -p /etc/systemd/coredump.conf.d")
+                    & ShellScript(
+                        f"echo '{COREDUMP_CONFIG}' | sudo tee"
```
therazix, line 97: Missing space. Without it, the script would look like this: `...sudo tee/etc/...`
```suggestion
                        f"echo '{COREDUMP_CONFIG}' | sudo tee "
```

psss:
> Addressed feedback, rebased on main.

Thanks for the update. Next time, could you please add new commits instead of amending and force-pushing? It helps with the review to see what has changed, especially when combined with rebase which brings other irrelevant changes so even github's `martinhoyer force-pushed` links do not help.

tests/test/check/data/main.fmf
```diff
@@ -61,6 +61,28 @@
...
+
+    /segfault:
+        test: |
+            sleep 10 & kill -SIGSEGV %1
+        summary: Test coredump check fails with segfault
+
+    /ignore-pattern:
+        check:
+          - how: coredump
+            ignore-patterns:
```
psss, line 80: Here the plural form is still used and the test has passed? Seems the implementation has not been adjusted accordingly.

tests/test/check/data/main.fmf
psss: Thanks!

tmt/checks/coredump.py
```diff
@@ -0,0 +1,614 @@
...
+                has_new_crashes = True
+
+            return has_new_crashes
+
+        except tmt.utils.RunError as exc:
+            logger.debug(f"Failed to check for crashes: {exc}")
+            return False
+
+    def _check_coredump(
+        self, invocation: "TestInvocation", event: CheckEvent, logger: tmt.log.Logger
```
therazix, line 397: The `event` parameter is not used anywhere. Is there a reason to keep it?

tmt/checks/coredump.py
```diff
@@ -0,0 +1,614 @@
...
+        invocation: "TestInvocation",
+        environment: Optional[tmt.utils.Environment] = None,
+        logger: tmt.log.Logger,
+    ) -> list[CheckResult]:
+        """Check for crashes after the test finishes."""
+        # TODO: Uncomment when PR #3498 is merged
+        # from tmt.utils.hints import get_hints
+
+        # Skip if coredumpctl is not available, as detected in before-test
+        sleep(1)
```
therazix, line 594: Forgotten `sleep`?

martinhoyer: Ha! Correct :)

martinhoyer:
> Next time, could you please add new commits instead of amending and force-pushing? It helps with the review to see what has changed, especially when combined with rebase which brings other irrelevant changes so even github's martinhoyer force-pushed links do not help.

@psss 
Yep, also not happy how it turned out, apologies. I thought that if I first add separate commits, push them and do the rebase+push it will look decent, but alas it's broken. I learned my lesson :saluting_face:

tests/test/check/data/main.fmf
martinhoyer: updated

tmt/checks/coredump.py
martinhoyer: fixed, thanks!

tests/test/check/data/main.fmf
psss: Seems to be fixed now. Thanks.

tmt/checks/coredump.py
therazix: Do we need it? The function is called only from `before_test` and `after_test`. And I don't see anything that says that the `_check_coredump` function must have the `event` parameter, so you can remove it from the call too. Or am I missing something?

psss (approved):

therazix (approved):

tmt/checks/coredump.py
psss: This one seems to be addressed.

tmt/checks/coredump.py
psss: Ok, just for reference: This one has be actually covered: `_check_for_crashes` already uses:

    check_files_path: Path,

And Martin adjusted the `filepath`  naming in 50eb7b616fef03bf1e90c7c6cbfb18b622bb99fe.

PR #3403, opened 2024-12-04
Enabling Perflint checks

martinhoyer:
Will probably have negligible performance improvements, but it seems like a low-hanging fruit.   

Fixed [PERF401](https://docs.astral.sh/ruff/rules/manual-list-comprehension/) and [PERF403](https://docs.astral.sh/ruff/rules/manual-dict-comprehension/) issues, and 2 of the [PERF203](https://docs.astral.sh/ruff/rules/try-except-in-loop/) issues. The remaining PERF203 issues are more complex and might require more significant refactoring, which could affect readability and functionality.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -303,6 +303,7 @@ select = [
     "W",  # pycodestyle warning
     "I",  # isort
     "N",  # pep8-naming
+    "PERF",  # Perflint
     "D",    # pydocstyle
     "UP",  # pyupgrade
     "YTT",  # flake8-2020
@@ -357,6 +358,7 @@ ignore = [
     "S607",  # Starting a process with a partial executable path
     "S105",  # Possible hardcoded password assigned to: "PASS"
     "SIM103",  # Return the condition directly - can hurt readability
+    "PERF203", # `try`-`except` within a loop incurs performance overhead
     "D200",  # One-line docstring should fit on one line
     "D212",  # Multi-line docstring summary should start at the first line
 
--- a/tests/unit/provision/testcloud/test_hw.py
+++ b/tests/unit/provision/testcloud/test_hw.py
@@ -26,11 +26,11 @@ from tmt.steps.provision.testcloud import (  # noqa: E402
 )
 
 if TPM_CONFIG_ALLOWS_VERSIONS:
-    allowed_combinations: list[tuple[str, Operator]] = []
-
-    for version in TPM_VERSION_SUPPORTED_VERSIONS[TPM_CONFIG_ALLOWS_VERSIONS]:
-        for op in virtual_TPM_VERSION_ALLOWED_OPERATORS:
-            allowed_combinations.append((version, op))
+    allowed_combinations: list[tuple[str, Operator]] = [
+        (version, op)
+        for version in TPM_VERSION_SUPPORTED_VERSIONS[TPM_CONFIG_ALLOWS_VERSIONS]
+        for op in virtual_TPM_VERSION_ALLOWED_OPERATORS
+    ]
 
     @pytest.mark.parametrize(
         ('version', 'op'),
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -60,8 +60,7 @@ def test_last_run_race(tmppath: Path, monkeypatch, root_logger: Logger):
             results.put(err)
 
     total = 20
-    for i in range(total):
-        threads.append(threading.Thread(target=create_last_run, args=(config, i)))
+    threads = [threading.Thread(target=create_last_run, args=(config, i)) for i in range(total)]
     for t in threads:
         t.start()
     for t in threads:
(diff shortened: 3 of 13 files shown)


tmt/export/nitrate.py
```diff
@@ -84,22 +84,17 @@ def _nitrate_find_fmf_testcases(test: 'tmt.Test') -> Iterator[Any]:
     """
     import tmt.base
     assert nitrate
-    for component in test.component:
-        try:
+    with suppress(tmt.utils.StructuredFieldError, nitrate.NitrateError):
```
martinhoyer, line 87: Does this change make sense @happz? I'm thinking  
```
try:
    with suppress(tmt.utils.StructuredFieldError):

except nitrate.NitrateError:
    pass
``` 
but not sure if there is an advantage in doing so.

lukaszachy:
@martinhoyer Huston, a little problem - https://stackoverflow.com/questions/34566806/why-use-contextlib-suppress-as-opposed-to-try-except-with-pass

I've checked same timeit command on python3.13 (my laptop) and it indeed is much worse with suppress. 
So I'd say different approach is necessary if the whole point is to get better performance :/

martinhoyer:
> @martinhoyer Huston, a little problem - https://stackoverflow.com/questions/34566806/why-use-contextlib-suppress-as-opposed-to-try-except-with-pass
> 
> I've checked same timeit command on python3.13 (my laptop) and it indeed is much worse with suppress. So I'd say different approach is necessary if the whole point is to get better performance :/

Nice, good info. Thanks!

Actually the supress wasn't added as part of performance optimization. The perflint is complaining about try except being inside the loop.  
We can do the "same" changes without the supress, by moving the loops within try-except.

btw, have you run he timeits with some of the changed code here? Just curious if it's also slower.

tmt/export/nitrate.py
martinhoyer: changed in https://github.com/teemtee/tmt/pull/3403/commits/64f624261f0d734330586a9420c0055ec1b99a16

martinhoyer:
Out of curiosity, I've tried to solve the remaining PERF203 and tested modifying this function:
```
    for i in range(attempts):
        try:
            return func(*args, **kwargs)
        except Exception as exc:
            exceptions.append(exc)
            logger.debug(
                'retry',
                f"{label} failed, {attempts - i} retries left, "
                f"trying again in {interval:.2f} seconds.")
            logger.fail(str(exc))
            time.sleep(interval)
    raise RetryError(label, causes=exceptions)
```
to
```
    retries_left = attempts

    while retries_left > 0:
        retries_left -= 1
        try:
            return func(*args, **kwargs)
        except Exception as exc:
            exceptions.append(exc)
            logger.debug(
                'retry',
                f"{label} failed, {retries_left} retries left, "
                f"trying again in {interval:.2f} seconds.")
            logger.fail(str(exc))

            if retries_left > 0:
                time.sleep(interval)

    raise RetryError(label, causes=exceptions)
```
An the test results over a milion iterations:
Average 1: 0.0000056165 seconds
Average 2: 0.0000056047 seconds

image

Anyway, I'm going to do something useful.

happz:
Rebase will be a PITA...

martinhoyer:
> Rebase will be a PITA...

Indeed, I've started from scratch on top of https://github.com/teemtee/tmt/pull/3662 Marking as draft until that one is merged.

martinhoyer:
@coderabbitai full review

martinhoyer:
@coderabbitai full review

tmt/steps/prepare/install.py
```diff
@@ -473,15 +473,21 @@ def install_local(self) -> None:
...
-                local_packages_installed.append(package)
+                return True
             except tmt.utils.RunError as error:
-                self.warn(f"Local package '{package}' not installed: {error.stderr}")
+                self.warn(f"Local package '{pkg}' not installed: {error.stderr}")
+                return False
+
+        local_packages_installed = [
+            package for package in self.local_packages if try_install_local_package(package)
+        ]
```
thrix, line 490: @martinhoyer this does not really avoid the try/except in the loop right? :)

pyproject.toml
```diff
@@ -357,6 +358,7 @@ ignore = [
     "S607",  # Starting a process with a partial executable path
     "S105",  # Possible hardcoded password assigned to: "PASS"
     "SIM103",  # Return the condition directly - can hurt readability
+    "PERF203", # `try`-`except` within a loop incurs performance overhead
```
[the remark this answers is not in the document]
thrix: yep, and the refactoring does not remove the try/execept from the loop anyway ...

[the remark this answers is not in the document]
martinhoyer: Yeah that was pretty stupid. Dropped that commit.

tmt/steps/prepare/install.py
martinhoyer: dropped, thanks

thrix (approved):
LGTM

therazix (approved):

happz:
@martinhoyer please, check out the failed pre-commit check. I tried, but I can't convince my `gh` to work with Github to even checkout your branch :/

PR #3404, opened 2024-12-04
Do not use `mypy` cache for `ruamel.yaml`

martinhoyer:
Trying to fix the annoying mypy fails every second run. Tested locally by running `pre-commit run --all-files` repeatably.

See https://github.com/python/mypy/issues/12664#issuecomment-2518393807

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,6 @@ repos:
           - "types-lxml"
 
         pass_filenames: false
-        args: [--config-file=pyproject.toml]
 
   - repo: https://github.com/RobertCraigie/pyright-python
     rev: v1.1.390
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -235,6 +235,11 @@ module = [
     ]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+# Workaround for https://github.com/python/mypy/issues/12664
+module = "ruamel.*"
+cache_dir = "/dev/null"
+
 [tool.pyright]
 include = [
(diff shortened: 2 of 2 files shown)


martinhoyer:
Things got a bit messy when another module override was added to the mypy config in pyproject.toml. Adding back the`mypy.ini` file. Any better ideas?

happz:
> Things got a bit messy when another module override was added to the mypy config in pyproject.toml. Adding back the`mypy.ini` file. Any better ideas?

How does `pyproject.toml` looks like with the change? I’m not a big fan of fracturing the configuration into smaller files, maybe the price wouldn’t be that high.

martinhoyer:
> > Things got a bit messy when another module override was added to the mypy config in pyproject.toml. Adding back the`mypy.ini` file. Any better ideas?
> 
> How does `pyproject.toml` looks like with the change? I’m not a big fan of fracturing the configuration into smaller files, maybe the price wouldn’t be that high.

I actually don't know how working pyproject.toml would look like, as I didn't manage to get it working. 
Based on [docs](https://mypy.readthedocs.io/en/stable/config_file.html#example-pyproject-toml), it should just be adding 
```
[[tool.mypy.overrides]]
module = "ruamel.*"
cache_dir = "/dev/null"
```
under the existing overrides section, but that breaks things.

martinhoyer:
> args: [--config-file=pyproject.toml]

This is redundant as mypy looks there if mypy.ini or .mypy.ini is not present. Was there any explicit reason for it in the first place?

psss (approved):
Nice! Thanks much for getting rid of this!

Issue #3407, opened 2024-12-05
Store all `contact` emails when uploading results in `reportportal`


psss:
Seems that just the first email address is stored in the test when uploading results to ReportPortal. It would be good to store all email addresses if the `contact` field is a list.

Related: https://issues.redhat.com/browse/TT-227

PR #3412, opened 2024-12-08
Add possibility to export multiple test contacts into the ReportPortal

seberm:
Add the possibility of exporting multiple test contacts into the report portal instance.

Closes: https://github.com/teemtee/tmt/issues/3407

TODO:
- [ ] Fix the local reportportal "contact" attribute test

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tests/report/reportportal/data/test.fmf
+++ b/tests/report/reportportal/data/test.fmf
@@ -1,11 +1,13 @@
 /bad:
     summary: Failing test
-    contact: tester@redhat.com
+    contact:
+      - tester1@redhat.com
+      - tester2@redhat.com
     test: echo "Something bad happened!"; false
 
 /good:
     summary: Passing test
-    contact: tester_2@redhat.com
+    contact: tester3@redhat.com
     test: echo "Everything's fine!"
     id: 63f26fb7-69c4-4781-a06e-098e2b58129f
 
--- a/tests/report/reportportal/test.sh
+++ b/tests/report/reportportal/test.sh
@@ -191,6 +191,14 @@ rlJournalStart
                     key="contact"
                     value="$(yq -r ".\"$test_name\".$key" test.fmf)"
                     if [[ $value != null ]]; then
+                        # TODO: Fix the test here! The output of yq is list of
+                        # contact in the case of multiple contacts (test==/bad)
+                        # and a string in a case when one contact is provided
+                        # for the test (test==/good). We must somehow handle
+                        # both situations.
+                        #
+                        # If yq returns list, we must check both contacts in the response JSON
+                        # If yq returns just a string value, we must check only one contact in the response JSON
                         rlAssertGrep "$key" tmp_attributes.json -A1 > tmp_attributes_selection
                         rlAssertGrep "$value" tmp_attributes_selection
                     else
@@ -200,6 +208,10 @@ rlJournalStart
                     key="TMT_TREE"
                     rlAssertNotGrep "$key" tmp_attributes.json
                 fi
+
+                # REMOVE ME: just for testing
+                #echo "EXITING OMG"
+                #exit 99
 
                 rm tmp_attributes*
             done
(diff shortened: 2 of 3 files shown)


happz (approved):

psss (approved):
Looks good, thanks!

psss:
A short release note would be nice.

seberm:
> A short release note would be nice.

@psss Added, thanks for the review!

psss:
Hm, `/tests/finish/prune` started to [fail](https://artifacts.dev.testing-farm.io/4c6daa21-8494-4852-a454-ba2cd603a1b1/) for no obvious reason with:

    :: [ 09:13:23 ] :: [   FAIL   ] :: File /tmp/tmp.5qqARmNv2o/plan/execute/data/guest/default-0/default-2/write/test-data-3/data/out-test.txt should exist 

This does not seem to be related to this change. When executed locally, everything it just fine :face_with_spiral_eyes:

lukaszachy:
@psss fails for me locally as well. I `make rpm` current content, installed (just tmt main package) into fedora:41 container and executed the `tests/finish/prune/test.sh`

It fails because that asserted file exists as `test-data-4/data/out-test.txt`, not `test-data-3`

happz:
> @psss fails for me locally as well. I `make rpm` current content, installed (just tmt main package) into fedora:41 container and executed the `tests/finish/prune/test.sh`
> 
> It fails because that asserted file exists as `test-data-4/data/out-test.txt`, not `test-data-3`
> 
> Fastest way to check is (failing version)
> 
> ```
> $ cd tests/finish/prune/data
> $ tmt run discover -vvv
>         summary: 4 tests selected
>             /default-0/test/test1
>             /default-1/test/test2
>             /default-2/write/plan-data
>             /default-2/write/test-data
> ```
> 
> Passing version has the last two tests in switched order.

The problem with the changed order of tests has been mentioned this evening on tmt Slack channel, this seems like the same issue. I suspect fmf library is a culprit here, but I didn't follow fmf development closely enough for me to point a finger at a commit (and no tmt change rings a bell). But it should be easy to check, the test would pass with tmt & an older fmf version.

> `shell `plugin takes the list of tests, constructs an fmf tree out of them, and populates internal `Test` instances from this tree - and if there was a change sorting of fmf nodes, items, keys, that might have affected this process

lukaszachy:
@happz @psss  Yes indeed, fmf-1.4.1  vs 1.5.0 return different order of tests.

happz:
> @happz @psss Yes indeed, fmf-1.4.1 vs 1.5.0 return different order of tests.

Which sounds like a breaking change in case of lists, as demonstrated by the `shell` disover plugin, where the order of tests was apparently given by their order in the list of `tests`.

psss:
So the problem is that previously the data key sorting happened when the `fmf.Tree` was being built and now we sort the keys when the `Tree.climb()` method is called.

https://github.com/teemtee/fmf/blob/95c9745db00b0e2b545a8e93021544c10d8665ac/fmf/base.py#L710-L716

The change does not affect trees which are built directly from `fmf` files on disk because those were always sorted by node name. However, when the `Tree.child()` method is used for creating child nodes manually, those nodes were kept in the original order as they were created. This is what the `shell` discover plugin uses:

https://github.com/teemtee/tmt/blob/01d38b80c43f522072845939beecf662acd4219a/tmt/steps/discover/shell.py#L389-L394

So these nodes are now sorted as well, which for this specific use case probably does not make sense: User provided the tests in the form of a list and they most probably expect the order should be preserved.

I see two possible ways to fix this:

* do not sort keys when climbing the Tree
* make the sorting optional / configurable

If we go the first way, it would break tests which rely on the sorting. And I guess this approach can be quite widely used. So, let's make it configurable? Or any other/better ideas?

lukaszachy:
> make the sorting optional / configurable

I'd say it should be configurable. Caller of climb method should know if they want to keep the order in which they created the data (eg. as this discover.shell) or sorted (majority of other use cases).

psss:
> I'd say it should be configurable. Caller of climb method should know if they want to keep the order in which they created the data (eg. as this discover.shell) or sorted (majority of other use cases).

Agreed. The `fmf` part should be covered here:

* https://github.com/teemtee/fmf/pull/271

Issue #3414, opened 2024-12-10
bootc plugin should delete qcow2 image when plan finished


henrywang:
The bootc plugin qcow2 image is built from `Containerfile` or bootc container image. So every tmt plan has its own qcow2 image to run its test. That needs a new qcow2 built by bib. The old one coming from previous plan should be removed.

There should be configurable bool option `delete-qcow2` in provision.

This issue has been worked around by https://github.com/teemtee/tmt/blob/94951abdf36e72da28bc984e79172ab6b47a9a3f/tests/provision/bootc/test.sh#L22, so it's not found in test.

For now in bootc test we use another work around for that https://github.com/containers/bootc/blob/3f5eb52a1712e1adf28f841cdf27211e7c170d57/xtask/src/xtask.rs#L208.

PR #3419, opened 2024-12-11
Use unique names for the `bootc` images

henrywang:
Plan with bootc provision builds its own qcow2 before test run. Each plan has different Containerfile or container image configured, so each plan should use its own qcow2 built from Contaienrfile or container image. The old or previous plan used qcow2 image should be removed when plan is done.

Resolve: https://github.com/teemtee/tmt/issues/3414

### DIFF
--- a/tmt/steps/provision/bootc.py
+++ b/tmt/steps/provision/bootc.py
@@ -11,7 +11,7 @@ import tmt.steps.provision
 import tmt.steps.provision.testcloud
 import tmt.utils
 from tmt.steps.provision.testcloud import GuestTestcloud
-from tmt.utils import field
+from tmt.utils import Path, field
 from tmt.utils.templates import render_template
 
 if TYPE_CHECKING:
@@ -329,7 +329,6 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
         self._check_if_podman_is_rootless()
 
         data = BootcData.from_plugin(self)
-        data.image = f"file://{self.workdir}/qcow2/disk.qcow2"
         data.show(verbose=self.verbosity_level, logger=self._logger)
 
         if self._rootless:
@@ -354,6 +353,24 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
         else:
             raise tmt.utils.ProvisionError(
                 "Either 'container-file' or 'container-image' must be specified.")
+
+        # Set unique disk file name, each plan will have its own disk file
+        disk_file_name = Path(render_template(
+            'disk-{{ PHASE.parent.plan.my_run.unique_id }}'
+            '-{{ PHASE.parent.plan.pathless_safe_name }}'
+            '-{{ PHASE.safe_name }}.qcow2',
+            PHASE=self))
+
+        assert self.workdir is not None
+
+        image_dir = self.workdir / 'qcow2'
+
+        # Rename disk file name to unique file name
+        built_image = image_dir / 'disk.qcow2'
+        renamed_image = image_dir / disk_file_name
+
+        built_image.rename(renamed_image)
+        data.image = f"file://{renamed_image}"
 
         self._guest = GuestBootc(
             logger=self._logger,
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -4,8 +4,6 @@
 IMAGE_NEEDS_DEPS="localhost/tmt-bootc-needs-deps"
 IMAGE_INCLUDES_DEPS="localhost/tmt-bootc-includes-deps"
 
-TESTCLOUD_IMAGE="/var/tmp/tmt/testcloud/images/disk.qcow2"
-
 
 rlJournalStart
     rlPhaseStartSetup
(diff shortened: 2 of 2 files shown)


tmt/steps/provision/bootc.py
```diff
@@ -75,6 +75,12 @@ def remove(self) -> None:
             logger=self._logger,
             env=PODMAN_ENV if self._rootless else None)
 
+        # plan with bootc provision should build its own qcow2 before test run
+        # test qcow2 image should be removed when plan is done
+        tmt.utils.Command(
+            "rm", "-vf", f"{DEFAULT_TMP_PATH}/images/disk.qcow2"
```
happz, line 81: 1. `unlink` should work as well: `(DEFAULT_TMP_PATH / 'images/disk.qcow2').unlink(missing_ok=True)`
2. but what caught my eye: is this a single path shared by all plans and all `bootc` plugin invocations? That seems like a huge risk of conflict, being a shared, mutable state without proper synchronization. Can the name be changed, to reflect which run, plan, and guest it belongs to? Because this is not good, if all runs and plans all compete over the same file, we are just asking for trouble. Either its name or path should be different, perhaps the file should be moved under a workdir, something like this.

henrywang: Thanks for the comment.

The bib build `disk.qocw2` and save into `{self.workdir}/qcow2/disk.qcow2`. The `workdir` is different for each plan.

The `disk.qcow2` should be copied to DEFAULT_TMP_PATH}/images by `virtual.testcloud` when testcloud run VM. Is this correct?

What happens when defined qcow2 path https://github.com/teemtee/tmt/blob/5dc0a1f4441a7a3727fc36160302355d7d737363/tmt/steps/provision/bootc.py#L318C31-L318C62 for each plan and run it https://github.com/teemtee/tmt/blob/5dc0a1f4441a7a3727fc36160302355d7d737363/tmt/steps/provision/bootc.py#L344?

The name `disk.qcow2` is hardcoded by bib.

happz: `data.image = f"file://{self.workdir}/qcow2/disk.qcow2"` looks good, that's what I would expect to exist.

> The disk.qcow2 should be copied to DEFAULT_TMP_PATH}/images by virtual.testcloud when testcloud run VM. Is this correct?

I see. No, I think this spells trouble. I don't mean now, `bootc` plugin is basically a tech preview, but once more people start playing with it and someone tries more than one `provision` phase in their plan, multiple `bootc`, it may easily become a problem as `virtual` would use the same qcow2 file for all of them. Which may not necessarily be the right choice...

I think it's something left to resolve, sooner rather than later, and we might need help from testcloud owners.

henrywang: Thanks @happz! I'm not tmt expert, so anyone can help on this?

thrix: Discussed here:
https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1736243756312569

happz: As discussed: bootc plugins should rename the file before using it. IIUIC, the filename as produced by the builder will remain `disk.qcow2`, but then the image is used and copied by testcloud. Plugin should rename the file before using it, to prevent testcloud - or anything else - from heading into a conflict.

Therefore, would it be possible to rename the image file to something else before doing `data.image = f"file://{self.workdir}/qcow2/disk.qcow2"`? The filename should be unique enough to not conflict between runs, plans and guests. Check out `tmt.steps.safe_filename()` which adds phase and guest name, with a custom template we could get safe name, e.g. `safe_filename('disk.qcow2-{{ self.parent.plan.my_run.unique_id }}-{{ self.parent.plan.pathless_safe_name }}', self.parent, self)`.

tmt/steps/provision/bootc.py
```diff
@@ -355,6 +354,20 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
+        disk_file_name = safe_filename(
+            'disk.qcow2-{{ self.parent.plan.my_run.unique_id }}-{{ self.parent.plan.pathless_safe_name }}',
+            self.parent,
+            self)
+
+        # Rename disk file name to unique file name
+        os.rename(
+            f"file://{self.workdir}/qcow2/disk.qcow2",
+            f"file://{self.workdir}/qcow2/{disk_file_name}"
+            )
```
happz, line 367: We try to use `pathlib` as much as possible when it comes to interacting with filesystems, so I would propose the following change for this part. It should also drop `file://` from paths (it's just for the `image` assignment) & the repetition of various strings is asking for a typo:

```python
image_dir = self.workdir / 'qcow2'

# Rename disk file name to unique file name
built_image = image_dir / 'disk.qcow2'
renamed_image = image_dir / disk_file_name

built_image.rename(renamed_image)
```

What do you think?

henrywang: That's better. Thanks.

tmt/steps/provision/bootc.py
```diff
@@ -355,6 +355,23 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
             raise tmt.utils.ProvisionError(
                 "Either 'container-file' or 'container-image' must be specified.")
 
+        # Set unique disk file name, each plan will have its own disk file
+        template = 'disk.qcow2-{{ self.parent.plan.my_run.unique_id }}-' \
+                   '{{ self.parent.plan.pathless_safe_name }}'
+        disk_file_name = safe_filename(
+            template,
+            self.parent,
+            self)
```
psss, line 364: The third parameter should be `guest`, but we don't have it yet, right? @happz, any thoughts? I tried to run this but ended with:

    'jinja2.runtime.TemplateReference object' has no attribute 'parent'

happz: Right, I was too quick. So how about this?

```diff
diff --git a/tmt/steps/provision/bootc.py b/tmt/steps/provision/bootc.py
index 16686730..af864dc5 100644
--- a/tmt/steps/provision/bootc.py
+++ b/tmt/steps/provision/bootc.py
@@ -10,9 +10,8 @@ import tmt.steps
 import tmt.steps.provision
 import tmt.steps.provision.testcloud
 import tmt.utils
-from tmt.steps import safe_filename
 from tmt.steps.provision.testcloud import GuestTestcloud
-from tmt.utils import field
+from tmt.utils import Path, field
 from tmt.utils.templates import render_template
 
 if TYPE_CHECKING:
@@ -356,13 +355,13 @@ class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
                 "Either 'container-file' or 'container-image' must be specified.")
 
         # Set unique disk file name, each plan will have its own disk file
-        template = 'disk.qcow2-{{ self.parent.plan.my_run.unique_id }}-' \
-                   '{{ self.parent.plan.pathless_safe_name }}'
-        disk_file_name = safe_filename(
-            template,
-            self.parent,
-            self)
+        disk_file_name = Path(render_template(
+            'disk-{{ PHASE.parent.plan.my_run.unique_id }}'
+            '-{{ PHASE.parent.plan.pathless_safe_name }}'
+            '-{{ PHASE.safe_name }}.qcow2',
+            PHASE=self))
 
+        assert self.workdir is not None
         image_dir = self.workdir / 'qcow2'
 
         # Rename disk file name to unique file name
```

psss: Is there any special reason for using `render_template()` here instead of a pure f-string?

happz: Mostly a copy & paste from the `safe_filename()` code, here and now. Although I didn't expect it to be replaced with an f-string even when moved into the improved `safe_filename()`. It's a template, and using `render_template()` was a natural choice for me, as it's easy to prepare from multiple components, and then evaluate.

psss (approved):
Now looks good and works nice. Thanks!

happz (approved):

PR #3420, opened 2024-12-11
Timestamp schema pattern should allow Z instead of zero offset

happz:
`2024-11-29T12:32:11.444600549+01:00` is accepted, `2024-11-29T11:32:11.444600549Z` is not. But it should, see [1], it's a valid format. tmt ingests the values, but does not parse them, just passes them around, so changing the pattern should be enough to accept this format.

[1] https://github.com/fedora-java/javapackages-validator/pull/176

Pull Request Checklist

* [x] implement the feature


psss (approved):

psss:
Trivial change, full test passed before rebase, merging.

PR #3423, opened 2024-12-12
Define a new context dimension for image mode

psss:
Introduce a new context dimension key `deployment-mode` in order to allow enabling/disabling/adjusting tests and plans based on the distribution mode.

Pull Request Checklist

* [x] include a release note

### DIFF
--- a/spec/context/dimension.fmf
+++ b/spec/context/dimension.fmf
@@ -14,14 +14,10 @@ description: |
         runs (fedora, fedora-33, centos, centos-8, centos-8.4,
         centos-stream, centos-stream-9, rhel, rhel-8, rhel-8.4).
 
-    distro-mode
-        Distribution mode, ``package`` for the standard way using
-        packages or ``image`` for the new way of distributing
-        using container images.
-
-    image-mode
-        Flag denoting how new content is distributed. Use ``true``
-        for container images and ``false`` for standard packages.
+    deployment-mode
+        Deployment mode of the distribution, ``package`` for the
+        standard way using packages or ``image`` for the new way
+        of distributing using container images.
 
     variant
         The distribution variant (Client, Desktop, Server
--- a/spec/context/dimension.fmf
+++ b/spec/context/dimension.fmf
@@ -14,6 +14,15 @@ description: |
         runs (fedora, fedora-33, centos, centos-8, centos-8.4,
         centos-stream, centos-stream-9, rhel, rhel-8, rhel-8.4).
 
+    distro-mode
+        Distribution mode, ``package`` for the standard way using
+        packages or ``image`` for the new way of distributing
+        using container images.
+
+    image-mode
+        Flag denoting how new content is distributed. Use ``true``
+        for container images and ``false`` for standard packages.
+
     variant
         The distribution variant (Client, Desktop, Server
         Workstation, Silverblue, CoreOS).
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -11,6 +11,11 @@ tmt-1.42.0
 The ``tmt show`` command now prints in verbose mode manual test
 instructions as well.
 
+A new context :ref:`/spec/context/dimension` ``deployment-mode``
+has been added to the specification. It can be used to
+:ref:`/spec/core/adjust` test and plan metadata for the
+``package`` or ``image`` mode context.
+
 
 tmt-1.41.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


psss:
Two versions were proposed/discussed on the stakeholder session let's review both and vote/select the final candidate here.

spec/context/dimension.fmf
```diff
@@ -14,6 +14,15 @@ description: |
         runs (fedora, fedora-33, centos, centos-8, centos-8.4,
         centos-stream, centos-stream-9, rhel, rhel-8, rhel-8.4).
 
+    distro-mode
+        Distribution mode, ``package`` for the standard way using
+        packages or ``image`` for the new way of distributing
+        using container images.
```
psss, line 20: Vote here for `distro-mode`.

spec/context/dimension.fmf
```diff
@@ -14,6 +14,15 @@ description: |
...
+
+    image-mode
+        Flag denoting how new content is distributed. Use ``true``
+        for container images and ``false`` for standard packages.
```
psss, line 24: Vote here for `image-mode`.

spec/context/dimension.fmf
phracek: Thanks for bringing it this. Do we really need to have two variables?

In my understanding distro-mode means installing packages in standard way, like RPM.
What about if user specifies:
```
distro-mode: true
image-mode: true
```
or opposite way:
```
distro-mode: false
image-mode: false
```

I would prefer to have only one variable like `image-mode` that would be mean either `image` or `package`.

happz: I believe that's the case, the question - and voting - is about "which one of these two dimensions?".

comps:
What about using an OS installation style instead?

If I understood it correctly, "image mode" is basically just rpm-ostree ( https://docs.fedoraproject.org/en-US/fedora-silverblue/technical-information/ ) built in some special way. So for tests (or other software on the system), the difference is not "distro mode" vs "image mode", but the visible filesystem structure the test sees, ie. "FHS / standard" vs "rpm-ostree".

The product name for that ("Atomic", "Silverblue", "image mode", .. ??) might be less relevant.

psss:
@kkaarreell, as you raised this topic on the stakeholder session, do you have any specific use cases on mind for e.g. enabling / disable tests based on this new dimension? Are they possibly related to the filesystem structure the test sees as @comps mentioned above?

lukaszachy:
Can the dimension name be `mode` (or similar) with possible values `image|package` ?

kkaarreell:
> Can the dimension name be `mode` (or similar) with possible values `image|package` ?

Please don't. `mode` is so vague and overloaded term and maybe some users are already using it (`mode=ro|rw|...`). Please, make it at least a bit more specific and less prone to collisions. RH docs speaks about it as a deployment method. While `deployment_method=package_mode|image_mode` is long, it would be very much handy when inspecting logs or RP launch attributes, trying to identify how a test environment look like.

spec/context/dimension.fmf
thrix: @phracek the proposal is `image-mode: true/false` **or** `distro-mode: package/image`.

lukaszachy:
> > Can the dimension name be `mode` (or similar) with possible values `image|package` ?
> 
> Please don't. `mode` is so vague and overloaded term and maybe some users are already using it (`mode=ro|rw|...`). Please, make it at least a bit more specific and less prone to collisions. RH docs speaks about it as a deployment method. While `deployment_method=package_mode|image_mode` is long, it would be very much handy when inspecting logs or RP launch attributes, trying to identify how a test environment look like.

psss:
> as long as we don't have boolean `*-mode` representing each possibility.

Yeah, I share this point with @lukaszachy. But, hopefully, there won't be that many deployment modes? Or? :)

lukaszachy:
> > as long as we don't have boolean `*-mode` representing each possibility.
> 
> Yeah, I share this point with @lukaszachy. But, hopefully, there won't be that many deployment modes? Or? :)

More painful is that one would need to set at least two dimensions each time.

kkaarreell:
Just FYI, in Polarion this is going to be tracked in a field called "deploymentmode".

psss:
> Just FYI, in Polarion this is going to be tracked in a field called "deploymentmode".

Thanks for the info. It would make sense to keep things consistent.

psss:
To sum up: There were strong objections raised on the hacking session against the boolean approach as it would be impractical if there was another value introduced in the future. Thus we are left with the values. Duplicating `mode` in the values does not make much sense, thus it should go to the dimension name. We have brainstormed `distro-mode` and now `deploymentmode` has popped up. I'd say let's make it close to what was already proposed for Polarion and go with `deployment-mode` to keep our multiword naming consistent. Any objections?

thrix (approved):
LGTM. no objections for the final solution

psss:
Summary from the hacking session: Let's keep the two values `image` and `package` for now. We can add `container` later if there is a clear use case and definition for it.

PR #3424, opened 2024-12-13
Allow dynamic plugin data classes & convert feature plugins

happz:
Plugins' `_data_class` is no longer the source of plugins' data classes. Instead, new `get_data_class()` method is added (see [1]). `_data_class` is not gone, it serves as the default for `get_data_class()`, but plugins now can provide their own implementation.

Which is exactly what `prepare/feature` and feature plugins do now, the plugin builds its data class from collected smaller data classes, one for each feature plugin.

There are some loose ends, namely type annorations, but the code works, both in runtime and when rendering docs.

[1] a `data_class` property would be nice, but the attribute must be a class-level attribute and those cannot be properties.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/docs/scripts/generate-plugins.py
+++ b/docs/scripts/generate-plugins.py
@@ -120,7 +120,11 @@ def _create_step_plugin_iterator(registry: tmt.plugins.PluginRegistry[tmt.steps.
         for plugin_id in registry.iter_plugin_ids():
             plugin = registry.get_plugin(plugin_id).class_
 
-            yield plugin_id, plugin, plugin._data_class
+            if hasattr(plugin, 'get_data_class'):
+                yield plugin_id, plugin, plugin.get_data_class()
+
+            else:
+                yield plugin_id, plugin, plugin._data_class
 
     return plugin_iterator
 
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -1027,7 +1027,8 @@ class Step(tmt.utils.MultiInvokableCommon, tmt.export.Exportable['Step']):
                 else:
                     debug3('incompatible step data')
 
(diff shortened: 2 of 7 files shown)


falconizmi:
Tested locally with WIP feature plugin buildroot https://gitlab.cee.redhat.com/baseos-qe/tmt/-/merge_requests/106 and it successfully detects it (verified in output of `TMT_PLUGINS=./tmt_redhat/buildroot/ tmt run prepare --how feature --help` ).
But it couldn't run the playbook of the feature plugin, this PR https://github.com/teemtee/tmt/pull/3457 has temporarily workaround for it.

happz:
@psss https://artifacts.osci.redhat.com/testing-farm/1f44fca6-1bba-4f99-b60e-ec6444f2a4ac/work-bootcjx6em3nm/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/output.txt looks like a "bug" in `bootc` plugin, could it be calling `dnf` while images changed and `dnf` has been replaced with `dnf5`?

@henrywang @ckyrouac ^^ any ideas?

psss:
> @psss https://artifacts.osci.redhat.com/testing-farm/1f44fca6-1bba-4f99-b60e-ec6444f2a4ac/work-bootcjx6em3nm/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/output.txt looks like a "bug" in `bootc` plugin, could it be calling `dnf` while images changed and `dnf` has been replaced with `dnf5`?

Yeah, looks like that. But I just tested locally with fresh images and everything just works nice. Which is weird.

psss:
Anyway, this is definitely no related to this change. Merging.

psss (approved):

PR #2850, opened 2024-04-12
Support `--workdir-root` in the `tmt clean` command

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [x] extend the test coverage
* [x] include a release note

PR #2838, opened 2024-04-09
Accept and apply `workdir-root` in `testcloud`

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

PR #3426, opened 2024-12-15
Support `--workdir-root` in the `tmt clean images` command

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/tmt/cli/_root.py
+++ b/tmt/cli/_root.py
@@ -1853,7 +1853,7 @@ def clean_guests(
 @workdir_root_options
 @verbosity_options
 @dry_options
-def clean_images(context: Context, _workdir_root: Optional[str], **kwargs: Any) -> None:
+def clean_images(context: Context, workdir_root: Optional[Path], **kwargs: Any) -> None:
     """
     Remove images of supported provision methods.
 
@@ -1866,7 +1866,6 @@ def clean_images(context: Context, _workdir_root: Optional[str], **kwargs: Any) 
     #        cleaned, similarly to guests.
     assert context.obj.clean_logger is not None  # narrow type
 
-    workdir_root = Path(_workdir_root) if _workdir_root is not None else None
     if workdir_root and not workdir_root.exists():
         raise tmt.utils.GeneralError(f"Path '{workdir_root}' doesn't exist.")
 
--- a/tmt/cli/_root.py
+++ b/tmt/cli/_root.py
@@ -1644,8 +1644,6 @@ def clean(
 
     if last and id_:
         raise tmt.utils.GeneralError("Options --last and --id cannot be used together.")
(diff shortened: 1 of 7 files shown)


skycastlelily:
Block #2838 and the failing tests seem unrelated:)

tmt/cli/_root.py
```diff
@@ -1434,12 +1435,17 @@ def clean_images(context: Context, **kwargs: Any) -> None:
     #        cleaned, similarly to guests.
     assert context.obj.clean_logger is not None  # narrow type
 
+    workdir_root = Path(_workdir_root) if _workdir_root is not None else None
```
thrix, line 1438: @skycastlelily I am wondering if 

```suggestion
    workdir_root = Path(_workdir_root) if _workdir_root else None
```

would not be better. Feels like an empty string should also be evaluated to not set? WDYT?

skycastlelily: Ah, that line is from  Milos,yeah,I copied his suggestion from #2850  :sweat_smile: 
`workdir_root = Path(_workdir_root) if _workdir_root is not None else None
`
will  make tmt take the current dir ie, '.'  as workdir_root,
while, `workdir_root = Path(_workdir_root) if _workdir_root else None`
will make tmt take Path(os.environ['TMT_WORKDIR_ROOT']) or WORKDIR_ROOT .
And, with "workdir_root = Path(_workdir_root) if _workdir_root is not None else None"
users would run into an error: Run workdir 'run-001' must not be inside fmf root '/home/lnie/tmt'. if they run 
`tmt run --workdir-root ''
`
Which is the supposed behavior? @thrix  @happz 
Or, maybe, we should raise an error when users pass an empty string?:)

thrix: I see, yeah, rather then error maybe?

happz: I believe this will become a moot point after https://github.com/teemtee/tmt/pull/3489

thrix: yep, true, resolving

thrix:
@skycastlelily so this did not get enough review, but according to comments it would be good to cover this with tests, as seems it is not covered

optak (approved):
LGGTM

skycastlelily:
> @skycastlelily so this did not get enough review, but according to comments it would be good to cover this with tests, as seems it is not covered

Sorry for the late reply, I was on two-weeks-long vacation,and sure,updated https://github.com/teemtee/tmt/pull/3426/commits/885071446b8211df8317877b9f28c80ef4153f33,and the failure seems unrelated:)

happz (approved):

PR #3430, opened 2024-12-18
Use shorter package names for `latest` copr builds

psss:
To keep the naming consistent with `fmf` packages. There's no need to track miliseconds, the `devX` suffix bump is enough to keep the latest package to appear as newest.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -86,6 +86,7 @@ jobs:
     trigger: commit
     branch: main
     project: latest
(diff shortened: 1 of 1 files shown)


psss:
For reference, this is the difference:

    tmt-1.40.0.dev23+ge12ded3f-1.20241209130719980946.main.23.ge12ded3f
    fmf-1.5.1.dev2+g95c9745-main

happz (approved):

therazix (approved):

PR #3434, opened 2024-12-22
Document predefined order values of special phases

happz:
I threw in a bit of renaming, because some constants were not about plugins only, some were not defaults, i.e. it was not possible to change them, and so on.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation


therazix (approved):

seberm (approved):

spec/core/order.fmf
```diff
@@ -14,3 +17,26 @@ example: |
...
+        60
+            Application of patches by ``rpm-build`` command invoked by
+            :ref:`discover/fmf</plugins/discover/fmf>` plugin when
+            ``dist-git-source`` key is used.
+
+        70
+            Installation of packages :ref:`required</spec/tests/require>` by tests.
+
+        75
+            Installation of packages :ref:`recommended</spec/tests/recommend>` by tests.
```
psss, line 42: All these values, except for default `50`, are related to the `prepare` step only. Would it make sense to move them closer to the `prepare` steps docs? And link them from here?

tmt/steps/__init__.py
```diff
@@ -68,8 +68,25 @@
...
+# prefix, so it's easy to find all `PHASE_ORDER_*` constants for documentation.
+#
+
+#: The default order of any object.
+# TODO: this is a duplication of tmt.base.DEFAULT_ORDER. Unfortunately, tmt.base
+# imports tmt.steps, not the other way around.
+# `PHASE_ORDER_DEFAULT = tmt.base.DEFAULT_ORDER` would be way better.
+PHASE_ORDER_DEFAULT = 50
+#: Installation of essential plugin and check requirements.
+PHASE_ORDER_INSTALL_ESSENTIAL_REQUIRES = 30
```
psss, line 85:
```suggestion
PREPARE_PHASE_ORDER_INSTALL_ESSENTIAL_REQUIRES = 30
```

These are not generic for all phases, should we be more specific?

spec/core/order.fmf
happz: I tried, but given how "plan" specification is composed from many stories, I was unable to position this piece correctly under the `prepare` section. Even with `order` it migrated to the beginning or the end of the whole page. I can make it a "plans" topic, but I'm not able to make it a "prepare" topic.

tmt/steps/__init__.py
psss: Thanks!

spec/core/order.fmf
psss: What about including it in the `prepare` step description? They are already partially covered there:

https://github.com/teemtee/tmt/blob/686d3bc0161e70507124efb152e9b33a46e8a7a4/spec/plans/prepare.fmf#L15-L22

happz: Moved in 8fc3b73c

tmt/steps/prepare/distgit.py
```diff
@@ -19,6 +19,11 @@
     import tmt.steps.discover
     import tmt.steps.prepare.install
 
+#: Application of patches by ``rpm-build`` command invoked by
+#: :ref:`discover/fmf</plugins/discover/fmf>` plugin when
+#: ``dist-git-source`` key is used.
+PHASE_ORDER_RPMBUILD_BP = 30
```
psss, line 25: The default was previously `60` (see below). Why the value has been updated to `30`?

happz: A copy & paste issue, it's correct in docs. Fixed in d4f6cbab.

psss: Good!

happz:
> @happz, what about defining the dist git constant with the rest at one place? Outlined in [0b2875c](https://github.com/teemtee/tmt/commit/0b2875c1d65e863c52b6f92e24e199d54a659b40).

Now that you mentioned it: I would even remove the dist-git one from the list in `prepare.fmf`, I would make sure it's documented in the plugin docstring, and instead of listing it here, I'd add a note to `prepare.fmf`, "hey, be aware that plugins may also add phases, with predefined `order` values of their own, if you're curious, be sure to check docs (link to /plugins) of plugins you use in your plans". Because there will be - there already are - plugins that would add their custom phases, and some of those plugins are even outside of tmt repo, so there is no chance in hell we would keep this list up-to-date and cover all **plugin-specific** `order` values. I would therefore give up, not even try it, and point users to the right source.

(The `errata` plugin defines itself with the default `order` of 60...)

spec/core/order.fmf
thrix: @psss @happz is this resolved?

thrix:
> > @happz, what about defining the dist git constant with the rest at one place? Outlined in [0b2875c](https://github.com/teemtee/tmt/commit/0b2875c1d65e863c52b6f92e24e199d54a659b40).
> 
> Now that you mentioned it: I would even remove the dist-git one from the list in `prepare.fmf`, I would make sure it's documented in the plugin docstring, and instead of listing it here, I'd add a note to `prepare.fmf`, "hey, be aware that plugins may also add phases, with predefined `order` values of their own, if you're curious, be sure to check docs (link to /plugins) of plugins you use in your plans". Because there will be - there already are - plugins that would add their custom phases, and some of those plugins are even outside of tmt repo, so there is no chance in hell we would keep this list up-to-date and cover all **plugin-specific** `order` values. I would therefore give up, not even try it, and point users to the right source.
> 
> (The `errata` plugin defines itself with the default `order` of 60...)

@psss  PTAL pls!

thrix (approved):
LGTM

spec/core/order.fmf
psss: Yes, this has been moved.

psss:
> Now that you mentioned it: I would even remove the dist-git one from the list in `prepare.fmf`, I would make sure it's documented in the plugin docstring, and instead of listing it here, I'd add a note to `prepare.fmf`, "hey, be aware that plugins may also add phases, with predefined `order` values of their own, if you're curious, be sure to check docs (link to /plugins) of plugins you use in your plans".

Yeap, agreed.

KwisatzHaderach (approved):

happz:
Hopefully all addressed in c1ba183a

tmt/steps/__init__.py
```diff
@@ -68,8 +68,25 @@
 
 
 DEFAULT_ALLOWED_HOW_PATTERN: Pattern[str] = re.compile(r'.*')
-DEFAULT_PLUGIN_METHOD_ORDER: int = 50
 
+#
+# Following are default and predefined order values for various special phases
+# recognized by tmt. When adding new special phase, add its order below, and
+# do not forget to update the list in /spec/core/order. Please, keep the name
```
psss, line 75: This `/spec/core/order` part is outdated and should be dropped from the comment.

tmt/steps/__init__.py
```diff
@@ -68,8 +68,25 @@
...
+# prefix, so it's easy to find all `PHASE_ORDER_*` constants for documentation.
+#
+
+#: The default order of any object.
+# TODO: this is a duplication of tmt.base.DEFAULT_ORDER. Unfortunately, tmt.base
+# imports tmt.steps, not the other way around.
```
psss, line 81: What about dropping `DEFAULT_ORDER = 50` from `base.py` instead?

tmt/steps/discover/shell.py
```diff
@@ -60,7 +60,7 @@ class TestDescription(
     description: Optional[str] = None
     enabled: bool = True
     order: int = field(
-        # TODO: ugly circular dependency (see tmt.base.DEFAULT_ORDER)
+        # TODO: ugly circular dependency (see tmt.steps.PHASE_ORDER_DEFAULT)
```
psss, line 63: Perhaps this one would be resolved as well if it's defined under `tmt.steps`?

psss (changes requested):
Looks good, just one minor outdated comment and perhaps we could get rid of the circular dependency problem by removing the default order from `base.py`?

tmt/steps/__init__.py
happz: It feels correctly placed there, it's the value of `order` after all, which applies to all objects, not just to steps.

tmt/steps/discover/shell.py
happz: Nice, fixed in 0fb39379

tmt/steps/__init__.py
happz: Nope, fixed in 8e3586a6

tmt/steps/discover/shell.py
happz: Nope, fixed in 8e3586a6

tmt/steps/__init__.py
psss: Ah, right, it's not just steps.

psss (approved):

PR #3435, opened 2025-01-02
Add CentOS Stream 10 among test images

happz:
Pull Request Checklist

* [x] implement the feature


seberm (approved):

containers/centos/stream10/Containerfile
```diff
@@ -0,0 +1,10 @@
+#
+# A CentOS Stream 10 image tailored for tmt test suite
+#
+# tmt/tests/centos/stream10/upstream:latest
+#
+
+FROM quay.io/centos/centos:stream10
+
+    # Populate dnf cache
```
thrix, line 9: Weird indent

happz: Expected, aligns with the command, not `RUN`.

thrix: ok, we have different state, but I can live with it :)

thrix (approved):
LGTM

tests/unit/test_package_managers.py
```diff
@@ -148,6 +150,8 @@ def assert_output(
     (CONTAINER_FEDORA_39, PACKAGE_MANAGER_YUM),
 
     # CentOS Stream
+    (CONTAINER_CENTOS_STREAM_10, PACKAGE_MANAGER_DNF),
+    (CONTAINER_CENTOS_STREAM_10, PACKAGE_MANAGER_YUM),
```
psss, line 154: Do we really need to exercise `yum` for `centos-stream-10`?

psss (approved):
Just one question, otherwise looks good.

tests/unit/test_package_managers.py
happz: Good point, dropped in 187ef946

psss:
Tests were failing because `epel-10` was not enabled in the testing copr repo. Should be fixed by 19947572.

psss:
Heh, now it fails because of `fedora-39`:

    Error: It wasn't possible to enable this project.
    Repository 'fedora-39-x86_64' does not exist in project '@teemtee/stable'.

But that one is EOL, so I guess, that should be dropped?

happz:
> Heh, now it fails because of `fedora-39`:
> 
> ```
> Error: It wasn't possible to enable this project.
> Repository 'fedora-39-x86_64' does not exist in project '@teemtee/stable'.
> ```
> 
> But that one is EOL, so I guess, that should be dropped?

I'm not sure about dropping the whole Fedora 39, that would require more changes, so for now I just disabled this Copr test in eb398abf.

Issue #3442, opened 2025-01-04
Provision beaker `hostname` doesn't work


guazhangRH:
```
tmt -c compose=RHEL-10.0-20241220.0  run --all -dvv   provision -h beaker --hardware hostname=storageqe-100.rhts.eng.pek2.redhat.com --image RHEL-10.0-20241220.0    plan --name plan/reserve
```

```
 tmt -c compose=RHEL-10.0-20241220.0  run --all -dvv   provision -h beaker --hardware hostname=storageqe-100.rhts.eng.pek2.redhat.com --image RHEL-10.0-20241220.0    plan --name plan/reserve 
Using tree '/home/guazhang/git/tmt_tests'.
/var/tmp/tmt/run-004
tmt version: 1.40.0
tmt command line: /usr/bin/tmt -c compose=RHEL-10.0-20241220.0 run --all -dvv provision -h beaker --hardware hostname=storageqe-100.rhts.eng.pek2.redhat.com --image RHEL-10.0-20241220.0 report --insert --how html plan --name plan/reserve
Run data not found.
Enabled steps: prepare, execute, finish, report, discover and provision
Workdir '/var/tmp/tmt/run-004/plan/reserve' created.
Found 1 plan.

/plan/reserve
summary: kernel storage test plan reserve server
    Workdir '/var/tmp/tmt/run-004/plan/reserve/discover' created.
    Workdir '/var/tmp/tmt/run-004/plan/reserve/provision' created.
    Workdir '/var/tmp/tmt/run-004/plan/reserve/prepare' created.
    Workdir '/var/tmp/tmt/run-004/plan/reserve/execute' created.
    Workdir '/var/tmp/tmt/run-004/plan/reserve/report' created.
    Report step always force mode enabled.
    Workdir '/var/tmp/tmt/run-004/plan/reserve/report' created.
    Workdir '/var/tmp/tmt/run-004/plan/reserve/finish' created.
    discover
    workdir: /var/tmp/tmt/run-004/plan/reserve/discover
        how: fmf
        Workdir '/var/tmp/tmt/run-004/plan/reserve/discover/default-0' created.
        directory: /home/guazhang/git/tmt_tests
        Copy '/home/guazhang/git/tmt_tests' to '/var/tmp/tmt/run-004/plan/reserve/discover/default-0/tests'.
        hash: 45ce571
        tests: /tests/reboot
[...]

PR #3443, opened 2025-01-04
Workaround an apparent problem with Mrack conversion of HW requirements

happz:
It seems that when only an actual filtering element is the filter, mrack renders an element without attributes. It might be a mrack bug, or us not understanding mrack API, but a quick workaround to unblock tmt is possible. If it's a mrack bug, the proper fix may come later; if it's us not giving mrack the correct structure, well, this patch fixes that. We should be covered in both directions.

Fixes https://github.com/teemtee/tmt/issues/3442

Pull Request Checklist

* [x] implement the feature


skycastlelily (approved):
LGTM^^

skycastlelily:
And,here is the mrack fix:https://github.com/neoave/mrack/pull/305^^

psss (approved):
Looks good, thanks!

PR #3450, opened 2025-01-07
Extend `yamllint` coverage to prepare/feature playbooks

happz:
These playbooks are not examples or test-only playbooks, but user-facing playbooks tmt and other tools may use in real workflows. They are covered by ansible-lint, let's add also yamllint to the mix because Ansible playbooks should be valid YAML files, and yamllint takes a slightly different view and reports slightly different set of issues. Whatever passes ansible-lint should be perfectly valid YAML file, but not necessarily nicely formatted, presentable, maintainable, readable, and/or consistent with the rest of playbooks shipped with tmt. yamllint should improve this side of the picture.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/ansible/packages.yml
+++ b/ansible/packages.yml
@@ -1,3 +1,5 @@
+---
+
 - name: Install packages
   hosts: all
   tasks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -125,7 +125,7 @@ repos:
     rev: v1.35.1
     hooks:
       - id: yamllint
-        files: ^tmt/schemas/.*\.yaml
+        files: ^tmt/(?:schemas|steps/prepare/feature)/.*\.yaml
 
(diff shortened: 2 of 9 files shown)


therazix (approved):

seberm (approved):

ansible/packages.yml
```diff
@@ -1,3 +1,5 @@
+---
```
psss, line 1: Why are these necessary?

psss: I understand the multiple files thing. But... why would you like to separate multiple documents if there's just one? I mean, I've seen this a couple times, but still don't get it why is this supposed to be a good practice. Extra lines which help... nothing? No push against, I'm just wondering... and trying to understand.

psss (approved):

psss:
Failures are irrelevant, merging.

PR #3452, opened 2025-01-08
Enable the `stable` copr repository for building

psss:
This would allow us to test with fresh `fmf` bits as soon as they are released. We would not have to wait until packages reach the stable fedora repositories.


therazix (approved):

happz (approved):

psss:
Rawhide failure is irrelevant TF problem, everything else green, merging.

PR #3454, opened 2025-01-08
Switch package manager tests from diffutils to another package

happz:
diffutils now cannot be removed from coreos images, therefore we are picking another packages to be used for package manager units tests. Not in all cases, some images are not affected.

Pull Request Checklist

* [x] implement the feature


psss (approved):
Looks good, thanks!

PR #3459, opened 2025-01-09
Add `system.management-interface` hardware property

thrix:
Users wanting to select a system with available system management interface can use this field.

Being added to be able to select Beaker machines with `IPMI`.

```
provision:
  hardware:
    system:
      management-interface: IPMI
```

Seems there are other management interfaces which could have in the future, like:

```
Redfish
Intel Active Management Technology (AMT)
Dell iDRAC (Integrated Dell Remote Access Controller)
HP iLO (Integrated Lights-Out)
Lenovo XClarity
Supermicro IPMI/BMC
OpenBMC
Cisco UCS Manager
SMASH (System Management Architecture for Server Hardware)
```

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/spec/hardware/system.fmf
+++ b/spec/hardware/system.fmf
@@ -26,11 +26,17 @@ description: |
                # String, name of the management interface protocol
                protocol: IPMI
 
-               # Integer or string, management interface vendor name or id
-               vendor: "~ MontaVista Software, Inc."
+               # Number or string, an ID of the management interface vendor.
+               vendor: "0x04b3"
 
-               # Integer or string, management interface product name or id
-               product: 0x4010
+               # String, a name of the management interface vendor.
+               vendor-name: "~ MontaVista Software, Inc."
+
+               # Number or string, an ID of the management interface device (product).
+               device: "0x04b3"
+
+               # String, a name of the management interface device (product).
+               device-name: "~ Some device name"
 
     .. versionchanged:: 1.45
        added ``management-interface`` specification
--- a/spec/hardware/system.fmf
+++ b/spec/hardware/system.fmf
@@ -20,6 +20,21 @@ description: |
            # Integer or string, required number of NUMA nodes.
            numa-nodes: 2|">= 2"
 
+           # System management interface
+           management-interface:
+
+               # String, name of the management interface protocol
+               protocol: IPMI
+
+               # Integer or string, management interface vendor name or id
+               vendor: "~ MontaVista Software, Inc."
+
+               # Integer or string, management interface product name or id
+               product: 0x4010
+
+    .. versionchanged:: 1.45
+       added ``management-interface`` specification
+
     .. versionchanged:: 1.39
        ``beaker`` plugin supports ``vendor-name``
 
@@ -43,6 +58,12 @@ example:
         vendor-name: "~ HPE"
         numa-nodes: ">= 4"
 
+  - |
+    # Select any system with available Intelligent Platform Management Interface (IPMI).
+    system:
+        management-interface:
+            protocol: IPMI
+
 link:
   - implemented-by: /tmt/steps/provision/mrack.py
-    note: "``system.vendor`` and ``system.model`` not implemented yet"
+    note: "``system.vendor``, ``system.model`` and ``management-interface`` not implemented yet"
--- a/spec/hardware/system.fmf
+++ b/spec/hardware/system.fmf
@@ -38,7 +38,7 @@ description: |
                # String, a name of the management interface device (product).
                device-name: "~ Some device name"
 
-    .. versionchanged:: 1.45
+    .. versionchanged:: 1.46
        added ``management-interface`` specification
 
     .. versionchanged:: 1.39
--- a/spec/hardware/system.fmf
+++ b/spec/hardware/system.fmf
@@ -20,8 +20,8 @@ description: |
            # Integer or string, required number of NUMA nodes.
            numa-nodes: 2|">= 2"
 
-           # System management interface
-           management-interface:
+           # System management controller
+           management-controller:
 
                # String, name of the management interface protocol
                protocol: IPMI
@@ -39,7 +39,7 @@ description: |
                device-name: "~ Some device name"
 
     .. versionchanged:: 1.46
-       added ``management-interface`` specification
+       added ``management-controller`` specification
 
     .. versionchanged:: 1.39
        ``beaker`` plugin supports ``vendor-name``
@@ -67,9 +67,9 @@ example:
   - |
     # Select any system with available Intelligent Platform Management Interface (IPMI).
     system:
-        management-interface:
+        management-controller:
             protocol: IPMI
 
 link:
   - implemented-by: /tmt/steps/provision/mrack.py
-    note: "``system.vendor``, ``system.model`` and ``management-interface`` not implemented yet"
+    note: "``system.vendor``, ``system.model`` and ``management-controller`` not implemented yet"
--- a/tmt/schemas/provision/hardware.yaml
+++ b/tmt/schemas/provision/hardware.yaml
@@ -310,7 +310,7 @@ definitions:
           - type: string
           - type: integer
 
-      management-interface:
+      management-controller:
         type: object
 
         properties:
--- a/tmt/schemas/provision/hardware.yaml
+++ b/tmt/schemas/provision/hardware.yaml
@@ -310,6 +310,23 @@ definitions:
           - type: string
           - type: integer
 
+      management-interface:
+        type: object
+
+        properties:
+          protocol:
+            type: string
+
+          product:
+            anyOf:
+              - type: string
+              - type: integer
+
+          vendor:
+            anyOf:
+              - type: string
+              - type: integer
+
     additionalProperties: false
 
     # enforce at least one property
--- a/tmt/schemas/provision/hardware.yaml
+++ b/tmt/schemas/provision/hardware.yaml
@@ -317,15 +317,21 @@ definitions:
           protocol:
             type: string
 
-          product:
+          device:
             anyOf:
               - type: string
               - type: integer
+
+          device-name:
+            type: string
 
           vendor:
             anyOf:
               - type: string
               - type: integer
+
+          vendor-name:
+            type: string
 
     additionalProperties: false
 


spec/hardware/system.fmf
```diff
@@ -20,6 +20,9 @@ description: |
            # Integer or string, required number of NUMA nodes.
            numa-nodes: 2|">= 2"
 
+           # System management interface
```
happz, line 23: Should be `String, system management interface.`

spec/hardware/system.fmf
```diff
@@ -20,6 +20,9 @@ description: |
...
+           management-interface: IPMI
+
     .. versionchanged:: 1.39
```
happz, line 26: Add also `versionchanged` block for the new constraint:

```
    .. versionchanged:: 1.33
       Added ``location.lab-controller`` into specification.
```

psss (commented):
The naming sounds reasonable. The possible values should be probably defined as well, once we have a tangible list. I've quickly checked those brainstormed values and for example `Lenovo XClarity` is described as `Lenovo's implementation of IPMI` so there might be some clarifications needed.

pcahyna:
I don't understand some entries in the suggested list.
```
Dell iDRAC (Integrated Dell Remote Access Controller)
HP iLO (Integrated Lights-Out)
Lenovo XClarity
```
iDRAC etc. is not a management interface, it is the management platform on DELL servers, i.e. the hardware:
> The controller has its own processor, memory, network connection, and access to the system bus. Key features include power management, virtual media access and remote console capabilities, all available through a supported web browser or command-line interface

(https://en.wikipedia.org/wiki/Dell_DRAC)

It also implements an IPMI interface, in addition to web etc., which is what one would generally be interested in when requesting a machine with a management interface, so one would ask for it by requesting `IPMI`. Just as like `boot.method` does not have values like `American Megatrends` or `Phoenix`, but `bios` and `uefi`, `system.management-interface` should have values like IPMI and Redfish and not `iDRAC` or `iLO`.

Other questions:
- will it be possible to support machines that implement multiple management interfaces? It is very common for servers today to support both IPMI and Redfish (as the industry transitions from IPMI to Redfish, it is needed to have both at the same time for a long period)
- if yes, will it be possible to request a machine that supports both IPMI and Redfish, if one needs to test how the two interact? Not sure how the syntax would look like, if one needs to ask for `management-interface: IPMI` and `management-interface: Redfish` at the same time.
- how to support additional attributes of the management interface? For example here is the `dmidecode` output showing IPMI:
```
Handle 0x0012, DMI type 38, 18 bytes
IPMI Device Information
        Interface Type: KCS (Keyboard Control Style)
        Specification Version: 2.0
        I2C Slave Address: 0x10
        NV Storage Device: Not Present
        Base Address: 0x0000000000000CA2 (I/O)
        Register Spacing: Successive Byte Boundaries
```
at least the `Specification Version: ` might be interesting to request. `bmc-info` shows more attributes:
```
Device ID             : 0
Device Revision       : 3
Device SDRs           : unsupported
Firmware Revision     : 9.08
Device Available      : yes (normal operation)
IPMI Version          : 2.0
Sensor Device         : supported
SDR Repository Device : supported
SEL Device            : supported
FRU Inventory Device  : supported
IPMB Event Receiver   : supported
IPMB Event Generator  : unsupported
Bridge                : unsupported
Chassis Device        : supported
Manufacturer ID       : MontaVista Software, Inc. (4753)
Product ID            : 3842
Auxiliary Firmware Revision Information : 00000000h

System GUID : efcdab89-6745-23a1-efcd-ab89674523a1
```
I suppose at least `Manufacturer ID` and `Product ID` would be interesting to have (it would be a more standard way for asking for a specific manufacturer if one needs to test quirks of a specific manufacturer than to ask for `iDRAC` or `iLO`). And here is the output for Redfish (XClarity controller):
```
Handle 0x006F, DMI type 42, 169 bytes
Management Controller Host Interface
        Host Interface Type: Network
        Device Type: USB
        idVendor: 0x04b3
        idProduct: 0x4010
        Protocol ID: 04 (Redfish over IP)
                Service UUID: f84555e2-779c-11ec-859c-c77fb77990a8
                Host IP Assignment Type: Static
                Host IP Address Format: IPv4
                IPv4 Address: 169.254.95.120
                IPv4 Mask: 255.255.0.0
                Redfish Service IP Discovery Type: Static
                Redfish Service IP Address Format: IPv4
                IPv4 Redfish Service Address: 169.254.95.118
                IPv4 Redfish Service Mask: 255.255.0.0
                Redfish Service Port: 443
                Redfish Service Vlan: 0
                Redfish Service Hostname: XCC-7D8F-1234567890

Handle 0x0070, DMI type 42, 169 bytes
Management Controller Host Interface
        Host Interface Type: Network
        Device Type: USB
        idVendor: 0x04b3
        idProduct: 0x4010
        Protocol ID: 04 (Redfish over IP)
                Service UUID: f84555e2-779c-11ec-859c-c77fb77990a8
                Host IP Assignment Type: Static
                Host IP Address Format: IPv4
                IPv4 Address: 169.254.95.120
                IPv4 Mask: 255.255.0.0
                Redfish Service IP Discovery Type: Static
                Redfish Service IP Address Format: IPv6
                IPv6 Redfish Service Address: fe80::67b:cbff:fe7b:8354
                IPv6 Redfish Service Mask: ffff::
                Redfish Service Port: 443
                Redfish Service Vlan: 0
                Redfish Service Hostname: XCC-7D8F-1234567890
```
again, idVendor and idProduct may be interesting to expose. Output for iDRAC:
```
Handle 0x2A00, DMI type 42, 118 bytes
Management Controller Host Interface
        Host Interface Type: Network
        Device Type: USB
        idVendor: 0x413c
        idProduct: 0xa102
        Protocol ID: 04 (Redfish over IP)
                Service UUID: Not Settable
                Host IP Assignment Type: DHCP
                Host IP Address Format: IPv4
                Redfish Service IP Discovery Type: Static
                Redfish Service IP Address Format: IPv4
                IPv4 Redfish Service Address: 169.254.1.1
                IPv4 Redfish Service Mask: 255.255.255.0
                Redfish Service Port: 443
                Redfish Service Vlan: 4294967295
                Redfish Service Hostname: idrac.local
```
I am not asking to have the additional attributes available now, just to think how they could be added to the syntax - I don't see how they would fit into the proposed syntax, maybe something like
```
provision:
  hardware:
    system:
      management-interface:
        IPMI:
          <additional attributes go here>
```
instead? dunno.
- maybe `management-protocol` instead of `management-interface`? `interface` is correct, but one may think that it refers to the actual physical interface on the machine (usually the Ethernet port and its NIC), not to an interface as an abstraction.
- last note, why IPMI in all caps? It is the official name of the protocol and is written in caps, sure, but you don't write `bios` and `uefi` in caps either.

happz:
> ...
>
> Other questions:
> 
> * will it be possible to support machines that implement multiple management interfaces? It is very common for servers today to support both IPMI and Redfish (as the industry transitions from IPMI to Redfish, it is needed to have both at the same time for a long period)
> * if yes, will it be possible to request a machine that supports both IPMI and Redfish, if one needs to test how the two interact? Not sure how the syntax would look like, if one needs to ask for `management-interface: IPMI` and `management-interface: Redfish` at the same time.

I believe this will be supported:

```yaml
hardware:
  and:
    - system:
        management-interface: IPMI_v1
    - system:
        management-interface: IPMI_v1
```

> * how to support additional attributes of the management interface? For example here is the `dmidecode` output showing IPMI:
> 
> ```
> Handle 0x0012, DMI type 38, 18 bytes
> IPMI Device Information
>         Interface Type: KCS (Keyboard Control Style)
>         Specification Version: 2.0
>         I2C Slave Address: 0x10
>         NV Storage Device: Not Present
>         Base Address: 0x0000000000000CA2 (I/O)
>         Register Spacing: Successive Byte Boundaries
> ```
> 
> at least the `Specification Version: ` might be interesting to request. `bmc-info` shows more attributes:
> 
> ```
> Device ID             : 0
> Device Revision       : 3
> Device SDRs           : unsupported
> Firmware Revision     : 9.08
> Device Available      : yes (normal operation)
> IPMI Version          : 2.0
> Sensor Device         : supported
> SDR Repository Device : supported
> SEL Device            : supported
> FRU Inventory Device  : supported
> IPMB Event Receiver   : supported
> IPMB Event Generator  : unsupported
> Bridge                : unsupported
> Chassis Device        : supported
> Manufacturer ID       : MontaVista Software, Inc. (4753)
> Product ID            : 3842
> Auxiliary Firmware Revision Information : 00000000h
> 
> System GUID : efcdab89-6745-23a1-efcd-ab89674523a1
> ```
> 
> I suppose at least `Manufacturer ID` and `Product ID` would be interesting to have (it would be a more standard way for asking for a specific manufacturer if one needs to test quirks of a specific manufacturer than to ask for `iDRAC` or `iLO`). And here is the output for Redfish (XClarity controller):
> 
> ```
> Handle 0x006F, DMI type 42, 169 bytes
> Management Controller Host Interface
>         Host Interface Type: Network
>         Device Type: USB
>         idVendor: 0x04b3
>         idProduct: 0x4010
>         Protocol ID: 04 (Redfish over IP)
>                 Service UUID: f84555e2-779c-11ec-859c-c77fb77990a8
>                 Host IP Assignment Type: Static
>                 Host IP Address Format: IPv4
>                 IPv4 Address: 169.254.95.120
>                 IPv4 Mask: 255.255.0.0
>                 Redfish Service IP Discovery Type: Static
>                 Redfish Service IP Address Format: IPv4
>                 IPv4 Redfish Service Address: 169.254.95.118
>                 IPv4 Redfish Service Mask: 255.255.0.0
>                 Redfish Service Port: 443
>                 Redfish Service Vlan: 0
>                 Redfish Service Hostname: XCC-7D8F-1234567890
> 
> Handle 0x0070, DMI type 42, 169 bytes
> Management Controller Host Interface
>         Host Interface Type: Network
>         Device Type: USB
>         idVendor: 0x04b3
>         idProduct: 0x4010
>         Protocol ID: 04 (Redfish over IP)
>                 Service UUID: f84555e2-779c-11ec-859c-c77fb77990a8
>                 Host IP Assignment Type: Static
>                 Host IP Address Format: IPv4
>                 IPv4 Address: 169.254.95.120
>                 IPv4 Mask: 255.255.0.0
>                 Redfish Service IP Discovery Type: Static
>                 Redfish Service IP Address Format: IPv6
>                 IPv6 Redfish Service Address: fe80::67b:cbff:fe7b:8354
>                 IPv6 Redfish Service Mask: ffff::
>                 Redfish Service Port: 443
>                 Redfish Service Vlan: 0
>                 Redfish Service Hostname: XCC-7D8F-1234567890
> ```
> 
> again, idVendor and idProduct may be interesting to expose. Output for iDRAC:
> 
> ```
> Handle 0x2A00, DMI type 42, 118 bytes
> Management Controller Host Interface
>         Host Interface Type: Network
>         Device Type: USB
>         idVendor: 0x413c
>         idProduct: 0xa102
>         Protocol ID: 04 (Redfish over IP)
>                 Service UUID: Not Settable
>                 Host IP Assignment Type: DHCP
>                 Host IP Address Format: IPv4
>                 Redfish Service IP Discovery Type: Static
>                 Redfish Service IP Address Format: IPv4
>                 IPv4 Redfish Service Address: 169.254.1.1
>                 IPv4 Redfish Service Mask: 255.255.255.0
>                 Redfish Service Port: 443
>                 Redfish Service Vlan: 4294967295
>                 Redfish Service Hostname: idrac.local
> ```
> 
> I am not asking to have the additional attributes available now, just to think how they could be added to the syntax - I don't see how they would fit into the proposed syntax, maybe something like
> 
> ```
> provision:
>   hardware:
>     system:
>       management-interface:
>         IPMI:
>           <additional attributes go here>
> ```
> 
> instead? dunno.

tmt already supports `vendor`, `vendor-name`, `device`, and `device-name` attributes across several types of devices, I see them as the right fit for `Manufacturer ID`, `Product ID`, and similar fields. The position is another question, since `management-interface` is string-only. And it's a good question you mentioned it, because it seems we should change the specification to something like this:

```yaml
system:
  management-interface:
    # Now what is `IPMI`? How should we call it, a protocol, implementation? `protocol` was suggested below:
    protocol: IPMI
    # Now we can support more fields here, not mixing IPMI with them:
    vendor-name: "~ MontaVista Software, Inc."

# A machine with both IPMI and Redfish?
and:
  - system:
      management-interface:
        protocol: IPMI
  - system:
      management-interface:
        protocol: Redfish
        vendor: 0x04b3
```

And, in the future, more keys can start appearing here if necessary.

> * maybe `management-protocol` instead of `management-interface`? `interface` is correct, but one may think that it refers to the actual physical interface on the machine (usually the Ethernet port and its NIC), not to an interface as an abstraction.

I agree, and I like `protocol` better. Plus, once under `management-interface`, I'd prefer `protocol`, i.e. not repeating `management`.

> * last note, why IPMI in all caps? It is the official name of the protocol and is written in caps, sure, but you don't write `bios` and `uefi` in caps either.

Eeee, I do :) I believe tmt is case-insensitive in this area, or better should be, it should give you an Intel CPU with `vendor-name: Intel` and `vendor-name: iNTEL`, but IIRC, this has not been stated anywhere, and there might be places where case-sensitivity is not handled correctly. Something to look into.

thrix:
@pcahyna @psss @happz  thanks for the discussion! I will try to incorporate what Milos advised. @pcahyna can you check if it sounds good to you?

psss:
@pcahyna ping?

psss:
@thrix, do you still want to squeeze this into `1.45`? To be released today.

spec/hardware/system.fmf
thrix: @happz addressed

spec/hardware/system.fmf
thrix: added

thrix:
> > @thrix, do you still want to squeeze this into `1.45`? To be released today.
> 
> I am working on it right now, eta 1-2 hours

Sorry, these meetings are a performance killer.

I updated it according to the comments, the best solution I could come up with according to the discussion. Went with 3 subfields:

* protocol
* vendor
* product

thrix:
@pcahyna @psss @happz kindly asking for review

spec/hardware/system.fmf
```diff
@@ -20,6 +20,21 @@ description: |
            # Integer or string, required number of NUMA nodes.
            numa-nodes: 2|">= 2"
 
+           # System management interface
+           management-interface:
+
+               # String, name of the management interface protocol
```
thrix, line 26:
```suggestion
               # String, a name of the management interface protocol.
```

spec/hardware/system.fmf
```diff
@@ -20,6 +20,21 @@ description: |
...
+               protocol: IPMI
+
+               # Integer or string, management interface vendor name or id
```
thrix, line 29:
```suggestion
               # Integer or string, an management interface vendor name or id.
```

spec/hardware/system.fmf
```diff
@@ -20,6 +20,21 @@ description: |
...
+               vendor: "~ MontaVista Software, Inc."
+
+               # Integer or string, management interface product name or id
```
thrix, line 32:
```suggestion
               # Integer or string, an management interface product name or id.
```

spec/hardware/system.fmf
```diff
@@ -20,6 +20,21 @@ description: |
            # Integer or string, required number of NUMA nodes.
            numa-nodes: 2|">= 2"
 
+           # System management interface
+           management-interface:
+
+               # String, a name of the management interface protocol.
+               protocol: IPMI
+
+               # Integer or string, an management interface vendor name or id.
```
thrix, line 29:
```suggestion
               # Integer or string, an ID or name of the management interface vendor.
```

spec/hardware/system.fmf
```diff
@@ -20,6 +20,21 @@ description: |
...
+               vendor: "~ MontaVista Software, Inc."
+
+               # Integer or string, an management interface product name or id.
```
thrix, line 32:
```suggestion
               # Integer or string, an ID or name of the management interface product.
```

psss:
@thrix, I'm trying to finish `1.45` in #3622. Would you mind if we handle this one in the next release?

spec/hardware/system.fmf
```diff
@@ -20,6 +20,21 @@ description: |
...
            numa-nodes: 2|">= 2"
 
+           # System management interface
+           management-interface:
+
+               # String, a name of the management interface protocol.
+               protocol: IPMI
+
+               # Integer or string, an ID or name of the management interface vendor.
+               vendor: "~ MontaVista Software, Inc."
```
psss, line 30: We're already using pairs like `vendor` and `vendor-name` or `model` and `model-name`. Shall we possibly apply the same consistent approach here as well?

happz: Same for `product`, that's `device` and `device-name`.

pcahyna:
> @pcahyna @psss @happz thanks for the discussion! I will try to incorporate what Milos advised. @pcahyna can you check if it sounds good to you?

sorry, I have not realized that you have been waiting on me. yes, this looks good.

pcahyna:
sorry for the late comment. Wopuld it be appropriate to rename the property to `management-controller` ? Then we could in the future have properties under it that correspond to the actual interfaces (like network interface) of the controller, like IPv4 interface, IPv6 interface, serial interface, whatever.
Also, I think it could make sense to ask the maintainer of the kernel driver. He probably works with IPMI more than I do.

happz:
> sorry for the late comment. Would it be appropriate to rename the property to `management-controller` ? Then we could in the future have properties under it that correspond to the actual interfaces (like network interface) of the controller, like IPv4 interface, IPv6 interface, serial interface, whatever.

`management-controller` sounds good to me.

spec/hardware/system.fmf
pcahyna: so, will it be `model`(`-name`) or `device`(`-name`) ? I don't see the difference between the two. Is the specification already inconsistent?

happz: We are using `device` and `vendor` for ID-like properties, and `device-name` and `vendor-name` for label-like properties, names and such.

`model` and `model-name` are defined only for CPU, and it's true that's pretty much the same meaning. I'd recommend adding `device` and `device-name` properties to CPU group rather then adding `model` and `model-name` anywhere else.

happz: Dammit. No, `device` and `device-name` it is. We shall root this out. I'm not going to support seven different aliases.

pcahyna: For disks it is not even an alias, `device-name` is not in the document.

thrix: So changed to :

* protocol
* device and device-name
* vendor and vendor-name

thrix:
@psss @happz @pcahyna pls review

spec/hardware/system.fmf
```diff
@@ -20,6 +20,27 @@ description: |
...
+               # String, a name of the management interface vendor.
+               vendor-name: "~ MontaVista Software, Inc."
+
+               # Number or string, an ID of the management interface device (product).
+               device: "0x04b3"
+
+               # String, a name of the management interface device (product).
+               device-name: "~ Some device name"
+
+    .. versionchanged:: 1.45
```
thrix, line 41:
```suggestion
    .. versionchanged:: 1.46
```

happz (approved):

happz:
@thrix https://github.com/teemtee/tmt/pull/3459#issuecomment-2758835618 is left without an answer.

therazix (approved):

psss:
> Would it be appropriate to rename the property to `management-controller` ? Then we could in the future have properties under it that correspond to the actual interfaces (like network interface) of the controller, like IPv4 interface, IPv6 interface, serial interface, whatever.

Agreed, `management-controller` sounds a bit better to me, as we're trying to describe not only they ways we interact with it (interface) but also the `device` itself. @thrix, would you be ok with `management-controller`?

thrix:
@happz @psss @pcahyna renamed to `management-controller`

happz (approved):

tmt/schemas/provision/hardware.yaml
```diff
@@ -310,6 +310,29 @@ definitions:
...
+      management-controller:
+        type: object
+
+        properties:
+          protocol:
+            type: string
+
+          device:
+            anyOf:
+              - type: string
```
pcahyna, line 322: sorry for being late, but why string?

tmt/schemas/provision/hardware.yaml
```diff
@@ -310,6 +310,29 @@ definitions:
...
+              - type: integer
+
+          device-name:
+            type: string
+
+          vendor:
+            anyOf:
+              - type: string
```
pcahyna, line 330: ditto, why string?

tmt/schemas/provision/hardware.yaml
happz:
```yaml
# I despise Intel's devices:
device: "!= 32902"
```

PR #3465, opened 2025-01-13
Recommend `ansible-core` in the main `tmt` package

### LINKED ISSUE #3425: Unusable `prepare -h ansible` in tmt minimal install
tmt doesn't require 'ansible' in minimal install, making this prepare step (and all features?) unusable. 

```
# dnf install -y tmt
# tmt --feeling-safe  run prepare -h ansible --playbook playbook.yaml provision -h local
...
        how: ansible
        playbook: playbook.yaml
        fail: Command 'sudo -E ansible-playbook -c local -i localhost, /playbook.yaml' returned 1.
...
        # stderr (1/1 lines)
        sudo: ansible-playbook: command not found
```

bajertom:
Pull Request Checklist

Fixes #3425 
* [x] implement the feature
* [x] include a release note
* [x] write the documentation

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -11,6 +11,10 @@ tmt-1.42.0
 The ``tmt show`` command now prints in verbose mode manual test
 instructions as well.
 
+Introducing new subpackage ``tmt+ansible``, to pull in Ansible
+requirements for plugins that use Ansible under the hood,
+:ref:`/plugins/prepare/ansible`, :ref:`/plugins/finish/ansible`,
+and :ref:`/plugins/prepare/feature`.
 
 tmt-1.41.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -12,7 +12,7 @@ The ``tmt show`` command now prints in verbose mode manual test
 instructions as well.
 
 The ``ansible-core`` package is now a recommended dependency package
-for tmt. It is usedby plugins that use Ansible under the hood,
+for tmt. It is used by plugins that use Ansible under the hood,
 :ref:`prepare/ansible</plugins/prepare/ansible>`,
 :ref:`finish/ansible</plugins/finish/ansible>`,
 and :ref:`prepare/feature</plugins/prepare/feature>`.
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -11,10 +11,11 @@ tmt-1.42.0
 The ``tmt show`` command now prints in verbose mode manual test
 instructions as well.
 
-Introducing new subpackage ``tmt+ansible``, to pull in Ansible
-requirements for plugins that use Ansible under the hood,
-:ref:`/plugins/prepare/ansible`, :ref:`/plugins/finish/ansible`,
-and :ref:`/plugins/prepare/feature`.
+The ``ansible-core`` package is now a recommended dependency package
+for tmt. It is usedby plugins that use Ansible under the hood,
+:ref:`prepare/ansible</plugins/prepare/ansible>`,
+:ref:`finish/ansible</plugins/finish/ansible>`,
+and :ref:`prepare/feature</plugins/prepare/feature>`.
 
 tmt-1.41.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tmt.spec
+++ b/tmt.spec
@@ -36,6 +36,7 @@ It contains no code, just makes sure the dependencies are installed.}
 The tmt Python module and command line tool implement the test
 metadata specification (L1 and L2) and allows easy test execution.
 
+%pyproject_extras_subpkg -n tmt ansible
 %pyproject_extras_subpkg -n tmt export-polarion
 %pyproject_extras_subpkg -n tmt report-junit
 %pyproject_extras_subpkg -n tmt report-polarion
@@ -111,17 +112,6 @@ Requires:       tmt == %{version}-%{release}
 Requires:       python3-mrack-beaker
 
 %description -n tmt+provision-beaker %_metapackage_description
-
-%package -n     tmt+ansible
-Summary:        Dependencies required for tmt using ansible
-Provides:       tmt+ansible == %{version}-%{release}
-%if 0%{?fedora} < 40
-Obsoletes:      tmt+ansible < %{version}-%{release}
-%endif
-Requires:       tmt == %{version}-%{release}
-Requires:       ansible-core
-
-%description -n tmt+ansible %_metapackage_description
 
 # Replace with pyproject_extras_subpkg at some point
 %package -n     tmt+all
--- a/tmt.spec
+++ b/tmt.spec
@@ -25,6 +25,7 @@ Obsoletes:      tmt-report-reportportal < %{version}-%{release}
 %endif
 
 Recommends:     bash-completion
+Recommends:	ansible-core
 
 %py_provides    python3-tmt
 
@@ -65,7 +66,6 @@ Obsoletes:      tmt-provision-container < %{version}-%{release}
 %endif
 Requires:       tmt == %{version}-%{release}
 Requires:       podman
-Requires:       tmt+ansible == %{version}-%{release}
 Requires:       (ansible-collection-containers-podman or ansible)
 
 %description -n tmt+provision-container %_metapackage_description
@@ -80,7 +80,6 @@ Requires:       tmt == %{version}-%{release}
 Requires:       python3-testcloud >= 0.11.3
 Requires:       libvirt-daemon-config-network
 Requires:       openssh-clients
-Requires:       tmt+ansible == %{version}-%{release}
 # Recommend qemu system emulators for supported arches
 Recommends:     qemu-kvm-core
 %if 0%{?fedora}
--- a/tmt.spec
+++ b/tmt.spec
@@ -64,7 +64,8 @@ Obsoletes:      tmt-provision-container < %{version}-%{release}
 %endif
 Requires:       tmt == %{version}-%{release}
 Requires:       podman
-Requires:       (ansible or ansible-collection-containers-podman)
+Requires:       tmt+ansible == %{version}-%{release}
+Requires: 	ansible-collection-containers-podman or ansible
 
 %description -n tmt+provision-container %_metapackage_description
 
@@ -78,7 +79,7 @@ Requires:       tmt == %{version}-%{release}
 Requires:       python3-testcloud >= 0.11.3
 Requires:       libvirt-daemon-config-network
 Requires:       openssh-clients
-Requires:       (ansible or ansible-core)
+Requires:       tmt+ansible == %{version}-%{release}
 # Recommend qemu system emulators for supported arches
 Recommends:     qemu-kvm-core
 %if 0%{?fedora}
@@ -110,6 +111,17 @@ Requires:       tmt == %{version}-%{release}
 Requires:       python3-mrack-beaker
 
 %description -n tmt+provision-beaker %_metapackage_description
+
+%package -n     tmt+ansible
+Summary:        Dependencies required for tmt using ansible
+Provides:       tmt+ansible == %{version}-%{release}
+%if 0%{?fedora} < 40
+Obsoletes:      tmt+ansible < %{version}-%{release}
+%endif
+Requires:       tmt == %{version}-%{release}
+Requires:       ansible-core
+
+%description -n tmt+ansible %_metapackage_description
 
 # Replace with pyproject_extras_subpkg at some point
 %package -n     tmt+all
--- a/tmt.spec
+++ b/tmt.spec
@@ -66,7 +66,7 @@ Obsoletes:      tmt-provision-container < %{version}-%{release}
 Requires:       tmt == %{version}-%{release}
 Requires:       podman
 Requires:       tmt+ansible == %{version}-%{release}
-Requires: 	ansible-collection-containers-podman or ansible
+Requires:       (ansible-collection-containers-podman or ansible)
 
 %description -n tmt+provision-container %_metapackage_description
 
--- a/tmt.spec
+++ b/tmt.spec
@@ -37,7 +37,6 @@ It contains no code, just makes sure the dependencies are installed.}
 The tmt Python module and command line tool implement the test
 metadata specification (L1 and L2) and allows easy test execution.
 
-%pyproject_extras_subpkg -n tmt ansible
 %pyproject_extras_subpkg -n tmt export-polarion
 %pyproject_extras_subpkg -n tmt report-junit
 %pyproject_extras_subpkg -n tmt report-polarion
@@ -66,7 +65,7 @@ Obsoletes:      tmt-provision-container < %{version}-%{release}
 %endif
 Requires:       tmt == %{version}-%{release}
 Requires:       podman
-Requires:       (ansible-collection-containers-podman or ansible)
+Requires:       ansible-collection-containers-podman
 
 %description -n tmt+provision-container %_metapackage_description
 
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -77,6 +77,11 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
     Run Ansible playbooks against the guest, by running
     ``ansible-playbook`` for all given playbooks.
 
+    .. note::
+
+       The plugin requires a working Ansible to be available, which can
+       be achieved by installing ``tmt+ansible`` package.
+
     .. warning::
 
         When specifying playbooks with paths:
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -80,7 +80,7 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
     .. note::
 
        The plugin requires a working Ansible to be available on the
-       :ref:`test-runner`.
+       test runner.
 
     .. warning::
 
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -79,8 +79,8 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
 
     .. note::
 
-       The plugin requires a working Ansible to be available, which can
-       be achieved by installing ``tmt+ansible`` package.
+       The plugin requires a working Ansible to be available on the
+       :ref:`test-runner`.
 
     .. warning::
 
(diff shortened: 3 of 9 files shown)


tmt.spec
```diff
@@ -111,6 +111,18 @@ Requires:       python3-mrack-beaker
...
 
+%package -n     tmt+ansible
+Summary:        Dependencies required for tmt using ansible
+Provides:       tmt-ansible == %{version}-%{release}
+%if 0%{?fedora} < 40
+Obsoletes:      tmt-ansible < %{version}-%{release}
+%endif
+Requires:       tmt == %{version}-%{release}
+Requires:       (ansible or ansible-core)
+Requires:       ansible-collection-containers-podman
```
happz, line 122: Hmm, I'm not sure this is the correct combination. In the original scenario, after installing both plugins: if `ansible` is available, both would pull in just the `ansible` package; if `ansible` is not available, they would pull in the alternatives, `ansible-core` and `ansible-collection-container-podman`. With `tmt+ansible`, we get either `ansible` + `ansible-collection-containers-podman`, or `ansible-core` + `ansible-collection-containers-podman`. Which doesn't match the original situation.

Plus, installing `tmt+ansible` just to make `--how local` work, as requested in the issue`, would pull in `ansible-collection-container-podman` which is required for containers, and I did not plan to ever use containers :)

Maybe we should use `tmt+ansible` to pull in the generic parts - `ansible` or `ansible-core`, and leave `ansible-collection-container-podman` required just by `tmt+-provision-container`?

tmt.spec
```diff
@@ -64,7 +64,7 @@ Obsoletes:      tmt-provision-container < %{version}-%{release}
 %endif
 Requires:       tmt == %{version}-%{release}
 Requires:       podman
-Requires:       (ansible or ansible-collection-containers-podman)
+Requires:       tmt-ansible == %{version}-%{release}
```
psss, line 67:
```suggestion
Requires:       tmt+ansible == %{version}-%{release}
```

The plus sign should be used in the `Requires` sections as well, right?

tmt.spec
psss: I believe the `ansible or ansible-core` hack was only needed for a specific situation around `fedora-36` where the package renaming happened. Now the `ansible` package always depends on `ansible-core`. So... I guess we can only require `ansible-core` now?

psss: > Maybe we should use `tmt+ansible` to pull in the generic parts - `ansible` or `ansible-core`, and leave `ansible-collection-container-podman` required just by `tmt+-provision-container`?

That collection is really a small package, so not much extra content. But... at the second look... @happz is right, this is really closely related to the `provision-container` subpackage.

happz: What is the situation with `ansible-collection-container-podman`? It's `ansible or ansible-collection-container-podman` still needed?

psss: For executing in a container I believe it is. Here's the change which was adding it: https://github.com/teemtee/tmt/commit/da225ee9843f288a27dd46d463fa85c9dbb298d2

happz: There is no such package on my Fedora 41 laptop. `tmt+provision-container` pulls in `ansible` which provides podman-related plugins. `ansible-core` is indeed unusable for containers, therefore depending on `tmt+ansible` which depends on `ansible-core` will not be enough. It seems to me `tmt+provision-container` should keep its `ansible or ansible-collection-container-podman` requirement, and gain new `tmt+ansible` just to pull in whatever "Ansible" means beyond podman integration.

psss: Summary from the review session:

```
tmt+ansible
Require: ansible-core

tmt+provision-virtual
Require: tmt+ansible

tmt+provision-container
Require: tmt+ansible
Require: ansible-collection-containers-podman or ansible
```

lukaszachy:
All (?) [feature](https://tmt.readthedocs.io/en/stable/spec/plans.html#feature) are implemented as ansible playbooks , 
so do I understand correctly that  will not work with tmt minimal install? Could it be also documented then?
Or should we have a better name for this package?

bajertom:
@lukaszachy  As far as I understand it, it will not work with tmt minimal install. I will document this,  for the prepare/feature too.

lukaszachy:
@psss Is the 'prepare feature' a "core feature"? We say "Minimal tmt installation with core features only" as use case for minimal install - https://tmt.readthedocs.io/en/stable/stories/install.html#minimal

Do we even say which are those?

psss:
> Is the 'prepare feature' a "core feature"?

I don't think so, as it requires `ansible` and we want the minimal package to have really minimum dependencies.

> Do we even say which are those?

Hm, good point, this has not been precisely defined. Let's extend the description with at least a basic list.

* #3472

docs/releases.rst
```diff
@@ -25,6 +25,8 @@ the given job group.
 The ``note`` field of tmt :ref:`/spec/plans/results` changes from
 a string to a list of strings, to better accommodate multiple notes.
 
+New subpackage ``tmt+ansible`` introduced. It covers prepare step
+with ``--how ansible`` option by installing ansible requirements.
```
happz, line 29: * Should mention other Ansible-dependant plugins too, `prepare/feature`.
* Do not use command-line `--how`, this applies to plugins and let's assume users know how to select a plugin :)

How about the following?

```
Introducing new subpackage ``tmt+ansible``, to pull in Ansible requirements for plugins
that use Ansible under the hood, :ref:`/plugins/prepare/ansible`,:ref:`/plugins/finish/ansible`,
and :ref:`/plugins/prepare/feature`.
```

spec/plans/prepare.fmf
```diff
@@ -106,6 +108,10 @@ example: |
         supported, ``crb`` and ``fips`` are coming in the near
         future.
 
+        .. note::
+            Subpackage tmt+ansible need to be installed for this
+            feature to work.
+
```
happz, line 114: > Subpackage tmt+ansible need to be installed for this feature to work.

No, it doesn't :) What the plugin need is a working Ansible, which *may* be installed by installing `tmt+ansible`.

Also, these notes belong to plugin documentation, i.e. plugin docstrings, they are not part of the specification. Somewhere around here, I guess, https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/ansible.py#L79 ? And the same for `prepare/feature`.

How about this?

```
.. note::

   The plugin requires a working Ansible to be available, which can be achieved
   by installing ``tmt+ansible`` package.
```

docs/releases.rst
```diff
@@ -11,6 +11,10 @@ tmt-1.42.0
 The ``tmt show`` command now prints in verbose mode manual test
 instructions as well.
 
+Introducing new subpackage ``tmt+ansible``, to pull in Ansible
+requirements for plugins that use Ansible under the hood,
+:ref:`/plugins/prepare/ansible`,:ref:`/plugins/finish/ansible`,
```
happz, line 16: Missing space before second `:ref:`

bajertom:
Core tests fail on unavailability of tmt+ansible, "⚠ Test environment installation failed: Install packages", [log](https://artifacts.dev.testing-farm.io/69228a15-c022-47ba-a7f7-84659a8c6fa9/guest-setup-bfeb92f3-bc52-4eb6-85ba-667ce3cff11a/artifact-installation-bfeb92f3-bc52-4eb6-85ba-667ce3cff11a/4-Install-packages.txt).

Is this a logical outcome of the fact, that this package is getting introduced by this PR and therefore it is not yet available in any repository? It seems to be using copr repo to download and install dependencies for tmt.

Or is it something else?

happz:
> Core tests fail on unavailability of tmt+ansible, "⚠ Test environment installation failed: Install packages", [log](https://artifacts.dev.testing-farm.io/69228a15-c022-47ba-a7f7-84659a8c6fa9/guest-setup-bfeb92f3-bc52-4eb6-85ba-667ce3cff11a/artifact-installation-bfeb92f3-bc52-4eb6-85ba-667ce3cff11a/4-Install-packages.txt).
> 
> Is this a logical outcome of the fact that this package is getting introduced by this PR and therefore it is not yet available in any repository? It seems to be using copr repo to download and install dependencies for tmt.
> 
> Or is it something else?

Hmm, I would expect this package to be delivered with all other packages of the tmt build under the test, as yet another artifact to test. If I check the RPM build log, https://download.copr.fedorainfracloud.org/results/packit/teemtee-tmt-3465/fedora-41-x86_64/08527142-tmt/builder-live.log.gz, I don't see any mention of tmt+ansible though, which explains why it's not installed: it was never built.

@psss @lukaszachy any ideas why it's skipped?

martinhoyer:
Thanks for taking this Tom! Some general thoughts:

Wouldn't it make sense to have it simply as a "Suggests" or "Recommends" weak dependency of the tmt main package?  

Failing gracefully with a generic message that ansible is required to be installed would play with pypi/non-rpm installations.

IIRC this one is a bit different, as other extra packages are required for specific provisioning, report plugins, while ansible can also be used by "stuff in core/minimal tmt". Might need a discussion wether or not it should be more of a core dependency?

psss:
> IIRC this one is a bit different, as other extra packages are required for specific provisioning, report plugins, while ansible can also be used by "stuff in core/minimal tmt". Might need a discussion wether or not it should be more of a core dependency?

I don't think that `ansible` should be required by the minimal package see also:

* https://github.com/teemtee/tmt/pull/3472

Turning it into a recommended package sounds a bit better, but still, would result in installing the ansible deps by default, as users usually do not turn off optional dependencies. Handling the missing ansible with a reasonable message for pypi installed package sounds good to me as well.

psss:
> @psss @lukaszachy any ideas why it's skipped?

I think the reason is that there are no files included in the rpm package. This part is covered by the `%files` section:

https://github.com/teemtee/tmt/blob/e9c53af76652f5235087763b7b1522e7774e8d79/tmt.spec#L163-L169

Thinking about it a bit more: Both `ansible` and `ansible-core` are available on `pypi`, so, I guess, we could/should add the corresponding extra require? Which would take care of this and make the spec file shorter. Trying in aefecda9.

docs/releases.rst
```diff
@@ -11,6 +11,10 @@ tmt-1.42.0
 The ``tmt show`` command now prints in verbose mode manual test
 instructions as well.
 
+Introducing new subpackage ``tmt+ansible``, to pull in Ansible
+requirements for plugins that use Ansible under the hood,
+:ref:`/plugins/prepare/ansible`, :ref:`/plugins/finish/ansible`,
+and :ref:`/plugins/prepare/feature`.
```
psss, line 17: When rendered, this is a bit confusing to the user:

    Ansible under the hood, ansible, ansible, and feature.

What about mentioning `prepare` and `finish` steps to differentiate between the two `ansible` links?

tmt/steps/prepare/ansible.py
```diff
@@ -77,6 +77,11 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
     Run Ansible playbooks against the guest, by running
     ``ansible-playbook`` for all given playbooks.
 
+    .. note::
+
+       The plugin requires a working Ansible to be available, which can
```
psss, line 82:
```suggestion
       The plugin requires a working Ansible to be available on the :ref:`test-runner`, which can
```

psss (changes requested):
Seems to be working now, core tests passed. Added a couple ideas for improvement.

docs/releases.rst
happz: We can do that:

```
`:ref:`prepare/ansible</plugins/prepare/ansible>`, :ref:`finish/ansible</plugins/finish/ansible>`,
and :ref:`prepare/feature</plugins/prepare/feature>`
```

tmt/steps/prepare/feature/__init__.py
```diff
@@ -134,6 +134,11 @@ class PrepareFeature(tmt.steps.prepare.PreparePlugin[PrepareFeatureData]):
     """
     Enable or disable common features like repositories on the guest.
 
+    .. note::
+
+       The plugin requires a working Ansible to be available, which can
+       be achieved by installing ``tmt+ansible`` package.
```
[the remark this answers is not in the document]
happz: That should be possible, I think it would require places where `ansible-playbook` is called to be wrapped with something like https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/__init__.py#L372 ? Which means mostly `_run_ansible()` methods.

thrix:
@bajertom looking for the update as annouced, so we can get this in

psss:
Summary from the hacking session discussion:

* Do not introduce a new `tmt+ansible` subpackage
* Use `Recommend: ansible-core` for the main tmt package
* Document that this is an optional/weak dependency
* For `pip` installation include this in the main package
* Keep the nice error messages when ansible is not found

@bajertom, could you please update the pull request accordingly? Thanks.

tmt/steps/prepare/feature/__init__.py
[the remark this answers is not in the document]
bajertom: Now that tmt+ansible is out of the question, `tmt[ansible]` is what I left in the note. However I feel like repeating it 3x in the code is a bit against DRY principle. Should I do it more elegantly or is this OK?

tmt/steps/prepare/ansible.py
bajertom: I had to remove the :ref:`test-runner` and change it to "test runner" to keep the information there, otherwise it resulted in errors 
```
<rst-doc>:8: (ERROR/3) Unknown interpreted text role "ref".
Failed to import the 'tmt.steps.prepare.ansible' module from '/home/runner/work/tmt/tmt/docs' 
```

tmt/steps/provision/local.py
```diff
@@ -54,20 +54,26 @@ def _run_ansible(
...
+                    playbook),
+                env=self._prepare_environment(),
+                friendly_command=friendly_command,
+                log=log,
+                silent=silent)
+        except tmt.utils.RunError as exc:
+            if exc.stderr and 'ansible-playbook: command not found' in exc.stderr:
+                raise tmt.utils.GeneralError(
+                    "Seems that ansible is not installed on the test runner. "
+                    "Install `tmt[ansible]` optional dependency.") from exc
```
psss, line 75: Shall we cover both `pip` and `rpm` use case here? Perhaps something like this?

    Use 'dnf install ansible-core' or 'pip install tmt[ansible]' to install it.

Yeah, the same message is repeated three times, perhaps we could define it as a constant in `tmt/steps/provision/__init__.py`?

psss (changes requested):
Thanks for the update. Added three comments.

tmt/steps/provision/local.py
happz: Or, we could add it to `tmt.options.show_step_method_hints()`. It's not a step, but it's where hints like this are stored today, so extending this function, calling it with some `ansible` string instead of step, and then raising an exception?

WRT pip vs RPM, all these hints speak about RPMs. I think Python package is also important and we should fix them all. I hope `tmt about` would be able to help with more verbosity in the future, for now I for one would be fine with mentioning just the RPM.

psss: Yeah, sounds good, let's re-use that helper.

tmt.spec
```diff
@@ -64,7 +66,7 @@ Obsoletes:      tmt-provision-container < %{version}-%{release}
 %endif
 Requires:       tmt == %{version}-%{release}
 Requires:       podman
-Requires:       (ansible or ansible-collection-containers-podman)
+Requires:       (ansible-collection-containers-podman or ansible)
```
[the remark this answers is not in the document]
bajertom: Removed

tmt.spec
```diff
@@ -36,6 +37,7 @@ It contains no code, just makes sure the dependencies are installed.}
 The tmt Python module and command line tool implement the test
 metadata specification (L1 and L2) and allows easy test execution.
 
+%pyproject_extras_subpkg -n tmt ansible
```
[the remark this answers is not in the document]
bajertom: Removed

tmt/steps/provision/local.py
bajertom: I've used the helper function. Now looking at it, isn't it enough to use it only in https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/__init__.py#L1300?

The result now is doubled (used tmt with ansible playbook on a container without ansible installed):
```
        how: ansible
        playbook: playbook.yaml
        hint: Seems that ansible is not installed on the test runner. Use 'dnf install ansible-core'.
        hint: Use the 'local' method to execute tests directly on your localhost.
        hint: See 'tmt run provision --help' for all available provision options.
        hint: Seems that ansible is not installed on the test runner. Use 'dnf install ansible-core'.
        hint: Use the 'local' method to execute tests directly on your localhost.
        hint: See 'tmt run provision --help' for all available provision options.
        fail: Command 'sudo -E ansible-playbook -c local -i localhost, /root/tmt/playbook.yaml' returned 1.
    report
        how: display
        summary: no results found
    finish
    
        summary: 0 tasks completed

plan failed

Cause number 1:

    prepare step failed

    Cause number 1:

        Command 'sudo -E ansible-playbook -c local -i localhost, /root/tmt/playbook.yaml' returned 1.

        # stderr (1/1 lines)
        sudo: ansible-playbook: command not found
```

psss: > hint: Use the 'local' method to execute tests directly on your localhost.
>         hint: See 'tmt run provision --help' for all available provision options.

Yes, that would make sense. Less copy & paste and you would get rid of the duplication.

psss: > hint: Use the 'local' method to execute tests directly on your localhost.
>         hint: See 'tmt run provision --help' for all available provision options.

This hint actually is not very related/useful when `ansible-playbook` is missing. What about adding `if how == 'ansible'` as the top level condition in `show_step_method_hints()`?

happz: WRT duplication, yes, it makes sense to work with the hint at in one method family only, i.e. either in (all) `ansible()` methods, or in all `_run_ansible()` methods. Some plugins have their own implementations, so exnahncing all implementations is necessary, but `ansible()` boils down to `_run_ansible()`, so doing it in `ansible()` should be safe.

bajertom: I've made the changes and used "plugin" as a step name even though it is not a step, just to make it go through.

When run on a localhost, the error message appears in exc.stderr (sudo: ansible-playbook: command not found) and exc.message is different (Command 'sudo -E ansible-playbook -c local -i localhost, /home/tom/tmt/playbook.yaml' returned 1.).

When run in container, the exc.stderr is None and the error message appears in exc.message (File 'ansible-playbook' not found.)

tmt/steps/provision/podman.py
```diff
@@ -274,7 +274,6 @@ def _run_ansible(
             *self._ansible_extra_args(extra_args),
             '-c', 'podman', '-i', f'{self.container},', playbook
             ])
-
```
happz, line 277: Several lines removed, unrelated change :)

tmt/steps/provision/local.py
happz: And that basically tells us the check belongs to `_run_ansible()` because its actual form depends on how and what the provisioning plugin does and runs :/ So, if this was on my table, I would sigh and move it out of `ansible()` back to all `_run_ansible()` implementations :/

psss:
Nice, now everything's working as expected:

    hint: Install 'ansible-core' to prepare guests using ansible playbooks.
    fail: File 'ansible-playbook' not found.

tmt/steps/provision/podman.py
psss: Fixed in 78a9ca02e80c1a06b93406367a342e8d83c0a30e.

happz (approved):

psss (approved):

tmt/steps/provision/local.py
psss: Thanks! And sorry for the complications.

psss:
All failures are irrelevant, merging.

PR #3466, opened 2025-01-13
Run the `report` step upon failures as well

### LINKED ISSUE #3401: The `report` step should be executed upon errors
Currently, if a failure happens during the `provision`, `prepare` and `execute` steps, the rest of the steps is skipped, except for `finish` which takes care of guest cleanup. There might be partial test results (if the failure happened in the middle of the `execute` step) but user does not get any because the `report` step is skipped completely.

Let's run the `report` step always (if enabled) and include all completed test results. For tests which were not executed we should probably report `skip` or `error` as the result.

psss:
In order to report at least partial test results, let's execute the `report` step also when the test run is interrupted in the middle of the test execution (for example when the guest becomes unresponsive).

Fix #3401.

Pull Request Checklist

* [x] implement the feature
* [x] include a release note


KwisatzHaderach (approved):
Code looks good, this might actually want a release note though.

happz (approved):
Mostly harmless. What could go wrong.

psss:
> Code looks good, this might actually want a release note though.

Good, point added in c6920cce.

PR #2671, opened 2024-02-08
Support boot.method in mrack

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

PR #3469, opened 2025-01-15
Refactor order of actions in guest reboot implementations

happz:
Things worked, but the code was not easy to understand and follow:

* it wasn't obvious custom reboot `command` and hard reboots are not compatible;
* it wasn't obvious that custom reboot command can come only from reboot request data stored by `tmt-reboot`;
* the ordering of hard, soft and soft-with-custom-command was unclear,
* `command` of `reboot()` was not the same as `command` of `perform_reboot()`;
* the relationship between `reboot()` and `perform_reboot()` was unclear;
* often all known arguments were passed to `super().reboot()` or `perform_reboot()`, which was not incorrect, but harder for humans and linters to reason about;
* docstrings of `reboot()` methods were incomplete or outdated,
* plugins did not mention their special behavior in docstrings of `reboot()` method.

To hopefully improve the situation, the patch makes the following changes:

* uses `@overload` to clearly mark that `hard=True` means no custom reboot command;
* it makes reboot handling in `TestInvocation` code more verbose to make the distinction between hard and soft reboots which should make the custom reboot command clearly attached to soft reboots only;
* all `reboot()` methods now follow the same order of actions: handling hard reboot first, soft reboots second, with less dense code;
* `perform_reboot()` now accepts "action" callback, not "command", and it is responsibility of `reboot()` to feed `perform_reboot()` wit the proper action (running remote command, local command, or calling testcloud API, and so on);
* reduces "blind" passing of known values: `super().reboot(hard=hard)` is correct, but impossible for a type checker to map to the "hard or custom command" split; using `hard=True|False` helps linters narrow the situation;
* cleans up docstrings, makes them as same as possible with plugin-specific additions and notes where necessary;
* and mention reboot support levels in plugin docstrings.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring

### DIFF
--- a/tmt/steps/provision/connect.py
+++ b/tmt/steps/provision/connect.py
@@ -134,12 +134,12 @@ class GuestConnect(tmt.steps.provision.GuestSsh):
             # ignore[union-attr]: mypy still considers `self.hard_reboot` as possibly
             # being `None`, missing the explicit check above.
             return self.perform_reboot(
-                lambda: self._run_guest_command(
-                    self.hard_reboot.to_shell_command()),  # type: ignore[union-attr]
+                lambda: self._run_guest_command(self.hard_reboot.to_shell_command()),  # type: ignore[union-attr]
                 timeout=timeout,
                 tick=tick,
                 tick_increase=tick_increase,
-                fetch_boot_time=False)
+                fetch_boot_time=False,
+            )
 
         if command is not None:
             return super().reboot(
@@ -147,7 +147,8 @@ class GuestConnect(tmt.steps.provision.GuestSsh):
                 command=command,
                 timeout=timeout,
                 tick=tick,
-                tick_increase=tick_increase)
+                tick_increase=tick_increase,
+            )
 
         if self.soft_reboot is not None:
             self.debug(f"Soft reboot using the soft reboot command '{self.soft_reboot}'.")
@@ -155,17 +156,13 @@ class GuestConnect(tmt.steps.provision.GuestSsh):
             # ignore[union-attr]: mypy still considers `self.soft_reboot` as possibly
             # being `None`, missing the explicit check above.
             return self.perform_reboot(
-                lambda: self._run_guest_command(
-                    self.soft_reboot.to_shell_command()),  # type: ignore[union-attr]
+                lambda: self._run_guest_command(self.soft_reboot.to_shell_command()),  # type: ignore[union-attr]
                 timeout=timeout,
                 tick=tick,
-                tick_increase=tick_increase)
+                tick_increase=tick_increase,
+            )
 
-        return super().reboot(
-            hard=False,
-            timeout=timeout,
-            tick=tick,
-            tick_increase=tick_increase)
+        return super().reboot(hard=False, timeout=timeout, tick=tick, tick_increase=tick_increase)
 
     def start(self) -> None:
         """
--- a/tmt/steps/provision/connect.py
+++ b/tmt/steps/provision/connect.py
@@ -96,14 +96,26 @@ class GuestConnect(tmt.steps.provision.GuestSsh):
         """
         Reboot the guest, and wait for the guest to recover.
 
+        .. note::
+
+           Custom reboot command can be used only in combination with a
+           soft reboot. If both ``hard`` and ``command`` are set, a hard
+           reboot will be requested, and ``command`` will be ignored.
+
         :param hard: if set, force the reboot. This may result in a loss of
             data. The default of ``False`` will attempt a graceful reboot.
-        :param command: a command to run on the guest to trigger the reboot.
+
+            Plugin will use :py:attr:`ConnectGuestData.hard_reboot`,
+            set via ``hard-reboot`` key. Unlike ``command``, this command
+            would be executed on the runner, **not** on the guest.
+        :param command: a command to run on the guest to trigger the
+            reboot. If ``hard`` is also set, ``command`` is ignored.
+
             If not set, plugin would try to use
-            :py:attr:`ConnectGuestData.soft_reboot` or
-            :py:attr:`ConnectGuestData.hard_reboot` (``--soft-reboot`` and
-            ``--hard-reboot``, respectively), if specified. Unlike ``command``,
-            these would be executed on the runner, **not** on the guest.
+            :py:attr:`ConnectGuestData.soft_reboot`, set via
+            ``soft-reboot`` key. Unlike ``command``,
+            this command would be executed on the runner, **not** on the
+            guest.
         :param timeout: amount of time in which the guest must become available
             again.
         :param tick: how many seconds to wait between two consecutive attempts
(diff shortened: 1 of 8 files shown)


tmt/steps/provision/mrack.py
```diff
@@ -1363,8 +1372,8 @@ def reboot(
         :returns: ``True`` if the reboot succeeded, ``False`` otherwise.
         """
 
-        if not command and hard:
-            self.debug("Reboot using the reboot command 'bkr system-power --action reboot'.")
+        if hard:
+            self.debug("Hard reboot using the reboot command 'bkr system-power --action reboot'.")
```
skycastlelily, line 1376: According to https://beaker-project.org/docs/man/bkr-system-power.html
```suggestion
            self.debug("Hard reboot using the reboot command 'bkr system-power --action reboot --force'.")
```

tmt/steps/provision/mrack.py
```diff
@@ -1363,8 +1372,8 @@ def reboot(
...
 
             reboot_script = ShellScript(f'bkr system-power --action reboot {self.primary_address}')
```
skycastlelily, line 1378:
```suggestion
            reboot_script = ShellScript(f'bkr system-power --action --force reboot {self.primary_address}')
```

tmt/steps/provision/testcloud.py
```diff
@@ -1135,16 +1135,60 @@ def reboot(
...
+                tick=tick,
+                tick_increase=tick_increase,
+                fetch_boot_time=False,
+            )
+
         if command:
-            return super().reboot(hard=hard, command=command, timeout=timeout)
-        if not self._instance:
+            return super().reboot(
+                command=command, timeout=timeout, tick=tick, tick_increase=tick_increase
```
skycastlelily, line 1178:
```suggestion
                hard=False, command=command, timeout=timeout, tick=tick, tick_increase=tick_increase
```

tmt/steps/provision/connect.py
```diff
@@ -187,6 +208,24 @@ class ProvisionConnect(tmt.steps.provision.ProvisionPlugin[ProvisionConnectData]
...
+    .. code-block:: yaml
+
+        provision:
+          how: connect
+          hard-reboot: virsh reboot my-example-vm
+          soft-reboot: shutdown -r now
+
+    .. warning::
+
+        Both ``hard-reboot`` and ``soft-reboot`` commands are executed
```
skycastlelily, line 227: You sure?^^ the above example "shutdown -r now" seems like a command supposed to run on the guest.

skycastlelily (approved):
Otherwise LGTM^^: I tested most of the plugins,and it works well.

tmt/steps/provision/mrack.py
happz: > Normally this command will refuse to run if another user is using the system. Pass this option to perform the power action anyway.

I don't believe `--force` would make sense: if we do not own the system, we should probably stop using it rather than forcing a reboot.

tmt/steps/provision/mrack.py
happz: See https://github.com/teemtee/tmt/pull/3469/files#r1955955938

tmt/steps/provision/connect.py
happz: Yes, both run on the runner. Updated the example in f2995139

tmt/steps/provision/mrack.py
skycastlelily: > if we do not own the system

This is  mrack (not connect/local) plugin, we would own the system if would us who provision it.
and I thought hard reboot mean we want to reboot the system anyway.

However, I don't have a strong opinion on this, and I vote for what you professional would choose^^

happz: I think the difference is the key: `--force` does not mean just "force (hard) reboot of a machine", but "force (hard) reboot of a machine *used by someone else but you*", and I don't believe we want that amendment clause.

tmt/steps/provision/mrack.py
```diff
@@ -1373,11 +1382,11 @@ def reboot(
...
                 tick=tick,
                 tick_increase=tick_increase,
-                hard=True,
+                fetch_boot_time=False,
             )
 
         return super().reboot(
-            hard=hard,
+            hard=False,
             command=command,
```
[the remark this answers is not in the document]
happz: Hmm, why aren't we having a problem with `reboot` so far? First time hearing about `rhts-reboot`, and I think we could switch to it for Beaker, but is it that necessary over `reboot` or e.g. `shutdown -h now`?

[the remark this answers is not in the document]
skycastlelily: > Hmm, why aren't we having a problem with reboot so far? 

I guess that is  partially because "restart-with-reboot" is the way for users to require a reboot in a test and that key have hard as default(https://tmt.readthedocs.io/en/stable/spec/tests.html#restart).Btw, tmt cli doesn't  accept  a reboot command,and here is my code to implement it ,feel free to take it, or ask me to open a new mr:)

```
diff --git a/tmt/steps/__init__.py b/tmt/steps/__init__.py
index c6af4989..3f54e0c4 100644
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -48,14 +48,7 @@ from tmt.container import (
     option_to_key,
 )
 from tmt.options import option, show_step_method_hints
-from tmt.utils import (
-    DEFAULT_NAME,
-    Environment,
-    EnvVarValue,
-    GeneralError,
-    Path,
-    RunError,
-)
+from tmt.utils import DEFAULT_NAME, Command, Environment, EnvVarValue, GeneralError, Path, RunError
 from tmt.utils.templates import render_template

 if TYPE_CHECKING:
@@ -2032,6 +2025,12 @@ class Reboot(Action):
         @option(
             '--hard', is_flag=True, help='Hard reboot of the machine. Unsaved data may be lost.'
         )
+        @option(
+            '--command',
+            type=str,
+            default='reboot',
+            help='A command to run on the guest to trigger the reboot.',
+        )
         def reboot(context: 'tmt.cli.Context', **kwargs: Any) -> None:
             """
             Reboot the guest.
@@ -2065,7 +2064,7 @@ class Reboot(Action):
         assert hasattr(self.parent, 'plan')
         assert self.parent.plan is not None
         for guest in self.parent.plan.provision.guests():
-            guest.reboot(hard=self.opt('hard'))
+            guest.reboot(hard=self.opt('hard'), command=Command(self.opt('command')))
         self.info('reboot', 'Reboot finished', color='yellow')
```

> but is it that necessary over reboot or e.g. shutdown -h now?

yeah, it's necessary , "The rhts-reboot command also sets “BootNext” to the entry added by Anaconda rather than booting from the network. "   
Here is the rhts-reboot on efi beaker server:

```
[root@hpe-apollo-cn99xx-15-vm-17 ~]# rhts-reboot
<13>Feb 16 23:38:50 root: efibootmgr -n 0008
BootNext: 0008
BootCurrent: 0008
Timeout: 3 seconds
BootOrder: 0003,0004,0005,0006,0002,0008,0000,0001,0007
Boot0000* UiApp	FvVol(64074afe-340a-4be6-94ba-91b5b4d0f71e)/FvFile(462caa21-7614-4503-836e-8ab6f4662331)
Boot0001* UEFI Misc Device	VenHw(93e34c7e-b50e-11df-9223-2443dfd72085,00){auto_created_boot_option}
Boot0002* UEFI Misc Device 2	PciRoot(0x0)/Pci(0x1,0x3)/Pci(0x0,0x0){auto_created_boot_option}
Boot0003* UEFI PXEv4 (MAC:5254002A1511)	PciRoot(0x0)/Pci(0x1,0x0)/Pci(0x0,0x0)/MAC(5254002a1511,1)/IPv4(0.0.0.0,0,DHCP,0.0.0.0,0.0.0.0,0.0.0.0){auto_created_boot_option}
Boot0004* UEFI PXEv6 (MAC:5254002A1511)	PciRoot(0x0)/Pci(0x1,0x0)/Pci(0x0,0x0)/MAC(5254002a1511,1)/IPv6([::],0,Static,[::],[::],64){auto_created_boot_option}
Boot0005* UEFI HTTPv4 (MAC:5254002A1511)	PciRoot(0x0)/Pci(0x1,0x0)/Pci(0x0,0x0)/MAC(5254002a1511,1)/IPv4(0.0.0.0,0,DHCP,0.0.0.0,0.0.0.0,0.0.0.0)/Uri(){auto_created_boot_option}
Boot0006* UEFI HTTPv6 (MAC:5254002A1511)	PciRoot(0x0)/Pci(0x1,0x0)/Pci(0x0,0x0)/MAC(5254002a1511,1)/IPv6([::],0,Static,[::],[::],64)/Uri(){auto_created_boot_option}
Boot0007* EFI Internal Shell	FvVol(64074afe-340a-4be6-94ba-91b5b4d0f71e)/FvFile(7c04a583-9e3e-4f1c-ad65-e05268d0b4d1)
Boot0008* Fedora	HD(1,GPT,6341dbb8-4e20-4bfb-8377-4739a19cd807,0x800,0x12c000)/\EFI\fedora\shimaa64.efi
<13>Feb 16 23:38:50 root: NEXTBOOT_VALID_TIME is zero. BootNext setting will persist.
```

As you can see,before reboot the system,rhts-reboot will run " efibootmgr -n  0008" ,which will set Bootnext to fedora entry, without that, the system will boot from 0003.

Feel free to check it  by yourself:
tmt doesn't support [bootmethod](https://github.com/teemtee/tmt/pull/2671),here is a workaround:find an efi system on through beaker web, use that system name as hostname in the plan.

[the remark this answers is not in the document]
happz: > > Hmm, why aren't we having a problem with reboot so far?
> 
> I guess that is  partially because "restart-with-reboot" is the way for users to require a reboot in a test and that key have hard as default(https://tmt.readthedocs.io/en/stable/spec/tests.html#restart).

That is a very special case of a reboot, and yes, it's using a hard reboot. The assumption here is that when the test crashed, it's very hard to reason about the state of the guest, and a hard reboot seems like the option with the least potential issues.

> Btw, tmt cli doesn't  accept  a reboot command,and here is my code to implement it ,feel free to take it, or ask me to open a new Mr:)

Looks more like a dedicated MR ;) BTW what is the use case for this new key?

> > but is it that necessary over reboot or e.g. shutdown -h now?
> 
> yeah, it's necessary , "The rhts-reboot command also sets “BootNext” to the entry added by Anaconda rather than booting from the network. " Here is the rhts-reboot on efi beaker server:

https://beaker-project.org/docs/architecture-guide/provisioning-process.html#boot-order was helpful.

So we should change the `beaker` plugin to use `rhts-reboot` instead of plain `reboot` (unless a special command has been set by the test, of course). That sounds like a new PR, I would like to keep this one as a refactoring rather than adding new stuff as well.

therazix (approved):

lukaszachy:
Failures in tests using  virtual-centos-stream-10 and one container with epel-9 repodata -> lets merge

PR #3470, opened 2025-01-16
Add `tmt about`, command showing things about tmt itself

happz:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] include a release note

### DIFF
--- a/tests/core/about/test.sh
+++ b/tests/core/about/test.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+EXPECTED_PLUGIN_LIST=\
+"export.plan:
+  - dict
+  - json
+  - template
+  - yaml
+export.story:
+  - dict
+  - json
+  - template
+  - rst
+  - yaml
+export.test:
+  - dict
+  - json
+  - nitrate
+  - polarion
+  - template
+  - yaml
+package_managers:
+  - apk
+  - apt
+  - dnf
+  - dnf5
+  - yum
+  - rpm-ostree
+plan_shapers:
+  - max-tests
+step.discover:
+  - fmf
+  - shell
+step.execute:
+  - tmt
+  - upgrade
+step.finish:
+  - ansible
+  - shell
+step.prepare:
+  - install
+  - ansible
+  - shell
+  - feature
+step.provision:
+  - artemis
+  - virtual.testcloud
+  - bootc
+  - connect
+  - local
+  - beaker
+  - container
+step.report:
+  - display
+  - html
+  - junit
+  - polarion
+  - reportportal
+test.check:
+  - avc
+  - dmesg
+  - watchdog
+test.framework:
+  - beakerlib
+  - shell"
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "set -o pipefail"
+    rlPhaseEnd
+
+    rlPhaseStartTest "List plugins as human-readable output"
+        rlRun -s "tmt about plugin ls"
+
+        # Just a brief check, hard to test expected formatting
+        rlAssertGrep "Export plugins for story" $rlRun_LOG
+        rlAssertGrep "Export plugins for plan" $rlRun_LOG
+        rlAssertGrep "Export plugins for test" $rlRun_LOG
+        rlAssertGrep "Test check plugins" $rlRun_LOG
+        rlAssertGrep "Test framework plugins" $rlRun_LOG
+        rlAssertGrep "Plan shapers" $rlRun_LOG
+        rlAssertGrep "Package manager plugins" $rlRun_LOG
+        rlAssertGrep "Discover step plugins" $rlRun_LOG
+        rlAssertGrep "Provision step plugins" $rlRun_LOG
+        rlAssertGrep "Prepare step plugins" $rlRun_LOG
+        rlAssertGrep "Execute step plugins" $rlRun_LOG
+        rlAssertGrep "Finish step plugins" $rlRun_LOG
+        rlAssertGrep "Report step plugins" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartSetup "List plugins as YAML"
+        rlRun -s "tmt about plugins ls --how yaml"
+
+        # Hello, traveller. Is this test failing for you? Then you
+        # probably added new plugin, or changed existing ones, e.g.
+        # by renaming their registry or moving plugins around. The
+        # test is a sanity one, making sure tmt discovers all it can,
+        # updating $EXPECTED_PLUGIN_LIST should turn the tide of
+        # misfortune.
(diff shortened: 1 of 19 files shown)


psss:
This sounds like a very nice idea! Are there any limitations about the usage? I tried the new command within the venv but I get no output.

happz:
> This sounds like a very nice idea! Are there any limitations about the usage? I tried the new command within the venv but I get no output.

```
$ tmt about plugins --how rest
Export plugins for story

* `dict`
* `json`
* `template`
* `rst`
* `yaml`

----

Export plugins for plan

* `dict`
* `json`
* `template`
* `yaml`

----
...

$ tmt about plugins --how json | jq .
{
  "export.story": [
    "dict",
    "json",
    "template",
    "rst",
    "yaml"
  ],
  "export.plan": [
    "dict",
    "json",
    "template",
    "yaml"
  ],
  "export.test": [
    "dict",
    ...
```

The current limitation is, it's just listing plugins, nothing more. I hope to get it to share more info, maybe render plugin docs, specs docs, or explanation of various errors, just like `rustc --explain` does:

```
$ PAGER= rustc --explain E0277
You tried to use a type which doesn't implement some trait in a place which expected that trait.

Erroneous code example:

// here we declare the Foo trait with a bar method
trait Foo {
    fn bar(&self);
}

// we now declare a function which takes an object implementing the Foo trait
fn some_func<T: Foo>(foo: T) {
    foo.bar();
}

fn main() {
    // we now call the method with the i32 type, which doesn't implement
    // the Foo trait
    some_func(5i32); // error: the trait bound `i32 : Foo` is not satisfied
}

In order to fix this error, verify that the type you're using does implement the trait. Example:

trait Foo {
    fn bar(&self);
}

// we implement the trait on the i32 type
impl Foo for i32 {
    fn bar(&self) {}
}

fn some_func<T: Foo>(foo: T) {
    foo.bar(); // we can now use this method since i32 implements the
               // Foo trait
}

fn main() {
    some_func(5i32); // ok!
}

...
```

thrix (dismissed):
Unfortunately does not work well for me:

* `tmt about` shows nothing, let's print maybe the help?

* tracebacks with plaing `tmt about plugins ls`

```
$ tmt about plugins ls                                                                                                                                                                                    
Unhandled ReST node '<title_reference>dict</title_reference>'.                                                                                                                                                  (shell returned 2)
```

happz:
> Unfortunately does not work well for me:
> 
> * `tmt about` shows nothing, let's print maybe the help?
> * tracebacks with plaing `tmt about plugins ls`
> 
> ```
> $ tmt about plugins ls                                                                                                                                                                                    
> Unhandled ReST node '<title_reference>dict</title_reference>'.                                                                                                                                                  (shell returned 2)
> ```

Both issues should be fixed now.

therazix (approved):
Works well for me.

skycastlelily (approved):
LGTM^^,and confirmed:

    Unfortunately does not work well for me:

        tmt about shows nothing, let's print maybe the help?
        tracebacks with plaing tmt about plugins ls

    $ tmt about plugins ls                                                                                                                                                                                    
    Unhandled ReST node '<title_reference>dict</title_reference>'.                                                                                                                                                  (shell returned 2)

> Both issues should be fixed now.

tmt/cli/about.py
```diff
@@ -0,0 +1,107 @@
...
+    if context.invoked_subcommand is None:
+        echo(context.get_help(), color=context.color)
+
+
+def _render_plugins_list_rest() -> str:
+    registry_intro_map: dict[str, str] = {
+        r'export\.([a-z]+)': 'Export plugins for {{ MATCH.group(1).lower() }}',
+        r'test.check': 'Test check plugins',
+        r'test.framework': 'Test framework plugins',
+        r'package_manager': 'Package manager plugins',
```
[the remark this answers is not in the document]
happz: Correct, but it would not hurt the workflow, as the pattern would still match - it does not need to be a full match.

psss (changes requested):
Could we include at least a short release note? The command is very simple, true, but still, it's a new `tmt` subcommand and users deserve to learn about it, hm?

tmt/plugins/__init__.py
```diff
@@ -437,3 +445,40 @@ def __call__(self, logger: Logger) -> ModuleT:
...
+    from tmt.checks import _CHECK_PLUGIN_REGISTRY
+    from tmt.frameworks import _FRAMEWORK_PLUGIN_REGISTRY
+    from tmt.package_managers import _PACKAGE_MANAGER_PLUGIN_REGISTRY
+    from tmt.plugins.plan_shapers import _PLAN_SHAPER_PLUGIN_REGISTRY
+    from tmt.steps.discover import DiscoverPlugin
+    from tmt.steps.execute import ExecutePlugin
+    from tmt.steps.finish import FinishPlugin
+    from tmt.steps.prepare import PreparePlugin
+    from tmt.steps.provision import ProvisionPlugin
+    from tmt.steps.report import ReportPlugin
```
psss, line 464: Seems the `prepare.feature` registry is missing?

happz:
> Could we include at least a short release note? The command is very simple, true, but still, it's a new `tmt` subcommand and users deserve to learn about it, hm?

True, but I hesitate. `tmt about` is indeed new, but only knows how to list discovered plugins, which is not that much. I hope to add more functionality users more visible and helpful for users, maybe a release note would be better then?

tmt/plugins/__init__.py
happz: True, added in 484329b5.

psss:
> True, but I hesitate. `tmt about` is indeed new, but only knows how to list discovered plugins, which is not that much. I hope to add more functionality users more visible and helpful for users, maybe a release note would be better then?

Ok, understood. Let's make it more visible later.

tmt/plugins/__init__.py
psss: Thanks!

psss (approved):

tests/core/about/test.sh
```diff
@@ -0,0 +1,64 @@
...
+        rlAssertGrep "Execute step plugins" $rlRun_LOG
+        rlAssertGrep "Finish step plugins" $rlRun_LOG
+        rlAssertGrep "Report step plugins" $rlRun_LOG
+    rlPhaseEnd
+
+    # Hello, traveller. Is this test failing for you? Then you
+    # probably added new plugin, or changed existing ones, e.g.
+    # by renaming their registry or moving plugins around. The
+    # test is a sanity one, making sure tmt discovers all it can,
+    # updating $EXPECTED_PLUGIN_LIST should turn the tide of
```
martinhoyer, line 52: Looks like bootc is missing
```
- package_managers: apk apt dnf dnf5 rpm-ostree yum
+ package_managers: apk apt bootc dnf dnf5 rpm-ostree yum
```

martinhoyer: ff440aaf7eb84228bd7b8a4169bb71319b52d4d7

PR #3472, opened 2025-01-17
Describe features covered by the minimal package

psss:
Let's be a bit more specific about the features which are covered by the minimal ``tmt`` package.

Pull Request Checklist

* [x] write the documentation

### DIFF
--- a/stories/install.fmf
+++ b/stories/install.fmf
@@ -15,10 +15,13 @@ link:
         * execute tests on the :ref:`/plugins/provision/local`
           host or :ref:`/plugins/provision/connect` to a remote
           guest using ssh
-        * prepare guest using :ref:`/plugins/prepare/shell` and
-          :ref:`/plugins/prepare/install` plugins
+        * prepare guest using the
+          :ref:`/plugins/prepare/ansible`,
+          :ref:`/plugins/prepare/feature`,
+          :ref:`/plugins/prepare/install` and
+          :ref:`/plugins/prepare/shell` plugins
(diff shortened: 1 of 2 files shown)


thrix:
@psss I belive we discussed that we will drop this and go the `Recommends` way?

psss:
I'd suggest to update the description anyway, to be more specific what the `tmt` package actually provides. I'll update the description to be aligned with the #3465 change. Let's include this simple docs update in `1.42`, hm?

psss:
Description updated and linked from the Guide in 78ae66be.

stories/install.fmf
```diff
@@ -6,6 +6,23 @@ link:
...
+        The minimal package provides basic functionality which
+        covers mainly these areas:
+
+        * explore and lint existing tests, plans and stories
+        * create new tests, plans and stories from templates
+        * execute tests on the :ref:`/plugins/provision/local`
+          host or :ref:`/plugins/provision/connect` to a remote
+          guest using ssh
+        * prepare guest using the
+          :ref:`/plugins/prepare/ansible`,
```
lukaszachy, line 19: How should we document that ansible is possible only if weak dependencies are also installed?

thrix (approved):
LGTM

stories/install.fmf
thrix: @psss do you want to address this?

psss: Let's explicitly mention it. Added in 09cbc5c0.

lukaszachy (approved):

PR #3475, opened 2025-01-21
Define the `releaser` role

psss:
In order to get the releases more predictable and smooth, let's put together a couple of recommendations for the person who accepts the release lead role.

Pull Request Checklist

* [x] write the documentation

### DIFF
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -766,3 +766,24 @@ fix needs to be released before the regular schedule:
 * Tag the commit and publish the release in the same way as for
   regular release
 * Create a pull request with the hotfix release notes changes
+
+
+Releaser
+------------------------------------------------------------------
+
+Taking care of a new ``tmt`` release is not just about performing
+the final steps described above. In this role you should shepherd
+the issues and pull requests like sheep so that they make it to
+the ``main`` branch by the proposed deadline. Here's a couple of
+recommendations which could help you to make the release process
+smooth and timely:
+
+* continually watch the issues & pull requests and gently push
+  them forward if any of them seems to get stuck
+* bring attention especially to those with the high priority, the
+  ``must-have`` issues and pull requests should be finished
+  ideally one week before the release deadline
+* regularly check the pull request progress and highlight those
+  which are waiting for feedback on the review sessions
+* if there is anything not clear and needs discussion bring it to
+  the chat or raise the topic on the weekly sessions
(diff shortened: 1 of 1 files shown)


martinhoyer (approved):
Much needed, thanks! 

I like the sheep analogy.  
How about "managing" instead of "taking care of"?  
Do you think it's worth mentioning contacting assignees directly on chat(s), to learn about the current and planned progress and potentially raising it for a reassignment (or something like that)

docs/contribute.rst
```diff
@@ -766,3 +766,24 @@ fix needs to be released before the regular schedule:
...
+the final steps described above. In this role you should shepherd
+the issues and pull requests like sheep so that they make it to
+the ``main`` branch by the proposed deadline. Here's a couple of
+recommendations which could help you to make the release process
+smooth and timely:
+
+* continually watch the issues & pull requests and gently push
+  them forward if any of them seems to get stuck
+* bring attention especially to those with the high priority, the
+  ``must-have`` issues and pull requests should be finished
```
happz, line 784: I'm not sure we have a literal `must-have`, the label says `priority | must` and I couldn't find `must-have` in tmt repo.

docs/contribute.rst
```diff
@@ -766,3 +766,24 @@ fix needs to be released before the regular schedule:
...
+  ideally one week before the release deadline
+* regularly check the pull request progress and highlight those
+  which are waiting for feedback on the review sessions
+* if there is anything not clear and needs discussion bring it to
+  the chat or raise the topic on the weekly sessions
```
thrix, line 789: What about checking for ready merge requests which are not in included in release?

docs/contribute.rst
psss: Ah, good point, fixed in e3fc5429.

psss:
> I like the sheep analogy.

Good! :+1: 

> How about "managing" instead of "taking care of"?

I wanted to keep the wording close to the sheep analogy, so would rather keep this if it's ok?

> Do you think it's worth mentioning contacting assignees directly on chat(s), to learn about the current and planned progress and potentially raising it for a reassignment (or something like that)

Good idea, added one more bullet in e3fc5429. Please, check if the wording sounds ok.

thrix (approved):
LGTM

therazix (approved):

psss:
Documentation change, no need to wait for tests.

PR #3477, opened 2025-01-22
Fix reportportal schema for traceback and log size limits

happz:
Pull Request Checklist

* [x] implement the feature


thrix:
@happz can you share more light why size limits should not be an integer?

happz:
> @happz can you share more light why size limits should not be an integer?

User may also provide a string, a number with a unit: `--traceback-limit=1MB`. The default value itself is not an integer, `50 kB`.

psss (approved):

KwisatzHaderach (approved):

PR #3480, opened 2025-01-24
Check disk space for `/tests/provision/bootc`

psss:
The test started to fail recently with the following error:

    qemu-img: error while writing at byte 6421348352:
    No space left on device

Let's investigate the problem and adjust hardware requirements accordingly if needed.

### DIFF
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -12,6 +12,7 @@ rlJournalStart
         rlRun "mkdir -p /var/tmp/tmt"
         rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
         rlRun "pushd data"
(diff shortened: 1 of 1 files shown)


thrix:
@psss I believe this should be marked as draft, as does not seem to be anything we want to get merged right?

psss:
I think we should merge it. Here's a recent [failure](https://artifacts.osci.redhat.com/testing-farm/11ecadb8-5b78-444c-9849-513293ec42ea/work-bootce7mxm21z/plans/provision/bootc/execute/data/guest/default-0/tests/provision/bootc-1/output.txt). Seems they happen randomly and would be good to have a better idea about what disk is used on the system. Perhaps we should add more of these for each subtest. Will update the pull request.

happz (approved):

therazix (approved):

psss:
This one is a no-brainer and already passed before. Merging right away.

PR #3483, opened 2025-01-26
Keep container uid the same as host to allow ops on workdir root

carlosrodfern:
Volume mounts on the container are by default read only for non-root users.
This prevents the necessary access for rootless container. The change in this
commit makes the container uid be the same as the host uid so that the user
under which podman runs, which owns the workdir root, is the same user uid on
the container, so the permissions also match.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

Related: #3252

### DIFF
--- a/tests/provision/facts/test-guest.sh
+++ b/tests/provision/facts/test-guest.sh
@@ -44,7 +44,7 @@ rlJournalStart
             build_container_image "fedora/41/upstream\:latest"
 
             provision_options="--image localhost/tmt/container/test/fedora/41/upstream:latest"
-            bfu_provision_options="$provision_options --user=nobody"
+            bfu_provision_options="--image localhost/tmt/container/test/fedora/41/unprivileged:latest --user=fedora"
 
             arch="$(arch)"
             distro="Fedora Linux 41 (Container Image)"
--- a/tests/provision/facts/test-guest.sh
+++ b/tests/provision/facts/test-guest.sh
@@ -42,6 +42,7 @@ rlJournalStart
 
         elif [ "$PROVISION_HOW" = "container" ]; then
             build_container_image "fedora/41/upstream\:latest"
+            build_container_image "fedora/41/unprivileged\:latest"
 
             provision_options="--image localhost/tmt/container/test/fedora/41/upstream:latest"
             bfu_provision_options="--image localhost/tmt/container/test/fedora/41/unprivileged:latest --user=fedora"
--- a/tmt/steps/provision/podman.py
+++ b/tmt/steps/provision/podman.py
@@ -9,6 +9,8 @@ import tmt.steps
 import tmt.steps.provision
 import tmt.utils
 from tmt.container import container, field
+from tmt.options import show_step_method_hints
+from tmt.package_managers import FileSystemPath
 from tmt.steps.provision import GuestCapability
 from tmt.utils import Command, OnProcessStartCallback, Path, ShellScript, retry
 
@@ -215,6 +217,9 @@ class GuestContainer(tmt.Guest):
                 self._logger,
             )
 
+        # Mount the whole plan directory in the container
+        workdir = self.parent.plan.workdir
+
         self.verbose('name', self.container, 'green')
 
         additional_args = []
@@ -233,13 +238,25 @@ class GuestContainer(tmt.Guest):
                 '--name',
                 self.container,
                 '-v',
-                f'{self.workdir_root}:{self.workdir_root}:z',
+                f'{workdir}:{workdir}:z',
                 '-itd',
                 '--user',
                 self.user,
                 self.image,
             )
         )
+
+    def setup(self) -> None:
+        if not self.facts.is_superuser:
+            self.package_manager.install(FileSystemPath('/usr/bin/setfacl'))
+            self.execute(
+                ShellScript(
+                    f"""
+                    sudo chmod o+rwX {self.workdir_root};
+                    sudo setfacl -d -m o:rwX {self.workdir_root}
+                    """
+                )
+            )
 
     def reboot(
         self,
--- a/tmt/steps/provision/podman.py
+++ b/tmt/steps/provision/podman.py
@@ -252,6 +252,7 @@ class GuestContainer(tmt.Guest):
             self.execute(
                 ShellScript(
                     f"""
+                    sudo mkdir -p {self.workdir_root};
                     sudo chmod o+rwX {self.workdir_root};
                     sudo setfacl -d -m o:rwX {self.workdir_root}
                     """
--- a/tmt/steps/provision/podman.py
+++ b/tmt/steps/provision/podman.py
@@ -247,6 +247,7 @@ class GuestContainer(tmt.Guest):
         )
 
     def setup(self) -> None:
+        super().setup()
         if not self.facts.is_superuser:
             self.package_manager.install(FileSystemPath('/usr/bin/setfacl'))
             self.execute(
--- a/tmt/steps/provision/podman.py
+++ b/tmt/steps/provision/podman.py
@@ -215,9 +215,6 @@ class GuestContainer(tmt.Guest):
                 self._logger,
             )
 
-        # Mount the whole plan directory in the container
-        workdir = self.parent.plan.workdir
-
         self.verbose('name', self.container, 'green')
 
         additional_args = []
@@ -231,10 +228,12 @@ class GuestContainer(tmt.Guest):
             Command(
                 'run',
                 *additional_args,
+                '--userns',
+                'keep-id',
                 '--name',
                 self.container,
                 '-v',
-                f'{workdir}:{workdir}:z',
+                f'{self.workdir_root}:{self.workdir_root}:z',
                 '-itd',
                 '--user',
                 self.user,
--- a/tmt/steps/provision/podman.py
+++ b/tmt/steps/provision/podman.py
@@ -9,7 +9,6 @@ import tmt.steps
 import tmt.steps.provision
 import tmt.utils
 from tmt.container import container, field
-from tmt.options import show_step_method_hints
 from tmt.package_managers import FileSystemPath
 from tmt.steps.provision import GuestCapability
 from tmt.utils import Command, OnProcessStartCallback, Path, ShellScript, retry


tmt/steps/provision/podman.py
```diff
@@ -218,6 +219,15 @@ def start(self) -> None:
             self.image
             ))
 
+    def setup(self) -> None:
```
happz, line 222: `super().setup()` should be called as well.

tmt/steps/provision/podman.py
```diff
@@ -218,6 +219,15 @@ def start(self) -> None:
...
+        if not self.facts.is_superuser:
+            self.package_manager.install(FileSystemPath('/usr/bin/setfacl'))
+            self.execute(ShellScript(
+                f"""
+                    sudo chmod -R o+rwX {self.workdir_root};
```
happz, line 227: Wouldn't this also change read-only files to writable? Is that an expected, or even desired course of action?

thrix:
moved the milestone, seems this is not making the release today

tmt/steps/provision/podman.py
carlosrodfern: That's a really good point. Because the directory is mounted in the host, it will make `/var/tmp/tmt` on the host world writable.

The other solution I can think of is a deeper change, basically, making the logic around `TEST_WRAPPER_TEMPLATE` different for container, leveraging the fact that the container directory is mounted.

happz: > The other solution I can think of is a deeper change, basically, making the logic around `TEST_WRAPPER_TEMPLATE` different for container, leveraging the fact that the container directory is mounted.

That we can do as well, the wrapper does not have to be the same for everything. It's a Jinja template, two shell files, we can do whatever we need to do to make all things work - permissions, TTY, duration, signals, and whatever else is waiting there to bite us. I tried to document and improve wrappers a bit, if you wish to give this avenue a try, please, start on top of this PR.

carlosrodfern: So the logic creating the pid files would need to _not_ be executed on the container, but on the host, assuming it would be available to the container, and assuming flock works on the mount that podman does. That is necessary so it can be placed in the `/var/tmp/tmt` directory as the other PR is trying to do.

Basically, it is a change not just to the `TEST_WRAPPER_TEMPLATE` but to the code using it as well, adding different execution paths for containers. I can look into it.

happz: Pid file exists so `tmt-reboot`, called by a different tmt instance, could safely notify about upcoming reboot, and kill the test process our tmt manages. This presumes both the test and `tmt-reboot` run on the same guest, sharing it, like a baremetal machine or a VM from a cloud - thinking about it now, it does make much less sense in the context of containers.

Seems like a topic for tomorrow's tmt hacking meeting, make we could introduce some feature flag signaling to tmt when keeping pidfile makes sense, maybe provided by the guest class implementation.

thrix: added to hacking

carlosrodfern: @happz , did you have a chance to look into it?

happz: I didn't, let's give it a try now.

There are several `provision` plugins, what does the existence of the PID file and the out-of-session `tmt-reboot` mean for them? The PID file exists so `tmt-reboot` can update the running test with reboot command and timeout, and then kill the test process spawned by a tmt managing the test. `tmt-reboot` runs on the guest, not on the runner, and the killed process was running on the guest, not on the runner. And the more I think about it, the more I'm getting lost in various combinations :/

Let's focus on the workdir permissions and the original issue, "Containers that mount the workdir but run with non-root users are not able to operate on the workdir because the mounted volume appears as read only." Isn't it possible to fix this by some careful `podman run` setup? Like mapping ownership, permissions, etc?

SSH-capable guests do have something similar, but they seem to get away with just read+executable permissions:

```python
    def setup(self) -> None:
        if self.is_dry_run:
            return
        if not self.facts.is_superuser and self.become:
            self.package_manager.install(FileSystemPath('/usr/bin/setfacl'))
            workdir_root = effective_workdir_root()
            self.execute(
                ShellScript(
                    f"""
                    mkdir -p {workdir_root};
                    setfacl -d -m o:rX {workdir_root}
                    """
                )
            )
```

carlosrodfern: I did try to devise a solution using uidmaps but no luck. I'll give `--userns` a try.

carlosrodfern: @happz I just pushed a solution using `--userns`. Let me know your thoughts.

tmt/steps/provision/podman.py
```diff
@@ -232,10 +229,12 @@ def start(self) -> None:
             Command(
                 'run',
                 *additional_args,
+                '--userns',
+                'keep-id',
                 '--name',
                 self.container,
                 '-v',
-                f'{workdir}:{workdir}:z',
+                f'{self.workdir_root}:{self.workdir_root}:z',
```
happz, line 237: Why the change from `self.parent.plan.workdir` to `self.workdir_root`? I don't believe the yare the same thing, `workdir_root` will contain files completely unrelated to the current run.

carlosrodfern: Mainly because the new pid code is trying to create the pid in the `/var/tmp/tmt` dir, which doesn't match any of the permission if the mount is /var/tmp/tmt/run-001/etc...`.

carlosrodfern: Hmmm, I believe I can still do it with the `workdir` if I bring the `setup(..)` method back without the recursive change. It would be a little bit more over the place since it is modifying the parent dir created by the mount with some permissions but with a different set of permission for the subdirectories as inherited by the userns and the mount itself, but it might work.

carlosrodfern: OK, I added a new commit bringing `setup(..)` back, but without the recursive chmod. It works as well on the pid change branch. This approach may feel a little bit less consistent in the sense that the `workdir_root` has now permissions to allow modifications in it outside the plan execution, but the plan execution itself (a subdir) is mounted with a different set of permissions as set by volume mount with the keep-id userns option.

@happz, Let me know, I can revert that last commit.

happz: Sure. BTW is there a test that would demonstrate the problem, a test we could watch turn green once the fix is correct? :)

carlosrodfern: `provision/facts/guest` is failing because the preparation step to install setfacl is doing a `sudo` but the user is `nobody`, so it is asking for the sudo password in that step.

```
            provision_options="--image fedora:39"
            bfu_provision_options="$provision_options --user=nobody"
```
Should the test be changed to use a similar unprivileged container image with a passwordless sudo fedora user? I can make that change if that's OK.


The `provision/user` one is failing because the `workdir` is actually not a subdirectory of `self.workdir_root`, so I pushed a change to add `mkdir -p {self.workdir_root}`, and it is going to be needed for the pid change as well.

happz: The failing test, `/tests/provision/facts/guest`, seems less than perfect because it does not have the most fitting image. That should get better with https://github.com/teemtee/tmt/pull/3518, after a rebase & rerun of tests we should get a better picture.

tmt/steps/provision/podman.py
happz: Unresolving, `super().setup()` is not called.

martinhoyer:
@carlosrodfern Hey, looks like there is one unresolved [comment](https://github.com/teemtee/tmt/pull/3483#discussion_r1987103298), and well, will need a rebase. Would you have time to look into it, or should we try to find a volunteer to do the "finishing touches"? :)

carlosrodfern:
> @carlosrodfern Hey, looks like there is one unresolved [comment](https://github.com/teemtee/tmt/pull/3483#discussion_r1987103298), and well, will need a rebase. Would you have time to look into it, or should we try to find a volunteer to do the "finishing touches"? :)

Just pushed the fix :+1: 
Thx!

happz:
@carlosrodfern please, rebase when you find some spare time. I briefly checked the test results, it seems like there is actually a bug in the test, the following patch should fix it: `*:unprivileged` image needs to be used for testing of unprivileged actions, plus the username is incorrect.

```diff
diff --git a/tests/provision/facts/test-guest.sh b/tests/provision/facts/test-guest.sh
index 8f621b50..92931fc9 100755
--- a/tests/provision/facts/test-guest.sh
+++ b/tests/provision/facts/test-guest.sh
@@ -44,7 +44,7 @@ rlJournalStart
             build_container_image "fedora/41/upstream\:latest"

             provision_options="--image localhost/tmt/container/test/fedora/41/upstream:latest"
-            bfu_provision_options="$provision_options --user=nobody"
+            bfu_provision_options="--image localhost/tmt/container/test/fedora/41/unprivileged:latest --user=fedora"

             arch="$(arch)"
             distro="Fedora Linux 41 (Container Image)"
```

carlosrodfern:
> @carlosrodfern please, rebase when you find some spare time. I briefly checked the test results, it seems like there is actually a bug in the test, the following patch should fix it: `*:unprivileged` image needs to be used for testing of unprivileged actions, plus the username is incorrect.
> 
> ```diff
> diff --git a/tests/provision/facts/test-guest.sh b/tests/provision/facts/test-guest.sh
> index 8f621b50..92931fc9 100755
> --- a/tests/provision/facts/test-guest.sh
> +++ b/tests/provision/facts/test-guest.sh
> @@ -44,7 +44,7 @@ rlJournalStart
>              build_container_image "fedora/41/upstream\:latest"
> 
>              provision_options="--image localhost/tmt/container/test/fedora/41/upstream:latest"
> -            bfu_provision_options="$provision_options --user=nobody"
> +            bfu_provision_options="--image localhost/tmt/container/test/fedora/41/unprivileged:latest --user=fedora"
> 
>              arch="$(arch)"
>              distro="Fedora Linux 41 (Container Image)"
> ```

@happz , I just did a rebase and I still see the `bfu_provision_options="$provision_options --user=nobody"` line. Do you want me to provide the fix in this PR?

carlosrodfern:
> > @carlosrodfern please, rebase when you find some spare time. I briefly checked the test results, it seems like there is actually a bug in the test, the following patch should fix it: `*:unprivileged` image needs to be used for testing of unprivileged actions, plus the username is incorrect.
> > ```diff
> > diff --git a/tests/provision/facts/test-guest.sh b/tests/provision/facts/test-guest.sh
> > index 8f621b50..92931fc9 100755
> > --- a/tests/provision/facts/test-guest.sh
> > +++ b/tests/provision/facts/test-guest.sh
> > @@ -44,7 +44,7 @@ rlJournalStart
> >              build_container_image "fedora/41/upstream\:latest"
> > 
> >              provision_options="--image localhost/tmt/container/test/fedora/41/upstream:latest"
> > -            bfu_provision_options="$provision_options --user=nobody"
> > +            bfu_provision_options="--image localhost/tmt/container/test/fedora/41/unprivileged:latest --user=fedora"
> > 
> >              arch="$(arch)"
> >              distro="Fedora Linux 41 (Container Image)"
> > ```
> 
> @happz , I just did a rebase and I still see the `bfu_provision_options="$provision_options --user=nobody"` line. Do you want me to provide the fix in this PR?

OK, I provided the test fix in the PR. Let me know if you want me to remove that commit.

happz:
@carlosrodfern yeah, unfortunately github does not let me update your PR, otherwise I'd do it already.

tests/provision/facts/test-guest.sh
```diff
@@ -44,7 +44,7 @@ rlJournalStart
             build_container_image "fedora/41/upstream\:latest"
```
happz, line 44:
```suggestion
            build_container_image "fedora/41/upstream\:latest"
            build_container_image "fedora/41/unprivileged\:latest"
```

tmt/steps/provision/podman.py
```diff
@@ -231,6 +232,8 @@ def start(self) -> None:
             Command(
                 'run',
                 *additional_args,
+                '--userns',
+                'keep-id',
```
psss, line 236: Sorry, I'm not very familiar with the `--userns` option. Just want clarify if/why this is needed: Trying with a simple plan:

```yaml
provision:
    how: container
    image: localhost/tmt/container/test/fedora/41/unprivileged
    user: fedora

execute:
    how: tmt
    script: "getfacl /tmp/root && id && cat /etc/passwd"
```

And the following command:

    TMT_WORKDIR_ROOT=/tmp/root tmt run -vvv

When the option is present I can see user `psss` included in the `/etc/passwd` file:

```
nobody:x:65534:65534:Kernel Overflow User:/:/usr/sbin/nologin
fedora:x:1000:1000::/home/fedora:/bin/bash
psss:*:12559:12559::/:/bin/sh
```

But when `user: psss` is used in the plan I get this:

```
fail: Command 'podman run --network tmt-008-network --userns keep-id --name tmt-008-GQRwnRBq -v /tmp/root/run-008/plan:/tmp/root/run-008/plan:z -itd --user psss localhost/tmt/container/test/fedora/41/unprivileged' returned 126.
Error: unable to find user psss: no matching entries in passwd file
```

So, how does the mapping of local user `psss` to the container `/etc/passwd` help here?

carlosrodfern: The `--userns keep-id` allows the mounted directory permission to work for the local host user and the container user by having the same uid, which is especially necessary for hardened systems with `umask 0027` where the files are not world-readable by default. 

For example, on the host:
```
$ mkdir /tmp/test-0
$ umask 0027
$ touch /tmp/test-0/test-0.txt
```
... and on the container:

```
$ podman run -it -v /tmp/test-0:/test:z --user fedora localhost/tmt/tests/container/fedora/rawhide/unprivileged /bin/bash

(container)$ ls -lash /test
0 drwxr-xr-x. 2 root root 60 Mar 20 03:12 .
0 dr-xr-xr-x. 1 root root 14 Mar 20 03:11 ..
0 -rw-r-----. 1 root root  0 Mar 20 03:12 test-0.txt
(container)$ cat /test/test-0.txt
cat: /test/test-0.txt: Permission denied
```

However, if I use the `keep-id`:
```
$ podman run -it -v /tmp/test-0:/test:z --userns keep-id --user fedora localhost/tmt/tests/container/fedora/rawhide/unprivileged /bin/bash
(container)$ ls -lash /test
total 0
0 drwxr-xr-x. 2 fedora fedora 60 Mar 20 03:12 .
0 dr-xr-xr-x. 1 root   root   14 Mar 20 03:15 ..
0 -rw-r-----. 1 fedora fedora  0 Mar 20 03:12 test-0.txt

(container) $ cat /test/test-0.txt
# no issues

(container) $ umask 0027
(container) $ touch /test/test-1.txt
```

...and on the host:

```
$ ls -lash /tmp/test-0/
total 0
0 drwxr-xr-x.  2 user user  80 Mar 19 20:16 .
0 drwxrwxrwt. 27 root   root   680 Mar 19 20:11 ..
0 -rw-r-----.  1 user user   0 Mar 19 20:12 test-0.txt
0 -rw-r-----.  1 user user   0 Mar 19 20:16 test-1.txt
```


In your specific example, I can't see how the `cat /etc/passwd` contains the `psss` host user. It should simply contain the `fedora` user with the same uid as the host user the container is running under. That seems like a bug to me.

psss: Thank you for the detailed explanation. The thing is that we're usually running tests under a different user. So in my scenario: I'm running `tmt run` for the above-mentioned plan under user `psss` on my laptop. This uid gets synced to the `/etc/passwd` but the `/tmp/root` directory still has `root` as the owner. Probably because `fedora` user is used to connect to the container?

```
$ ls -l /tmp/root
total 0
drwxr-xr-t. 1 root root 8 Mar 20 12:52 run-011
```

martinhoyer: > I can't see how the cat /etc/passwd contains the psss host user.

Wouldn't that be due to `--passwd` is true by default? `podman run --help`:
` --passwd                  add entries to /etc/passwd and /etc/group (default true)`

martinhoyer (approved):
not my area of expertise, but lgtm. Thanks!

tmt/steps/provision/podman.py
carlosrodfern: > Btw, we're using `sudo` commands below to set the proper permissions (so for this `--userns` is not needed) and allow others to write to the directory and newly created files & directories, so inside the container our test user `fedora` should be able to successfully write to the workdir. Or am I still missing something essential?

In this PR in the above comments we ran into a couple of issues in containers:

* Sometimes `/var/tmp/tmt` is not used at all, not even created, when there are not actual outputs (I believe it is the case in inline scripts), but they were sent to /tmp. However, we still need `/var/tmp/tmt` created with the right permissions because of the up-coming pid feature.
* The `/var/tmp/tmt` is created in containers as a side effect of a mount that includes it as a parent. The mounts are specific to the plan, not to the entire `/var/tmp/tmt` directory. As result, the parent `/var/tmp/tmt` is created by default with the read-only permissions owned by root. In order to allow for writing to `/var/tmp/tmt` itself (like for the pid feature), we need to force the permissions. The permissions config works for the pid feature but doesn't for the rest of other things down the path that were already present before these commands are invoked as result of the volume mount. We can't just do a recursive chmod because that changes the perms on the host (it is volume mount), so the `keep-id` is the most secure option here.

In both cases, the solution should work for hardened (especially the umask 0027) and non-hardened containers and hosts.

martinhoyer: Just following up on the discussion. The `--passwd` seems to be doing what you mentioned @psss?
```
$ podman run --rm --userns=keep-id fedora:latest cat /etc/passwd 
root:x:0:0:Super User:/root:/bin/bash
bin:x:1:1:bin:/bin:/usr/sbin/nologin
<snip>
nobody:x:65534:65534:Kernel Overflow User:/:/usr/sbin/nologin
tss:x:59:59:Account used for TPM access:/:/usr/sbin/nologin
mhoyer:*:1000:1000:Martin Hoyer:/:/bin/sh

$ podman run --rm --passwd=false --userns=keep-id fedora:latest cat /etc/passwd 
root:x:0:0:Super User:/root:/bin/bash
bin:x:1:1:bin:/bin:/usr/sbin/nologin
<snip>
nobody:x:65534:65534:Kernel Overflow User:/:/usr/sbin/nologin
tss:x:59:59:Account used for TPM access:/:/usr/sbin/nologin
```

psss:
The pidfile fix [breaks virtual provision](https://github.com/teemtee/tmt/pull/3252#issuecomment-2753457463), this will need more time, moving to backlog.

Issue #3238, opened 2024-09-25
Include all discovered tests in the `results.yaml` file


thrix:
Currently, when tmt execution dies in the middle of the plan execution, the results.yml:

* contains only tests which were executed
* all other tests (running or skipped) are missing

This causes issues to the user as it is not possible to easily for consumers of the results.yml (like Testing Farm) to provide a reasonable output.

Blocker for VirtQE and KernelQE.

🔽 [TFT-2847](https://issues.redhat.com/browse/TFT-2847)
🔽 [TT-342](https://issues.redhat.com/browse/TT-342)

happz:
Adding an `error` result for the test that "killed" the `execute` should be well within our reach.

It was also discussed (briefly, IIRC, feel free to correct me) that tmt could emit placeholder results, with let's say `idle` outcome, for collected and soon-to-be-executed tests, converting them to actual results as it goes. To me, this sounds like a good idea, and it would be consistent with the `error` result mentioned above.

lukaszachy:
Just a note to the 'idle' state: With 'results:custom' the count of expected vs real results might differ as additional tests might be reported (new test names be the name of the test with results: custom appended by reported name).
This is no blocker by any means, just something to document so consumers of results.yaml are not surprised that the number of results after execute step is higher than before.

psss:
See also related issue #3401 which suggests the `report` step to be always executed upon failures.

PR #3485, opened 2025-01-26
Add pending state for results

### LINKED ISSUE #3304: Generate results.yaml in progress
Testing Farm uses the `results.yaml` to find and show test outputs in the result viewer. We want to show them as soon as possible, but we can show the log only after the plan is finished due to tmt generating the file at the end.

thrix:
Resolves #3304

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/discover/__init__.py
+++ b/tmt/steps/discover/__init__.py
@@ -1,5 +1,6 @@
 import dataclasses
 from collections.abc import Iterator
+from itertools import zip_longest
 from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
 
 import click
@@ -17,6 +18,7 @@ import tmt.steps
 import tmt.utils
 from tmt.options import option
 from tmt.plugins import PluginRegistry
+from tmt.result import Result, ResultOutcome
 from tmt.steps import Action
 from tmt.utils import GeneralError, Path, field, key_to_option
 
@@ -414,6 +416,15 @@ class Discover(tmt.steps.Step):
                                 test.name == result.name and
                                 test.serial_number == result.serial_number):
                             self._failed_tests[test_phase].append(test)
+
+        # Initialize results with pending outcome if no results for it available
+        results = self.plan.execute.results()
+        for index, (test, result) in enumerate(zip_longest(self.tests(), results)):
+            if result is None:
+                results.insert(index, Result(name=test.name, result=ResultOutcome.PENDING))
+
+        # save the results
+        self.plan.execute._save_results(results)
 
         # Give a summary, update status and save
         self.summary()
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -1129,9 +1129,11 @@ class Execute(tmt.steps.Step):
         """ Execute tests """
         super().go(force=force)
 
-        # Clean up possible old results
+        # Clean up possible old results, by switching them to pending state
         if force:
-            self._results.clear()
+            for index, result in enumerate(self._results):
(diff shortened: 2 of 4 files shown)


tmt/steps/execute/__init__.py
```diff
@@ -1129,9 +1129,11 @@ def go(self, force: bool = False) -> None:
         """ Execute tests """
         super().go(force=force)
 
-        # Clean up possible old results
+        # Clean up possible old results, by switching them to pending state
         if force:
-            self._results.clear()
+            for index, result in enumerate(self._results):
+                if result.result != ResultOutcome.PENDING:
+                    self._results[index] = Result(name=result.name, result=ResultOutcome.PENDING)
```
happz, line 1136: I'd say this should recreate the list from scratch, i.e. not overwriting the existing results. Let's say there was a test with custom results, creating 100 of results for that single test. Before running the test for the first time, there would be a single `PENDING` placeholder for the test; after the operation above, `results.yaml` may contain 100 results for that test, each with `PENDING` results. And since tmt does not control creation of these results, the test may easily decide to not produce several of them this time, and therefore that particular result should never appear in our `results.yaml`. Dropping all results except the single `PENDING` one per test seems like better option to me.

tmt/steps/discover/__init__.py
```diff
@@ -415,6 +417,15 @@ def go(self, force: bool = False) -> None:
                                 test.serial_number == result.serial_number):
                             self._failed_tests[test_phase].append(test)
 
+        # Initialize results with pending outcome if no results for it available
+        results = self.plan.execute.results()
+        for index, (test, result) in enumerate(zip_longest(self.tests(), results)):
```
happz, line 422: I believe it won't be as easy as `zip()`. Results may be listed in an arbitrary order - yes, results are added once their tests finish, but I wouldn't rely on it, and there is certainly no code enforcing the order.

Check the `Execute.results_for_tests()` which lists results and test pairs, you should be interested in creating new results for pairs where the test is set and the result is `None`, that's a test that has not run yet.

tmt/steps/execute/__init__.py
```diff
@@ -1201,7 +1203,11 @@ def go(self, force: bool = False) -> None:
         # which would make results appear several times. Instead, we can reach
         # into the original plugin, and use it as a singleton "entry point" to
         # access all collected `_results`.
-        self._results += execute_phases[0].results()
+        #
+        # Go through all results and replace those which were pending
+        for index, result in enumerate(execute_phases[0].results()):
+            if self._results[index].result == ResultOutcome.PENDING:
+                self._results[index] = result
```
happz, line 1210: Hmmm, this is weird. I think it will hit custom results again: for a single test, there may be 100 results in `execute_phases[0].results()` but not in `self._results. They would then land all in the very same index.

I believe we need a more robust approach here, based on matching the result's test serial numbers, constructing the list from two sources, `execute_phases[0]` and `self._results. I'd try walking `self._results`, and find out whether there are any results in `execute_phases[0].results()` for the same test: if there are, add them to future `self._results`, if there are not, copy the pending one. Eventually, I'd expect to be left with consumed `execute_phase[0].results()`, because all of those are valid results, and possibly not consumed `self._results` because some were moved to the new list, and some didn't run and remain `PENDING`. Those I would add to the new list as well.

psss:
This pull request covers https://github.com/teemtee/tmt/issues/3238 as well, right?

therazix:
Transfered to #3559.

PR #3489, opened 2025-01-27
Use custom option type for workdir root to deliver `Path` instances

happz:
Instead of strings needing to convert to paths "manually", we can ask to Click to use a custom type that would do the conversion for us, delivering `Path` instances to our commands.

Pull Request Checklist

* [x] implement the feature


psss (approved):
Looks good!

thrix:
Moved to 1.43, tests need a look

PR #3493, opened 2025-01-28
Execute must not use discover phases disabled by their `when`

### LINKED ISSUE #3492: Execute tries to run tests from 'turned off' discover config
When there is a discover config turned off by using `when` key, `execute` still tries to run tests defined in that config and fails as no such test exists according to the discover. 

Thanks to @happz for initial investigation and finding likely cause at https://github.com/teemtee/tmt/blob/384bb5682984c5f4ae54e25fb62a6abfc04403a3/tmt/steps/execute/__init__.py#L1173 

Reproducer plan:
```
execute:
  how: tmt

provision:
  how: container

discover:
    - how: shell
      tests:
      - name: nope
        test: echo
      when: flag == on
    - how: shell
      tests:
      - name: yes
        test: echo
      when: flag is not defined
```

happz:
This seems to be an omission from when `when` was added: it is evaluated by tmt, not by fmf, disabled phases are still present in the tree of steps and phases. Most places already ignore them, but `execute` did not. `execute` runs its single phase with all `discover` phases in sequence, and must make sure it does not try to run the disabled ones.

Fixes #3492.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/steps/when/data/plan.fmf
+++ b/tests/steps/when/data/plan.fmf
@@ -1,32 +1,52 @@
 execute:
   how: tmt
 
-discover:
-  how: shell
-  tests:
-    - name: /
-      test: touch $TMT_PLAN_DATA/when-in-discover
-  when: trigger == commit
-
 provision:
   how: local
 
-prepare:
-  - how: shell
-    script: touch $TMT_PLAN_DATA/when-prepare-always
-  - how: shell
-    script: touch $TMT_PLAN_DATA/when-prepare-fedora
-    when: distro == fedora
+/sanity:
+  discover:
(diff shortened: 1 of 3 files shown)


KwisatzHaderach (approved):

lukaszachy (approved):
Should we add tests? 
I verified it fixes the problem

lukaszachy (approved):

happz:
> Should we add tests? I verified it fixes the problem

Yep, it's there now. It took me some time to modify and extend the existing one.

KwisatzHaderach (approved):

PR #3495, opened 2025-01-29
Move hints to their own place, with API and unicorns

happz:
I wanted to change a couple of things about "hints" tmt shows to users when when a package is missing and functionality is limited:

* Static, stored in a single function. That made them detached from their origin, and 3rd party plugins would have no chance to add their own hints.
* A bit confined text of hints did not cover all possible venues. PyPI installation was ignored by some, other hints spoke just about using `pip install`.
* Hints are interesting and useful, but visible only when error strikes.

The patch turns the function into a "registry", a simple dictionary storing them. Plugins and tmt code in general can "register" their hints, and a nice tools are available for showing them.

Hints wow have IDs, and there are dedicated IDs for step-specific (``report``, ...) and plugin-specific (``report/foo``) hints, allowing tmt core to print them when step or plugins crashes on import.

Plugin-specific hints are now rendered both in their CLI and HTML documentation.

In the future, I would like to provide hints not just as "a package foo is missing, install it" guide, but also explaining various errors and issues tmt would report, e.g. pairing them with exceptions, "To learn more, run `tmt about E1234`".

Pull Request Checklist

* [x] implement the feature
* [ ] extend the test coverage - I plan to connect these to `tmt about`, together with listing all discovered plugins it should give me enough tools to test the features

### DIFF
--- a/tmt/utils/hints.py
+++ b/tmt/utils/hints.py
@@ -1,0 +1,181 @@
+"""
+Hints for users when facing installation-related issues.
+
+Plugins, steps, and tmt code in general can register hints to be shown
+to user when an important (or optional, but interesting) package is not
+available.
+
+Hints are shown when importing plugins fails, and rendered as part of
+both their CLI help and HTML documentation.
+"""
+
+# NOTE (happz): in my plan, this module would be an unfinished, staging
+# area for hints; eventually, I would like them to be managed under the
+# umbrella of `tmt about` subcommand. `print_hint()` would still exist,
+# but `tmt about` would be responsible for handling hints, therefore the
+# code below may change, the concept should not. And hints would cover
+# wider area, e.g. describing common errors and issues, not just when
+# a package is missing. They would be coupled with exceptions tmt
+# raises, providing more info on command-line and in HTML docs.
+
+import textwrap
+from collections.abc import Iterator
+from typing import Optional
+
+import tmt.log
+import tmt.utils
+import tmt.utils.rest
+
+HINTS: dict[str, str] = {
+    # Hints must be dedented & stripped of leading/trailing white space.
+    # For hints registered by plugins, this is done by `register_hint()`.
+    _hint_id: textwrap.dedent(_hint).strip()
+    for _hint_id, _hint in {
+        'provision': """
+            You can use the ``local`` method to execute tests directly on your localhost.
+
+            See ``tmt run provision --help`` for all available ``provision`` options.
+            """,
+        "report": """
+            You can use the ``display`` method to show test results on the terminal.
+
+            See ``tmt run report --help`` for all available report options.
+            """,
+        'ansible-not-available': """
+            Make sure ``ansible-playbook`` is installed, it is required for preparing guests using
+            Ansible playbooks.
+
+            To quickly test ``ansible-playbook`` presence, you can try running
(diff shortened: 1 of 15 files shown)


happz:
Adding `blocked`, I think I can do something with `dedent()` calls.

tmt/utils/hints.py
```diff
@@ -0,0 +1,181 @@
...
+
+    :param id_: step name for step-specific hints,
+        ``<step name>/< plugin name>`` for plugin-specific hints,
+        or an arbitrary string.
+    :param hint: a hint to register.
+    """
+
+    if id_ in HINTS:
+        raise tmt.utils.GeneralError(
+            "Registering hint '{id_}' collides with an already registered hint."
```
therazix, line 78:
```suggestion
            f"Registering hint '{id_}' collides with an already registered hint."
```

tmt/utils/hints.py
```diff
@@ -0,0 +1,181 @@
...
+
+        if hint is None:
+            return None
+
+        return _render_single_hint(hint)
+
+    hint = HINTS.get(id_)
+
+    if hint is None:
+        raise tmt.utils.GeneralError("Could not find hint '{id_}'.")
```
therazix, line 116:
```suggestion
        raise tmt.utils.GeneralError(f"Could not find hint '{id_}'.")
```

tmt/utils/hints.py
```diff
@@ -0,0 +1,181 @@
...
+                yield _render_single_hint(hint)
+
+        return '\n'.join(_render_optional_hints())
+
+    def _render_mandatory_hints() -> Iterator[str]:
+        for id_ in ids:
+            hint = HINTS.get(id_)
+
+            if hint is None:
+                raise tmt.utils.GeneralError("Could not find hint '{id_}'.")
```
therazix, line 157:
```suggestion
                raise tmt.utils.GeneralError(f"Could not find hint '{id_}'.")
```

therazix (approved):
A couple of `f` prefixes are missing for f-strings. Otherwise, looks good.

tmt/utils/hints.py
```diff
@@ -0,0 +1,181 @@
...
+    }.items()
+}
+
+
+def register_hint(id_: str, hint: str) -> None:
+    """
+    Register a hint for users.
+
+    :param id_: step name for step-specific hints,
+        ``<step name>/< plugin name>`` for plugin-specific hints,
```
therazix, line 71: There is an extra space.
```suggestion
        ``<step name>/<plugin name>`` for plugin-specific hints,
```

tmt/utils/hints.py
happz: Interestingly, I believed we have a `ruff` rule for that. We don't, it's in a preview mode.

Addressed in 3ae00ab3.

tmt/utils/hints.py
happz: Addressed in 3ae00ab3.

tmt/utils/hints.py
happz: Addressed in 3ae00ab3.

tmt/utils/hints.py
happz: Addressed in 3ae00ab3.

PR #1801, opened 2023-01-17
Modify rpm-ostree options for new parameter

BeeGrech:
rpm-ostree behavior has been modified, and implemented an assume yes param ie -y, implementing here as in later releases this will be required for invoking non interactive

PR #3497, opened 2025-01-29
Fix package manager tests assertions

thrix:
These were introduced in #1801 and merged by an error.


happz (approved):

psss (approved):

thrix:
Beaker machines causing stability issues I think, trying to address in #3499

PR #3498, opened 2025-01-29
Dmesg and AVC check add note explaining `skip` outcome

happz:
Pull Request Checklist

* [x] implement the feature


tmt/checks/dmesg.py
```diff
@@ -187,7 +187,10 @@ def before_test(
             environment: Optional[tmt.utils.Environment] = None,
             logger: tmt.log.Logger) -> list[CheckResult]:
         if not invocation.guest.facts.has_capability(GuestCapability.SYSLOG_ACTION_READ_ALL):
-            return [CheckResult(name='dmesg', result=ResultOutcome.SKIP)]
+            return [CheckResult(
+                name='dmesg',
+                result=ResultOutcome.SKIP,
+                note=['fetching of buffer content was not allowed'])]
```
thrix, line 193: I am wondering, would it not make more sense to link from the note to docs giving more explanation here?
As a user I will have no godd idea what this actually means. Docs could explain more?

tmt/checks/dmesg.py
```diff
@@ -202,10 +205,16 @@ def after_test(
...
 
         if not invocation.is_guest_healthy:
-            return [CheckResult(name='dmesg', result=ResultOutcome.SKIP)]
+            return [CheckResult(
+                name='dmesg',
+                result=ResultOutcome.SKIP,
+                note=['guest was not healthy'])]
```
thrix, line 217: Also this one, I had to look at code to understand what does it mean the guest is not healthy, currently that means that restart or hard reboot was requested. With this note, it would be hard to guess what this actually means.

tmt/checks/dmesg.py
happz: I agree, it would be nice. I will try tonimprove it - check out my PR about “hints”, I tried to improve this kind of messages, I’d say it’s related, and with `tmt about` I’d like to build a corpus of hints and helpful messages, exceptions, etc.

therazix (approved):

happz:
It works, but it's ugly, because it needs to be rendered better by different presentation layers, from `display` to `html` and TF, and there's a ton of work to make that happen **and** make it look nice, to improve things.

martinhoyer (approved):

martinhoyer (approved):
Very nice

psss:
@happz, is this one still blocked?

happz:
> @happz, is this one still blocked?

No, it is not anymore.

tmt/checks/dmesg.py
happz: Addressed, the text is now much richer.

tmt/checks/dmesg.py
happz: Addressed, the text is now much richer.

tmt/checks/dmesg.py
```diff
@@ -133,7 +134,17 @@ def _save_dmesg(
...
 
 
-@provides_check('dmesg')
+@provides_check(
+    'dmesg',
+    hints={
+        'fetching-skipped': """
+            Saving of the kernel ring buffer was skipped because of missing privileges.
+
+            The account, used by tmt, lacks necessary privileges. See ``man 2 syslog``
```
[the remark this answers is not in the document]
happz: This comment does not seem to me as having any relation to the code it comments on.

PR #3506, opened 2025-02-04
Support `epel-10` in the `feature` prepare plugin

lukaszachy:
I don't see epel-next listed on https://docs.fedoraproject.org/en-US/epel/getting-started/#_release_package_permalinks so only rhel-9/centos-stream-9 have epel-next parts enabled.

Pull Request Checklist

* [x] implement the feature


psss (approved):
Looks good, thanks!

therazix (approved):

PR #3513, opened 2025-02-06
Enable the `crb` repo when enabling `epel`

psss:
The `crb` repository is explicitly included in the official set of steps needed to enable the epel repository:

 * https://docs.fedoraproject.org/en-US/epel/getting-started/

Without it some packages, including `tmt` itself, cannot be installed because of missing dependencies.

Also added `ansible` to the `ansible-lint` check dependencies otherwise it would complain about unknown `community.general` collection.

### DIFF
--- a/tests/prepare/feature/epel/data/plans.fmf
+++ b/tests/prepare/feature/epel/data/plans.fmf
@@ -31,6 +31,17 @@ execute:
         environment+:
             STATE: disabled
 
+/flac:
+    summary: Install flac
+    description: Check that crb repo is enabled
+    prepare:
+      - how: feature
+        epel: enabled
+      - how: install
+        package: flac
+    execute:
+        script: flac --help
+
 /profile:
     discover:
       how: shell
--- a/tmt/steps/prepare/feature/epel-enable.yaml
+++ b/tmt/steps/prepare/feature/epel-enable.yaml
@@ -87,6 +87,12 @@
         - ansible_distribution == "CentOS"
         - ansible_distribution_major_version | int >= 8
       block:
+        - name: Ensure the crb repository is enabled
+          community.general.dnf_config_manager:
+            name: crb
+            state: enabled
+          when: ansible_distribution_major_version | int >= 9
+
         - name: Install package 'epel-release'
           ansible.builtin.dnf:
             name: epel-release
--- a/tmt/steps/prepare/feature/epel-enable.yaml
+++ b/tmt/steps/prepare/feature/epel-enable.yaml
@@ -64,6 +64,7 @@
 
         - name: Enable the crb repository
           ansible.builtin.command: crb enable
+          register: output
           changed_when: output.rc != 0
 
     - name: Enable EPEL repos on CentOS 7
@@ -124,4 +125,5 @@
 
         - name: Enable the crb repository
           ansible.builtin.command: crb enable
+          register: output
           changed_when: output.rc != 0
--- a/tmt/steps/prepare/feature/epel-enable.yaml
+++ b/tmt/steps/prepare/feature/epel-enable.yaml
@@ -62,6 +62,10 @@
           # Enable for Stream 10 once epel-next is available
           when: ansible_distribution_major_version | int == 9
 
+        - name: Enable the crb repository
+          ansible.builtin.command: crb enable
+          changed_when: output.rc != 0
+
     - name: Enable EPEL repos on CentOS 7
       when:
         - ansible_distribution == "CentOS"
@@ -87,12 +91,6 @@
         - ansible_distribution == "CentOS"
         - ansible_distribution_major_version | int >= 8
       block:
-        - name: Ensure the crb repository is enabled
-          community.general.dnf_config_manager:
-            name: crb
-            state: enabled
-          when: ansible_distribution_major_version | int >= 9
-
         - name: Install package 'epel-release'
           ansible.builtin.dnf:
             name: epel-release
@@ -123,3 +121,7 @@
           # EPEL Next is available for CentOS Stream 9
           # Enable for Stream 10 once epel-next is available
           when: ansible_distribution_major_version | int == 9
+
+        - name: Enable the crb repository
+          ansible.builtin.command: crb enable
+          changed_when: output.rc != 0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -137,6 +137,8 @@ repos:
           - --exclude=examples/redis/ansible/tasks/redis.yml
           # redis.yml example requires posix module to be installed
           # in order to be parsed by ansible-lint
+        additional_dependencies:
+          - "ansible"
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
     rev: v0.9.8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -137,8 +137,6 @@ repos:
           - --exclude=examples/redis/ansible/tasks/redis.yml
           # redis.yml example requires posix module to be installed
           # in order to be parsed by ansible-lint
-        additional_dependencies:
-          - "ansible"
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
     rev: v0.9.8
--- a/tests/prepare/feature/epel/test.sh
+++ b/tests/prepare/feature/epel/test.sh
@@ -29,6 +29,10 @@ rlJournalStart
         rlPhaseStartTest "Disable EPEL on $image"
             rlRun -s "tmt -vvv run -a plan --name '/epel/disabled' provision --how container --image $image"
         rlPhaseEnd
+
+        if is_centos_9; then
+            rlRun -s "tmt -vvv run -a plan --name '/flac' provision --how container --image $image"
+        fi
     done
 
     # Environment profiles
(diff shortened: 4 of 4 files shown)


lukaszachy:
But the 'crb' is a valid name only for Centos-stream. For RHEL it is different name and differs between subscribed machine (`codeready-builder-for-rhel-9-$(arch)-rpms`) and private images (usually `rhel-CRB`). 
Shoudln't this be fixed as dependency on `feature: crb` instead?

psss:
Hm, on `centos-stream-9` the job fails with this:

    ERROR! couldn't resolve module/action 'community.general.dnf_config_manager'.
    This often indicates a misspelling, missing collection, or incorrect module path.

I thought that the `community.general` collection is supposed to be available in Testing Farm:

 * https://docs.testing-farm.io/Testing%20Farm/0.1/test-runner.html#_supported_ansible_collections
 * https://gitlab.com/testing-farm/gluetool-modules/-/blob/main/container/requirements.yml

@thrix, @happz, @janhavlin, any idea why it's missing?

janhavlin (approved):

happz:
> Hm, on `centos-stream-9` the job fails with this:
> 
> ```
> ERROR! couldn't resolve module/action 'community.general.dnf_config_manager'.
> This often indicates a misspelling, missing collection, or incorrect module path.
> ```
> 
> I thought that the `community.general` collection is supposed to be available in Testing Farm:
> 
> * https://docs.testing-farm.io/Testing%20Farm/0.1/test-runner.html#_supported_ansible_collections
> * https://gitlab.com/testing-farm/gluetool-modules/-/blob/main/container/requirements.yml
> 
> @thrix, @happz, @janhavlin, any idea why it's missing?

I filed TFT-3140, this seems like a TF bug to me, although triggered by tmt.

lukaszachy:
Lets document that `community.general.dnf_config_manager` is required by `feature:epel` (https://tmt.readthedocs.io/en/stable/plugins/prepare.html#feature)

happz:
> Lets document that `community.general.dnf_config_manager` is required by `feature:epel` (https://tmt.readthedocs.io/en/stable/plugins/prepare.html#feature)

Filed https://github.com/teemtee/tmt/issues/3537 - I for one wouldn't block on it, as it would be nice if it would be based on improved documentation of `prepare/feature` plugins rather than added now and leaving me with a refactoring of documentation texts.

martinhoyer (approved):
Hmm, this one is tricky.
Not that I'm suggesting using it, but was `ansible.builtin.yum_repository` considered?  
IIUIC, It works with just `.repo` files, while `dnf_config_manager` uses, well, dnf config manager. Are there any scenarios where dnf would be installed, while dnf-plugins-core is not? (not that it's important, but working with only .repo files would also make the same work on el8)

I've tried to figure out how to cover RHEL and others which might not have the same/unique name of the crb repo(s).
Some sort of best effort matching of name doesn't seem to be very pretty, but looking at c9s and RHEL .repo files, I think we should be able to match the correct repo(s), by looking at `CRB/x86_64/os/` (with ansible facts replacing arch).

...plus `/CRB/x86_64/debug/tree/` `/CRB/source/tree/`

EDIT: right, the above is more for #3513 discussion. Sorry.

Oh and shouldn't we remove the epel-next stuff since epel10, c10s is out?

psss:
> Filed #3537 - I for one wouldn't block on it, as it would be nice if it would be based on improved documentation of `prepare/feature` plugins rather than added now and leaving me with a refactoring of documentation texts.

The thing is that tests would be failing for `centos-stream-9` until TFT-3140 is fixed. So it does not make sense to merge this now. @thrix, can we move TFT-3140 forward?

psss:
I believe this one should cover #3626, for the `tmt try` use case it should work even without the necessary TF change.

happz:
> I believe this one should cover #3626, for the `tmt try` use case it should work even without the necessary TF change.

Yet it feels like stopping somewhere in the middle: should we also have `crb` feature? `prepare -h feature --crb=enabled` should be a thing.

psss:
> should we also have `crb` feature? `prepare -h feature --crb=enabled` should be a thing.

Yeap, agreed, there should be a `crb` feature plugin. Haven't found an existing issue so filed here: #3650

martinhoyer:
@psss fyi, installing `epel-release` rpm includes `/usr/bin/crb` executable script :tada: see below. Works on RHEL too.  
it's just `# crb enable` and voila.


```
#!/usr/bin/bash
#
#  Enable / Disable / Status the codeready-builder(CRB) or equivalent repo.
#
# Copyright (C) 2024
# SPDX-License-Identifier:      GPL-2.0-only
#
# Authors:
#     Troy Dawson <tdawson@redhat.com>
#


###############
# Show help
###############
usage() {
  echo "Usage `basename $0` [ enable | disable | status ] " >&2
  echo >&2
  echo "Enable, Disable, or give the status of the CodeReady Builder (CRB) repo" >&2
  echo >&2
  echo "Options:" >&2
  echo "  enable" >&2
  echo "    Enable the CRB repo" >&2
  echo "  disable" >&2
  echo "    Disable the CRB repo" >&2
  echo "  status" >&2
  echo "    Show if rhe CRB repo is enabled or disabled" >&2
  echo "  help" >&2
  echo "    Show this options menu" >&2
  echo >&2
  popd &>/dev/null
  exit 1
}


# Show the status
show_status()(
    # Determine if a version of CRB is enabled
    crb_repo=$(dnf repolist | grep -i -e crb -e powertools -e codeready | grep -v -i -e debug -e source -e eus -e virt -e rhui | awk '{print $1}')
    if [ "${crb_repo}" == "" ] ; then
        echo "CRB repo is disabled"
    else
        echo "CRB repo is enabled and named: ${crb_repo}"
    fi
)

# Enable or Disable the repo
enable_disable_repo(){
    command="$1"
    # Determine which repo we need to change
    crb_repo=$(dnf repolist --all | grep -i -e crb -e powertools -e codeready | grep -v -i -e debug -e source -e eus -e virt -e rhui | awk '{print $1}')

    # Determine if we are on RHEL or not, then do the correct steps
    source /etc/os-release
    if [[ "${NAME}" =~ "Red Hat" ]] ; then
        subscription-manager repos ${command} ${crb_repo}
    else
        # Determine if dnf-command(config-manager) is installed
        if ! rpm -q --whatprovides 'dnf-command(config-manager)' > /dev/null 2>&1 ; then
            echo "Error: Please run: dnf install 'dnf-command(config-manager)'" >&2
            echo "       before trying to enable/disable the CRB repo." >&2
            echo "Aborting"'!' >&2
            exit 1
        fi
        # Everything else uses dnf config-manager
        dnf config-manager ${command} ${crb_repo}
    fi
}

###############
# Get our arguments
###############
key="$1"
case $key in
enable | --enable )
    echo "Enabling CRB repo"
    enable_disable_repo --enable
    show_status
;;
disable | --disable )
    echo "Disabling CRB repo"
    enable_disable_repo --disable
    show_status
;;
status | --status )
    show_status
;;
* )
    usage
    exit 2
;;
esac

exit 0
```

psss:
Very nice! 👍

tmt/steps/prepare/feature/epel-enable.yaml
```diff
@@ -62,6 +62,10 @@
           # Enable for Stream 10 once epel-next is available
           when: ansible_distribution_major_version | int == 9
 
+        - name: Enable the crb repository
+          ansible.builtin.command: crb enable
+          changed_when: output.rc != 0
```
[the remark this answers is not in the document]
happz: Yep, absolutely, `register` is missing.

martinhoyer (approved):
@psss FYI https://src.fedoraproject.org/rpms/epel-release/pull-request/37

tmt/steps/prepare/feature/epel-enable.yaml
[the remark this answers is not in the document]
psss: Fixed in 826bbfd0.

thrix (approved):
LGTM

therazix (approved):

tests/prepare/feature/epel/data/plans.fmf
```diff
@@ -31,6 +31,17 @@ execute:
...
             STATE: disabled
 
+/flac:
+    summary: Install flac
+    description: Check that crb repo is enabled
+    prepare:
+      - how: feature
+        epel: enabled
+      - how: install
+        package: flac
```
martinhoyer, line 41: `dummy-package-canary` is in epel

martinhoyer (approved):
Not sure why installing `flac` is being tested :) Surely there is a small

tests/prepare/feature/epel/data/plans.fmf
psss: `flac` is in crb, we want to exercise installing a package from the `crb` repo.

martinhoyer: I saw flac in epel?

martinhoyer: compared to flac:
```
Installing:
 flac                       x86_64                  1.4.3-6.el10                       rhel-CRB-latest                  173 k
Installing dependencies:
 flac-libs                  x86_64                  1.4.3-6.el10                       rhel-AppStream                   269 k
 libogg                     x86_64                  2:1.3.5-10.el10                    rhel-AppStream                    36 k

Transaction Summary
==============================================================================================================================
Install  3 Packages

Total download size: 478 k
Installed size: 1.1 M
```

PR #3524, opened 2025-02-12
Force multiline docstrings to start on the second line

happz:
We enable all D* checks, but our selected convention disables some of
them again. Adding D213 to the list of enabled checks does the trick.
    
This will not affect singleline docstrings :/

Pull Request Checklist

* [x] implement the feature


psss (approved):
Yeah, starting on the first line does not look good.

skycastlelily (approved):
LGTM^^

PR #2967, opened 2024-05-29
Add _translate_constraint_by_config function

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

PR #3525, opened 2025-02-13
Use base model from configuration as basis of future metadata containers

happz:
Eventually, we should have as little custom code as possible, leaving `DataContainer` and custom machinery of validation and normalization behind us, switching to attrs and Pydantic. What started in `tmt.config.models`, we can use in general for containers populated by fmf and/or CLI.

Pull Request Checklist

* [x] implement the feature


skycastlelily (approved):
LGTM,and pytest tests/unit/test_config.py pass^^, maybe we should also add "blocking other work" label? As it is kind of blocking [translate hw mr](https://github.com/teemtee/tmt/pull/2967):)

therazix (approved):

PR #3526, opened 2025-02-14
Accept and apply `workdir-root` in `testcloud`

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -863,8 +863,10 @@ class GuestTestcloud(tmt.GuestSsh):
         self.config.DOWNLOAD_PROGRESS = self.debug_level > 2
         self.config.DOWNLOAD_PROGRESS_VERBOSE = False
 
+        # We can't assign the path to STORE_DIR because it must exist first
         data_dir = self.workdir_root / 'testcloud'
         store_dir = data_dir / 'images'
+
         # Make sure required directories exist
         os.makedirs(data_dir, exist_ok=True)
         os.makedirs(store_dir, exist_ok=True)
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -25,12 +25,12 @@ import tmt.utils
 import tmt.utils.wait
 from tmt.container import container, field
 from tmt.utils import (
-    WORKDIR_ROOT,
     Command,
     Path,
     ProvisionError,
     ShellScript,
     configure_constant,
+    effective_workdir_root,
     retry_session,
 )
 from tmt.utils.wait import Deadline, Waiting
@@ -105,12 +105,6 @@ def import_testcloud(logger: tmt.log.Logger) -> None:
     global TPM_CONFIG_ALLOWS_VERSIONS
     TPM_CONFIG_ALLOWS_VERSIONS = hasattr(TPMConfiguration(), 'version')
 
-
-# Testcloud cache to our tmt's workdir root
-TESTCLOUD_DATA = (
-    Path(os.environ['TMT_WORKDIR_ROOT']) if os.getenv('TMT_WORKDIR_ROOT') else WORKDIR_ROOT
-) / 'testcloud'
-TESTCLOUD_IMAGES = TESTCLOUD_DATA / 'images'
 
 TESTCLOUD_WORKAROUNDS: list[str] = []  # A list of commands to be executed during guest boot up
 
@@ -869,9 +863,15 @@ class GuestTestcloud(tmt.GuestSsh):
         self.config.DOWNLOAD_PROGRESS = self.debug_level > 2
         self.config.DOWNLOAD_PROGRESS_VERBOSE = False
 
+        data_dir = self.workdir_root / 'testcloud'
+        store_dir = data_dir / 'images'
+        # Make sure required directories exist
+        os.makedirs(data_dir, exist_ok=True)
+        os.makedirs(store_dir, exist_ok=True)
+
         # Configure to tmt's storage directories
-        self.config.DATA_DIR = TESTCLOUD_DATA
-        self.config.STORE_DIR = TESTCLOUD_IMAGES
+        self.config.DATA_DIR = data_dir
+        self.config.STORE_DIR = store_dir
 
         self.config.STOP_RETRIES = self.stop_retries
         self.config.STOP_RETRY_WAIT = self.stop_retry_delay
@@ -989,9 +989,6 @@ class GuestTestcloud(tmt.GuestSsh):
 
         if self.is_dry_run:
             return
-        # Make sure required directories exist
-        os.makedirs(TESTCLOUD_DATA, exist_ok=True)
-        os.makedirs(TESTCLOUD_IMAGES, exist_ok=True)
 
         # Prepare the console log
         assert self.logdir is not None  # Narrow type
@@ -1028,7 +1025,7 @@ class GuestTestcloud(tmt.GuestSsh):
             raise ProvisionError(f"Image '{self._image.local_path}' not found.") from error
         except (testcloud.exceptions.TestcloudPermissionsError, PermissionError) as error:
             raise ProvisionError(
-                f"Failed to prepare the image. Check the '{TESTCLOUD_IMAGES}' "
+                f"Failed to prepare the image. Check the '{self.config.STORE_DIR}' "
                 f"directory permissions."
             ) from error
         except KeyError as error:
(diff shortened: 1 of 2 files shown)


tmt/steps/provision/testcloud.py
```diff
@@ -834,9 +828,18 @@ def prepare_config(self) -> None:
...
+        # Make sure required directories exist
+        os.makedirs(data_dir, exist_ok=True)
+        os.makedirs(store_dir, exist_ok=True)
+
         # Configure to tmt's storage directories
-        self.config.DATA_DIR = TESTCLOUD_DATA
-        self.config.STORE_DIR = TESTCLOUD_IMAGES
+
+        self.config.DATA_DIR = effective_workdir_root(self.workdir_root) / 'testcloud'
+        self.config.STORE_DIR = self.config.DATA_DIR / 'images'
```
happz, line 840: 1. `DATA_DIR` seems to be already decided on line 831, shall we just do `self.config.DATA_DIR = data_dir`?
2. same goes for `store_dir`

It seems to me we can even assign to `DATA_DIR` and `STORE_DIR` and *then* call `makedirs()`:

```python
        self.config.DATA_DIR = self.workdir_root / 'testcloud'
        self.config.STORE_DIR = data_dir / 'images'

        # Make sure required directories exist
        os.makedirs(self.config.DATA_DIR, exist_ok=True)
        os.makedirs(self.config.STORE_DIR, exist_ok=True)
```

tmt/steps/provision/testcloud.py
```diff
@@ -1287,10 +1287,11 @@ def _print_local_images(self) -> None:
         Print images which are already cached
         """
 
+        store_dir = effective_workdir_root(self.workdir_root) / 'testcloud/images'
```
happz, line 1290: IIUIC, we should use `self.workdir_root` which has been already decided for us.

tmt/steps/provision/testcloud.py
skycastlelily: > It seems to me we can even assign to DATA_DIR and STORE_DIR and then call makedirs():

No, we can not now, due to the [testcloud code change](https://github.com/teemtee/testcloud/blob/main/testcloud/config.py#L108)

> DATA_DIR seems to be already decided on line 831, shall we just do self.config.DATA_DIR = data_dir?

Updated: https://github.com/teemtee/tmt/commit/bad3543f4543f84e803458236af3a487887e8001:)

tmt/steps/provision/testcloud.py
skycastlelily: Updated, https://github.com/teemtee/tmt/commit/bad3543f4543f84e803458236af3a487887e8001:)

tmt/steps/provision/testcloud.py
happz: > > It seems to me we can even assign to DATA_DIR and STORE_DIR and then call makedirs():
> 
> No, we can not now, due to the [testcloud code change](https://github.com/teemtee/testcloud/blob/main/testcloud/config.py#L108)

What does it mean? `DATA_DIR` has both a getter and setter, why can't we assign to it and then read it back?

skycastlelily: Sorry for making it confusing,  I should put more [info](https://github.com/teemtee/testcloud/blob/main/testcloud/sql.py#L56) :sweat_smile: 

If the  dir we are going to be set for self.config.DATA_DIR  does not exit, we will see:

"unable to open database file" error

skycastlelily: Update^^: https://github.com/teemtee/tmt/pull/3526/commits/da30430bc3732a99819c78d195b4cbc9276f18d4

therazix (approved):

tmt/steps/provision/testcloud.py
happz: Why not `store_dir = self.workdir_root / 'testcloud/images'`?

skycastlelily: right :sweat_smile:  updated^^ https://github.com/teemtee/tmt/pull/3526/commits/1948c2b928520e5033a5fdb40329698c263e43ad

tmt/steps/provision/testcloud.py
```diff
@@ -836,9 +829,17 @@ def prepare_config(self) -> None:
         self.config.DOWNLOAD_PROGRESS = self.debug_level > 2
         self.config.DOWNLOAD_PROGRESS_VERBOSE = False
 
+        # We can't assign the path to STORE_DIR because it must exist first
+        data_dir = self.workdir_root / 'testcloud'
+        store_dir = data_dir / 'images'
```
happz, line 834: Oh, I just realized these should not be local variables, but properties of `self`! See `store_dir = self.workdir_root / 'testcloud/images'` below, it's constructing the same path, and we should avoid that and provide canonical "path where images are stored" attribute. So, how about the following?

```python
@functools.cached_property
def testcloud_data_dirpath(self) -> Path:
    return self.workdir_root / 'testcloud'

@functools.cached_property
def testcloud_image_dirpath(self) -> Path:
    return self.testcloud_data_dirpath / 'testcloud'
```

And then use these, `os.makedirs(self.testcloud_data_dirpath, exist_ok=True)` and so on.

tmt/steps/provision/testcloud.py
```diff
@@ -981,7 +979,7 @@ def start(self) -> None:
             raise ProvisionError(f"Image '{self._image.local_path}' not found.") from error
         except (testcloud.exceptions.TestcloudPermissionsError, PermissionError) as error:
             raise ProvisionError(
-                f"Failed to prepare the image. Check the '{TESTCLOUD_IMAGES}' "
+                f"Failed to prepare the image. Check the '{self.config.STORE_DIR}' "
```
happz, line 982: `self.config.STORE_DIR` should become `self.testcloud_image_dirpath`.

tmt/steps/provision/testcloud.py
```diff
@@ -1346,10 +1349,11 @@ def _print_local_images(self) -> None:
         Print images which are already cached
         """
 
+        store_dir = self.workdir_root / 'testcloud/images'
```
happz, line 1352: Not needed, we can use `self.testcloud_image_dirpath` below.

skycastlelily: np, I'm afraid we can't ,testcloud_image_dirpath is a property of GuestTestcloud,  while _print_local_images is a function of ProvisionTestcloud :)

happz: Hmm, right. So how about moving `testcloud_data_dirpath` and `testcloud_image_dirpath` to `ProvisionTestcloud`? Guest class can then access them via `self.step`. The yare same fo all testcloud guests in the given run anyway.

skycastlelily: Updated^^: https://github.com/teemtee/tmt/pull/3526/commits/e002b8b8a14ce158f9cdfd024bc345bb4295516e

tmt/steps/provision/testcloud.py
thrix: @skycastlelily this one looks unresolved?

tmt/steps/provision/testcloud.py
thrix: looks resolved.

tmt/steps/provision/testcloud.py
thrix: @happz can we resolve?

tmt/steps/provision/testcloud.py
thrix: looks resolved

martinhoyer (approved):
Added one small commit replacing os.mkdirs with Path.mkdir.

skycastlelily:
> Added one small commit replacing os.mkdirs with Path.mkdir.

opps,thanks for your review , I forgot to fetch your commit before I roll back, would you please update this with your commit,please?  :sweat_smile .  The latest commit actually doesn't work well, so I roll back to previous solution:)

tmt/steps/provision/testcloud.py
```diff
@@ -836,9 +836,14 @@ def prepare_config(self) -> None:
         self.config.DOWNLOAD_PROGRESS = self.debug_level > 2
         self.config.DOWNLOAD_PROGRESS_VERBOSE = False
 
+        # We can't assign a non-existent path to STORE_DIR,
+        # so we should make sure required directories exist
+        self.testcloud_data_dirpath.mkdir(parents=True, exist_ok=True)
+        self.testcloud_data_dirpath.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
martinhoyer: hmm it does't autoclose?

happz (approved):

happz:
Unrelated failure, merging.

Issue #3349, opened 2024-11-08
Prepare playbook from URL does not work for imported plans


The-Mule:
When running a an imported plan that contains Ansible playbook specified as URL, prepare phase fails claiming that the playbook does not exist.

Importing plan:

```
/fips-enabled:
  plan:
    import:
      url: https://.../tests/libkcapi
      name: /plans/ci/fips-enabled
``` 

Actual plan:

```
summary: CI test plan

execute:
  how: tmt

/fips-enabled:
  context+:
    fips: enabled
  prepare+:
    - name: Enable FIPS mode
      how: ansible
      order: 99
      playbook:
        - https://.../playbooks/enable-fips.yaml
 ...
```

Playbook URL is correct, the playbook is reachable. This can be proved by running the actual plan directly - it just works:

```
   prepare
        ...
        queued prepare task #1: Enable beaker-tasks on default-0
        ... 
        prepare task #4: Enable FIPS mode on default-0
        how: ansible
        name: Enable FIPS mode
        playbook: https://.../playbooks/enable-fips.yaml
        playbook-path: playbook-isftw5ft.yml
        ok: 1 task
        changed: 1 task
        skipped: 1 task
        ...
        summary: 4 preparations applied
```
 
On the other hand when running the importing one that imports the actual one, it does not work:

```
prepare
        ...
        queued prepare task #4: Enable FIPS mode on default-0
        ...
        prepare task #4: Enable FIPS mode on default-0
        how: ansible
        name: Enable FIPS mode
        playbook: https://.../playbooks/enable-fips.yaml
        playbook-path: playbook-0ldv71eq.yml
        fail: Playbook '/var/ARTIFACTS/work-fips-enabled-buildroot-disabled0niufoxa/import/plans/ci/fips-enabled-buildroot-disabled/playbook-0ldv71eq.yml' does not exist.
```

I believe the issue is that TMT downloads the playbook into different directory than than "/import/plans/..." directory (maybe the actual run directory, ie. "/plans/..."?

Issue #3237, opened 2024-09-25
ansible plugin downloads playbook to the plan directory, not workdir


lukaszachy:
I use plan like https://github.com/lukaszachy/testing-farm-konflux/blob/main/examples/tmt-plan/konflux.fmf
and once ti finishes I always have files like `playbook-la0vgbpg.yml` in the directory with the plan (fmf root as well as git root).

tmt version: 1.37.dev26+g1238dcaa

The command line used is `tmt -c trigger=commit -c initiator=konflux run -avvv --environment-file envfile.yaml report -h html -o login -w error -w fail`

IMO tmt should refrain from modifying plan directory (or any input files/directories).

happz:
That was a price for the "all playbooks must be under fmf tree root" rule which has been relaxed a bit recently. I suppose `prepare/ansible` should use its workdir, correct? Root selection (https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/ansible.py#L171) is correct for file-backed playbooks, for the downloaded different root will be needed, but that's easy to achieve, set `root_path` before the `if` branch and set it to workdir for remote playbooks.

PR #3527, opened 2025-02-17
Save remote ansible playbooks to correct directory

therazix:
This PR changes the location where remote ansible playbooks are saved. Previously, they were saved in the plan's directory, but now they are saved in the workdir (`/var/tmp/tmt/...`) to ensure that tmt does not modify the plan directory.

This change also resolves the issue with remote ansible playbooks not functioning correctly when used with imported plans.

Fixes: #3237
Fixes: #3349

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation

### DIFF
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -186,10 +186,7 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
             lowercased_playbook = _playbook.lower()
 
             def normalize_remote_playbook(raw_playbook: str) -> Path:
-                assert self.step.plan.my_run is not None  # narrow type
-                assert self.step.plan.my_run.tree is not None  # narrow type
-                assert self.step.plan.my_run.tree.root is not None  # narrow type
-                root_path = self.step.plan.my_run.tree.root
+                root_path = self.step.plan.anchor_path
 
                 try:
                     with retry_session(
--- a/tmt/steps/prepare/ansible.py
+++ b/tmt/steps/prepare/ansible.py
@@ -186,7 +186,8 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
             lowercased_playbook = _playbook.lower()
 
             def normalize_remote_playbook(raw_playbook: str) -> Path:
-                root_path = self.step.plan.anchor_path
+                assert self.step.workdir is not None  # narrow type
+                root_path = self.step.workdir
 
                 try:
                     with retry_session(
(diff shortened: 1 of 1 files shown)


happz (approved):

therazix:
@happz I noticed there is a similar issue #3237 and I tried to fix it in 8855a08752699c9e18a4688d1113c6bc6e4ac2b4. Now, remote playbooks are saved in the prepare directory inside the workdir (`/var/tmp/tmt/run-xyz/plans/plan_name/prepare`), which gets purged during the finish step. Does this approach make sense, or would it be better to use a different directory so the playbook is preserved beyond the finish step?

happz:
> @happz I noticed there is a similar issue #3237 and I tried to fix it in [8855a08](https://github.com/teemtee/tmt/commit/8855a08752699c9e18a4688d1113c6bc6e4ac2b4).

Makes sense, it seems to be the same kind of trouble.

> Now, remote playbooks are saved in the prepare directory inside the workdir (`/var/tmp/tmt/run-xyz/plans/plan_name/prepare`), which gets purged during the finish step. Does this approach make sense, or would it be better to use a different directory so the playbook is preserved beyond the finish step?

I'd say workdir is a good option. We can always improve `finish` to not remove them when pruning its workdir, either by giving them some shared filename prefix (`ansible-remote-playbook-foo.yaml`?) or putting remote playbooks in a dedicated directory (`$workdir/remote-ansible-playbooks`?).

tmt/steps/prepare/ansible.py
```diff
@@ -222,8 +220,11 @@ def normalize_collection_playbook(raw_playbook: str) -> AnsibleCollectionPlayboo
                 return AnsibleCollectionPlaybook(raw_playbook)
 
             playbook: AnsibleApplicable
+            playbook_root = self.step.plan.anchor_path
 
             if lowercased_playbook.startswith(('http://', 'https://')):
+                assert self.step.workdir is not None  # narrow type
+                playbook_root = self.step.workdir
```
happz, line 227: As soon as `self.step.workdir` appears for the second time, I'd suggest changing the `normalize_*` family of functions to return both playbook root and playbook, i.e. `-> tuple[Path, AnsibleApplicable]`, and most of them would return `anchor_path` but `normalize_remote_playbook` would return `(self.step.workdir, Path(file.name).relative_to(root_path))`.

That way, the "this kind of playbook lives in step workdir" knowledge would be owned by the code responsible for processing the raw playbook path, instead of being "shared" by this code *and* code calling it.

What do you think?

therazix: Great idea. Updated in 590b020282e6c1139602379f8549d4bbd16704a7.

happz (approved):

thrix (approved):
LGTM

PR #3528, opened 2025-02-17
Do not use --squash when building test container images

happz:
`build --squash` effectively disables the caching layer. Since there are no intermediate layers that could be inspected, there is only the last layer, all commands seem to be reexuted every time images are built. Which results in no saving of resources.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ images/test/bases:  ## Download base images for custom test images
 # Build a single container: <image name> <containerfile>
 define do-build-container-image =
 @ echo "$(ccgreen)Building$(ccend) $(ccred)${1}$(ccend) $(ccgreen)image...$(ccend)"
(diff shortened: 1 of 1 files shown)


lukaszachy (approved):

martinhoyer (approved):
Very nice.  
Do you know if it make sense to look into caching the containers dir in between the individual runs? I assume some of the layers don't change that often.  

Not having experience with this, I imagine that if it was github pipeline, I'd use `cache` action to store `.local/share/containers/storage/**` (and/or containers/cache?) in order for build tasks to reuse layers with matching hash. ...but I really don't know what I'm talking about.

lukaszachy:
Failure of repofile metadata in  `/tests/prepare/feature` doesn't seem to be caused by this change (it passes in other jobs).

PR #3529, opened 2025-02-17
Fix incorrect directory in /tests/unit/with-development-packages/basic

happz:
It seems to be broken for some time, but since the test was part of CI on Fedora 39 only, and that the CI does not run on Fedora 39 for some time already, we simply did not know.

Pull Request Checklist

* [x] implement the feature


lukaszachy (approved):

happz:
@lukaszachy FYI, we need rawhide check to pass, the test in question runs only there (it runs against devel packages, and as such it's more or less fine if it runs on just one distro).

PR #3535, opened 2025-02-18
Define "deadline" and "waiting" classes for better (reboot) deadlines

happz:
tmt needs to stop doing something, usually waiting for a guest to become
alive, for the first time or once again after a reboot. The deadline is
not clearly defined, with several inputs, and the reboot waiting code is
not honoring deadlines correctly. It is unclear what happens when time
is spent working towards a goal while deadline remains the same.

Patch adds two classes, one to hold a deadline and provide some helper
methods, and other to hold info about how to wait for a deadline. This:

* deadlines instead of timeouts. No chance to forget decrease the
  timeout.
* less parameters passed between methods, as deadline, tick, tick
  increase are now bundled into a single package, simpler call sites.
* it is now easy to switch from one way of waiting to another one, e.g.
  when starting a guest: plugins may have different timeouts for booting
  a guest and the guest making its SSH available; together, they still
  need to fit into one "get guest up and running" deadline.

Pull Request Checklist

* [x] implement the feature


tmt/utils/wait.py
```diff
@@ -0,0 +1,262 @@
...
+
+    #: The deadline that limits the waiting.
+    deadline: Deadline
+
+    #: How many seconds to wait between two consecutive checks whether
+    #: the condition is satisfied.
+    tick: float = tmt.utils.DEFAULT_WAIT_TICK
+
+    #: A multiplier applied to :py:attr:`tick` after every attempt.
+    tick_increase: float = tmt.utils.DEFAULT_WAIT_TICK_INCREASE
```
therazix, line 159: Both `tick` and `tick_increase` should use constants from `tmt.utils.wait`. Constants in `tmt.utils` should be deleted.

happz: Yes, it looks like I forgot to move them.

happz: I didn't forget to move them, I forgot to remove the old ones. Fixed in ede74b7f.

therazix (approved):

skycastlelily (approved):
LGTM:)

PR #3536, opened 2025-02-18
Deduplicates the `finish` documentation

mcasquer:
Pull Request Checklist

* [ ] write the documentation
* [ ] adjust plugin docstring

### DIFF
--- a/spec/plans/finish.fmf
+++ b/spec/plans/finish.fmf
@@ -15,14 +15,7 @@ example: |
 
 /shell:
     summary:
-        Perform finishing tasks using shell (bash) scripts
-    description:
-        Execute arbitrary shell commands to finish the testing.
-
-        Default shell options are applied to the script, see the
-        :ref:`/spec/tests/test` key specification for more
-        details.
-
+        Perform finishing tasks using shell (bash) scripts.
     example: |
         finish:
             how: shell
@@ -35,15 +28,6 @@ example: |
 /ansible:
     summary:
         Perform finishing tasks using ansible
-    description: |
-        One or more playbooks can be provided as a list under the
-        ``playbook`` attribute.  Each of them will be applied
-        using ``ansible-playbook`` in the given order. The path
-        must be relative to the metadata tree root.
-
-        Remote playbooks can be referenced as well as the local ones,
-        and both kinds can be used at the same time.
-
     example: |
         finish:
             how: ansible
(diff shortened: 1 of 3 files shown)


happz (approved):

thrix (approved):
LGTM

spec/plans/finish.fmf
```diff
@@ -3,10 +3,7 @@ summary: Finishing tasks
 description:
     Additional actions to be performed after the test execution
     has been completed.  Counterpart of the ``prepare`` step
```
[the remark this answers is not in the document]
mcasquer: Updated !

spec/plans/finish.fmf
```diff
@@ -15,14 +12,7 @@ example: |
...
     summary:
-        Perform finishing tasks using shell (bash) scripts
-    description:
-        Execute arbitrary shell commands to finish the testing.
-
-        Default shell options are applied to the script, see the
-        :ref:`/spec/tests/test` key specification for more
-        details.
-
+        Perform finishing tasks using shell (bash) scripts.
```
psss, line 15: I'm thinking about simplifying this section even a little bit more: Here we still mention `shell` and `ansible` plugins, which means we need to maintain the list, add new items when a new `finish` plugin is implemented.

What about just explaining what the `finish` step is about, including a single concise example for user to get a quick idea, and pointing to the `/plugins/finish` section to learn all details.

The example could quickly demonstrate both plugins, for example like this:

```yaml
finish:
  - how: shell
    script: upload-logs.sh
  - how: ansible
    playbook: playbooks/cleanup.yaml
```

psss (changes requested):
Thanks for the cleanup. Looks good. What about dropping the list of available plugins from the spec section so that we do not have to maintain it here at all?

spec/plans/finish.fmf
mcasquer: In addition, if I remove all of them, especially the `where` I get the following warning when compiling
```bash
guide.rst:920: WARNING: undefined label: '/spec/plans/finish/where'
```

psss: > @psss If I remove all the items then the `finish` step becomes a draft, this could be confusing for people reading the documentation... what do you think?

I didn't mean to remove everything, the `where` and `when` keys should definitely stay. At least for now. Plus there would be the overall description and example, so that section should not be empty, right?

psss (approved):
Thanks! Added the `Finish Plugins` section header plus just a few minor adjustments in aa84ef0a. @mcasquer, does it look ok to you?

mcasquer:
> Thanks! Added the `Finish Plugins` section header plus just a few minor adjustments in [aa84ef0](https://github.com/teemtee/tmt/commit/aa84ef0a2c5074c9631301f53c4337909e2a72ef). @mcasquer, does it look ok to you?

@psss LGTM :)

PR #3540, opened 2025-02-18
Store check result modified during interpretation

psss:
When presenting results to the user it should be clear at the first sight which check did not fail as expected or which checks are informational only.

Pull Request Checklist

* [ ] implement the feature
* [ ] include a release note

### DIFF
--- a/tmt/result.py
+++ b/tmt/result.py
@@ -398,10 +398,11 @@ class Result(BaseResult):
         :returns: :py:class:`ResultOutcome` instance with the interpreted result
         """
 
+        # Process only relevant checks
+        checks = [check for check in self.check if check.name == check_name]
+
         # Reduce all check outcomes into a single worst outcome
-        reduced_outcome = ResultOutcome.reduce(
-            [check.result for check in self.check if check.name == check_name]
-        )
+        reduced_outcome = ResultOutcome.reduce([check.result for check in checks])
 
         # Now let's handle the interpretation
         interpret = interpret_checks[check_name]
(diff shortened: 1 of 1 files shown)


martinhoyer:
This doesn't feel right to me. Not sure I can pin-point why, but for one I presume this will break when multiple checks of same name are present. Also there are before/after phases and users don't specify expected outcome per phase.

Hard to wrap a head around this, but if I understand it correctly, when changing the check.result like this, it could mean that when 1 out of 4 things fails, we convert all four to fail. As it's now, results would stay and only the actual test result would be affected with added note as for why.

I'm guessing the pain-point is the disconnect between Test Result and Check Result, where in Test Result, we are simply flipping it (and creating original_outcome), while in Check Result it's currently not quite possible treat it the same way.  

iirc, there are more questions...
 - Will there be checks which do different things in before and after phase? i.e. coredump/abrt/kdump..., where before-test it's a "setup" and after-test is actual check if something went wrong?
 - Even if not, users would still benefit from being able to set expected outcome for each phase
 - If expected outcome is set to check as a whole and only one phase fails, I don't think we should flip results of both phases, it's just....wrong? :)  
 oh and what about scenario, where before-test dmesg check fails, which clears the logs; and then after test the check passes, but everything is red.

Maybe I'm missing something, but couldn't most of this be solved by doing Check.SubResult or PhaseResult? (alongside moving away from using name as a identifier)

psss:
Seems this one deserves a discussion about to agree on how to approach it.

psss:
Next step as suggested above:

> We really need that unit test to collect all combinations and corner cases and test what we expect to happen.

Tentatively proposed for the housekeeping sprint.

psss:
Blocked on #4578.

PR #3543, opened 2025-02-20
Support ausearch --checkpoint method of noticing AVC denials

happz:
It has some advantages, namely no need to deal with locale. Still, the timestamp-based method is available and can be selected.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/tests/test/check/test-avc.sh
+++ b/tests/test/check/test-avc.sh
@@ -18,12 +18,12 @@ rlJournalStart
     rlPhaseEnd
 
     for method in checkpoint timestamp; do
-        rlPhaseStartTest "Run /avc tests with $PROVISION_HOW"
+        rlPhaseStartTest "Run /avc tests with $PROVISION_HOW ($method method)"
             rlRun "tmt -c provision_method=$PROVISION_HOW run --id $run --scratch -a -vvv provision -h $PROVISION_HOW test -n /avc/$method" "1"
             rlRun "cat $results"
         rlPhaseEnd
 
-        rlPhaseStartTest "Test harmless AVC check with $PROVISION_HOW"
+        rlPhaseStartTest "Test harmless AVC check with $PROVISION_HOW ($method method)"
             rlRun "avc_log=$run/plan/execute/data/guest/default-0/avc/$method/harmless-2/checks/avc.txt"
             rlAssertExists "$avc_log"
             rlRun "cat $avc_log"
@@ -40,7 +40,7 @@ rlJournalStart
             fi
         rlPhaseEnd
 
-        rlPhaseStartTest "Test nasty AVC check with $PROVISION_HOW"
+        rlPhaseStartTest "Test nasty AVC check with $PROVISION_HOW ($method method)"
             rlRun "avc_log=$run/plan/execute/data/guest/default-0/avc/$method/nasty-1/checks/avc.txt"
             rlAssertExists "$avc_log"
             rlRun "cat $avc_log"
--- a/tests/test/check/test-avc.sh
+++ b/tests/test/check/test-avc.sh
@@ -18,6 +18,10 @@ rlJournalStart
     rlPhaseEnd
 
     for method in checkpoint timestamp; do
+        # TODO: for some reason, the checkpoint method does not seem to report the expected AVC
+        # denials. There must be something wrong, but it's not clear what and where.
+        if [ "$method" = "checkpoint" ] && [ "$PROVISION_HOW" = "local" ]; then continue; fi
+
         rlPhaseStartTest "Run /avc tests with $PROVISION_HOW ($method method)"
             rlRun "tmt -c provision_method=$PROVISION_HOW run --id $run --scratch -a -vvv provision -h $PROVISION_HOW test -n /avc/$method" "1"
             rlRun "cat $results"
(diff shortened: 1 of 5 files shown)


lukaszachy (commented):
I'm not sure about the default, given problems to make it pass for localhost.

happz:
> I'm not sure about the default, given problems to make it pass for localhost.

Fair point, I will switch it to `timestamp`. I did not expect this kind of instability with `local`.

tmt/checks/avc.py
```diff
@@ -21,17 +27,102 @@
...
-AUSEARCH_TIMESTAMP_FILENAME = 'avc-timestamp.sh'
+#: The filename of the "mark" file ``ausearch`` on the guest.
+AUSEARCH_MARK_FILENAME = 'avc-mark.txt'
 
 #: Packages related to selinux and AVC reporting. Their versions would be made
 #: part of the report.
 INTERESTING_PACKAGES = ['audit', 'selinux-policy']
 
 
+SETUP_SCRIPT_TIMESTAMP = jinja2.Template(
```
martinhoyer, line 69: Wouldn't it make sense to de-duplicate the templates with conditions and have only one SETUP_SCRIPT? Something like:
```
SETUP_SCRIPT = jinja2.Template(
    textwrap.dedent("""
    set -x
    export LC_ALL=C
    {% if method == 'timestamp' %}
    echo "export AVC_SINCE=\\"$( date "+%x %H:%M:%S")\\"" > {{ mark_filepath }}
    {% else %}
    ausearch --input-logs --checkpoint {{ mark_filepath }} -m AVC -m USER_AVC -m SELINUX_ERR
    {% endif %}
    cat {{ mark_filepath }}
    """)
)
How do I decide whether to keep the template in code versus in separate .j2 file?

File: tmt/checks/avc.py [hunk 2]
```diff
@@ -169,7 +268,12 @@ def create_final_report(
             "Test does not have start time recorded, cannot run AVC check."
         )
 
-    ausearch_timestamp_filepath = invocation.check_files_path / AUSEARCH_TIMESTAMP_FILENAME
+    ausearch_mark_filepath = invocation.check_files_path / AUSEARCH_MARK_FILENAME
+
+    # Wait one second before storing the mark because ausearch
+    # could catch denials from the previous test if they are executed
+    # during the same second
+    time.sleep(check.delay_before_report)
```
martinhoyer, line 276: isn't there a (easy) way to detect the previous test execution has finished  instead of using sleep?

martinhoyer (commented):
Looks like a great idea to me. Added a comment to the jinja templates. What is causing the timestamp method to pass on local?

tmt/checks/avc.py
happz: Hm, could be done, yes. I will do it once done with debugging the checkpoint.

tmt/checks/avc.py
happz: It did finish. But since `ausearch` granularity is one second, it may happen that the last test finished and the check for the new one started in the same second. Extremely unlikely, given all the delays here, but there is not guarantee it cannot happen. Hence the delay, to make sure this AVC check begins with timestamp that **is** different than the one the previous test might have seen.

tmt/checks/avc.py
happz: Addressed in 69c75731

tmt/checks/avc.py
martinhoyer: Ahh, I get it. How about nanosecond timestamps?  
```
mhoyer@fedora:~$ date -u '+%Y-%m-%dT%H:%M:%S.%N'
2025-02-24T11:30:48.122004339
mhoyer@fedora:~$ date --iso-8601=ns
2025-02-24T12:30:48,928138809+01:00
```

martinhoyer: hmm, ausearch doesn't like nanoseconds apparently

tests/test/check/test-avc.sh
```diff
@@ -17,35 +17,51 @@ rlJournalStart
...
     rlPhaseEnd
 
-    rlPhaseStartTest "Run /avc tests with $PROVISION_HOW"
-        rlRun "tmt -c provision_method=$PROVISION_HOW run --id $run --scratch -a -vvv provision -h $PROVISION_HOW test -n /avc" "1"
-        rlRun "cat $results"
-    rlPhaseEnd
+    for method in checkpoint timestamp; do
+        # TODO: for some reason, the checkpoint method does not seem to report the expected AVC
+        # denials. There must be something wrong, but it's not clear what and where.
+        if [ "$method" = "checkpoint" ] && [ "$PROVISION_HOW" = "local" ]; then continue; fi
```
lukaszachy, line 23: Will you enable again this on localhost?

happz: I didn't plan to, in this PR that is, because I have **zero** idea what's wrong. If there are any hints, I can give it a try, but I am out of ideas, and for the sake of getting the timestamp fixed, I am willing to let `local` + `avc` + the non-default `checkpoint` method let in.

therazix (approved):

martinhoyer (approved):

tests/test/check/test-avc.sh
martinhoyer: Would be cool to get to the bottom of this, but if we can't in time for the release, +1 for fixing timestamp.

PR #3544, opened 2025-02-21
Enable preview lint checks of `ruff`

happz:
See [1]: preview rules are sometimes interesting rules, but they are not enabled and ruff community collects opinions before deciding whether a rule is worth eing enabled by default.

And there are some useful rules, I'd like to start enabling them in smaller groups, to make review easier. Starting with this patch and already enabled rules.

[1] https://docs.astral.sh/ruff/preview/#preview

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/report/junit.py
+++ b/tmt/steps/report/junit.py
@@ -384,7 +384,6 @@ def make_junit_xml(
 
     xml_parser = etree.XMLParser(**xml_parser_kwargs)
     try:
-        # S320: Parsing of untrusted data is known to be vulnerable to XML attacks.
         tree_root: XMLElement = etree.fromstring(xml_data, xml_parser)
 
     except etree.XMLSyntaxError as e:


happz:
> I'm usually using --preview locally, but not sure about enabling it config, even manually, as it might be hard to maintain. The preview rules can often be moved to other rules, or be dropped completely. That said, given the cadence of ruff releases, perhaps we should focus on some autoupdate solution. One could be move to https://github.com/pre-commit/action ?
> 
> Is there a way to only show preview violation as a warning, not failing the check?

No idea.

Enabling the preview rules showed very few violations, and TBH, since February, I forgot **why** I wanted to enable this, what were the issues I discovered there's a check for which is preview-only. So it's a kind of sad story now. Anyway, maybe my memory will return, and it can just as well rot in PR queue, marking it as ready for review and moving on.

happz:
@coderabbitai full review

bajertom:
All changes seem reasonable to me. I have one question though:
removing the `# noqa: S320` in tmt/steps/report/junit.py will now make the linter complain about vulnerability of XML attacks, right? Why were we ignoring it in the first place?

happz:
> All changes seem reasonable to me. I have one question though: removing the `# noqa: S320` in tmt/steps/report/junit.py will now make the linter complain about vulnerability of XML attacks, right? Why were we ignoring it in the first place?

Interesting, I didn't realize this, and ruff should complain. Unless there is something wrong with that rule; I'll check. Plus, if it's to be removed, the comment should go as well.

It was disabled because we intended to parse some strings into XML trees. It's pretty much a "be aware, know what you are doing" message.

happz:
> > All changes seem reasonable to me. I have one question though: removing the `# noqa: S320` in tmt/steps/report/junit.py will now make the linter complain about vulnerability of XML attacks, right? Why were we ignoring it in the first place?
> 
> Interesting, I didn't realize this, and ruff should complain. Unless there is something wrong with that rule; I'll check. Plus, if it's to be removed, the comment should go as well.
> 
> It was disabled because we intended to parse some strings into XML trees. It's pretty much a "be aware, know what you are doing" message.

@bajertom aha, S320 rule has been deprecated, and will be removed from ruff. Apparently, lxml has been fixed. So it was removed by the ruff itself.

bajertom (approved):

psss (approved):

martinhoyer:
I don't think this is a good idea, but let's see :)
@happz fwiw, quite a lot of cool rules in new ruff versions (also weird that we are enabling --preview instead keeping ruff up-to-date), but since it requires line changes, reviews, etc. I believe it's better to first focus on a) renovate and b) removing EPEL9 rpms and Python 3.9 support.

happz:
> also weird that we are enabling --preview instead keeping ruff up-to-date

It doesn't seem like one-or-the-other choice to me. Even the current ruff docs, which presumably speak about the current up-to-date ruff, list over a hundred "preview" rules. As I understand it, "preview" rules become regular ones at some point, but it seems new "preview" rules appear as well. So, to me, it seems we can have both.

PR #3548, opened 2025-02-25
Do not remove plugin's workdir if there are preserved members

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -1383,6 +1383,8 @@ class BasePlugin(Phase, Generic[StepDataT, PluginReturnValueT]):
 
     _data_class: type[StepDataT]
 
+    preserved_members: list[str] = []
+
     @classmethod
     def get_data_class(cls) -> type[StepDataT]:
         """
@@ -1876,11 +1878,26 @@ class BasePlugin(Phase, Generic[StepDataT, PluginReturnValueT]):
 
         if self.workdir is None:
             return
-        logger.debug(f"Remove '{self.name}' workdir '{self.workdir}'.", level=3)
-        try:
-            shutil.rmtree(self.workdir)
-        except OSError as error:
-            logger.warning(f"Unable to remove '{self.workdir}': {error}")
+        remove_workdir = True
+        for member in self.workdir.iterdir():
+            if member.name in self.preserved_members:
+                # We only want to remove the workdir, if there is no preserved members
+                remove_workdir = False
+            else:
+                logger.debug(f"Remove '{member}'.", level=3)
+                try:
+                    if member.is_file() or member.is_symlink():
+                        member.unlink()
+                    else:
+                        shutil.rmtree(member)
+                except OSError as error:
+                    logger.warning(f"Unable to remove '{member}': {error}")
+        if remove_workdir:
+            logger.debug(f"Remove '{self.name}' workdir '{self.workdir}'.", level=3)
+            try:
+                shutil.rmtree(self.workdir)
+            except OSError as error:
+                logger.warning(f"Unable to remove '{self.workdir}': {error}")
 
 
 class GuestlessPlugin(BasePlugin[StepDataT, PluginReturnValueT]):
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -2613,6 +2613,8 @@ class ProvisionPlugin(tmt.steps.GuestlessPlugin[ProvisionStepDataT, None]):
     # TODO: Generics would provide a better type, https://github.com/teemtee/tmt/issues/1437
     _guest: Optional[Guest] = None
 
+    preserved_members = ['logs']
+
     @classmethod
     def base_command(
         cls,


tmt/steps/__init__.py
```diff
@@ -1876,11 +1878,26 @@ def prune(self, logger: tmt.log.Logger) -> None:
...
         if self.workdir is None:
             return
-        logger.debug(f"Remove '{self.name}' workdir '{self.workdir}'.", level=3)
-        try:
-            shutil.rmtree(self.workdir)
-        except OSError as error:
-            logger.warning(f"Unable to remove '{self.workdir}': {error}")
+        remove_workdir = True
+        for member in self.workdir.iterdir():
+            if member.name in self.preserved_members:
```
happz, line 1883: The name should be `_preserved_workdir_members`, to align with `Step.prune()`.

tmt/steps/__init__.py
```diff
@@ -1876,11 +1878,26 @@ def prune(self, logger: tmt.log.Logger) -> None:
 
         if self.workdir is None:
             return
-        logger.debug(f"Remove '{self.name}' workdir '{self.workdir}'.", level=3)
-        try:
-            shutil.rmtree(self.workdir)
-        except OSError as error:
-            logger.warning(f"Unable to remove '{self.workdir}': {error}")
+        remove_workdir = True
+        for member in self.workdir.iterdir():
```
happz, line 1882: Hm, a bit more wasteful, but definitely more readable:

```python
preservable_members = [member for member in self.workdir.iterdir() if member in self._preserved_workdir_members]
removable_members = [member for member in self.workdir.iterdir() if member not in preservable_members]
```

Then, if `preservable_members` is empty, you can remove the whole workdir, or remove everything from `removable_members`. It should make the code less dense, with fewer conditions.

skycastlelily: much better than mine,and updatedhttps://github.com/teemtee/tmt/pull/3548/commits/3fa627f21493bbd3e5de844c06190d001daeccfc^^

tmt/steps/provision/__init__.py
```diff
@@ -2613,6 +2613,8 @@ class ProvisionPlugin(tmt.steps.GuestlessPlugin[ProvisionStepDataT, None]):
     # TODO: Generics would provide a better type, https://github.com/teemtee/tmt/issues/1437
     _guest: Optional[Guest] = None
 
+    _preserved_workdir_members = ['logs']
```
happz, line 2616: We do not create this directory - yet ;) This addition belongs to the guest log patch.

happz (approved):

tmt/steps/provision/__init__.py
skycastlelily: yeah, and updated ;) https://github.com/teemtee/tmt/pull/3548/commits/0c5ea3b1934f579d9a207fdee8115a27d415567c

tmt/steps/__init__.py
```diff
@@ -1876,11 +1878,30 @@ def prune(self, logger: tmt.log.Logger) -> None:
...
+            member
+            for member in self.workdir.iterdir()
+            if member.name in self._preserved_workdir_members
+        ]
+        removable_members = [
+            member for member in self.workdir.iterdir() if member not in preservable_members
+        ]
+        if preservable_members:
+            for member in removable_members:
+                logger.debug(f"Remove '{member}'.", level=3)
```
martinhoyer, line 1891: Again, apologies. It is wrong from me to comment on perfectly fine code. It's just that it links to the above set/link and that I think we can do one less iterdir() and avoid intermediary list, if that make sense. 

Also wanted to to ask about `unlink(missing_ok=True)` parameter, which was added in 3.8, but then I realized FileNotFoundError is being caught by the OSError, so...nevermind.

```suggestion

        workdir_members = self.workdir.iterdir()

        preservable_members = (
            member for member in workdir_members if member.name in self._preserved_workdir_members
        )

        if preservable_members:
            for member in workdir_members:
                if member in preservable_members:
                    continue

```

martinhoyer (approved):
lgtm, thanks! I've added a couple of comments, just for me (and maybe others) to hopefully learn things.  

@happz Praying for blessing of your wisdom :)

As discussed, `workdir` type is a bit hazy at the moment and it's for a different discussion, just wanted to ask if we're ok with just `if self.workdir is None:`, if the workdir type can be `Literal[True]`.    Once again, apologise for unrelated comments :duck:

happz:
> lgtm, thanks! I've added a couple of comments, just for me (and maybe others) to hopefully learn things.
> 
> @happz Praying for blessing of your wisdom :)
> 
> As discussed, `workdir` type is a bit hazy at the moment and it's for a different discussion, just wanted to ask if we're ok with just `if self.workdir is None:`, if the workdir type can be `Literal[True]`. Once again, apologise for unrelated comments 🦆

I believe we are. It is not expressed with types, but `workdir` should be always `Optional[Path]`, because `True` would be converted into a workdir that did not exist yet, and the path would be assigned to `workdir`.

tmt/steps/__init__.py
```diff
@@ -1383,6 +1383,8 @@ class BasePlugin(Phase, Generic[StepDataT, PluginReturnValueT]):
 
     _data_class: type[StepDataT]
 
+    _preserved_workdir_members: list[str] = []
```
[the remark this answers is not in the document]
happz: > Sorry to pick up on your PR on this, it looks like a good place to bring up this question.
> 
> I've been trying to understand list vs set usage in cases like this. Looking at https://wiki.python.org/moin/TimeComplexity, I understanding is that:
> 
> * `list` will always go though all elements until it finds a match - `O(n)` time complexity
> * `set` will use hash table so it doesn't do additional "scanning/comparing/matching - `O(1)`.
> 
> @happz I remember you told me before, at some other piece of code, that the difference will not be noticeable. Would it be also the case here, in scenario with huge number of files to go through?

If code walks through a lot of directory members, checking which of them are in `_preserved_workdir_members`, `set()` will have a measurable effect. Whether it would have measurable effect with two items in the list and three files in a workdir, that's another question :) But this is a very cheap optimization, let's do it.

I believe there is a ruff check for exactly this, but it's behind the `--preview` option. I have a PR somewhere trying to enable them, but it conflicted with the regular bump of pre-commit checks, so I will resurrect it once that one lands.

tmt/steps/__init__.py
martinhoyer: Ha! I've overoptimized it. Thanks!
> we can do list(workdir.iterdir())

or, well, `set(workdir.iterdir())`, right? When order of elements does not matter. /half-joking :) 
fwiw, I think the memory overhead wouldn't be worth in majority of cases, not going to fire up profiler though.

@happz Big thanks for the write-up. 100% with you on all points. I wouldn't bother you all if these file operations weren't so common in tmt and if I understood it well enough. (same with ssh ;) )

tmt/steps/__init__.py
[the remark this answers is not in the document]
skycastlelily: > Sorry to pick up on your PR on this, it looks like a good place to bring up this question.

Never mind, actually thanks for bringing it up here, I also need to learn:)

> But this is a very cheap optimization, let's do it.

Done^^ https://github.com/teemtee/tmt/pull/3548/commits/a27ec853a1fe550ee8d45691f5dd180221bda45c

skycastlelily:
> @happz Praying for blessing of your wisdom :)

Same here :beers:

tmt/steps/__init__.py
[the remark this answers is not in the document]
happz: @martinhoyer please, re-review.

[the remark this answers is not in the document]
martinhoyer: Thanks both!

martinhoyer (approved):

KwisatzHaderach (approved):

psss (approved):

psss:
Failures are irrelevant errors, merging.

PR #3557, opened 2025-02-27
Turn hints internally into classes

happz:
This should bring better API when working with them. While adding better hints and notes to test and check results, I realized there need to be finer API, to better support hints that will not be rendered, and hints that need to present their summary without a wall of text.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/docs/templates/plugins.rst.j2
+++ b/docs/templates/plugins.rst.j2
@@ -159,7 +159,7 @@ The following keys are accepted by all plugins of the ``{{ STEP }}`` step.
 {% if plugin_full_id in HINTS %}
 .. note::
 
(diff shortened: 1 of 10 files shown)


skycastlelily (approved):
print_hints function works, not able to test the newly added summary related functions,as they are not called anywhere, but the code LGTM^^

happz:
Summaries will be used from https://github.com/teemtee/tmt/pull/3498 and https://github.com/teemtee/tmt/pull/3596

therazix (approved):

PR #3562, opened 2025-02-28
Improve audit events handling in avc check

martinhoyer:
When troubleshooting ausearch --checkpoint failing tests in certain environments, I've observed changes in avc check behaviour when many events are being created for processing. For example `sshd` from tmt ssh connections or `sudo` usage.  
I haven't found a way a better check to find if all events are 'synced' before running ausearch commands, so I'm parsing `auditctl -s` output, which includes: "backlog field tells how many event records are currently queued waiting for auditd to read them" and make sure it is 0 before proceeding.  
Unfortunatelly that's not enough to make it work reliably. 

Using `--just-one` was a bad idea as ausearch always starts from beginning of the log, with no way to start with the last one. Redirecting stdout to /dev/null seems to be just as fast.  
I've also tried to get the event id with `tail -1 audit.log`, and use `auseach -a $id`, but that also didn't solve race conditions(I think) that arise from running all this through ssh. Neither does various --format and --escape options...or any other options for that matter. 
Using `-ts checkpoint` is ultimately the only way I've found in this sunk-cost-fallacy-rabbit-black-hole.  Well, unless we want to have dedicated log file for just avc failures, just for the duration of the test.

Needs tests and I'm not quite happy with the bash script and report. Would mark is as draft, but that would not trigger pipelines.

Pull Request Checklist

* [ ] implement the feature

### DIFF
--- a/tmt/checks/avc.py
+++ b/tmt/checks/avc.py
@@ -1,7 +1,6 @@
 import datetime
 import enum
 import textwrap
-import time
 from typing import TYPE_CHECKING, Any, Optional, Union
 
 import jinja2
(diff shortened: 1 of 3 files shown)


happz:
> Needs tests and I'm not quite happy with the bash script and report. Would mark is as draft, but that would not trigger pipelines.

I believe draft isn't the issue, it's the lack of `ci | full test` label. With it, all checks should run even for drafts.

tmt/checks/avc.py
```diff
@@ -66,36 +65,73 @@ def normalize(
...
-{% if CHECK.test_method.value == 'timestamp' %}
-source {{ MARK_FILEPATH }}
-ausearch -i --input-logs -m AVC -m USER_AVC -m SELINUX_ERR -ts $AVC_SINCE
-{% else %}
-cat {{ MARK_FILEPATH }}
-ausearch --input-logs --checkpoint {{ MARK_FILEPATH }} -m AVC -m USER_AVC -m SELINUX_ERR -i -ts checkpoint
-{% endif %}
-"""  # noqa: E501
-    )
+    textwrap.dedent(f"""
```
happz, line 123: My only comment so far, the mixture of Jinja template and f-strings. I'd pick one of them and stick to it, most likely Jinja. How about passing the "library" variable to `render()` below, or we could have just one script, doing both things in if/else branches, and just set a variable for it when rendering it as before and after test (`ACTION=prepare` vs `ACTION=check` + `if [ "$ACTION" = "prepare" ]; ...`). Both should let us avoid using two template flavors.

martinhoyer: Is this ok? :) 7b51424ba359f863b774925cbd4ecbc0632aaece

LecrisUT:
> Well, unless we want to have dedicated log file for just avc failures, just for the duration of the test.

It has been some time, and I think this is the case now? There have been some improvements on this front. Maybe can check with a [current example](https://artifacts.dev.testing-farm.io/0102b287-0e63-46c9-9ab2-0634c6cc6d9c/) and see if there is anything else that is worth improving? Feel free to give some new feedback or reopen as needed.

PR #3565, opened 2025-03-03
Reset AVC denials after building test images

happz:
Also add timestamps to image building logging.

This should help identify which images cause trouble with AVC denials,
and let us build images whose base images we don't control and can break
our tests.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@
 # Prepare variables
 TMP = $(CURDIR)/tmp
 
-ccred=$(shell tput setaf 1)
-ccgreen=$(shell tput setaf 2)
-ccend=$(shell tput sgr0)
+ccred=$(shell env TERM="$${TERM:-linux}" tput setaf 1)
+ccgreen=$(shell env TERM="$${TERM:-linux}" tput setaf 2)
+ccend=$(shell env TERM="$${TERM:-linux}" tput sgr0)
 
 # Define special targets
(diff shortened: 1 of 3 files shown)


martinhoyer (approved):
Thanks!

KwisatzHaderach (approved):

Makefile
```diff
@@ -146,9 +146,9 @@ images/test/bases:  ## Download base images for custom test images
 
 # Build a single container: <image name> <containerfile>
 define do-build-container-image =
-@ echo "$(ccgreen)Building$(ccend) $(ccred)${1}$(ccend) $(ccgreen)image...$(ccend)"
+@ echo "$(ccred)$$(date '+%x %T')$(ccend) $(ccgreen)Building$(ccend) $(ccred)${1}$(ccend) $(ccgreen)image...$(ccend)"
```
psss, line 149: Could we use date in the `YYY-MM-DD` format? I get this in my locale:

```
6.3.2025 13:15:34 Building tmt/container/test/alpine:latest image...
```

Seems a bit weird.

happz: > Could we use date in the YYY-MM-DD format?

Yes.

> Seems a bit weird.

Yes. Blame your locale of choice ;)

happz: Changed in c988014f

PR #3567, opened 2025-03-04
Add --build-disk-image-only option to bootc plugin

qcheng-redhat:
Add --build-disk-image-only option to bootc plugin

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/bootc.py
+++ b/tmt/steps/provision/bootc.py
@@ -1,6 +1,8 @@
 import os
 from typing import TYPE_CHECKING, Optional, cast
 
+import click
+
 import tmt
 import tmt.base
 import tmt.hardware
@@ -152,6 +154,16 @@ class BootcData(tmt.steps.provision.testcloud.ProvisionTestcloudData):
              """,
     )
 
+    build_disk_image_only: bool = field(
+        default=False,
+        is_flag=True,
+        option='--build-disk-image-only',
+        help="""
+             Only build a bootc disk image from a container image and quit.
+             Guest VM will not start.
+             """,
+    )
+
 
 @tmt.steps.provides_method('bootc')
 class ProvisionBootc(tmt.steps.provision.ProvisionPlugin[BootcData]):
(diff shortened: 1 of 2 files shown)


tmt/steps/provision/bootc.py
```diff
@@ -410,6 +429,9 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         built_image.rename(renamed_image)
         data.image = f"file://{renamed_image}"
 
+        if data.build_disk_image_only:
```
lukaszachy, line 432: Does it makes sense to print path to the generated image?  
If one runs just `tmt run provision -h bootc ... --build-disk-image-only` path is not printed.

qcheng-redhat: Thank you for the review! Yes, make sense. Updated.

tmt/steps/provision/bootc.py
```diff
@@ -152,6 +152,15 @@ class BootcData(tmt.steps.provision.testcloud.ProvisionTestcloudData):
              """,
     )
 
+    build_disk_image_only: bool = field(
+        default=False,
+        is_flag=True,
+        option=('--build-disk-image-only'),
+        help="""
+             Only build a bootc disk image from a container image.
```
happz, line 160: To me, this wasn't clear until I read the code, as it does not say the important part - only build an image, *and will not boot it up*. E.g. "Only build a bootc disk image from a container image and quit. Guest VM will not start." would be clearer to me.

tmt/steps/provision/bootc.py
```diff
@@ -410,6 +429,10 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         built_image.rename(renamed_image)
         data.image = f"file://{renamed_image}"
 
+        if data.build_disk_image_only:
+            self.info(f"The disk image is saved at: {data.image}")
```
happz, line 433: Ha, maybe we should use `self.print()` instead - `info` & co. are rather logging, progress-or-event reporting set of functions, but this piece of information looks like the actual output of the plugin, similar to `tmt * export` or `minute --print-flavors`.

I would also drop the "The disk image ..." part, to make things easier for machine consumers of the output.

thrix:
@qcheng-redhat howdy, this is a must have issue for the release, can you please take a look at the review comments? ty

tmt/steps/provision/bootc.py
qcheng-redhat: Updated. Please review.

tmt/steps/provision/bootc.py
qcheng-redhat: Updated.

qcheng-redhat:
> @qcheng-redhat howdy, this is a must have issue for the release, can you please take a look at the review comments? ty

Thank you for the reminder. Updated.

PR #3569, opened 2025-03-04
Allow to ignore duration in tests

lukaszachy:
There are other ways to adjust this value, but sometimes it is just easier to just let test run.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] include a release note


lukaszachy:
Envar `TMT_PLUGIN_EXECUTE_IGNORE_DURATION=1` is the right way to specify this via envars, right? But it doesn't seem to work for me :/

lukaszachy:
> Envar `TMT_PLUGIN_EXECUTE_IGNORE_DURATION=1` is the right way to specify this via envars, right? But it doesn't seem to work for me :/

Ah, because I didn't have `execute -h tmt` on command line, so  plugin variable didn't work for a default command....
I guess this is something we have to fix at least for this use case, right?

tmt/steps/execute/internal.py
```diff
@@ -303,6 +304,17 @@ class ExecuteInternalData(tmt.steps.execute.ExecuteStepData):
...
     )
+    ignore_duration: bool = field(
+        default=configure_bool_constant(False, 'TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION'),
+        option=('--ignore-duration'),
+        is_flag=True,
+        help="""
+             Ignore test duration value and allow test to run forever.
+             Can be set by ``TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION=1`` even
+             when step is not specified on the commandline. This environment variable
+             will be replaced by plan.fmf or CLI arguments.
```
happz, line 315: * Is this field needed? IIUIC, `ExecuteInternalData` inherits from `ExecuteStepData`, including the fields listed there.
* envvar should be passed in via `envvar` parameter - it will make nicer rendering on command line, in HTML docs, everywhere. Click will be able to consume it too, and it will be mentioned in help, highlighted nicely. When it's incorporated in the help text, nothing like that is possible.

docs/releases.rst
```diff
@@ -5,6 +5,15 @@
 ======================
 
 
+tmt-1.44.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Execute tmt option ``--ignore-duration`` makes tmt to execute
+the test as long as it needs. This can be enabled via environment
+``TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION=1`` even when the execute
+step is not stated on the commandline.
```
happz, line 14: * I'd recommend sticking to "key" rather than "option", because fmf is meant to be the primary source (https://tmt.readthedocs.io/en/stable/contribute.html#docs)
* I'd not include the note about envvar, on the basis that every plugin and step option has this feature. It feels like an exceptional granularity to me.

lukaszachy: > I'd not include the note about envvar, on the basis that every plugin and step option has this feature

I want to make clear that it will work even when plugin is not mentioned on CLI, which in this case is the majority of executions. Otherwise, if one knows this limitation, they will either modify their plans or avoid using option in TF

tmt/utils/__init__.py
```diff
@@ -120,6 +120,24 @@ def configure_constant(default: int, envvar: str) -> int:
         ) from exc
 
 
+def configure_bool_constant(default: bool, envvar: str) -> bool:
```
martinhoyer, line 123: Just a un-educated thought - since we are deducing bool from `1`,`0` str, why not include `true`/`false`, `yes`/`no` (case insensitively)? It could be useful in other parts of the code, no?

tmt/steps/execute/__init__.py
```diff
@@ -709,6 +715,7 @@ def go(
         logger: tmt.log.Logger,
     ) -> None:
         self.go_prolog(logger)
+        logger.verbose('ignore-duration', self.data.exit_first, 'green', level=2)
```
martinhoyer, line 718:
```suggestion
        logger.verbose('ignore-duration', self.data.ignore_duration, 'green', level=2)
```
copy-paste leftover?

martinhoyer (commented):
Nothing useful to add to the option/key, envvar discussion. Maybe just a question about whether there is something related being cooked in @happz's stash that could be merged before this to avoid later changes ;)

tmt/utils/__init__.py
happz: Might be silently allowing unexpected input: Click's environment variable ingestion will recognize only `0` and `1` for boolean flags. tmt recognizing more inputs might lead someone to use `export FOO=true` and it would not work with Click processing the options.

tmt/steps/execute/internal.py
lukaszachy: Took me a while to realize, but why should I add `envvar` when this variable name is produced by step+plugin+option names? Each plugin option should have it, we don't want to extend each field manually ,do we?

tmt/steps/execute/__init__.py
lukaszachy: :facepalm:

docs/releases.rst
lukaszachy: Reworded in field message and add note into overview section in 2051375e

tmt/steps/execute/__init__.py
```diff
@@ -261,6 +261,12 @@ class ExecuteStepData(tmt.steps.WhereableStepData, tmt.steps.StepData):
         option='--duration',
         help='The maximal time allowed for the test to run.',
     )
+    ignore_duration: bool = field(
+        default=False,
+        option=('--ignore-duration'),
```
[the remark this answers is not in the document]
lukaszachy: fixed in 2051375e

tmt/steps/execute/__init__.py
lukaszachy: thanks for spotting this, fixed in 2051375e

tmt/utils/__init__.py
martinhoyer: Ack, thanks for the info!

tmt/steps/execute/internal.py
lukaszachy: removed from ExecuteInternalData in 2051375e

tmt/utils/__init__.py
lukaszachy: Should we raise an error ? If parsed via click...
```
$ TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION=foo tmt_DUR run -a execute -h tmt
Usage: tmt_DUR run execute [OPTIONS]
Try 'tmt_DUR run execute --help' for help.

Error: Invalid value for '--ignore-duration': 'foo' is not a valid boolean.

```

tmt/steps/execute/internal.py
happz: > Took me a while to realize, but why should I add `envvar` when this variable name is produced by step+plugin+option names? Each plugin option should have it, we don't want to extend each field manually ,do we?

Hmm, good point, that is indeed correct. And it's done like this for plugin options, but not for step options. So I guess it doesn't matter that much for now.

happz (approved):

docs/releases.rst
```diff
@@ -5,6 +5,15 @@
 ======================
 
 
+tmt-1.44.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Execute tmt option ``--ignore-duration`` makes tmt to execute
+the test as long as it needs. Execute plugin doesn't need to be
+specified on the commandline for :ref:`plugin-variables` to work
+for this option.
```
psss, line 14: Shouldn't we explicitly mention the name of the environment variable which can be used to tweak this? I think those long names are not quite obvious to everybody.

tests/execute/ignore-duration/test.sh
```diff
@@ -0,0 +1,45 @@
...
+        rlRun "tmt run -vv plan -n /via-plan-false" "2"
+        rlAssertGrep 'errr /demo/test (timeout)' $rlRun_LOG '-F'
+    rlPhaseEnd
+
+    rlPhaseStartTest "With IGNORE_DURATION=1"
+        export TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION=1
+        rlRun "tmt run -vv plan -n /no-option"
+        rlRun "tmt run -vv plan -n /via-plan-true"
+        # ENV should win over CLI or file values, but to be consistent with
+        # reporportal/polarion plugin envar is weaker than plan.fmf
```
psss, line 27: Hm, interesting. Do we have this documented somewhere?

tmt/schemas/execute/tmt.yaml
```diff
@@ -23,6 +23,9 @@ properties:
   exit-first:
     type: boolean
 
+  ignore-duration:
```
psss, line 26: Just thinking aloud, shouldn't we be more specific and mention `test` as well?
```suggestion
  ignore-test-duration:
```

tmt/steps/execute/__init__.py
```diff
@@ -261,6 +262,18 @@ class ExecuteStepData(tmt.steps.WhereableStepData, tmt.steps.StepData):
...
         help='The maximal time allowed for the test to run.',
     )
+    ignore_duration: bool = field(
+        default=configure_bool_constant(False, 'TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION'),
+        option='--ignore-duration',
+        is_flag=True,
+        envvar="TMT_PLUGIN_EXECUTE_TMT_IGNORE_DURATION",
+        help="""
+             Ignore test duration value and allow test to run forever.
+             Can be set by envvar even when step is not specified on the
```
psss, line 272:
```suggestion
             Can be set by environment variable even when step is not specified on the
```

Let's not use abbreviations in the docs.

tmt/steps/execute/__init__.py
```diff
@@ -261,6 +262,18 @@ class ExecuteStepData(tmt.steps.WhereableStepData, tmt.steps.StepData):
...
+             commandline. This environment variable
+             will be replaced by plan.fmf or CLI arguments.
```
psss, line 274:
```suggestion
             will be replaced by fmf config file or CLI arguments.
```

A bit better? `plan.fmf` sounds like an exact file name.

psss (changes requested):
Looks good, thanks! Added a few suggestions.

docs/releases.rst
happz: It was there, I opposed: https://github.com/teemtee/tmt/pull/3569#discussion_r1980884950

tmt/schemas/execute/tmt.yaml
lukaszachy: Had a chat on hacking session about this, not accepting on the basis there is no other 'duration'.
We have just 'Test' duration, other 'kill it if it exceeds some time' use term "timeou" (eg reboot, connect...)

tests/execute/ignore-duration/test.sh
lukaszachy: I have filled #3591 for it now, IMO this is out-of-scope of the PR

docs/releases.rst
lukaszachy: I'll keep it as is.

KwisatzHaderach (approved):

tests/execute/ignore-duration/test.sh
psss: Thanks!

docs/releases.rst
psss: Ok.

psss (approved):

psss:
Rebased, 3 acks, should be ready for merge.

PR #3579, opened 2025-03-05
tmt-report-result should put test output files into subdirectories

happz:
Instead of using `fileprefix` as a prefix of filenames, use it to construct a directory, and put files into it. This should present files organized by their `$TESTNAME` rather thana  flat structure, which becomes hard to navigate in case of tests with many phases and logs.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/execute/scripts/tmt-report-result
+++ b/tmt/steps/execute/scripts/tmt-report-result
@@ -36,14 +36,15 @@ copy_outputfile_to_data_dir () {
     # Cut first char if underscore.
     [[ ${fileprefix:0:1} == '_' ]] && fileprefix=$(echo "$fileprefix" | cut -c 2-)
     # Construct the final outputfile.
-    finalOutputFile="${TMT_TEST_DATA}/${fileprefix}/${filename}.txt"
+    relativeFinalOutputFile="${fileprefix}/${filename}.txt"
+    finalOutputFile="${TMT_TEST_DATA}/${relativeFinalOutputFile}"
     # make sure the directory exists.
     mkdir -p "$(dirname "$finalOutputFile")"
     # Copy outputfile to data dir.
     cp -f "$outputFile" "$finalOutputFile"
     # Make sure that the log file is readable for guest.pull()
     chmod a+r "$finalOutputFile"
-    outputFile="$finalOutputFile"
+    outputFile="${relativeFinalOutputFile}"
 }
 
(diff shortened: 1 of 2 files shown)


tmt/steps/execute/scripts/tmt-report-result
martinhoyer, line 1: I'm not able to 'suggest' on the whole function, so here:
```
copy_outputfile_to_data_dir() {
    # Extract the filename from the output file path
    filename=$(basename "$outputFile")
    # Replace slashes and spaces in test name with underscores
    fileprefix=$(echo "$TESTNAME" | tr '/ ' '_')
    # Remove leading underscore if present
    [[ ${fileprefix:0:1} == '_' ]] && fileprefix=${fileprefix:1}
    # Construct the final outputfile path
    finalOutputFile="${TMT_TEST_DATA}/${fileprefix}/${filename}.txt"
    # Copy outputfile to data dir and set proper permissions in one step
    install -D -m 644 "$outputFile" "$finalOutputFile"
    # Update the outputFile variable to the new location
    outputFile="$finalOutputFile"
}
```
Mostly interested in using `install` instead of `cp`, wdyt?

martinhoyer (approved):

tmt/steps/execute/scripts/tmt-report-result
happz: No opinion. Why is `install` better than `cp`?

therazix (approved):

tmt/steps/execute/scripts/tmt-report-result
martinhoyer: Apologies, missed this. 
Not my forte and don't remember, but probably was thinking single command instead of mkdir, cp and chmod. Could be rsync as well I guess. Non issue, just a weak suggestion.

fwiw, there is also no `awk`.

martinhoyer (approved):

skycastlelily (approved):
LGTM^^

tmt/steps/execute/scripts/tmt-report-result
happz: I don't have much time to invest into improving this code. The goal was to address a proposal from a user, and this felt like the simplest change possible. Feel free to file an issue if you think it's worth switching to `install` (or another approach), I for one would move on in this PR.

happz:
Unrelated failures, merging.

Issue #3558, opened 2025-02-27
File bomb occurs when invoking `tmt run --last report`


pholica:
Whenever `tmt run --last report` is called, the whole test repository is copied to every plan in `/var/tmp/tmt/run-XXX`. It's not a big deal in case of small repos, however I ran out of inodes on my filesystem when I executed TMT with larger repository where each run had over 50000 files in it.

Steps to reproduce:

1. tmt run -a
2. check number of files in created /var/tmp/tmt (eg. by running `find /var/tmp/tmt/run-XXX | wc -l`)
3. run tmt run --last report
4. check number of files in created /var/tmp/tmt (eg. by running `find /var/tmp/tmt/run-XXX | wc -l`)

Example from some test repo (https://github.com/pholica/tmt_testrepo/):
```
$ tmt run -a provision --feeling-safe --how local plans --name plan1
/var/tmp/tmt/run-001

/plans/example/plan1
    discover
        how: fmf
        directory: /home/pholica/git/pholica/tmt_testrepo
        summary: 1 test selected
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: local
        primary address: localhost
        multihost name: default-0
        arch: x86_64
        distro: Fedora Linux 41 (Forty One)
    
        summary: 1 guest provisioned
    prepare
        queued push task #1: push to default-0
        
        push task #1: push to default-0
    
        queued prepare task #1: essential-requires on default-0
        
        prepare task #1: essential-requires on default-0
        how: install
        summary: Install essential required packages
        name: essential-requires
        where: default-0
        package: /usr/bin/flock
    
        queued pull task #1: pull from default-0
        
        pull task #1: pull from default-0
    
        summary: 1 preparation applied
    execute
        queued execute task #1: default-0 on default-0
        
        execute task #1: default-0 on default-0
        how: tmt
        progress:                                     
    
        summary: 1 test executed
    report
        how: display
        summary: 1 test passed
    finish
    
        summary: 0 tasks completed

total: 1 test passed
$ find /var/tmp/tmt/run-001 | wc -l
38
$ find /var/tmp/tmt/run-001 -name LICENSE
$ tmt run --last report --how display
/var/tmp/tmt/run-001

/plans/example/plan1
    report
        how: display
        summary: 1 test passed

total: 1 test passed
$ find /var/tmp/tmt/run-001 | wc -l
57
$ find /var/tmp/tmt/run-001 -name LICENSE
[...]

PR #3581, opened 2025-03-05
Refactoring the file copy operations - reflinks with fallback

martinhoyer:
More or less follow-up on https://github.com/teemtee/tmt/pull/2791  
The way tmt is copying files appears to be quite a performance bottleneck in some scenarios and can cause issues like https://github.com/teemtee/tmt/issues/3558 (well, reflinks wouldn't help with running out of inodes there). 

What makes `reflinks` interesting, is that it behaves the same as normal `cp` operation, but doesn't take additional storage space, unless the file is changed.

They do need to be supported by the underlying filesystem. Notably:
✅ [Btrfs](https://btrfs.readthedocs.io/en/latest/Reflink.html) - Fedora default since F33
✅ XFS - reflinks supported since CentOS Stream 8

:x: ext4 - Ubuntu GitHub action runners
:x: EFS - AWS 

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/utils/filesystem.py
+++ b/tmt/utils/filesystem.py
@@ -65,6 +65,7 @@ def copy_tree(
     Copy directory efficiently, trying different strategies.
 
     Attempts strategies in order:
+
     #. ``cp -a --reflink=auto`` (copy-on-write, with ``cp``'s own
        fallback).
 
--- a/tmt/utils/filesystem.py
+++ b/tmt/utils/filesystem.py
@@ -21,17 +21,14 @@ def _copy_tree_cp(
     not supported by the filesystem.
     Returns True if successful, False if `cp` command fails with `RunError`.
     """
-    cmd = ['cp', '-a', '--reflink=auto']
-
-    logger.debug(f"Attempting copy from '{src}' to '{dst}' with {cmd}")
-
     try:
         # The '/./' at the end of the source path tells cp to copy the *contents* of the directory
         # rather than creating a new subdirectory in the destination
-        Command(*cmd, f"{src}/./", str(dst)).run(cwd=None, logger=logger, join=True, silent=True)
+        Command('cp', '-a', '--reflink=auto', f"{src}/./", str(dst)).run(
+            cwd=None, logger=logger, join=True, silent=True
+        )
         return True
-    except RunError as error:
-        logger.debug(f"cp copy command failed: {error}, falling back")
+    except RunError:
         return False
     # Let other exceptions (e.g. permissions, disk full) propagate
 
@@ -44,15 +41,10 @@ def _copy_tree_shutil(
     """
     Perform copy using shutil.copytree.
 
-    This is typically a fallback strategy. It ensures that the destination
-    directory `dst` and its parents exist before attempting the copy.
+    This is typically a fallback strategy. The destination directory must
+    exist before calling this function.
     """
     logger.debug(f"Performing shutil.copytree from '{src}' to '{dst}'")
-
-    # Ensure destination directory `dst` and its parents exist.
-    # shutil.copytree with dirs_exist_ok=True will then copy contents into `dst`,
-    # merging if `dst` already contains files (e.g. from a previous failed attempt).
-    dst.mkdir(parents=True, exist_ok=True)
 
     shutil.copytree(
         src,
@@ -112,6 +104,7 @@ def copy_tree(
 
     # 1. Try 'cp -a --reflink=auto' copy.
     #    'cp' itself handles fallback from reflink to standard copy if reflink=auto is used.
+    logger.debug(f"Attempting copy from '{src}' to '{dst}' using 'cp' with reflink")
     if _copy_tree_cp(src, dst, logger):
         logger.debug(
             "Copy finished using 'cp -a --reflink=auto' strategy (or its internal fallback)."
@@ -119,7 +112,7 @@ def copy_tree(
         return
 
     # 2. Fallback to shutil.copytree
-    logger.debug("Falling back to shutil.copytree strategy.")
+    logger.debug("cp command failed, falling back to shutil.copytree strategy.")
     try:
         _copy_tree_shutil(src, dst, logger)
         logger.debug("Copy finished using shutil.copytree strategy.")
--- a/tmt/utils/filesystem.py
+++ b/tmt/utils/filesystem.py
@@ -72,10 +72,26 @@ def copy_tree(
 
     Attempts strategies in order:
     1. 'cp -a --reflink=auto' (copy-on-write, with cp's own fallback).
+       - Reflinks provide fast, space-efficient copies that behave like normal copies
+       - They don't use additional storage space unless the file is modified
+       - Supported on btrfs (Fedora default since F33) and XFS (CentOS Stream 8+)
+       - Using '--reflink=auto' means 'cp' automatically falls back to standard copy
+         if reflinks aren't supported by the filesystem
     2. shutil.copytree as a final fallback.
+       - Used if the 'cp' command fails for any reason
+       - Maintains symlinks (symlinks=True)
+       - Merges with existing destination directories (dirs_exist_ok=True)
 
     Symlinks are always preserved. The destination directory `dst` and its
-    parents will be created if they do not exist.
+    parents will be created if they do not exist. File permissions and timestamps
+    are preserved in all copy strategies.
+
+    Example usage:
+        # Copy a directory tree with all its content
+        copy_tree(Path("/path/to/source"), Path("/path/to/destination"), logger)
+
+        # Copy with relative paths
+        copy_tree(workdir / "original", workdir / "backup", logger)
 
     :param src: Source directory path. Must exist and be a directory.
     :param dst: Destination directory path.
--- a/tmt/utils/filesystem.py
+++ b/tmt/utils/filesystem.py
@@ -1,0 +1,116 @@
+"""
+Utility functions for filesystem operations.
+"""
+
+import shutil
+
+import tmt.log
+from tmt._compat.pathlib import Path
+from tmt.utils import Command, GeneralError, RunError
+
+
+def _copy_tree_cp(
+    src: Path,
+    dst: Path,
+    logger: tmt.log.Logger,
+) -> bool:
+    """
+    Attempt to copy directory using 'cp -a --reflink=auto'.
+
+    The 'cp' command itself will fall back to a standard copy if reflink is
+    not supported by the filesystem.
+    Returns True if successful, False if `cp` command fails with `RunError`.
+    """
+    cmd = ['cp', '-a', '--reflink=auto']
+
+    logger.debug(f"Attempting copy from '{src}' to '{dst}' with {cmd}")
+
+    try:
+        # The '/./' at the end of the source path tells cp to copy the *contents* of the directory
+        # rather than creating a new subdirectory in the destination
+        Command(*cmd, f"{src}/./", str(dst)).run(cwd=None, logger=logger, join=True, silent=True)
+        return True
+    except RunError as error:
+        logger.debug(f"cp copy command failed: {error}, falling back")
+        return False
+    # Let other exceptions (e.g. permissions, disk full) propagate
+
+
+def _copy_tree_shutil(
+    src: Path,
+    dst: Path,
+    logger: tmt.log.Logger,
+) -> None:
+    """
+    Perform copy using shutil.copytree.
+
+    This is typically a fallback strategy. It ensures that the destination
+    directory `dst` and its parents exist before attempting the copy.
+    """
+    logger.debug(f"Performing shutil.copytree from '{src}' to '{dst}'")
+
+    # Ensure destination directory `dst` and its parents exist.
+    # shutil.copytree with dirs_exist_ok=True will then copy contents into `dst`,
+    # merging if `dst` already contains files (e.g. from a previous failed attempt).
+    dst.mkdir(parents=True, exist_ok=True)
+
+    shutil.copytree(
+        src,
+        dst,
+        symlinks=True,
+        dirs_exist_ok=True,
+    )
+
+
+def copy_tree(
+    src: Path,
+    dst: Path,
+    logger: tmt.log.Logger,
+) -> None:
+    """
+    Copy directory efficiently, trying different strategies.
+
+    Attempts strategies in order:
+    1. 'cp -a --reflink=auto' (copy-on-write, with cp's own fallback).
+    2. shutil.copytree as a final fallback.
+
+    Symlinks are always preserved. The destination directory `dst` and its
+    parents will be created if they do not exist.
+
+    :param src: Source directory path. Must exist and be a directory.
+    :param dst: Destination directory path.
+    :param logger: Logger to use for debug messages.
+    :raises GeneralError: when copying fails using all strategies, or if `src`
+                         does not exist or is not a directory.
+    """
+    logger.debug(f"Copying directory tree from '{src}' to '{dst}'")
+
+    if not src.is_dir():
+        # Add an explicit check for src, as 'cp' or 'shutil.copytree' might give
+        # less clear or varied errors. This ensures a consistent error message.
+        raise GeneralError(f"Source '{src}' for copy_tree is not a directory or does not exist.")
+
+    # Ensure destination directory `dst` and its parents exist.
+    # This is crucial for `cp` and helpful for `shutil.copytree` (as it creates parent dirs).
+    dst.mkdir(parents=True, exist_ok=True)
+
+    # 1. Try 'cp -a --reflink=auto' copy.
+    #    'cp' itself handles fallback from reflink to standard copy if reflink=auto is used.
+    if _copy_tree_cp(src, dst, logger):
+        logger.debug(
+            "Copy finished using 'cp -a --reflink=auto' strategy (or its internal fallback)."
+        )
+        return
+
+    # 2. Fallback to shutil.copytree
+    logger.debug("Falling back to shutil.copytree strategy.")
+    try:
+        _copy_tree_shutil(src, dst, logger)
+        logger.debug("Copy finished using shutil.copytree strategy.")
+    except Exception as error:
+        # Catching a broad Exception here because shutil.copytree can raise various errors
+        # (e.g., OSError, FileExistsError if dst is a file after mkdir, etc.)
+        # and we want to wrap them all in GeneralError.
+        raise GeneralError(
+            f"Failed to copy directory tree from '{src}' to '{dst}' using all strategies."
+        ) from error
--- a/tmt/utils/filesystem.py
+++ b/tmt/utils/filesystem.py
@@ -15,11 +15,13 @@ def _copy_tree_cp(
     logger: tmt.log.Logger,
 ) -> bool:
     """
-    Attempt to copy directory using 'cp -a --reflink=auto'.
+    Attempt to copy directory using ``cp -a --reflink=auto``.
 
-    The 'cp' command itself will fall back to a standard copy if reflink is
-    not supported by the filesystem.
-    Returns True if successful, False if `cp` command fails with `RunError`.
+    The ``cp`` command itself will fall back to a standard copy if
+    reflink is not supported by the filesystem.
+
+    :returns: ``True`` if successful, ``False`` if ``cp`` command fails
+        with :py:class:`RunError`.
     """
     try:
         # The '/./' at the end of the source path tells cp to copy the *contents* of the directory
@@ -63,22 +65,32 @@ def copy_tree(
     Copy directory efficiently, trying different strategies.
 
     Attempts strategies in order:
-    1. 'cp -a --reflink=auto' (copy-on-write, with cp's own fallback).
-       - Reflinks provide fast, space-efficient copies that behave like normal copies
-       - They don't use additional storage space unless the file is modified
-       - Supported on btrfs (Fedora default since F33) and XFS (CentOS Stream 8+)
-       - Using '--reflink=auto' means 'cp' automatically falls back to standard copy
-         if reflinks aren't supported by the filesystem
-    2. shutil.copytree as a final fallback.
-       - Used if the 'cp' command fails for any reason
-       - Maintains symlinks (symlinks=True)
-       - Merges with existing destination directories (dirs_exist_ok=True)
+    #. ``cp -a --reflink=auto`` (copy-on-write, with ``cp``'s own
+       fallback).
+
+       * Reflinks provide fast, space-efficient copies that behave like
+         normal copies
+       * They don't use additional storage space unless the file is
+         modified
+       * Supported on btrfs (Fedora default since F33) and XFS (CentOS
+         Stream 8+)
+       * Using ``--reflink=auto`` means ``cp`` automatically falls back
+         to standard copy if reflink isn't supported by the filesystem
+
+    #. :py:func:`shutil.copytree` as a final fallback.
+
+       * Used if the ``cp`` command fails for any reason
+       * Maintains symlinks (``symlinks=True``)
+       * Merges with existing destination directories (``dirs_exist_ok=True``)
 
     Symlinks are always preserved. The destination directory `dst` and its
     parents will be created if they do not exist. File permissions and timestamps
     are preserved in all copy strategies.
 
     Example usage:
+
+    .. code-block:: python
+
         # Copy a directory tree with all its content
         copy_tree(Path("/path/to/source"), Path("/path/to/destination"), logger)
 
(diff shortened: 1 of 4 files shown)


tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+    for item in src_path.rglob('*'):
+        relative_path = item.relative_to(src_path)
+        dst_item = dst_path / relative_path
+
+        # Create a path in cache using a hash of the relative path to avoid path length issues
+        cache_key = relative_path_to_cache_key(relative_path)
+        cache_path = cache_dir / cache_key
+
+        if item.is_dir():
+            dst_item.mkdir(parents=True, exist_ok=True)
```
happz, line 86: I don't understand this bit: what happens to the content of `item`, being a directory?

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+            logger.debug(f"Reflink copy failed: {error}, falling back to hardlink strategy")
+
+    # Fallback to hardlink-based copy for unchanged files when reflinks aren't available
+    # This fallback is particularly important for CI/CD environments like GitHub Actions
+    # which typically run on Ubuntu with ext4 filesystems that don't support reflinks.
+    # Without this fallback, workflows running in such environments would experience
+    # significantly higher inode consumption and potential performance issues.
+    # Use a persistent cache for hardlinks
+    cache_dir = Path(tempfile.gettempdir()) / 'tmt_file_cache'
+    cache_dir.mkdir(exist_ok=True)
```
happz, line 69: When is `cache_dir` going to be removed?

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+        src: Source directory path
+        dst: Destination directory path
+        logger: Logger to use for debug messages (optional)
+    """
+    src_path = Path(src)
+    dst_path = Path(dst)
+
+    # Try reflink copy first (supported by btrfs, xfs with reflink, and some other filesystems)
+    try:
+        if logger:
```
happz, line 40: Logger should be mandatory, if there are debug messages, let's get them logged.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+    falling back to hardlinks for unchanged files.
+
+    This function aims to reduce inode consumption by leveraging
+    filesystem-specific features like reflinks (copy-on-write)
+    when available. Symlinks are always preserved.
+
+    Args:
+        src: Source directory path
+        dst: Destination directory path
+        logger: Logger to use for debug messages (optional)
```
happz, line 33: The more spread format is `:param foo: description...`, with no `Args`.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+    hardlink_count = 0
+    regular_copy_count = 0
+
+    for item in src_path.rglob('*'):
+        relative_path = item.relative_to(src_path)
+        dst_item = dst_path / relative_path
+
+        # Create a path in cache using a hash of the relative path to avoid path length issues
+        cache_key = relative_path_to_cache_key(relative_path)
+        cache_path = cache_dir / cache_key
```
happz, line 83: `cache_path = cache_dir / relative_path_to_cache_key(relative_path)` ? `cache_key` does not seem to be used anywhere else.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+                f"Directory copied using hardlinks: {hardlink_count} hardlinked, "
+                f"{regular_copy_count} regular copies"
+            )
+        else:
+            logger.debug("Directory copied using regular copy method")
+
+    return
+
+
+def relative_path_to_cache_key(relative_path: Union[str, Path]) -> str:
```
happz, line 146: `hashlib` is full of hashing functions, why not using one from this standard library package?

martinhoyer:
@happz hold on, it's still a draft :)

tmt/utils/filesystem.py
martinhoyer: again, still a draft - especially the hardlink, cache.

tmt/utils/filesystem.py
martinhoyer: Right, leftover from a change. Thanks.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,161 @@
...
+import zlib
+from typing import Optional, Union
+
+import tmt.log
+from tmt._compat.pathlib import Path
+
+
+def copy_tree(
+    src: Union[str, Path],
+    dst: Union[str, Path],
```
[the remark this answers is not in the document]
martinhoyer: Make sense, I was thinking about shutil.copytree, which is dumb.

tmt/utils/filesystem.py
martinhoyer: I have no good solution at this point, other than it being in tmpfs? Ideas welcome :)

tmt/utils/filesystem.py
martinhoyer: addressed

tmt/utils/filesystem.py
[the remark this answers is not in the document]
martinhoyer: addressed

tmt/utils/filesystem.py
martinhoyer: addressed

tmt/utils/filesystem.py
martinhoyer: To be honest I'm not entirely confident about the whole "fallback" code. Hoping someone more knowledgeable can validate it. Too bad testing-farm is running on aws filesystem and it's kinda needed.

happz: I wouldn't blame just AWS. I think we shouldn't lock us with supporting just two very modern filesystems. ext4 is by no means rare.

martinhoyer: No blame at all. I meant it more like we could keep the existing shutil.copytree as fallback instead if of this custom cache that could be hard to maintain.

martinhoyer:
@coderabbitai full review

tmt/utils/filesystem.py
happz: I can't find the answer: what happens to directories?

```python
        if item.is_dir():
            dst_item.mkdir(parents=True, exist_ok=True)
            # Copy directory attributes (mode, timestamps)
            shutil.copystat(item, dst_item)
```

But I can't find a code that will copy the content of the directory. Am I missing something?

BTW

martinhoyer: @happz trying something new, but barely know what I'm doing. Does https://github.com/teemtee/tmt/pull/3581/commits/ad71da2df1142239b5c5d1cf2aea57c750f87a36 make any sense at all? 

Initially, I only wanted to do opportunistic reflinks - a few lines of code, but the implementing the fallback "caching" - I'm too dumb for this.

tmt/utils/filesystem.py
martinhoyer: the idea in https://github.com/teemtee/tmt/commit/ad71da2df1142239b5c5d1cf2aea57c750f87a36 is that we'd use `/var/tmp/tmt/cache` (or whatever is work root dir) and have tmt clean --cache? Otherwise I don't know - if we use unique temp dir, it would be culled in Finish prune?

martinhoyer:
@coderabbitai review

tmt/utils/filesystem.py
happz: TBH, I'm getting lost as well, and I would recommend going for smaller goals here, before it gets out of hand and you leave it.

IMO, too much stuff is happening in a single function - hard to navigate, hard to test, and also hard to get everything in place when trying to address several issues at once. How about starting with just the reflinks and `shutil.copy()` as a fallback? You can always add the hardlink/cache method as a follow-up patch, in isolation, and easier to test and review.

```python
def _copy_tree_reflinks(
    src: Path,
    dst: Path,
    logger: 'tmt.log.Logger',
    workdir_root: Path,
) -> bool:
  # Do the magic, return True on success, False otherwise.

def _copy_tree_basic(
    src: Path,
    dst: Path,
    logger: 'tmt.log.Logger',
    workdir_root: Path,
) -> bool:
  shutil.copy(...)

  return True

def copy_tree(
    src: Path,
    dst: Path,
    logger: 'tmt.log.Logger',
    workdir_root: Path,
) -> None:
    if _copy_tree_reflink(...):
        return

    # Maybe we will have more methods, like the one with hardlinks and cache
    ...

    # Everything else failed, fall back to the basics.
    _copy_tree_basic(...)

Comment (martinhoyer) [reply tmt/utils/filesystem.py:104]: > How about starting with just the reflinks and shutil.copy() as a fallback?

Yes please! :) I'll do that.

Comment (happz) [reply tmt/utils/filesystem.py:104]: I would start this "small". I can also imagine an envvar in `copy_tree()`, e.g. `TMT_COPY_TREE_METHOD=reflink tmt run a-big-fat-kernel-test-plan`, which would let us run and compare different methods. But that can also wait, starting small, internal `copy_tree` wrapper + two methods, one of them `shutils.copy_tree()`, that should be a nice improvement and cleanup.

File: pyproject.toml [hunk 8]
```diff
@@ -42,6 +42,7 @@ dependencies = [              # F39 / PyPI
     "ruamel.yaml>=0.16.6",    # 0.17.32 / 0.17.32
     "urllib3>=1.26.5, <3.0",  # 1.26.16 / 2.0.4
     "typing-extensions>=4.9.0; python_version < '3.13'",
+    "filelock>=3.7.1",
```
martinhoyer, line 45: This wouldn't be needed if we keep hardlink disabled

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,332 @@
...
+    Returns True on successful completion.
+    """
+    logger.debug(f"Performing basic copy from '{src}' to '{dst}'")
+
+    if not dst.exists():
+        dst.mkdir(parents=True, exist_ok=True)
+
+    if not src.exists():
+        logger.debug(f"Source directory '{src}' does not exist, skipping basic copy")
+        return True
```
happz, line 242: (I thought I did submit these comments, but I don't see them anywhere in the PR, so, I probably didn't...)

Would it make sense to move these two "checks" into `copy_tree()`? It seems both methods check whether source exists and create the destination, so it could happen before we start copying anything.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,332 @@
...
+    # (e.g., cache dir issue) or if we explicitly add conditions to skip hardlinks.
+    logger.debug("Falling back to basic copy strategy.")
+    if _copy_tree_basic(src, dst, logger, workdir_root):
+        logger.debug("Copy finished using basic copy strategy.")
+        return
+
+    # Should not be reached if _copy_tree_basic always returns True on completion,
+    # unless it encounters an error and returns False.
+    logger.warning(f"All copy strategies failed for '{src}' to '{dst}'.")
+    # Raise exception here?
```
happz, line 317: I think we should raise an exception, we were tasked with copying a tree and we failed.

I'm not sure about individual methods and them catching exceptions. The reflink does have a genuine interest in catching the "reflink not supported" issue, and turning it into `False`, but the rest of the errors should probably not result in a fallback. Like running out of disk space. For example:

```python
    except Exception as e:
        logger.warning(f"Basic copy failed during processing: {e}")
        return False
```

I'd argue this should be an exception, e.g. `GeneralError(...) from e`. And I would even say that these methods should not bother with exceptions unless they can do something better than logging and reraising them, I'd think that `copy_tree()` could wrap calls with `try` and `except` and wrap whatever is raised in individual methods with `GeneralError("Failed to copy tree") from exc` or something like that.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Use cp with --reflink=auto which falls back to regular copy if reflinks not supported
+        # The '/./' at the end of the source path tells cp to copy the *contents* of the directory
+        # rather than creating a new subdirectory in the destination
+        subprocess.run(
+            ['cp', '-a', '--reflink=auto', f"{src}/./", str(dst)],
+            check=True,
+            stderr=subprocess.PIPE,
+        )
+        logger.debug("Directory copied using reflink")
```
happz, line 48: Method logs success, `copy_tree()` logs success as well, I would opt for dropping it from here.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,332 @@
...
+        if not src.exists():
+            logger.debug(f"Source directory '{src}' does not exist, skipping reflink copy")
+            return False  # Indicate failure as source doesn't exist
+
+        dst.mkdir(parents=True, exist_ok=True)
+
+        # Use cp with --reflink=auto which falls back to regular copy if reflinks not supported
+        # The '/./' at the end of the source path tells cp to copy the *contents* of the directory
+        # rather than creating a new subdirectory in the destination
+        subprocess.run(
```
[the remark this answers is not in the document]
martinhoyer: Thanks, I'm dumb. Using Command in https://github.com/teemtee/tmt/pull/3581/commits/88c1c15bf66cde2e1942ca9732403bae7bc74a39

tmt/utils/filesystem.py
martinhoyer: Dropped

martinhoyer:
@coderabbitai full review

tmt/utils/filesystem.py
martinhoyer: Nice, thank you! https://github.com/teemtee/tmt/pull/3581/commits/88c1c15bf66cde2e1942ca9732403bae7bc74a39

martinhoyer:
@coderabbitai configuration

tmt/utils/filesystem.py
happz: Yep, absolutely.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Handles files, directories, and symlinks using standard shutil operations.
+    Raises exceptions on failure.
+    """
+    logger.debug(f"Performing basic copy from '{src}' to '{dst}'")
+
+    copied_count = 0
+    for item in src.rglob('*'):
+        relative_path = item.relative_to(src)
+        dst_item = dst / relative_path
```
happz, line 56: Having `item` and `dst_item` - would you mind renaming `item` to `src_item`? We have a special name for *dst* item, but not for *src* item, that really does not feel right.

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,113 @@
...
+import tmt.log
+from tmt._compat.pathlib import Path
+from tmt.utils import Command, GeneralError, RunError
+
+
+def _copy_tree_reflinks(
+    src: Path,
+    dst: Path,
+    logger: 'tmt.log.Logger',
+    workdir_root: Path,  # Keep signature consistent, even if not used here
```
happz, line 17: Is it actually used by any code in this file? I see it as a parameter, and being passed to calls to `_copy_tree_*`, but that seems to be it. A relict of the caching method? Dropping it would make the patch simpler, less changes in calls to `copy_tree()`.

martinhoyer:
I'm dumb, the 3 fallbacks are useless, as there cp with `--reflink=auto` already falls back to non-reflink `cp`

martinhoyer:
Aaand done. Hopefully last refactor - so sorry!  
Had to squash things, as it became too messy even for my PR. 

Thinking about @jscotka's feedback - I think it should realistically never get to the shutil.copytree fallback, as `cp --reflink=auto` will automatically falls-back to 'hardlink' when filesystem is missing support. Regardless, it should be as close to cp behaviour as is possible with shutil, by using symlinks=True and making sure the destination dir exists. 

@therazix The workdir should be gone finally :)

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,116 @@
...
+
+    This is typically a fallback strategy. It ensures that the destination
+    directory `dst` and its parents exist before attempting the copy.
+    """
+    logger.debug(f"Performing shutil.copytree from '{src}' to '{dst}'")
+
+    # Ensure destination directory `dst` and its parents exist.
+    # shutil.copytree with dirs_exist_ok=True will then copy contents into `dst`,
+    # merging if `dst` already contains files (e.g. from a previous failed attempt).
+    dst.mkdir(parents=True, exist_ok=True)
```
happz, line 55: Isn't the very same call, with different comment, in `copy_tree()`?

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Ensure destination directory `dst` and its parents exist.
+    # This is crucial for `cp` and helpful for `shutil.copytree` (as it creates parent dirs).
+    dst.mkdir(parents=True, exist_ok=True)
+
+    # 1. Try 'cp -a --reflink=auto' copy.
+    #    'cp' itself handles fallback from reflink to standard copy if reflink=auto is used.
+    if _copy_tree_cp(src, dst, logger):
+        logger.debug(
+            "Copy finished using 'cp -a --reflink=auto' strategy (or its internal fallback)."
```
happz, line 101: This is a bit schizofrenic: the failure is logged by `_copy_tree_cp`, `logger.debug(f"cp copy command failed: {error}, falling back")`, but success is logged here. Plus, `_copy_tree_cp` mentions a fallback - what if there's no fallback available or enabled? `_copy_tree_cp` definitely does not control whether there will be any fallback or not - it tries to copy stuff, and either it succeeds or fails, and returns a value. What happens next is out of its scope, we may very well decide the reflink is the only way and remove all other methods - and we would need to dive into `_copy_tree_cp` to remove the mention of a fallback.

The logging kind of peeks beyond the scope of that single function, so I would recommend doing logging here: `copy_tree` will share with the world whether the given method succeeded, or whether it failed and there will be a fallback, or whether that's all we were able to do and we're giving up and raising an exception.

So, in my view, `logger.debug(f"Attempting copy from '{src}' to '{dst}' with {cmd}")` and `logger.debug(f"Performing shutil.copytree from '{src}' to '{dst}'")` are good and where they should be, but `logger.debug(f"cp copy command failed: {error}, falling back")` isn't, and `copy_tree` itself should do the reporting on whether a method failed and what happens next, like it does now with messages like `"Copy finished using 'cp -a --reflink=auto' strategy (or its internal fallback)."`, they are good.

martinhoyer:
@happz did some changes based on the feedback, hopefully some of it was addressed without introducing more issues, but not getting my hopes up :)

tmt/utils/filesystem.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Copy with relative paths
+        copy_tree(workdir / "original", workdir / "backup", logger)
+
+    :param src: Source directory path. Must exist and be a directory.
+    :param dst: Destination directory path.
+    :param logger: Logger to use for debug messages.
+    :raises GeneralError: when copying fails using all strategies, or if `src`
+                         does not exist or is not a directory.
+    """
```
happz, line 93: A couple of changes for better docstrings. It started as a fix to docstrings breaking the documentation build, turned into polishing & taking advantage of ReST features.

```diff
diff --git a/tmt/utils/filesystem.py b/tmt/utils/filesystem.py
index 3852afd2..11ed5f9a 100644
--- a/tmt/utils/filesystem.py
+++ b/tmt/utils/filesystem.py
@@ -15,11 +15,13 @@ def _copy_tree_cp(
     logger: tmt.log.Logger,
 ) -> bool:
     """
-    Attempt to copy directory using 'cp -a --reflink=auto'.
+    Attempt to copy directory using ``cp -a --reflink=auto``.

-    The 'cp' command itself will fall back to a standard copy if reflink is
-    not supported by the filesystem.
-    Returns True if successful, False if `cp` command fails with `RunError`.
+    The ``cp`` command itself will fall back to a standard copy if
+    reflink is not supported by the filesystem.
+
+    :returns: ``True`` if successful, ``False`` if ``cp`` command fails
+        with :py:class:`RunError`.
     """
     try:
         # The '/./' at the end of the source path tells cp to copy the *contents* of the directory
@@ -63,22 +65,33 @@ def copy_tree(
     Copy directory efficiently, trying different strategies.

     Attempts strategies in order:
-    1. 'cp -a --reflink=auto' (copy-on-write, with cp's own fallback).
-       - Reflinks provide fast, space-efficient copies that behave like normal copies
-       - They don't use additional storage space unless the file is modified
-       - Supported on btrfs (Fedora default since F33) and XFS (CentOS Stream 8+)
-       - Using '--reflink=auto' means 'cp' automatically falls back to standard copy
-         if reflinks aren't supported by the filesystem
-    2. shutil.copytree as a final fallback.
-       - Used if the 'cp' command fails for any reason
-       - Maintains symlinks (symlinks=True)
-       - Merges with existing destination directories (dirs_exist_ok=True)
-
-    Symlinks are always preserved. The destination directory `dst` and its
-    parents will be created if they do not exist. File permissions and timestamps
-    are preserved in all copy strategies.
+
+    #. ``cp -a --reflink=auto`` (copy-on-write, with ``cp``'s own
+       fallback).
+
+       * Reflinks provide fast, space-efficient copies that behave like
+         normal copies
+       * They don't use additional storage space unless the file is
+         modified
+       * Supported on btrfs (Fedora default since F33) and XFS (CentOS
+         Stream 8+)
+       * Using ``--reflink=auto`` means ``cp`` automatically falls back
+         to standard copy if reflinks aren't supported by the filesystem
+
+    #. :py:func:`shutil.copytree` as a final fallback.
+
+       * Used if the ``cp`` command fails for any reason
+       * Maintains symlinks (``symlinks=True``)
+       * Merges with existing destination directories (``dirs_exist_ok=True``)
+
+    Symlinks are always preserved. The destination directory ``dst`` and
+    its parents will be created if they do not exist. File permissions
+    and timestamps are preserved in all copy strategies.

     Example usage:
+
+    .. code-block:: python
+
         # Copy a directory tree with all its content
         copy_tree(Path("/path/to/source"), Path("/path/to/destination"), logger)

@@ -88,8 +101,8 @@ def copy_tree(
     :param src: Source directory path. Must exist and be a directory.
     :param dst: Destination directory path.
     :param logger: Logger to use for debug messages.
-    :raises GeneralError: when copying fails using all strategies, or if `src`
-                         does not exist or is not a directory.
+    :raises GeneralError: when copying fails using all strategies, or if
+        ``src`` does not exist or is not a directory.
     """
     logger.debug(f"Copying directory tree from '{src}' to '{dst}'")
```

martinhoyer: There needs to be a linter check for this. Especially if we are not doing markdown.

happz: I'm not aware of a ReST linter. There can be one, don't mind using it. -1 from me to markdown, though, ReST is kind of a lingua franca of the Python world, thanks to Sphinx and wide adoption, including Python's own standard library. Feel free to raise the point on the meeting, but I don't see us switching to another markup language.

martinhoyer:
Another few hours wasted trying to decipher how to work with tmt tests without any progress. Giving up.

martinhoyer:
> /home/runner/work/tmt/tmt/tmt/utils/filesystem.py:docstring of tmt.utils.filesystem.copy_tree:5:Unexpected indentation.

of course

happz:
> Another few hours wasted trying to decipher how to work with tmt tests without any progress. Giving up.

Ask on the channel. Ask sooner. There is no point wasting hours just to hit a wall and get nowhere. If it’s unclear and complicated, we can document and/or change it. Struggling alone and eventually giving up helps noone, it just leaves people pissed off, angry and disappointed. And not just you.

happz:
> > /home/runner/work/tmt/tmt/tmt/utils/filesystem.py:docstring of tmt.utils.filesystem.copy_tree:5:Unexpected indentation.
> 
> of course

It's the missing empty line between `Attempts strategies in order:` and `#. ``cp -a --reflink=auto`` (copy-on-write, with ``cp``'s own`.

(I'm sorry, but it was in my proposed patch... A linter would indeed be nice, but white space does matter.)

happz:
> > > /home/runner/work/tmt/tmt/tmt/utils/filesystem.py:docstring of tmt.utils.filesystem.copy_tree:5:Unexpected indentation.
> > 
> > 
> > of course
> 
> It's the missing empty line between `Attempts strategies in order:` and `` #. ``cp -a --reflink=auto`` (copy-on-write, with ``cp``'s own ``.
> 
> (I'm sorry, but it was in my proposed patch... A linter would indeed be nice, but white space does matter.)

https://github.com/teemtee/tmt/pull/3760 - and it would detect the issue in question.

therazix (approved):
LGTM

PR #3583, opened 2025-03-06
Support `dist-git-source` for repos without GitLab remote

CtrlZmaster:
I do not set up GitLab repo as a remote for RHEL dist-git because there is no `rhel-X-main` branch and all of the branches are protected anyways, so I don't see the point. I didn't run tmt locally for some time and yesterday, I was hit with a `No known remote ...` error when I attempted to run my test plan. Because I didn't know what it means, I searched in the source code and found the issue. Before 9cad874ff5fa3a33812073e8c5dffabcb8dabb13, RHEL dist-git remotes were specified by both `pkgs.devel.redhat.com` and GitLab URI substrings. Therefore, I believe that this is a regression. I wanted to create an issue first but then I realized that there is a simple fix, so I took a stab at it.

Maybe the omission of `pkgs.devel.redhat.com` was intentional in the rewrite, so feel free to reject this PR. There is a simple workaround of just adding the GitLab dist-git repo as a remote after all.

Pull Request Checklist

* [x] implement the feature
* [x] ~~write the documentation~~
* [x] ~~update the specification~~
* [x] ~~adjust plugin docstring~~
* [x] ~~modify the json schema~~
* [x] ~~mention the version~~
* [x] ~~include a release note~~


lukaszachy:
@CtrlZmaster Internal dist-git should be handled by `RHELInternalDistGit` from `tmt-redhat-utils` package.
With that said, as dist-git url is not a secret for some time already, this PR will make it easier for users so +1 to have it public.

CtrlZmaster:
@lukaszachy well, that's embarrassing :facepalm: I knew of the plugins but I didn't see any use for them in my workflow. Thinking about it, this might be the first time I ran tmt from RHEL dist-git. I started developing tmt plans in CentOS and never touched them in RHEL z-stream, I guess. 

Since the dist-git URL is not a secret, it would be nicer if this worked out-of-the-box. The additional class in the plugin is pretty much just duplicated code and if a change like this was merged, the plugin could be eliminated. I think that that could make life of packagers easier, because the plugin wouldn't be needed. Just installing tmt from RPM would be enough to start, no need to setup repos for the plugins.

I can rewrite the commit messages so that they are correct and rebase... I suppose that the internal plugin was always needed and this is not a regression? Should I take the PR out of draft? Can you please guide me on what to do next?

lukaszachy:
I see we have `class *DistGit` defined both in `tmt.utils` and `tmt.utils.git`
 @happz The right place should be `tmt.utils.git` ?

@CtrlZmaster Lets wait until we figure out  where is the right single place to `RedHatDistGit` defined. After that I think this PR needs to drop the other location and it is good to go.

happz:
> I see we have `class *DistGit` defined both in `tmt.utils` and `tmt.utils.git` @happz The right place should be `tmt.utils.git` ?

Most likely, that looks like those classes were moved to `tmt.utils.git` but I forgot to remove the originals. Something might be still referring to them. In any case, `tmt.utils.git` it should be, IMO.

CtrlZmaster:
@happz I was wondering why are the classes defined twice, so I just decided to be cautious and fix both places. The only references I could find yesterday were from inside the `tmt.utils.git` module. I'll double check and remove the classes from `tmt.utils`.

lukaszachy (approved):
Local testing shows it works :)

lukaszachy:
@CtrlZmaster We have quite horrible PR->review->merge turnaround so please bear with us

happz:
@CtrlZmaster please, rebase the PR when you find some spare time. It should be on its way to `main` just after that.

CtrlZmaster:
@happz Second time is the charm, I didn't rebase correctly on the first try  :sweat_smile: should be ok now

happz:
> @happz Second time is the charm, I didn't rebase correctly on the first try 😅 should be ok now

I believe you did, but there are still some failing tests: https://artifacts.dev.testing-farm.io/bb659df4-9d20-44ff-8e14-60a642cfebb6/work-corenggh_x6e/plans/features/core/execute/data/guest/default-0/tests/unit/with-system-packages/basic-41/output.txt I'd say it's related to the removal of code, fixing imports in tests should be enough to resolve this issue.

thrix:
@CtrlZmaster PTAL pls, we will be very happy to merge this in once this last thing resolved.

CtrlZmaster:
Omg, I put `git` in the wrong place. So much for doing a quick fix :roll_eyes:

CtrlZmaster:
Actually, I rebased that whole block wrong again even on the second try. 

It just wasn't apparent because of how that test is written. In pcs, we also used to have mega tests that had multiple asserts in them. We (mostly) phased them out in favor of small tests that each test just one thing. The problem with these bigger tests is that once it starts failing somewhere in the beginning, it can mask other failures, leading to fixing the same test multiple times, just moving down through the multiple failing asserts. I wanted to take a stab at also separating the tests for different handlers but this is `pytest` and I'm not familiar with how it works, we use `unittest` in pcs and it seems to be really different.

happz:
And the next hurdle is, downstream plugins import distgit classes, and they are moved now \o/ https://artifacts.osci.redhat.com/testing-farm/6736cdc7-d1bd-4be7-9772-d0989fe851a5/ I guess we need a patch for downstream plugins, to help them transition to new locations.

lukaszachy:
> And the next hurdle is, downstream plugins import distgit classes, and they are moved now \o/ https://artifacts.osci.redhat.com/testing-farm/6736cdc7-d1bd-4be7-9772-d0989fe851a5/ I guess we need a patch for downstream plugins, to help them transition to new locations.

Internal tmt/-/merge_requests/117 should take care of this

psss (approved):
Thanks for simplifying this!

PR #3586, opened 2025-03-07
Add the `bootc` method for installing packages

ukulekek:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note
  * When `tmt` works with the image mode images, it now uses native package installation method instead of using `rpm-ostree`: `tmt` creates a `Containerfile` based on the booted image, adds needed packages there, builds the image and reboot a system to use the fresh image with needed packages.

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -5,6 +5,16 @@
 ======================
 
 
+tmt-1.47.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When `tmt` works with the image mode images, it now uses native package
+installation method instead of using `rpm-ostree`: `tmt` creates a
+`Containerfile` based on the booted image, adds needed packages there,
+builds the image and reboots the system to use the fresh image with the
+necessary packages.
+
+
 tmt-1.46.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -45,6 +55,7 @@ hardware property has been proposed to allow specifying the desired
 system management interface (e.g., IPMI) when provisioning hardware.
 While not yet implemented, this feature aims to support more precise
 hardware selection in the future.
+
 
 tmt-1.45.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/provision/bootc/data/plans.fmf
+++ b/tests/provision/bootc/data/plans.fmf
@@ -5,6 +5,8 @@ provision:
     disk: 20
 execute:
     how: tmt
+environment:
+    PATTERN: localhost/tmt/bootc/
 
 
 /image:
@@ -14,16 +16,12 @@ execute:
         provision+:
             add-tmt-dependencies: true
             container-image: localhost/tmt-bootc-needs-deps
-        environment:
-            PATTERN: localhost/tmtmodified
 
     /includes-deps:
         summary: "Image that already includes dependencies"
         provision+:
             add-tmt-dependencies: false
             container-image: localhost/tmt-bootc-includes-deps
-        environment:
-            PATTERN: localhost/tmt-bootc-includes-deps
 
 /containerfile:
 
@@ -32,8 +30,6 @@ execute:
         provision+:
             add-tmt-dependencies: true
             container-file: needs-deps.containerfile
-        environment:
-            PATTERN: localhost/tmtmodified
 
     /includes-deps:
         summary: "Containerfile that already includes dependencies"
@@ -41,5 +37,3 @@ execute:
             how: bootc
             add-tmt-dependencies: false
             container-file: includes-deps.containerfile
-        environment:
-            PATTERN: localhost/tmtbase
--- a/tests/provision/bootc/data/test.fmf
+++ b/tests/provision/bootc/data/test.fmf
@@ -1,2 +1,4 @@
 summary: Check that booted image matches expected pattern
+require:
+  - tree
 test: "bootc status && bootc status | grep $PATTERN"
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -17,20 +17,45 @@ rlJournalStart
 
     rlPhaseStartTest "Image that needs dependencies"
         rlRun "podman build . -f needs-deps.containerfile -t $IMAGE_NEEDS_DEPS"
-        rlRun "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
+        rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
+        # Testing the output of the bootc package manager
+        rlAssertGrep "building container image with dependencies" $rlRun_LOG
+        rlAssertGrep "STEP 1/2: FROM containers-storage:localhost/tmt/bootc" $rlRun_LOG
+        rlAssertGrep "Successfully tagged localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "switching to new image localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "rebooting to apply new image" $rlRun_LOG
+
     rlPhaseEnd
 
     rlPhaseStartTest "Image that already includes dependencies"
         rlRun "podman build . -f includes-deps.containerfile -t $IMAGE_INCLUDES_DEPS"
-        rlRun "tmt -vvv run --scratch -i $run plan --name /plans/image/includes-deps"
+        rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/image/includes-deps"
+        # Testing the output of the bootc package manager
+        rlAssertGrep "building container image with dependencies" $rlRun_LOG
+        rlAssertGrep "STEP 1/2: FROM containers-storage:localhost/tmt-bootc-includes-deps" $rlRun_LOG
+        rlAssertGrep "Successfully tagged localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "switching to new image localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "rebooting to apply new image" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "Containerfile that needs dependencies"
-        rlRun "tmt -vvv run --scratch -i $run plan --name /plans/containerfile/needs-deps"
+        rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/containerfile/needs-deps"
+        # Testing the output of the bootc package manager
+        rlAssertGrep "building container image with dependencies" $rlRun_LOG
+        rlAssertGrep "STEP 1/2: FROM containers-storage:localhost/tmtmodified" $rlRun_LOG
+        rlAssertGrep "Successfully tagged localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "switching to new image localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "rebooting to apply new image" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "Containerfile that already includes dependencies"
-        rlRun "tmt -vvv run --scratch -i $run plan --name /plans/containerfile/includes-deps"
+        rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/containerfile/includes-deps"
+        # Testing the output of the bootc package manager
+        rlAssertGrep "building container image with dependencies" $rlRun_LOG
+        rlAssertGrep "STEP 1/2: FROM containers-storage:localhost/tmtbase-" $rlRun_LOG
+        rlAssertGrep "Successfully tagged localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "switching to new image localhost/tmt/bootc/" $rlRun_LOG
+        rlAssertGrep "rebooting to apply new image" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tmt/package_managers/bootc.py
+++ b/tmt/package_managers/bootc.py
@@ -1,0 +1,264 @@
+import json
+import re
+import uuid
+from typing import Any, Optional, cast
+
+import tmt.utils
+from tmt.package_managers import (
+    FileSystemPath,
+    Installable,
+    Options,
+    PackageManager,
+    PackageManagerEngine,
+    dnf,
+    provides_package_manager,
+)
+from tmt.utils import Command, CommandOutput, GeneralError, Path, RunError, ShellScript
+
+
+class BootcEngine(PackageManagerEngine):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize bootc engine for package management"""
+        super().__init__(*args, **kwargs)
+
+        self.aux_engine = dnf.DnfEngine(*args, **kwargs)
+
+        self.initialize_containerfile_directives()
+
+    def initialize_containerfile_directives(self) -> None:
+        """Initialize containerfile directives"""
+        self.containerfile_directives = self._get_base_containerfile_directives()
+
+    def prepare_command(self) -> tuple[Command, Command]:
+        """
+        Prepare installation command for bootc
+        """
+        command = Command()
+
+        if self.guest.facts.is_superuser is False:
+            command += Command('sudo')
+
+        command += Command('bootc')
+        return (command, Command(''))
+
+    def _get_current_bootc_image(self) -> str:
+        """Get the current bootc image running on the system"""
+
+        command, _ = self.prepare_command()
+        command += Command('status', '--json')
+        output = self.guest.execute(command, silent=True)
+
+        if not output.stdout:
+            raise tmt.utils.PrepareError("Failed to get current bootc status: empty output.")
+
+        try:
+            image_status = json.loads(output.stdout)
+        except json.JSONDecodeError as error:
+            raise tmt.utils.PrepareError(f"Failed to parse bootc status JSON: {error}")
+
+        if not image_status:
+            raise tmt.utils.PrepareError("Failed to get current bootc status: empty JSON.")
+
+        # Extract nested information with clear error messages for each missing key
+        try:
+            booted = image_status.get('status', {}).get('booted', {})
+            if not booted:
+                raise KeyError("'booted' key")
+
+            image_info = booted.get('image')
+            if not image_info:
+                raise KeyError("'image' key in booted status")
+
+            image_data = image_info.get('image', {})
+
+            base_image = cast(str, image_data.get('image', ''))
+            if not base_image:
+                raise KeyError("'image' name in image data")
+
+            return base_image
+        except KeyError as error:
+            raise tmt.utils.PrepareError(f"Failed to extract bootc image info: missing {error}")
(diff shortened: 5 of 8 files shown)


ukulekek:
Hi, @happz. I wanted to ask what the relationship is between `Install*` classes and corresponding package managers? Respectively, what logic should I move to the package manager out of the `InstallBootc` class?

ukulekek:
@happz, the idea with adapters sounds great to me as I was struggling to come up with a solution on how not to execute commands immediately. I'll try to move all logic to the package manager next week, thank you!

happz:
> @happz, the idea with adapters sounds great to me as I was struggling to come up with a solution on how not to execute commands immediately.

Sounds good, but it didn't work. I'm not smart enough to appease all type-related issues I was hitting. So, I took different approach: https://github.com/teemtee/tmt/pull/3587 - not tested, just locally...

Package manager we have are now split into two parts: "engine" which does what PM does today, prepares commands to install stuff, but it does not run anything - instead, it returns these commands. The rest of code then works with "package manager" part, that is just a "call the engine and run what it gives you" frontend for its engine. tmt still uses this interface, calls `guest.package_manager.install(...)` which turns out to be `self.guest.execute(self.engine.install(...))`.

And for your use case, I imagined your code could take the engine itself, e.g. `tmt.package_manager.dnf.DnfEngine`, and then use its methods, e.g. `install(...)`. It has the very same input as package manager's install, but it returns a shell script instead of running it. I imagine you would use the engine and its methods in the same way `Install*` does, except you would take their return values - scripts like `dnf -y install foo bar baz` - and inject them into Containerfiles you are building.

Something like this:

```python
    # Let's say we collect instructions for the final Containerfile in self._containerfile_directives, and that we have self.engine which is the DnfEngine

    def install_from_repository(self) -> None:
        script = self.engine.install(
            *self.list_installables("package", *self.packages),
            options=Options(
                excluded_packages=self.exclude,
                skip_missing=self.skip_missing,
            ),
        )

        self._containerfile_directives.append(f'RUN {script}')
```

Feel free to play with it, I didn't experiment with your patch yet.

Also, all those `\n`s throughout the code are very hard to keep up with. Check out what `install()` and `_install()` do, they call sequence of methods of the selected installer - if I would working on this, I would take advantage of this: have a list of strings to collect Containerfile commands, populate it with some `FROM ...` and `ENV ...` directives as necessary, and then methods like `install_local()` or `install_debuginfo()` would talk to the engine (can it always be dnf? or do we need to detect it somehow?), and slap `RUN ...` before all scripts returned by the engine. Eventually, you'd have a list of strings, it should be enough to do some `containerfile = '\n'.join(self._containerfile_directives)`, save it into a file and let `podman build` do its job.

Plus, you will not need an extra flag for tracking whether there are any commands to include - if `self._containerfile_directives` has just one item, it's your initial `FROM ...` and there is nothing else to do.

Let me know about your progress or any idea you get about the solution.

psss:
@ukulekek, is it realistic to finish by the end of the week so that there is some time for review before the release deadline?

tests/unit/test_package_managers.py
```diff
@@ -380,7 +380,7 @@ def _parametrize_test_install() -> Iterator[
                 container,
                 package_manager_class,
                 Package('tree'),
-                r"export DEBIAN_FRONTEND=noninteractive; dpkg-query --show tree \|\| apt install -y  tree",  # noqa: E501
+                r"set -x\s+export DEBIAN_FRONTEND=noninteractive\s+installable_packages=\"tree\"\s+dpkg-query --show \$installable_packages \\\s+\|\| apt install -y  \$installable_packages\s+exit \$\?",  # noqa: E501
```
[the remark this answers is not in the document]
happz: Ok, this applies to more files, including package managers. Proper rebase is needed before review.

tmt/steps/prepare/install.py
```diff
@@ -582,6 +583,90 @@ def install_debuginfo(self) -> None:
...
+        # Filelist for packages on the guest
+        filelist = [
+            PackagePath(self.package_directory / filename.name) for filename in self.local_packages
+        ]
+
+        cast(BootcEngine, self.engine).containerfile_directives.append(
+            f'RUN mkdir -p {self.package_directory}'
+        )
+        cast(BootcEngine, self.engine).containerfile_directives.append(
+            f'COPY {" ".join([str(file) for file in filelist])} {self.package_directory}'
```
happz, line 623: I believe you don't need to materialize a list in `join()`, a generator should work as well: `join(str(file) for file in filelist)`

tmt/steps/prepare/install.py
```diff
@@ -582,6 +583,90 @@ def install_debuginfo(self) -> None:
...
 
 
+class InstallBootc(InstallBase):
+    """Install packages using bootc container image mode"""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize bootc installation for package management"""
+        super().__init__(*args, **kwargs)
+
+        self.engine = self.guest.package_manager.engine
```
happz, line 593: `self.engine` seems to be rarely used, most of the calls below uses `self.guest.package_manager.engine`.

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,262 @@
...
+                raise tmt.utils.PrepareError("Failed to detect current bootc image.")
+
+            return base_image
+
+        except json.JSONDecodeError as error:
+            raise tmt.utils.PrepareError(f"Failed to parse bootc status: {error}")
+
+    def _get_base_containerfile_directives(self) -> list[str]:
+        base_image = self._get_current_bootc_image()
+        return [f'FROM {base_image}']
```
happz, line 82: How about `return [f'FROM {self._get_current_bootc_image()}']`?

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+class BootcEngine(PackageManagerEngine):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize bootc engine for package management"""
+        super().__init__(*args, **kwargs)
+
+        self.aux_engine = dnf.DnfEngine(*args, **kwargs)
+
+        self.containerfile_directives = self._get_base_containerfile_directives()
```
happz, line 27: This should be a call to `self.initialize_containerfile_directives()` - if you have a method that initializes an internal state, you should use it and nothing else.

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,262 @@
...
+        base_image = self.engine._get_current_bootc_image()
+
+        try:
+            # Check if image exists and store it from bootc of don't
+            self.guest.execute(
+                ShellScript(
+                    f'podman images {base_image} | grep {base_image} || bootc image copy-to-storage --target {base_image}'  # noqa: E501
+                )
+            )
+            self.guest.execute(ShellScript(f'mkdir -p {containerfile_dir!s}'))
```
happz, line 191: You will need different paths for runners and guests. Or put the remote path into the current workdir. The path generated by `tempfile.gettempdir()` is node-specific; it was unused and available on the runner, but it may very easily be used, by a sheer accident, by something completely unrelated on the guest. So either use path that is truly same on both side, like workdir, or you need different temporary directory on thee guest, e.g. `Guest.mkdtemp`.

tests/unit/test_package_managers.py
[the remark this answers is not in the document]
ukulekek: Yeah, automated rebase went badly, rebased properly now

tmt/steps/prepare/install.py
ukulekek: Right, thanks, fixed

tmt/steps/prepare/install.py
ukulekek: Yes, I removed it and make all calls use `self.guest.package_manager.engine`

tmt/package_managers/bootc.py
ukulekek: Ty, changed

tmt/package_managers/bootc.py
ukulekek: Right, I was waiting for the `Guest.mkdtemp` implementation. I used it to create a temporary directory

happz:
@ukulekek please, mark this PR as ready for reviews, no longer a draft, if it's ready for reviews.

tests/provision/bootc/test.sh
```diff
@@ -17,7 +17,8 @@ rlJournalStart
 
     rlPhaseStartTest "Image that needs dependencies"
         rlRun "podman build . -f needs-deps.containerfile -t $IMAGE_NEEDS_DEPS"
-        rlRun "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
+        rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
+        rlAssertGrep "building container image with dependencies" $rlRun_LOG
```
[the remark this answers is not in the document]
ukulekek: Thanks, I added more checks throughout the whole test

martinhoyer (approved):
Nice! lgtm.

Tested the rpm locally from clean tmt init and worked fine.
Added just a couple of nitpicks, can be ignored.

It looks like there is an extra space between `options` and package (two whitespaces before `vim`). Not sure if it's rpm-ostree or bootc though. 
```
        how: install
        summary: Install required packages
        name: requires
        where: default-0
            cmd: rpm -q --whatprovides vim
        package: 1 package requested
            vim
            cmd: rpm -q --whatprovides vim || rpm-ostree install --apply-live --idempotent --allow-inactive --assumeyes  vim
```

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,259 @@
...
+            if match is not None:
+                results[installable] = False
+                continue
+
+            results[installable] = True
+
+        return results
+
+    def build_container(self) -> None:
+        image_tag = f"localhost/tmt-deps-{uuid.uuid4()}"
```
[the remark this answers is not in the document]
happz: Testing images have prefix `localhost/tmt/container/test`, I still hope we would get also some `localhost/tmt/vm/test`, I'd recommend using a similar prefix for images tmt builds, e.g. `localhost/tmt/bootc/...`.

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,259 @@
...
+                # Reboot into the new image
+                self.info("package", "rebooting to apply new image", "green")
+                self.guest.reboot()
+
+            finally:
+                # Re-intitialize containerfile directives
+                self.engine.initialize_containerfile_directives()
+
+    def refresh_metadata(self) -> CommandOutput:
+        self.guest.warn("Metadata refresh is not supported with rpm-ostree.")
```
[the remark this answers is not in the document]
ukulekek: Thanks, I actually added an implementation for the function

tests/provision/bootc/test.sh
```diff
@@ -17,20 +17,42 @@ rlJournalStart
 
     rlPhaseStartTest "Image that needs dependencies"
         rlRun "podman build . -f needs-deps.containerfile -t $IMAGE_NEEDS_DEPS"
-        rlRun "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
+        rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
```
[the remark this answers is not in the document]
ukulekek: Thank you, I added comments

tests/provision/bootc/data/test.fmf
```diff
@@ -1,2 +1,4 @@
 summary: Check that booted image matches expected pattern
+require:
+  - vim
```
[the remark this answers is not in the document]
ukulekek: Sure, changed to this package

tmt/package_managers/bootc.py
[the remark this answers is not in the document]
ukulekek: Thanks, changed to `localhost/tmt/bootc/{uuid.uuid4()}`

psss:
@ukulekek, could you please update the pull request checklist? I believe this would definitely deserve a release note. Could you please add one? Thanks!

therazix (approved):

tests/provision/bootc/data/plans.fmf
```diff
@@ -15,15 +15,15 @@ execute:
...
+            PATTERN: localhost/tmt-deps
 
     /includes-deps:
         summary: "Image that already includes dependencies"
         provision+:
             add-tmt-dependencies: false
             container-image: localhost/tmt-bootc-includes-deps
         environment:
-            PATTERN: localhost/tmt-bootc-includes-deps
+            PATTERN: localhost/tmt-deps
```
psss, line 26: If the image includes all necessary dependencies, shouldn't it be used directly? This suggests that we create a new image even if it is not necessary. I guess there could be situations when user really wants to boot their prepared image without any changes performed by `tmt`? @thrix, any thoughts on this?

ukulekek:
@psss I added a release note, please check.

tests/provision/bootc/data/test.fmf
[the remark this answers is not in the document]
therazix: Is there a reason why this was changed back to `vim`?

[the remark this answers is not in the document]
ukulekek: Hi, the dummy-package-canary is available only in EPEL in CentOS composes, and I didn't want to enable epel in CI as it will slow things down more. Would you happen to have any alternatives in mind?

therazix:
> @psss I added a release note, please check.

The release note should be added to `/docs/releases.rst`

tests/provision/bootc/data/test.fmf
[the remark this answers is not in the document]
therazix: Ah, makes sense. We can keep `vim` then, unless @martinhoyer has some suggestions.

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,257 @@
...
+
+            containerfile = '\n'.join(self.engine.containerfile_directives)
+
+            base_image = self.engine._get_current_bootc_image()
+
+            try:
+                # Check if image exists and store it from bootc of don't
+                self.guest.execute(
+                    ShellScript(
+                        f'podman images {base_image} | grep {base_image} || bootc image copy-to-storage --target {base_image}'  # noqa: E501
```
thrix, line 194: Unfortunately `bootc image copy-to-storage` does not copy all the files :( Dropping at least yum repositories, so this breaks for RHEL image mode images from Testing Farm.

```
$ git clone https://pkgs.devel.redhat.com/cgit/rpms/bash 
$ cd bash
$ tmt run provision -h virtual -i http://images.osci.redhat.com/RHEL-9.6.0-image-mode-x86_64.qcow2
```

tests/provision/bootc/data/test.fmf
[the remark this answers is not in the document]
psss: We're often using `tree` as a test package as it's quite small.

tmt/package_managers/bootc.py
ukulekek: I tried to come up with a different solution, but I always ended up breaking the /plans/provision/bootc workflow, which is next:
* Build a local container image `localhost/container-foo` and run tmt, which boots into the container and tries to build a new image on top of it with bootc. But it always ends up in error ([link to test run](https://artifacts.osci.redhat.com/testing-farm/ab72efa6-3682-4047-959f-8aaccb909ba0/)):
```
>>> FROM localhost/container-foo
ERROR: failed to solve: localhost/container-foo: failed to resolve source metadata for localhost/container-foo:latest: failed to do request: Head "https://localhost/v2/container-foo/manifests/latest": dial tcp [::1]:443: connect: connection refused
```
This is because the booted container image doesn't have access to itself; thus, using `copy-to-storage` seems like the only option.

I tried adding this as a workaround:
```
                # Workaround of https://github.com/bootc-dev/bootc/issues/1259
                try:
                    if self.guest.execute(ShellScript("ls /etc/yum.repos.d")):
                        self.guest.execute(
                            ShellScript(
                                f'echo "COPY /etc/yum.repos.d /etc/yum.repos.d" >> {containerfile_path!s}'  # noqa: E501
                            )
                        )
                except RunError:
                    pass
```
But it doesn't work because the `/etc/yum.repos.d` is out of building context

thrix:
@ukulekek tests are still failing?

docs/releases.rst
```diff
@@ -46,6 +46,13 @@ system management interface (e.g., IPMI) when provisioning hardware.
 While not yet implemented, this feature aims to support more precise
 hardware selection in the future.
 
+When `tmt` works with the image mode images, it now uses native package
+installation method instead of using `rpm-ostree`: `tmt` creates a
+`Containerfile` based on the booted image, adds needed packages there,
+builds the image and reboot a system to use the fresh image with needed
+packages.
```
thrix, line 53:
```suggestion
builds the image and reboots the system to use the fresh image with the
necessary packages.
```

tmt/package_managers/bootc.py
```diff
@@ -0,0 +1,257 @@
...
+        # Write the final Containerfile
+        with self.guest.mkdtemp() as containerfile_dir:
+            containerfile_path = Path(containerfile_dir, 'Containerfile')
+
+            containerfile = '\n'.join(self.engine.containerfile_directives)
+
+            base_image = self.engine._get_current_bootc_image()
+
+            try:
+                # Check if image exists and store it from bootc of don't
```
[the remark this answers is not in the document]
ukulekek: Sure, added

tests/provision/bootc/data/plans.fmf
```diff
@@ -15,15 +15,15 @@ execute:
             add-tmt-dependencies: true
             container-image: localhost/tmt-bootc-needs-deps
         environment:
-            PATTERN: localhost/tmtmodified
+            PATTERN: localhost/tmt/bootc/
```
janhavlin, line 18: Since the variable value is now shared across all plans, it can be defined just once on the root level:
```
discover:
    how: fmf
provision:
    how: bootc
    disk: 20
execute:
    how: tmt
environment:
    PATTERN: localhost/tmt/bootc/
```

janhavlin (approved):

thrix:
I was able to run `systemd` tests:


```
   .... (snip) ....
   prepare
        
        queued prepare task #2: requires on default-0
        queued prepare task #3: recommends on default-0
        
        how: install
        name: essential-requires
        where: default-0
        package: audit, policycoreutils, beakerlib and 1 more
        package: building container image with dependencies
        package: switching to new image localhost/tmt/bootc/6dde5b01-8c03-424f-897d-07f8a616ca76
        package: rebooting to apply new image
        
        prepare task #2: requires on default-0
        how: install
        summary: Install required packages
        name: requires
        where: default-0
        package: policycoreutils, rpm, netlabel_tools and 6 more
        package: building container image with dependencies
        package: switching to new image localhost/tmt/bootc/184033dd-ef2b-451d-9261-787d8c0ed555
        package: rebooting to apply new image
        
        prepare task #3: recommends on default-0
        how: install
        summary: Install recommended packages
        name: recommends
        where: default-0
        package: libvirt-client, glibc-langpack-sk, cups and 53 more
        package: building container image with dependencies
        package: switching to new image localhost/tmt/bootc/c6dd803b-9a27-4b5e-9d81-6f1b340dcc5c
        package: rebooting to apply new image
    
        summary: 3 preparations applied
    execute
        queued execute task #1: default-0 on default-0
        
        how: tmt
        progress: /Regression/leaking-scope-units [15/59]                                    
    .... (snip) ....
```

thrix (changes requested):

thrix (approved):
LGTM

thrix:
@happz pls do a last check on the code, otherwise it has review

thrix:
all tests passed, except provision which we saw already passing, but openstack today does not play well, will be doing a final run on #3693

PR #3593, opened 2025-03-11
Add a helper for creating temporary directories on guest

happz:
It will come handy when building image mode image.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -1,4 +1,5 @@
 import ast
+import contextlib
 import dataclasses
 import datetime
 import enum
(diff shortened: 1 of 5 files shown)


tmt/steps/provision/__init__.py
```diff
@@ -1788,6 +1789,84 @@ def fetch_logs(
...
+    ) -> Command:
+        template = template or 'tmp.XXXXXXXXXX'
+
+        if prefix is not None:
+            template = f'{prefix}{template}'
+
+        options: list[str] = ['--directory']
+
+        if parent is not None:
+            options += ['-p', str(parent)]
```
martinhoyer, line 1806:
```suggestion
        if parent is not None:
            # Using --tmpdir instead of deprecated -p
            options += ['--tmpdir', str(parent)]
```
The manpage says `-p` is deprecated:
```
       -p DIR, --tmpdir[=DIR]
              interpret TEMPLATE relative to DIR; if DIR is not specified, use $TMPDIR if set, else  /tmp.   With  this
              option,  TEMPLATE  must not be an absolute name; unlike with -t, TEMPLATE may contain slashes, but mktemp
              creates only the final component

       -t     interpret TEMPLATE as a single file name component, relative to a directory: $TMPDIR, if  set;  else  the
              directory specified via -p; else /tmp [deprecated]
```

tmt/steps/provision/__init__.py
```diff
@@ -1788,6 +1789,84 @@ def fetch_logs(
...
+
+        if not output.stdout:
+            raise GeneralError(f"Failed to create temporary directory on guest: {output.stderr}")
+
+        path = Path(output.stdout.strip())
+
+        try:
+            yield path
+
+        except Exception as exc:
```
martinhoyer, line 1864: What do you think about keeping the directory on failure? Not sure if it should be optional, and if yes, opt-in or opt-out.
```suggestion
        except Exception as exc:
            if not keep_on_error:
                self.execute(Command('rm', '-rf', path))
            else:
                self.debug(f"Keeping temporary directory {path} due to error")
```

martinhoyer: my IDE gave me this unit test to cover it fwiw:
```
diff --git a/tests/unit/test_guest.py b/tests/unit/test_guest.py
index 5058982b..0c981901 100644
--- a/tests/unit/test_guest.py
+++ b/tests/unit/test_guest.py
@@ -100,6 +100,7 @@ def test_mkdtemp(
 ) -> None:
     guest.execute(ShellScript('mkdir -p /tmp/qux'))
 
+    # Test basic functionality
     with guest.mkdtemp(
         prefix='bar',
         template='XXXXXXbazXXXXXX',
@@ -107,7 +108,43 @@ def test_mkdtemp(
     ) as path:
         guest.execute(ShellScript(f'ls -al {path}'))
 
+    # Verify directory was removed
     with pytest.raises(RunError) as exc_context:
         guest.execute(ShellScript(f'ls -al {path}'))
 
     assert re.match(r'(?i).*?No such file or directory.*', exc_context.value.stderr or '')
+
+
+@pytest.mark.containers
+@pytest.mark.parametrize(
+    ('container',),  # noqa: PT006
+    [(TEST_CONTAINERS['CONTAINER_FEDORA_41'],)],  # Test only on one container to save time
+    indirect=["container"],
+    ids=[TEST_CONTAINERS['CONTAINER_FEDORA_41'].url],
+)
+def test_mkdtemp_keep_on_error(
+    container: ContainerData,
+    guest: GuestContainer,
+    root_logger: Logger,
+    caplog: _pytest.logging.LogCaptureFixture,
+) -> None:
+    error_message = "Simulated error for testing"
+    
+    try:
+        with guest.mkdtemp(
+            prefix='keep-me-',
+            keep_on_error=True,
+        ) as path:
+            # Record the path for later verification
+            temp_path = path
+            # Create a marker file to verify the directory exists
+            guest.execute(ShellScript(f'touch {path}/marker'))
+            # Simulate an error
+            raise RuntimeError(error_message)
+    except RuntimeError as exc:
+        assert str(exc) == error_message
+        # Verify the directory still exists with our marker
+        output = guest.execute(ShellScript(f'ls -la {temp_path}/marker'))
+        assert output.stdout.strip() != ""
+        # Clean up manually for the test
+        guest.execute(Command('rm', '-rf', temp_path))
```

martinhoyer (approved):
Nice. Adding just a couple of ignorable thoughts.

tmt/steps/provision/__init__.py
happz: That is true, and any modern `mktemp` will follow. But Alpine `mktemp` does not accept `--tmpdir`:

```
/ # mktemp --help
BusyBox v1.36.1 (2024-06-12 06:28:12 UTC) multi-call binary.

Usage: mktemp [-dt] [-p DIR] [TEMPLATE]

Create a temporary file with name based on TEMPLATE and print its name.
TEMPLATE must end with XXXXXX (e.g. [/dir/]nameXXXXXX).
Without TEMPLATE, -t tmp.XXXXXX is assumed.

        -d      Make directory, not file
        -q      Fail silently on errors
        -t      Prepend base directory name to TEMPLATE
        -p DIR  Use DIR as a base directory (implies -t)
        -u      Do not create anything; print a name

Base directory is: -p DIR, else $TMPDIR, else /tmp
```

I think even I started with `--tmpdir` or something similar, and Alpine tests did not work.

therazix (approved):

tmt/steps/provision/__init__.py
happz: @martinhoyer ping

PR #3596, opened 2025-03-12
Use the `display` report plugin to display results during execution

happz:
`execute/tmt` is using very, very similar output format - test, duration, check results, subresults. The patch changes `display` to be reusable from other places, and changes `execute/tmt` to use `display` instad of its custom code. With a bit of templating, what gets rendered how should now be easier to follow, and we are reusing existing code.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -4,6 +4,16 @@
     Releases
 ======================
 
+tmt-1.49.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Output of the :ref:`/plugins/report/internal` and
+:ref:`/plugins/report/display` is changing in this release, to provide
+slightly more details, headers and timestamps. ``execute`` now starts
+using ``display`` for its own progress reporting, providing the unified
+formatting and simplified code.
+
+
 tmt-1.48.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -7,7 +7,7 @@
 tmt-1.49.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Output of the :ref:`/plugins/report/internal` and
+Output of the :ref:`/plugins/report/display` and
 :ref:`/plugins/report/display` is changing in this release, to provide
 slightly more details, headers and timestamps. ``execute`` now starts
 using ``display`` for its own progress reporting, providing the unified
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -7,7 +7,7 @@
 tmt-1.49.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Output of the :ref:`/plugins/report/display` and
+Output of the :ref:`/plugins/execute/tmt` and
 :ref:`/plugins/report/display` is changing in this release, to provide
 slightly more details, headers and timestamps. ``execute`` now starts
 using ``display`` for its own progress reporting, providing the unified
--- a/tests/execute/result/basic.sh
+++ b/tests/execute/result/basic.sh
@@ -2,6 +2,23 @@
 # vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
 . /usr/share/beakerlib/beakerlib.sh || exit 1
 
+function test_result () {
+    pattern="$1"
+    extra_lines="$2"
+
+    shift 2
+
+    rlRun "grep -P -A$extra_lines \"\s+\d\d:\d\d:\d\d\s+$pattern\" $run/report.txt > $run/test.txt"
+    rlRun "cat $run/test.txt"
+
+    for i in `seq 1 $extra_lines`; do
+        pattern="$1"
+        shift
+
+        rlRun "grep -P \"\s+$pattern\" $run/test.txt"
+    done
+}
+
 run()
 {
     res=$1  # expected final result of test
@@ -9,13 +26,16 @@ run()
     orig=$3 # original result
     ret=$4  # tmt return code
 
-    rlRun -s "tmt run -a --scratch --id \${run} test --name ${tn} provision --how local report -v 2>&1 >/dev/null | grep report -A2 | tail -n 1" \
+    if [ -z "${orig}" ]; then extra_lines=2; else extra_lines=3; fi
+
+    rlRun -s "tmt run -a --scratch --id \${run} test --name ${tn} provision --how local report -v 2>&1 >/dev/null | grep report -A$extra_lines" \
         ${ret} "Result: ${res}, Test name: ${tn}, Original result: '${orig}', tmt return code: ${ret}"
 
     if [ -z "${orig}" ]; then # No original result provided
         rlAssertGrep "${res} ${tn}$" $rlRun_LOG
     else
-        rlAssertGrep "${res} ${tn} (.*original test result: ${orig}.*)$" $rlRun_LOG
+        rlAssertGrep "${res} ${tn}$" $rlRun_LOG
+        rlAssertGrep "Note: ${orig}" $rlRun_LOG
     fi
 
     echo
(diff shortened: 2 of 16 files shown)


tmt/utils/templates.py
```diff
@@ -310,6 +316,74 @@ def _template_filter_web_git_url(  # type: ignore[reportUnusedFunction,unused-ig
...
+        # {"duration": None} -> '    '
+        {{ {"duration": None} | format_duration }}
+
+        # {"duration": "12:34:56"} -> '12:34:56'
+        {{ {"duration": "12:34:56"} | format_duration }}
+    """
+
+    # If test duration information is missing, print 8 spaces to keep indentation
+
+    return result.duration if result.duration else 6 * ' '
```
therazix, line 384: Shouldn't there be `8` spaces instead of `6` to keep the indentation same? The comment above also says that there are 8 spaces.
```suggestion
    return result.duration if result.duration else 8 * ' '
```

Or perhaps we can print some placeholder value like `--:--:--`.

tmt/steps/report/display.py
```diff
@@ -33,136 +67,254 @@ class ReportDisplayData(tmt.steps.report.ReportStepData):
...
+        outcome = 'errr' if result.result == ResultOutcome.ERROR else result.result.value
+
+        yield render_template(
+            self.subresult_header_template,
+            environment=self.environment,
+            CONTEXT=self,
+            RESULT=result,
+            OUTCOME=outcome,
+            OUTCOME_COLOR=RESULT_OUTCOME_COLORS[result.result],
+            INDENT=INDENT * ' ',
```
therazix, line 225: Subresult header template doesn't have `INDENT` variable anywhere. Is there a reason for passing it?

tests/test/check/data/main.fmf
```diff
@@ -1,8 +1,11 @@
 /dmesg:
-    test: /bin/true
+    test: echo "some text"
```
therazix, line 2: Is this test edited on purpose? I don't see how it's relevant to this PR.

therazix (approved):
LGTM, this approach makes sense. Just a few comments.

martinhoyer (approved):
Very nice! Nothing to add to the existing comments.

thrix (approved):
I tried to run bash tests and seems to look nice and correct as before. I see no changes in progress output. LGTM

thrix:
@happz this has some unresolved comments + failing tests, but already enough karma. Can you take a look pls?

happz:
Working on addressing comments and fixing tests, but I discovered some indentation issues and related ugliness in code I need to fix as well.

tmt/utils/templates.py
happz: Addressed in 02cc62d3 and bdec874b

tmt/steps/report/display.py
happz: Dropped in 02cc62d3

tests/test/check/data/main.fmf
happz: Reverted in 02cc62d3

tests/execute/result/basic.sh
```diff
@@ -2,20 +2,40 @@
...
+
+    rlRun "grep -P -A$extra_lines \"\s+\d\d:\d\d:\d\d\s+$pattern\" $run/report.txt > $run/test.txt"
+    rlRun "cat $run/test.txt"
+
+    for i in `seq 1 $extra_lines`; do
+        pattern="$1"
+        shift
+
+        rlRun "grep -P \"\s+$pattern\" $run/test.txt"
+    done
```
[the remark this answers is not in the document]
happz: Hmmm, I don't see why the order would cause `$1` to be always the same value. There is a `shift 2` before the loop, that "unrolls" the arguments already consumed, then `pattern="$1"` + `shift` pick the next argument in line. And it works in the test:

```
:: [ 10:24:28 ] :: [   PASS   ] :: Command 'grep -P -A2 "\s+\d\d:\d\d:\d\d\s+errr /test/always-error \(on default-0\) \[1/12\]" /tmp/tmp.HyRrmRItMS/report.txt > /tmp/tmp.HyRrmRItMS/test.txt' (Expected 0, got 0)
:: [ 10:24:28 ] :: [  BEGIN   ] :: Running 'cat /tmp/tmp.HyRrmRItMS/test.txt'
            00:00:00 errr /test/always-error (on default-0) [1/12]
                Note: test result overridden: error
                Note: original test result: pass
:: [ 10:24:28 ] :: [   PASS   ] :: Command 'cat /tmp/tmp.HyRrmRItMS/test.txt' (Expected 0, got 0)
:: [ 10:24:28 ] :: [  BEGIN   ] :: Running 'grep -P "\s+Note: test result overridden: error" /tmp/tmp.HyRrmRItMS/test.txt'
                Note: test result overridden: error
:: [ 10:24:28 ] :: [   PASS   ] :: Command 'grep -P "\s+Note: test result overridden: error" /tmp/tmp.HyRrmRItMS/test.txt' (Expected 0, got 0)
:: [ 10:24:28 ] :: [  BEGIN   ] :: Running 'grep -P "\s+Note: original test result: pass" /tmp/tmp.HyRrmRItMS/test.txt'
                Note: original test result: pass
:: [ 10:24:28 ] :: [   PASS   ] :: Command 'grep -P "\s+Note: original test result: pass" /tmp/tmp.HyRrmRItMS/test.txt' (Expected 0, got 0)
```

Both `Note: ...` patterns are tested.

tmt/steps/report/display.py
martinhoyer: @therazix probably can be resolved? :)

tests/execute/result/basic.sh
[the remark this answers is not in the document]
martinhoyer: Our fancy new LLM analysis:

>     Understanding shift and $1:
> 
>         $1, $2, etc., are positional parameters, representing the arguments passed to the script or function.
> 
>         shift discards $1 and shifts all other parameters down by one (e.g., $2 becomes $1, $3 becomes $2, and so on).
> 
>         pattern="$1" assigns the current value of the first positional parameter to the variable pattern.
> 
>     Tracing Version A (pattern="$1"; shift):
> 
>         Iteration 1:
> 
>             pattern="$1": pattern gets pattern_A.
> 
>             shift: $1 becomes pattern_B, $2 becomes pattern_C.
> 
>             grep uses pattern_A.
> 
>         Iteration 2:
> 
>             pattern="$1": pattern gets pattern_B (which is the current $1).
> 
>             shift: $1 becomes pattern_C.
> 
>             grep uses pattern_B.
>             This version correctly processes pattern_A then pattern_B. It's the standard and idiomatic way to iterate through positional parameters.
> 
>     Tracing Version B (shift; pattern="$1"):
> 
>         Iteration 1:
> 
>             shift: $1 becomes pattern_B, $2 becomes pattern_C. (The original pattern_A is discarded before being used by grep in this loop).
> 
>             pattern="$1": pattern gets pattern_B.
> 
>             grep uses pattern_B.
> 
>         Iteration 2:
> 
>             shift: $1 becomes pattern_C.
> 
>             pattern="$1": pattern gets pattern_C.
> 
>             grep uses pattern_C.
>             This version skips pattern_A (the initial $1 when the loop started) and processes pattern_B then pattern_C.
> 
> Critique of the Initial Argument:
> 
> Comment stated: "There's a logic error in the loop: the shift operation is performed after assigning pattern="$1", which means only the first pattern is used for all iterations."
> 
> This statement is incorrect. When shift is performed after pattern="$1":
> 
>     In the current iteration, the pattern variable correctly holds the value that $1 had at the beginning of that iteration.
> 
>     The shift then prepares $1 for the next iteration by loading the subsequent argument into it.
>     This ensures that each argument is processed in turn. If shift were not present, then indeed, only the first pattern would be used repeatedly.
> 

Let's give graphite a thumbs down and resolve imho.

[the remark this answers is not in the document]
happz: Very well, resolving.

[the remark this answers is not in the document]
happz: This is expected behavior, the very first argument is not a pattern to test in this loop.

tmt/steps/report/display.py
```diff
@@ -33,143 +60,261 @@ class ReportDisplayData(tmt.steps.report.ReportStepData):
...
 
-    _data_class = ReportDisplayData
+    #: When 2 or more, log info - name and path - would be printed out.
+    #: When 3 or more, log output would be printed out as well.
+    verbosity: int = 0
+    #: Whether guest from which results originated should be printed out.
+    display_guest: bool = True
+
+    #: Additional variables to use when rendering templates.
+    variables: dict[str, Any] = simple_field(default_factory=dict[str, Any])
```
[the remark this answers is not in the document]
happz: This is now requested by pyright, and it works.

thrix:
checking with packit why the results did not appear for 2 days

thrix:
> checking with packit why the results did not appear for 2 days

And the reporting made sense, seems like a bug (possible race condition): 
https://github.com/packit/packit-service/issues/2799

docs/releases.rst
```diff
@@ -4,6 +4,16 @@
     Releases
 ======================
 
+tmt-1.49.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Output of the :ref:`/plugins/report/display` and
+:ref:`/plugins/report/display` is changing in this release, to provide
```
psss, line 11: The same plugin is reported twice, seems like a copy & paste typo?

happz: Yes, it's a typo. Fixed in 95de5a01.

happz:
Neither of two failures is related to this PR, merging.

Issue #2271, opened 2023-08-16
Run shellcheck in pre-commit


martinhoyer:
Wouldn't it make sense to add [shellcheck-precommit hook](https://github.com/koalaman/shellcheck-precommit) to pre-commit config instead of having a dedicated github action and all that goes with it?

From what I understand, the point of [differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck) is that it checks only changed files, which is what pre-commit is doing.

We could also remove the .shellcheckrc file, which only includes `external-sources=true` and add it as an argument in pre-commit-config.yaml. (assuming people are not running shellcheck manually and instead rely on pre-commit locally)

PR #3604, opened 2025-03-15
Switch to hadolint-py and add shellcheck-py

martinhoyer:
> See: https://pre-commit.ci/
> 
> In order for it to work, it requires adding pre-commit.ci in gh organization settings. (not doing that without consensus)  
> 
> Can go hand-in-hand with https://github.com/teemtee/tmt/pull/2923
> 
> One motivation is to keep the checks autoupdated, for example https://github.com/teemtee/tmt/pull/3544#issuecomment-2713452237
> 
> GitHub action:
> >this action is in maintenance-only mode and will not be accepting new features.
> generally you want to use [pre-commit.ci](https://pre-commit.ci/) which is faster and has more features.
> 
> > pre-commit.ci will always be free for open source repositories.
> Also adding shellcheck.py while keeping the shellcheck action - Closes https://github.com/teemtee/tmt/issues/2271

pre-commit.ci doesn't play ball with pyright. For now, reverting to github workflow. Keeping hadolint-py and shellcheck-py

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -188,19 +188,13 @@ repos:
           - single-statement-per-line
           - "--"
 
-  - repo: https://github.com/hadolint/hadolint
-    rev: v2.13.1-beta
+  - repo: https://github.com/AleksaC/hadolint-py
+    rev: v2.12.1b3
     hooks:
-      # * hadolint-docker depends hard on Docker, ignores podman
-      # * pre-commit check does not install hadolint binary,
-      #   see https://github.com/hadolint/hadolint/issues/886
-      #
-      # As of now, installing hadolint "manually" is the only viable
-      # option. It is required to exist on developer's workstation,
-      # and it is installed by a dedicated step in Github pre-commit
-      # action (.github/workflows/pre-commit.yml).
+
       - id: hadolint
         files: ^containers/.*
+
   - repo: https://github.com/shellcheck-py/shellcheck-py
     rev: v0.10.0.1
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -201,3 +201,15 @@ repos:
       # action (.github/workflows/pre-commit.yml).
       - id: hadolint
         files: ^containers/.*
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.10.0.1
+    hooks:
+      - id: shellcheck
+        # Local shellcheck for developers - keep exclusions in sync with .github/workflows/shellcheck.yml
+        exclude: |
+          (?x)^(
+            tests/.*|
+            examples/.*|
+            tmt/steps/execute/scripts/.*\.sh\.j2|
+            tmt/templates/.*
+          )$
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -26,9 +26,10 @@ jobs:
         name: Differential ShellCheck
         uses: redhat-plumbers-in-action/differential-shellcheck@v4
         with:
+          # Keep exclusions in sync with shellcheck-py hook in .pre-commit-config.yaml
           exclude-path: |
             tests/**
-            examples/**/test.sh
+            examples/**
             tmt/steps/execute/scripts/*.sh.j2
             tmt/templates/**
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,17 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+ci:
+  autofix_commit_msg: |
+    [pre-commit.ci] auto fixes from pre-commit hooks
+
+    For more information, see https://pre-commit.ci
+  autofix_prs: true
```
LecrisUT, line 8: This one should be discussed if we want to autofix. It's mostly a style choice.

LecrisUT (commented):
I have not had much experience with pre-commit-ci, but iirc it rebases open PRs, so I am ok with either this or renovate as a tool.

.pre-commit-config.yaml
```diff
@@ -191,13 +203,17 @@ repos:
...
     rev: v2.13.1-beta
     hooks:
-      # * hadolint-docker depends hard on Docker, ignores podman
-      # * pre-commit check does not install hadolint binary,
-      #   see https://github.com/hadolint/hadolint/issues/886
-      #
-      # As of now, installing hadolint "manually" is the only viable
-      # option. It is required to exist on developer's workstation,
-      # and it is installed by a dedicated step in Github pre-commit
-      # action (.github/workflows/pre-commit.yml).
```
happz, line 201: Is GH action going to take care of pulling in `hadolint` automagically?

.pre-commit-config.yaml
thrix: I am against an autofix, contributors should install and use pre-commit locally, and can include the fixes themselves. I would not let pre-commit ci to step in

.pre-commit-config.yaml
martinhoyer: I wanted to deal with hadolint one way or another. In the meantime found  hadolint-py. It works for me locally, similarly to shellcheck-py

.pre-commit-config.yaml
martinhoyer: explicitly set to false

.pre-commit-config.yaml
happz: Ok, I missed the switch to hadolint-py.

happz (approved):

.pre-commit-config.yaml
martinhoyer: nono, I switched to it today, your feedback was very much valid :)

mcasquer (approved):
Since renovate addition should handle the rebases part, LGTM

martinhoyer:
Added  pre-commit ci app to teemtee/tmt

martinhoyer:
> exceeds tier max size 250MiB: 253.5MiB

oof nodejs binaries 190MB. Stupid javascript.

martinhoyer:
Nothing is ever easy.
For context - original failures: https://results.pre-commit.ci/run/github/206777717/1745941884.ILAZsf-0RJOXnlmnttSSTQ

And now: https://results.pre-commit.ci/run/github/206777717/1745947600.41gGYLi9Qe2q90n6v-OCrw
https://github.com/RobertCraigie/pyright-python/issues/345

Not sure if the tmt-lint works with the `- GIT_DEPTH: "1"` or not now.

martinhoyer:
ok, reverting to github workflow. Keeping hadolint and shellcheck changes.

mcasquer (approved):

.pre-commit-config.yaml
```diff
@@ -188,8 +188,8 @@ repos:
           - single-statement-per-line
           - "--"
 
-  - repo: https://github.com/hadolint/hadolint
-    rev: v2.13.1-beta
+  - repo: https://github.com/AleksaC/hadolint-py
```
happz, line 191: Actually, with the switch to `hadolint-py`: do we need the comment below, and `hadolint` among packages to install in https://github.com/teemtee/tmt/blob/main/Makefile#L262?

martinhoyer: good point!

martinhoyer: https://github.com/teemtee/tmt/pull/3604/commits/b9cacc7f89b23d961096cc9f0bf56e6f54a72785
Pretty cool. @thrix 
Screenshot From 2025-05-03 21-17-22 I just clicked on the first one pretty much.

happz:
Seems harmless & pre-commit only, merging without more checks.

PR #3605, opened 2025-03-17
Use containers to run tmt on older / unsupported systems?

martinhoyer:
This is mostly and experiment/proof-of-concept for an alternative to "python virtualenv + pypi wheels", that could be useful for older rpm-based systems, as tmt has non-python dependencies that can't be installed through pip.

That said, the bf732795ba1184f0504211ab1a96eefadd990397 is worth looking at as it's tangibly smaller - using fedora:minimal and microdnf.

Other thing I'd like to bring up is using copr repo for "stable", this is not ideal especially because of being "x86_64" only, but also, why not have the images in sync with what's in Fedora repos?

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


containers/proxy/tmt-proxy.sh
```diff
@@ -0,0 +1,76 @@
...
+    local run_command=("podman" "run" "--rm" "-it")
+
+    # Add user namespace configuration to keep host UID/GID
+    run_command+=("--userns=keep-id")
+
+    # Add default volume mounts
+    run_command+=("-v" "$workdir:$workdir")
+    run_command+=("-v" "$HOME/.ssh:/root/.ssh:ro")
+    run_command+=("-v" "/etc/tmt:/etc/tmt")
+    run_command+=("-v" "$HOME:$HOME")
```
martinhoyer, line 53: This is not amazing - left it there while in draft as I had issues with just $workdir iirc.

thrix:
Some dicussion from the call:

* maintaining older releases is a pain
* could be significant amount of work
* it is unclear if libvirt and other system interactions would be possible to support (bootc, etc.)

Issue #3573, opened 2025-03-04
tmt-abort is not honored across the boundary between discover phases


bgoncalv:
example: https://artifacts.osci.redhat.com/testing-farm/59e4191a-4e2e-41df-a82a-e003b38e8562/

tests from the first group were aborted, but then the next discover phase and its tests run.

kernelinstall test aborted, the next discovered tests continues to run.

PR #3606, opened 2025-03-18
Tests that abort or fail (with `exit-first`) will skip remaining tests

happz:
Before the patch, an aborted test or a failed test with `exit-first` would skip tests remaining in the current `discover` phase. Tests from remaining `discover` phases would still be executed.

The patch adds a signal for `execute` plugins to notify the step queue to not continue by running more phases. It's implemented by raising an exception, but that may change in the future, once we teach plugins and steps to return values instead of using object attributes to communicate results.

Fixes https://github.com/teemtee/tmt/issues/3573

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tests/execute/restraint/tmt-abort/data/plan.fmf
+++ b/tests/execute/restraint/tmt-abort/data/plan.fmf
@@ -1,9 +1,13 @@
 discover:
-    how: fmf
+  - how: fmf
     test:
      - abort
-     - do-not-run
+     - do-not-run/1
+  - how: fmf
+    test:
+     - do-not-run/2
 provision:
     how: container
+    image: localhost/tmt/container/test/fedora/rawhide:latest
 execute:
     how: tmt
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -285,6 +285,14 @@ class ExecuteStepData(tmt.steps.WhereableStepData, tmt.steps.StepData):
 ExecuteStepDataT = TypeVar('ExecuteStepDataT', bound=ExecuteStepData)
 
 
+class AbortExecute(tmt.utils.GeneralError):
+    """
+    Raised by ``execute`` phases to when the whole step should abort.
+    """
+
+    pass
+
+
 @container
 class TestInvocation:
     """
@@ -1059,15 +1067,6 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
         # (e.g. saves them as a tmt subresults).
         return invocation.test.test_framework.extract_results(invocation, results, logger)
 
-    def check_abort_file(self, invocation: TestInvocation) -> bool:
-        """
-        Check for an abort file created by tmt-abort
-
-        Returns whether an abort file is present (i.e. abort occurred).
-        """
-
-        return (invocation.test_data_path / TMT_ABORT_SCRIPT.created_file).exists()
-
     def timeout_hint(self, invocation: TestInvocation) -> None:
         """
         Append a duration increase hint to the test output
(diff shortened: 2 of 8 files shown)


tests/execute/restraint/tmt-abort/data/plan.fmf
```diff
@@ -1,9 +1,13 @@
...
     test:
      - abort
-     - do-not-run
+     - do-not-run/1
+  - how: fmf
+    test:
+     - do-not-run/2
 provision:
     how: container
+    image: localhost/tmt/container/test/fedora/rawhide:latest
```
thrix, line 11: Why we are forcing a rawhide image here?

tmt/steps/execute/__init__.py
```diff
@@ -285,6 +285,14 @@ class ExecuteStepData(tmt.steps.WhereableStepData, tmt.steps.StepData):
 ExecuteStepDataT = TypeVar('ExecuteStepDataT', bound=ExecuteStepData)
 
 
+class AbortExecute(tmt.utils.GeneralError):
+    """
+    Raised by ``execute`` phases to when the whole step should abort.
```
thrix, line 290:
```suggestion
    Raised by ``execute`` phases when the entire step should abort.
```

tests/execute/restraint/tmt-abort/data/plan.fmf
happz: Unless we set up some kind of "default image" configuration, we should force images in our plans. As a test image, `localhost/tmt/container/test/fedora/rawhide:latest` should contain the most up-to-date repository metadata, and save us some time in testing. And it's predictable, it's clear which image is used. I could use Fedora 41 too, but that one will go away one day, and we have no `localhost/tmt/container/test/fedora/current` image alias yet.

therazix (approved):

KwisatzHaderach (approved):

tests/execute/restraint/tmt-abort/data/plan.fmf
martinhoyer: > Fedora 41 too, but that one will go away one day

Wouldn't `fedora:latest` be ideal?

martinhoyer (approved):
nice

tests/execute/restraint/tmt-abort/data/plan.fmf
happz: Sounds good, but we don't have such an image yet. And we should have one, because a lot of tests just blindly use any default container image of a provisioning plugin. It's worthy its own patch, I'd say, to get such a test image, plus making it the default in our tests via tmt configuration...

happz: In other words, @thrix @martinhoyer, do you propose any change here?

martinhoyer: Well, yeah. Make sense to me to replace 40,41 with `latest`. Not related to this PR of course.

happz: Especially when we do not have `localhost/tmt/container/test/fedora/latest` alias yet.

happz: Proposed https://github.com/teemtee/tmt/pull/3617, one day we might switch tests to use `fedora/latest`.

happz: I'm closing this conversation - the reason is mentioned above, but I agree it would be possible to do it in a better way, if we dive into providing some "default foo" configuration for tmt.

PR #3617, opened 2025-03-20
Add "latest Fedora" container image for tests to use

happz:
To avoid hardcoding `fedora/41` or `fedora/rawhide`, we can use `fedora/latest` and have the released Fedora at our disposal.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/containers/fedora/latest/Containerfile
+++ b/containers/fedora/latest/Containerfile
@@ -15,8 +15,4 @@ dnf makecache
 
 # Make sure the image is built with the latest packages
 dnf update -y
-
-# Inject `dnf5` to make things more complicated for `dnf` family of
-# package manager implementations
-dnf install -y dnf5
 EOF
--- a/containers/fedora/latest/Containerfile
+++ b/containers/fedora/latest/Containerfile
@@ -1,0 +1,22 @@
+#
+# A latest Fedora image tailored for tmt test suite
+#
+# tmt/tests/fedora/latest:latest
+#
+
+FROM quay.io/fedora/fedora:latest
+
+# hadolint ignore=DL3040
+RUN <<EOF
+set -ex
+
+# Populate dnf cache
+dnf makecache
+
+# Make sure the image is built with the latest packages
+dnf update -y
+
+# Inject `dnf5` to make things more complicated for `dnf` family of
+# package manager implementations
+dnf install -y dnf5
+EOF
--- a/containers/fedora/latest/Containerfile.unprivileged
+++ b/containers/fedora/latest/Containerfile.unprivileged
@@ -1,0 +1,26 @@
+#
+# A latest Fedora image tailored for tmt test suite, with unprivileged account & password-less sudo
+#
+# tmt/tests/fedora/latest/unprivileged:latest
+#
+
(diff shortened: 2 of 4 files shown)


containers/fedora/latest/Containerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN <<EOF
+set -ex
+
+# Populate dnf cache
+dnf makecache
+
+# Make sure the image is built with the latest packages
+dnf update -y
+
+# Inject `dnf5` to make things more complicated for `dnf` family of
```
martinhoyer, line 19: Isn't this redundant with current `latest` being 41?

containers/fedora/latest/Containerfile.unprivileged
```diff
@@ -0,0 +1,26 @@
...
+dnf makecache
+
+# Make sure the image is built with the latest packages
+dnf update -y
+
+# Create unprivileged user and setup sudo for it
+dnf install -y /usr/sbin/useradd
+useradd fedora
+usermod -aG wheel fedora
+printf "fedora\tALL=(ALL)\tNOPASSWD: ALL" >> /etc/sudoers
```
martinhoyer, line 23: Perhaps it's better to use /etc/sudoers.d/?
```bash
echo "fedora ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/fedora
chmod 0440 /etc/sudoers.d/fedora
```

martinhoyer (commented):
Thanks for this!  

Looking at it now, aren't we missing `dnf clean all` in the ContainerFiles to make the images smaller?   

Sorry for commenting on things not related to the changes.

martinhoyer:
FWIW, I have two ideas I'd like to bounce:
 - instead of rawhide, using "latest + 1", so in current context, we'd run F41 and F42 (F42 being in beta and F43 images already available). Shouldn't be hard to automate?
 - aarch64 for testing some parts?
 - super not related, but since I'm doing a comment...how about aditional `test-expanded` that would between the default pipeline and `test-full` and only a subset of tests could be run when working on PR, adding full test label only when things are green. It probably doesn't make sense, but wanted to bring it up as I feel bad about the number of failed full test pipelines I had last week. :)

happz:
> Thanks for this!
> 
> Looking at it now, aren't we missing `dnf clean all` in the ContainerFiles to make the images smaller?

Not these images. We want them big, with fully-populated, up-to-date metadata, to speed up package installations during tests.

> Sorry for commenting on things not related to the changes.

No problem, I just think it might be better to open an issue or discussion, because the nature of PRs is to be forgotten once merged. All the good ideas may become lost because we forget about this PR and unfinished chats.

containers/fedora/latest/Containerfile.unprivileged
happz: True, and since it's not just this image, I filed https://github.com/teemtee/tmt/issues/3683 to fix all of them.

containers/fedora/latest/Containerfile
happz: Dropped in b2a5db27

thrix (approved):
LGTM

PR #3623, opened 2025-03-27
Allow users override default `Mrack` hardware translations with config

skycastlelily:
Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


tmt/steps/provision/mrack.py
```diff
@@ -757,23 +759,11 @@ def constraint_to_beaker_filter(
...
-    name, _, child_name = constraint.expand_name()
-
-    if child_name:
-        transformer = _CONSTRAINT_TRANSFORMERS.get(f'{name}.{child_name}')
-
-    else:
-        transformer = _CONSTRAINT_TRANSFORMERS.get(name)
-
-    if transformer:
-        return transformer(constraint, logger)
```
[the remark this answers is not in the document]
skycastlelily: wow, impressive again, I realized we may need to rename the function, but didn't realize we could implement it in such a beautiful way^^ :  https://github.com/teemtee/tmt/pull/3623/commits/a9486672f7979f95ba1faacdbbd1eb3c84c3bc5f

tests/provision/beaker/data/config/hardware.fmf
```diff
@@ -0,0 +1,6 @@
+beaker:
+  translations:
+    - requirement : cpu.processors
+      template: '{"cpu": {"processors": {"_op": "{{ BEAKER_OPERATOR }}", "_value": "{{ BEAKER_VALUE }}"}}}'
+    - requirement : cpu.model_name
+      template: '{"cpu": {"dummyname": {"_op": "{{ BEAKER_OPERATOR }}", "_value": "{{ BEAKER_VALUE }}"}}}'
```
happz, line 6: We should make this eaiser to read for us, machines won't care. `template` is supposed to be a YAML, so how about formatting it that way?

```yaml
beaker:
  translations:
    - requirement : cpu.processors
      template: |
        cpu:
          processors:
            _op: "{{ BEAKER_OPERATOR }}"
            _value: "{{ BEAKER_VALUE }}"
```

tmt/steps/provision/mrack.py
```diff
@@ -756,23 +761,15 @@ def constraint_to_beaker_filter(
...
-    return _translate_constraint_by_config(constraint, config, logger)
-
-    return _transform_unsupported(constraint, logger)
+    suitable_translations = [
+        translation
+        for translation in config
+        if translation.requirement == constraint.printable_name.replace('-', '_')
+    ]
+    if suitable_translations:
+        return _translate_constraint_by_config(constraint, suitable_translations, logger)
```
happz, line 771: Leave the translations to `_translate_constraint_by_config()`, please, it was exactly where it is supposed to be. Now you have "configuration" spread across two places: this function reads it and looks for translations, and `_translate_constraint_by_config()` applies them :)

Looking into configuration really belongs into `_translate_constraint_by_config()`. `constraint_to_beaker_filter()` will just call them, that's all:

```python
return _translate_constraint_by_config(constraint, logger) or _translate_constraint_by_transformer(constraint, logger)
```

You have a patch somewhere that changes the "unsuported" to an empty dictionary, which makes this possible: `_translate_constraint_by_config()` opens configuration if it finds a suitable translation, it will use it and return the resulting dictionary. Otherwise, it will simply `return _transform_unsupported(constraint, logger)` - which will be an empty dictionary, and the combined `return` in the proposal above will simply evaluate it as false-ish, and `or` will cause `_translate_constraint_by_transformer()` to be called. `_translate_constraint_by_transformer()` again use the same approach: checks translators, if there is a suitable one, `return` its output, otherwise `return _transform_unsupported()`.

Eventually the first one wins: `_translate_constraint_by_config()`, or `_translate_constraint_by_transformer()`, or we end up with `_transform_unsupported()`.

```python
def _translate_constraint_by_config(
    constraint: tmt.hardware.Constraint[Any],
    logger: tmt.log.Logger,
) -> BeakerizedConstraint:
    config = _get_constraint_translations()

    if not config:
        return _transform_unsupported(constraint, logger)

    suitable_translations = [
        translation
        for translation in constraint_translations
        if translation.requirement == constraint.printable_name
    ]

    if not suitable_translations:
        return _transform_unsupported(constraint, logger)

    beaker_operator, actual_value, negate = operator_to_beaker_op(
        constraint.operator, constraint.value
    )

    return tmt.utils.yaml_to_dict(
        render_template(
            suitable_translations[0].template,
            CONSTRAINT=constraint,
            BEAKER_OPERATOR=beaker_operator,
            BEAKER_VALUE=actual_value,
            BEAKER_NEGATE=negate,
        )
    )


def _translate_constraint_by_transformer(
    constraint: tmt.hardware.Constraint[Any],
    logger: tmt.log.Logger,
) -> BeakerizedConstraint:
    transformer = _CONSTRAINT_TRANSFORMERS.get(constraint.printable_name.replace('-', '_'))

    if not transformer:
        return _transform_unsupported(constraint, logger)

    return transformer(constraint, logger)


...
    return _translate_constraint_by_config(constraint, logger) or _translate_constraint_by_transformer(constraint, logger)

```

skycastlelily: > You have a patch somewhere that changes the "unsuported" to an empty dictionary, which makes this possible:

As the patch owner, I was only thinking to make the job xml a bit cleaner, but failed to realize it could help us do the magic here, thanks a lot, mentor :saluting_face: :sweat_smile: 

Updated and gonna to add block label^^
https://github.com/teemtee/tmt/pull/3623/commits/3de925312a49360017795d96ffdbb350f5b6629e

tmt/steps/provision/mrack.py
```diff
@@ -623,7 +646,8 @@ def _transform_iommu_is_supported(
     if test in [(tmt.hardware.Operator.EQ, False), (tmt.hardware.Operator.NEQ, True)]:
         return MrackHWKeyValue('VIRT_IOMMU', '==', '0')
 
-    return _transform_unsupported(constraint, logger)
+    logger.warning(f"Hardware requirement '{constraint.printable_name}' will have no effect.")
```
happz, line 649: If `_transform_virtualization_hypervisor()` or `_transform_beaker_pool()` return "unsupported" value, wouldn't the warning be printed by `constraint_to_beaker_filter()`, thanks to its final check? It seems to me these two warnings are superfluous.

```python
    if not transformed:
        logger.warning(f"Hardware requirement '{constraint.printable_name}' will have no effect.")
```

skycastlelily: yeah,updated^^ :https://github.com/teemtee/tmt/pull/3623/commits/c57794c740eb91effafe317a8500a25b52a8c90a

tmt/steps/provision/mrack.py
```diff
@@ -251,37 +251,39 @@ class MrackHWNotGroup(MrackHWGroup):
...
 
 
-def _transform_unsupported(
-    constraint: tmt.hardware.Constraint[Any], logger: tmt.log.Logger
-) -> dict[str, Any]:
+def _transform_unsupported(constraint: tmt.hardware.Constraint[Any]) -> dict[str, Any]:
     # Unsupported constraint has been already logged via report_support(). Make
     # sure user is aware it would have no effect, and since we have to return
     # something, return an {}- no harm done, composable with other elements.
-    logger.warning(f"Hardware requirement '{constraint.printable_name}' will have no effect.")
```
happz, line 260: Now the comment above is not really fitting, right? It needs to be moved with the logging call, i.e. into `constraint_to_beaker_filter()` :)

skycastlelily: ah, report_support! we should not log "Hardware requirement '{constraint.printable_name}' is not supported." any more,right? As  tmt will support everything  users put into the custom file^^

happz: Yes, we should because the user might not have a translation rule for a constraint, and that will make it unsupported. My point here is about the comment alone: it speaks about making the user aware, but the action making the user aware - `logger.warning` - has moved to a different place. The comment needs to move with it.

skycastlelily: yeah, I know what you mean, it's just I was thinking if we should remove  report_support, I could address both in one comment :smile: 
> Yes, we should because the user might not have a translation rule for a constraint, and that will make it
 > unsupported.

yeah,  I thought if user don't have  a translation (and there is no transformer),  we will logger.warning "have no effect", which should be enough for them to know  they need to do something, anyway, updated^^:
https://github.com/teemtee/tmt/pull/3623/commits/523bccfee2b462ccf315b1b8dbfd990a5b6fc707

happz: Ah, I see now, I misunderstood. I would probably leave it for some other occasion and another PR. I'd like to get this one merged.

skycastlelily: > Ah, I see now,  I misunderstood. 

Sorry for making you feel confused, and yeah, agree,let's get it merged^^

happz (approved):

therazix (commented):
Shouldn't there be a `main.fmf` file in `/tests/provision/beaker`? Otherwise, the test will not run in the pipeline.

skycastlelily:
> Shouldn't there be a `main.fmf` file in `/tests/provision/beaker`? Otherwise, the test will not run in the pipeline.

Updated:) https://github.com/teemtee/tmt/pull/3623/commits/eff451d7a9825906f59677ea0cc23b5bdc7da10f

psss:
Tests are failing with:

    [Errno 2] No such file or directory: '/etc/beaker/client.conf'

Seems that even in `--dry` mode the `mrack` plugin depends on the client configuration. Maybe we could fake something? Or convince it to look somewhere else (so that we don't need root permissions for this test)?

psss:
> Would the file be provided if we add Beaker client package as a test requirement?

I'm afraid that this does not help: The config file is provided by `beaker-redhat` and it seems there's no default config shipped in fedora repositories:

```
# dnf install /etc/beaker/client.conf
Updating and loading repositories:
 Fedora 41 openh264 (From Cisco) - x86_64             100% |   5.3 KiB/s |   8.6 KiB |  00m02s
 Fedora 41 - x86_64 - Updates                         100% |   8.2 MiB/s |  34.8 MiB |  00m04s
 Fedora 41 - x86_64                                   100% |  12.6 MiB/s |  62.0 MiB |  00m05s
Repositories loaded.
Failed to resolve the transaction:
No match for argument: /etc/beaker/client.conf
```

skycastlelily:
sorry for the late reply,I was on state holiday vacation,turns out the[ beaker reboot  test](https://github.com/teemtee/tmt/blob/main/tests/provision/reboot/test.sh#L29) won't be ran .

>>Would the file be provided if we add Beaker client package as a test requirement?

>I'm afraid that this does not help: The config file is provided by beaker-redhat and it seems there's no default config >shipped in fedora repositories:

yeah.

> Maybe we could fake something? 
   
  seems like we need mock, but not sure how to do that with hardware.sh

Or convince it to look somewhere else (so that we don't need root permissions for this test)?

we could use "BEAKER_CONF" ,just I did with latest update:https://github.com/teemtee/tmt/pull/3623/commits/b755ecafc8399bcedd7ff15ba541448f69d69388

However we need use a real account, so I guess my update won't help :sweat_smile:

happz:
> we could use "BEAKER_CONF" ,just I did with latest update:[b755eca](https://github.com/teemtee/tmt/commit/b755ecafc8399bcedd7ff15ba541448f69d69388)
> 
> However we need use a real account, so I guess my update won't help 😅

Is real user account needed even for `--dry`?

skycastlelily:
> > we could use "BEAKER_CONF" ,just I did with latest update:[b755eca](https://github.com/teemtee/tmt/commit/b755ecafc8399bcedd7ff15ba541448f69d69388)
> > However we need use a real account, so I guess my update won't help 😅
> 
> Is real user account needed even for `--dry`?

yeah, I'm afraid so :neutral_face: , because [self.conf ](https://github.com/neoave/mrack/blob/main/src/mrack/providers/beaker.py#L123)will be checked during [init](https://github.com/neoave/mrack/blob/main/src/mrack/providers/beaker.py#L102)

happz:
> > > we could use "BEAKER_CONF" ,just I did with latest update:[b755eca](https://github.com/teemtee/tmt/commit/b755ecafc8399bcedd7ff15ba541448f69d69388)
> > > However we need use a real account, so I guess my update won't help 😅
> > 
> > 
> > Is real user account needed even for `--dry`?
> 
> yeah, I'm afraid so 😐 , because [self.conf ](https://github.com/neoave/mrack/blob/main/src/mrack/providers/beaker.py#L123)will be checked during [init](https://github.com/neoave/mrack/blob/main/src/mrack/providers/beaker.py#L102)

Hmm, I guess we are stuck then, until mrack learns a way to not talk to remote API unless needed.

happz:
Ok, so what can we do here? To test the feature, we need a working Beaker account. That puts it in the same category as actual provisioning tests, like those for the `artemis` plugin, in the sense that it needs the (live) infrastructure. If we can manage to update mrack not to use API for mere building of XML, we could use the test & `--dry` and move, but until then, it seems to me all we can do is disable the test for now.

skycastlelily:
> until mrack learns a way to not talk to remote API unless needed.

Here is the mrack mr: https://github.com/neoave/mrack/pull/311 ^^

skycastlelily:
> > Ok, so what can we do here? To test the feature, we need a working Beaker account. That puts it in the same category as actual provisioning tests, like those for the `artemis` plugin, in the sense that it needs the (live) infrastructure. If we can manage to update mrack not to use API for mere building of XML, we could use the test & `--dry` and move, but until then, it seems to me all we can do is disable the test for now.
> 
> Yeah, ageed, let's include the test but disable for now in automation until the `mrack` change is merged.

+1,and disabled:https://github.com/teemtee/tmt/pull/3623/commits/8ec00e4d8221e15a6c9b5545ddc2dbb808b24f35 :)

therazix (approved):

PR #3624, opened 2025-03-27
Support '--fetch-logs' option, logs will be fetched on users' desire

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/finish/__init__.py
+++ b/tmt/steps/finish/__init__.py
@@ -234,7 +234,8 @@ class Finish(tmt.steps.Step):
 
         # Stop and remove provisioned guests, even the partially provisioned ones.
         for guest in self.plan.provision.guests:
-            guest.fetch_logs(logger=self._logger)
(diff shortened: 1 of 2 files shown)


happz:
Aren't the logs already fetched? Or is the use case to *disable* fetching of logs?

skycastlelily:
> Aren't the logs already fetched? Or is the use case to _disable_ fetching of logs?

Sorry for  the so late reply, I didn't see your comment until just now :sweat_smile: ,  yeah, tmt fetch logs by default now, If  I remember correctly, we are supposed to  disable logs  fetching by default, and only fetch logs when users specify the fetch_logs key :)

therazix:
There is currently no use case for this. If there is one, feel free to reopen.

PR #3628, opened 2025-03-27
Add anchor link to artifacts page for tests in ReportPortal

therazix:
This also needs a change on Testing Farm part.

Related to [TFT-2757](https://issues.redhat.com/browse/TFT-2757).

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -481,6 +481,24 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
                 curr_description = self.data.launch_description
         return curr_description
 
+    def append_test_url(
+        self,
+        description: str,
+        base_url: str,
+        serial_number: int,
+        guest_name: str,
+    ) -> str:
+        """
+        Extend text with the test URL (if provided)
+        """
+
+        assert self.step.plan.my_run is not None
+        assert self.step.plan.my_run.workdir is not None
+        run_id = self.step.plan.my_run.workdir.name
+        plan_name = self.step.plan.pathless_safe_name
+        url = f'{base_url}#{run_id}_{plan_name}_{serial_number}_{guest_name}'
+        return f'{description}<br>{url}' if description else url
+
     def upload_result_logs(
         self,
         result: tmt.result.BaseResult,
@@ -755,6 +773,14 @@ class ReportReportPortal(tmt.steps.report.ReportPlugin[ReportReportPortalData]):
                         self.data.upload_to_launch and launch_per_plan
                     ) or self.data.upload_to_suite:
                         test_description = self.append_description(test_description)
+
+                    if self.data.artifacts_url and result:
+                        test_description = self.append_test_url(
+                            test_description,
+                            self.data.artifacts_url,
+                            serial_number,
+                            result.guest.name,
+                        )
 
                     # Create a test item
                     self.info("test", test_name, color="cyan")


docs/releases.rst
```diff
@@ -16,6 +16,13 @@ results are represented as ``<testsuite>`` tags, each containing a
 ``<testcase>`` tag for the main result, along with additional
 ``<testcase>`` tags for any subresults.
 
+When reporting results to ReportPortal, each test result can now
+directly link to ``oculus`` test result. To achieve this, a new
+option ``--link-template`` was added to the
+:ref:`/plugins/report/reportportal` plugin, which can be used
+to provide a template that will be rendered for each test result
+and appended to the end of its description.
```
happz, line 24: * I'd say the scope is much bigger than Oculus, correct? If I give my `reportportal` plugin `link-template: "https://github.com/my/project/issues/{{ TMT_PLAN_NAME }}_{{ TMT_TEST_SERIAL_NUMBER }}_{{ TMT_GUEST_NAME }}"`, I can do it. It would not work, but who knows, maybe there will be another service like TF. The point is, the template can be unrelated to TF. I'd try to sell it like that - "each test result can now link directly to a URL. To achieve this, key, plugin, yada yada yada..." - then I would add TF as the motivation: "In cooperation with Testing Farm, this will provide results, reported to Report Portal, with direct links to their respective artifacts." or something similar.
* "option ``--link-template``" should be "key ``link-template``" (trying to root out the CLI-ish lingo when docs are mostly about fmf).

tmt/steps/report/reportportal.py
```diff
@@ -268,6 +269,17 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
         help="Enable/disable uploading of tmt subresults into the ReportPortal.",
     )
 
+    link_template: Optional[str] = field(
+        metavar="LINK_TEMPLATE",
```
happz, line 273: `TEMPLATE` would be enough - I already know it's a link template :)

tmt/steps/report/reportportal.py
```diff
@@ -268,6 +269,17 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
...
+        option="--link-template",
+        default=_str_env_to_default('link_template', None),
+        help="""
+             Jinja template that will be rendered for each test result and appended to the end
+             of its description. It can contain the following variables for tmt to render:
+             ``TMT_PLAN_NAME``, ``TMT_TEST_SERIAL_NUMBER``, ``TMT_GUEST_NAME``.
```
happz, line 279: * I think we don't need `TMT_*` prefix, no one else but tmt will be rendering this template
* hmm, how about providing the whole `Result` instance? We would still need a variable for the plan name, but the rest limits what data the template can use, and that's often prone to missing something interesting. We could just pass in `result.to_serialized()` as `RESULT`, and change the template: `https://github.com/my/project/issues/{{ PLAN_NAME }}_{{ RESULT.serial_number }}_{{ RESULT.guest.name }}`

docs/releases.rst
therazix: Changed in d1a17dc3d9caa1911c685284c47cfc6dca213d76.

tmt/steps/report/reportportal.py
therazix: Changed in d1a17dc3d9caa1911c685284c47cfc6dca213d76.

tmt/steps/report/reportportal.py
therazix: I agree, providing the whole result is better. Currently passing as `RESULT=result` because `to_serialized()` replaces underscores with dashes. Changed in d1a17dc3d9caa1911c685284c47cfc6dca213d76.

happz (approved):

psss:
Oh, just realized we've not updated the schema. Should be fixed in #3641.

PR #3631, opened 2025-03-28
Deduplicate the `provision` step documentation

mcasquer:
Deduplicates the `provision` step documentation, which includes the following:
* Moving all the information from the specs to the plugins section
* Updates the header in the plugins section
* Remove completely (whenever is possible) the references in the specs
* Fixes some references to the removed documentation
* Directly refer the reader to the `provision plugins` and add `implemented-by` link 

Pull Request Checklist

* [x] adjust plugin docstring


mcasquer:
> @psss I've realized that for the `provision` step the `kickstart` component is not present in the plugins section. As the intention of the PR is 'documentation move' I've not added new components to the plugins. Is there a class for this plugin? Or what approach should we take here?

With this comment I mean that if we are going to remove all the info from the specs, `kickstart` will be no longer present in the TMT docs :)

thrix:
@mcasquer could you add more information to the description so reviewers who do not understand where is the duplication can more easily follow up to understand why and how we are addressing this. The description is basically empty, and thus I need to understand myself. Please be verbose, it makes the review then easier!

mcasquer:
> @mcasquer could you add more information to the description so reviewers who do not understand where is the duplication can more easily follow up to understand why and how we are addressing this. The description is basically empty, and thus I need to understand myself. Please be verbose, it makes the review then easier!

@thrix sorry for that, description updated !

spec/plans/provision.fmf
```diff
@@ -29,73 +20,8 @@ example: |
...
-    link:
-      - implemented-by: /tmt/steps/provision/testcloud.py
-
-/local:
-    summary: Use the localhost for testing
-    description:
-        Do not provision any system. Tests will be executed
-        directly on the localhost. Note that for some actions like
-        installing additional packages you need root permission or
-        enabled sudo.
```
psss, line 69: This text should be moved to the plugin.

spec/plans/provision.fmf
```diff
@@ -29,73 +20,8 @@ example: |
...
-            how: container
-            image: fedora:latest
-      - |
-        # Use an image with a non-root user with sudo privileges,
-        # and run scripts with sudo.
-        provision:
-            how: container
-            image: image with non-root user with sudo privileges
-            user: tester
-            become: true
```
psss, line 96: This example has been lost in the migration.

spec/plans/provision/artemis.fmf
```diff
@@ -1,25 +1,4 @@
 summary: Provision a guest via an Artemis service
-description: |
```
psss, line 2: This file should be dropped completely.

tmt/steps/provision/artemis.py
```diff
@@ -736,6 +751,12 @@ class ProvisionArtemis(tmt.steps.provision.ProvisionPlugin[ProvisionArtemisData]
             image: Fedora
             api-url: https://your-artemis.com/
 
+    .. note::
+
+        When used together with TF infrastructure
```
psss, line 756:
```suggestion
        When used together with the Testing Farm infrastructure
```

Let's prevent using too much abbreviations in the docs.

psss: Similar below.

psss (changes requested):
Thanks for preparing this. Added a couple comments. Please, also remove the `openstack` mention. It was only included as an example and never implemented.

spec/plans/provision.fmf
mcasquer: Retrieved !

spec/plans/provision.fmf
mcasquer: Done !

spec/plans/provision/artemis.fmf
mcasquer: Done !

tmt/steps/provision/artemis.py
mcasquer: Updated !

psss (approved):
Nice, thanks for addressing all comments! We're good to go!

therazix (approved):

PR #3635, opened 2025-04-01
fmf: remove the tailing slashes from the test names

lxbsz:
Before this the tests will be missed without giving any logs and it's hard to debug exactly what has happened.

This just remove the tailing the slashed from the test names.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -538,7 +538,7 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
         for filter_ in filters:
             self.info('filter', filter_, 'green')
         # Names of tests selected by --test option
-        names = self.get('test', [])
+        names = [name.rstrip('/') for name in self.get('test', [])]
         if names:
             self.info('tests', fmf.utils.listed(names), 'green')
 


happz:
I'm not sure we could do this, a trailing slash may very well be there on purpose:

```
(dev) happz@multivac /tmp/foo $ cat main.fmf
/foo:
  test: /bin/true
  duration: 5m

/foo/bar:
  duration: 10m

/foo/baz:
  duration: 10m

/foo-quax:
  test: /bin/true
  duration: 15m
(dev) happz@multivac /tmp/foo $ tmt test ls
/foo/bar
/foo/baz
/foo-quax
(dev) happz@multivac /tmp/foo $ tmt test ls /foo/
/foo/bar
/foo/baz
(dev) happz@multivac /tmp/foo $ tmt test ls /foo
/foo/bar
/foo/baz
/foo-quax
(dev) happz@multivac /tmp/foo $
```

With your patch, `/foo/quax` would be included in `tmt test ls /foo/` which is not what I asked for. And the `test` key of `discover/fmf` behaves in the very same way.

Can you share your use case or reproducer so we can better understand what you are facing?

lxbsz:
> I'm not sure we could do this, a trailing slash may very well be there on purpose:
> 
> ```
> (dev) happz@multivac /tmp/foo $ cat main.fmf
> /foo:
>   test: /bin/true
>   duration: 5m
> 
> /foo/bar:
>   duration: 10m
> 
> /foo/baz:
>   duration: 10m
> 
> /foo-quax:
>   test: /bin/true
>   duration: 15m
> (dev) happz@multivac /tmp/foo $ tmt test ls
> /foo/bar
> /foo/baz
> /foo-quax
> (dev) happz@multivac /tmp/foo $ tmt test ls /foo/
> /foo/bar
> /foo/baz
> (dev) happz@multivac /tmp/foo $ tmt test ls /foo
> /foo/bar
> /foo/baz
> /foo-quax
> (dev) happz@multivac /tmp/foo $
> ```
> 
> With your patch, `/foo/quax` would be included in `tmt test ls /foo/` which is not what I asked for. And the `test` key of `discover/fmf` behaves in the very same way.
> 
> Can you share your use case or reproducer so we can better understand what you are facing?

Hi @happz 

I just was trying to run some tests recently and added a `/` in the end by mistake and found that the corresponding test was never triggerred without any logs or hint. 

And I didn't see any doc about this too. It just took me a lot of time to figure out what exactly has cuased this.

Such as the following problem test:

```
$ cat test.fmf

summary: kernel tests
discover:
  - name: kernel tests
    how: fmf
    url: https://XXX/kernel.git
    ref: master
    test:
      - /vm/page_alloc/
...
```

happz:
> > With your patch, `/foo/quax` would be included in `tmt test ls /foo/` which is not what I asked for. And the `test` key of `discover/fmf` behaves in the very same way.
> > Can you share your use case or reproducer so we can better understand what you are facing?
> 
> Hi @happz
> 
> I just was trying to run some tests recently and added a `/` in the end by mistake and found that the corresponding test was never triggerred without any logs or hint.

I see. But that is complicated, right? It is really hard for tmt to realize you might have made a mistake and didn't really want to run `/foo/`, but `/foo`.

> And I didn't see any doc about this too. It just took me a lot of time to figure out what exactly has cuased this.

You did :) This is definitely a user mistake, and while I, for one, would like to help you out, I don't see the removal of trailing `/` as the right move because it will bite another group of users, the group that uses the trailing `/` on purpose.

-1 from me for mangling the name pattern, but maybe we could find some heuristic and guess when the user wanted something else (yeah, and that's the complicated bit). Maybe if the `tests` key ends with a slash, we could run the name-based filter twice, the second time without the trailing slash, and if the set of filtered names would not match the original list, the one with the slash included, emit a warning or a hint?

lxbsz:
> > > With your patch, `/foo/quax` would be included in `tmt test ls /foo/` which is not what I asked for. And the `test` key of `discover/fmf` behaves in the very same way.
> > > Can you share your use case or reproducer so we can better understand what you are facing?
> > 
> > 
> > Hi @happz
> > I just was trying to run some tests recently and added a `/` in the end by mistake and found that the corresponding test was never triggerred without any logs or hint.
> 
> I see. But that is complicated, right? It is really hard for tmt to realize you might have made a mistake and didn't really want to run `/foo/`, but `/foo`.
> 
> > And I didn't see any doc about this too. It just took me a lot of time to figure out what exactly has cuased this.
> 
> You did :) This is definitely a user mistake, and while I, for one, would like to help you out, I don't see the removal of trailing `/` as the right move because it will bite another group of users, the group that uses the trailing `/` on purpose.
> 
> -1 from me for mangling the name pattern, but maybe we could find some heuristic and guess when the user wanted something else (yeah, and that's the complicated bit). Maybe if the `tests` key ends with a slash, we could run the name-based filter twice, the second time without the trailing slash, and if the set of filtered names would not match the original list, the one with the slash included, emit a warning or a hint?

Yeah, agree. I didn't notice that the tailing `/` has other purpose. Maybe we should also document this somewhere ?

lukaszachy:
There is one use case for `/` but it is a bit niche. With https://fmf.readthedocs.io/en/stable/features.html#select  one can make parent node to be presented as test itself:
```
$ cat foo.fmf
/:
  select: true
test: echo
/child1:
```

So using `/` as `foo/` can be used to select only `/foo/child1` .

happz:
Adding `discuss` so we could resolve the questions above, and either proceed or close the PR.

xiubli:
@lukaszachy @happz Thanks for your feedback, sorry I missed last replies. 

@lukaszachy Yeah, you are right, this `/` is useful for our recently project, we will use it freqently. 

Let's close this PR.

PR #3636, opened 2025-04-02
Logger's print() no longer shifts the output

happz:
As discussed offline, `print()` will no longer shift the output depending on `shift` or `level` parameters. It will remain focused on the output rather than logging - and logging is the tree-like pattern when emitted by tmt, while output should not be indented this way.

If a particular callsite wishes to indent (part of) its output, e.g. to improve readability of the output, it needs to do it on its own.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/log.py
+++ b/tmt/log.py
@@ -767,7 +767,6 @@ class Logger:
         self,
         text: str,
         color: Optional[str] = None,
-        shift: int = 0,
     ) -> str:
         """
         Format the given text in a way suitable for :py:meth:`print`
@@ -777,7 +776,7 @@ class Logger:
             text,
             # Always apply colors - message can be decolorized later.
             color=color,
-            level=shift + self._base_shift,
+            level=0,
             labels=self.labels,
             labels_padding=self.labels_padding,
         )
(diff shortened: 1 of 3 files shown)


martinhoyer (approved):
@happz how do we feel about [Rich](https://github.com/Textualize/rich)?

happz:
> @happz how do we feel about [Rich](https://github.com/Textualize/rich)?

I like it a lot, and Textual even more. I was thinking about using it instead of colorization and logging "subsystems" in tmt, but when I needed to fix logging, Rich versions available were pretty outdated, so I went in and did a custom implementation. Plus, tmt logging is weird, indentations and stuff, so the benefits wouldn't be that good. One day, it would be nice to use it instead of colorama, for example - I personally target colors (once I'm done with CLI themes), traceback printing (again, needs care, we need history, multiple causes, and logging into file(s), so who knows what Rich would save us), there are one or two progress bars, ReST rendering for command line might be slightly simpler, etc. Nothing big, switching to exception rendering to Rich would affect a couple of functions in `tmt.utils`, not the rest of code, but colors and Rich's style tags in general would be more impactful (kind of trying to get closer to it with `tmt.utils.themes`, I imagine one day we would just change `Style` internals from colorama/click to rich and that would be it).

So, yeah, I have plans and ideas, and I'm slowly moving in that direction, although still in the phase of sorting out internal loose ends.

psss:
One more test needed adjustment, fixed in adc575ce.

psss:
Printed messages can contain labels: One more adjustment: e71dea0e

PR #3661, opened 2025-04-11
lint check fail instead of warn

bajertom:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -5,6 +5,13 @@
 ======================
 
 
+tmt-1.47.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``tmt lint`` command now fails with an exit code 1 instead of a
+warn with an exit code 0 when an invalid key is encountered in fmf
+metadata.
+
 tmt-1.46.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -1086,7 +1086,7 @@ class Core(
                         yield from detect_unallowed_properties_with_pattern(suberror)
                         yield from detect_enum_violations(suberror)
 
-            yield LinterOutcome.WARN, 'fmf node failed schema validation'
+            yield LinterOutcome.FAIL, 'fmf node failed schema validation'
 
             return
 


bajertom:
* **#3661** <a href="https://app.graphite.dev/github/pr/teemtee/tmt/3661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/teemtee/tmt/3661?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

happz:
I'm not sure we're ready for schema validation being a hard fail lint check because it will report many false positives :/ We need at least a way users could override it for a given repo, in-repo config file, or something, otherwise, it may be complicated. Imagine a `provision` without `how` - tmt will work, it will fill in `virtual`, but schema will complain...

martinhoyer (approved):

happz:
As agreed, we will first teach tmt to recognize when to use the corresponding default for the given step.

* extend schema-based validation first, to understand the default value of `how`, and use proper schema
* plan schema is composed in https://github.com/teemtee/tmt/blob/4a3fd3b994d6aaec0e3911c32b6924cfdba4b18e/tmt/utils/__init__.py#L4455
* default for `provision` is `virtual`, https://github.com/teemtee/tmt/blob/4a3fd3b994d6aaec0e3911c32b6924cfdba4b18e/tmt/schemas/provision/virtual.yaml#L17 - maybe load `virtual` schema and patch `how` to trigger for `how: null`

bajertom:
I have used the `_process_step_collection` method to check whether the step's `how` is `None` and if so, it is loaded with the default from the above `_process_step`. This makes the linter not complain if the `how` is not set, because it always defaults to something, but will fail if anything incorrect is given.

However it can result in something like this (`execute:` with nothing in plan):
```
/bad
fail P002 execute step must be defined with "how"
```
It passes C000, because `tmt` gets filled in, but then fails on P002, because it was not defined specifically with `how: something`. Maybe this check is redundant since tmt always fills in the default value for `how`?

tests/lint/plan/test.sh
```diff
@@ -30,7 +30,7 @@ rlJournalStart
 
     rlPhaseStartTest "Bad"
         rlRun -s "$tmt plan lint bad" 1
-        rlAssertGrep "warn C000 fmf node failed schema validation" $rlRun_LOG
+        rlAssertGrep "pass C000 fmf node passes schema validation" $rlRun_LOG
```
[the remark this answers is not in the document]
bajertom: It actually does pass due to the changes in PR. Fails elsewhere, is it's OK (maybe, I've addressed this in a comment above)

tmt/utils/__init__.py
```diff
@@ -4825,6 +4825,15 @@ def _process_step_collection(step_name: str, step_collection: Any) -> None:
 
             return
 
+        # Handle None/empty step configuration (e.g., "provision:" with no value)
+        # comment
```
tcornell-bus: comment?

docs/releases.rst
```diff
@@ -13,6 +13,10 @@ limit for each test.
 The provision plugins now support setting an ``environment`` key.
 See the documentation of :ref:`/plugins/provision`.
 
+The ``tmt lint`` command now fails with an exit code 1 instead of a
+warn with an exit code 0 when an invalid key is encountered in fmf
+metadata.
```
tcornell-bus: Should something like this be mentioned as well? "tmt lint schema validation is now aware of default values"

tmt/utils/__init__.py
bajertom: Good catch, leftover from when I was forcing pre-commit^^. Removed.

docs/releases.rst
bajertom: Addressed in https://github.com/teemtee/tmt/pull/3661/commits/b62187996659e38ec0080f72d9d32a86fd67af1a

tmt/utils/__init__.py
```diff
@@ -4825,6 +4825,14 @@ def _process_step_collection(step_name: str, step_collection: Any) -> None:
 
             return
 
+        # Handle None/empty step configuration (e.g., "provision:" with no value)
+        if step_collection is None:
+            empty_step: dict[Any, Any] = {}
```
happz: A list would be better, the natural content of a step is a list of mappings, a list of phases. A step being a mapping is a special case to make users happier.

Would this work?

```python
if step_collection is None:
    step_collection = node.data[step_name] = []
```

And then leave the rest to `if isinstance(step_collection, list):` below?

bajertom: It works well, I'm adjusting the tests now to reflect the new reality.

tcornell-bus (approved):

tmt/utils/__init__.py
```diff
@@ -4826,6 +4826,12 @@ def _process_step_collection(step_name: str, step_collection: Any) -> None:
 
             return
 
+        # Handle None/empty step configuration (e.g., "provision:" with no value)
+        if step_collection is None:
+            step_collection = node.data[step_name] = []
```
happz:
```suggestion
            node.data[step_name] = []
```

happz (approved):

PR #3662, opened 2025-04-11
Update pre-commit check versions and resolve failures

martinhoyer:
More or less required for https://github.com/teemtee/tmt/pull/3403 and https://github.com/teemtee/tmt/pull/2923#

Notable is https://docs.astral.sh/ruff/rules/custom-type-var-for-self/ and unused imports reported by new versions of Pyright.

Definitely needs @happz review.

### DIFF
--- a/docs/scripts/generate-hardware-matrix.py
+++ b/docs/scripts/generate-hardware-matrix.py
@@ -48,9 +48,9 @@ def main() -> None:
         if hw_requirement == 'arch':
             continue
 
-        matrix[hw_requirement] = {method: (False, int) for method in known_methods}  # noqa: C420
         # C420: current implementation creates a new tuple for each method
         # https://github.com/teemtee/tmt/pull/3662#discussion_r2040669353
+        matrix[hw_requirement] = {method: (False, int) for method in known_methods}  # noqa: C420
 
         if not story.link:
             pass
--- a/docs/scripts/generate-hardware-matrix.py
+++ b/docs/scripts/generate-hardware-matrix.py
@@ -48,7 +48,7 @@ def main() -> None:
         if hw_requirement == 'arch':
             continue
 
-        matrix[hw_requirement] = {method: (False, int) for method in known_methods}
+        matrix[hw_requirement] = dict.fromkeys(known_methods, (False, int))
 
         if not story.link:
             pass
--- a/docs/scripts/generate-hardware-matrix.py
+++ b/docs/scripts/generate-hardware-matrix.py
@@ -48,7 +48,9 @@ def main() -> None:
         if hw_requirement == 'arch':
             continue
 
-        matrix[hw_requirement] = dict.fromkeys(known_methods, (False, int))
+        matrix[hw_requirement] = {method: (False, int) for method in known_methods}  # noqa: C420
+        # C420: current implementation creates a new tuple for each method
+        # https://github.com/teemtee/tmt/pull/3662#discussion_r2040669353
 
         if not story.link:
             pass
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -786,17 +786,19 @@ def test_get_distgit_handler():
             tmt.utils.git.get_distgit_handler([])
     # Fedora detection
     returned_object = tmt.utils.git.get_distgit_handler(
-        """
-        remote.origin.url ssh://lzachar@pkgs.fedoraproject.org/rpms/tmt
-        remote.lzachar.url ssh://lzachar@pkgs.fedoraproject.org/forks/lzachar/rpms/tmt.git
-        """.split('\n')
+        [
+            "remote.origin.url ssh://lzachar@pkgs.fedoraproject.org/rpms/tmt",
+            "remote.lzachar.url ssh://lzachar@pkgs.fedoraproject.org/forks/lzachar/rpms/tmt.git",
+        ]
     )
     assert isinstance(returned_object, tmt.utils.git.FedoraDistGit)
     # CentOS detection
     returned_object = tmt.utils.git.get_distgit_handler(
-        """
-        remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git
-        """.split('\n')
+        [
+            "",
+            "        remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git",
+            "        ",
+        ]
     )
     assert isinstance(returned_object, tmt.utils.git.CentOSDistGit)
     # RH Gitlab detection
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -795,9 +795,7 @@ def test_get_distgit_handler():
     # CentOS detection
     returned_object = tmt.utils.git.get_distgit_handler(
         [
-            "",
-            "        remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git",
-            "        ",
+            "remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git",
         ]
     )
     assert isinstance(returned_object, tmt.utils.git.CentOSDistGit)
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -58,6 +58,7 @@ import tmt.templates
 import tmt.utils
 import tmt.utils.git
 import tmt.utils.jira
+from tmt._compat.typing import Self
 from tmt.checks import Check
 from tmt.container import (
     SerializableContainer,
@@ -788,7 +789,7 @@ class Core(
         return self.name
 
     @classmethod
-    def from_tree(cls: type[T], tree: 'tmt.Tree') -> list[T]:
+    def from_tree(cls, tree: 'tmt.Tree') -> list[Self]:
         """
         Gather list of instances of this class in a given tree.
 
@@ -799,7 +800,7 @@ class Core(
         :param tree: tree to search for objects.
         """
 
-        return cast(list[T], getattr(tree, f'{cls.__name__.lower()}s')())
+        return cast(list[Self], getattr(tree, f'{cls.__name__.lower()}s')())
 
     def _update_metadata(self) -> None:
         """
@@ -3999,7 +4000,7 @@ class Run(tmt.utils.Common):
             self._plans = self.tree.plans(run=self, names=plan_names)
 
         # Initialize steps only if not selected on the command line
-        step_options = 'all since until after before skip'.split()
+        step_options = ['all', 'since', 'until', 'after', 'before', 'skip']
         selected = any(self.opt(option) for option in step_options)
         assert self._cli_context_object is not None  # narrow type
         if not selected and not self._cli_context_object.steps:
--- a/tmt/cli/__init__.py
+++ b/tmt/cli/__init__.py
@@ -96,7 +96,7 @@ class ContextObject:
     common: tmt.utils.Common
     fmf_context: tmt.utils.FmfContext
     tree: tmt.Tree
-    steps: set[str] = simple_field(default_factory=lambda: set[str]())
+    steps: set[str] = simple_field(default_factory=set[str])
     clean: Optional[tmt.Clean] = None
     clean_logger: Optional[tmt.log.Logger] = None
     clean_partials: collections.defaultdict[str, list[tmt.base.CleanCallback]] = simple_field(
--- a/tmt/cli/__init__.py
+++ b/tmt/cli/__init__.py
@@ -96,7 +96,7 @@ class ContextObject:
     common: tmt.utils.Common
     fmf_context: tmt.utils.FmfContext
     tree: tmt.Tree
-    steps: set[str] = simple_field(default_factory=set)
+    steps: set[str] = simple_field(default_factory=set[str])
     clean: Optional[tmt.Clean] = None
     clean_logger: Optional[tmt.log.Logger] = None
     clean_partials: collections.defaultdict[str, list[tmt.base.CleanCallback]] = simple_field(
--- a/tmt/cli/__init__.py
+++ b/tmt/cli/__init__.py
@@ -96,7 +96,7 @@ class ContextObject:
     common: tmt.utils.Common
     fmf_context: tmt.utils.FmfContext
     tree: tmt.Tree
-    steps: set[str] = simple_field(default_factory=set[str])
+    steps: set[str] = simple_field(default_factory=lambda: set[str]())
     clean: Optional[tmt.Clean] = None
     clean_logger: Optional[tmt.log.Logger] = None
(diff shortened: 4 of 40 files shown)


tests/unit/test_utils.py
```diff
@@ -789,17 +789,19 @@ def test_get_distgit_handler():
...
     assert isinstance(returned_object, tmt.utils.git.FedoraDistGit)
     # CentOS detection
     returned_object = tmt.utils.git.get_distgit_handler(
-        """
-        remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git
-        """.split('\n')
+        [
+            "",
+            "        remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git",
+            "        ",
```
happz, line 803: This one's weird, especially the leading whitespace, empty first line, and whitespace-only last line. Could be just one string, the rest coming from the multiline string without decent?

tmt/cli/__init__.py
```diff
@@ -96,7 +96,7 @@ class ContextObject:
     common: tmt.utils.Common
     fmf_context: tmt.utils.FmfContext
     tree: tmt.Tree
-    steps: set[str] = simple_field(default_factory=set)
+    steps: set[str] = simple_field(default_factory=set[str])
```
happz, line 99: Interesting, first time seeing this kind of annotation. AFAIK, `default_factory` is supposed to be a parameter-less callable returning the default value, can `set[str]` be that?

tmt/hardware.py
```diff
@@ -514,15 +515,15 @@ class Constraint(BaseConstraint, Generic[ConstraintValueT]):
 
     @classmethod
     def _from_specification(
-        cls: type[T],
+        cls,
         name: str,
         raw_value: str,
         as_quantity: bool = True,
-        as_cast: Optional[Callable[[str], ConstraintValueT]] = None,
+        as_cast: Optional[Callable[[str], Any]] = None,
```
happz, line 522: This change seems like a downgrade, instead of `int, 'Size', str, bool, float`, any type is allowed, which is less strict and opens the door to unexpected types sneaking in. Is it something new linters complained about?

tmt/result.py
```diff
@@ -366,9 +365,9 @@ def from_test_invocation(
         default_ids: ResultIds = {tmt.identifier.ID_KEY: invocation.test.id}
 
         for key in EXTRA_RESULT_IDENTIFICATION_KEYS:
-            value: Any = invocation.test.node.get(key)
+            value: Any = cast(Any, invocation.test.node.get(key))
 
-            default_ids[key] = None if value is None else str(value)
+            default_ids[key] = None if value is None else str(cast(object, value))
```
happz, line 370: Hmm, `str()` not accepting `Any`? New stricter rule somewhere? That `cast()` is really weird, maybe we could use `object` instead of `Any` on the line above?

tests/unit/test_utils.py
martinhoyer: So the one above was also formatted like this and I missed this one. I suspect it was a bug (leading and trailing empty strings), but didn't look too deep into get_distgit_handler().

```
>>> """
... remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git
... """.split('\n')
['', 'remote.origin.url git+ssh://git@gitlab.com/redhat/centos-stream/rpms/ruby.git', '']
```

tmt/cli/__init__.py
martinhoyer: pyright hates it when it's just `set`:
```
tmt/cli/__init__.py:99:5 - error: Type of "steps" is partially unknown
    Type of "steps" is "set[Unknown]" (reportUnknownVariableType)
```
This passes: `steps: set[str] = simple_field(default_factory=lambda: set())` , but then, looking at the usage of `steps`, it should be set[str], so maybe ` steps: set[str] = simple_field(default_factory=lambda: set[str]())` ?  

I'll do whatever you tell me @happz . Not super confident in this honestly :)

tmt/log.py
```diff
@@ -321,7 +320,7 @@ def format(self, record: logging.LogRecord) -> str:
 
         # Otherwise render the message.
         elif record.msg and record.args:
-            record.message = record.msg % record.args
+            record.message = cast(str, record.msg % record.args)
```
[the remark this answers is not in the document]
martinhoyer: oh, it passed now. Probably a leftover. Removing.

tmt/result.py
martinhoyer: huh, this one also is not needed. Reverting and double checking the rest of the changes.

tmt/cli/__init__.py
happz: Just surprised, but it does seem to work :O

```
>>> set()
set()
>>> type(set())
<class 'set'>
>>> set[str]
set[str]
>>> type(set[str])
<class 'types.GenericAlias'>
>>> type(set[str]())
<class 'set'>
>>> set[str]((1, 2, 3))
{1, 2, 3}
>>> type(set[str]((1, 2, 3)))
<class 'set'>
>>>
```

Soo, yep, sounds good.

tests/unit/test_utils.py
martinhoyer: be6b147c91eb200319cea7eed5492a90324ec963

tmt/hardware.py
martinhoyer: indeed. I should have just add noqa here instead of using Any.  
Because I'm stupid, I've tried to solve properly here, even though I'm just bumping pre-commit versions, but have issues with pint.Quantity, 'Size' TypAlias and all. Perhaps it'll be obvious to you @happz: 9db1e1a31f276c36791cc8f2859bec2ef1a5eb06

tmt/log.py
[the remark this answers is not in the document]
martinhoyer: gone in 8b6e03c7a3ac88b938026c8c009b9c408b7f82f6

tmt/hardware.py
happz: Seems harmless - the goal is to have different value types for different subclasses of `Constraint`, and this version seems to the job as well.

tmt/cli/__init__.py
```diff
@@ -96,7 +96,7 @@ class ContextObject:
     common: tmt.utils.Common
     fmf_context: tmt.utils.FmfContext
     tree: tmt.Tree
-    steps: set[str] = simple_field(default_factory=set)
+    steps: set[str] = simple_field(default_factory=lambda: set[str]())
```
happz, line 99: Nono, `default_factory` is a callable returning the default value, i.e. `set` - or `set[str]` after this patch - but not `set()` which is the actual value, and that would be fitting to `default` but not `default_factory`.

docs/scripts/generate-hardware-matrix.py
```diff
@@ -48,7 +48,7 @@ def main() -> None:
         if hw_requirement == 'arch':
             continue
 
-        matrix[hw_requirement] = {method: (False, int) for method in known_methods}
+        matrix[hw_requirement] = dict.fromkeys(known_methods, (False, int))
```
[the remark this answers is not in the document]
happz: Which is what we had there in the first place...

tmt/cli/__init__.py
martinhoyer: Thanks! using `set[str]` in a375f5b0edb88e08b1bba69daf078bb1d378dbbc

docs/scripts/generate-hardware-matrix.py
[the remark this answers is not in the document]
martinhoyer: Yay, graphite is useful. Reverted and added comment in a375f5b0edb88e08b1bba69daf078bb1d378dbbc

martinhoyer:
@coderabbitai review

falconizmi (approved):

tmt/hardware.py
martinhoyer: Any ideas for the
```
        constraint.operator,  # TODO Argument 2 to "operator_to_beaker_op" has incompatible type "Union[int, Quantity, str, bool, float]"; expected "str"  [arg-type]
```
?

happz: I'd say `str(constraint.value)` is the right move here.

docs/scripts/generate-hardware-matrix.py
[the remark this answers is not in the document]
thrix: Seems resolved

tmt/cli/__init__.py
thrix: Seems resolved.

tmt/hardware.py
thrix: @martinhoyer is this one addressed?

martinhoyer: @thrix it is now. Rebased as well.

happz:
> @happz one more pyright error appeared after rebase - [390ba10](https://github.com/teemtee/tmt/commit/390ba107f7d6e0a85e04ea1c13c755bef3a5e15c) - I'm not super confident about these bits, if you'd have a sec to verify it's not supid, it would be appreciated 🙏

* I'm not sure what the problem is there, but isn't `default_factory=list[str]` pretty much the same as that `lambda` dance, and is pyright complaining about it while allowing `default_factory=set[str]` elsewhere in the patch?
* the `help_example_values` seems to be correct - new, unexpected, because we're not used to this form, but it does make sense.

martinhoyer:
> > @happz one more pyright error appeared after rebase - [390ba10](https://github.com/teemtee/tmt/commit/390ba107f7d6e0a85e04ea1c13c755bef3a5e15c) - I'm not super confident about these bits, if you'd have a sec to verify it's not supid, it would be appreciated 🙏
> 
>     * I'm not sure what the problem is there, but isn't `default_factory=list[str]` pretty much the same as that `lambda` dance, and is pyright complaining about it while allowing `default_factory=set[str]` elsewhere in the patch?
> 
>     * the `help_example_values` seems to be correct - new, unexpected, because we're not used to this form, but it does make sense.

fwiw, the error was this:
```
tmt/container/__init__.py:102:5 - error: Type of "help_example_values" is partially unknown
    Type of "help_example_values" is "list[Unknown]" (reportUnknownVariableType)
```

From what I _(not really)_ understand, it's `list[str]` vs `list[Unknown]`/empty list literal `[]`

happz:
> > > @happz one more pyright error appeared after rebase - [390ba10](https://github.com/teemtee/tmt/commit/390ba107f7d6e0a85e04ea1c13c755bef3a5e15c) - I'm not super confident about these bits, if you'd have a sec to verify it's not supid, it would be appreciated 🙏
> > 
> > 
> > ```
> > * I'm not sure what the problem is there, but isn't `default_factory=list[str]` pretty much the same as that `lambda` dance, and is pyright complaining about it while allowing `default_factory=set[str]` elsewhere in the patch?
> > 
> > * the `help_example_values` seems to be correct - new, unexpected, because we're not used to this form, but it does make sense.
> > ```
> 
> fwiw, the error was this:
> 
> ```
> tmt/container/__init__.py:102:5 - error: Type of "help_example_values" is partially unknown
>     Type of "help_example_values" is "list[Unknown]" (reportUnknownVariableType)
> ```
> 
> From what I _(not really)_ understand, it's `list[str]` vs `list[Unknown]`/empty list literal `[]`

I think it's consistent with the changes, `foo = list()` should be `list[Unknown]` as well, because it does not say anything about the items. `help_example_values = []` is the same case, if I understand the move to `list[x]` and `set[x]` correctly. And it does make sense, there were some `cast()` calls that tried to amend exactly this issue.

tmt/container/__init__.py
```diff
@@ -98,7 +99,7 @@ class FieldMetadata(Generic[T]):
 
     #: Specific values that should be shown in the documentation as
     #: interesting examples of the field usage.
-    help_example_values: list[str] = simple_field(default_factory=list)
+    help_example_values: list[str] = simple_field(default_factory=lambda: list[str]())
```
happz, line 102: This I don't understand: `default_factory=list[str]` did not work? There's `default_factory=dict[str, Any]` below, and `default_factory=set[str]` somewhere else in the PR.

happz: It is really suspicious, so I try to dive into your branch, and pyright doesn't mind:

```
happz@multivac [precommitbump] ~/git/tmt $ git df
diff --git a/tmt/container/__init__.py b/tmt/container/__init__.py
index 90cc59a0..37e51bb4 100644
--- a/tmt/container/__init__.py
+++ b/tmt/container/__init__.py
@@ -99,7 +99,7 @@ class FieldMetadata(Generic[T]):
 
     #: Specific values that should be shown in the documentation as
     #: interesting examples of the field usage.
-    help_example_values: list[str] = simple_field(default_factory=lambda: list[str]())
+    help_example_values: list[str] = simple_field(default_factory=list[str])
 
     #: If field accepts a value, this string would represent it in documentation.
     #: This stores the metavar provided when field was created - it may be unset.

happz@multivac [precommitbump] ~/git/tmt $ pre-commit run --all-files pyright
pyright..................................................................Passed
happz@multivac [precommitbump] ~/git/tmt $ 
```

And I can even commit it.

docs/scripts/generate-hardware-matrix.py
```diff
@@ -48,7 +48,9 @@ def main() -> None:
         if hw_requirement == 'arch':
             continue
 
-        matrix[hw_requirement] = {method: (False, int) for method in known_methods}
+        matrix[hw_requirement] = {method: (False, int) for method in known_methods}  # noqa: C420
+        # C420: current implementation creates a new tuple for each method
+        # https://github.com/teemtee/tmt/pull/3662#discussion_r2040669353
```
thrix, line 53: I find it weird to have comments after the commented code, rather before

```suggestion
        # C420: current implementation creates a new tuple for each method
        # https://github.com/teemtee/tmt/pull/3662#discussion_r2040669353
        matrix[hw_requirement] = {method: (False, int) for method in known_methods}  # noqa: C420
```

tmt/base.py
```diff
@@ -58,6 +58,7 @@
 import tmt.utils
 import tmt.utils.git
 import tmt.utils.jira
+from tmt._compat.typing import Self
```
thrix, line 61: @martinhoyer did the update of pre-commit checks cause really the need of this change? 
If not, I would suggest not to add random changes to ease review.

martinhoyer: @thrix yes, it's not random. Ruff iirc

happz (approved):

tmt/base.py
happz: @thrix happy with the answer?

happz:
> Remaining test is stuck and is known issue (F42 provision)

FTR, I don't believe it's 1) a known issue, or 2) related to F42. AFAICT, it's TF not delivering the requested machine, with requests being stuck in Beaker. There are no errors, no failed installations, just a queue.

PR #3665, opened 2025-04-14
Make sure os boot entry is not at the top of boot order

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/execute/scripts/tmt-reboot
+++ b/tmt/steps/execute/scripts/tmt-reboot
@@ -25,6 +25,13 @@ done
 if [ $efi = True ]; then
     if efibootmgr &>/dev/null ; then
         os_boot_entry=$(efibootmgr | awk '/BootCurrent/ { print $2 }')
+	first_boot=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f1)
+	# if somehow os boot entry is configured to the top of boot order,
+	# we move it to the end of the order, to avoid breaking the system.
+	if [[ $first_boot == $os_boot_entry ]]; then
+	       BOOT_LIST=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f2-)
+	       efibootmgr -o "${BOOT_LIST},${BOOTC_BOOT}"
+	fi
         # fall back to /root/EFI_BOOT_ENTRY.TXT if it exists and BootCurrent is not available
         if [[ -z "$os_boot_entry" && -f /root/EFI_BOOT_ENTRY.TXT ]] ; then
             os_boot_entry=$(</root/EFI_BOOT_ENTRY.TXT)


happz:
See https://github.com/teemtee/tmt/pull/3645, seems like you and @martinhoyer are dancing around the same problem.

tmt/steps/execute/scripts/tmt-reboot
```diff
@@ -25,6 +25,13 @@
 if [ $efi = True ]; then
     if efibootmgr &>/dev/null ; then
         os_boot_entry=$(efibootmgr | awk '/BootCurrent/ { print $2 }')
+	first_boot=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f1)
+	# if somehow os boot entry is configured to the top of boot order,
+	# we move it to the end of the order, to avoid breaking the system.
+	if [[ $first_boot == $os_boot_entry ]]; then
```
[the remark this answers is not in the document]
siteshwar: Same warning is reported through [OpenScanHub](https://openscanhub.fedoraproject.org/task/49742/log/added.html).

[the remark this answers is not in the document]
skycastlelily: Thanks for your review, addressed : https://github.com/teemtee/tmt/pull/3665/commits/a659eb2f6adc4c83a6a3e67bd509368e8dedbcbb :)

skycastlelily:
>  I just saw both of you poking the same issue and the same part of the code

hmmm, np, IIUC , we are not poking the same issue, the main part of  Martin's mr is efibootmgr -n (setting next boot) to make sure efi system boot correctly to the os,while mine  is  efibootmgr -o  make sue the os boot entry not at the top of order. And looks like we are not poking the same code, though I didn't get the full picture, but it seems that he is trying to add a new script, while mine is make tmt-reboot work  more robust :)
However, I may be wrong,  Martin, feel free to ping me to talk about this :smiley:

tmt/steps/execute/scripts/tmt-reboot
```diff
@@ -25,6 +25,13 @@ done
 if [ $efi = True ]; then
     if efibootmgr &>/dev/null ; then
         os_boot_entry=$(efibootmgr | awk '/BootCurrent/ { print $2 }')
+	first_boot=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f1)
+	# if somehow os boot entry is configured to the top of boot order,
+	# we move it to the end of the order, to avoid breaking the system.
```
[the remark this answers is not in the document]
skycastlelily: Sure, user reported that bootc provision of aarch64 beaker servers will break the boot order on these aarch64 machines, requiring manual intervention to fix the pxe boot order to make the servers usable again.The original report is from slack channel:https://redhat-internal.slack.com/archives/C03BRN71JAF/p1743073805031049.
The guy who responsible for the bootc jobs make a fix on their code, and Milos think the problem should be fixed on tmt side, so do I, so here is the mr.Please note, this mr will Not make the user's fix ineffective:)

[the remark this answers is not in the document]
skycastlelily: addressed:https://github.com/teemtee/tmt/pull/3665/commits/054e337ef10530d9c8d3843618f0295a39bf5f59^^

tmt/steps/execute/scripts/tmt-reboot
```diff
@@ -25,6 +25,16 @@ done
...
+	first_boot=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f1)
+	# if somehow os boot entry is configured to the top of boot order,
+	# we move it to the end of the order. We need to make sure netboot is
+	# at the top of the boot order list. If os boot entry is at the top,
+	# the following recipes trying to reserve that server  will boot into
+	# that os instead of performing an installation on users' desire.
+	if [[ $first_boot == "$os_boot_entry" ]]; then
+	       BOOT_LIST=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f2-)
+	       efibootmgr -o "${BOOT_LIST},${BOOTC_BOOT}"
+	fi
```
happz, line 37: Hmhm, I would say the indentation needs one more level, to put new code correctly into the context of the `if` test above. `if` and `endif` are more important to bash, but to humans, the indentation matters.

skycastlelily: Addressed:https://github.com/teemtee/tmt/pull/3665/commits/62eb32040328636f0f23b7bac4443d782babfdef ^^

tmt/steps/execute/scripts/tmt-reboot
```diff
@@ -25,6 +25,16 @@ done
 if [ $efi = True ]; then
     if efibootmgr &>/dev/null ; then
         os_boot_entry=$(efibootmgr | awk '/BootCurrent/ { print $2 }')
+	first_boot=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f1)
+	# if somehow os boot entry is configured to the top of boot order,
+	# we move it to the end of the order. We need to make sure netboot is
+	# at the top of the boot order list. If os boot entry is at the top,
+	# the following recipes trying to reserve that server  will boot into
+	# that os instead of performing an installation on users' desire.
+	if [[ $first_boot == "$os_boot_entry" ]]; then
```
martinhoyer, line 34: oh wait, we are always forcing os boot entry to not be first? This seems specific to bare metal beaker-like workflow. IMHO tmt-reboot should only touch 'next-boot', no?

skycastlelily: hmm, TBO, I do overlooked the non-beaker scenario when I submitted the mr, as the whole efi part of tmt-reboot is from beakerlib's rhts-reboot . Now,after think about it carefully,it seems that this patch won't  do harm to non-beaker workflow, for virtual/container instance, it will boot/fall back to boot from os entry, and  for other bare metal workflow(I'm not sure tmt have such use case), we have bootnext to make sure it boot from os entry.
Will it make you feel safer if I remove the set boot order part under[ this line ](https://github.com/teemtee/tmt/blob/main/tmt/steps/execute/scripts/tmt-reboot#L32) ?

martinhoyer: Even some beaker machines are configured for os boot.  
imho, if somebody needs to change boot order of their machine, they might as well do themselves, but if there is a need to have it in tmt, I wouldn't put it as part of `tmt-reboot`.  

Maybe I'm missing something, but wouldn't https://github.com/teemtee/tmt/pull/3645 be enough for everyone?

martinhoyer (commented):
Needs some discussion

skycastlelily:
@happz @thrix @psss Miro mentioned that there are some complaints in beaker-admin list about the boot order issue, and there are still several days left before the end of this month, I decided to look into this today,so..here are the updates:
1) xiaofeng didn't see the problem again after he has his [changing boot order mr](https://gitlab.com/fedora/bootc/tests/bootc-workflow-test/-/merge_requests/619/diffs) merged
2)qinghua saw the problem, because she use her team's code which has  bootc install to-existing-root , but don't have similar move bootc boot option to the end of the list code(ie, my mr)
3) the complaints from the beaker-admin is  likely  because some other teams who have  bootc install to-existing-root in their code, but don't have similar code in this mr.
4) Both qinghua and xiaofeng's team use tmt-reboot, which  I guess also by other teams who are using/will use bootc install  
5)
 > Maybe I'm missing something, but wouldn't https://github.com/teemtee/tmt/pull/3645 be enough for everyone?

np, your mr will help those who want to change boot next, while mine is for those  who want to make sure os entry is not on the top of the order list 

In conclusion, both Martin and this mr are needed. Just as Milos said,We need to fix the bootc on top of the boot order list issue from tmt side, if not, other teams will run into  similar problem.Ah, maybe some of you feel confused about what the problem is, telling from the channel messages.
The problem is  with bootc install to-existing-root,  bootupd added a new boot options and configured as first boot, so after the job is finished, the following jobs will not be able to provision the server again, unless change the boot order, and make pxe as first order manually 
The reason why users don't see this on so called package mode installation, is because in that way ,beaker will add a changing- boot-order code  in [each job's  kickstart %post part](https://beaker.engineering.redhat.com/kickstart/16517840)

For Martin's concern "This seems specific to bare metal beaker-like workflow." , I could have this mr  changes boot order only if /root/EFI_BOOT_ENTRY.TXT(yeah,beaker stuff) exists,if you folks want:)

tmt/steps/execute/scripts/tmt-reboot
```diff
@@ -25,6 +25,16 @@ done
...
+	    BOOT_LIST=$(efibootmgr | awk '/BootOrder/ { print $2 }' | cut -d, -f2-)
+	    efibootmgr -o "${BOOT_LIST},${BOOTC_BOOT}"
```
qcheng-redhat: BOOTC_BOOT is not defined here ?

skycastlelily: ah, good catch, thanks for your review :smile: Addressed: https://github.com/teemtee/tmt/pull/3665/commits/bc319042137f2ce62f7a8fa3099069c661526df8

martinhoyer:
> The problem is with bootc install to-existing-root, bootupd added a new boot options and configured as first boot, so after the job is finished, the following jobs will not be able to provision the server again, unless change the boot order, and make pxe as first order manually

But why would this be solved in `tmt-reboot`?  
This will ultimately be yet another wonky workaround for beaker being dumb and obsolete. In any case, if it is being done, it should be set after first boot imho, preferably through something like kickstart post-install scripts and let `tmt-reboot` set next-boot to the booted entry as the least-worst workaround.

skycastlelily:
Just back from vacation.

> But why would this be solved in tmt-reboot?

Because users  will call  tmt-reboot  every time  after   "bootc install to-existing-root", which is the cause of the boot order issue. 
FYI,AFAIK,we only have two users,whose jobs  caused the boot order issue,both their ( xiaofeng and qinghua) issues are fixed after the code are added

> This will ultimately be yet another wonky workaround for beaker being dumb and obsolete.

I'm sorry, but I don't see how it could be wonky,tbo,I don't see how we could do it in a more proper way :)

>  In any case, if it is being done, it should be set after first boot imho,  

You mean first boot after bootudp added a new boot entry,right? I'm sorry  but I don't see how it would be better.

>let tmt-reboot set next-boot to the booted entry as the least-worst workaround.

tmt-reboot is already doing it.

martinhoyer:
Summary, as requested:  

- There is an assumption that the provisioned beaker machine always have network boot as 1st boot entry to prevent system not being available for provisioning.

- This is why tmt-reboot, and potentially [tmt-ensure-efiboot-order](https://github.com/teemtee/tmt/pull/3645) script change the "next" boot entry, so the run can continue even after boot/crash.

- This assumption can be broken, be it bootc related or otherwise.
 
 The proposed fix is to always set network boot as 1st boot entry when running tmt-reboot if a `/root/EFI_BOOT_ENTRY.TXT` file exists, suggesting this is a beaker-provisioned machine.
 
 My issues with it:

-  Not all beaker machines are set to be always re-provisioned and tmt has no way to know what is the expected boot order. 
    - example scenario: I have a dedicated bare metal machine, which is not being re-provisioned, but instead `tmt provision -h connect` is being used to run tests on it. Using this workaround would change the boot order unexpectedly, at least requiring manual intervention, setting the boot order through system mgmt/console or in worst case, leading to data loss do to beaker re-provisining the system.  
- This should be solved where it's getting broken and tmt should work with the above assumption.
- If we *really* want to have this in tmt, it should be done as part of the provisioning code, not hard-coding boot order on every tmt-reboot execution.

That's about from me on this PR. With all the wasted time on this and things like this, we could have beaker replaced with a suitable modern platform ten times already. :)

skycastlelily:
And here is my feedback summary,as requested.

> With all the wasted time on this

Yeah, I do feel I have wasted way more time than I expected

> example scenario: I have a dedicated bare metal machine, which is not being re-provisioned, but instead tmt provision -h connect is being used to run tests on it.

I do overlooked the connect scenario, if I noted it or  this info  is provided earlier,I would already close this mr :neutral_face: ,now I agree with you : tmt-reboot is not the best place  to put the code, and maybe we could move this mr's changing boot order  code into [your mr](https://github.com/teemtee/tmt/pull/3645), and broadcast users that if their test code are likely to change boot order, then they need to run your  script , or maybe just broadcast users they need to pay more attention if their code is going to change boot-order,in anyway, my mr is not needed anymore, feel free to close it.

> If we really want to have this in tmt, it should be done as part of the provisioning code, not hard-coding boot order on every tmt-reboot execution.

On xiaofeng and qinghua's scenario, the bootc boot entry dose not exist on provision process, so tmt is not able to fix the issue in that way,  however, I now agree that tmt should not hard-coding boot order on tmt-reboot:)

LecrisUT:
Is this still relevant after 32bac2ecb6566cfc5f9d024cec2a147e67fe8761?

skycastlelily:
> Is this still relevant after [32bac2e](https://github.com/teemtee/tmt/commit/32bac2ecb6566cfc5f9d024cec2a147e67fe8761)?

closing as not planed, based on https://github.com/teemtee/tmt/pull/3665#issuecomment-3105596500

PR #3669, opened 2025-04-15
Copy files with same destination in single rsync

### LINKED ISSUE #3290: Copy scripts in batch and add executability with rsync
Copying scripts one by one is not efficient and take quite some time, we could do just one `guest.push` instead of multiple, see tmt/steps/execute/__init__.py:
```
        with script as source:
            for dest in [script.path, *script.aliases]:
```
Additionally, we could implement `--executability` or `--perms`, `--chmod` rsync options instead of running chmod command.

martinhoyer:
tmt scripts, like `tmt-reboot`, etc are currently being copied one by one. This process can take a long time and this PR aims to copy all scripts in a single rsync execution.

Resolves #3290

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -117,7 +117,10 @@ class Script:
         :param filepath: Path where the script should be copied
         :returns: None for regular scripts, a Path for templates to be tracked for cleanup
         """
-        shutil.copy2(SCRIPTS_SRC_DIR / self.source_filename, filepath)
+        source_path = SCRIPTS_SRC_DIR / self.source_filename
+        # Make sure the parent directory exists
+        filepath.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(source_path, filepath)
         return None
 
 
@@ -172,6 +175,8 @@ class ScriptTemplate(Script):
         :returns: The path to the rendered template for cleanup tracking
         """
         assert self._rendered_script_path is not None
+        # Make sure the parent directory exists
+        filepath.parent.mkdir(parents=True, exist_ok=True)
         shutil.copy2(self._rendered_script_path, filepath)
         return self._rendered_script_path
 
@@ -915,9 +920,10 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
             # Push the entire staging directory content to the guest's root
             # Use options to preserve structure, links, and permissions.
             # -a implies -rlptgoD (archive mode)
-            self.debug(f"Pushing staged scripts from '{staging_root}' to guest")
+            self.debug(f"Pushing staged scripts from '{staging_root}/' to guest")
             guest.push(
-                source=staging_root,
+                # Add trailing slash to ensure directory contents are copied
+                source=Path(f"{staging_root}/"),
                 destination=Path('/'),
                 options=[
                     "-a",  # Archive mode (-rlptgoD), includes preserving permissions
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -832,7 +832,6 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
         to the guest's root directory in a single operation.
         """
         staging_root: Optional[Path] = None
-        rendered_templates: list[Path] = []
 
         try:
             # Create a temporary local staging directory
@@ -880,11 +879,7 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
                 # Copy or render the script into the staging area
                 with script:
                     # Copy the script to the staging area
-                    rendered_path = script.copy_into(local_script_path)
-
-                    # Track rendered templates for cleanup if needed
-                    if rendered_path is not None:
-                        rendered_templates.append(rendered_path)
+                    script.copy_into(local_script_path)
 
                 self.debug(f"Staged script '{script.source_filename}' to '{local_script_path}'")
 
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -2,8 +2,10 @@ import copy
 import functools
 import json
 import os
+import shutil
 import signal as _signal
 import subprocess
+import tempfile
 import threading
 from contextlib import suppress
 from tempfile import NamedTemporaryFile
@@ -28,7 +30,6 @@ from tmt.steps import Action, ActionTask, PhaseQueue, PluginTask, Step
 from tmt.steps.discover import Discover, DiscoverPlugin, DiscoverStepData
 from tmt.steps.provision import Guest
 from tmt.utils import (
-    Command,
     Path,
     ShellScript,
     Stopwatch,
@@ -149,8 +150,8 @@ class ScriptTemplate(Script):
         return self._rendered_script_path
 
     def __exit__(self, *args: object) -> None:
-        assert self._rendered_script_path
-        os.unlink(self._rendered_script_path)
+        if self._rendered_script_path and self._rendered_script_path.exists():
+            self._rendered_script_path.unlink()
 
 
 def effective_scripts_dest_dir(default: Path = DEFAULT_SCRIPTS_DEST_DIR) -> Path:
@@ -801,28 +802,106 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
 
     def prepare_scripts(self, guest: "tmt.steps.provision.Guest") -> None:
         """
-        Prepare additional scripts for testing
+        Prepare additional scripts locally in a staging directory and push them.
+
+        Creates a temporary local directory structure mirroring the intended
+        guest filesystem layout. Scripts and their aliases (as symbolic links)
+        are placed into this structure. The entire structure is then pushed
+        to the guest's root directory in a single operation.
         """
+        staging_root: Optional[Path] = None
+        rendered_templates: list[Path] = []
 
-        # Make sure scripts directory exists
-        command = Command("mkdir", "-p", f"{guest.scripts_path}")
+        try:
+            # Create a temporary local staging directory
+            staging_root = Path(tempfile.mkdtemp(prefix='tmt-scripts-staging-'))
+            self.debug(f"Created local script staging directory: {staging_root}")
 
-        if not guest.facts.is_superuser:
-            command = Command("sudo") + command
+            default_scripts_dest_dir_relative = guest.scripts_path.relative_to('/')
 
-        guest.execute(command)
+            for script in self.scripts:
+                if not script.enabled(guest):
+                    continue
 
-        # Install all scripts on guest
-        for script in self.scripts:
-            with script as source:
-                for filename in [script.source_filename, *script.aliases]:
-                    if script.enabled(guest):
-                        guest.push(
-                            source=source,
-                            destination=script.destination_path or guest.scripts_path / filename,
-                            options=["-p", "--chmod=755"],
-                            superuser=guest.facts.is_superuser is not True,
-                        )
+                # Determine the final path relative to the guest root '/'
+                if script.destination_path:
+                    # Custom destination path provided
+                    relative_script_path = script.destination_path.relative_to('/')
+                else:
+                    # Default destination path
+                    relative_script_path = (
+                        default_scripts_dest_dir_relative / script.source_filename
+                    )
+
+                # Construct the full path within the local staging directory
+                local_script_path = staging_root / relative_script_path
+
+                # Ensure the parent directory exists in the staging area
+                local_script_path.parent.mkdir(parents=True, exist_ok=True)
+
+                # Copy or render the script into the staging area
+                with script as source_path:
+                    if isinstance(script, ScriptTemplate):
+                        # ScriptTemplate renders to a temporary file in __enter__
+                        rendered_path = script._rendered_script_path
+                        assert rendered_path is not None
+                        shutil.copy2(rendered_path, local_script_path)
+                        # Track rendered templates for cleanup, even though __exit__ handles it
+                        rendered_templates.append(rendered_path)
+                    else:
+                        # Regular script, copy the source file
+                        shutil.copy2(source_path, local_script_path)
+
+                self.debug(f"Staged script '{script.source_filename}' to '{local_script_path}'")
+
+                # Create aliases as relative symbolic links within the staging area
+                # Aliases always go into the default scripts directory
+                if script.aliases:
+                    alias_dir_local = staging_root / default_scripts_dest_dir_relative
+                    alias_dir_local.mkdir(parents=True, exist_ok=True)
+
+                    # Calculate the relative path from the alias location to the script location
+                    # within the staging area using os.path.relpath, as pathlib's relative_to
+                    # doesn't directly compute paths between arbitrary directories easily.
+                    # Use noqa as os.path is needed here for correct relative path calculation.
+                    target_relative_to_alias_dir = os.path.relpath(  # noqa: TID251
+                        local_script_path, start=alias_dir_local
+                    )
+
+                    for alias in script.aliases:
+                        alias_path_local = alias_dir_local / alias
+                        try:
+                            # Create relative symlink using pathlib
+                            alias_path_local.symlink_to(target_relative_to_alias_dir)
+                            self.debug(
+                                f"Created alias symlink '{alias_path_local}' -> "
+                                f"'{target_relative_to_alias_dir}'"
+                            )
+                        except OSError as e:
+                            self.warn(f"Failed to create alias symlink '{alias_path_local}': {e}")
+
+            # Push the entire staging directory content to the guest's root
+            # Use options to preserve structure, links, and permissions.
+            # -a implies -rlptgoD (archive mode)
+            # --chmod=... ensures execute permissions are set correctly during transfer
+            # Trailing slash on source ensures contents are copied *into* destination
+            self.debug(f"Pushing staged scripts from '{staging_root}/' to guest '/'")
+            guest.push(
+                source=staging_root,  # Push the whole directory
+                destination=Path('/'),
+                options=[
+                    "-a",  # Archive mode (-rlptgoD), includes preserving permissions
+                    "--links",  # Copy symlinks as symlinks
+                ],
+                superuser=guest.facts.is_superuser is not True,
+            )
+
+        finally:
+            # Clean up the local staging directory
+            if staging_root and staging_root.exists():
+                self.debug(f"Cleaning up local script staging directory: {staging_root}")
+                shutil.rmtree(staging_root)
+            # Note: ScriptTemplate.__exit__ handles cleanup of individual rendered files
 
     def _tmt_report_results_filepath(self, invocation: TestInvocation) -> Path:
         """
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -150,8 +150,9 @@ class ScriptTemplate(Script):
         return self._rendered_script_path
 
     def __exit__(self, *args: object) -> None:
-        if self._rendered_script_path and self._rendered_script_path.exists():
-            self._rendered_script_path.unlink()
+        if self._rendered_script_path:
+            with suppress(FileNotFoundError):
+                self._rendered_script_path.unlink()
 
 
 def effective_scripts_dest_dir(default: Path = DEFAULT_SCRIPTS_DEST_DIR) -> Path:
@@ -871,6 +872,9 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
                     for alias in script.aliases:
                         alias_path_local = alias_dir_local / alias
                         try:
+                            # Remove existing file or symlink if present
+                            if alias_path_local.exists() or alias_path_local.is_symlink():
+                                alias_path_local.unlink()
                             # Create relative symlink using pathlib
                             alias_path_local.symlink_to(target_relative_to_alias_dir)
                             self.debug(
@@ -883,15 +887,14 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
             # Push the entire staging directory content to the guest's root
             # Use options to preserve structure, links, and permissions.
             # -a implies -rlptgoD (archive mode)
-            # --chmod=... ensures execute permissions are set correctly during transfer
-            # Trailing slash on source ensures contents are copied *into* destination
-            self.debug(f"Pushing staged scripts from '{staging_root}/' to guest '/'")
+            self.debug(f"Pushing staged scripts from '{staging_root}' to guest")
             guest.push(
-                source=staging_root,  # Push the whole directory
+                source=staging_root,
                 destination=Path('/'),
                 options=[
                     "-a",  # Archive mode (-rlptgoD), includes preserving permissions
                     "--links",  # Copy symlinks as symlinks
+                    "--chmod=755",  # Just in case
                 ],
                 superuser=guest.facts.is_superuser is not True,
             )
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -110,6 +110,16 @@ class Script:
     def __exit__(self, *args: object) -> None:
         pass
 
+    def copy_into(self, filepath: Path) -> Optional[Path]:
+        """
+        Copy the script to the specified filepath.
+
+        :param filepath: Path where the script should be copied
+        :returns: None for regular scripts, a Path for templates to be tracked for cleanup
+        """
+        shutil.copy2(SCRIPTS_SRC_DIR / self.source_filename, filepath)
+        return None
+
 
 @container
 class ScriptCreatingFile(Script):
@@ -153,6 +163,17 @@ class ScriptTemplate(Script):
         if self._rendered_script_path:
             with suppress(FileNotFoundError):
                 self._rendered_script_path.unlink()
+
+    def copy_into(self, filepath: Path) -> Optional[Path]:
+        """
+        Copy the rendered template script to the specified filepath.
+
+        :param filepath: Path where the script should be copied
+        :returns: The path to the rendered template for cleanup tracking
+        """
+        assert self._rendered_script_path is not None
+        shutil.copy2(self._rendered_script_path, filepath)
+        return self._rendered_script_path
 
 
 def effective_scripts_dest_dir(default: Path = DEFAULT_SCRIPTS_DEST_DIR) -> Path:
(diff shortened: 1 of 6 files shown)


martinhoyer:
@coderabbitai full review

tmt/steps/execute/__init__.py
```diff
@@ -133,8 +133,11 @@ class ScriptTemplate(Script):
     context: dict[str, str]
 
     _rendered_script_path: Optional[Path] = None
+    _delete_on_exit: bool = True  # Flag to control cleanup
```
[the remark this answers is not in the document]
martinhoyer: @coderabbitai could you reduce the amount of nitpicks you are adding in your reviews please?

tmt/steps/execute/__init__.py
```diff
@@ -794,28 +808,107 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
+                    final_target_path_on_guest = target_dir / script.source_filename
+                    # Store aliases associated with the final target path in the default dir
+                    if script.aliases:
+                        aliases_to_create[final_target_path_on_guest] = script.aliases
+                else:
+                    # Custom destination path - treat it as the final file path
+                    target_dir = destination_path.parent
+                    final_target_path_on_guest = destination_path
+                    # Ensure the custom parent directory exists
+                    mkdir_parent_cmd = Command("mkdir", "-p", f"{target_dir}")
```
happz, line 866: `str(target_dir)`?

tmt/steps/provision/__init__.py
```diff
@@ -2441,43 +2445,47 @@ def push(
         options = options or DEFAULT_RSYNC_PUSH_OPTIONS
         if destination is None:
             destination = Path("/")
+
+        sources: list[Path]
+        log_message: str
```
happz, line 2450: Repeating `self.debug(...)` does not seem like a bad approach to me. I can see why `sources` is initialized in each branch, but I feel like `log_message` has much fewer reasons to live...

tmt/steps/provision/local.py
```diff
@@ -160,7 +160,7 @@ def reboot(
 
     def push(
         self,
-        source: Optional[Path] = None,
+        source: Optional[Union[Path, list[Path]]] = None,
```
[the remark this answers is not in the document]
martinhoyer: hmm, I don't love it

tmt/steps/execute/__init__.py
martinhoyer: Gave it a thought and I like this more. I believe it's ultimately the same thing, calling __str__, but nicer/more readable?

tmt/steps/execute/__init__.py
```diff
@@ -147,8 +150,19 @@ def __enter__(self) -> Path:
...
+
+    def cleanup(self) -> None:
+        """Explicitly cleans up the rendered script file."""
+        if self._rendered_script_path and self._rendered_script_path.exists():
+            self._rendered_script_path.unlink()
+            self._rendered_script_path = None
+
+    def keep_rendered_file(self) -> None:
+        """Prevents automatic deletion on __exit__."""
+        self._delete_on_exit = False
```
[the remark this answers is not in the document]
martinhoyer: Thanks, it felt weird and now I know why :)

tmt/steps/execute/__init__.py
happz: Yes, essentialy the same. We already do have similar occurences, I don't like them, but it can be argued it's more readable.

tmt/steps/provision/__init__.py
martinhoyer: https://github.com/teemtee/tmt/commit/224287bf19fe54b49b2948bf4a854f3ed2db221e

martinhoyer:
@coderabbitai review

tmt/steps/execute/__init__.py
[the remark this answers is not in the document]
happz: BTW maybe we could use `delete_on_exit=True`, that would be `tempfile` where `delete_on_close=True` is the way to prevent temporary files to go away. The resemblance would be nice and familiar.

martinhoyer:
@coderabbitai review

tmt/steps/execute/__init__.py
```diff
@@ -794,28 +804,109 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
+                    single_alias_cmd = ShellScript(
+                        f"ln -sf {target_path.as_posix()} {link_path.as_posix()}"
+                    )
+                    if alias_count == 0:
+                        full_alias_command = single_alias_cmd
+                    else:
+                        full_alias_command &= single_alias_cmd  # Use the '&' operator (__and__)
+                    alias_count += 1
+
+            if alias_count > 0:
```
happz, line 902: If this counter exists just to prevent running commands when there are no aliases to create, and to start the command correctly, would it make sense to test their collection instead? With an accumulator, it would be more streamlined, with fewer tests (and with a few more touches):

```python
if aliases_to_create:
    alias_commands: list[ShellScript] = []

    # Use absolute path for symlink target for simplicity/robustness
    for target_path, aliases in aliases_to_create.items():
        alias_commands += [
            ShellScript(
                f"ln -sf {target_path.as_posix()} {scripts_dest_dir / alias.as_posix()}"
            )
            for alias in aliases
        ]

    self.debug("Creating script aliases on guest.")
    guest.execute(
        ' && '.join(script.to_element() for script in alias_commands),
        friendly_command="Create script aliases"
    )
```

tmt/steps/provision/__init__.py
```diff
@@ -2441,43 +2445,44 @@ def push(
...
-                    *cmd,
-                    *options,
-                    "-e",
-                    self._ssh_command.to_element(),
-                    source,
-                    f"{self._ssh_guest}:{destination}",
-                ),
-                silent=True,
+            # Construct the rsync command with potentially multiple sources
+            rsync_cmd = Command(
```
happz, line 2475: Why the move outside of `self._run_guest_command()`? Now we have a new variable that is initialized once, never modified, and used once as well. It does not help with line width, AFAICT. And the diff is also larger, because more lines have changed, but didn't, actually, just their indentation, with the exception of the line with source paths.

tmt/steps/provision/__init__.py
```diff
@@ -1607,13 +1607,19 @@ def execute(
...
         destination: Optional[Path] = None,
         options: Optional[list[str]] = None,
         superuser: bool = False,
     ) -> None:
         """
-        Push files to the guest
+        Push files or directories to the guest.
+
+        :param source: Path or list of paths on the local machine to push.
+                       If None, the plan workdir is pushed.
```
happz, line 1619: 4 spaces of indentation are enough. Well, even one is fine, Sphinx will handle it. Indenting to the beginning of the first line of the description robs you of valuable space when the parameter name gets too long:

```python
        :param source: Path or list of paths on the local machine to push.
            If None, the plan workdir is pushed.
```

Also "None" whouls be a literal, i.e. two backticks on each side. It's not just any "None", it's `None`.

tmt/steps/provision/__init__.py
```diff
@@ -1607,13 +1607,19 @@ def execute(
...
+        :param destination: Path on the guest where to push. If None, defaults to '/'.
+        :param options: List of rsync options to use.
+        :param superuser: If True, run rsync with sudo on the guest.
```
happz, line 1622: "True" should be a literal (and "None" in `destination` as well), or you can say "When set, ...".

tmt/steps/provision/__init__.py
martinhoyer: > Also "None" whouls be a literal, i.e. two backticks on each side. It's not just any "None", it's None.

Aren't we going a bit too far? :)

The indentation is...weird. You could say that like this, the second line of the sentence continues in front of the first line, as the start is Path, but what do I know. Don't like ReST anyway. 
https://peps.python.org/pep-0287/#questions-answers

https://github.com/teemtee/tmt/pull/3669/commits/dbf46bbc3ca9fc784364f1b0dca678d7a8c5c563

tmt/steps/provision/__init__.py
martinhoyer: https://github.com/teemtee/tmt/pull/3669/commits/dbf46bbc3ca9fc784364f1b0dca678d7a8c5c563

tmt/steps/execute/__init__.py
martinhoyer: ah, make sense thanks! Is this ok? https://github.com/teemtee/tmt/pull/3669/commits/d8cbf3bac02de8f9f28afb46d1caf21135091088

The ShellScript object is bane of my existence :)

I've tried to do yours, which and  the `&` thing, and ultimately went with newlines and single script,  but don't like it anyway.
Your snippet wouldn't work as `"str" has no attribute "as_posix"`

tmt/steps/provision/__init__.py
martinhoyer: indeed, thanks for noticing. b7e5dde33bbcbec8a3b257d7363cf9231c00c899

martinhoyer:
@coderabbitai review

tmt/steps/execute/__init__.py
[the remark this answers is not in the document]
martinhoyer: Forgot to mention 2d94f3f754ae7ddcfce87dde3a586a1e12093b8a, thanks

thrix:
> Well, it wasn't as simple as I initially thought, hope this works. The current tests should verify functionality I believe.
> 
> Resolves #3290
> 

This is not a very useful description of the merge request, could you provide reasonable description so one does nto need to look at issue etc to get a initial "look and feel" what you are trying to address and how? Thanks

thrix (changes requested):
I am not very happy with the code, it become a lot more hard to read and understand :( Personally I am not sure the approach is great, maybe it would be simpler to:

1. prepare on the control host the list of files to copy (including aliases) per target directory
2. copy in one rsync command ^ to the destination

?

Then we could omit this dance with handling differently the aliases and make it thus more readable and understandable? Plu, we could omit the linking execution on the host.

tmt/steps/execute/__init__.py
happz: Ok, `ShellScript(' && '.join(script.to_element() for script in alias_commands))`

martinhoyer:
> I am not very happy with the code, it become a lot more hard to read and understand :( Personally I am not sure the approach is great, maybe it would be simpler to:
> 
>     1. prepare on the control host the list of files to copy (including aliases) per target directory
> 
>     2. copy in one rsync command ^ to the destination
> 
> 
> ?
> 
> Then we could omit this dance with handling differently the aliases and make it thus more readable and understandable? Plu, we could omit the linking execution on the host.

Agreed, also not happy, but I think it would need a deeper refactoring to make it "not complicated". Tried the suggested "prepare it locally first" approach, does it look better @thrix? I barely know what I'm doing here. Not a one-liner I thought it will be :'(

tmt/steps/execute/__init__.py
```diff
@@ -797,28 +807,105 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
+            for script in self.scripts:
+                if not script.enabled(guest):
+                    continue
 
-        # Make sure scripts directory exists
-        command = Command("mkdir", "-p", f"{guest.scripts_path}")
+            with script as source:
+                # If it's a template, prevent immediate cleanup and track it
+                if isinstance(script, ScriptTemplate):
+                    script._delete_on_exit = False  # Prevent cleanup before push
```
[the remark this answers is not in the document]
martinhoyer: ping @thrix  https://github.com/teemtee/tmt/pull/3669#issuecomment-2839173792

martinhoyer:
@coderabbitai full review

tmt/steps/execute/__init__.py
```diff
@@ -797,28 +807,105 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
+        # Store aliases: key is the target path on guest, value is list of alias names
+        # Aliases are created in the default scripts_dest_dir
+        aliases_to_create: dict[Path, list[str]] = {}
+        # Keep track of ScriptTemplate instances for later cleanup
+        templates_to_cleanup: list[ScriptTemplate] = []
+
+        try:
+            for script in self.scripts:
+                if not script.enabled(guest):
+                    continue
```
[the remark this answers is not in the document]
martinhoyer: If it's not enabled, it skips processing that script. Am I missing something?

tmt/steps/execute/__init__.py
```diff
@@ -801,28 +802,106 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
+            self.debug(f"Pushing staged scripts from '{staging_root}/' to guest '/'")
+            guest.push(
+                source=staging_root,  # Push the whole directory
+                destination=Path('/'),
+                options=[
+                    "-a",  # Archive mode (-rlptgoD), includes preserving permissions
+                    "--links",  # Copy symlinks as symlinks
+                ],
+                superuser=guest.facts.is_superuser is not True,
+            )
```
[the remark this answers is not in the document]
martinhoyer: indeed. I'm trying something in f7bbc5786753739a55dc5408bd2929025b6edbf0, wdyt @happz?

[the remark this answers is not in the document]
happz: I don't know. I don't want to ruin your party, but to me it feels like `push()` is trying to guess now, making assumptions about what the caller might have meant - if the source path is a directory, we must be pushing its content. I would prefer forcing the caller to be clear about their intentions, I'm always worried when lower layers "guess" their users' intentions :/

On the other hand, I guess we rarely want to not push the content if the source path is a directory - so, if that becomes the rule, "this is how `push()` behaves, always and everywhere", then it should be fine.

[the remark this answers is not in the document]
martinhoyer: I'm happy to get the party ruined as I very much dislike this code. I thought about making it an explicit argument, but decided against.  
It might be best to drop it from the release, as it bloated out of proportions and merging it hours before release feels unwise.

I'm going to check the internet for rsync wrappers and best practices. It can't be that "hard" :)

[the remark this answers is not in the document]
happz: My unhappiness is rooted deep in `Common` class, where `workdir` is `Path`, and it can be `Optional`. This is the root of all path evil, all those `assert self.workdir is not None` and so on. This "rsync vs directory content" is just another chapter :/

Anyway, as I said, we probably always want to push content of a directory to remote path, so stating "`push()` always pushes content of a directory" would be a valid position, given it's not a generic `rsync` wrapper but our tmt wrapper, suited for our needs. It would require at least a brief review of `push()` calls, to make sure they don't expect different behavior, and it would be nice to do so in a dedicated patch, focusing on that change and nothing else. But it would be correct.

tmt/steps/execute/__init__.py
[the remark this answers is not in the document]
thrix: yeah, I would swear it was not idented this way, when I looked, now it looks ok (or I got confused)

tmt/steps/execute/__init__.py
```diff
@@ -801,28 +803,108 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
+                    if isinstance(script, ScriptTemplate):
+                        # ScriptTemplate renders to a temporary file in __enter__
+                        rendered_path = script._rendered_script_path
+                        assert rendered_path is not None
+                        shutil.copy2(rendered_path, local_script_path)
+                        # Track rendered templates for cleanup, even though __exit__ handles it
+                        rendered_templates.append(rendered_path)
+                    else:
+                        # Regular script, copy the source file
+                        shutil.copy2(source_path, local_script_path)
```
happz, line 854: How about adding a method to `Script`, something like `render_into(filepath: Path)`? The default implementation would do this `shutil.copy2(SCRIPTS_SRC_DIR / self.source_filename, filepath)` oneliner (`source_path` is the `SCRIPTS_SRC_DIR / self.source_filename`), and implementation in `ScriptTemplate` would use `self._rendered_script_path` instead. Why: the simple task, "store script in a given file", is performed in two ways, depending on which class the `script` is. That screams class-specific implementation is leaking beyond the `Script*` classes.

tmt/steps/provision/__init__.py
```diff
@@ -2541,47 +2545,101 @@ def push(
...
+            final_command_args = [
+                *cmd,
+                *options,
+                "-e",
+                self._ssh_command.to_element(),
+                source_str,
+                remote_target,
+            ]
+
+            self.debug(f"Executing rsync: {' '.join(map(str, final_command_args))}", level=3)
```
happz, line 2595: `_run_guest_command()` and its workhorse `Command.run()` **will** log the command, in its entirety, with all its options: `logger.debug(f'Run command: {self!s}', level=2)`. It is not necessary to log it all before the command itself, and I am very much opposed to any attempts like this (https://github.com/teemtee/tmt/pull/3669#pullrequestreview-2786213904), because `Command.run()` does this so we can have unified logging of "a command is executed" actions rather than seven different ways of presenting commands and their options. If you miss any information in what `Command.run()` provides, let's discuss and extend that method so all commands would benefit from it. It's possible to use the `message` parameter to pass in custom "intro" logging message, e.g. `message='Executing rsync to push to guest'` would do, `Command.run()` will add the actual command right after this command line.

And once removed, `final_command_args` can be dropped as well & its content moved back to `Command(...)` call below...

happz, line 2595: `Command.run()` will log the command, in its entirety, with all its options: `logger.debug(f'Run command: {self!s}', level=2)`. It is not necessary to log it extra; the reason why `Command.run()` does it is to avoid this exact situation where we log "running external command" in seven different ways which is a PITA during investigation. If there is anything missing from the info provided by `Command.run()`, let's discuss and fix that, so other invocations could benefit from it, but let's not pollute logging with duplicate messages.

It's also possible to use `message` parameter to add extra intro-ish header into logs, if you want to record it's not just any `rsync` call, but `rsync` pushing to guest, `message='Executing rsync to push to guest'` would do, and `Command.run()` will record the command and its options as well.

And once removed, `final_command_args` could be dropped and moved back into `Command(...)` call below, as mentioned in https://github.com/teemtee/tmt/pull/3669#pullrequestreview-2786213904.

martinhoyer: Right, my bad. Thanks. https://github.com/teemtee/tmt/pull/3669/commits/c819c326713cd85e07bdc69284d73f2d7e26988b

martinhoyer:
@coderabbitai review

tmt/steps/provision/__init__.py
```diff
@@ -2541,47 +2545,97 @@ def push(
...
-            Run the rsync command
-            """
+            # Check if the provided source exists and is a directory locally
+            # to decide if we sync contents (add slash) or the item itself
+            try:
+                if source.exists() and source.is_dir():
+                    sync_source_contents = True
+                    self.debug(f"Source '{source}' is a directory, syncing contents.")
+                else:
+                    self.debug(f"Source '{source}' is a file or does not exist, syncing item.")
```
psss, line 2555: Shouldn't `sync_source_contents` be set to `False` here?

tmt/steps/provision/__init__.py
```diff
@@ -2541,47 +2545,97 @@ def push(
...
                 ),
                 silent=True,
             )
 
-        # Try to push twice, check for rsync after the first failure
         try:
             rsync()
-        except tmt.utils.RunError:
+            self.debug(
+                f"Successfully pushed '{source_str}' to '{self._ssh_guest}:{dest_str}'", level=1
```
psss, line 2601: I think similar debug messages should go to a deeper debug level. Otherwise we would get too much noise.

tmt/steps/provision/__init__.py
```diff
@@ -2541,47 +2545,97 @@ def push(
...
+                if rsync_check_outcome == CheckRsyncOutcome.ALREADY_INSTALLED:
+                    self.debug(
+                        "rsync already installed, first error likely connection/permission issue."
+                    )
+                    raise first_error
+
+                # If check passed, retry the command
+                self.debug("rsync check passed or installed, retrying command.")
+                rsync()  # Second attempt
+                self.debug(f"Pushed '{source_str}' to '{self._ssh_guest}:{dest_str}' on retry.")
```
psss, line 2623: Similar as above, perhaps `level=2` would be appropriate?

tmt/steps/execute/__init__.py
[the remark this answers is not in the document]
martinhoyer: Tried removing the unhappiness - not sure if it make sense - https://github.com/teemtee/tmt/pull/3744
(it's on top of this branch)

tmt/steps/execute/__init__.py
```diff
@@ -801,28 +824,120 @@ def prepare_tests(self, guest: Guest, logger: tmt.log.Logger) -> list[TestInvoca
...
-                    if script.enabled(guest):
-                        guest.push(
-                            source=source,
-                            destination=script.destination_path or guest.scripts_path / filename,
-                            options=["-p", "--chmod=755"],
-                            superuser=guest.facts.is_superuser is not True,
-                        )
+                    # Track rendered templates for cleanup if needed
+                    if rendered_path is not None:
+                        rendered_templates.append(rendered_path)
```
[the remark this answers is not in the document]
martinhoyer: https://github.com/teemtee/tmt/pull/3669/commits/5f36703d9821cd3e431d62a4a73810b69a09b205

tmt/steps/provision/__init__.py
martinhoyer: Thanks! https://github.com/teemtee/tmt/pull/3669/commits/5f36703d9821cd3e431d62a4a73810b69a09b205

tmt/steps/provision/__init__.py
martinhoyer: https://github.com/teemtee/tmt/pull/3669/commits/5f36703d9821cd3e431d62a4a73810b69a09b205

tmt/steps/provision/__init__.py
martinhoyer: https://github.com/teemtee/tmt/pull/3669/commits/5f36703d9821cd3e431d62a4a73810b69a09b205

tmt/steps/execute/__init__.py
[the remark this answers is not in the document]
happz: Resolving, more to follow in another PR, I guess.

tmt/steps/provision/__init__.py
psss: Seems that `/tests/plan/import/basic` still [fails](https://artifacts.dev.testing-farm.io/909f0713-7927-480f-bd01-15b6a7ab7aa7/) with:

    cannot access local variable 'sync_source_contents' where it is not associated with a value

martinhoyer: looks green now. Adding full test label.

tmt/steps/provision/local.py
```diff
@@ -159,13 +159,16 @@ def reboot(
...
         destination: Optional[Path] = None,
         options: Optional[list[str]] = None,
         superuser: bool = False,
     ) -> None:
         """
-        Nothing to be done to push workdir
+        Copy files or directories locally.
+
+        For local guests, this method copies files or directories to the specified
+        destinationNothing to be done to push workdir
```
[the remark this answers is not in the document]
martinhoyer: yeah, didn't want to commit this file...letting the test fail first.

tmt/steps/provision/__init__.py
```diff
@@ -2507,77 +2564,96 @@ def push(
...
+            full_cmd_list = [
+                *cmd_base,
+                *current_rsync_options,
+                "-e",
+                ssh_command_str,
+                source_str,
+                remote_target,
+            ]
 
+            self.debug(f"Executing rsync: {' '.join(repr(arg) for arg in full_cmd_list)}", level=2)
```
happz, line 2602: Logging is done by `Command.run()` under the hood of `self._run_guest_command()`, and `full_cmd_list` does not need to exist and can be expanded in `Command(...)` below...

tmt/steps/provision/__init__.py
```diff
@@ -2507,77 +2564,96 @@ def push(
...
 
-        def rsync() -> None:
-            """
-            Run the rsync command
-            """
+        try:
+            source_str = self._prepare_rsync_source_path(source)
+            dest_str = self._prepare_rsync_destination_path(destination)
+        except ValueError as path_prep_error:
+            raise GeneralError(f"Failed to prepare rsync paths: {path_prep_error}")
```
happz, line 2578: Please, use `raise ... from path_prep_error` instead.

tmt/steps/provision/__init__.py
```diff
@@ -2499,6 +2499,63 @@ def execute(
...
+        Prepares the destination path string for rsync.
+        """
+        effective_dest_path = destination if destination is not None else Path("/")
+        dest_path_str = str(effective_dest_path)
+
+        if dest_path_str == "/":
+            return "/"
+
+        if not dest_path_str.endswith('/'):
+            return dest_path_str + '/'
```
happz, line 2556: Isn't this the issue we discussed briefly on slack, i.e. every destination path, even if it's a file, will be turned into a directory by getting a trailing `/`? The code below wishes to do a file -> file push, it will fail on this.

```python
            guest.push(
                source=prepare_wrapper_path,
                destination=prepare_wrapper_path,
            )
```

tmt/steps/provision/__init__.py
```diff
@@ -2507,77 +2564,96 @@ def push(
...
+                    f"Initial rsync error: {first_error}\n"
+                    f"Subsequent error during check/retry: {second_error}"
+                )
+                self.fail(error_message)
+                raise second_error from first_error
+            except Exception as check_retry_exception:
                 self.fail(
-                    f"Failed to push workdir to the guest. This usually means "
-                    f"that login as '{self.user}' to the guest does not work."
+                    f"An unexpected error during rsync check or retry: {check_retry_exception}"
```
happz, line 2649: No need for including `check_retry_exception` in the error message, the exception is reraised on the next line, it will be recorded.

tmt/steps/provision/__init__.py
```diff
@@ -2507,77 +2564,96 @@ def push(
...
                 )
                 raise
+        except Exception as initial_unexpected_err:
+            self.fail(
+                f"An unexpected error during the initial rsync attempt: {initial_unexpected_err}"
```
happz, line 2654: No need for including `check_retry_exception` in the error message, the exception is reraised on the next line, it will be recorded.

Issue #3652, opened 2025-04-08
Plans with both `execute` and `plan.import` are accepted, causing duplication


happz:
```yaml
execute:
    how: tmt

/binutils:
    /common:
        plan:
            import:
                url: ...
                ref: main
                name: /plans/build-gating/common

    /kernel-rebuild:
        plan:
            import:
                url: ...
                ref: main
                name: /plans/build-gating/kernel-rebuild
```

tmt will accept this fmf input, duplicating `/binutils/common` and `/binutils/kernel-rebuild`: one instance will come from being "local" plans, aka having `execute` key, the second instance because of being "remote reference" aka having `plan.import` keys. tmt should at least emit a warning for this, if not even reject it.

thrix:
Aha, I see, the `execute` phase is inherited, nm

PR #3675, opened 2025-04-16
Plans should not be both local and importing remote plans

happz:
If such a plan is detected, tmt emits a warning and skips the "local" aspect of the given plan.

Fixes https://github.com/teemtee/tmt/issues/3652.

Pull Request Checklist

* [x] implement the feature


therazix (approved):

tmt/base.py
```diff
@@ -3564,6 +3564,20 @@ def plans(
             sources = None
 
         # Build the list, convert to objects, sort and filter
```
thrix, line 3566: @happz this comment should ideally move also to line #3580?

thrix (approved):
LGTM

tmt/base.py
happz: I wouldn't think so, it describes the code below - collect plans, convert them to `Plan` instances, filter and so on.

thrix: ok, as you wish.

psss (approved):

PR #3644, opened 2025-04-03
Deduplicate the `report` step documentation

mcasquer:
Deduplicates the `report` step documentation, which includes the following:
* Moving all the information from the specs to the plugins section
* Updates the header in the plugins section
* Remove completely (whenever is possible) the references in the specs
* Fixes some references to the removed documentation
* Directly refer the reader to the `report plugins` and add `implemented-by` link.

Pull Request Checklist

* [x] update the specification
* [x] adjust plugin docstring

PR #3677, opened 2025-04-16
Fix invalid `link` of `/spec/plans/report/file`

happz:
Introduced in https://github.com/teemtee/tmt/pull/3644

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/spec/plans/report.fmf
+++ b/spec/plans/report.fmf
@@ -68,7 +68,7 @@ link:
                         log:
                           - LOG_PATH
     # Not implemented yet, do not inherit link from parent
-    link:
+    link: []
 
 /when:
     summary: Conditional step configuration


psss:
Isn't it expected that `None` would be normalized into an empty list?

happz:
> Isn't it expected that `None` would be normalized into an empty list?

Hmm, good point, that did not occur to me. I will look into that, I already see `links` is internally `Optional[Links]`, there's no point in being `Optional`...

happz:
> > Isn't it expected that `None` would be normalized into an empty list?
> 
> Hmm, good point, that did not occur to me. I will look into that, I already see `links` is internally `Optional[Links]`, there's no point in being `Optional`...

Filed https://github.com/teemtee/tmt/issues/3679 to track the follow-up work.

PR #3678, opened 2025-04-16
Fix logging of early import errors

happz:
In some conditions, exceptions raised while importing plugins could have been reported as just the exception message, without any traceback. This was caused by the necessary loggers being defined in `tmt.cli` package, which must be imported correctly first for the loggers to be available. But, `tmt.cli` runs plugin discovery, and if for whatever reason import of a plugin fails, subsequent call to `show_exception()` would also raise an exception, because it would try to import unavailable `tmt.cli.EXCEPTION_LOGGER`, resulting in the most simple error logging implemented.

By moving bootstrap and exception loggers into their own module, both plugin discovery spawned from `tmt.cli` and the error reporting code can reach loggers they need, as their existence no longer depends on `tmt.cli` and plugins being imported successfully.

Make a small typo in `tmt/config/themes/default.yaml`, e.g. change `restructuredtext-literal` to `restructuredtext-literal-`. `tmt lint` will crash with a single line or output:

```
Failed to import the 'tmt.steps.discover.fmf' module from '/home/happz/git/tmt'.
```

With this patch, one gets full chain of errors:

```
$ tmt lint

Failed to import the 'tmt.steps.discover.fmf' module from '/home/happz/git/tmt'.

Cause number 1:

    Invalid theme configuration.

    Cause number 1:

        1 validation error for Theme
        restructuredtext-admonition-warning-
          extra fields not permitted (type=value_error.extra)
```

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1094,7 +1094,7 @@ def test_import_member(root_logger):
 
 def test_import_member_no_such_module(root_logger):
     with pytest.raises(
-        SystemExit,
+        tmt.utils.GeneralError,
         match=rf"Failed to import the 'tmt\.steps\.nope_does_not_exist'"
         rf" module from '{Path.cwd()}'.",
     ):
--- a/tmt/_bootstrap.py
+++ b/tmt/_bootstrap.py
@@ -1,0 +1,36 @@
+"""
+Loggers for use during tmt startup.
+
+Very special loggers used by tmt plugin discovery and configuration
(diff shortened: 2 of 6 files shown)


therazix (approved):

psss (approved):
Nice!

psss:
Just two irrelevant failures, the prepare freeze timeout appeared repeatedly, proposing to extend the outer timeout a bit in #3762.

PR #3680, opened 2025-04-17
Move the `report.yaml` spec into deferred stories

psss:
Let's move the definition out of the main specification as it's outdated, not implemented and thus might be confusing for users. We can revisit that later when/if needed.

Pull Request Checklist


therazix (approved):

happz (approved):

happz:
Specs-only, merging.

PR #3681, opened 2025-04-17
Add `awk` as an essential requirement of tmt `execute` plugin

happz:
`awk` is used several times in scripts the plugin installs and offers to users, namely `tmt-reboot`. It seems that `awk` is no longer available by default in Fedora 42 containers, so we need to be explicit about its necessity.

Pull Request Checklist

* [x] implement the feature


therazix (approved):

martinhoyer (approved):
Might be worth getting rid of it at some point

happz:
Skipping virtual checks, there's a queue in Beaker :/

PR #3686, opened 2025-04-21
Remove the duplication leftovers from the specs

mcasquer:
Removes completely the duplicated information from the following specs:
* discover (shell)
* provision (openstack)
* execute (upgrade & tmt)

Also updates a `tmt` spec mention, so now points to the proper plugin

Pull Request Checklist

### DIFF
--- a/spec/plans/discover.fmf
+++ b/spec/plans/discover.fmf
@@ -39,54 +39,6 @@ description: |
           description: Long test description.
           ...
 
-/shell:
-    summary: Provide a manual list of shell test cases
-    description: |
-        List of test cases to be executed can be defined manually
-        directly in the plan as a list of dictionaries containing
-        test ``name`` and actual ``test`` script. Optionally it is
-        possible to define any other :ref:`/spec/tests` attributes
-        such as ``path`` or ``duration`` here as well. The default
-        :ref:`/spec/tests/duration` for tests defined directly in
-        the discover step is ``1h``.
-
-        It is possible to fetch code from a remote git repository
-        using ``url``. In that case repository is cloned and all
-        paths are relative to the remote git root. Using remote
-        repo and local test code at the same time is not possible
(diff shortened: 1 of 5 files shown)


spec/plans/discover.fmf
```diff
@@ -39,54 +39,6 @@ description: |
...
-            url: https://github.com/teemtee/tmt
-            tests:
-              - name: Use tests/full/test.sh from the remote repo
-                path: /tests/full
-                test: ./test.sh
-
-    link:
-      - implemented-by: /tmt/steps/discover/shell.py
-
 /fmf:
```
psss, line 42: The `fmf` section should go away as well, right?

psss (changes requested):

spec/plans/discover.fmf
mcasquer: @psss that's right, fmf removed too :)

skycastlelily (approved):
Looks like [spec/plans/main.fmf](https://github.com/teemtee/tmt/blob/main/spec/plans/main.fmf#L15 ) also need to be modified, otherwise LGTM

martinhoyer (approved):
image
It's beautiful

therazix (approved):

mcasquer:
> Looks like [spec/plans/main.fmf](https://github.com/teemtee/tmt/blob/main/spec/plans/main.fmf#L15) also need to be modified, otherwise LGTM

We've added references to the plugins for each spec so IMO it's not incorrect still referencing the plans, the user should be able to access the plugins docs for more info

skycastlelily:
> 
> We've added references to the plugins for each spec so IMO it's not incorrect still referencing the plans, the user should be able to access the plugins docs for more info

hmm, sure, I don't have a strong opinion on this:)

happz:
Specs-only, merging.

PR #3687, opened 2025-04-22
Disable the `virt-customize` speed-up due to DNS issues

psss:
As a hotfix, let's use `fedora-41` instead of `fedora-42` to work around the weird dns issues and unblock the release.

For reference the error was:

```
>>> Curl error (6): Could not resolve hostname for https://mirrors.fedoraproject
>>> Librepo error: Cannot prepare internal mirrorlist: Curl error (6): Could notFailed to download metadata (metalink: "https://mirrors.fedoraproject.org/metalink?repo=updates-released-f41&arch=x86_64") for repository "updates"
Librepo error: Cannot prepare internal mirrorlist: Curl error (6): Could not resolve hostname for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f41&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
virt-customize: error: dnf --refresh install -y beakerlib: command exited with an error
```

PR #3690, opened 2025-04-22
Workaround virt-customize issues in CI

thrix:
The problem seems to be DNS, let's try to avoid `systemd-resolved` by setting nameserver to google DNS.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


thrix:
so does not seems to be DNS :( needs more debugging, keeping open, we will go with #3687

PR #3695, opened 2025-04-24
CRB Feature Prepare Plugin

### LINKED ISSUE #3650: Implement `crb` prepare for the `feature` plugin
Add support for easily enabling the [CodeReady Linux Builder](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository#CodeReadyLinuxBuilder-repository) repository.
This should work for both `centos-stream` and `rhel`.

martinhoyer:
Closes #3650 

This implements `crb` prepare feature, similarly to the [epel](https://tmt.readthedocs.io/en/stable/plugins/prepare-feature.html)

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] modify the json schema
* [x] mention the version
* [x] include a release note


tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,99 @@
...
+        command_str = f"curl -sS {UPSTREAM_SCRIPT_URL} | {DEFAULT_SHELL} -s -- {action}"
+
+        try:
+            # Execute the command via shell. RunError is raised on failure.
+            guest.execute(Command(DEFAULT_SHELL, '-c', command_str))
+            logger.info(f"Successfully executed {action} using upstream script.")
+
+        except RunError as error:
+            # Log the original error details for better debugging
+            logger.debug(f"RunError details: {error}")
```
happz, line 88: `guest.execute()` will log the error. It will log the whole output of the command executed, exit code, etc.

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,99 @@
...
+
+        logger.info(f"{action.capitalize()} CRB repository using upstream script.")
+
+        # Command to download and execute the script, passing the action (enable/disable)
+        # sh -s -- action: passes 'action' as an argument ($1) to the script executed by sh
+        command_str = f"curl -sS {UPSTREAM_SCRIPT_URL} | {DEFAULT_SHELL} -s -- {action}"
+
+        try:
+            # Execute the command via shell. RunError is raised on failure.
+            guest.execute(Command(DEFAULT_SHELL, '-c', command_str))
```
happz, line 83: Make `command_str` `command: ShellScript(...)` and you can pass it to `guest.execute()` directly. `guest.execute()` will take care of adding `DEFAULT_SHELL` if needed.

martinhoyer: Thanks. I know I've been asking already, but how can I tell when ShellScript is more appropriate than Command? b5108a63b5ff9c0d392a34a83642601cb2e3e86b

happz: I'd say it's often a personal preference, both have their pros and cons. The good thing is they both have methods to convert one to the other, and all our `execute()` or `run()` methods should be able to handle both.

```python
# Both are fine
Command('git', 'rebase', '-i', 'HEAD~2')
ShellScript('git rebase -i HEAD~2')
```

`ShellScript` might be better for more complicated commands, it's nicer compound commands with pipes and stuff, and it accepts docstrings.

In this case, it's the use of `DEFAULT_SHELL` and `-c` - it's correct, it will work, no problem there, it's just more verbose than necessary because 1. `Guest.execute()` accepts shell scripts, and 2. `f"curl -sS {UPSTREAM_SCRIPT_URL} | {DEFAULT_SHELL} -s -- {action}"` certainly looks like one. That points *me* towards using `ShellScript` right away.

martinhoyer: Thank You :pray:

tmt/steps/prepare/feature/crb.py
martinhoyer: ping @happz

lukaszachy:
Could we bundle the `https://src.fedoraproject.org/rpms/epel-release/raw/epel10/f/crb` into tmt package?

Using existing script is a great idea, but there is risk that src.fedora will not be reachable (outage, restricted network access) and this feature will not work. Another risk is in fetching latest script, which might get broken by accidental update.

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,91 @@
...
+        """Enable or disable the repository using the upstream script"""
+        cls._check_distro(guest, logger)
+
+        logger.info(f"{action.capitalize()} CRB repository using upstream script.")
+
+        # Command to download and execute the script, passing the action (enable/disable)
+        # sh -s -- action: passes 'action' as an argument ($1) to the script executed by sh
+        guest.execute(
+            ShellScript(
+                f"curl -sS {UPSTREAM_SCRIPT_URL} | FORCE_DNF=1 {DEFAULT_SHELL} -s -- {action}"
```
thrix, line 81: Would it make sense to fail here more nicely if the `UPSTREAM_SCRIPT_URL` would become unaccessible?
i.e. add a check before that the script is available?
Or at least add `tmt.utils.DEFAULT_SHELL_OPTIONS` to use `set -o pipefail`, so it does not fail weirdly and we know it failed on `curl` and it would stop piping it into the shell??

martinhoyer: sounds good to me. There can also be a fallback to checking of crb command is available, installing the epel-latest rpm if not, etc. Don't know where is the line for this, but it being jank by design by not having a standardized identifier for the repo I wouldn't overengineer it?

maybe we can use https://dl.fedoraproject.org/pub/epel/ to download the rpm and only extract the crb script from it, I presume it has a good CDN....overthinking it

martinhoyer:
> Could we bundle the `https://src.fedoraproject.org/rpms/epel-release/raw/epel10/f/crb` into tmt package?
> 
> Using existing script is a great idea, but there is risk that src.fedora will not be reachable (outage, restricted network access) and this feature will not work. Another risk is in fetching latest script, which might get broken by accidental update.

Whatever is the consensus. I'm personally against vendoring it.

tmt/steps/prepare/feature/crb.py
thrix: from discussion: vendor it and ideally test it in all combinations (via containers?)

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,87 @@
...
+
+        logger.info(f"{action.capitalize()} CRB repository using upstream script.")
+
+        # Inspired by crb executable from https://src.fedoraproject.org/rpms/epel-release
+        guest.execute(
+            ShellScript(
+                f"sudo dnf config-manager --{action} $(dnf repolist --all |"
+                f" grep -i -e crb -e powertools -e codeready |"
+                f" grep -v -i -e debug -e source -e eus -e virt -e rhui |"
+                f" sed 's/^\\s*\\([^ ]*\\).*/\1/')"  # do not use awk
```
happz, line 77: I'm afraid this is a pet peeve of mine, I just don't like these multiline-but-not-really strings :/ Why not use a true multiline string? `ShellScript` doesn't care, every line wouldn't be polluted by starting with `f"`, I wouldn't need to think for a while why there are no trailing commas, it wouldn't look so bad, and between `"""` and `"""` we would have just what belongs there. Every single line now needs special Python controls to make it part of a larger string, `"""` does that out of the box. I'm sorry, IMO it's harming readability :( Also, a `r`aw string could help a bit with backslashes, and moving the closing `)` on its own line would make it visible - the first time I missed it completely and treated it as part of the `sed` expression.

```python
ShellScript(
    rf"""
    sudo dnf config-manager --{action} $(dnf repolist --all | \
      grep -i -e crb -e powertools -e codeready | \
      grep -v -i -e debug -e source -e eus -e virt -e rhui | \
      sed 's/^\\s*\\([^ ]*\\).*/\\1/' \
    )
    """
)
```

martinhoyer: funnily enough I had the exact opposite comment in https://github.com/teemtee/tmt/pull/3715#pullrequestreview-2817315935, where the multiline formatting also broke things due to not being de-dented.  
Happy to come up with some project-wide consensus (I think I've asked for one already in comments or chat).  
To me for example is this more readable than your multiline, leaving de-denting aside. But I'm hardly an authority on the matter. Internet and PEP8 is not super helpful, but I'd say it's leaning towards multiple f lines in parentheses.

happz: Put it for discussion at the Tuesday meeting, it would be nice to have official guidance. The community should decide which style is better, the beautiful one I am proposing, or the ugly, terrible one you prefer :)

> But I'm hardly an authority on the matter. Internet and PEP8 is not super helpful, but I'd say it's leaning towards multiple f lines in parentheses.

PEP8 states how priorities should be read, and https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds is probably my favorite part of the documentation of all formatting and style guides everywhere. We get some ideas, but eventually, it need to be acceptable to us. We will never please every single external contributor, hence picking one way, sticking to it, and the consistency wins even when it contradicts PEP8's proposals.

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,85 @@
...
+    .. code-block:: yaml
+
+        prepare:
+            how: feature
+            crb: enabled
+
+    .. code-block:: shell
+
+        prepare --how feature --crb enabled
+    """
```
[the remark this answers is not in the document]
happz: `.. versionadded:: 1.49` is the right form, e.g. https://github.com/teemtee/tmt/blob/main/tmt/checks/avc.py#L391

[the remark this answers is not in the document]
martinhoyer: It's already there, I thought graphite will auto-resolve this :/

tmt/steps/prepare/feature/crb.py
martinhoyer: I wonder what @LecrisUT would suggest :)

happz: Added to the hacking agenda.

LecrisUT: Beetleguised :P. My vote is to always pipe it through `textwrap.dedent` and always keep (and use) new lines.

For this one, it would be nice to split it and assign some random variables for readability, e.g.
```shell
crb_repos=$(dnf repolist --all |grep -i -e crb -e powertools -e codeready)
...
```

tests/prepare/feature/crb/data/plans.fmf
```diff
@@ -0,0 +1,36 @@
...
+        ! dnf repolist --enabled | grep -i -E 'crb|codeready'
+        # Verify it is in the disabled list
+        dnf repolist --disabled | grep -i -E 'crb|codeready'
+
+    environment+:
+      STATE: disabled
+
+  /crb_package:
+    prepare+:
+      - crb: enabled
```
psss, line 32:
```suggestion
        crb: enabled
```

The [list/dict merging](https://fmf.readthedocs.io/en/stable/features.html#merging) works as well, but let's not confuse the reader.

tests/prepare/feature/crb/data/plans.fmf
```diff
@@ -0,0 +1,36 @@
...
+    execute:
+      script: |
+        dnf install -y sgx-rpm-macros
+        rpm -q sgx-rpm-macros
```
psss, line 36: The setup is identical with the `/crb/enabled` plan above. What about movig these two command there? We would save a bit of time by provisioning/preparing just two containers instead of three.

tests/prepare/feature/crb/main.fmf
```diff
@@ -0,0 +1,4 @@
+summary: Common feature to enable/disable CRB
+description: Make sure enable/disable CRB works.
+tag+:
+  - provision-container
```
psss, line 4:
```suggestion
  - provision-container
  - provision-only
```

We don't need to run the test as part of non-provision plans. Without `provision-only` it would be also executed under `/plans/features/steps/prepare`, not only under `/plans/provision/container/the-rest` as expected.

tests/prepare/feature/crb/test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    images="$TEST_IMAGE_PREFIX/centos/stream9/upstream:latest $TEST_IMAGE_PREFIX/ubi/8/upstream:latest ubi9"
+
+    # CRB
+    for image in $images; do
+        # Run the '/crb/enabled' plan, overriding the provision image.
+        # The plan itself contains the check (dnf repolist enabled | grep ...).
+        # Expecting tmt run to succeed (exit code 0) as the check should pass.
+        rlPhaseStartTest "Test /crb/enabled on $image"
+            rlRun "tmt run --all provision --how container --image $image prepare execute plan --name /crb/enabled" 0 "Run /crb/enabled plan for $image"
```
psss, line 23:
```suggestion
            rlRun "tmt run --all provision --how container --image $image plan --name /crb/enabled" 0 "Run /crb/enabled plan for $image"
```

When `--all` is used we don't need to explicitly list `prepare` and `execute` steps.

tests/prepare/feature/crb/test.sh
```diff
@@ -0,0 +1,45 @@
...
+        rlPhaseEnd
+
+        # Run the '/crb/disabled' plan, overriding the provision image.
+        # The plan itself contains the checks (! dnf repolist enabled ... && dnf repolist disabled ...).
+        # Expecting tmt run to succeed (exit code 0) as the checks should pass.
+        rlPhaseStartTest "Test /crb/disabled on $image"
+            rlRun "tmt run --all provision --how container --image $image prepare execute plan --name /crb/disabled" 0 "Run /crb/disabled plan for $image"
```
psss, line 30: Similar here.

tests/prepare/feature/crb/test.sh
```diff
@@ -0,0 +1,45 @@
...
+                # This plan enables CRB and tries to install a package from it.
+                rlRun "tmt run --all provision --how container --image $image prepare execute plan --name /crb/crb_package" 0 "Run /crb/crb_package plan for $image"
+            rlPhaseEnd
+        fi
+    done
+
+    rlPhaseStartCleanup
+        rlRun "popd"
+    rlPhaseEnd
+rlJournalEnd
```
psss, line 45: Tried to run the test and it failed with the following:

```
        Command 'sudo dnf config-manager --disable $(dnf repolist --all | grep -i -e crb -e powertools -e codeready | grep -v -i -e debug -e source -e eus -e virt -e rhui | sed 's/^\s*\([^ ]*\).*//')' returned 127.

        # stderr (1/1 lines)
        /bin/bash: line 1: sudo: command not found
```

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+
+        guest.package_manager.install(Package("dnf-command(config-manager)"))
+
+    @classmethod
+    def _manage_repo(cls, guest: Guest, logger: tmt.log.Logger, action: str) -> None:
+        """Enable or disable the repository using the upstream script"""
+        cls._check_distro(guest, logger)
+
+        logger.info(f"{action.capitalize()} CRB repository using upstream script.")
```
psss, line 69:
```suggestion
        logger.info(f"{action.capitalize()} CRB repository using upstream script.")
```

Shall the `upstream script` be removed from the message as it's not downloaded anymore?

psss (changes requested):
Thanks for implementing this! Overall looks good to me. Added a couple comments.

tests/prepare/feature/crb/data/plans.fmf
psss: Ah, I've noticed this is relevant only for selected images, right? So feel free to ignore this.

tests/prepare/feature/crb/data/plans.fmf
martinhoyer: Should there be variety in the tests?

tests/prepare/feature/crb/test.sh
martinhoyer, line 30:
```suggestion
            rlRun "tmt run --all provision --how container --image $image plan --name /crb/disabled" 0 "Run /crb/disabled plan for $image"
```

tests/prepare/feature/crb/test.sh
martinhoyer: removed

tmt/steps/prepare/feature/crb.py
martinhoyer: thanks, removed

tests/prepare/feature/crb/test.sh
```diff
@@ -0,0 +1,45 @@
...
+    rlPhaseStartSetup
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
+
+        build_container_image "centos/stream9/upstream\:latest"
+        build_container_image "ubi/8/upstream\:latest"
+
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    images="$TEST_IMAGE_PREFIX/centos/stream9/upstream:latest $TEST_IMAGE_PREFIX/ubi/8/upstream:latest ubi9"
```
LecrisUT, line 15: How about a test case for fedora as well to check that it is either skipped or failing non-silently.

tests/prepare/feature/crb/test.sh
psss: What about rather adding `sudo` only if the user is not `root`? For the images we expect that either user is root or a password-less `sudo` is configured. It would be nice to cover the seconds scenario as well.

tests/prepare/feature/crb/test.sh
martinhoyer: that's a good point, but can we talk about how EPEL is ok with not having it @psss

tests/prepare/feature/crb/test.sh
martinhoyer: this was updated

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,87 @@
...
+        prepare:
+            how: feature
+            crb: enabled
+
+    .. code-block:: shell
+
+        prepare --how feature --crb enabled
+
+    .. versionadded:: 1.49
+    """
```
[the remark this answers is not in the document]
martinhoyer: wow

tests/prepare/feature/crb/test.sh
psss: Thanks!

tests/prepare/feature/crb/test.sh
martinhoyer: done

martinhoyer: or maybe not, there is this image prefix thing

martinhoyer: I see. Amazing.

tmt/steps/prepare/feature/crb.py
```diff
@@ -0,0 +1,84 @@
...
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def _check_distro(cls, guest: Guest, logger: tmt.log.Logger) -> None:
+        """Verify the guest distribution is supported and install config-manager if needed"""
+        if not (
+            guest.facts.distro
+            and any(pattern.match(guest.facts.distro) for pattern in SUPPORTED_DISTRO_PATTERNS)
+        ):
+            logger.warning('CRB prepare feature is supported on RHEL/CentOS-Stream 8, 9 or 10.')
```
happz, line 57: Emits a warning, but then proceeds and tries to make changes to guest that is reportedly not compatible. Since `_check_distro()` is called just once, would it make sense for its code to be merged into `_manage_repo()`, and `warning()` followed by `return`?

psss: Or return `bool` denoting if the check was successful and use it for the `return` decision:

```python
if not cls._check_distro(guest, logger):
    return
```

Incorporating sounds also ok, it's not that large.

psss: Should be covered by cb01e7ab.

psss (approved):
Adjusted the unsupported distro handling a bit in cb01e7ab and added a release note in 72bbd51c. Should be ready for merge.

LecrisUT (approved):

PR #3701, opened 2025-04-28
Refactor the `DiscoverFmf.go()` method

### LINKED ISSUE #2448: Refactor the `DiscoverFmf.go()` method
It has grown just too much (almost 400 lines) and it's very hard to read it. Let's split it into multiple methods.

bajertom:
Pull Request Checklist

Would it be also worth it to refactor the part between these two helper methods?  

* [x] implement the feature

Checking for `URL`  in the `tmt/steps/discover/fmf.py` was dropped completely and information about missing git repo has been moved into `tmt/steps/discover/__init__.py` instead. When `--fmf-id` is used in a directory without git root and there is no `URL` fed to `tmt run discover --how fmf` then the user gets notified about this.

During testing I've found out that two tests are identical in the `tests/discover/filtering.sh`, specifically the 
` # 2: w/o url in plan AND w/ url in plan: w/o url in CLI - w/ url in CLI` and `# 2: w/ url in plan AND w/o url in plan: w/o url in CLI - w/ url in CLI` scenarios.
They both use two plans, one with and one without `URL`, then calling `tmt run discover --how fmf --fmf-id` either with or without `--url` on a command line. 
The difference is that in first scenario the non-url-plan is called `a-non-url.fmf` and in the second scenario it it called `z-non-url.fmf`. The result is the same, so one of the test was dropped. (I think the original idea was about an order in which they were supposed to be discovered, first with url followed with non-url, and in the the other scenario vice-versa).

Changed `self.step.plan.node.root` into `fmf_root` and added the `if else` statement into  
`fmf_root = Path(self.step.plan.fmf_root) if self.step.plan.fmf_root else None` in `tmt/steps/discover/fmf.py`.
The reason is because `node.root` is not well annotated and `fmf_root` will give the same results. Adding the existence check is due to the fact that `fmf_root` can be None and tmt would fail with ugly exception message `argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'.`

Fixes #2448

### DIFF
--- a/tests/discover/filtering.sh
+++ b/tests/discover/filtering.sh
@@ -128,7 +128,7 @@ rlJournalStart
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id finish 2>&1 \
                2>&1 >/dev/null | tee output" 2
         rlAssertGrep "\`tmt run discover --fmf-id\` without \`url\` option \
-in plan \`/plans/example\` can be used only within git repo." output
+in plan \`/plans/example\` or on command line can be used only within git repo." output
 
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id \
                --url https://github.com/teemtee/fmf finish 2>&1 >/dev/null | tee output" 0
@@ -156,25 +156,7 @@ in plan \`/plans/example\` can be used only within git repo." output
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id finish 2>&1 >/dev/null \
                | tee output" 2
         rlAssertGrep "\`tmt run discover --fmf-id\` without \`url\` option \
-in plan \`/plans/a-non-url\` can be used only within git repo." output
-
-        rlRun "tmt run -rdvvv discover -h fmf --fmf-id \
-               --url https://github.com/teemtee/fmf finish 2>&1 >/dev/null | tee output" 0
-        rlRun "rm -rf $tmp_dir1 $tmp_dir2"
-
-        # 2: w/ url in plan AND w/o url in plan: w/o url in CLI - w/ url in CLI
-        tmp_dir1="$(mktemp -d)"
-        tmp_dir2="$(mktemp -d)"
-        rlRun "cd $tmp_dir1"
-        rlRun "tmt init --template full"
-        rlRun "cd $tmp_dir2"
-        rlRun "tmt init --template base"
-        rlRun "cp plans/example.fmf $tmp_dir1/plans/z-non-url.fmf"
-        rlRun "cd $tmp_dir1"
-        rlRun "tmt run -rdvvv discover -h fmf --fmf-id finish 2>&1 >/dev/null \
-               | tee output" 2
-        rlAssertGrep "\`tmt run discover --fmf-id\` without \`url\` option \
-in plan \`/plans/z-non-url\` can be used only within git repo." output
+in plan \`/plans/a-non-url\` or on command line can be used only within git repo." output
 
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id \
                --url https://github.com/teemtee/fmf finish 2>&1 >/dev/null | tee output" 0
--- a/tests/discover/filtering.sh
+++ b/tests/discover/filtering.sh
@@ -128,7 +128,7 @@ rlJournalStart
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id finish 2>&1 \
                2>&1 >/dev/null | tee output" 2
         rlAssertGrep "\`tmt run discover --fmf-id\` without \`url\` option \
-in plan \`/plans/example\` or on command line can be used only within git repo." output
+in plan \`/plans/example\` can be used only within git repo." output
 
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id \
                --url https://github.com/teemtee/fmf finish 2>&1 >/dev/null | tee output" 0
@@ -156,7 +156,7 @@ in plan \`/plans/example\` or on command line can be used only within git repo."
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id finish 2>&1 >/dev/null \
                | tee output" 2
         rlAssertGrep "\`tmt run discover --fmf-id\` without \`url\` option \
-in plan \`/plans/a-non-url\` or on command line can be used only within git repo." output
+in plan \`/plans/a-non-url\` can be used only within git repo." output
 
         rlRun "tmt run -rdvvv discover -h fmf --fmf-id \
                --url https://github.com/teemtee/fmf finish 2>&1 >/dev/null | tee output" 0
--- a/tmt/steps/discover/__init__.py
+++ b/tmt/steps/discover/__init__.py
@@ -463,11 +463,10 @@ class Discover(tmt.steps.Step):
                 for test_origin in self.tests(enabled=True):
                     fmf_id = test_origin.test.fmf_id
 
-                    if not (fmf_id.url or self.opt('url')):
+                    if not fmf_id.url:
                         raise tmt.utils.DiscoverError(
                             f"`tmt run discover --fmf-id` without `url` option "
-                            f"in plan `{self.plan}` or on command line can be used only "
-                            f"within git repo."
+                            f"in plan `{self.plan}` can be used only within git repo."
                         )
 
                     exported = test_origin.test.fmf_id.to_minimal_spec()
--- a/tmt/steps/discover/__init__.py
+++ b/tmt/steps/discover/__init__.py
@@ -476,6 +476,7 @@ class Discover(tmt.steps.Step):
                         exported.pop('ref')
 
                     export_fmf_ids.append(tmt.utils.dict_to_yaml(exported, start=True))
+
                 click.echo(''.join(export_fmf_ids), nl=False)
             return
 
--- a/tmt/steps/discover/__init__.py
+++ b/tmt/steps/discover/__init__.py
@@ -463,8 +463,12 @@ class Discover(tmt.steps.Step):
                 for test_origin in self.tests(enabled=True):
                     fmf_id = test_origin.test.fmf_id
 
-                    if not fmf_id.url:
-                        continue
+                    if not (fmf_id.url or self.opt('url')):
+                        raise tmt.utils.DiscoverError(
+                            f"`tmt run discover --fmf-id` without `url` option "
+                            f"in plan `{self.plan}` or on command line can be used only "
+                            f"within git repo."
+                        )
 
                     exported = test_origin.test.fmf_id.to_minimal_spec()
 
@@ -472,7 +476,6 @@ class Discover(tmt.steps.Step):
                         exported.pop('ref')
 
                     export_fmf_ids.append(tmt.utils.dict_to_yaml(exported, start=True))
-
                 click.echo(''.join(export_fmf_ids), nl=False)
             return
 
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -552,7 +552,7 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
             if path is not None:
                 fmf_root: Optional[Path] = path
             else:
-                fmf_root = Path(self.step.plan.node.root) if self.step.plan.node.root else None
+                fmf_root = Path(self.step.plan.fmf_root) if self.step.plan.fmf_root else None
             requires_git = self.opt('sync-repo') or any(
                 self.get(opt) for opt in self._REQUIRES_GIT
             )
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -2,7 +2,6 @@ import contextlib
 import glob
 import re
 import shutil
-from contextlib import suppress
 from typing import Any, Optional, cast
 
 import fmf
@@ -531,11 +530,6 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
                 "Cannot manipulate with dist-git without the `--dist-git-merge` option."
             )
 
-        # Raise an exception if --fmf-id is used w/o url and git root
-        # doesn't exist for discovered plan
-        if self.opt('fmf_id'):
-            self.validate_fmf_id_requirements(url)
-
         self.log_import_plan_details()
 
         # Clone provided git repository (if url given) with disabled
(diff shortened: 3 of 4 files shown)


tmt/steps/discover/fmf.py
```diff
@@ -571,38 +534,95 @@ def assert_git_url(plan_name: Optional[str] = None) -> None:
...
+
         # Discover tests
         self.do_the_discovery(path)
 
+    def validate_fmf_id_requirements(self, url: Optional[str] = None) -> None:
+        """
+        Validate environment requirements for the --fmf-id option.
+        """
+
+        def assert_git_url(plan_name: Optional[str] = None) -> None:
```
martinhoyer, line 550: I suspect this should be in `tmt/utils/git.py`?

bajertom: Moved the `assert_git_url` to `tmt/utils/git.py`.

bajertom:
I've found a typo in `tests/discover/adjust-tests.sh` so I've fixed it in this PR if that's okay.
Also `mypy` found unused import of `subprocess` in `steps/discover/fmf.py` and removed it.

tmt/utils/git.py
```diff
@@ -140,6 +140,25 @@ def run(command: Command) -> str:
         )
 
 
+def assert_git_url(plan_name: Optional[str] = None) -> None:
+    """
+    Assert that we're in a Git repository for fmf-id purposes.
+    """
+    try:
+        subprocess.run(
```
happz, line 148: Good, and since this piece of code was touched, please, convert it to use `tmt.utils.Command` or `tmt.utils.ShellScript`. They provide unified logging.

tmt/utils/git.py
```diff
@@ -140,6 +140,25 @@ def run(command: Command) -> str:
...
+            ['git', 'rev-parse', '--show-toplevel'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        raise tmt.utils.DiscoverError(
+            f"`tmt run discover --fmf-id` without `url` option in "
+            f"plan `{plan_name}` can be used only within"
```
happz, line 157: And if `plan_name` is `None`, the message would say "... option in plan `None` can be used..." which is not very nice.

AFAICT, the `Optional` is there because of `plan_name = attr.name` above, can we either test `plan_name` there, or use `assert plan_name is not None`? What do we expect `plan_name` to be there, can it really be `None`, or is just because linters assume it *might* be `None` because it comes from some `dict().get()`-like operation which may return `None`? If we know the plan name must exist at that point, we should use `assert` & drop `Optional`.

tmt/utils/git.py
```diff
@@ -140,6 +140,25 @@ def run(command: Command) -> str:
         )
 
 
+def assert_git_url(plan_name: Optional[str] = None) -> None:
```
happz, line 143: Weird naming - I think it's the move without reshaping the code to fit its new position. Now it lives in `tmt.utils.git`, and speaks about "git URL", but is there any URL involved? And the message below speaks about `tmt run discover`, so... Can I use it to assert something about git URL, from other steps, for example?

I'd propose changing the name and functionality a bit - the goal seems to be testing whether CWD is a git repo, the rest is very much `DiscoverFmf` wrapping & messaging. So, how about something `in_git` or `assert_git` or something along those lines? Run the command, and return `True` if in a git repo, `False` otherwise, and let the messaging to the caller.

tmt/steps/discover/fmf.py
```diff
@@ -571,38 +533,77 @@ def assert_git_url(plan_name: Optional[str] = None) -> None:
...
+                    if not plan_url:
+                        tmt.utils.git.assert_git_url(plan_name)
+                except AttributeError:
+                    pass
+
+        # All other cases are covered by this condition
+        if not url:
+            tmt.utils.git.assert_git_url(self.step.plan.name)
+
+    def process_distgit_source(self, git_root: Path, ref: Optional[str], sourcedir: Path) -> None:
```
lukaszachy, line 572: `ref` is used only in `distgit_dir=self.testdir if ref else git_root`, could caller decide where is  distgit_dir?

tmt/utils/git.py
martinhoyer: saw this one coming :)

tmt/utils/git.py
bajertom: I've added the assertion for `plan_name` not being `None`.

tmt/steps/discover/fmf.py
bajertom: Now it is decided before calling the function.

tmt/utils/git.py
bajertom: I've used the `get_git_root` mentioned above, but had to change the messaging a bit to accommodate for `plan_name` now not being passed to the helper function as an argument and to avoid duplication of the text. 

I've tried to used f" formatting, but couldn't figure out how so I settled for .format.  I'm not sure if the parentheses + r""" is the best way, but all the other ways I tried either resulted in the string being cut in half or erroring out on `plan_name` not being defined or going against the latest decision from tmt hacking session on formatting of long strings. Any idea?

martinhoyer (commented):
Some suggestions/nitpicks:
```
diff --git a/tmt/steps/discover/fmf.py b/tmt/steps/discover/fmf.py
index 1b56fddc..7300901b 100644
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -1,8 +1,8 @@
 import contextlib
 import glob
-import os
 import re
 import shutil
+from contextlib import suppress
 from typing import Any, Optional, cast

 import fmf
@@ -545,24 +545,20 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
     def validate_fmf_id_requirements(self, url: Optional[str] = None) -> None:
         """
         Validate environment requirements for the --fmf-id option.
-        """

-        validation_message = (
-            r"""`tmt run discover --fmf-id` without `url` option """
-            r"""in plan `{plan_name}` can be used only within git repo."""
-        )
-        # It covers only one case, when there is:
-        # 1) no --url on CLI
-        # 2) plan w/o url exists in test run
+        It covers only one case, when there is:
+        1) no --url on CLI
+        2) plan w/o url exists in test run
+        """
         if not self.opt('url'):
             try:
-                fmf_tree = fmf.Tree(os.getcwd())
+                fmf_tree = Path(fmf.Tree).cwd()
             except fmf.utils.RootError:
                 raise tmt.utils.DiscoverError(
                     "No metadata found in the current directory. Use 'tmt init' to get started."
                 )
             for attr in fmf_tree.climb():
-                try:
+                with suppress(AttributeError):
                     plan_url = attr.data.get('discover').get('url')
                     plan_name = attr.name
                     assert plan_name is not None
@@ -571,10 +567,9 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
                             self.get_git_root(directory=Path.cwd())
                         except tmt.utils.RunError:
                             raise tmt.utils.DiscoverError(
-                                validation_message.format(plan_name=plan_name)
+                                f"`tmt run discover --fmf-id` without `url` option in "
+                                f"plan `{plan_name}` can be used only within git repo."
                             )
-                except AttributeError:
-                    pass

         # All other cases are covered by this condition
         if not url:
@@ -582,7 +577,8 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
                 self.get_git_root(directory=Path.cwd())
             except tmt.utils.RunError:
                 raise tmt.utils.DiscoverError(
-                    validation_message.format(plan_name=self.step.plan.name)
+                    f"`tmt run discover --fmf-id` without `url` option in "
+                    f"plan `{plan_name}` can be used only within git repo."
                 )

     def process_distgit_source(self, distgit_dir: Path, sourcedir: Path) -> None:
```
 - I hope we don't do triple quotes for two line sentence? 
 - f-strings are generally prefered to .format, we don't add any extra lines if it's only used twice anyway.
  - Pathlib ftw
  - contextlib.supress

tmt/steps/discover/fmf.py
```diff
@@ -435,47 +434,10 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
                 "Cannot manipulate with dist-git without the `--dist-git-merge` option."
             )
 
-        # Raise an exception if --fmf-id uses w/o url and git root
+        # Raise an exception if --fmf-id used w/o url and git root
```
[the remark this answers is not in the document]
bajertom: Fixed

tmt/utils/git.py
bajertom: Switched back to f-string.

martinhoyer:
@coderabbitai full review

happz:
@coderabbitai full review

bajertom:
Addressed coderabbitai's comment in https://github.com/teemtee/tmt/pull/3701/commits/5d71903e89368b48ba0c5228dc5f8e0d9f3a6f1c 
@coderabbitai full review

tmt/steps/discover/fmf.py
psss: @lukaszachy, is this covered well?

tmt/steps/discover/fmf.py
```diff
@@ -608,6 +548,79 @@ def assert_git_url(plan_name: Optional[str] = None) -> None:
...
+                    plan_name = attr.name
+                    assert plan_name is not None
+                    if not plan_url:
+                        try:
+                            self.get_git_root(directory=Path.cwd())
+                        except tmt.utils.RunError as exc:
+                            raise tmt.utils.DiscoverError(
+                                f"`tmt run discover --fmf-id` without `url` option "
+                                f"in plan `{plan_name}` can be used only within git repo."
+                            ) from exc
```
psss: This part is irrelevant, we should not build `fmf` tree and try to parse `url` from the plan. Either use the attached parent plan directly or just check for the presence of git.

psss: I'd probably suggest to drop these `url` checks here completely and cover the missing git repo when exporting the `fmf` identifiers which is currently handled here:

https://github.com/teemtee/tmt/blob/3196e29a64ee9154eec5da2b8ada1afbab49c7e6/tmt/steps/discover/__init__.py#L457-L477

I think raising an exception when there is a test found which does not have `url` set could be sufficient:

https://github.com/teemtee/tmt/blob/3196e29a64ee9154eec5da2b8ada1afbab49c7e6/tmt/steps/discover/__init__.py#L466-L467

bajertom: Addressed in https://github.com/teemtee/tmt/pull/3701/commits/d602ac5707d21e35595fc13bafbbc76ee3549a9f.

Is an explicit git repo check (like in the original code) here necessary? If we are in a gitrepo, then it will take the url from there and continue, if not, then exception will occur.

bajertom: The accompanied failing tests check the Exception output, which I have slightly changed with addition of `or on command line` - should I keep it there and change the tests or keep the Exception the way it was?

tmt/steps/discover/__init__.py
```diff
@@ -463,16 +463,19 @@ def go(self, force: bool = False) -> None:
                 for test_origin in self.tests(enabled=True):
                     fmf_id = test_origin.test.fmf_id
 
-                    if not fmf_id.url:
-                        continue
+                    if not (fmf_id.url or self.opt('url')):
```
happz:
```suggestion
                    if not (fmf_id.url or self.data.url):
```

tmt/steps/discover/__init__.py
```diff
@@ -463,16 +463,19 @@ def go(self, force: bool = False) -> None:
...
+                            f"within git repo."
+                        )
 
                     exported = test_origin.test.fmf_id.to_minimal_spec()
 
                     if fmf_id.default_branch and fmf_id.ref == fmf_id.default_branch:
                         exported.pop('ref')
 
                     export_fmf_ids.append(tmt.utils.dict_to_yaml(exported, start=True))
-
```
happz: Empty lines are fine, they make code more spacey, less condensed, easier to read :)

bajertom: That is actually what a pre-commit did for me.

bajertom: I thought it might be some new configuration of pre-commit or something, so I've let it do its thing.

happz: pre-commit adding empty lines?? That's new...

tmt/steps/discover/__init__.py
psss:
```suggestion
                    if not fmf_id.url:
```

I would keep the condition as this and state in the exception that the `--fmf-id` option requires a git repository. Let's keep things simple. I don't think we need to describe all the ways how to get it.

tmt/steps/discover/__init__.py
bajertom: It removed it, didn't it?

tmt/steps/discover/fmf.py
psss: Please, adjust the tests to the new wording. Also feel free to simplify them or drop irrelevant ones.

tmt/steps/discover/__init__.py
happz: Right.

pre-commit removing empty lines?? That's new...

I don't understand which rule would it make to remove it from this place. Between methods, between classes, sure, but in a method body?

bajertom: I'd swear I had to run pre-commit at least two times because it complained twice and changed the files twice in a row and some of those changes were these lines, but of course I am unable to reproduce it now :D In any case, I am going to add the empty line back there.

tmt/steps/discover/fmf.py
bajertom: I have adjusted grepped outcome of two test scenarios to fit the new Exception message. I've also deleted one redundant test scenario for which there was no reason to exist to begin with - only difference was in a name of a copied file (a-non-url.fmf vs z-non-url.fmf), everything else including content of the files was the same.

tmt/steps/discover/__init__.py
bajertom: Resolved in 56d5f8589160cc98c081f778625770dc3ccb437a.

tmt/steps/discover/fmf.py
```diff
@@ -500,7 +456,7 @@ def assert_git_url(plan_name: Optional[str] = None) -> None:
             if path is not None:
                 fmf_root: Optional[Path] = path
             else:
-                fmf_root = Path(self.step.plan.node.root)
+                fmf_root = Path(self.step.plan.node.root) if self.step.plan.node.root else None
```
bajertom: I had to insert this check otherwise trying to use `tmt run discover --how fmf` in a directory without fmf root would result in an ugly 
`argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'`
 exception instead of a nice message, which by the way happens in the original code too.

happz: Use `self.step.plan.fmf_root`, that should give you the same outcome without peeking into `Plan` implementation.

happz: Plus, `node.root` is not well-annotated, `fmf_root` is. `fmf_root = Path(self.step.plan.node.root)` should be reported by mypy/pyright, because `node.root` can be `None`, but fmf does not offer good annotations yet, and mypy remains silent. Which is why it blew up for you, it *can* and sometimes *is* `None` in real-life situations.

bajertom: Addressed in d14f35976a63b5c0cc217c1ef55d70076839d25b.

tmt/steps/discover/__init__.py
bajertom: Addressed in https://github.com/teemtee/tmt/pull/3701/commits/eb1e4c1a8678a7be1cb3d97141d9d867f43dbe32.

tests/discover/filtering.sh
```diff
@@ -190,24 +190,6 @@ in plan \`/plans/a-non-url\` can be used only within git repo." output
...
-        rlRun "cp plans/example.fmf $tmp_dir1/plans/z-non-url.fmf"
-        rlRun "cd $tmp_dir1"
-        rlRun "tmt run -rdvvv discover -h fmf --fmf-id finish 2>&1 >/dev/null \
-               | tee output" 2
-        rlAssertGrep "\`tmt run discover --fmf-id\` without \`url\` option \
-in plan \`/plans/z-non-url\` can be used only within git repo." output
-
-        rlRun "tmt run -rdvvv discover -h fmf --fmf-id \
-               --url https://github.com/teemtee/fmf finish 2>&1 >/dev/null | tee output" 0
-        rlRun "rm -rf $tmp_dir1 $tmp_dir2"
```
thrix, line 181: @bajertom why are we dropping this test?

thrix: Even if the code change, maybe we should cover the case when `url` is not specified?

thrix:
@bajertom it would really help me if you would describe in the merge request description the changes done, and the logic behind them, kind of summary of the refactoring would be great.

tests/discover/filtering.sh
psss: @thrix, see: https://github.com/teemtee/tmt/pull/3701#discussion_r2172133003

thrix: yes, thanks, thus my ask to keep a nice up-to-date description so we the reviewers can more easily follow the changes.

bajertom:
@thrix

I have improved the summary to describe the refactoring changes. I welcome any other improvement idea.

happz (approved):

tmt/steps/discover/fmf.py
psss: Thanks, looks good.

psss (approved):
Now looks good, thanks!

thrix (approved):
LGTM

thrix:
Testing failures are flakes, see https://github.com/teemtee/tmt/pull/3861 which passed with raised timeouts

PR #3709, opened 2025-04-30
Add `tmt about hints` command

happz:
Another step in the "make tmt CLI nicer", `tmt about hints` will list and print hints known to tmt. Next, we will make hints more verbose, rendered in HTML, and self-referencing for more details.

Pull Request Checklist

* [x] implement the feature


tmt/cli/about.py
```diff
@@ -84,25 +97,59 @@ def plugins_ls(context: Context, how: str) -> None:
...
+
+    if hint_ids:
+        for hint in tmt.utils.hints.get_hints(*hint_ids):
+            context.obj.logger.print(hint.render(context.obj.logger))
+
+    elif context.invoked_subcommand is None:
+        echo(context.get_help(), color=context.color)
+
+
+@hints.command(name='ls')
```
lukaszachy, line 130: With other `show` vs `ls` subcommands in mind - isn't `hints ls` closer to how `hints show` would behave? 

I expected `ls` to just list all available hint ids without any hint text.

happz: I guess it gets complicated with the option to pick different output formats. It's pretty much both `show` and `export`, depending on whether you pick `yaml` or `pretty` output. I think changing the name of the command is possible, but I would prefer not to introduce other commands, i.e. one with a selectable format seems good enough to me.

PR #3715, opened 2025-05-06
Fix tests/execute/nonroot for boot based Fedora CoreOS

thrix:
The Fedora CoreOS has moved to `bootc` and that revealed some genuine issues:

1. the container images need to be prepared under `root` user, so they are visible to the `bootc switch` command, which must be run under `root`.

2. the `/tmp` directory used by the test does not persist between reboots, use `/var/tmp` instead.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/execute/nonroot/test.sh
+++ b/tests/execute/nonroot/test.sh
@@ -3,7 +3,8 @@
 
 rlJournalStart
     rlPhaseStartSetup
-        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        # bootc images do not persist /tmp on reboot, use /var/tmp instead
+        rlRun "run=\$(mktemp -d -p /var/tmp)" 0 "Create run directory"
         rlRun "pushd data"
     rlPhaseEnd
 
--- a/tests/execute/nonroot/test.sh
+++ b/tests/execute/nonroot/test.sh
@@ -9,7 +9,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest
-        if ! rlRun "TMT_SHOW_TRACEBACK=full tmt run -vvv --id $run"; then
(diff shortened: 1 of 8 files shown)


tests/execute/nonroot/test.sh
```diff
@@ -3,7 +3,8 @@
 
 rlJournalStart
     rlPhaseStartSetup
-        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        # bootc images do not persist /tmp on reboot, use /var/tmp instead
```
thrix, line 6:
```suggestion
        # use `/var/tmp` to persist the run dir during reboots, because `bootc` package manager used
```

martinhoyer (approved):
lgtm

therazix (approved):

happz (approved):

therazix (approved):

martinhoyer (approved):
I know the ShellScript dedents things, but I'm thinking triple quoted f-strings should be used only for longer 'scripts', just 2c. Non-blocking of course
```
                        f"{sudo} ( podman pull {base_image}"
                        f" || podman pull containers-storage:{base_image}"
                        f" || bootc image copy-to-storage --target {base_image} )"
```

thrix:
> I know the ShellScript dedents things, but I'm thinking triple quoted f-strings should be used only for longer 'scripts', just 2c. Non-blocking of course
> 
> ```
>                         f"{sudo} ( podman pull {base_image}"
>                         f" || podman pull containers-storage:{base_image}"
>                         f" || bootc image copy-to-storage --target {base_image} )"
> ```

Does not seems to dedent things ... 

```
cmd:
                
                sudo bash -c "(                             ( podman pull quay.io/fedora/fedora-coreos:stable || podman pull containers-storage:quay.io/fedora/fedora-coreos:stable )                             || bootc image copy-to-storage --target quay.io/fedora/fedora-coreos:stable                         )"
```

Issue #3659, opened 2025-04-10
tmt fails to report missing mrack


happz:
```
/var/tmp/tmt/run-013
Found 1 plan.

/default/plan
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: beaker
        order: 50
        arch: s390x
        image: RHEL-10.1-updates-20250409.d.3
        fail: name 'ProvisioningError' is not defined
        multihost name: default-0

...

            File /usr/lib/python3.13/site-packages/tmt/steps/provision/mrack.py, line 1394, in start
              self._create(self._tmt_name())

              self = <tmt.steps.provision.mrack.GuestBeaker object at 0x7fbc1a35b0e0>

            File /usr/lib/python3.13/site-packages/tmt/steps/provision/mrack.py, line 1292, in _create
              except ProvisioningError as exc:

              self = <tmt.steps.provision.mrack.GuestBeaker object at 0x7fbc1a35b0e0>
              tmt_name = 'tmt-013-BNZxxUlH'
              data = CreateJobParameters(...)
```

This is caused by missing `python3-mrack-beaker` package - I don't know how it's possible, I for one cannot remove `python3-mrack-beaker` without also removing tmt, but, well, here we are. So:

* mrack is not installed
* `beaker` plugin starts provisioning by spawning `BeakerAPI`
* before creating `BeakerAPI` instance, it tried to import mrack, but fails - exception si raised
* ```python
  with tmt.utils.signals.PreventSignals(self._logger):
      try:
          response = self.api.create(data)

      except ProvisioningError as exc:
  ```
  ^^ yeah, but `ProvisioningError` comes from mrack package, and that we fail to import, so Python will throw it on our head.

Changing the order fo things, e.g. explicitly calling `import_and_load_mrack_deps()` and then moving to `GuestBeaker.create()`, should provide the correct - well, any... - report WRT missing mrack.

PR #3718, opened 2025-05-06
Fix mrack import requiring mrack import

### LINKED ISSUE #3716: mrack restoring a guest fails because of chicken and egg issue of imports
`tmt --root . run --last login` would try to resume a guest from previous run. If that guest was provisioned via mrack aka `beaker` plugin, plugin would try to inspect the Beaker job:

```python
        try:
            response = self.api.inspect()

            if response["status"] == "Aborted":
                return False

            current = cast(GuestInspectType, response)
            state = current["status"]
                return False

            if state == 'Reserved':
                return True
            return False

            return False
```

Accessing `self.api` imports mrack, but the code tries to except `mrack.errors`, which cannot exist yet. Calling import_and_load_mrack_deps() first, before try, should help.

martinhoyer:
Resolves #3716

### DIFF
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -8,9 +8,10 @@ import re
 from collections.abc import Mapping
 from contextlib import suppress
 from functools import wraps
-from typing import Any, Callable, Optional, TypedDict, Union, cast
+from typing import Any, Callable, Optional, TypedDict, TypeVar, Union, cast
 
 import packaging.version
+from typing_extensions import ParamSpec
 
 import tmt
 import tmt.config
@@ -60,14 +61,17 @@ DEFAULT_API_SESSION_REFRESH = 3600
 # Store import parameters for lazy loading
 _MRACK_IMPORT_ARGS: Optional[tuple[Any, str, tmt.log.Logger]] = None
 
+P = ParamSpec('P')
+R = TypeVar('R')
 
-def ensure_mrack_imported(func: Callable[..., Any]) -> Callable[..., Any]:
+
+def ensure_mrack_imported(func: Callable[P, R]) -> Callable[P, R]:
     """
     Decorator to ensure mrack is imported before the function is called.
     """
 
     @wraps(func)
-    def wrapper(*args: Any, **kwargs: Any) -> Any:
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
         if not _MRACK_IMPORTED and _MRACK_IMPORT_ARGS is not None:
             workdir, name, logger = _MRACK_IMPORT_ARGS
             import_and_load_mrack_deps(workdir, name, logger)
@@ -1235,6 +1239,7 @@ class GuestBeaker(tmt.steps.provision.GuestSsh):
     _api_timestamp: Optional[datetime.datetime] = None
 
     @property
+    @ensure_mrack_imported
     def api(self) -> BeakerAPI:
         """
         Create BeakerAPI leveraging mrack
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -57,7 +57,26 @@ DEFAULT_PROVISION_TICK = 60  # poll job each minute
 #: Kerberos ticket.
 DEFAULT_API_SESSION_REFRESH = 3600
 
+# Store import parameters for lazy loading
+_MRACK_IMPORT_ARGS: Optional[tuple[Any, str, tmt.log.Logger]] = None
 
+
+def ensure_mrack_imported(func: Callable[..., Any]) -> Callable[..., Any]:
+    """
+    Decorator to ensure mrack is imported before the function is called.
+    """
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        if not _MRACK_IMPORTED and _MRACK_IMPORT_ARGS is not None:
+            workdir, name, logger = _MRACK_IMPORT_ARGS
+            import_and_load_mrack_deps(workdir, name, logger)
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+@ensure_mrack_imported
 def mrack_constructs_ks_pre() -> bool:
     """
     Kickstart construction has been improved in 1.21.0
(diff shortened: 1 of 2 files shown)


tmt/steps/provision/mrack.py
```diff
@@ -1244,24 +1244,37 @@ def is_ready(self) -> bool:
...
                 return True
+            # Any other state means not ready yet.
+            self._logger.debug(
+                f"Guest {self.name} job {self.job_id} status is '{state}', not 'Reserved' yet."
+            )
             return False
 
-        except mrack.errors.MrackError:
+        except mrack.errors.MrackError as e:
+            self._logger.warning(f"Guest {self.name} job {self.job_id} inspection failed: {e}")
```
martinhoyer, line 1277: does it make sense to do warning?

martinhoyer:
@coderabbitai full review

happz:
> > #3659 is the same issue, just different place.
> 
> So this is something which still needs to be covered?

Rather examined, whether the fix proposed in this PR applies to #3659 as well. I think it does, but since @martinhoyer was knee-deep in fixing this mrack issue, he could be in a better position to find out.

martinhoyer:
> > #3659 is the same issue, just different place.
> 
> So this is something which still needs to be covered?

Right, I knew I forgot something. 
@happz @psss I've tried to handle it holistically and it eventually spiraled out of control, where I created _compat/mrack.py, added mrack to banned api and it was just so many changed lines with type checkers going insane. Reverted back to this https://github.com/teemtee/tmt/pull/3718/commits/f2913fa7528825f5a4c681478a884b28ab60fd02, but I'm not sure if it make sense to go this way.

tmt/steps/provision/mrack.py
```diff
@@ -36,14 +36,213 @@
...
+        """Get mrack context module"""
+        if not self._imported or self._mrack is None:
+            raise RuntimeError("MrackProxy must be initialized before accessing context")
+        return self._mrack.context
+
+    def errors(self) -> Any:
+        """Get mrack errors module"""
+        if not self._imported or self._mrack is None:
+            raise RuntimeError("MrackProxy must be initialized before accessing errors")
+        return self._mrack.errors
```
happz, line 231: Both `context` and `errors` could decorated with `@property`, to more resemble the module names in the code below (`mrack_proxy.errors.NotAuthenticatedError`).

happz:
> > > #3659 is the same issue, just different place.
> > 
> > 
> > So this is something which still needs to be covered?
> 
> Right, I knew I forgot something. @happz @psss I've tried to handle it holistically and it eventually spiraled out of control, where I created _compat/mrack.py, added mrack to banned api and it was just so many changed lines with type checkers going insane. Reverted back to this [f2913fa](https://github.com/teemtee/tmt/commit/f2913fa7528825f5a4c681478a884b28ab60fd02), but I'm not sure if it make sense to go this way.
> 
> edit: I guess it's one of those things where it's better to set it on fire and start again, but don't mind me, I'm just tired :)

Yes, I believe there is another way. I tried with some lazy imports, and I failed. I think there is a chance for some smart solution, but I think we somehow lost the need for a big, generic approach: there used to be more optional imports, today we're down to mrack and testcloud. Which means, it's probably not necessary to spend more time here and now on the grand solution I'd like to see, and improving the situation is good enough. Which I think your patch does, so, LGTM.

tmt/steps/provision/mrack.py
```diff
@@ -36,14 +36,213 @@
...
+                                ks_components.append(host.kickstart[ks_section])
+
+                        if ks_components:
+                            req['ks_append'] = ['\n'.join(ks_components)]
+
+                # Whiteboard must be added *after* request preparation, to overwrite the default.
+                req['whiteboard'] = host.whiteboard
+
+                # Get logger from proxy
+                proxy_logger = mrack_proxy._logger
```
psss, line 190: Similarly here.

tmt/steps/provision/mrack.py
```diff
@@ -36,14 +36,213 @@
...
+            'mrack': self._mrack,
+            'providers': self._providers,
+            'BEAKER': self._beaker,
+            'BeakerProvider': self._beaker_provider,
+            'BeakerTransformer': self._beaker_transformer,
+            'TmtBeakerTransformer': self._tmt_beaker_transformer,
+        }
+
+        if name in attr_map:
+            return attr_map[name]
```
psss, line 217: What about using an exception instead?

```python
try:
    return attr_map[name]
except KeyError:
    ...
```

tmt/steps/provision/mrack.py
martinhoyer: Removed

tmt/steps/provision/mrack.py
martinhoyer: This part is also removed

tmt/steps/provision/mrack.py
martinhoyer: removed

tmt/steps/provision/mrack.py
```diff
@@ -57,7 +57,26 @@
...
+_MRACK_IMPORT_ARGS: Optional[tuple[Any, str, tmt.log.Logger]] = None
 
+
+def ensure_mrack_imported(func: Callable[..., Any]) -> Callable[..., Any]:
+    """
+    Decorator to ensure mrack is imported before the function is called.
+    """
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
```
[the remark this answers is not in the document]
martinhoyer: Need to learn about all this, thanks!

tmt/steps/provision/mrack.py
```diff
@@ -57,7 +58,29 @@
...
+def ensure_mrack_imported(func: Callable[P, R]) -> Callable[P, R]:
+    """
+    Decorator to ensure mrack is imported before the function is called.
+    """
+
+    @wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+        if not _MRACK_IMPORTED and _MRACK_IMPORT_ARGS is not None:
+            workdir, name, logger = _MRACK_IMPORT_ARGS
+            import_and_load_mrack_deps(workdir, name, logger)
```
psss, line 77: So `_MRACK_IMPORT_ARGS` is defined under `import_and_load_mrack_deps`. This means that this line will be called only if `import_and_load_mrack_deps` was called before. Does that make sense?

martinhoyer: yeah, looks like it's wrong. Giving up.

psss: Ok, then let's use #3754 for now and put this pull request into backlog. We can revisit the approach later.

martinhoyer:
closing in favor of https://github.com/teemtee/tmt/pull/3754

PR #3726, opened 2025-05-10
Proof of concept: using uv, poethepoet

martinhoyer:
Adding it here as a draft for visibility, in case somebody wants to play with it, and/or as a reference for possible future discussions.  

uv build backend is in preview, though there is quite a discussion around the idea: https://github.com/astral-sh/uv/issues/3957.   
Nevertheless, it could be used as a front-end where adequate, with hatchling or setuptools as build backend.

Poethepoet is used with uv as 'runner' to get rid of Makefile and keep things "Python".


martinhoyer:
Thanks for checking it out @happz! 

> uv

imho, let's revisit it once 
 - we have EPEL10
 - not building el9 rpms
 - see how the discussion, development around uv as build backend goes.

> just for the sake of "using" said language. 

What's wrong with that? You already have tooling, libs,  workflows for that language and developers/contributors that are familiar with it, + might not know, or don't want to learn makefile syntax.

I've played with [just](https://github.com/casey/just), which is lovely, but if the same functionality can be done in Python...

happz:
> Thanks for checking it out @happz!
> 
> > uv
> 
> imho, let's revisit it once
> 
> * we have EPEL10
> * not building el9 rpms
> * see how the discussion, development around uv as build backend goes.
> 
> > just for the sake of "using" said language.
> 
> What's wrong with that? You already have tooling, libs, workflows for that language and developers/contributors that are familiar with it, + might not know, or don't want to learn makefile syntax.

I don't think trying to stay in the ecosystem is wrong per se, I just think there are often alternatives that may be better, that's all. Like `uv`, or `ruff`.

> I've played with [just](https://github.com/casey/just), which is lovely, but if the same functionality can be done in Python...

Yes, the functionality would be the same, I'm not disputing that, when called, files would be created and RPMs built, and so on. On the other hand, `just` seems to offer dedicated files to hold the rules, which I think allows some syntax highlighting in various IDEs, maybe a linter as well. `poe` seems to rely on `pyproject.toml` & Python code to implement modules which are then into "scripts" in `pyproject.toml`, which, IIUIC, requires us to write more Python code to run commands, or squash them into strings in `pyproject.toml`. And post of the stuff we call from `Makefile` and `pyproject.toml` too, I think, are just some commands - `podman build` here, `cat` there, `sed`, `sudo`... So turning fully to Python adds extra level of code, more work, wrapping `cat` with `subprocess.check_call()`, this kind of stuff. It will work, no doubt about that, but I just wanted to run a couple of commands...

Well, whatever, take it any way you wish, I shared my thoughts and concerns. I would like to find out what is wrong with `make`, though, that it deserves to be replaced rather than improved or fixed. "Get rid of Makefile" is rather vague.

martinhoyer:
> I don't think trying to stay in the ecosystem is wrong per se, I just think there are often alternatives that may be better, that's all. Like uv, or ruff.

@happz Fair point, but these are made specifically to be used in python project.  (fwiw, one thing about poethepoet I like is that it automatically uses `uv` venvs when the project is using uv)

In general, I guess it's always about the balance of having a large number of best tools for any specific task in your toolbox versus having a lower amount of tools that can do many things. In tmt context, I believe we should strive for simplicity, but am not claiming I know what is the objectively best solution, just that uv+poe was it for me after some small amount of research and experimentation.

> Yes, the functionality would be the same, I'm not disputing that, when called, files would be created and RPMs built, and so on. On the other hand, `just` seems to offer dedicated files to hold the rules, which I think allows some syntax highlighting in various IDEs, maybe a linter as well. `poe` seems to rely on `pyproject.toml` & Python code to implement modules which are then into "scripts" in `pyproject.toml`, which, IIUIC, requires us to write more Python code to run commands, or squash them into strings in `pyproject.toml`. And post of the stuff we call from `Makefile` and `pyproject.toml` too, I think, are just some commands - `podman build` here, `cat` there, `sed`, `sudo`... So turning fully to Python adds extra level of code, more work, wrapping `cat` with `subprocess.check_call()`, this kind of stuff. It will work, no doubt about that, but I just wanted to run a couple of commands...

fwiw, I'm not sure if you're talking about about how the poe backend works, or about it's usage. 
To run commands, I don't see any additional Python code or boiler plate. You can also use `use_exec = true` instead of subprocess. 
Perhaps you mean how `manage_containers.py` look here? I'm quite sure it's not quite the right way to do it - the intention was to show that we could have `/scripts/` directory with python files, which would synergize with poe `script tasks`

But yeah, this is the only experience i have with poe, so all I can do is point to https://poethepoet.natn.io/

> Well, whatever, take it any way you wish, I shared my thoughts and concerns.  

Well your thoughts and concerns are my concerns as well. This draft is quickly put together proof of concept, or in other words a demonstration how things "could look", once/if we ever start discussing it. If we'll just have a meeting and verbally explain intentions to others in a few mins, the discussion might not be so productive.

> I would like to find out what is wrong with `make`, though, that it deserves to be replaced rather than improved or fixed. "Get rid of Makefile" is rather vague.

It's yet another addition to the barrier of entry for people not used to makefile syntax. I mean, why would you use it, unless you work on C project.  
Rust uses `cargo`, web people use project.json, etc, why should Python project use make? Let alone having to explicitly install it when I want to build it in CI.

psss:
@LecrisUT, is this one obsoleted by the recent `uv` changes?

LecrisUT:
Yes, and the current uv issues are being tracked. Biggest one being renovate compatibility. Btw pip will have pylock.toml support soon also

PR #3730, opened 2025-05-13
Tweaks of reboot-related logging

happz:
* core code should mention it is now going to trigger a reboot and with what action;
* `artemis` was duplicating output and deciding the reboot command, that is guaranteed by `Guest.reboot()` already;
* `virtual` did not log its soft-reboot code path;
* dropping the increased verbosity in /tests/execute/nonroot that was needed to debug bootc vs VM reboot.

Pull Request Checklist

* [x] implement the feature


happz:
@coderabbitai full review

therazix (approved):

PR #3738, opened 2025-05-14
Ansible plugin now saves output and produces prepare/finish result

happz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/finish/__init__.py
+++ b/tmt/steps/finish/__init__.py
@@ -1,5 +1,5 @@
 import copy
-from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
 
 import click
 import fmf
@@ -9,12 +9,12 @@ import tmt.steps
 from tmt.container import container
 from tmt.options import option
 from tmt.plugins import PluginRegistry
-from tmt.result import PhaseResult
+from tmt.result import PhaseResult, ResultOutcome
 from tmt.steps import (
     Action,
-    ActionTask,
     Method,
     PhaseQueue,
+    PluginOutcome,
     PluginTask,
     PullTask,
     sync_with_guests,
@@ -33,7 +33,7 @@ class FinishStepData(tmt.steps.WhereableStepData, tmt.steps.StepData):
 FinishStepDataT = TypeVar('FinishStepDataT', bound=FinishStepData)
 
 
-class FinishPlugin(tmt.steps.Plugin[FinishStepDataT, list[PhaseResult]]):
+class FinishPlugin(tmt.steps.Plugin[FinishStepDataT, PluginOutcome]):
     """
     Common parent of finish plugins
     """
@@ -76,10 +76,10 @@ class FinishPlugin(tmt.steps.Plugin[FinishStepDataT, list[PhaseResult]]):
         guest: 'Guest',
         environment: Optional[tmt.utils.Environment] = None,
         logger: tmt.log.Logger,
-    ) -> list[PhaseResult]:
+    ) -> PluginOutcome:
         self.go_prolog(logger)
 
-        return []
+        return PluginOutcome()
 
 
 class Finish(tmt.steps.Step):
@@ -172,7 +172,7 @@ class Finish(tmt.steps.Step):
 
                 guest_copies.append(guest_copy)
 
-            queue: PhaseQueue[FinishStepData, list[PhaseResult]] = PhaseQueue(
+            queue: PhaseQueue[FinishStepData, PluginOutcome] = PhaseQueue(
                 'finish', self._logger.descend(logger_name=f'{self}.queue')
             )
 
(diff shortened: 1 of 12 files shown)


psss:
Saving the ansible output seems quite useful, could be an interesting change for users, would it possibly deserve a short release note?

happz:
> Saving the ansible output seems quite useful, could be an interesting change for users, would it possibly deserve a short release note?

Possibly. It can be argued it would be interesting, on the other hand, tmt is still pretty silent about these logs as they and their paths are not shown, unlike test logs. So my instict is to not bother user with something they would need to dig into a workdir to find it, and wait with the release to the moment `prepare` results are reported on console in the same way as test logs are. TL;DR, no preference, both ways have their pros and cons.

psss:
> So my instict is to not bother user with something they would need to dig into a workdir to find it, and wait with the release to the moment `prepare` results are reported on console in the same way as test logs are.

Understand the point. Would that possibly be just a single `logger.verbose()` or more complicated? If simple enough I'd include it right away and mention in the release notes. If not, let's keep the release not for later.

happz:
> > So my instict is to not bother user with something they would need to dig into a workdir to find it, and wait with the release to the moment `prepare` results are reported on console in the same way as test logs are.
> 
> Understand the point. Would that possibly be just a single `logger.verbose()` or more complicated? If simple enough I'd include it right away and mention in the release notes. If not, let's keep the release not for later.

More complicated, at least in my vision: I would like to reuse the `display` plugin, if it's available, and task it with reporting the results:

```
$ tmt run -i /tmp/tmp.ABoXAKPXIA --scratch -av provision -h local report -h display -vvv
/tmp/tmp.ABoXAKPXIA
warn: User is feeling safe.
Found 1 plan.

/plan
    discover
        how: shell
        summary: 1 test selected
            /script-00
    provision
        
        how: local
        primary address: localhost
        multihost name: default-0
        arch: x86_64
        distro: Fedora Linux 41 (Workstation Edition)
        kernel: 6.15.4-100.fc41.x86_64
        package manager: dnf5
        selinux: yes
        systemd: yes
        is superuser: no
        is_container: no
    
        summary: 1 guest provisioned
    prepare
        
        how: install
        name: essential-requires
        where: default-0
        package: 3 packages requested
            /usr/bin/awk
            /usr/bin/flock
            /usr/bin/python3
    
        summary: 1 preparation applied
    execute
        queued execute task #1: default-0 on default-0
        
        how: tmt
            00:00:00 pass /script-00 (on default-0) [1/1]

    
        summary: 1 test executed
    report
        how: display
        order: 50
            00:00:00 pass /script-00
                logs (with content):
                    output.txt (/tmp/tmp.ABoXAKPXIA/plan/execute/data/guest/default-0/script-00-1/output.txt)
                        content: fake
                    failures.yaml (/tmp/tmp.ABoXAKPXIA/plan/execute/data/guest/default-0/script-00-1/failures.yaml)
        summary: 1 test passed
    finish
        queued finish task #1: State before is valid (no file) on default-0
        queued finish task #2: Ansible we want to test on default-0
        queued finish task #3: State after is as expected (file created) on default-0
        queued finish task #4: Create a file that is pulled during the finish stage on default-0
        
        finish task #1: State before is valid (no file) on default-0
        how: shell
        name: State before is valid (no file)
        overview: 1 script found
        script: bash -xc "! [ -f /tmp/finished ]"
        
        finish task #2: Ansible we want to test on default-0
        how: ansible
        name: Ansible we want to test
        playbook: playbook.yml
        ok: 1 task
        changed: 1 task
        
        finish task #3: State after is as expected (file created) on default-0
        how: shell
        name: State after is as expected (file created)
        overview: 1 script found
        script: bash -xc "[ -f /tmp/finished ]"
        
        finish task #4: Create a file that is pulled during the finish stage on default-0
        how: shell
        name: Create a file that is pulled during the finish stage
        overview: 1 script found
        script: touch /tmp/tmp.ABoXAKPXIA/plan/data/my_file.txt
    
        ..:..:.. pass Ansible we want to test / playbook.yml
            logs (with content):
                output.txt (/tmp/tmp.ABoXAKPXIA/plan/finish/Ansible-we-want-to-test/0/default-0/output.txt)
                    content: ## Ansible we want to test / playbook.yml
                    content: 
                    content: # exit code: finished successfully
                    content: 
                    content: # stdout (16 lines)
                    content: # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                    content: PLAY [A simple playbook in finish phase] ***************************************
                    content: 
                    content: TASK [Gathering Facts] *********************************************************
                    content: ok: [localhost]
                    content: 
                    content: TASK [Create a simple test file] ***********************************************
                    content: changed: [localhost]
                    content: 
                    content: TASK [Make sure a simple environment variable is respected] ********************
                    content: ok: [localhost]
                    content: 
                    content: TASK [Make sure variables with spaces are correctly handed over] ***************
                    content: ok: [localhost]
                    content: 
                    content: PLAY RECAP *********************************************************************
                    content: localhost                  : ok=4    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
                    content: # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    
        summary: 4 tasks completed

total: 1 test passed
```

Note how `display` runs not just under `report`, that's expected, but also under `finish` where it's given the list of results collected from `finish` plugins. There is just one, `shell` plugin does not provide results yet, but that's something I would like to see. Or extend the `report` plugin to report on results from `prepare`, `execute` and `finish`, but `report` runs *before* `finish`...

Anyway, what I'd like to see is not a oneliner, although not complex either, the raw patch I have has 160 lines.

tmt/steps/finish/__init__.py
```diff
@@ -186,30 +186,61 @@ def go(self, force: bool = False) -> None:
...
+                # Plugin was alive till the very end, and returned results.
                 if outcome.result:
-                    results += outcome.result
+                    results += outcome.result.results
+
+                    if outcome.result.exceptions:
+                        for exc in outcome.result.exceptions:
+                            _record_exception(outcome, exc)
+
+                        continue
```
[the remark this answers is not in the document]
happz: Results are added to the collection on line 230:

```
                if outcome.result:
                    results += outcome.result.results  # <<<

                    if outcome.result.exceptions:
                        for exc in outcome.result.exceptions:
                            _record_exception(outcome, exc)

                        continue
```

happz:
@coderabbitai full review

tmt/steps/prepare/__init__.py
```diff
@@ -429,31 +429,64 @@ def _emit_phase(
...
-                results += outcome.result
+                results += outcome.result.results
+
+                if outcome.result.exceptions:
+                    for exc in outcome.result.exceptions:
+                        _record_exception(outcome, exc)
+
+                    continue
 
             self.preparations_applied += 1
```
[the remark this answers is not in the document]
happz: `outcome.result` can be `None`, `result: Optional[TaskResultT]` where `TaskResultT` is `PluginOutcome` in the case of `prepare` and `finish`.

psss:
> Anyway, what I'd like to see is not a oneliner, although not complex either, the raw patch I have has 160 lines.

Ok, makes sense, let's handle that separately.

tmt/steps/prepare/ansible.py
```diff
@@ -168,25 +170,40 @@ class PrepareAnsible(tmt.steps.prepare.PreparePlugin[PrepareAnsibleData]):
 
     _data_class = PrepareAnsibleData
 
+    @property
+    def _preserved_workdir_members(self) -> set[str]:
+        return {
+            *super()._preserved_workdir_members,
+            # Include directories storing individual playbook logs.
+            *{str(i) for i in range(len(self.data.playbook))},
```
psss: These integer directory names are a bit confusing. Here's an example directory tree:

```
/var/tmp/tmt/run-232/plan/finish/
├── Ansible-we-want-to-test
│   └── 0
│       └── default-0
│           └── output.txt
```

Without reading the code I would not understand the extra levels and names. Not completely sure if it would be worth it but perhaps `playbook-0` and `guest-default-0` would make it clear. On the other hand, we probably do not want to make the paths tool long. So, just as a nitpick comment. Feel free to ignore.

psss (approved):
Looks good to me. Nice improvement! Just one nitpick comment which can safely be ignored.

tmt/steps/prepare/ansible.py
happz: Sure, renamed to `playbook-N` in bd884767. `default-0` is a guest name, and should stay - if we want a `guest-` prefix, we should add it in other places as well, e.g. in `execute`.

therazix: Shouldn't this also be renamed to match?
```suggestion
            *{f'playbook-{i}' for i in range(len(self.data.playbook))},
```

happz: Correct, I forgot about that. Fixed in b6b71b46bbf2d7f8a295f018298162fb79873ec9

psss: Thanks!

happz:
Unrelated error, merging.

PR #3741, opened 2025-05-15
Extract failures from the failing checks

### LINKED ISSUE #3666: Extract failures from the failing check logs
Failure log parsing doesn't work correctly with avc check enabled. Beakerlib log detection in `failure()` function in `results.py` searches for
`::  FAIL ::` but if the test failed only due to avc check (i.e. test itself has passed) then there won't be any FAIL. In this situation, the alternate branch for generic log processing is applied. I believe that we should improve beakerlib log detection, e.g. by searching also for other beakerlib typical strings (`"::  PASS ::". ":: INFO ::"` etc.).

I did some testing and I see these issues.

1. the plugin attempts to parse a failure/traceback log from the original test log even though the test had failed only due to an avc check. This is of course doomed to fail. Unfortunately, result object doesn't seem to provide a "nice" way of checking if the test has failed or just a check has failed.
2. dumb beakerlib log detection in `failure()` function in results.py. It searches for "::  FAIL ::" but if the test failed only due to avc check (i.e. test itself has passed) then there won't be any FAIL. In this situation, the alternate branch for generic log processing is applied. I believe that we should improve beakerlib log detection, e.g. by searching also for other beakerlib typical strings ("::  PASS ::". ":: INFO ::" etc.) or something even better (test framework value?).
 3. AVC failures should be part of the failure/traceback log. I am not sure about other check types. Probably they should implement own parsing function which would extend the failure log.
One additional issue I have notices is dumb beakerlib log detection in `failure()` function in results.py. It searches for "::  FAIL ::" but if the test failed only due to avc check (i.e. test itself has passed) then there won't be any FAIL. In this situation, the alternate branch for generic log processing is applied. I believe that we should improve beakerlib log detection, e.g. by searching also for other beakerlib typical strings ("::  PASS ::". ":: INFO ::" etc.).

Below, I am pasting a code that worked for me, definitely not ready for merge and possibly not matching the latest tmt code. Just wanted to illustrate what I did.

```
# diff -u /usr/lib/python3.9/site-packages/tmt/steps/report/reportportal.py.orig /usr/lib/python3.9/site-packages/tmt/steps/report/reportportal.py
--- /usr/lib/python3.9/site-packages/tmt/steps/report/reportportal.py.orig    2025-03-27 08:52:18.000000000 +0100
+++ /usr/lib/python3.9/site-packages/tmt/steps/report/reportportal.py    2025-04-10 16:32:40.811204108 +0200
@@ -481,6 +481,10 @@
                 curr_description = self.data.launch_description
         return curr_description
 
+    def _parse_avc(self, log: str) -> str:
+        matches = re.findall(r'(type=(USER)?AVC.*?)[\n$]', log)
+        return [m[0] for m in matches]
+
     def upload_result_logs(
         self,
         result: tmt.result.BaseResult,
@@ -518,13 +521,22 @@
 
             # Optionally write out failures only for results which implement the failures callable
             if hasattr(result, "failures") and index == 0 and write_out_failures:
-                message = _filter_log(
-                    result.failures(log),
-                    settings=LogFilterSettings(
-                        size=self.data.traceback_size_limit,
-                        is_traceback=True,
-                    ),
-                )
+                # in case the test failed only due to checks, no point parsing test log
+                message = ''
+                # seems we do not have better way of checking if to expect failures in the test log
+                if 'original test result: pass' not in result.note:
+                    message += _filter_log(
+                        result.failures(log),
+                        settings=LogFilterSettings(
+                            size=self.data.traceback_size_limit,
+                            is_traceback=True,
+                        ),
+                    )
+                for check in result.check:
+                    if check.name == 'avc' and check.result == ResultOutcome.FAIL:
+                        message += f'\n\nAVC messages:\n'
+                        l = self.step.plan.execute.read(check.log[0])
+                        message += '\n'.join(self._parse_avc(l))
 
                 self.rp_api_post(
                     session=session,
```

One more thing, even with the change above subresults reporting won't properly report test failing on avc check. Since check is not associated with any particular phase, all phases are reported as green.

therazix:
This PR adds support for extracting failures from failed checks and moves the existing code responsible for parsing failed test logs closer to the framework code. If a test fails due to a failed check, multiple checks, the test itself, or any combination of these, all relevant failures are now stored in the `failures.yaml` files, which are linked to the results. These failures are also reported to ReportPortal.

Resolves #3666 

Pull Request Checklist

* [x] implement the feature
* [x] include a release note


tmt/result.py
```diff
@@ -194,11 +193,17 @@ class BaseResult(SerializableContainer):
...
+    failure: Optional[str] = None
 
     start_time: Optional[str] = None
     end_time: Optional[str] = None
     duration: Optional[str] = None
 
+    def to_serialized(self) -> dict[str, Any]:
+        serialized = super().to_serialized()
+        serialized.pop('failure', None)
+        return serialized
```
therazix, line 205: Currently, failures are stored directly in the `Result` object during result creation, so no additional log parsing is required. I'm not sure if this is the best approach, since we now need to remove the `failure` attribute during serialization. I'd definitely appreciate any idea on how to improve this.

happz: Hmm, that's unfortunate. Let's discuss in today's meeting, there are some other points I'd like to resolve, but maybe they will become obsolete.

therazix: As discussed in the meeting, the failures are now saved inside the `failures.yaml` file, and the results only save a path to this file.

tmt/checks/__init__.py
```diff
@@ -63,6 +64,34 @@ def find_plugin(name: str) -> 'CheckPluginClass':
...
+    """
+
+    path = invocation.check_files_path / tmt.steps.execute.TEST_FAILURES_FILENAME
+    try:
+        invocation.phase.write(
+            path,
+            tmt.utils.dict_to_yaml(failures),
+            mode='a',
+        )
+    except tmt.utils.FileError as error:
```
happz, line 87: Hmhm, I would say that if we fail, we fail. Is there any meaning in not saving the failures? Checks do save files already, is `failures.yaml` any special among them?

tmt/checks/avc.py
```diff
@@ -311,9 +321,15 @@ def create_final_report(
     else:
         outcome = ResultOutcome.ERROR
 
+    assert invocation.phase.step.workdir is not None  # narrow type
     report_filepath = _save_report(invocation, report, report_timestamp, append=True)
+    paths = [report_filepath.relative_to(invocation.phase.step.workdir)]
+
+    failures_path = save_check_failures(invocation, failures, logger) if failures else None
```
happz, line 328: Right, if `save_check_failures()` stops returning `Optional[Path]`, this would be also much simpler and cleaner:

```python
    paths = [
        report_filepath.relative_to(invocation.phase.step.workdir)
        save_check_failures(invocation, failures, logger)
    ]
```

tmt/checks/coredump.py
```diff
@@ -377,27 +377,27 @@ def _check_for_crashes(
...
+            return crashes
 
         except tmt.utils.RunError as exc:
             logger.debug(f"Failed to check for crashes: {exc}")
-            return False
+            return []
 
     def _check_coredump(
-        self, invocation: "TestInvocation", logger: tmt.log.Logger
+        self, invocation: "TestInvocation", logger: tmt.log.Logger, add_failures: bool = True
```
happz, line 389: Is there a situations where we wouldn't want to "add failures"?

tmt/checks/dmesg.py
```diff
@@ -109,17 +125,24 @@ def _save_dmesg(
 
         else:
             outcome = ResultOutcome.PASS
-
-            if any(pattern.search(output.stdout or '') for pattern in self.failure_pattern):
+            text = output.stdout or ''
+            if any(pattern.search(text) for pattern in self.failure_pattern):
                 outcome = ResultOutcome.FAIL
+                failures = self._extract_failures(text)
```
happz, line 131: Or, reverse the order:

```python
failures = self._extract_failures(output.stdout or '')

if failures:
    outcome = ResultOutcome.FAIL

invocation.phase.write(
    path,
    '\n'.join(render_command_report(label=f'Acquired at {timestamp}', output=output)),
    mode='a',
)

log_paths = [
    path.relative_to(invocation.phase.step.workdir),
    save_check_failures(invocation, failures, logger)
]

return outcome, log_paths
```

We no longer want `any`, we want to pick them all, because we want to save them. And I would save the `failures.yaml` every time, even if it's empty: makes the code cleaner and the file always exists, it may be an empty list, but that's fine, machines do not care, it would still be list and consumers do not need to test whether it exists or not: it's always there, and it always contains failures. It may be an empty list, yes, but a list of failures too.

tmt/frameworks/__init__.py
```diff
@@ -37,6 +38,34 @@ def _provides_framework(framework_cls: TestFrameworkClass) -> TestFrameworkClass
     return _provides_framework
 
 
+def save_test_failures(
```
happz, line 41: Looks awfully similar to one in `tmt.checks`. Maybe it could be moved to `tmt.result` so both these places would use it from there?

tmt/frameworks/beakerlib.py
```diff
@@ -17,6 +17,50 @@
 BEAKERLIB_REPORT_RESULT_COMMAND = 'rhts-report-result'
 
 
+def _extract_failure(log_path: Path, invocation: 'TestInvocation') -> Optional[str]:
```
happz, line 20: Convert into returning `list[str]`, please. Even if it would be just a single item, it should still be a list: failures are transported in tmt code as a list of failures. If a particular piece of code is capable of producing just one "summary" failure, that is fine, but it should connect well with the rest of the code: `save_test_failures(invocation, [fail], logger)` - the need to rewrap "output of failure discovery function" as a list shows that `_extract_failure()` needs special handling, and it really isn't special enough.

tmt/frameworks/shell.py
```diff
@@ -1,12 +1,26 @@
...
+def _extract_failure(log_path: Path, invocation: 'TestInvocation') -> Optional[str]:
+    if not log_path.is_file():
+        return None
+    try:
+        log = invocation.phase.read(log_path)
+    except tmt.utils.FileError:
+        return None
+
+    failures = re.findall(r'.*\b(?:error|fail)\b.*', log, re.IGNORECASE | re.MULTILINE)
+    return '\n'.join(failures) if failures else None
```
happz, line 23: Same here, return a list. Here I think we can return individual findings, i.e. not bundling them into a single string and single item of the final list.

tmt/frameworks/shell.py
```diff
@@ -78,6 +95,17 @@ def _process_results_reduce(
 
                 break
 
+        failures: list[str] = []
+        for test_log in test_logs:
+            failure = _extract_failure(test_log, invocation)
```
happz, line 100: `failure += _extract_failure(...)` - `_extract_failure()` will return a list, and all this special handling needed to amend the impedance between `_extract_failure()` and the rest of failure handling will go away.

tmt/frameworks/shell.py
```diff
@@ -131,11 +159,19 @@ def extract_results(
             elif tmt.utils.ProcessExitCodes.is_pidfile(invocation.return_code):
                 note.append('pidfile locking')
 
+        log_path = invocation.relative_path / tmt.steps.execute.TEST_OUTPUT_FILENAME
+        paths = [log_path]
+
+        failure = _extract_failure(log_path, invocation)
+        failures_path = save_test_failures(invocation, [failure], logger) if failure else None
```
happz, line 166: Yup, `[failure]` strikes again. "Failures shall be reported as a list of failures" and the non-optional writing of failures into a file are good rules, all parties should stick to these and the collection will be much cleaner.

tmt/result.py
```diff
@@ -223,6 +222,18 @@ def show(self) -> str:
     def printable_note(self) -> str:
         return ', '.join(self.note)
 
+    def failures(self) -> list[Path]:
```
happz, line 225: Maybe `failure_logs`, as it does not return actual failures. And it can be a `@property`.

tmt/steps/report/junit.py
```diff
@@ -305,11 +305,24 @@ def _read_log_filter(log: Path) -> str:
...
+        Read the contents of a given failures file
+        """
+        failures: list[str] = []
+        for path in paths:
+            try:
+                failures.extend(tmt.utils.yaml_to_list(phase.step.plan.execute.read(path)))
+            except tmt.utils.FileError:
+                continue
+
+        return '\n'.join(failures)
```
happz, line 319: Can it still be a list? Leave the final formatting to the template, maybe it wishes to create a bullet list from these failures.

tmt/steps/report/junit.py
```diff
@@ -305,11 +305,24 @@ def _read_log_filter(log: Path) -> str:
...
+
     environment.filters.update(
         {
             'read_log': _read_log_filter,
             'duration_to_seconds': _duration_to_seconds_filter,
-            'failures': tmt.result.Result.failures,
+            'read_failures': _read_failures_filter,
```
happz, line 325: I would keep the filter name, `failures` returning a list of failures looks good to me.

tmt/result.py
```diff
@@ -575,57 +589,15 @@ def show(self, display_guest: bool = True) -> str:
 
         return ' '.join(components)
 
-    @staticmethod
-    def failures(log: Optional[str], msg_type: str = 'FAIL') -> str:
+    def failures(self) -> list[Path]:
```
happz, line 592: Not static method anymore, but it can be a `@property`.

tmt/steps/report/junit.py
happz: Plus, if you use my suggestion to use `failure_logs` for the property returning paths, you can keep `failures` as a result property, iterating over its own `failure_logs`, and yielding actual failures. And we wouldn't need this custom filter at all, because it ould move into results, and other code could also use to "iterate of result failures". The piece in the template would be something like `result.failures | join("\n") | e`.

tmt/checks/__init__.py
therazix: Changed in 4bd5606367c6306e86fcb276b1cef435b025dd73.

tmt/checks/coredump.py
therazix: From my testing, it resulted in duplicate failures. But it's possible I misunderstood how this check works. I agree that it would be better to always collect failures. Changed in 52a383a8469a251b763db0604db1c62c025f1f03.

tmt/frameworks/__init__.py
```diff
@@ -9,7 +9,6 @@
     from tmt.base import DependencySimple, Test
     from tmt.steps.execute import TestInvocation
 
-
```
happz, line 12: Unrelated change, one extra changed file.

tmt/checks/dmesg.py
therazix: Changed in ad8b7ff1a2ad517b85a01ed686626e158e527e59 and 52a383a8469a251b763db0604db1c62c025f1f03.

tmt/frameworks/__init__.py
therazix: Moved in 78af46b7280c82bc27ab1376340a94ba0bd2769e.

tmt/frameworks/beakerlib.py
therazix: Changed in ad8b7ff1a2ad517b85a01ed686626e158e527e59.

tmt/frameworks/shell.py
therazix: Changed in ad8b7ff1a2ad517b85a01ed686626e158e527e59.

tmt/frameworks/shell.py
therazix: Changed in ad8b7ff1a2ad517b85a01ed686626e158e527e59.

tmt/steps/report/junit.py
```diff
@@ -305,11 +305,24 @@ def _read_log_filter(log: Path) -> str:
...
+    def _read_failures_filter(paths: list[Path]) -> str:
+        """
+        Read the contents of a given failures file
+        """
+        failures: list[str] = []
+        for path in paths:
+            try:
+                failures.extend(tmt.utils.yaml_to_list(phase.step.plan.execute.read(path)))
+            except tmt.utils.FileError:
+                continue
```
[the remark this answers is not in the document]
therazix: Added in fbe169d6d5e4998c74187da285ba1e4f599e1cd8.

tmt/steps/report/junit.py
therazix: Changed in fbe169d6d5e4998c74187da285ba1e4f599e1cd8.

tmt/result.py
therazix: Changed in 9d612f5a47960c39b6d2d0460f1b38a47264403b.

tmt/steps/report/reportportal.py
```diff
@@ -608,26 +608,29 @@ def upload_result_logs(
...
-                message = _filter_log(
-                    result.failures(log),
-                    settings=LogFilterSettings(
-                        size=self.data.traceback_size_limit,
-                        is_traceback=True,
-                    ),
-                )
+            if index == 0 and write_out_failures:
+                for failure_file in result.failures():
+                    failures = tmt.utils.yaml_to_list(self.step.plan.execute.read(failure_file))
```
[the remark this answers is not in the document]
therazix: I'm not sure how to open the failure log file from `Result`, as it doesn't have access to the workdir, or methods that would allow that.

tmt/steps/report/junit.py
happz: I see, I missed that. Ok, let's stick to `failure_logs | failures`.

tmt/steps/report/reportportal.py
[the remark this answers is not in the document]
happz: Ack, makes sense.

tmt/frameworks/beakerlib.py
```diff
@@ -17,6 +17,50 @@
...
+            # name we need to do this in two steps.
+            if copy_phase_name and re.search('[:]{80}', line):
+                # read the next line containing phase name
+                line = next(iterator)
+                failure_log.append(f'\n{line}')
+                copy_phase_name = False
+        # reverse extracted lines to restore previous order
+        failure_log.reverse()
+        return ['\n'.join(failure_log).strip()]
+    return []
```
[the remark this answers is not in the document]
therazix: It is imported at the top of the file.

tmt/frameworks/__init__.py
therazix: Change reverted in 33af30a01681389eb76d825f7b935501c9589912.

tmt/result.py
```diff
@@ -575,57 +591,16 @@ def show(self, display_guest: bool = True) -> str:
...
-            return '\n'.join(failure_log).strip()
-
-        # Check for other failures and errors when not using beakerlib
-        for m in re.findall(
-            rf'.*\b(?=error|fail|{msg_type})\b.*', log, re.IGNORECASE | re.MULTILINE
-        ):
-            filtered += m + '\n'
-
-        return filtered or log
+        failures = super().failure_logs
```
happz, line 600: `failure_logs`

therazix: Renamed in 0ff0e329263ae9b88d7126ab6416ae212fd8c887.

tests/execute/result/failure-logs.sh
```diff
@@ -0,0 +1,54 @@
...
+}
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "pushd failure-logs"
+        rlRun "set -o pipefail"
+        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+    rlPhaseEnd
+
+    tmt_command="tmt run --scratch -a --id ${run} provision --how local execute -vv test --name"
```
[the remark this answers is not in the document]
therazix: Addded in 0a5f3c3de022c89f02e5ff3ec6bef4ff99ca30ee.

tmt/frameworks/beakerlib.py
```diff
@@ -96,6 +138,20 @@ def extract_results(
             if (invocation.path / filename).is_file()
         ]
 
+        # Check for failures in the beakerlib log
+        if (invocation.path / tmt.steps.execute.TEST_OUTPUT_FILENAME).exists():
+            # Save potential failures to the file
+            log.append(
+                save_failures(
+                    invocation,
+                    invocation.test_data_path,
```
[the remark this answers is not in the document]
therazix: Changed in 2b4b894c3c9b1629b6504bb7379503b2a9e6bdf0.

tmt/result.py
```diff
@@ -670,3 +645,28 @@ def results_to_exit_code(results: list[Result], execute_enabled: bool = True) ->
...
+    :param directory: directory to save the file in.
+    :param failures: list of failures to save.
+    """
+
+    path = directory / tmt.steps.execute.TEST_FAILURES_FILENAME
+
+    try:
+        existing_failures = tmt.utils.yaml_to_list(invocation.phase.read(path))
+    except tmt.utils.FileError:
+        existing_failures = []
```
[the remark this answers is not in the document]
therazix: As discussed with @psss, the test output logs also duplicate information with each execution, and addressing this would probably require some larger changes in the code. It's not worth blocking on this. We can fix it in a separate PR if it becomes an issue.

psss (approved):
Great! Thanks for addressing comments. Now looks good. Let's just mention this improvement in release notes.

thrix (approved):
LGTM, looks clean, tried manually.

PR #3742, opened 2025-05-15
Use `quay.io` as image registry for tmt images

happz:
Alignes with the test images, at least we would have just one registry to worry about.

Pull Request Checklist

* [x] implement the feature


skycastlelily (approved):
LGTM^^

PR #1606, opened 2022-10-18
Update specification of the `where` multihost key

adiosnb:
Updating specification related to #1545.

PR #3743, opened 2025-05-15
Move the alternative `where` key definition to deferred stories

psss:
The original draft of the `where` key specification in #1606 contained and alternative syntax which was never implemented and thus is confusing to the users. Let's move it to the deferred stories for now. We can return to the idea later if there is enough interest for that approach.

Pull Request Checklist

### DIFF
--- a/spec/plans/discover.fmf
+++ b/spec/plans/discover.fmf
@@ -107,6 +107,7 @@ description: |
 
 /where:
     summary: Execute tests on selected guests
+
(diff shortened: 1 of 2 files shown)


psss:
Raised on the stakeholder meeting. Let's merge soon to prevent more user confusion.

happz (approved):

therazix (approved):

PR #3745, opened 2025-05-16
Support `tasks` for `mrack` plugin

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/schemas/provision/beaker.yaml
+++ b/tmt/schemas/provision/beaker.yaml
@@ -57,5 +57,10 @@ properties:
   beaker-job-group:
     type: string
 
+  tasks:
+    type: array
+    items:
+      type: object
+
 required:
   - how
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -1022,6 +1022,15 @@ class BeakerGuestData(tmt.steps.provision.GuestSshData):
              """,
     )
 
+    tasks: Optional[list[dict[str, Any]]] = field(
+        default_factory=list,
+        option='--tasks',
+        metavar='TASKS',
+        help="""
+             Tasks added to the beaker job.
+             """,
+    )
+
 
 @container
 class ProvisionBeakerData(BeakerGuestData, tmt.steps.provision.ProvisionStepData):
@@ -1061,6 +1070,7 @@ class CreateJobParameters:
     beaker_job_owner: Optional[str]
     public_key: list[str]
     group: Optional[str]
+    tasks: Optional[list[dict[str, Any]]]
 
     def to_mrack(self) -> dict[str, Any]:
         data = dataclasses.asdict(self)
@@ -1080,6 +1090,8 @@ class CreateJobParameters:
                 data['beaker']['ks_append'] = kickstart
         if self.public_key:
             data['beaker']['pubkeys'] = self.public_key
+        if self.tasks:
+            data['beaker']['tasks'] = self.tasks
 
         return data
 
@@ -1203,6 +1215,7 @@ class GuestBeaker(tmt.steps.provision.GuestSsh):
     provision_tick: int
     public_key: list[str]
     api_session_refresh_tick: int
+    tasks: list[dict[str, Any]]
 
     _api: Optional[BeakerAPI] = None
     _api_timestamp: Optional[datetime.datetime] = None
@@ -1280,6 +1293,7 @@ class GuestBeaker(tmt.steps.provision.GuestSsh):
             beaker_job_owner=self.beaker_job_owner,
             public_key=self.public_key,
             group=self.beaker_job_group,
+            tasks=self.tasks,
         )
 
         if self.is_dry_run:


happz:
Why?

Please, use commit messages and PR description to provide the motivation and what the goal is here.

skycastlelily:
> Why?
> 
> Please, use commit messages and PR description to provide the motivation and what the goal is here.


I know tmt is trying to obsolete beaker's task, however, if we don't pass a task to mrack, it will use 
dummy task,and the [bootc job](https://beaker.engineering.redhat.com/jobs/11079286) will end up with warn, though the system is installed successfully, which can be told from [the console log ](https://lab-02.hosts.prod.upshift.rdu2.redhat.com/beaker/logs/recipes/18712+/18712137/console.log) or the green complete [on the install page](https://beaker.engineering.redhat.com/recipes/18712137#installation) 

How about I'm gonna to update the commit messages and pr description if we decided to accept this mr?^^

skycastlelily:
> I disagree. Usable commit and PR description should not depend on whether the PR will be eventually accepted or not. If you become sick and aren't around to answer the "what is this about" question, we would be stuck waiting for your return. #3636 has a useful description, while its hopes to getting merged reasonably soon are minimal, but others can review without asking me what's going on in that patch.

Sure, make sense, updated^^ https://github.com/teemtee/tmt/pull/3745/commits/8d47bc0703226e0979c73dcdb08a08f08e5cf6e5

happz:
Do the tasks need to be passed in by a user? IIUIC, this aims at Beaker & image mode, a task needs to be added to the Beaker jobs - can `beaker` plugin pick the task and add it to the job, not user? Is it going to be the same task all the time?

skycastlelily:
> can beaker plugin pick the task and add it to the job, not user?

Here is how mrack works for the  task stuff now:https://github.com/neoave/mrack/blob/main/src/mrack/transformers/beaker.py#L249
we could make the default one to check-system which is needed by bootc-beaker, however, I'm not sure 
mrack wants it, according to mrack maintainer,  he also work mrack  for other users:)

happz:
> > can beaker plugin pick the task and add it to the job, not user?
> 
> Here is how mrack works for the task stuff now:https://github.com/neoave/mrack/blob/main/src/mrack/transformers/beaker.py#L249 we could make the default one to check-system which is needed by bootc-beaker, however, I'm not sure mrack wants it, according to mrack maintainer, he also work mrack for other users:)

Well, tmt can make it the default when image mode is requested. It does not need to be injected by mrack, it can be injected by tmt. I'm not sure user should be bothered with the necessity of adding a task - a well-known task, probably always the same one. Giving them a chance to use different tasks, sure, but making it mandatory would make it less user-friendly.

skycastlelily:
> . It does not need to be injected by mrack, it can be injected by tmt. I'm not sure user should be bothered with the >necessity of adding a task - a well-known task, probably always the same one.  Giving them a chance to use different >tasks, sure, but making it mandatory would make it less user-friendly.

make sense, however adding the default task part should be gonna to [the boot-beaker mr](https://github.com/teemtee/tmt/pull/3728),right?:)

and yes, users would need to use  different task:https://gitlab.com/fedora/bootc/tests/bootc-beaker-test/-/blob/main/template/beaker-job-switch-test.xml.template?ref_type=heads#L35 :)

happz:
> > . It does not need to be injected by mrack, it can be injected by tmt. I'm not sure user should be bothered with the >necessity of adding a task - a well-known task, probably always the same one.  Giving them a chance to use different >tasks, sure, but making it mandatory would make it less user-friendly.
> 
> make sense, however adding the default task part should be gonna to [the boot-beaker mr](https://github.com/teemtee/tmt/pull/3728),right?:)

Could be, this PR is mostly about allowing users to list tasks that should be added to a Beaker job, correct? I'm not sure want to support users in picking arbitrary task, like any task from Beaker's task library. I would add this topic to tomorrow's meeting to get more input and unblock this.

> and yes, users would need to use different task:https://gitlab.com/fedora/bootc/tests/bootc-beaker-test/-/blob/main/template/beaker-job-switch-test.xml.template?ref_type=heads#L35 :)

Is it the same one for all jobs, or does it need to match the distro or other keys? In other words, how  I, as a user, would pick the right task?

skycastlelily:
> Is it the same one for all jobs, or does it need to match the distro or other keys?

yeah, the same one

>  In other words, how I, as a user, would pick the right task?

Basically, the tasks are created by the users who pick them, different teams use different tasks, there are some common tasks used by all users, for sure.

@happz @thrix @psss And here is the update:

Just as Miro expected, the tasks in [xiaofeng's job  ](https://gitlab.com/fedora/bootc/tests/bootc-beaker-test/-/blob/main/template/beaker-job-switch-test.xml.template?ref_type=heads#L35
) could be done in tmt way. So, gonna to close this mr:)

However, beaker need a task to submit job,it's beaker's problem,we need to use beaker, so we need to pass a default check-system task ,if not, mrack will use its default /distribution/dummy task which will make the bootc-on-beaker job fail
we could handle that in [boot-beaker mr](https://github.com/teemtee/tmt/pull/3728), pass check-system task, under  
if host.bootc:

PR #3602, opened 2025-03-14
Handle guest logs

skycastlelily:
Fetch log content,save it to provided location or current dir

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage
* [ ] update the specification
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

PR #3747, opened 2025-05-16
Refactor get_url_content

skycastlelily:
This improve it in two ways:
 1. Add a verify param so users could download their trusted content, instead of getting the certificate verify failed error.
 2. show the error to users, if the download failed,so they could get a clue how to fix the problem.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -1587,7 +1587,7 @@ class GuestLogBeaker(tmt.steps.provision.GuestLog):
         :returns: content of the log, or ``None`` if the log cannot be retrieved.
         """
         try:
-            return tmt.utils.get_url_content(self.url)
+            return tmt.utils.get_url_content(self.url, verify=False)
         except Exception as error:
             logger.warning(f'Failed to fetch log: {error}')
             return None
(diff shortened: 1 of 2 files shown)


skycastlelily:
Due to the issues mentioned in the commit message, [fetch-log function](https://github.com/teemtee/tmt/pull/3602) doesn't work well, 
users will get error and logs are not stored:
` HTTPSConnectionPool(host='[beaker.engineering.redhat.com](http://beaker.engineering.redhat.com/)', port=443): Max retries exceeded with url: /recipes/18700772/logs/console.log (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1020)')))`
Not sure we should disable verify, though I think it's safe, at least we should log error message, so that users 
don't need to find the  fail reason manually:)

skycastlelily:
Turns out the warnings only happens on development environment, so, no need to disable the verify, and for  show error part,we have Milos' https://github.com/teemtee/tmt/pull/3752,closing:)

PR #3751, opened 2025-05-19
Support `--dry` option in the `bootc` provision plugin

### LINKED ISSUE #3750: bootc will provision an instance even if users specify "--dry"
Reproducer:

(dev) lnie@fedora:~/tmt$ tmt run --dry -vvv plan --name plans/bootc$

you will the above command stuck  at the provision step for some time, and boot disk is built.

skycastlelily:
Diffrent from virtual plugin, bootc plugin do many work, _build_bootc_disk for instance, before self._guest.start(), we need to check is_dry_run on all places where it takes real action and prevent it to do so.

Fixes #3750.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -12,7 +12,7 @@ rlJournalStart
         rlRun "mkdir -p /var/tmp/tmt"
         rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
         rlRun "dry_run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create dry run directory"
-        rlRun "dry_run_image='quay.io/fedora/fedora-bootc:41'" 0 "Create dry run directory"
+        rlRun "dry_run_image='quay.io/fedora/fedora-bootc:41'" 0 "Set dry run container image"
         rlRun "pushd data"
         rlRun "df -h" 0 "Check available disk space"
     rlPhaseEnd
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -11,8 +11,16 @@ rlJournalStart
         # in the podman machine mount
         rlRun "mkdir -p /var/tmp/tmt"
         rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
+        rlRun "dry_run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create dry run directory"
         rlRun "pushd data"
         rlRun "df -h" 0 "Check available disk space"
+    rlPhaseEnd
+    # This test must be ran first, or podman machine and container image will exist.
+    rlPhaseStartTest "Test dry run for bootc works as expected"
+        rlRun -s "tmt -vvv run --dry --scratch -i $dry_run plan --name /plans/containerfile/includes-deps"
+	rlRun "ls $dry_run/plans/containerfile/includes-deps/provision/default-0/ | grep qcow2" "1"
+	rlRun "podman machine ls  | grep podman-machine-tmt" "1"
+	rlRun "podman inspect fedora-bootc:41" "125"
     rlPhaseEnd
 
     rlPhaseStartTest "Image that needs dependencies"
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -12,15 +12,17 @@ rlJournalStart
         rlRun "mkdir -p /var/tmp/tmt"
         rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
         rlRun "dry_run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create dry run directory"
+        rlRun "dry_run_image='quay.io/fedora/fedora-bootc:41'" 0 "Create dry run directory"
         rlRun "pushd data"
         rlRun "df -h" 0 "Check available disk space"
     rlPhaseEnd
+
     # This test must be ran first, or podman machine and container image will exist.
     rlPhaseStartTest "Test dry run for bootc works as expected"
-        rlRun -s "tmt -vvv run --dry --scratch -i $dry_run plan --name /plans/containerfile/includes-deps"
-	rlRun "ls $dry_run/plans/containerfile/includes-deps/provision/default-0/ | grep qcow2" "1"
-	rlRun "podman machine ls  | grep podman-machine-tmt" "1"
-	rlRun "podman inspect fedora-bootc:41" "125"
+        rlRun -s "tmt -vvv run --dry --scratch -i $dry_run provision -h bootc --container-image $dry_run_image"
+        rlRun "ls $dry_run/plans/containerfile/includes-deps/provision/default-0/ | grep qcow2" "1"
+        rlRun "podman machine ls  | grep podman-machine-tmt" "1"
+        rlRun "podman inspect fedora-bootc:41" "125"
     rlPhaseEnd
 
     rlPhaseStartTest "Image that needs dependencies"
@@ -69,6 +71,7 @@ rlJournalStart
     rlPhaseStartCleanup
         rlRun "popd"
         rlRun "rm -r $run" 0 "Remove run directory"
+        rlRun "rm -r $dry_run" 0 "Remove dry run directory"
         rlRun "podman rmi $IMAGE_INCLUDES_DEPS" 0,1
         rlRun "podman rmi $IMAGE_NEEDS_DEPS" 0,1
     rlPhaseEnd
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 . /usr/share/beakerlib/beakerlib.sh || exit 1
+. ../../images.sh || exit 1
 
 IMAGE_NEEDS_DEPS="localhost/tmt-bootc-needs-deps"
 IMAGE_INCLUDES_DEPS="localhost/tmt-bootc-includes-deps"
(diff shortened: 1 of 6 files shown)


tmt/steps/provision/bootc.py
```diff
@@ -71,23 +71,23 @@ def __init__(
...
-        )
-
-        try:
-            tmt.utils.Command("podman", "machine", "rm", "-f", PODMAN_MACHINE_NAME).run(
-                cwd=self.workdir, stream_output=True, logger=self._logger
-            )
-        except BaseException:
-            self._logger.debug(
-                "Unable to remove podman machine '{PODMAN_MACHINE_NAME}', it might not exist."
+        if self._instance:
```
happz, line 74: Wouldn't it be better to reverse the condition? With `if self._instance is None`, you can immediately `return` if there's nothing to do when there's no instance, and the whole code below would not need to be indented.

tmt/steps/provision/bootc.py
```diff
@@ -429,7 +436,8 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         built_image = image_dir / 'disk.qcow2'
         renamed_image = image_dir / disk_file_name
 
-        built_image.rename(renamed_image)
+        if not self.is_dry_run:
+            built_image.rename(renamed_image)
         data.image = f"file://{renamed_image}"
```
happz, line 441: I don't think this line makes sense after the condition is added, there will be no "renamed image" with `--dry`. I'm not sure what to do here, but `data.image` will be pointing to an image that will not exist, and things will crash eventually.

skycastlelily: hmm, imo, for --dry, it's a common agreement that no real action will be taken, ie,data.image will only be defined, but not used, as start will [return immediately(https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/testcloud.py#L996).

tmt/steps/provision/bootc.py
skycastlelily: addressed^^ :https://github.com/teemtee/tmt/pull/3751/commits/73bce94e9e0ddedab3dc9a7fba721255cc25e5b9

tmt/steps/provision/bootc.py
```diff
@@ -71,23 +71,23 @@ def __init__(
...
+                logger=self._logger,
+                env=PODMAN_ENV if self._rootless else None,
             )
+            try:
+                tmt.utils.Command("podman", "machine", "rm", "-f", PODMAN_MACHINE_NAME).run(
+                    cwd=self.workdir, stream_output=True, logger=self._logger
+                )
+            except BaseException:
+                self._logger.debug(
+                    "Unable to remove podman machine '{PODMAN_MACHINE_NAME}', it might not exist."
```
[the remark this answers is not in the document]
skycastlelily: self._logger.debug will take care of the formatting.

tmt/steps/provision/bootc.py
thrix: @skycastlelily if you would add a test, maybe the concern would be lower :) I like the change, but because there is no test for it, it is easier to argue :D

skycastlelily: yeah, I really very hate people argue , as none of the involved people will be happy but hurt, so I'm gonna to add a test tomorrow :)

martinhoyer (approved):
It's one of those things that doesn't feel right - to have to add dry-run specific conditions in individual plugin methods. Unfortunately I don't understand the code well enough to make an in-depth review. 

The test I have no idea, but probably shouldn't have hardcoded Fedora version?

tests/provision/bootc/test.sh
```diff
@@ -11,9 +11,17 @@ rlJournalStart
         # in the podman machine mount
         rlRun "mkdir -p /var/tmp/tmt"
         rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
+        rlRun "dry_run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create dry run directory"
         rlRun "pushd data"
         rlRun "df -h" 0 "Check available disk space"
     rlPhaseEnd
+    # This test must be ran first, or podman machine and container image will exist.
```
happz, line 18: Please, add an empty line above this one to give the code some breathing space.

tests/provision/bootc/test.sh
```diff
@@ -11,9 +11,17 @@ rlJournalStart
         # in the podman machine mount
         rlRun "mkdir -p /var/tmp/tmt"
         rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
+        rlRun "dry_run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create dry run directory"
```
happz, line 14: The directory is not removed in the cleanup phase.

tmt/steps/provision/bootc.py
```diff
@@ -252,6 +253,8 @@ def _build_derived_image(self, base_image: str) -> str:
         """
         Build a "derived" container image from the base image with tmt dependencies added
         """
+        if self.is_dry_run:
+            return ''
```
happz, line 257: I dare to say that `return ''` is not the right move: if it makes no sense to "build derived image", as suggested by the empty string signalling "nope", then whoever called this method should not have done so. Returning an empty string makes it look like the method run, but the output is dummy and harder for linters to distinguish from a regular outcome.

tmt/steps/provision/bootc.py
```diff
@@ -343,6 +350,8 @@ def _build_bootc_disk(self, containerimage: str, image_builder: str, rootfs: str
         )
 
     def _init_podman_machine(self) -> None:
+        if self.is_dry_run:
+            return
```
happz, line 354: Belongs to `go()` as well, IMHO.

tmt/steps/provision/bootc.py
```diff
@@ -385,7 +394,6 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
 
         data = BootcData.from_plugin(self)
         data.show(verbose=self.verbosity_level, logger=self._logger)
-
```
happz, line 388: A previous empty line dropped.

tests/provision/bootc/test.sh
skycastlelily: Addressed^^ https://github.com/teemtee/tmt/pull/3751/commits/9bfbca76cfae0d3be881fa33a42600928bbf12ce

tests/provision/bootc/test.sh
skycastlelily: Addressed^^ https://github.com/teemtee/tmt/commit/9bfbca76cfae0d3be881fa33a42600928bbf12ce

tmt/steps/provision/bootc.py
```diff
@@ -290,6 +293,8 @@ def _build_base_image(self, containerfile: str, workdir: str) -> str:
         """
         Build the "base" or user supplied container image
         """
+        if self.is_dry_run:
+            return ''
```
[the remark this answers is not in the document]
skycastlelily: Addressed^^ https://github.com/teemtee/tmt/commit/9bfbca76cfae0d3be881fa33a42600928bbf12ce

tmt/steps/provision/bootc.py
```diff
@@ -386,24 +388,27 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         data = BootcData.from_plugin(self)
         data.show(verbose=self.verbosity_level, logger=self._logger)
 
-        if self._rootless:
+        if self._rootless and not self.is_dry_run:
             self._init_podman_machine()
 
         # Use provided container image
+        containerimage = ''
```
happz, line 395: An empty string should be a sign of "unknown", "unset" or "not initialized", `None` is more fitting and type checkers can trace it. I think there are two options:

* `containerimage: Optional[str] = None`, or
* `containerimage: str` - no value assigned, just declaring the name and type.

Then continue as you did. I think the latter option might be better, is clearly expected by `GuestBootc`, and type checkers and linters will tell you if it's left uninitialized somehow. Setting it to an empty string may be masking the error of passing the empty string where a string is expected, which would hardly be a valid value.

tests/provision/bootc/test.sh
```diff
@@ -11,10 +11,20 @@ rlJournalStart
...
+        rlRun "dry_run_image='quay.io/fedora/fedora-bootc:41'" 0 "Set dry run container image"
```
happz, line 15: As mentioned in https://github.com/teemtee/tmt/pull/3751#pullrequestreview-2868034139, the image name should not really be hardcoded in the test like this, and the upstream image shouldn't be used directly either.

Since we do not have any `bootc` image ready for tests, we should get one, `containers/fedora/latest/bootc/Containerfile` maybe? With a corresponding additions to `Makefile`, you would then add `build_container_image "fedora/latest/bootc\:latest"` to the setup phase, and then use `$TEST_IMAGE_PREFIX/fedora/latest/bootc:latest` in your test, instead of the hardcoded versioned image name.

skycastlelily:
> @skycastlelily I dare to say you forgot to `git add` the `fedora/latest/bootc/Containerfile` containerfile :)

Yeah, :sweat_smile: ,updated^^ https://github.com/teemtee/tmt/pull/3751/commits/3cfa9dc2ff7f5ca3e92670b682f7d5c4da1b4214

tmt/steps/provision/bootc.py
```diff
@@ -429,7 +436,8 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         built_image = image_dir / 'disk.qcow2'
         renamed_image = image_dir / disk_file_name
 
-        built_image.rename(renamed_image)
+        if not self.is_dry_run:
```
psss, line 439: Would it make sense to add a short release note?

skycastlelily: sure, addressed  :)https://github.com/teemtee/tmt/pull/3751/commits/a9b85a7fb97ff75483327ba6cd6df4623069b9a3

psss: Great, thanks!

psss: Btw, the checklist in the description should help with reminding about the missing release note. Would you updating it? It's completely fine to remove irrelevant items from it.

psss (approved):

psss:
The `rawhide` failure seems to be an unrelated ubuntu repository issue.

The `/tests/provision/bootc` test itself [failed](https://artifacts.osci.redhat.com/testing-farm/ddddd5ee-79fa-43a4-aca3-97806866325e/) in a weird way:

```python
Installing image: docker://localhost/tmt-bootc-includes-deps
Initializing ostree layout
ERROR Installing to filesystem: Creating ostree deployment: invalid reference format
  File "/run/osbuild/bin/org.osbuild.bootc.install-to-filesystem", line 53, in <module>
    r = main(args["options"], args["inputs"], args["paths"])
  File "/run/osbuild/bin/org.osbuild.bootc.install-to-filesystem", line 48, in main
    subprocess.run(pargs, env=env, check=True)
```

Sounds like some building hiccup? Rescheduled.

psss:
> See Slack from yesterday, it will fail again.

Ah, I see, thanks. Just read it. For reference, it should be this issue

* bootc-dev/bootc/issues/1308

And should be fixed in `1.3.0`.

happz:
No genuine issue related to this PR, https://github.com/teemtee/tmt/pull/3751#issuecomment-2921375120 -> merging.

PR #3754, opened 2025-05-19
Import `mrack` when the `GuestBeaker` instance is created

### LINKED ISSUE #3659: tmt fails to report missing mrack

### LINKED ISSUE #3716: mrack restoring a guest fails because of chicken and egg issue of imports

psss:
Let's perform the conditional `mrack` import when a new `GuestBeaker` instance is being created. In this way we ensure that the functionality will be available for all its methods. Included a simple manual test for now.

Fix #3716.
Fix #3659.

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -20,6 +20,10 @@ new tmt command would then spawn its own and conflict with the forgotten
 one. tmt no longer leaves the SSH master process running, preventing the
 issue.
 
+An issue in the :ref:`/plugins/provision/beaker` provision plugin
+prevented reconnecting to running guests. This has been fixed so
+now it's possible to fully work with existing tmt runs as well.
+
 
 tmt-1.48.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/provision/beaker/main.fmf
+++ b/tests/provision/beaker/main.fmf
@@ -1,6 +1,17 @@
-summary: translate hardware by config tests
-description:
-    Verify translate hardware constraints using custom config works well,
-    and override the default translations.
-test: ./hardware.sh
+# These tests are manual only for now as they need a running
+# Beaker instance
 enabled: false
+
+/reconnect:
+    summary: Verify reconnecting to a provisioned guest
+    description:
+        Make sure that it's possible to reconnect to an already
+        provisioned guest using an open tmt run.
+    test: ./reconnect.sh
+
+/hardware:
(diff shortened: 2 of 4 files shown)


martinhoyer (approved):
nice

happz (approved):

psss:
All test failures seem to be related to broken mirrors, everything else green, merging.

Issue #3759, opened 2025-05-20
Support custom keys inside `.fmf` files


comps:
FMF/TMT currently seem to ignore several keys prefixed with `extra-`, ie. `extra-hardware`, `extra-nitrate`, etc. -- see https://github.com/teemtee/tmt/blob/7e8920c72d564643491974cf80fceb9266ea9113/tmt/base.py#L118-L126 .

Please make this a more standard feature for other projects to use, rather than a specific list of hardcoded names.

Currently, `tmt lint` fails on anything custom:
```
test: ./test.sh
my_key:
  bla_bla: 1
extra-my_key:
  bla_bla: 2
```
```
fail T001 unknown key "my_key" is used
fail T001 unknown key "extra-my_key" is used
```

Alternatively, a different prefix could be used, but `extra-` seems like a sort of undocumented standard for that, in the TMT world.

PR #3761, opened 2025-05-20
Allow the use of 'extra-' metadata key prefix for custom keys

comps:
This attempts to implement https://github.com/teemtee/tmt/issues/3759 by simply ignoring `extra-*` keys in the lint check.

I'm somewhat unsure how to document it (hence the DRAFT) - should it be in `docs/questions.rst` near where other `extra-*` keys are mentioned (in the context of Nitrate) or somewhere else?

I'm also not sure which of the checklist items are relevent here - ie. I can see
```yaml
patternProperties:
  ^extra-:
    type: string
```
already being part of `tmt/schemas/test.yaml`, so does it need updating?

Thanks!

---

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


tmt/base.py
```diff
@@ -1018,7 +1013,11 @@ def _lint_keys(self, additional_keys: list[str]) -> list[str]:
 
         known_keys.extend(additional_keys)
 
-        return [key for key in self.node.get() if key not in known_keys]
+        return [
+            key
+            for key in self.node.get()
+            if key not in known_keys and not key.startswith('extra-')
```
happz, line 1019: I'd put the `'extra-'` string somewhere on the module level, e.g. below `EXTRA_TEST_KEYS`, something like `EXTRA_KEYS_PREFIX`. We should avoid magic values in the code.

happz:
> I'm somewhat unsure how to document it (hence the DRAFT) - should it be in `docs/questions.rst` near where other `extra-*` keys are mentioned (in the context of Nitrate) or somewhere else?

I'm not sure, @psss will know better. I would put a paragraph somewhere under `/spec`, because keys following the `extra-*` pattern should be mentioned as accepted by it ignored by tmt in general. `docs/questions.rst` seems to be speaking about some actual examples and migration, while you are formalizing the whole concept.

> 
> I'm also not sure which of the checklist items are relevent here - ie. I can see
> 
> ```yaml
> patternProperties:
>   ^extra-:
>     type: string
> ```
> 
> already being part of `tmt/schemas/test.yaml`, so does it need updating?

Correct, it does not need an update. I'd leave the following items:

* [x]  implement the feature
* [ ]  write the documentation

comps:
I have abstracted the `extra-` constant to a module-level `EXTRA_KEYS_PREFIX`, wrote some documentation and added a WIP test (to be extended with plans/stories and actually working test code).

The documentation could probably be improved, ie. I wasn't sure whether to capitalize "TMT", or use "Tmt" at the start of a sentence, or if "tmt" should be lowercase even when it starts a sentence.

Since I would like the linting exclusion to extend to any metadata under `extra-` (have it be not just a string, see the docs example), I removed the string enforcement from `patternProperties` that was causing

```
warn: /valid-test:extra-myproject - 123 is not of type 'string'
```

but that just led to

```
warn: /valid-test: - Additional properties are not allowed ('extra-myproject' was unexpected)
```

which I'm currently unsure how to solve without some deeper rewrites. Any ideas welcome.

happz:
> Since I would like the linting exclusion to extend to any metadata under `extra-` (have it be not just a string, see the docs example), I removed the string enforcement from `patternProperties` that was causing
> 
> ```
> warn: /valid-test:extra-myproject - 123 is not of type 'string'
> ```
> 
> but that just led to
> 
> ```
> warn: /valid-test: - Additional properties are not allowed ('extra-myproject' was unexpected)
> ```
> 
> which I'm currently unsure how to solve without some deeper rewrites. Any ideas welcome.

I believe the entry needs to stay, otherwise the key is indeed unexpected. But if I get my JSON schema right, there is a way to tell the value can be anything:

```yaml
patternProperties:
  ^extra-: {}
```

comps:
> ```yaml
> patternProperties:
>   ^extra-: {}
> ```

That did the trick, thanks!

I'll finish the test and un-DRAFT this so it can be properly reviewed.

comps:
I believe this is now ready for a proper review.

I spent a few hours trying to execute tests, and found a seemingly unrelated fail in the `feeling-safe` test,
```
:: [ 16:53:49 ] :: [  BEGIN   ] :: Running 'TMT_FEELING_SAFE=0 tmt run -i /tmp/tmp.PMBrBQ0oJQ --scratch provision -h local plan -n /plans/features/core'
/tmp/tmp.PMBrBQ0oJQ
warn: User is feeling safe.

/plans/features/core
    provision
        
        how: local
        primary address: localhost
        multihost name: default-0
        arch: x86_64
        distro: Fedora Linux 41 (Forty One)
    
        summary: 1 guest provisioned
:: [ 16:53:51 ] :: [   FAIL   ] :: Command 'TMT_FEELING_SAFE=0 tmt run -i /tmp/tmp.PMBrBQ0oJQ --scratch provision -h local plan -n /plans/features/core' (Expected 2, got 0)
```
No clue what could have broken it, but I guess we'll see if it fails on Github CI or if it's an issue with my environment.

martinhoyer:
@coderabbitai full review

spec/core/extra.fmf
```diff
@@ -0,0 +1,39 @@
...
+    plan properties for use by external systems or tmt plugins.
+
+description: |
+    When a test or a plan contains metadata keys with names starting
+    with ``extra-``, the keys are automatically ignored by ``tmt lint``.
+    This allows the user to define ie. ``extra-myproject`` as a top
+    level YAML key and specify any arbitrary YAML-compliant metadata
+    under that key.
+
+    TMT does not otherwise interact with these keys except for a short
```
psss, line 14:
```suggestion
    tmt does not otherwise interact with these keys except for a short
```

therazix (approved):

spec/core/extra.fmf
psss: We're always using `tmt` in lower case.

psss (changes requested):
Nice extension! Thanks for improving this. Looks good. Just one minor consistency hint. Plus I think this would also deserve at least a short release note. Could you please add one to `docs/releases.rst`?

comps:
I have changed TMT to lowercase and added a release note, as requested.

docs/releases.rst
```diff
@@ -4,6 +4,16 @@
     Releases
 ======================
 
+tmt-1.50.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is now possible to use ``extra-*`` metadata keys in tests, plans
+and stories for arbitrary user-defined data, within the limits of
+what YAML allows. These keys are always ignored by ``tmt lint``.
+See the :ref:`/spec/core/extra` key specification for details and
```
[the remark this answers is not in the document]
psss: Sounds completely irrelevant. The `:ref:` syntax is working as expected.

PR #3762, opened 2025-05-21
Extend timeout for a couple of slow tests

psss:
It already happened several times that the test failed because of the outer beakerlib timeout, still during installing the essential required packages. Let's extend the timeout a bit to give enough time when repositories or network are slow.

### DIFF
--- a/tests/prepare/freeze/test.sh
+++ b/tests/prepare/freeze/test.sh
@@ -8,7 +8,7 @@ rlJournalStart
     rlPhaseEnd
     rlPhaseStartTest
         # Timeout large enough to boot the VM
-        rlRun "rlWatchdog \"rlRun 'tmt run -vfi $tmp' 2\" 60"
+        rlRun "rlWatchdog \"rlRun 'tmt run -vfi $tmp' 2\" 120"
         # ^ exitcode has to be 2 (prepare failed)
         rlAssertGrep 'status:.*done' "$tmp/plan/provision/step.yaml"
         rlAssertGrep 'status:.*todo' "$tmp/plan/prepare/step.yaml"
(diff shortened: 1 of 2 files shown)


psss:
Here's a [recent example](https://artifacts.dev.testing-farm.io/c8d408a9-6853-4ede-bcad-296e1dbe3908/work-prepareleg4htry/plans/features/steps/prepare/execute/data/guest/default-0/tests/prepare/freeze-4/output.txt) for reference:

```
    prepare
        
        queued prepare task #2: default-0 on default-0
        
        how: install
        name: essential-requires
        where: default-0
        package: 2 packages requested
            /usr/bin/awk
            /usr/bin/flock
:: [ 05:19:05 ] :: [   LOG    ] :: Command is still running, I am killing it with KILL
```

tests/prepare/freeze/test.sh
```diff
@@ -8,7 +8,7 @@ rlJournalStart
     rlPhaseEnd
     rlPhaseStartTest
         # Timeout large enough to boot the VM
-        rlRun "rlWatchdog \"rlRun 'tmt run -vfi $tmp' 2\" 60"
+        rlRun "rlWatchdog \"rlRun 'tmt run -vfi $tmp' 2\" 120"
```
happz, line 11: Does it even need to have a timeout? Wouldn't `rlRun "tmt run -vfi $tmp" 2` work as well?

psss: The thing is that if `tmt` is not able to handle the freeze (which the test is verifying) we would get the whole test stuck here. So, I believe, simple `rlRun` is not enough here.

therazix (approved):

psss:
Hmmm, `/tests/provision/bootc` again [failed](https://artifacts.osci.redhat.com/testing-farm/5c368023-a886-42b6-aa82-9cd4f5f16152/) because of the timeout, at least for the third time this week, we need to adjust that one as well: c9bde534

therazix (approved):

bajertom (approved):

PR #3763, opened 2025-05-21
Add test for changing workdirroot with virtual plugin

skycastlelily:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -179,18 +179,8 @@ INDENT = 4
 # Default name for step plugins
 DEFAULT_NAME = 'default'
 
-WORKDIR_ROOT_MODE = (
-    stat.S_ISVTX
-    | stat.S_IRUSR
-    | stat.S_IWUSR
-    | stat.S_IXUSR
-    | stat.S_IRGRP
-    | stat.S_IWGRP
-    | stat.S_IXGRP
-    | stat.S_IROTH
-    | stat.S_IWOTH
-    | stat.S_IXOTH
-)
+#: Permissions applied by tmt on its workdir root.
+WORKDIR_ROOT_MODE: int = 0o1777
 
 # Special process return codes
 
@@ -2180,7 +2170,7 @@ class Common(_CommonBase, metaclass=_CommonMeta):
             if not self.workdir_root.is_dir():
                 try:
                     self.workdir_root.mkdir(exist_ok=True, parents=True)
-                    self.workdir_root.chmod(0o1777)
+                    self.workdir_root.chmod(WORKDIR_ROOT_MODE)
                 except OSError as error:
                     raise FileError(f"Failed to prepare workdir '{self.workdir_root}': {error}")
             else:
@@ -2190,7 +2180,7 @@ class Common(_CommonBase, metaclass=_CommonMeta):
                 except PermissionError:
                     raise FileError(
                         f"Failed to change workdir root '{self.workdir_root}' permission to "
-                        "0o1777, you need to change it manually using chmod."
+                        f"'{oct(WORKDIR_ROOT_MODE)}', you need to change it manually using chmod."
                     )
 
         if id_ is None:
--- a/tmt.spec
+++ b/tmt.spec
@@ -79,7 +79,6 @@ Requires:       tmt == %{version}-%{release}
 Requires:       python3-testcloud >= 0.11.7
 Requires:       libvirt-daemon-config-network
 Requires:       openssh-clients
-Requires:       coreutils
 # Recommend qemu system emulators for supported arches
 Recommends:     qemu-kvm-core
 %if 0%{?fedora}


skycastlelily:
Just found out that there was a clash between me and Milos regarding to change the label, he added the labels like 0.001 sec before I was trying to add the ci/full-test label, anyway, have test coverage  label back:)

skycastlelily:
The failures should be unrelated.
Thanks Brab, [the failure is fixed now](https://artifacts.osci.redhat.com/testing-farm/212f2f7e-cfb1-4291-bc12-8da6dcc393cd/). The cause of the problem is that mktemp -d creates the temp dir with u+rwx (see man mktemp) and such permissions are not enough. The workdir should have 1777. If we create a temp dir and use that as workdir root, due to it's an existing dir, the [chmod line](https://github.com/teemtee/tmt/blob/main/tmt/utils/__init__.py#L2139) won't be executed.
Please note, we could use the following code if we want to make the check code more readable, 

`if stat.S_IMODE(self.workdir_root.stat().st_mode) == 0o1777`

but we need import stat, which means we need to add a new dependence, I'm not sure it's worth:)

tmt/utils/__init__.py
```diff
@@ -2139,6 +2139,15 @@ def _check_or_create_workdir_root_with_perms() -> None:
                     self.workdir_root.chmod(0o1777)
                 except OSError as error:
                     raise FileError(f"Failed to prepare workdir '{self.workdir_root}': {error}")
+            else:
+                try:
+                    if self.workdir_root.stat().st_mode != 17407:
+                        self.workdir_root.chmod(0o1777)
```
happz: I would strongly recommend using `S_*` helpers from https://docs.python.org/3/library/stat.html - `17407` as a value is very, very hard to decode for human readers, `0o1777` is the same case. Instead, `chmod(stat.S_ISVTX | stat.S_IRUSR | ...)` should be used - preferably by moving this `stat.*` construct into a global constant, like `WORKDIR_ROOT_MODE` or something.

Check also `stat.S_IMODE(...)`, that should yield just the bits that can be set by `chmod()`, i.e. the value should be much simpler than 17407 - can it even be `0o1777`?

happz:
> but we need import stat, which means we need to add a new dependence, I'm not sure it's worth:)

Importing `stat` is cheap, and it is the way permissions and mode should be inspected and modified. `0o1777` is cool and nice, but when debugging things, the numbers tend to become blury and hard to follow...

tmt/utils/__init__.py
skycastlelily: Updated :https://github.com/teemtee/tmt/pull/3763/commits/1c65aaf92b5c822ff067acd8de428e914cac9f9c ^^

skycastlelily:
>Importing stat is cheap

I would think so if coreutils is  a dependence of tmt, it's not,hmm, add dependence, and use it anyway

> but when debugging things, the numbers tend to become blury and hard to follow..

 Not sure how it could become blury and hard to follow, but workdir_root_mode way is truly clear, so updated accordingly :)

happz:
> > Importing stat is cheap
> 
> I would think so if coreutils is  a dependence of tmt, it's not,hmm, add dependence, and use it anyway

Ah, I see. We are not importing `stat` binary, the import has nothing to do with coreutils; we are importing a package from Python's own standard library, https://github.com/python/cpython/blob/main/Lib/stat.py

> > but when debugging things, the numbers tend to become blury and hard to follow..
> 
>  Not sure how it could become blury and hard to follow,

Easily. Right there, in your patch, for a common reader `17407` is absolutely magic number. `0o1777` is slightly better, but only in its octal form, but ok. And so on. Magic numbers don't belong in code, magic numbers shall be turned into constants and equipped with a comment...

tmt/utils/__init__.py
```diff
@@ -175,6 +176,18 @@ def configure_bool_constant(default: bool, envvar: str) -> bool:
 # Default name for step plugins
 DEFAULT_NAME = 'default'
 
+WORKDIR_ROOT_MODE = (
```
happz: Oh, and this will be even better - we don't need `stat` bits for this \o/ Thanks to `S_IMODE`, we get just the mode, and `WORKDIR_ROOT_MODE` can be set to `0o1777` - this magic number is understandable.

tmt/utils/__init__.py
```diff
@@ -2139,6 +2152,15 @@ def _check_or_create_workdir_root_with_perms() -> None:
...
                 except OSError as error:
                     raise FileError(f"Failed to prepare workdir '{self.workdir_root}': {error}")
+            else:
+                try:
+                    if stat.S_IMODE(self.workdir_root.stat().st_mode) != WORKDIR_ROOT_MODE:
+                        self.workdir_root.chmod(WORKDIR_ROOT_MODE)
+                except PermissionError:
+                    raise FileError(
+                        f"Failed to change workdir root '{self.workdir_root}' permission to "
+                        "0o1777, you need to change it manually using chmod."
```
happz: `0o1777` should be replaced with `WORKDIR_ROOT_MODE`, formatted as an octal number. Single source of truth...

skycastlelily: Updated^^：https://github.com/teemtee/tmt/pull/3763/commits/5db785e5a1dcd3b39c3121f51bc03577980f3d08

tmt/utils/__init__.py
skycastlelily: Updated^^：https://github.com/teemtee/tmt/commit/5db785e5a1dcd3b39c3121f51bc03577980f3d08

tcornell-bus (approved):

PR #3767, opened 2025-05-21
Run dmesg and coredump commands with `silent=True`

martinhoyer:
This is to keep it consistent with avc test check and have less noisy output for users by default, especially from coredump, which prints `err: No match found.` when, it passes by not finding any new coredumps.

Was not able to verify or modify tests as even on main branch, running `PROVISION_HOW=virtual ./test-coredump` is completely broken failed shell stuff. (at least for me)

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] mention the version
* [ ] include a release note


lukaszachy:
My mistake to push the test fix into this branch, but it is related, right? :D

happz (approved):

lukaszachy (approved):

Issue #1721, opened 2022-11-30
New guide chapter: Efficient Metadata Handling


psss:
Discussed yesterday with @lukaszachy that it would be nice to summarize all features which are available for efficient metadata handling and preventing data duplication. The new guide chapter should cover these:

 * using `fmf` inheritance
 * `adjust` based on `context`
 * `yaml` referencing features
 * sharing tests across repositories

psss:
Simple example demonstrating `yaml` referencing:

```yaml
discover:
  - name: upstream
    how: fmf
    filter: &selection tier:1 & tag:container
    url: https://github.com/teemtee/tmt
  - name: fedora
    how: fmf
    filter: *selection
    url: https://src.fedoraproject.org/rpms/tmt/
```

psss:
The proposed topics are partially covered in the Guide:

> * using `fmf` inheritance

https://tmt.readthedocs.io/en/stable/guide.html#inheritance

> * `adjust` based on `context`

not covered yet

> * `yaml` referencing features

https://tmt.readthedocs.io/en/stable/guide.html#anchors-and-aliases

> * sharing tests across repositories

not covered yet

psss:
@mcasquer, you might be interested in this one.

mcasquer:
> > * `adjust` based on `context`
> 
> not covered yet
> 

It seems the adjust based on context is already covered at
https://tmt.readthedocs.io/en/stable/spec/core.html#adjust

psss:
> It seems the adjust based on context is already covered at
> https://tmt.readthedocs.io/en/stable/spec/core.html#adjust

That's true but it would be good to cover this topic in the Guide as well. The main purpose of the guide is to provide a nicely structured information with enough context and comments so that especially newcomers are able to grasp the features quickly. Let's add a new section `Adjust Metadata` with one or two inspiring examples and a pointer to the `adjust` key specification for details.

mcasquer:
> That's true but it would be good to cover this topic in the Guide as well. The main purpose of the guide is to provide a nicely structured information with enough context and comments so that especially newcomers are able to grasp the features quickly. Let's add a new section `Adjust Metadata` with one or two inspiring examples and a pointer to the `adjust` key specification for details.

Done at [#3818](https://github.com/teemtee/tmt/pull/3818)

mcasquer:
Closing this issue as all the points are covered now

PR #3778, opened 2025-05-28
Add the `Share Tests` section into the guide

mcasquer:
Partially covers #1721

Pull Request Checklist

* [x] write the documentation
* [x] include a release note

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -446,28 +446,7 @@ location:
 See the :ref:`/spec/tests/require` attribute specification for
 detailed description of the syntax and available keys.
 
-Share Tests
-------------------------------------------------------------------
 
-Tests can be shared across different repositories, significantly enhancing
-efficiency, reducing data duplication, and improving maintainability in the
-testing workflow.
-
-tmt allows references to external repositories directly within the ``discover``
-step of the plan. By specifying the url of a remote Git repository, tmt
-can fetch and integrate tests defined within that repository.
-
-.. code-block:: yaml
-
-    discover:
-        # Fetch common tests from a shared repository
-        - name: core-tests
-          how: fmf
-          url: https://github.com/my-org/core-tests.git
-
-        # Discover tests located within this project's own repository
-        - name: project-specific
-          how: fmf
 
 Plans
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -446,7 +446,28 @@ location:
 See the :ref:`/spec/tests/require` attribute specification for
 detailed description of the syntax and available keys.
 
+Share Tests
+------------------------------------------------------------------
 
+Tests can be shared across different repositories, significantly enhancing
+efficiency, reducing data duplication, and improving maintainability in the
+testing workflow.
+
+``TMT`` allows references to external repositories directly within the ``discover``
+step of the ``plan``. By specifying the url of a remote ``Git repository``, tmt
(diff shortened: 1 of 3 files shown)


docs/examples.rst
```diff
@@ -446,7 +446,28 @@ location:
...
 
+    discover:
+    # Fetch common tests from a shared repository
+    - name: core-tests
+        how: fmf
+        url: https://github.com/my-org/core-tests.git
+
+    # Discover tests located within this project's own repository
+    - name: project-specific
+        how: fmf
```
happz: The content of `discover` should be indented by 2 spaces.

mcasquer: @happz updated !

bajertom: I think you've moved `how` and `url` by 4 spaces back instead of 2. 
`name`, `how` and `url` should be aligned on the first letter (not the dash).

docs/examples.rst
```diff
@@ -446,7 +446,28 @@ location:
...
 
+Share Tests
+------------------------------------------------------------------
+
+Tests can be shared across different repositories, significantly enhancing
+efficiency, reducing data duplication, and improving maintainability in the
+testing workflow.
+
+``tmt`` allows references to external repositories directly within the ``discover``
+step of the ``plan``. By specifying the url of a remote ``Git repository``, tmt
```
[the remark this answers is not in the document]
bajertom: I wouldn't drop the double tick, but maybe lowered capital G in `Git repository` to keep it consistent with the rest of the documentation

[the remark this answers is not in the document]
happz: A couple of formatting notes: "tmt" shouldn't be wrapped with double ticks, we don't do that in docs; "discover" is correctly wrapped, but "plan" should not be, it's a plan, not a literal string `plan`; "url" should be "URL"; "Git repository" shouldn't be wrapped by any backticks; capitalized or not, that's the question, I don't know what we use more, I found both.

docs/examples.rst
mcasquer: Thanks for pointing it out, updated !

docs/examples.rst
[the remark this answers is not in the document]
mcasquer: Thanks for the feedback, updated the invalid literals, let me know if the capitalized is fine or lowered one is preferred :)

docs/examples.rst
bajertom: Now it's actually indented by 6 spaces? I think the `key: value` pairs of the dictionaries are indented by 4 spaces (from the "motherword") and if there is a list of them, each list item (dictionary) has a dash indented by 2 spaces (with the `key: value` pairs being still indented by 4 spaces).
Like here for example: https://github.com/teemtee/tmt/blob/main/plans/integration/main.fmf 

Maybe I'm getting to deep into this :nerd_face:

happz: Desired state would be this:

```
.. code-block:: yaml

    discover:
      # Fetch common tests from a shared repository
      - name: core-tests
        how: fmf
        url: https://github.com/my-org/core-tests.git

      # Discover tests located within this project's own repository
      - name: project-specific
        how: fmf
```

psss (approved):
Thanks for preparing this. Looks good. We're just trying to slowly but surely get rid of the examples page and move the content to the guide where it will be nicely structured with enough context and comments. I've moved the content in 7e3f9515 and added a short release note about the new section.

mcasquer:
> Thanks for preparing this. Looks good. We're just trying to slowly but surely get rid of the examples page and move the content to the guide where it will be nicely structured with enough context and comments. I've moved the content in [7e3f951](https://github.com/teemtee/tmt/commit/7e3f9515f6ad86089b4b385692d0a5274ca27bd8) and added a short release note about the new section.

@psss thanks a lot !

PR #3781, opened 2025-05-29
Force using IPv4 for apt

thrix:
Recently we are hitting weird network issues in CI, try this as a workaround force using IPv4, seems the problem is related somehow to IPv6.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


comps:
Maybe set a system-wide ipv4 preference instead of forcing it for one tool?

```sh
echo precedence ::ffff:0:0/96 100 > /etc/gai.conf
```

See https://datatracker.ietf.org/doc/html/rfc3484#section-10.3

psss (approved):

PR #3786, opened 2025-06-02
Use the `fedora-bootc:latest` image in the `bootc` test

psss:
The fix for a recent issue has been released in `fedora-bootc:42`. Let's use `fedora-bootc:latest` instead, so that we don't have to update the versions every half a year.

Related: https://github.com/bootc-dev/bootc/issues/1308

### DIFF
--- a/tests/provision/bootc/data/includes-deps.containerfile
+++ b/tests/provision/bootc/data/includes-deps.containerfile
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/fedora-bootc:42
+FROM quay.io/fedora/fedora-bootc:latest
 
 RUN dnf -y install cloud-init rsync vim && \
     ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants && \
--- a/tests/provision/bootc/data/includes-deps.containerfile
+++ b/tests/provision/bootc/data/includes-deps.containerfile
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/fedora-bootc:41
+FROM quay.io/fedora/fedora-bootc:42
 
 RUN dnf -y install cloud-init rsync vim && \
     ln -s ../cloud-init.target /usr/lib/systemd/system/default.target.wants && \


tests/provision/bootc/data/includes-deps.containerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/fedora-bootc:41
+FROM quay.io/fedora/fedora-bootc:42
```
happz: How about using `quay.io/fedora/fedora-bootc:latest`?

psss: That sounds even better. Updated.

henrywang: `latest` == `rawhide/43` in this case.

tests/provision/bootc/data/includes-deps.containerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/fedora-bootc:41
+FROM quay.io/fedora/fedora-bootc:latest
```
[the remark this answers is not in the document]
psss: I guess this is irrelevant as we don't want to regularly update the release versions.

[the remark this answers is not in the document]
happz: I'd say it's relevant in the sense that you did not update the PR subject before the all-mighty AI read the code, and the outdated-ness was real. I guess it could point out the PR subject is wrong, not the patch, given that `latest` is probably better than `42`, but...

psss:
The [bootc plan](https://artifacts.osci.redhat.com/testing-farm/289dd9f0-7418-48c1-9588-b59d5167de49/) is green! \o/

happz (approved):

happz:
Added to 1.50 as it's ruining our CI.

cgwalters:
This change makes sense, but I would also consider using quay.io/centos-bootc/centos-bootc:stream10 by default actually.

PR #3787, opened 2025-06-02
Retry fetching url for distro name in `testcloud`

psss:
Recently the `provision` job failed many times because the image `url` could not be fetched because of some network hiccups. Yet some other subtests around the failing one successfully finished. Let's do a couple retries before giving up.

### DIFF
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -212,9 +212,9 @@ TPM_VERSION_SUPPORTED_VERSIONS = {
     False: ['2.0', '2'],
 }
 
-#: Default retry attempts and interval
-RETRY_ATTEMPTS = 5
-RETRY_INTERVAL = 5
+#: Image url fetch retry attempts and interval
+IMAGE_URL_FETCH_RETRY_ATTEMPTS = 5
+IMAGE_URL_FETCH_RETRY_INTERVAL = 5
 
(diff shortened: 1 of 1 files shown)


psss:
This was blocking merging last week. Proposing with the `must` priority.

tmt/steps/provision/testcloud.py
```diff
@@ -212,6 +212,10 @@ def import_testcloud(logger: tmt.log.Logger) -> None:
     False: ['2.0', '2'],
 }
 
+#: Default retry attempts and interval
+RETRY_ATTEMPTS = 5
+RETRY_INTERVAL = 5
```
happz: Seeing this, I need to ask: retry of what? See line 191, there are already some retries, they do not relate to these "default retry" values, so just "retry" is puzzling. In other words, if I change `RETRY_ATTEMPTS`, what behavior would it affect?

I would try to pick more specialized names for these two new constants. Maybe something like `IMAGE_URL_FETCH_{ATTEMPTS,INTERVAL}`? 

(Hmhm, time has come to unify the naming in general, across the board, we have attempts, retries, delays, ticks, intervals... Not good.)

psss: Ok, renamed in e7c584d5.

happz (approved):

Issue #3672, opened 2025-04-16
Create an `inventory.yaml` file with all provisioned guests


psss:
Similar to the [Guest Topology Format](https://tmt.readthedocs.io/en/stable/spec/plans.html#guest-topology-format), the `inventory.yaml` file should contain information about all provisioned guests in the `ansible` format including guest roles so that a single `ansible-playbook` call could be used to apply playbooks across all guests in parallel.

psss:
Tentatively assigning to @dav-pascual as he's planning to look into this area.

PR #3796, opened 2025-06-05
Describe Ansible inventory generation feature

dav-pascual:
Stories for generation of Ansible inventory feature described in https://github.com/teemtee/tmt/issues/3672

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation


spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
...
+
+  - |
+    # path/to/the/layout.yaml
+    inventory_layout:
+      all:
+        children:
+          server:
+            children:
+              master: {}
+              replica: {}
```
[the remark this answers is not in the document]
dav-pascual: The layout can be a file meant to be reused for multiple plans, and not all groups have to contain a host all the time

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+                vars:
+                  var1: value1
+                  var2: value2
+
+          - name: client
+              role: client
+              ansible:
+                vars:
+                  ansible_user: fedora
+                  ansible_port: 2222
```
dav-pascual: How can I explicitly specify `status: idea` ?

happz: I believe by not including any of the `implemented-by` or `verified-by` links. If you leave `link` key empty, the story should be treated as "idea".

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
...
+          client: {}
+          runner: {}
```
happz: What I miss is the connection between `provision` content above, and these groups: is tmt somehow "filling the blanks" based on guests' names and/or roles? I can imagine a small template to "connect the dots", to fill `master`, `replica`, `client` and `runner` with `ansible_ssh_host` and other variables for respective guests, but as the example is now, it's static, and I don't see how it would get filled with data.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
+summary: Define Ansible general settings for the run
```
happz:
```suggestion
summary: Define general Ansible settings for a plan
```

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
...
+
+story:
+    As a user, I want to configure Ansible related settings for the test run,
```
happz:
```suggestion
    As a user, I want to configure Ansible-related settings for the plan,
```

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
...
+    for example, the ability to define the structure of the generated inventory
```
happz:
```suggestion
    for example, the ability to define the structure of the Ansible inventory
```

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
...
+    using a customizable layout template.
+
+description: |
+    When running tests, a static inventory file is generated from
+    provision metadata. The ``ansible.inventory.layout`` key allows
+    users to specify a YAML file defining the desired inventory group
```
happz: Will the file be treated as a template? IIRC, that was the idea, I would mention it here right from the beginning.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,45 @@
...
+        role: master
+      - name: test-runner
+        role: runner
+      - name: test-client
+        role: client
+
+    execute:
+      how: tmt
+
+  - |
```
happz, line 37: The content below belongs into the example above, it's just one example, just spread across two files.

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+/ansible:
+    summary: Per-guest Ansible customization
+
+    story:
+        As a user I want to assign an Ansible group and host variables
+        for a provisioned guest to be reflected in the generated inventory.
+
+    description: |
+        Each provisioned guest may define an optional ``ansible`` key
+        to control its behavior in the generated Ansible inventory.
```
happz:
```suggestion
        to define its corresponding Ansible inventory section.
```

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+
+        - ``group``: Assigns the guest to a specific group. If omitted,
```
happz:
```suggestion
        - ``group``: Assigns the guest to a Ansible specific group. If omitted,
```

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+          the ``role`` value from the guest is used as the group name.
```
happz:
```suggestion
          the ``role`` key would serve as the Ansible group name.
```

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+          If no role is defined either, the host will be placed in the
```
happz:
```suggestion
          If ``role`` is not set either, the host will be placed in the
```

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+          'ungrouped' section of the inventory.
```
happz: What is "ungrouped" section of the inventory? Will there be guest name directly under `all`?

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+
+        - ``vars``: Defines host-specific Ansible variables to include
+            under that host in the inventory.
```
happz: Two extra spaces make the rendered output way different, see https://tmt--3796.org.readthedocs.build/en/3796/spec/plans.html#spec-plans-provision-ansible

spec/plans/ansible.fmf
dav-pascual: tmt puts hosts (guests) on groups based on information provided in plan file, in the provision step. With order of preference:
1. ansible.group
2. role

If none of above are defined, host goes to 'ungrouped'. All host are always added to 'all' group with the correspondent vars.

The main use case of this is having a lot of preexisting playbooks which target certain groups. All these playbooks should be run against the same inventory, which has a certain group hierarchy.

For example. If we provide this layout:

```
all:
  children:
    ad:
      children:
        ad_root: {}
        ad_rootbackup: {}
        ad_subdomain: {}
        ad_subsubdomain: {}
        ad_treedomain: {}
        ad_member: {}
    ipa:
      children:
        ipaclient: {}
        ipaserver: {}
    linux:
      children:
        ipa: {}
        runner: {}
        client: {}
        server: {}
        replica: {}
    windows:
      children:
        ad: {}
```

with this plan:

```
discover:
  - how: fmf

ansible:
  inventory:
    layout: test-layout.yaml

provision:
  - name: test-master
    ansible:
      group: server
      vars:
        var1: value1
        var2: value2
  - name: test-runner
    role: runner
  - name: test-client1
    role: client
  - name: test-client2
    role: client

prepare:
  - how: ansible
    playbook:
      - twd/ansible/setup_server.yml
      - twd/ansible/setup_clients.yml
      - twd/ansible/setup_runner.yml
      - twd/ansible/all_hosts.yml
    extra-args: '-vvv'

execute:
  how: tmt
```

this inventory file is generated:
```
all:
    children:
        ad:
            children:
                ad_root: {}
                ad_rootbackup: {}
                ad_subdomain: {}
                ad_subsubdomain: {}
                ad_treedomain: {}
                ad_member: {}
        ipa:
            children:
                ipaclient: {}
                ipaserver: {}
        linux:
            children:
                ipa: {}
                runner:
                    hosts:
                        test-runner: {}
                client:
                    hosts:
                        test-client1: {}
                        test-client2: {}
                server:
                    hosts:
                        test-master: {}
                replica: {}
        windows:
            children:
                ad: {}
    hosts:
        test-master:
            ansible_host: 127.0.0.1
            ansible_user: root
            ansible_port: 10028
            var1: value1
            var2: value2
        test-client1:
            ansible_host: 127.0.0.1
            ansible_user: root
            ansible_port: 10031
        test-client2:
            ansible_host: 127.0.0.1
            ansible_user: root
            ansible_port: 10032
        test-runner:
            ansible_host: 127.0.0.1
            ansible_user: root
            ansible_port: 10033
```

happz: Isn't the layout in your example exactly the same as the final inventory file? :O It feels like a copy & paste issue, the layout being actually the final inventory by mistake.

dav-pascual: My bad, you're right. Updated.

I've borrowed part of this logic from [mrack project](https://github.com/neoave/mrack/blob/main/src/mrack/outputs/ansible_inventory.py). And as we discussed, even if an inventory layout is passed in its final state, tmt processes it and only does required changes.

happz: I see, now it's much clearer.

spec/plans/ansible.fmf
dav-pascual: Output file is static. Input layout can be just the template layout or the inventory in it's final state

In principle, [dynamic inventory](https://docs.ansible.com/ansible/latest/inventory_guide/intro_dynamic_inventory.html) is not supported. That's why I specified static

spec/plans/provision.fmf
dav-pascual: It's a default group: https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html#default-groups 

`The all group contains every host. The ungrouped group contains all hosts that don’t have another group aside from all. Every host will always belong to at least 2 groups (all and ungrouped or all and some other group). `

happz: I see. Makes sense. I would then speak about "group" rather than "section":

```suggestion
          ``ungrouped`` group.
```

psss:
Thanks much for outlining the spec! I believe we should finish the review soon as there is #3839 already in progress, proposing to squeeze into the current sprint. No implementation here so it should not consume that much time.

spec/plans/provision.fmf
dav-pascual: Empty `link` gives a warning of wrong schema. I will update this doc with the right file in https://github.com/teemtee/tmt/pull/3839

dav-pascual:
> Thanks much for outlining the spec! I believe we should finish the review soon as there is #3839 already in progress, proposing to squeeze into the current sprint. No implementation here so it should not consume that much time.

Thanks for the feedback. Any more action item before merging?

spec/plans/ansible.fmf
tcornell-bus: Would be possible to add the example you described in the above [comment](https://github.com/teemtee/tmt/pull/3796#discussion_r2164679264) in the documentation? I was having a hard time understanding what the process and outcome was supposed to be, but this thread cleared a lot of that up. It doesn't seem like the end-to-end example you described above is represented in the documentation code. 

> For example. If we provide this layout:
> [...]
> with this plan:
> [...]
> this inventory file is generated:
> [...]


Specifically this helped me to connect the dots.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,44 @@
+summary: Define general Ansible settings for a plan
+
+story:
```
happz: Should be marked as a multiline string, `|`, same as `description` below.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,44 @@
...
+
+    # path/to/the/layout.yaml
+    all:
+      children:
+        server:
+          children:
+            master: {}
+            replica: {}
+        client: {}
+        runner: {}
```
happz: I would definitely include the eventual outcome in this example. Ok, I see what keys I need to add to my plan, and how the structure looks like in the layout template, can I also see the actual inventory file produced by these pieces?

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,44 @@
...
+    execute:
+      how: tmt
+
+    # path/to/the/layout.yaml
+    all:
+      children:
+        server:
+          children:
+            master: {}
+            replica: {}
```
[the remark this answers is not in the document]
happz: @dav-pascual can you add a comment why this was resolved? Because I would have the same question, and I see no answer that would help me understand what is expected from `replica` alongside of `master`.

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
       - implemented-by: /tmt/steps/__init__.py
       - implemented-by: /tmt/queue.py
       - verified-by: /tests/multihost
+
+/ansible:
+    summary: Per-guest Ansible customization
+
+    story:
```
happz: A multiline string, needs `|` (yeah, it's a common issue in our fmf files...)

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+        for a provisioned guest to be reflected in the generated inventory.
+
+    description: |
+        Each provisioned guest may define an optional ``ansible`` key
+        to define its corresponding Ansible inventory section.
+
+        - ``group``: Assigns the guest to an Ansible specific group. If omitted,
+          the ``role`` key would serve as the Ansible group name.
+          If ``role`` is not set either, the host will be placed in the
+          'ungrouped' group.
```
happz: Please, include link to https://tmt.readthedocs.io/en/stable/plugins/provision.html#common-keys` so users can transition from here to the documentation of `role` key (plus, `ungrouped` is marked as a literal name, but we should use backticks for such highlight):

```
the :ref:`role </plugins/provision/common-keys>` key would serve as the Ansible group name. If ``role`` is not set either, the host will be placed in the ``ungrouped`` group.
```

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+
+        - ``vars``: Defines host-specific Ansible variables to include
+          under that host in the inventory.
+
+    example: |
+        provision:
+          - name: master
+            role: server
+            ansible:
+              group: server
```
happz: I would definitely use different value here, because when it's the same as `role`, what's the point of setting `group` explicitly?

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+              vars:
+                var1: value1
+                var2: value2
+
+          - name: client
+            role: client
+            ansible:
+              vars:
+                ansible_user: fedora
+                ansible_port: 2222
```
happz: Please, include also a dummy inventory outcome, to demonstrate how the `group` and `vars` propagate.

happz: Also, I would recommend adding a guest without `role` or `group`, to show how it lands in `ungrouped` group.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,44 @@
...
+    users to specify a YAML template file defining the desired inventory group
+    hierarchy and layout.
+
+    This layout will be used to organize provisioned guests under
+    Ansible groups accordingly. If no layout is provided, a simple
+    default basic layout is used.
+
+example:
+  - |
+    # sample_plan.fmf
```
psss:
```suggestion
    # sample-plan.fmf
```

Nitpick, in general we're using dashes for separating words in filenames.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,44 @@
...
+    ansible:
+      inventory:
+        layout: path/to/the/layout.yaml
```
psss: Let's do 4-space indentation to keep the example consistent with the rest of the specification.

spec/plans/ansible.fmf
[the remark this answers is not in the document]
psss: Yeap, I just came here with the same question: `replica` included in the layout but not defined under the `provision` step :-)

spec/plans/ansible.fmf
psss: Yes, that would help a lot. I was also wondering to myself: Now what the resulting file would look like?

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+    description: |
+        Each provisioned guest may define an optional ``ansible`` key
+        to define its corresponding Ansible inventory section.
+
+        - ``group``: Assigns the guest to an Ansible specific group. If omitted,
+          the ``role`` key would serve as the Ansible group name.
+          If ``role`` is not set either, the host will be placed in the
+          'ungrouped' group.
+
+        - ``vars``: Defines host-specific Ansible variables to include
```
psss:
```suggestion
        - ``variables``: Defines host-specific Ansible variables to include
```

If possible, I would recommend to avoid using abbreviations. Or is this to keep consistency with `ansible` itself?

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+
+    story:
+        As a user I want to assign an Ansible group and host variables
+        for a provisioned guest to be reflected in the generated inventory.
+
+    description: |
+        Each provisioned guest may define an optional ``ansible`` key
+        to define its corresponding Ansible inventory section.
+
+        - ``group``: Assigns the guest to an Ansible specific group. If omitted,
```
psss: What about using a definition list instead of the bullet points?

```rst
group
    Assigns the guest to an Ansible specific group. If omitted,
    the role key would serve as the Ansible group name. If role is
    not set either, the host will be placed in the ‘ungrouped’
    group.
```

spec/plans/provision.fmf
```diff
@@ -75,3 +75,39 @@ link:
...
+
+    example: |
+        provision:
+          - name: master
+            role: server
+            ansible:
+              group: server
+              vars:
+                var1: value1
+                var2: value2
```
psss: 4-space indent, please.

spec/plans/provision.fmf
dav-pascual: Yes, in Ansible they are called vars, so I kept the same name. Would you still prefer having ´variables´ instead?

dav-pascual:
@psss @happz @tcornell-bus I have expanded current example in ansible.fmf file with your feedback. Hopefully, it's more clear now.

spec/plans/ansible.fmf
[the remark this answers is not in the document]
dav-pascual: Making a point that even if a group is defined in a layout, it doesn't mean it'll have hosts under it (the opposite case to having a host with no group). The same layout file can be used consistently for many plans. The important thing in this usecase is the group hierarchy.

Any suggestion to make this less confusing? Or would you prefer to have it removed?

psss:
One hint from the [Pull Requests](https://tmt.readthedocs.io/en/stable/contribute.html#pull-requests) section of the contribution guide:

> During the pull request review it is recommended to add new commits with your changes on the top of the branch instead of amending the original commit and doing a force push. This will make it easier for the reviewers to see what has recently changed.

spec/plans/ansible.fmf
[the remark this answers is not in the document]
psss: I see, understood. Perhaps adding a short comment into the resulting `inventory.yaml` would help? Perhaps something like this?

    # An empty group is created if no matching guests are found

spec/plans/provision.fmf
psss: Ok, then let's stick here with what is natural for ansible.

spec/plans/provision.fmf
psss: I don't think this is covered. The example `inventory.yaml` output for the per-guest `ansible` key is missing.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,93 @@
...
+example:
+  - |
+    # sample-plan.fmf (input fmf file)
+    ansible:
+        inventory:
+            layout: path/to/the/layout.yaml
+
+    provision:
+        - name: test-master
+          ansible:
```
psss: I'm not completely sure it's a good idea to mix both `plan.ansible` and `provision.ansible` keys in a single example. It's a bit confusing to understand what affects what. Especially here, in the individual key specification. Wouldn't it be better to clearly separate the two concepts? I believe users will be able to connect the dots. Maybe later we could include the combined example in the guide with links to both specifications. What do you think?

psss:
I'm proposing a couple of minor enhancements in e1f08e2e. @dav-pascual, does it look good to you?

spec/plans/ansible.fmf
dav-pascual: Sounds good. Let's expand later with a complete example in the guide

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,93 @@
...
+              group: master
+              vars:
+                  var1: value1
+                  var2: value2
```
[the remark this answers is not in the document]
dav-pascual: We'll add this edge case in guide later as mentioned above

spec/plans/provision.fmf
dav-pascual: added

dav-pascual:
> I'm proposing a couple of minor enhancements in [e1f08e2](https://github.com/teemtee/tmt/commit/e1f08e2e276f9121804a717481f41ceed4751545). @dav-pascual, does it look good to you?

Looks good. Updated with the enhancements

tcornell-bus (approved):
The examples are very clear to me now. :slightly_smiling_face:

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,87 @@
...
+          role: runner
+        - name: test-client
+          role: client
+        - name: sidecar-vm  # if no group is specified, the guest is added to 'ungrouped'
+
+    execute:
+        how: tmt
+
+  - |
+    # path/to/the/layout.yaml (input inventory layout, relative to fmf file)
```
happz: Shouldn't the path be relative to fmf root? I think we discussed it somewhere, wasn't that the conclusion? It would be aligned with Ansible playbooks.

spec/plans/ansible.fmf
```diff
@@ -0,0 +1,87 @@
...
+                children:
+                    server:
+                        children:
+                            master: {}
+                            replica: {}
+                    client: {}
+                    runner: {}
+
+  - |
+    # path/to/run/artifacts/provision/inventory.yaml (output inventory file)
```
happz: Path itself is probably not interesting enough to be mentioned, "Output inventory file" should be enough.

spec/plans/ansible.fmf
dav-pascual: My bad. I mix it up. Fixed

spec/plans/ansible.fmf
dav-pascual: fixed

psss (approved):
Thanks for the updates, now looks good!

spec/plans/ansible.fmf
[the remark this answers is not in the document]
psss: Maybe just adding this (or similar) comment to make the `replica` example completely clear?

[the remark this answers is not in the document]
dav-pascual: Done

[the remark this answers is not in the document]
psss: Thanks!

psss:
Spec only change, core tests passed, failing job are irrelevant errors, merging.

PR #3798, opened 2025-06-06
Polish provision/connect and provision/local docs

happz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -178,6 +178,9 @@ SSH_MASTER_SOCKET_MIN_HASH_LENGTH = 4
 #: filename.
 SSH_MASTER_SOCKET_MAX_HASH_LENGTH = 64
 
+#: Default username to use in SSH connections.
+DEFAULT_USER = 'root'
+
 
 @overload
 def _socket_path_trivial(
@@ -2027,14 +2030,11 @@ class GuestSshData(GuestData):
              """,
         normalize=tmt.utils.normalize_optional_int,
     )
-    user: Optional[str] = field(
-        default=None,
+    user: str = field(
+        default=DEFAULT_USER,
         option=('-u', '--user'),
         metavar='NAME',
-        help="""
-             Username to use for SSH connections instead of the account
-             name running tmt.
-             """,
+        help='A username to use for all guest operations.',
     )
     key: list[Path] = field(
         default_factory=list,
--- a/tmt/steps/provision/connect.py
+++ b/tmt/steps/provision/connect.py
@@ -8,8 +8,6 @@ from tmt.container import container, field
 from tmt.utils import Command, ShellScript
 from tmt.utils.wait import Waiting
 
-DEFAULT_USER = "root"
-
 
 @container
 class ConnectGuestData(tmt.steps.provision.GuestSshData):
@@ -24,13 +22,6 @@ class ConnectGuestData(tmt.steps.provision.GuestSshData):
         option=('-g', '--guest'),
         metavar='HOSTNAME|IP',
         help='A preexisting machine to connect to.',
-    )
-
-    user: str = field(
-        default=DEFAULT_USER,
-        option=('-u', '--user'),
-        metavar='NAME',
-        help='A username to use for all guest operations.',
     )
 
     soft_reboot: Optional[ShellScript] = field(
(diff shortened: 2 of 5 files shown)


tmt/steps/provision/connect.py
```diff
@@ -172,11 +172,21 @@ def start(self) -> None:
...
+    # https://tmt.readthedocs.io/en/stable/contribute.html#docs
+    #
     """
     Connect to a provisioned guest using SSH.
 
-    Do not provision a new system. Instead, use provided
-    authentication data to connect to a running machine.
+    Do not provision any system, tests will be executed directly on the
+    machine that has been already provisioned. Use provided
+    authentication information to connect to it over SSH.
```
mcasquer, line 178: So SSH in capitals shouldn't be a literal? such as `ssh`in: https://github.com/teemtee/tmt/blob/8a3ec52a5478c40670ce4edd71eabc4bbb1b8e96/tmt/steps/provision/__init__.py#L2038

tmt/steps/provision/connect.py
```diff
@@ -221,6 +233,12 @@ class ProvisionConnect(tmt.steps.provision.ProvisionPlugin[ProvisionConnectData]
           hard-reboot: virsh reboot my-example-vm
           soft-reboot: ssh root@my-example-vm 'shutdown -r now'
 
+    .. code-block:: shell
```
mcasquer, line 227: Missing the precedent "On command-line:" used for other blocks like this?

tmt/steps/provision/local.py
```diff
@@ -184,34 +184,49 @@ def pull(
...
 
     .. code-block:: yaml
 
         provision:
             how: local
 
-    Note that ``tmt run`` is expected to be executed under a regular user.
-    If there are admin rights required (for example in the prepare step)
-    you might be asked for a ``sudo`` password.
+    .. code-block:: shell
```
mcasquer, line 225: Same here, another  "On command-line:"  to separate the blocks?

tmt/steps/provision/connect.py
happz: Nope. "SSH" is a protocol name, acronym even, like TCP or URL, on the other hand, "ssh" and "ssh_config" are literally names of command and a man page.

tmt/steps/provision/connect.py
happz: That one is added by docs rendering of individual keys, in plugin docs it's not needed.

tmt/steps/provision/connect.py
mcasquer: Fair enough

tmt/steps/provision/connect.py
mcasquer: Got it !

therazix (approved):

tmt/steps/provision/__init__.py
```diff
@@ -1990,36 +1995,48 @@ class GuestSshData(GuestData):
...
     )
     user: Optional[str] = field(
         default=None,
         option=('-u', '--user'),
-        metavar='USERNAME',
-        help='Username to use for all guest operations.',
+        metavar='NAME',
+        help="""
+             Username to use for SSH connections instead of the account
+             name running tmt.
```
psss: By default we're using `root` for basically all provision plugins. What about mentioning this here and removing the duplicated definitions from the individual plugins inheriting from `GuestSshData`?

https://github.com/teemtee/tmt/blob/f413e79d71892f2deacc0d937577799aa3b34e54/tmt/steps/provision/connect.py#L29-L34

https://github.com/teemtee/tmt/blob/f413e79d71892f2deacc0d937577799aa3b34e54/tmt/steps/provision/testcloud.py#L327-L332

psss (commented):
Nice! Just one suggestion.

tmt/steps/provision/__init__.py
happz: Indeed, every single plugin has its own default, and it's always `root`. Need to check 1mt plugin, but reducing the boilerplate is good.

happz: Addressed in e032f147

psss: Nice, thanks!

psss (approved):

happz:
Unrelated failure, merging.

PR #3799, opened 2025-06-06
Allow dotenv files from `--environment`

Lorquas:
Pull Request Checklist

* [x] implement the feature
* [x] adjust docstring
* [ ] mention the version

According to the man/help both yaml and dotenv are supported:
```
$ tmt --version
tmt version: 1.49.0
$ tmt run --help 
...
  -e, --environment KEY=VALUE|@FILE
                                  Set environment variable. Can be specified
                                  multiple times. The "@" prefix marks a file
                                  to load (yaml or dotenv formats
                                  supported).
```
however it only accepts YAML:
```
$ echo MYVAR=foobar > test.env
$ tmt -vvv run -a --environment @test.env --dry
/var/tmp/tmt/run-008

Failed to extract variables from YAML format, YAML defining variables must be a dictionary.
$ echo $?
2
```

This PR fixes this bug.


therazix (approved):
LGTM. Works as expected.

tcornell-bus (changes requested):
A quick test added would be really helpful to make sure .env files stay supported. I think there is already tests for .yaml here `tests/core/env`. Maybe a simple .env test can be added there as well?

Otherwise works as expect for me!

happz:
Indeed, a test would be nice. @Lorquas would you be able to extend `/tests/core/env` to cover also dotenv files?

tests/core/env/test.sh
```diff
@@ -56,6 +56,20 @@ rlJournalStart
...
+                        plan --name $plan \
+                        test --name $test"
+                    rlAssertGrep '>>>O0<<<' $rlRun_LOG
+                done
+            done
+        rlPhaseEnd
+
         rlPhaseStartTest "Empty environment file ($execute)"
             rlRun -s "tmt run -r -e @empty.yaml"
             rlAssertGrep "warn: Empty environment file" $rlRun_LOG
```
thrix: @Lorquas also please add a test for empty environment file

Lorquas: done!

tests/core/env/test.sh
```diff
@@ -56,10 +56,29 @@ rlJournalStart
...
+                    rlRun -s "tmt run -avvvr -e @vars.env \
+                        execute --how $execute \
+                        plan --name $plan \
+                        test --name $test"
+                    rlAssertGrep '>>>O0<<<' $rlRun_LOG
+                done
+            done
+        rlPhaseEnd
+
         rlPhaseStartTest "Empty environment file ($execute)"
```
thrix:
```suggestion
        rlPhaseStartTest "Empty YAML environment file ($execute)"
```

tests/core/env/test.sh
```diff
@@ -56,10 +56,29 @@ rlJournalStart
...
             rlRun -s "tmt run -r -e @empty.yaml"
             rlAssertGrep "warn: Empty environment file" $rlRun_LOG
         rlPhaseEnd
+
+        rlPhaseStartTest "Empty environment file ($execute)"
```
thrix:
```suggestion
        rlPhaseStartTest "Empty DOTENV environment file ($execute)"
```

PR #3810, opened 2025-06-11
Document handling of variants in hardware requirements

### LINKED ISSUE #929: HW requirements, `and`, `or` and expected outcome
Imagine following `provision.hardware` setup:

```yaml
hardware:
  or:
    - disk:
        - size: >= 20 GiB
        - size: >= 512 GiB
        - size: >= 512 GiB
    - disk:
        - size: >= 20 GiB
        - size: >= 1024 GiB
```

Some provisioning services, like OpenStack and AWS, come with instance types that have a relatively rigid root disk set up, but besides that one, it's possible to create additional "disks" with quite a precise size, and a number of these additional volumes can go to low tens. There are also provisioners like Vagrant that support even more freedom, but the principle is the same, it's easy to add more volumes with their size and type under control.

It is, therefore, possible to satisfy both branches of `hardware.or` directive from the example above, with the comparable amount of work - both would use the same instance type, providing the root disk, with an additional volume or two, respectively. Which one of these would you expect to be the final answer? Is there some order of precedence that should be followed?

Also, it's possible to construct constraints like this one:

```yaml
hardware:
  or:
    - disk:
        - size: >= 20 GiB
        - size: >= 512 GiB
        - size: >= 512 GiB
    - disk:
        - size: >= 20 GiB
        - size: >= 1024 GiB
    - cpu:
          model: 6
```

A reduced list of suitable instance types would consist of both types that can hold additional volumes, but also instance types that cannot but do have matching CPU. What would be your expectation in this case?

* For example, should less work be the rule here, picking virtually any instance type with `cpu.model == 6`, rather than those that can hold additional volumes, and avoid "wasting" time creating those volumes?
* Or some kind of "resource usage" weight to be applied, picking an instance type that uses the lowest amount of resources (2 volumes vs 3 volumes, but with the same combined size, or an instance type with just the root disk and matching CPU)?
* Also, how about "any of the matching types is a valid answer"? I guess it could be argued that all matching instance types are valid, provisioning is then free to pick one, no questions asked, and if the user is surprised by the final VM setup, it's their fault and they should refine the content of `hardware` in their plan.

I admit, both examples are slightly on the extreme side of things, but having clear guidelines for these situations would help a lot.

happz:
Fixes #929.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/spec/hardware/main.fmf
+++ b/spec/hardware/main.fmf
@@ -194,6 +194,52 @@ description: |
     the only one in existence, and guest architecture may allow
     for additional device and vendor ID schemas.
 
+    Priority of variants
+    --------------------
+
+    When the ``hardware`` field can be satisfied by more than way,
+    i.e. allowing multiple correct answers - variants. The actual guest
+    hardware configuration depends on the behavior of the ``provision``
+    plugin used. Consider the following examples:
+
+    .. code-block:: yaml
+
+        hardware:
+          memory: '>= 4 GB'
+
+        hardware:
+          or:
+            - memory: '>= 4 GB'
+            - memory: '< 4 GB'
+
+    The first example is simple: there is only one way a potential guest
+    can satisfy it, i.e. by having at least 4 GB of RAM, and no other
+    hardware configuration would be suitable.
+
+    The second example offers two ways a guest can can satisfy the
+    requirements: by having at least 4 GB of RAM, or having strictly
+    less than 4 GB of RAM. Both answers would be correct. Handling of
+    these variants splits :ref:`/plugins/provision` into two groups:
+
+    * Plugins that construct guests from the requirements:
+      :ref:`/plugins/provision/virtual.testcloud`
+
+      These plugins will take the first variant - ``memory: '>= 4 GB'``
+      - and they will construct the guest to match this subset of rules.
+      Other variants would be ignored.
+    * Plugins that select guests from a preexisting inventory of
+      machines: :ref:`/plugins/provision/beaker`
+
+      These plugins are capable of finding a machine that fits either of
+      the variants, and the guest may have either at least 4 GB or RAM,
+      or strictly less.
+
+    .. note::
+
+        :ref:`/plugins/provision/artemis` is a special case that may
+        support both ways, depending on the actual pool selected for the
+        provisioning.
+
     Notes
     -----
 


spec/hardware/main.fmf
```diff
@@ -194,6 +194,52 @@ description: |
...
+        hardware:
+          or:
+            - memory: '>= 4 GB'
+            - memory: '< 4 GB'
+
+    The first example is simple: there is only one way a potential guest
+    can satisfy it, i.e. by having at least 4 GB of RAM, and no other
+    hardware configuration would be suitable.
+
+    The second example offers two ways a guest can satisfy the
```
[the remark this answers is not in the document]
happz: Not anymore.

spec/hardware/main.fmf
```diff
@@ -194,6 +194,52 @@ description: |
...
+    requirements: by having at least 4 GB of RAM, or having strictly
+    less than 4 GB of RAM. Both answers would be correct. Handling of
+    these variants splits :ref:`/plugins/provision` into two groups:
+
+    * Plugins that construct guests from the requirements:
+      :ref:`/plugins/provision/virtual.testcloud`
+
+      These plugins will take the first variant - ``memory: '>= 4 GB'``
+      - and they will construct the guest to match this subset of rules.
+      Other variants would be ignored.
```
tcornell-bus: Should `provision --virtual` behave this way already? Or is this documentation for a feature to come later?

happz: It should behave this way already: `variant = self.hardware.constraint.variant()` (https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/testcloud.py#L944) -> https://github.com/teemtee/tmt/blob/main/tmt/hardware.py#L388. The plugin asks the HW-related code to return the first of all known variants, and ignores the rest.

tcornell-bus: I tried to check whether this behavior is already there for `virtual` using the above memory example, and I noticed this warning that might be misleading:
```
$ tmt run --all provision --how virtual --image fedora plan -n all
[...]
    provision
        
        how: virtual
        hardware:
            or:
              - memory: '>= 4 GB'
              - memory: < 4 GB
        memory: 2048 MB
        disk: 40 GB
        warn: Hardware requirement 'memory' is not supported.
        progress: booting...
```
This makes me think that the `memory` requirement is being ignored and it was created with 2048 MB memory.

But when I turn on debug and verbose:
```
$ tmt -v --debug run --all provision --how virtual --image fedora plan -n all
[...]
    provision
    workdir: /var/tmp/tmt/run-026/plans/all/provision
        
        how: virtual
        hardware:
            or:
              - memory: '>= 4 GB'
              - memory: < 4 GB
        memory: 2048 MB
        disk: 40 GB
        warn: Hardware requirement 'memory' is not supported.
        Workdir '/var/tmp/tmt/run-026/plans/all/provision/default-0' created.
        testcloud version: 0.11.7
        effective hardware:
            or:
              - memory: >= 4 GB
              - memory: < 4 GB
        final domain memory: 4000000
        final domain root disk size: 40
        final domain disk #0 size: 40
        name: tmt-026-wFmdYxSm
        Generating an ssh key.
        key: /var/tmp/tmt/run-026/plans/all/provision/default-0/id_ecdsa
        progress: booting...
```
The `final domain memory: 4000000` tells me it actually did follow the hardware requirement.
Should I create an issue about emitting warnings that aren't true? Unless I am doing this wrong of course.

tcornell-bus: In addition to the above possible issue, when I swapped the memory requirements to see if it picks the first every time, I got an operator not supported error:
```
$ tmt -v --debug run --all provision --how virtual --image fedora plan -n all
[...]
    provision
    workdir: /var/tmp/tmt/run-025/plans/all/provision
        
        how: virtual
        hardware:
            or:
              - memory: < 4 GB
              - memory: '>= 4 GB'
        memory: 2048 MB
        disk: 40 GB
        warn: Hardware requirement 'memory' is not supported.
        Workdir '/var/tmp/tmt/run-025/plans/all/provision/default-0' created.
        testcloud version: 0.11.7
        effective hardware:
            or:
              - memory: < 4 GB
              - memory: >= 4 GB
        fail: Cannot apply hardware requirement 'memory: < 4 GB', operator not supported.
```
I am just bringing this up in case the example in the documentation should be adjusted. Or if this syntax should be supported I can make an issue.

happz: Full of surprises :O

> ```
> $ tmt run --all provision --how virtual --image fedora plan -n all
> [...]
>     provision
>         queued provision.provision task #1: default-0
>         
>         provision.provision task #1: default-0
>         how: virtual
>         hardware:
>             or:
>               - memory: '>= 4 GB'
>               - memory: < 4 GB
>         memory: 2048 MB
>         disk: 40 GB
>         warn: Hardware requirement 'memory' is not supported.
>         progress: booting...
> ```
> 
> This makes me think that the `memory` requirement is being ignored and it was created with 2048 MB memory.

`memory` should be supported, no idea why the warning is emitted :O

> ```
> $ tmt -v --debug run --all provision --how virtual --image fedora plan -n all
> [...]
>     provision
>     workdir: /var/tmp/tmt/run-026/plans/all/provision
>         queued provision.provision task #1: default-0
>         
>         provision.provision task #1: default-0
>         how: virtual
>         hardware:
>             or:
>               - memory: '>= 4 GB'
>               - memory: < 4 GB
>         memory: 2048 MB
>         disk: 40 GB
>         warn: Hardware requirement 'memory' is not supported.
>         Workdir '/var/tmp/tmt/run-026/plans/all/provision/default-0' created.
>         testcloud version: 0.11.7
>         qcow: Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2
>         effective hardware:
>             or:
>               - memory: >= 4 GB
>               - memory: < 4 GB
>         final domain memory: 4000000
>         final domain root disk size: 40
>         final domain disk #0 size: 40
>         name: tmt-026-wFmdYxSm
>         Generating an ssh key.
>         key: /var/tmp/tmt/run-026/plans/all/provision/default-0/id_ecdsa
>         progress: booting...
> ```
> 
> The `final domain memory: 4000000` tells me it actually did follow the hardware requirement.

Exactly, it works, and I see no reason for the warning.

> Unless I am doing this wrong of course.

Nope, it looks legit, the warning is indeed incorrect.

> In addition to the above possible issue, when I swapped the memory requirements to see if it picks the first every time, I got an operator not supported error:
> 
> ```
> $ tmt -v --debug run --all provision --how virtual --image fedora plan -n all
> [...]
>     provision
>     workdir: /var/tmp/tmt/run-025/plans/all/provision
>         queued provision.provision task #1: default-0
>         
>         provision.provision task #1: default-0
>         how: virtual
>         hardware:
>             or:
>               - memory: < 4 GB
>               - memory: '>= 4 GB'
>         memory: 2048 MB
>         disk: 40 GB
>         warn: Hardware requirement 'memory' is not supported.
>         Workdir '/var/tmp/tmt/run-025/plans/all/provision/default-0' created.
>         testcloud version: 0.11.7
>         qcow: Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2
>         effective hardware:
>             or:
>               - memory: < 4 GB
>               - memory: >= 4 GB
>         fail: Cannot apply hardware requirement 'memory: < 4 GB', operator not supported.
> ```
> 
> I am just bringing this up in case the example in the documentation should be adjusted. Or if this syntax should be supported I can make an issue.

This one is actually correct, `>` is not supported. But it can be, I think it's just laziness, `> N` can be easily satisfied by getting a VM with (N + 1 byte) of RAM, right? Feel free to fix that, check `_report_hw_requirement_support()` and `_apply_hw_memory()`.

Aaah, and now the warning makes sense: it's not that `memory` is not supported, `memory  > N` is not, the operator - check `_report_hw_requirement_support()`, it's right there :) So it's two issues, one extending support to new operators - if you do that, please update `link` in `/spec/hardware/memory.fmf` - and the second is that there could be a better warning message emitted by `Hardware.report_support()`, instead of `printable_name` it could show the whole constraint, including the value and operator and giving user better picture.

tcornell-bus: Made issues https://github.com/teemtee/tmt/issues/3812 and https://github.com/teemtee/tmt/issues/3813 :smiley:

happz:
@coderabbitai full review

psss (approved):
Nicely written! Thanks.

psss:
Core tests passed, rawhide failure is irrelevant, merging.

PR #3816, opened 2025-06-16
Do not run tests against `fedora-rawhide`

psss:
Several times already we've run into issues which are out of our scope, breakages we often cannot influence and they are just slowing us down. Let's skip rawhide builds for now.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -22,7 +22,7 @@ actions: &base-actions
     - hatch version
 
 targets: &all-targets
-  - fedora-all
+  - fedora-stable
   - epel-9
 
 # Uncomment below line if OpenScanHub scans are failing


psss:
The current example is the broken `python-pydantic`:

* https://bugzilla.redhat.com/show_bug.cgi?id=2372054

Bloked by the upstream issue:

* https://github.com/pydantic/pydantic/issues/11613

Unclear when the fix will be available.

.packit.yaml
```diff
@@ -22,7 +22,7 @@ actions: &base-actions
     - hatch version
 
 targets: &all-targets
```
psss, line 24: Hmmm... just realized that `all-targets` is used for `propose_downstream` and `koji_build` jobs as well. This would mean that we would not get rawhide pull requests created for the dist git pagure.

bajertom (approved):

.packit.yaml
psss: Agreed on the hacking session to disable both Rawhide testing and building for now.

PR #3817, opened 2025-06-16
Document the `discover` step key `test`

### LINKED ISSUE #3794: Document the behaviour of the `discover` key `test`

psss:
Add a new section and examples demonstrating how the `test` key of the `fmf` discover step can be used to filter tests. Provide an example showing how a list of regular expressions can be used to execute a single test multiple times.

Fix #3794.

Pull Request Checklist

* [x] write the documentation

### DIFF
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -344,11 +344,11 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
         discover:
             how: fmf
             test:
-              - /test/one
-              - /test/two
-              - /special/setup
-              - /test/one
-              - /test/two
+              - ^/test/one$
+              - ^/test/two$
+              - ^/special/setup$
+              - ^/test/one$
+              - ^/test/two$
 
     Link Filter
     ^^^^^^^^^^^
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -94,7 +94,9 @@ class DiscoverFmfStepData(tmt.steps.discover.DiscoverStepData):
             List of test names or regular expressions used to
             select tests by name. Duplicate test names are allowed
             to enable repetitive test execution, preserving the
-            listed test order.
+            listed test order. The search mode is used for pattern
+            matching. See the :ref:`regular-expressions` section for
+            details.
             """,
         normalize=tmt.utils.normalize_string_list,
     )
@@ -356,7 +358,11 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
 
     .. code-block:: shell
 
-        tmt run discover -h fmf -v -t ^/test/one$ -t ^/special/setup$ -t ^/test/two$
+        tmt run discover -h fmf -v -t '^/test/one$' -t '^/special/setup$' -t '^/test/two$'
+
+    The ``test`` key uses search mode for matching patterns. See the
+    :ref:`regular-expressions` section for detailed information about
+    how exactly the regular expressions are handled.
 
     Link Filter
     ^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


tmt/steps/discover/fmf.py
```diff
@@ -320,6 +323,36 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
...
+    .. code-block:: yaml
+
+        discover:
+            how: fmf
+            test:
+              - /test/one
+              - /test/two
+              - /special/setup
+              - /test/one
+              - /test/two
```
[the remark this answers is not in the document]
psss: Ok, makes sense, added in 1c61217a.

tmt/steps/discover/fmf.py
```diff
@@ -320,6 +323,36 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
...
+    .. code-block:: yaml
+
+        discover:
+            how: fmf
+            test:
+              - ^/test/one$
+              - ^/test/two$
+              - ^/special/setup$
+              - ^/test/one$
+              - ^/test/two$
```
happz, line 357: Please, add a CLI example as well, if it's supported - see similar parts of `prepare/ansible`, e.g. https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/ansible.py#L133

psss: Good point, covered by 24a91660.

happz: Isn't `$` going to be a problem on command line, not being wrapped with single-quotes? I would never dare to use `-t ^/test/one$`, out of the fear of Bash gods I would stick to `-t '^/test/one$'` - is the former actually working?

mcasquer (approved):
Agreed with Milos review, others LGTM

tmt/steps/discover/fmf.py
bajertom: It is working both with and without single-quotes on command line. However if encapsulating it in single-quotes is safer due to $ being in there, I would opt for this. Otherwise LGTM.

tmt/steps/discover/fmf.py
```diff
@@ -320,6 +323,44 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
...
+    ^^^^^^^^^^^
+
+    Use the ``test`` key to limit which tests should be executed by
+    providing regular expression matching the test name:
+
+    .. code-block:: yaml
+
+        discover:
+            how: fmf
+            test: ^/tests/area/.*
```
thrix, line 338: Could we mention in a note what is used for matching? I assume it is `re.search`?

thrix (changes requested):

tmt/steps/discover/fmf.py
happz: Such info should be mentioned in the key help text, together with a reference to 

```
:ref:`regular-expressions`
```

It's how regular expressions should be documented - a key foo accepts regular expressions, uses "match" or "search" mode, see details on this page.

tmt/steps/discover/fmf.py
psss: > Isn't `$` going to be a problem on command line, not being wrapped with single-quotes?

Should not be, if there's a space after the `$`. But let's make it double-safe with quotes. Done in 76820c60.

tmt/steps/discover/fmf.py
psss: Yeah, good point. Added in 76820c60.

tmt/steps/discover/fmf.py
```diff
@@ -320,6 +325,48 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
...
+            test:
+              - ^/test/one$
+              - ^/test/two$
+              - ^/special/setup$
+              - ^/test/one$
+              - ^/test/two$
+
+    .. code-block:: shell
+
+        tmt run discover -h fmf -v -t '^/test/one$' -t '^/special/setup$' -t '^/test/two$'
```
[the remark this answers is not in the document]
psss: It would be too long, shortened intentionally.

bajertom (approved):

happz (approved):

psss:
Failure irrelevant, merging.

PR #3824, opened 2025-06-18
Use shell scripts proper for shell constructs in guest fact probes

happz:
Some probes were using `DEFAULT_SHELL` directly - not needed, `ShellScript.to_shell_command()` does the conversion nicely - and at least one command apparently did not work correctly without shell involved (is-container probe).

Pull Request Checklist

* [x] implement the feature


tcornell-bus (approved):

psss (approved):
Looks good, thanks!

Issue #902, opened 2021-09-23
Support reboot from shell during `prepare` and `finish`


FrNecas:
In one of the discussions around reboot, it was mentioned that it may be nice to also support it not only in execute -h internal but also in prepare -h shell. @psss mentioned that the same effect should be achieved using ansible and its reboot feature but it could be useful to provide the same thing for shell preparation (more options are nice, aren't they?). I see it as a nice-to-have, should not be hard to implement (logic can be reused) but not a priority.

```
prepare:
  - name: Prepare the system and reboot
    how: shell
    script: |
      if [ -z "$REBOOT_COUNT" ] || [ "$REBOOT_COUNT" -eq 0 ]; then
          dnf install -y something
          touch /etc/some-file
          tmt-reboot
      fi
```

lkotek:
@FrNecas Hi, I would like to take this issue if possible. I was looking for something to start with. This functionality is something I have a use for and it is already marked like good first issue :-)

FrNecas:
Sure, I just hope that I estimated it correctly to be an easy issue and that no problems arise :) I think that generalising the logic for reboot handling a bit and moving it from `execute/internal.py` elsewhere (maybe introducing a new class -- `PluginSupportingReboot` or something along those lines) should suffice. Also do not forget about the finish step as mentioned in #1351 . Feel free to ping me if you need any introduction to tmt's code base or any help with the issue @lkotek

lkotek:
> Sure, I just hope that I estimated it correctly to be an easy issue and that no problems arise :) I think that generalising the logic for reboot handling a bit and moving it from `execute/internal.py` elsewhere (maybe introducing a new class -- `PluginSupportingReboot` or something along those lines) should suffice. Also do not forget about the finish step as mentioned in #1351 . Feel free to ping me if you need any introduction to tmt's code base or any help with the issue @lkotek

@FrNecas Thanks for sharing the ideas! I will let you know if there is anything 'unexpected' I need to help with :)

lkotek:
Hi @FrNecas, I prepared PR containing the changes for support of reboot in case of prepare and finish steps. I extracted the functionality of the key methods responsible for reboot (the most important `_handle_reboot`), modified them a bit, so it was possible to use it in prepare steps and finally put them into new class ``CommonReboot``. Regarding support in finish step I used the same approach we currently use for ansible step (class from the prepare step is completely reused). 

[...]

PR #3832, opened 2025-06-23
Move tmt scripts out of the execute step

happz:
As a preparation for #902, moving tmt scripts - tmt-reboot, tmt-abort and so on - outside of the `execute` step, as we realize the user-triggered reboot needs to be supported in other steps as well.

Pull Request Checklist

* [x] implement the feature


therazix (approved):

tcornell-bus (approved):
It looks to me like all these scripts are covered in some form or another by tests, which all pass. LGTM!

PR #3833, opened 2025-06-23
Prevent tmt scripts from running outside of test context

happz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/scripts/tmt-abort
+++ b/tmt/steps/scripts/tmt-abort
@@ -1,4 +1,10 @@
 #!/bin/bash
+
+if [ -z "$TMT_TEST_PIDFILE" ]; then
+    echo "tmt-abort can be used only in the context of a running test."
+    exit 1
+fi
+
 # Create the abort file.
 touch "$TMT_TEST_DATA/abort"
 # Kill PPID to exit test execution.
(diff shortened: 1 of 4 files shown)


thrix (approved):
Seems like a no brainer change to me.

@happz please add it to release notes, it should be there in case somebody would be using it outside the `tmt-reboot` of the context now for some custom workflows. We used that before in reservations (we do not anymore), anyway, rather inform users about this change.

happz:
> Seems like a no brainer change to me.
> 
> @happz please add it to release notes, it should be there in case somebody would be using it outside the `tmt-reboot` of the context now for some custom workflows.

Such a custom workflow would need to set environment variables `tmt-reboot` consumes, and that is what makes it the "test context". So such a user should be perfectly fine. So there should be no group that would be using `tmt-reboot` from outside *without* pretending it's called under the hood of tmt; everyone else should see no difference.

The change should prevent this group to appear - by uploading scripts to guest in `provision`, they are now visible to `prepare` scripts, and I want them to crash and burn if anyone calls them from `prepare` shell script. It was not possible before, as scripts did not exist, they do no and I just want to prevent that.

Does it still need a release note? There can't be a user today, and with the change, new one won't appear.

psss (approved):
Yeah, makes sense, thanks!

happz:
Unrelated Rawhide issue, merging.

PR #3835, opened 2025-06-23
Refactor plan environment composition

happz:
* clean separation of various environment sources,
* more fitting names for the corresponding methods,
* fix double expansion of node keys of imported plans.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/policy/test.sh
+++ b/tests/policy/test.sh
@@ -14,7 +14,9 @@ rlJournalStart
             local filter="$3"
             local expected="$4"
 
-            rlRun    "tmt -vv test export --policy-file ../policies/test/$1 $plan"
+            rlRun -s "tmt -vv test export --policy-file ../policies/test/$1 $plan"
+            rlAssertGrep "Apply tmt policy '../policies/test/$1' to tests." $rlRun_LOG
+
             rlRun -s "tmt -vv test export --policy-file ../policies/test/$1 $plan 2> /dev/null | yq -cSr '.[] | $filter'"
 
             rlAssertEquals \
--- a/tmt/policy.py
+++ b/tmt/policy.py
@@ -244,9 +244,8 @@ class Policy(MetadataContainer):
         """
 
         logger.info(
-            f"Apply tmt policy '{self.name}'.",
+            f"Apply tmt policy '{self.name}' to tests.",
             color='green',
-            topic=Topic.POLICY,
         )
 
         self._apply(tests, self.test_policy, logger.descend())


thrix (approved):
LGTM, not easily graspable, but looks correct to me.  I was firstly confused if the old code doing:

```
return Plan(
....
    importing_environment=self.environment if reference.inherit_environment else None,
....
)
```

vs 

```
return Plan(
....
    importing_environment=inherited_environment,
....
)
```

where the inherited plan now does not actually use the whole `self.environment` but excludes some fields. But it looks ok I think, because we do not inherit environment variables from all the sources from the imported plans. Still the claim of "no functional change" is debatable in this case :)

happz:
> LGTM, not easily graspable, but looks correct to me.

It is complicated, several sources add to the final outcome, btu I hope it would be clearer now, as that was the goal of this PR: to give these sources more fitting names, clearly show how they are added up, and make it clear which sources are inheritable.

> I was firstly confused if the old code doing:
> 
> ```
> return Plan(
> ....
>     importing_environment=self.environment if reference.inherit_environment else None,
> ....
> )
> ```
> 
> vs
> 
> ```
> return Plan(
> ....
>     importing_environment=inherited_environment,
> ....
> )
> ```
> 
> where the inherited plan now does not actually use the whole `self.environment` but excludes some fields.

Hmm, I'm going to change that `importing_environment`, it's correct but the naming scheme changed a bit, and it might be indeed puzzling.

> But it looks ok I think, because we do not inherit environment variables from all the sources from the imported plans.

Well, we do, actually, but only some of those sources are owned by the importing plan, it's just its own environment variables (`environment`, `environment-file`) and variables the importing plan inherited. The rest of sources are shared - CLI (`--environment`), plan environment file, run. Effectively we do "inherit" these from the importing plan, in a sense that they will be part of the imported plan's environment, but they are not taken from the importing plan, as all plans add them to their own environment in `environment` property, therefore we do not need to extract them from the importing plan. Therefore `_inheritable_environment` provides only what is owned by the importing plan, environment we cannot get fromanywhere else.

> Still the claim of "no functional change" is debatable in this case :)

I believe it's fitting, the goal is to refactor internals for better maintenance, for users nothign shall change.

therazix (approved):

tcornell-bus:
It is difficult to compare the old and new changes, but only looking at the new it is much easier to understand!

psss (approved):
Looks much better! Thanks for improving the readability here! 👍

PR #3836, opened 2025-06-23
Refactor plan fmf context composition

happz:
* clean separation of various context sources,
* more fitting names for the corresponding methods.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -2175,6 +2175,19 @@ class Plan(
     # Plan environment and its components
     #
 
+    # Q: what part of plan environment or context should be "inheritable"
+    #    by imported plans?
+    #
+    # A: Only what the plan truly and fully owns: its own `environment`
+    #    and `environment-file` keys, `context` key, and environment
+    #    and context inherited from its importing plan if this one was
+    #    also imported.
+    #
+    #    No "shared" environment shall be part of the "inheritable" bundle.
+    #    By leaving CLI inputs or plan environment file out, we make the
+    #    inheritance clearer, without duplicities, and the composition
+    #    is then easier to extend or debug.
+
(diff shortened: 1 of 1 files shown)


tmt/base.py
```diff
@@ -3130,21 +3171,31 @@ def _convert_node(node: fmf.Tree) -> 'Plan':
             self.debug(f"Turning node '{node.name}' into a plan.", level=3)
 
             # Prepare fmf context and environment the imported node would inherit.
-            inherited_fmf_context = self._fmf_context if reference.inherit_context else None
+            inherited_fmf_context = (
+                self._inheritable_fmf_context if reference.inherit_context else None
```
tcornell-bus, line 3188: The only thing I can see that might be a functional change is this. The old `_fmf_context` has the cli context included whereas `_inheritable_fmf_context` does not have it. I would assume though based on the name `inherited_fmf_context` that the caller of `_convert_node` should not be expecting cli context to be included in this Plan member. The `tests/plan/import/inheritance.sh` passes so cli context must be accounted for later. Just wanted to make sure my understanding of this is correct.

thrix (approved):
LGTM

tmt/base.py
thrix: @happz an outstanding question, ptal pls

happz: Yes, the CLI contribution to context/environment is excluded from the inheritance, because it is supposed to be always included. It is added in `fmf_context` and `environment` properties.

I added a comment in 3db13269, let me know whether it helped.

vaibhavdaren (approved):

thrix:

* for Rawhide and F43 see https://issues.redhat.com/browse/TFT-3621

PR #3837, opened 2025-06-23
Emit a more specific warning when a hardware constraint is not supported

### LINKED ISSUE #3812: Emit a better warning message when a particular hardware contraint is not supported
Issue originates from this [discussion](https://github.com/teemtee/tmt/pull/3810/files#r2143466602). 
Take this hardware requirement as an example:
```
        hardware:
            or:
              - memory: '>= 4 GB'
              - memory: < 4 GB
```
Currently a warning like this `warn: Hardware requirement 'memory' is not supported.` is emitted, but the first requirement `memory: '>= 4 GB'` is still honored, it is actually warning about no support for the `<` operator in `memory: < 4 GB`. The warning could be made more clear to the user. Something like this `warn: Hardware requirement 'memory: < 4 GB' is not supported.` instead would suffice.

tcornell-bus:
Include the full hardware constraint in the warning when it is not supported.

Add a unit test to verify the more detailed warning message.

Fix #3812.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/unit/test_hardware.py
+++ b/tests/unit/test_hardware.py
@@ -1,6 +1,8 @@
+import logging
 import textwrap
 from typing import Any
 
+import _pytest.logging
 import pytest
 
 import tmt.hardware
@@ -8,6 +10,9 @@ import tmt.steps
 import tmt.steps.provision
 import tmt.utils
 from tmt.hardware import Hardware
+from tmt.log import Logger
+
+from . import MATCH, assert_log
 
 
 def parse_hw(text: str) -> Hardware:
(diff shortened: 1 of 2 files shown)


tests/unit/test_hardware.py
```diff
@@ -359,3 +364,30 @@ def test_parse_or_constraint() -> None:
...
+    root_logger: Logger,
+    caplog: _pytest.logging.LogCaptureFixture,
+) -> None:
+    # Spec to test against
+    hw_spec = """
+     or:
+       - memory: '>= 4 GB'
+       - memory: '!= 4 GB'
+    """
+    hw = parse_hw(hw_spec)
```
happz: Since `hw_spec` is not needed for any kind of comparison, would it be possible to drop it and let `parse_hw()` crunch the string itself?

happz (approved):

PR #792, opened 2021-05-31
Install `rsync` before guest.pull()

thrix:
@mruprich reported that after running installability test the `tmt`
fails on missing rsync when pulling out artifacts.

After investiation, it turned out, installability for `rsync` removes
`rsync` from the test machine, but `tmt` still expects it to be around.

This tries to fix the problem and make sure `rsync` is still installed
on the machine before trying to pull out the artifacts.

Todo:
- [x] review comments
- [x] fix the issues with current tests
- [x] add a new test covering the problem

Signed-off-by: Miroslav Vadkerti <mvadkert@redhat.com>

PR #3840, opened 2025-06-25
Do not attempt to install rsync when push/pull fails

happz:
When guest push or pull fails, tmt assumes it is because of a missing rsync. tmt continues by trying to install rsync package, and trying again.

Patch removes this attempt, and replaces it with a guest fact - push/pull then check the fact, and if rsync is missing, it tries to install it, just once, before doing the first push/pull ever.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/execute/unresponsive/test.sh
+++ b/tests/execute/unresponsive/test.sh
@@ -8,7 +8,7 @@ rlJournalStart
 
     rlPhaseStartTest "Test must error out if machine becomes unresponsive during execution (#3647)."
         rlRun -s "tmt run -vv -a provision -h $PROVISION_HOW" "2"
-        rlAssertGrep 'fail: Failed to verify rsync presence on the guest. This often means there is a problem with its package manager, or logging in as 'root' does not work, or the network connection itself.' $rlRun_LOG '-F'
+        rlAssertGrep 'Failed to verify rsync presence on the guest.' $rlRun_LOG '-F'
         rlAssertGrep 'errr /unresponsive/test/error' $rlRun_LOG '-F'
         rlAssertGrep 'pending /unresponsive/test/pending' $rlRun_LOG '-F'
     rlPhaseEnd


psss:
This makes the log investigation much harder, we should merge as soon as possible, proposing `must` priority.

tmt/steps/provision/__init__.py
tcornell-bus, line 1: Should the `class CheckRsyncOutcome(enum.Enum)` be removed as well since it looks like all uses of it have been removed?

Otherwise LGTM!

psss:
One of the [failing](https://artifacts.osci.redhat.com/testing-farm/247498ee-1c87-4a9b-b257-aa9d343671bb/work-executeztp6semg/plans/provision/virtual/execute/execute/data/guest/default-0/tests/execute/rsync-11/output.txt) tests is exercising a use case when the `rsync` package is removed during the test execution:

https://github.com/teemtee/tmt/blob/32ca248bd0e14ef8b9fe00f3d0235dfc3c0c2e2e/tests/execute/rsync/test.sh#L10-L14

I remember adding this test coverage was motivated by a real-life problem which happened during the component testing. Do we want/need to cover this scenario or is this such a big corner case that we should decide to not support it and just document it? @thrix, @happz, @lukaszachy, any thoughts here?

happz:
> One of the [failing](https://artifacts.osci.redhat.com/testing-farm/247498ee-1c87-4a9b-b257-aa9d343671bb/work-executeztp6semg/plans/provision/virtual/execute/execute/data/guest/default-0/tests/execute/rsync-11/output.txt) tests is exercising a use case when the `rsync` package is removed during the test execution:
> 
> https://github.com/teemtee/tmt/blob/32ca248bd0e14ef8b9fe00f3d0235dfc3c0c2e2e/tests/execute/rsync/test.sh#L10-L14
> 
> I remember adding this test coverage was motivated by a real-life problem which happened during the component testing. Do we want/need to cover this scenario or is this such a big corner case that we should decide to not support it and just document it? @thrix, @happz, @lukaszachy, any thoughts here?

I guess the question is, what's the expectation here? "Correctly handle rsync removed during test" - what does "correctly" mean? Is tmt expected to undo the change, to "fix" the guest and install rsync back? Or was that just a happy coincidence we do not need to preserve? That will decide the fate of this PR as it is now.

tmt/steps/provision/__init__.py
happz: Addressed in 76f5dbd8

psss:
> Is tmt expected to undo the change, to "fix" the guest and install rsync back?

Yeah, so this was the current approach: Any time `rsync` is missing on the guest we install it. Even if this happens during the test execution.

> Or was that just a happy coincidence we do not need to preserve?

I'm not sure I understand the "happy coincidence" part. This handling was added intentionally in #792 because of a specific failure reported by user releated to the installability test.

happz:
> > Is tmt expected to undo the change, to "fix" the guest and install rsync back?
> 
> Yeah, so this was the current approach: Any time `rsync` is missing on the guest we install it. Even if this happens during the test execution.
> 
> > Or was that just a happy coincidence we do not need to preserve?
> 
> I'm not sure I understand the "happy coincidence" part. This handling was added intentionally in #792 because of a specific failure reported by user releated to the installability test.

Ok, then this PR needs to change. The goal was to drop the blind assumption the problem was caused by a missing rsync. That should stay, but it needs to become smarter and preserve the installation of rsync when it's actually missing.

happz:
> > > Is tmt expected to undo the change, to "fix" the guest and install rsync back?
> > 
> > 
> > Yeah, so this was the current approach: Any time `rsync` is missing on the guest we install it. Even if this happens during the test execution.
> > > Or was that just a happy coincidence we do not need to preserve?
> > 
> > 
> > I'm not sure I understand the "happy coincidence" part. This handling was added intentionally in #792 because of a specific failure reported by user releated to the installability test.
> 
> Ok, then this PR needs to change. The goal was to drop the blind assumption the problem was caused by a missing rsync. That should stay, but it needs to become smarter and preserve the installation of rsync when it's actually missing.

Updated: after each test, the `has-rsync` fact is invalidated, and the first attempt to use push/pull will retest it. If a test removes rsync, this will trigger its explicit installation; if rsync is still present, tmt should move on.

tcornell-bus (approved):

tmt/steps/provision/__init__.py
```diff
@@ -747,27 +756,36 @@ def _query_capabilities(self, guest: 'Guest') -> dict[GuestCapability, bool]:
             GuestCapability.SYSLOG_ACTION_READ_CLEAR: True,
         }
 
-    def sync(self, guest: 'Guest') -> None:
+    def sync(self, guest: 'Guest', *facts: str) -> None:
         """
         Update stored facts to reflect the given guest
```
psss: What about documenting the `facts` parameter?

psss (approved):
Looks nice! Perhaps one nitpick?

tmt/steps/provision/__init__.py
happz: Added in dbf860a0e08876b09d7f7feb7b4a41ee43fbb60b

tmt/steps/provision/__init__.py
```diff
@@ -2548,6 +2552,20 @@ def execute(
...
+        """
+
+        if self.facts.has_rsync:
+            return
+
+        self.debug('rsync has not been confirmed on the guest, try installing it')
+
+        self.package_manager.install(Package('rsync'))
+
+        self.facts.sync(self, 'has_rsync')
```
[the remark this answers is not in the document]
happz: If `install()` call fails, it would happen to raise an exception, which would be propagated upwards. I don't believe there is a need for extra error handling.

tmt/steps/provision/__init__.py
psss: Thanks!

happz:
@coderabbitai full review

tmt/steps/provision/__init__.py
```diff
@@ -750,27 +759,41 @@ def _query_capabilities(self, guest: 'Guest') -> dict[GuestCapability, bool]:
...
+            self.bootc_builder = self._query_bootc_builder(guest)
+            self.has_selinux = self._query_has_selinux(guest)
+            self.has_systemd = self._query_has_systemd(guest)
+            self.has_rsync = self._query_has_rsync(guest)
+            self.is_superuser = self._query_is_superuser(guest)
+            self.is_ostree = self._query_is_ostree(guest)
+            self.is_toolbox = self._query_is_toolbox(guest)
+            self.toolbox_container_name = self._query_toolbox_container_name(guest)
+            self.is_container = self._query_is_container(guest)
+            self.capabilities = self._query_capabilities(guest)
```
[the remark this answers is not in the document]
happz: Improved in 17b098d1

tmt/utils/__init__.py
```diff
@@ -1110,7 +1110,9 @@ def __init__(self, *elements: RawCommandElement) -> None:
     def __str__(self) -> str:
         return self.to_element()
 
-    def __add__(self, other: Union['Command', RawCommand, list[str]]) -> 'Command':
+    def __add__(
+        self, other: Union['Command', RawCommand, Sequence[RawCommandElement]]
+    ) -> 'Command':
```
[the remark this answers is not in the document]
happz: Correct, `Sequence` was not good. Addressed in `17b098d1`.

tmt/steps/provision/__init__.py
```diff
@@ -750,27 +759,48 @@ def _query_capabilities(self, guest: 'Guest') -> dict[GuestCapability, bool]:
...
+            for fact in facts:
+                if not hasattr(self, fact):
+                    raise GeneralError(f"Cannot sync unknown guest fact '{fact}'.")
+
+                method_name = f'_query_{fact}'
+
+                if not hasattr(self, method_name):
+                    raise GeneralError(
+                        f"Cannot sync guest fact '{fact}', query method '{method_name} not found."
+                    )
```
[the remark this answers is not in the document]
happz: Fixed in f008e520

psss:
Seems there are some tests failing, `/tests/execute/unresponsive` is checking for the original error message `Failed to pull workdir` which should be updated, right?

```

Cause number 1:

    execute step failed

    Cause number 1:

        Command 'rpm -q --whatprovides rsync || dnf5 install -y  rsync' returned 255.

        # stderr (2 lines)
        kex_exchange_identification: read: Connection reset by peer
        Connection reset by 127.0.0.1 port 10023

:: [ 15:18:15 ] :: [   PASS   ] :: Command 'tmt run -vv -a provision -h virtual' (Expected 2, got 2)
:: [ 15:18:15 ] :: [   FAIL   ] :: File '/var/tmp/rlRun_LOG.aXUn4Qfp' should contain 'fail: Failed to pull workdir from the guest.' 
```

The orther one seems irrelevant image download failure. Although not sure how long `fedora-40` will still be with us.

happz:
> Seems there are some tests failing, `/tests/execute/unresponsive` is checking for the original error message `Failed to pull workdir` which should be updated, right?
> 
> ```
> The exception was caused by 1 earlier exceptions
> 
> Cause number 1:
> 
>     execute step failed
> 
>     The exception was caused by 1 earlier exceptions
> 
>     Cause number 1:
> 
>         Command 'rpm -q --whatprovides rsync || dnf5 install -y  rsync' returned 255.
> 
>         # stderr (2 lines)
>         kex_exchange_identification: read: Connection reset by peer
>         Connection reset by 127.0.0.1 port 10023
> 
> :: [ 15:18:15 ] :: [   PASS   ] :: Command 'tmt run -vv -a provision -h virtual' (Expected 2, got 2)
> :: [ 15:18:15 ] :: [   FAIL   ] :: File '/var/tmp/rlRun_LOG.aXUn4Qfp' should contain 'fail: Failed to pull workdir from the guest.' 
> ```

Hopefully addressed in bea9596.

psss:
> Hopefully addressed in [bea9596](https://github.com/teemtee/tmt/commit/bea95964fbb92db8afc5a7421ee07454ad368d98).

Ouch, seems that [quotes](https://artifacts.osci.redhat.com/testing-farm/dfcb27db-0cf9-415f-aa3c-8e23d97281b5/work-executeb8dslg97/plans/provision/virtual/execute/execute/data/guest/default-0/tests/execute/unresponsive-13/output.txt) are missing :face_with_peeking_eye: 

```patch
-fail: Failed to verify rsync presence on the guest. This often means there is a problem with its package manager, or logging in as root does not work, or the network connection itself
+fail: Failed to verify rsync presence on the guest. This often means there is a problem with its package manager, or logging in as 'root' does not work, or the network connection itself.
```

Perhaps checking for `Failed to verify rsync presence on the guest.` would be just enough?

psss:
Fedora 40 download [failed](https://artifacts.osci.redhat.com/testing-farm/dfcb27db-0cf9-415f-aa3c-8e23d97281b5/work-prepare1um7q5mx/plans/provision/virtual/prepare/execute/data/guest/default-0/tests/prepare/ansible-1/output.txt) again:

    qcow: Fedora-Cloud-Base-Generic.x86_64-40-1.14.qcow2
    progress: downloading...
    fail: Image download failed after 3 attempts.
    multihost name: default-0

I guess it's time to update the list of container images:

 * #3914

psss:
Failing tests are `fedora-40` image downloads, to be fixed in #3914.

PR #3842, opened 2025-06-25
Mention the public tmt web service instance

psss:
The public instance of the tmt web service has been deployed. Let's mention it in the release notes and add a couple example links so that users can get a better idea of how it looks like.

Pull Request Checklist

* [ ] write the documentation
* [ ] include a release note


happz:
Documentation-only, merging.

Issue #3820, opened 2025-06-17
tmt-1.50: formatting issue in Report Portal logs


smyle10:
In tmt-1.50, the log in Report Portal isn't formatted/wrapped well. Some xml markers are shown in logs.
Check https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1750145627736829 for the screenshots of the issue.

Reproducer:
```
git clone https://gitlab.cee.redhat.com/kernel-qe/networking/nmstate
export TMT_PLUGIN_REPORT_REPORTPORTAL_XXX=YOUR_RP_SETTINGS
tmt -c distro=RHEL-9.7 run plan -n /anyhost test -n /test_bond_[px]
```

Run the same tmt command in tmt-1.49, everything's good.

kkaarreell:
This comes from here https://github.com/teemtee/tmt/blob/main/tmt/steps/report/reportportal.py#L589 where the following logs are being processed (added): 
```
output.txt
journal.txt
journal.xml
```
Clearly not all of them are needed but it depends on a test type I guess. For beakerlib test just `output.txt` should be enough.

EDIT: Originates here https://github.com/teemtee/tmt/blob/f962b28fe02800bfd34cca10f270f0206035bf00/tmt/frameworks/beakerlib.py#L137

Issue #3674, opened 2025-04-16
Upload AVC logs to Report Portal


brakogh:
With tmt and Report Portal, we are now able to enable AVC check but we don't upload AVC logs into Report Portal. Can you start uploading AVC logs? In TF, these are present as avc.txt [1]

1. https://artifacts.osci.redhat.com/testing-farm/b18f9ab4-fed4-4d01-bf41-f7eeab1baea5/work-squidqumtz689/plans/cs_stacks/tier2/squid/execute/data/guest/default-0/internal/Security/CVE-2018-19131-XSS-when-generating-HTTPS-response-messages-about-TLS-errors/system-19/checks/avc.txt

Related issue:

* #3666 
* https://issues.redhat.com/browse/TT-349

psss:
During the discussion it was proposed to upload the AVC logs to `ReportPortal` only if the check is failing to prevent storing too many unnecessary logs.

martinky82:
With the test profiles heading into TMT I suppos this issue gains weight. Without the logs the AVC failure is just a nuisance for the user, who don't bother more than to 'waive' the result. I agree that for passed results, the logs bring no value at all so we need them uploaded only when the check fails.

PR #3844, opened 2025-06-26
Allow log selection for `reportportal` plugin

therazix:
This PR adds a new `upload-log-pattern` key to ReportPortal plugin. This key allows users to select which logs should be uploaded to ReportPortal by specifying the pattern to search for in the log file names. Check result logs are also affected by this key but are uploaded only if the check fails or if an error occurs during execution.

Fixes: #3674 
Fixes: #3820 

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -8,11 +8,11 @@ tmt-1.52.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The :ref:`/plugins/report/reportportal` report plugin now supports
-a new ``upload-log-pattern`` option. This option allows users to select
-which logs should be uploaded by specifying the pattern to search for
-in the log file names. Check result logs are also affected by this
-option but are uploaded only if the check fails or if an error occurs
-during execution.
+a new ``upload-log-pattern`` key. This key allows users to select
+which logs should be uploaded by specifying the pattern to search
+for in the log file names. Check result logs are also affected by
+this key but are uploaded only if the check fails or if an error
+occurs during execution.
 
 
 tmt-1.52.0
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -4,6 +4,15 @@
     Releases
 ======================
 
+tmt-1.52.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`/plugins/report/reportportal` report plugin now supports
+a new ``log`` option. This option allows users to select which logs
+should be uploaded by specifying their names. Check result logs are
+also affected by this option but are uploaded only if the check fails
+or if an error occurs during execution.
+
 
 tmt-1.52.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -8,10 +8,11 @@ tmt-1.52.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The :ref:`/plugins/report/reportportal` report plugin now supports
-a new ``log`` option. This option allows users to select which logs
-should be uploaded by specifying their names. Check result logs are
-also affected by this option but are uploaded only if the check fails
-or if an error occurs during execution.
+a new ``upload-log-pattern`` option. This option allows users to select
+which logs should be uploaded by specifying the pattern to search for
+in the log file names. Check result logs are also affected by this
+option but are uploaded only if the check fails or if an error occurs
+during execution.
 
 
 tmt-1.52.0
--- a/tmt/schemas/report/reportportal.yaml
+++ b/tmt/schemas/report/reportportal.yaml
@@ -83,7 +83,7 @@ properties:
   link-template:
     type: string
 
-  log:
+  upload-log-pattern:
     $ref: "/schemas/common#/definitions/array_of_strings"
 
   when:
--- a/tmt/schemas/report/reportportal.yaml
+++ b/tmt/schemas/report/reportportal.yaml
@@ -83,6 +83,9 @@ properties:
   link-template:
     type: string
 
+  log:
+    $ref: "/schemas/common#/definitions/array_of_strings"
+
   when:
     $ref: "/schemas/common#/definitions/when"
 
--- a/tmt/schemas/report/reportportal.yaml
+++ b/tmt/schemas/report/reportportal.yaml
@@ -84,7 +84,7 @@ properties:
     type: string
 
   upload-log-pattern:
-    $ref: "/schemas/common#/definitions/array_of_strings"
+    $ref: "/schemas/common#/definitions/one_or_more_strings"
 
   when:
     $ref: "/schemas/common#/definitions/when"
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -1,7 +1,7 @@
 import datetime
 import os
 import re
-from typing import TYPE_CHECKING, Any, Optional, overload
+from typing import TYPE_CHECKING, Any, Optional, Union, overload
 
 import requests
 import urllib3
@@ -11,6 +11,7 @@ import tmt.log
 import tmt.steps.report
 import tmt.utils
 import tmt.utils.templates
+from tmt._compat.pathlib import Path
 from tmt.container import container, field
 from tmt.result import ResultOutcome
 from tmt.utils import (
@@ -28,6 +29,13 @@ if TYPE_CHECKING:
 JSON: 'TypeAlias' = Any
 DEFAULT_LOG_SIZE_LIMIT: 'Size' = tmt.hardware.UNITS('1 MB')
 DEFAULT_TRACEBACK_SIZE_LIMIT: 'Size' = tmt.hardware.UNITS('50 kB')
+
+DEFAULT_LOG_NAMES: list[str] = [
+    r'avc\.txt',
+    r'dmesg-.*\.txt',
+    r'output\.txt',
+    r'tmt-watchdog\.txt',
+]
 
 
 def _flag_env_to_default(option: str, default: bool) -> bool:
@@ -52,6 +60,13 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
     if env_var not in os.environ or os.getenv(env_var) is None:
         return default
     return str(os.getenv(env_var))
+
+
+def _list_env_to_default(option: str, default: list[str]) -> list[str]:
+    env_var = 'TMT_PLUGIN_REPORT_REPORTPORTAL_' + option.upper()
+    if env_var not in os.environ or os.getenv(env_var) is None:
+        return default
+    return [item.strip() for item in str(os.getenv(env_var)).split(',') if item.strip()]
 
 
 def _size_env_to_default(option: str, default: 'Size') -> 'Size':
@@ -283,6 +298,17 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
              Jinja template that will be rendered for each test result and appended to the end
              of its description. The following variables are passed to the template:
              ``PLAN_NAME``, ``RESULT``.
+             """,
+    )
+
+    log: list[str] = field(
+        metavar="NAMES",
+        option="--log",
+        default_factory=lambda: _list_env_to_default('log', DEFAULT_LOG_NAMES),
+        help="""
+             List of result log names or regular expressions that should be uploaded
+             to ReportPortal. Check result logs will be uploaded only if the check failed or
+             if an error occurred during the execution.
              """,
     )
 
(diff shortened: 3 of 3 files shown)


tmt/steps/report/reportportal.py
```diff
@@ -54,6 +62,13 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
     return str(os.getenv(env_var))
 
 
+def _list_env_to_default(option: str, default: list[str]) -> list[str]:
```
happz: Probably `list_pattern`, `-> list[Pattern[str]]`, an employ `normalize_pattern_list()`. to turn input into actual patterns.

tmt/steps/report/reportportal.py
```diff
@@ -286,6 +301,17 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
              """,
     )
 
+    log: list[str] = field(
```
happz: `list[Pattern[str]]` - see e.g. `failure_pattern` in the dmesg check.

tmt/steps/report/reportportal.py
```diff
@@ -586,53 +612,56 @@ def upload_result_logs(
...
+                        "itemUuid": item_uuid,
+                        "launchUuid": launch_uuid,
+                        "level": "ERROR" if is_traceback else "INFO",
+                        "time": timestamp,
+                    },
+                )
+
+        # Upload result logs
+        for log_path in result.log:
+            if any(re.search(name, log_path.name) for name in self.data.log):
```
happz: `if any(pattern.search(log_path.name)) for pattern in self.data.log`, because we would have patterns already compiled.

tmt/steps/report/reportportal.py
happz: Oh, and we should have a better name, `log` can mean anything. `include-log-pattern`, `upload-log-pattern`? We want to express the field is for patterns that, if log name matches, the log would be uploaded to RP.

tmt/steps/report/reportportal.py
```diff
@@ -29,6 +30,13 @@
 DEFAULT_LOG_SIZE_LIMIT: 'Size' = tmt.hardware.UNITS('1 MB')
 DEFAULT_TRACEBACK_SIZE_LIMIT: 'Size' = tmt.hardware.UNITS('50 kB')
 
+DEFAULT_LOG_NAMES: list[str] = [
```
happz: * `DEFAULT_LOG_NAME_PATTERNS` would be more fitting
* `list[Pattern[str]]` + apply `re.compile()` to each pattern to build a list of patterns.

tmt/steps/report/reportportal.py
therazix: Changed in 5ab0e75349d0bba101c342a57e49876fb7dbf54a.

tmt/steps/report/reportportal.py
therazix: Changed in 5ab0e75349d0bba101c342a57e49876fb7dbf54a.

tmt/steps/report/reportportal.py
therazix: Changed in 5ab0e75349d0bba101c342a57e49876fb7dbf54a.

tmt/steps/report/reportportal.py
therazix: Changed in 5ab0e75349d0bba101c342a57e49876fb7dbf54a.

docs/releases.rst
```diff
@@ -4,6 +4,16 @@
     Releases
 ======================
 
+tmt-1.52.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`/plugins/report/reportportal` report plugin now supports
+a new ``upload-log-pattern`` option. This option allows users to select
```
```suggestion
a new ``upload-log-pattern`` option. This key allows users to select
```

`key` instead of `option` - try to be fmf-centric, options are easy to derive from key names.

tmt/steps/report/reportportal.py
```diff
@@ -286,6 +307,24 @@ class ReportReportPortalData(tmt.steps.report.ReportStepData):
...
+            'upload_log_pattern', DEFAULT_LOG_PATTERNS[:]
+        ),
+        normalize=tmt.utils.normalize_pattern_list,
+        serialize=lambda patterns: [pattern.pattern for pattern in patterns],
+        unserialize=lambda serialized: [re.compile(pattern) for pattern in serialized],
+        help="""
+             List of regular expressions to look for in result log names. If any of the
+             patterns is found in a log file name, the log will be uploaded to ReportPortal.
+             Check result logs will be uploaded only if the check failed or if an error
+             occurred during the execution.
```
thrix: Because `re.search` is used for regular expression matching. Could we make sure the documentation is clear about this, and it does not cause wrong expectations (for example users expecting `re.match`?).

tmt/steps/report/reportportal.py
```diff
@@ -54,6 +66,15 @@ def _str_env_to_default(option: str, default: Optional[str]) -> Optional[str]:
     return str(os.getenv(env_var))
 
 
+def _pattern_list_env_to_default(option: str, default: list[Pattern[str]]) -> list[Pattern[str]]:
+    env_var = 'TMT_PLUGIN_REPORT_REPORTPORTAL_' + option.upper()
+    if env_var not in os.environ or os.getenv(env_var) is None:
+        return default
+    return [
+        re.compile(item.strip()) for item in str(os.getenv(env_var)).split(',') if item.strip()
+    ]
```
thrix: I wonder, if we should not guard this by possible issues compiling the regular expression:

```
      try:
          return [
              re.compile(item.strip())
              for item in str(os.getenv(env_var)).split(',')
              if item.strip()
          ]
      except re.error as e:
          raise tmt.utils.GeneralError(
              f"Invalid regex pattern in environment variable {env_var}: {e}"
          )
```

happz: `tmt.utils.normalize_pattern_list` exists for this exact purpose.

tmt/steps/report/reportportal.py
happz: Yep, include a reference to regular expressions docs, see e.g. `test` key in `discover/fmf`:

```
The search mode is used for pattern matching. See the :ref:`regular-expressions` section for details.
```

docs/releases.rst
therazix: Changed in c95a9ba5abe9280de1eca4a6f9ebd99175eed53f.

tmt/steps/report/reportportal.py
therazix: Added in c95a9ba5abe9280de1eca4a6f9ebd99175eed53f.

tmt/steps/report/reportportal.py
therazix: Changed in c95a9ba5abe9280de1eca4a6f9ebd99175eed53f, but I don't see a way to pass a logger to the `tmt.utils.normalize_pattern_list` function. Is there a better approach than creating a new logger? https://github.com/teemtee/tmt/blob/c95a9ba5abe9280de1eca4a6f9ebd99175eed53f/tmt/steps/report/reportportal.py#L76

happz: Use `tmt.log.Logget.get_bootstrap_logger()` - it's not right, it's caused by evaluating environment variables so soon, but it's better than inventing new way to normalize a list of patterns.

therazix: Changed in 50747025cc10ca982f608d4e2c766c9eed20c9f3.

tmt/steps/report/reportportal.py
happz: LGTM

tmt/steps/report/reportportal.py
happz: LGTM

thrix (approved):

tmt/schemas/report/reportportal.yaml
```diff
@@ -83,6 +83,9 @@ properties:
   link-template:
     type: string
 
+  upload-log-pattern:
+    $ref: "/schemas/common#/definitions/array_of_strings"
```
happz: `/schemas/common#/definitions/one_or_more_strings` will be better, we want support both `upload-log-pattern: foo` as well as a list of patterns.

therazix: Changed in 5ab51469c1f0548bb6dc3fbac885cbf2d0e646e9.

tmt/steps/report/reportportal.py
```diff
@@ -29,6 +31,16 @@
 DEFAULT_LOG_SIZE_LIMIT: 'Size' = tmt.hardware.UNITS('1 MB')
 DEFAULT_TRACEBACK_SIZE_LIMIT: 'Size' = tmt.hardware.UNITS('50 kB')
 
+DEFAULT_LOG_PATTERNS: list[Pattern[str]] = [
+    re.compile(pattern)
+    for pattern in [
+        r'avc\.txt',
+        r'dmesg-.*\.txt',
+        r'output\.txt',
+        r'tmt-watchdog\.txt',
```
psss, line 40: Shall we mention these somewhere in the plugin documentation? So that users can find out which logs are uploaded by default?

tmt/steps/report/reportportal.py
```diff
@@ -586,53 +628,56 @@ def upload_result_logs(
...
 
-        for log_path in result.log:
-            if log_path.name == tmt.steps.execute.TEST_FAILURES_FILENAME:
-                continue
+        def upload_log(log_path: Path, is_traceback: bool = False) -> None:
             try:
-                log = self.step.plan.execute.read(log_path)
-            except tmt.utils.FileError:
-                continue
+                if log_path.suffix in ('.yaml', '.yml'):
```
psss: I'm not completely sure about this autodetection based on the suffix. What if a test submits for example a custom `test-output.yaml` file? It would be included in `result.log` and here it would crash.

happz: Maybe an attempt to handle the failure log? That one is saved as a YAML file.

tmt/steps/report/reportportal.py
happz: It's propagated to HTML docs, but not to CLI. `reportportal` needs to be fixed anyway, the way it mixes defaults and envvars is just a workaround to deal with tmt issues...

tmt/steps/report/reportportal.py
psss: Yes, but I think this is a bit fragile. I'd say the original way which explicitly parsed failure logs would be more reliable / appropriate:

https://github.com/teemtee/tmt/blob/50d53a9ac3dc8b6c4d3ee6f67584d653469974cf/tmt/steps/report/reportportal.py#L614-L616

tmt/steps/report/reportportal.py
psss: > It's propagated to HTML docs, but not to CLI.

Nice, I can see it included in the [reportportal plugin section](https://tmt--3844.org.readthedocs.build/en/3844/plugins/report.html#reportportal).

tmt/steps/report/reportportal.py
therazix: Changed in 7ccc6f24cc3ef2f0d23e46ecc506360aad54cfe7.

thrix: @psss pls re-review

psss: Looks better, thanks!

therazix:
As discussed in the review meeting, I updated the code so that each log now includes the log file name. 530d7a91896e544dd1a48c25b71be41939009bbf

tmt/steps/report/reportportal.py
```diff
@@ -586,53 +628,58 @@ def upload_result_logs(
...
-                },
+            filter_settings = LogFilterSettings(
+                size=self.data.traceback_size_limit if is_traceback else self.data.log_size_limit,
+                is_traceback=is_traceback,
             )
 
-        # Optionally write out failures
+            for log in logs:
+                # Add file name to the log
+                log = f'# `{log_path.name}`\n{log}'
```
happz: From RP API docs for `/v1/{projectName}/log/entry`:

```
  "file": {
    "name": "string"
  },
  "itemUuid": "string",
  "launchUuid": "string",
  "level": "error",
  "message": "string",
  "time": "2025-07-01T10:49:44.988Z",
  "uuid": "string"
}
```

Would you be willing to add `file` object into the payload below, and set its `name` key to the log name, instead of adding it to the content? How would that look?

therazix: I tried that and I don't see the log file name anywhere:

happz: Ok. We tried.

psss (approved):

therazix:
As discussed on Slack, I’ve removed the file name prefix from the error logs. 4be1e2a385996a1c2192e7e396cb66a446df3de4

psss:
> As discussed on Slack, I’ve removed the file name prefix from the error logs. [4be1e2a](https://github.com/teemtee/tmt/commit/4be1e2a385996a1c2192e7e396cb66a446df3de4)

Thanks, should be good to go. I've also checked how the output looks like with an avc check failure:

image

The formatting might be a little bit unexpected (several headings at the beginning), but we can improve it later.

psss (approved):

psss:
All red jobs are irrelevant rawhide `Checksum error` failures. Merging.

PR #3845, opened 2025-06-26
Use `self.data` instead of `self.opt()` in `fmf` discover

psss:
Drop the obsolete usage of `self.opt()` for getting values of the discover plugin keys and directly use `self.data` instead. This also fixes the `sync-repo` key to work not only from the command line but from the `fmf` files as well.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -444,7 +444,7 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
         Enable standalone mode when listing fmf ids
         """
 
-        if self.opt('fmf_id'):
+        if self.data.fmf_id:
             return True
         return super().is_in_standalone_mode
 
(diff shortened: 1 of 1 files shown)


tmt/steps/discover/fmf.py
```diff
@@ -548,15 +548,14 @@ def assert_git_url(plan_name: Optional[str] = None) -> None:
...
         # Copy git repository root to workdir
         else:
             if path is not None:
                 fmf_root: Optional[Path] = path
             else:
                 fmf_root = Path(self.step.plan.node.root)
-            requires_git = self.opt('sync-repo') or any(
-                self.get(opt) for opt in self._REQUIRES_GIT
-            )
+            requires_git = self.data.sync_repo or any(self.get(opt) for opt in self._REQUIRES_GIT)
```
happz: `or any(getattr(self.data, option_to_key(key)) for key in self._REQUIRES_GIT)` ?

Or change `_REQUIRES_GIT` to contain field names rather than key names, i.e. `-` => `_`?

psss: Good point, fixed in ebbc2c17.

AthreyVinay (approved):

happz (approved):

Issue #2398, opened 2023-10-11
Add files submitted via tmt-file-submit as logs to results


happz:
Files processed by `tmt-file-submit` - or `rlFileSubmit` - seem to be unreachable through results. They are available in `data` directory, but they are not references. How about adding some kind of list which would then be added to `Result.log`?

happz:
I consider this to be a fairly big gap in reporting. Kicking of with `priority | must`, adding to triage meeting, maybe the priority gets reconsidered.

comps:
A related consideration might be - added to *which* result/report?

What if the user calls 3x `tmt-file-submit` followed by `tmt-report-result` (to report a subresult)? What if the user calls `tmt-report-result` with `-o`?

(I know the TMT interface is just a list of `log` entries, the above is rhetorical. Maybe a small refactor of how CLI scripts handle logs overall might be good, I hit the 1-file limit of the rhts compatibility layer several times now. Even old Beaker HTTP API [supports](https://github.com/RHSecurityCompliance/contest/blob/07f85a5c7fba35b4a90869903be510f9dfbf98fb/lib/results.py#L162-L178) multiple log files per subresult, but restraint also chose to retain the rhts layer.)

psss:
> What if the user calls 3x `tmt-file-submit` followed by `tmt-report-result` (to report a subresult)?

I'd suggest to keep the implementation rather simple: If the `tmt-file-submit` is called, I would add the log file to the main test result, just as an additional log.

> What if the user calls `tmt-report-result` with `-o`?

In this case the `-o` parameter should add the main output of the subresult so this should go (and I believe this is already implemented) to the respective subresult.

> I know the TMT interface is just a list of `log` entries, the above is rhetorical.

The [results specification](https://tmt.readthedocs.io/en/stable/spec/results.html) defines the `log` key for `subresult` as well:

```yaml
subresult:
    # String, name of the test phase.
  - name: First test case

    # String, the effective outcome of the phase execution.
    result: "pass"|"fail"|"info"|"warn"|"error"|"skip"|"pending"

    # String, the original outcome of the phase execution.
    original-result: "pass"|"fail"|"info"|"warn"|"error"|"skip"|"pending"

    # List of strings, optional comments to report with the result.
    note:
      - Things were great.

    # List of strings, paths to log files.
    log:
      - path/to/subresult/log1.txt
      - path/to/subresult/log2.log
        ...
```

[...]

PR #3779, opened 2025-05-28
Add log path for journal.xml to results.yaml

tcornell-bus:
Fixes #3748 

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] extend the test coverage

Issue #1853, opened 2023-02-15
Make `tmt-*` helper scripts available for the `local` provision


sopos:
The local provision method does not prepare all the necessary helper files/scripts for beakerlib to work correctly.

In this particular case the `rlFileSubmit` and `rlBundleLogs` do not work correctly.
```
:: [ 10:59:58 ] :: [   INFO   ] :: Sending /tmp/libssh.tar.gz as tmp-libssh.tar.gz
bash: /usr/local/bin/tmt-file-submit: No such file or directory
:: [ 10:59:58 ] :: [  ERROR   ] :: rlBundleLog: Submit wasn't successful
```

lukaszachy:
Outcome of the discussion on hacking session:

Pack the tmt-* scripts into the tmt.rpm
Document that rstrnt-* scripts not supported
Reboot still is not supposed to be working for provision --how local

psss:
Ran into this problem again [today](https://github.com/teemtee/tmt/pull/3846#discussion_r2177737574), proposing `should` priority and putting to the backlog.

psss:
Missing local tmt scripts together with changes in #3846 result into `journal.xml` not to be included in the logs when the `local` provision method is used. I suggest to fix this asap, raising priority to `must`.

psss:
> Document that rstrnt-* scripts not supported

@lukaszachy, what about the `rhts-*` aliases? Currently the `rhts-report-result` is intentionally used for submitted logs at the end of each `beakerlib` phase:

https://github.com/teemtee/tmt/blob/5bf17e45631244ade9e05a5dd99cc3e7c871e857/tmt/frameworks/beakerlib.py#L84-L87

So, in order to make the `local` provision work, I'm afraid we will have to package `rhts-*` scripts as well. Or is there a better / cleaner approach?

thrix:
slipped to 1.57, unfortunately we could not get it implemented on time, but 1.57 is sure

happz:
Bumped from 1.57, will be reviewed on the planning meeting.

PR #3846, opened 2025-06-27
Add submitted files as logs

bajertom:
Fixes https://github.com/teemtee/tmt/issues/2398 

Pull Request Checklist

* [x] implement the feature
* [ ] adjust plugin docstring
* [ ] include a release note

Should this be covered by tests? Mentioned in the release notes?

Also - while playing with the submitting of files and displaying them, I've found out that journal.xml is copied to $TMT_TEST_DATA from the folder sitting above it. I had to specifically filter this out in the tmt-file-submit script so it wouldn't get displayed at report step twice. Is this behavior intended? 

If not, then moving
```
# Silent skip for journal.xml so it doesn't get saved
# into submitted-files.log and displayed twice at report
basename_file=$(basename "$FILENAME")
if [[ "$basename_file" == "journal.xml" ]]; then
    exit 0
fi
```
above the
```
[ -d "$TMT_TEST_DATA" ] || mkdir -p "$TMT_TEST_DATA"
cp -f "$FILENAME" "$TMT_TEST_DATA"
echo "File '$FILENAME' stored to '$TMT_TEST_DATA'."
```
would fix it.

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -1185,6 +1185,23 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
         # (e.g. saves them as a tmt subresults).
         return invocation.test.test_framework.extract_results(invocation, results, logger)
 
+    def collect_submitted_files(self, invocation: TestInvocation) -> list[Path]:
+        """
+        Collect paths of all files submitted during test
+        """
+
+        submitted_files = []
+        test_data_dir = invocation.test_data_path
+        submission_log = test_data_dir / "submitted-files.log"
+
+        if submission_log.exists():
+            for line in submission_log.read_text().splitlines():
+                file_path = Path(line)
+                relative_path = invocation.relative_test_data_path / file_path.name
+                submitted_files.append(relative_path)
+
+        return submitted_files
+
     def timeout_hint(self, invocation: TestInvocation) -> None:
         """
         Append a duration increase hint to the test output
--- a/tmt/steps/execute/internal.py
+++ b/tmt/steps/execute/internal.py
@@ -648,7 +648,12 @@ class ExecuteInternal(tmt.steps.execute.ExecutePlugin[ExecuteInternalData]):
         # and it's hard to pick the main one, who knows what custom results might
         # cover, so let's make sure every single result can lead to check results
         # related to its lifetime.
+
+        submitted_files = self.collect_submitted_files(invocation)
+
         for result in invocation.results:
+            if result.name == invocation.test.name:
+                result.log.extend(submitted_files)
             result.check = invocation.check_results
 
         return invocation.results
--- a/tmt/steps/execute/scripts/tmt-file-submit
+++ b/tmt/steps/execute/scripts/tmt-file-submit
@@ -31,3 +31,13 @@ fi
 [ -d "$TMT_TEST_DATA" ] || mkdir -p "$TMT_TEST_DATA"
 cp -f "$FILENAME" "$TMT_TEST_DATA"
 echo "File '$FILENAME' stored to '$TMT_TEST_DATA'."
+
+# Silent skip for journal.xml so it doesn't get saved
+# into submitted-files.log and displayed twice at report
+basename_file=$(basename "$FILENAME")
+if [[ "$basename_file" == "journal.xml" ]]; then
+    exit 0
+fi
+
+# Collecting paths of submitted files into a single log
+echo "$FILENAME" >> "$TMT_TEST_DATA/submitted-files.log"


tmt/steps/execute/scripts/tmt-file-submit
```diff
@@ -31,3 +31,13 @@ fi
 [ -d "$TMT_TEST_DATA" ] || mkdir -p "$TMT_TEST_DATA"
 cp -f "$FILENAME" "$TMT_TEST_DATA"
 echo "File '$FILENAME' stored to '$TMT_TEST_DATA'."
+
+# Silent skip for journal.xml so it doesn't get saved
+# into submitted-files.log and displayed twice at report
+basename_file=$(basename "$FILENAME")
+if [[ "$basename_file" == "journal.xml" ]]; then
```
happz: Do you need `basename_file` anywhere else? If not, I'd go for `if [[ "$(basename "$FILENAME")" == "journal.xml" ]]; then` instead.

tmt/steps/execute/__init__.py
```diff
@@ -1185,6 +1185,23 @@ def extract_results(self, invocation: TestInvocation, logger: tmt.log.Logger) ->
         # (e.g. saves them as a tmt subresults).
         return invocation.test.test_framework.extract_results(invocation, results, logger)
 
+    def collect_submitted_files(self, invocation: TestInvocation) -> list[Path]:
```
happz: I would definitely make this a method of `TestInvocation`. Note that the method does not touch `self` at all, on the other hand it needs access to `invocation` fields several times. There should be a `submission_log_path` property added to `TestInvocation` - check out e.g. `check_files_path`, same story - and `collect_submitted_files` should become a property as well: declarative (`submitted_files`) rather than imperative (`collect_submitted_files`).

tmt/steps/execute/scripts/tmt-file-submit
bajertom: Addressed in b31340693d1be146f5506ca8fc4614a26b81dae1.

tmt/steps/execute/__init__.py
bajertom: Addressed in b31340693d1be146f5506ca8fc4614a26b81dae1. 
I wasn't sure about naming of some stuff, like SUBMITTED_FILES_FILENAME for example, but decided to stick with the format of similar variables.

tmt/steps/execute/internal.py
```diff
@@ -648,7 +648,11 @@ def pull_from_guest() -> None:
         # and it's hard to pick the main one, who knows what custom results might
         # cover, so let's make sure every single result can lead to check results
         # related to its lifetime.
+
+        submitted_files = invocation.submitted_files
         for result in invocation.results:
+            if result.name == invocation.test.name:
+                result.log.extend(submitted_files)
```
happz: You can use `invocation.submitted_files` here, directly, without a temporary variable.

tmt/steps/execute/__init__.py
happz: Naming is hard...

tmt/steps/execute/scripts/tmt-file-submit
```diff
@@ -31,3 +31,12 @@ fi
 [ -d "$TMT_TEST_DATA" ] || mkdir -p "$TMT_TEST_DATA"
 cp -f "$FILENAME" "$TMT_TEST_DATA"
 echo "File '$FILENAME' stored to '$TMT_TEST_DATA'."
+
+# Silent skip for journal.xml so it doesn't get saved
+# into submitted-files.log and displayed twice at report
+if [[ "$(basename "$FILENAME")" == "journal.xml" ]]; then
```
happz: Hmhm, which part of tmt is adding `journal.xml` among test logs? I suppose it's the `beakerlib` framework, `for filename in [tmt.steps.execute.TEST_OUTPUT_FILENAME, 'journal.txt', 'journal.xml']` - did you see `journal.txt` gets duplicated as well?

I would maybe drop `journal.xml` from that list in `beakerlib` framework, and also this exceptional handling here: if Beakerlib submits the journal using `tmt-file-submit` (which it does, as you observed), then it would be submitted and collected by the regular flow of test logs, and we do not need it to be handled in a special way.

What I don't understand: I would expect `journal.txt` to suffer the same fat, i.e. be submitted through `tmt-file-submit` *and* added by the list above, appearing twice in the final list of logs. If it's not, then I'm missing something...

tmt/steps/execute/internal.py
bajertom: Resolved

tmt/steps/execute/scripts/tmt-file-submit
bajertom: This approach works and `journal.xml` shows up only once now. However it's strange - as you mentioned, one would expect `journal.txt` to appear in the `submitted-files.log` as well and to get duplicated as well, but it does not happen. Only `journal.xml` gets submitted via `tmt-file-submit`.

tmt/steps/execute/__init__.py
```diff
@@ -278,6 +289,22 @@ def is_guest_healthy(self) -> bool:
...
+
+        submitted_files = []
+
+        if self.submission_log_path.exists():
+            for line in self.submission_log_path.read_text().splitlines():
+                file_path = Path(line)
+                relative_path = self.relative_test_data_path / file_path.name
+                submitted_files.append(relative_path)
+
+        return submitted_files
```
thrix: Maybe this would be shorter?

```suggestion
        if not self.submission_log_path.exists():
            return []
            
        return [
            self.relative_test_data_path / Path(line).name
            for line in self.submission_log_path.read_text().splitlines():
        ]
```

tmt/steps/execute/__init__.py
```diff
@@ -278,6 +289,22 @@ def is_guest_healthy(self) -> bool:
...
+    def submitted_files(self) -> list[Path]:
+        """
+        Paths of all files submitted during test
+        """
+
+        submitted_files = []
+
+        if self.submission_log_path.exists():
+            for line in self.submission_log_path.read_text().splitlines():
+                file_path = Path(line)
```
thrix: I wonder what is the point to convert the `line` to a Path and then using it's name, according to the script the file already containers just a list of filenames anyway? Maybe that coversion is not needed at all?

thrix (changes requested):

tmt/steps/execute/scripts/tmt-file-submit
happz: Interesting. Maybe it's simply not submitted by beakerlib. I'm asking on Slack.

tmt/steps/execute/__init__.py
happz: `submitted_files.append(self.relative_test_data_path / line)` should also work.

tcornell-bus:
A quick test would be nice. I used this test `tests/execute/nonroot` to verify the `submitted-files.log` file was being written where I expected and contained what I expected, so an added check in that test.sh could do it?

Edit: That test is unrelated to this change, it just happened to be using `tmt-file-submit`.

psss:
> Should this be covered by tests?

Yes, at least a basic test would be good.

> Mentioned in the release notes?

Yeah, the change affects which logs are available to the user, so it would good to give a heads-up.

tmt/steps/execute/scripts/tmt-file-submit
psss: > I would maybe drop `journal.xml` from that list in `beakerlib` framework

Including `journal.xml`  among logs as well was introduced just recently in #3779. Do we really want to drop it?

tmt/steps/execute/__init__.py
```diff
@@ -206,6 +209,14 @@ def check_files_path(self) -> Path:
 
         return self.path / CHECK_DATA
 
+    @functools.cached_property
+    def submission_log_path(self) -> Path:
+        """
+        A path to log containing submitted files paths
+        """
+
+        return self.test_data_path / SUBMITTED_FILES_FILENAME
```
psss: The test data path should be reserved solely for stuff produced by the test itself (to prevent any file name collisions). Can we use the test invocation path instead?

tmt/frameworks/beakerlib.py
```diff
@@ -135,7 +135,7 @@ def extract_results(
         note: list[str] = []
         log: list[Path] = [
             invocation.relative_path / filename
-            for filename in [tmt.steps.execute.TEST_OUTPUT_FILENAME, 'journal.txt', 'journal.xml']
+            for filename in [tmt.steps.execute.TEST_OUTPUT_FILENAME, 'journal.txt']
```
psss, line 141: Including `journal.xml` was explicitly requested and imlemented just recently in #3779. Why are we removing it?

tmt/steps/execute/__init__.py
psss: Sounds good.

psss (changes requested):
Thanks for looking into this! Overall looks good. Added a couple questions / suggestions.

tmt/steps/execute/scripts/tmt-file-submit
happz: We learned that Beakerlib submits `journal.xml`, which makes it appear on the list automagically. Framework adding it explicitly makes it appear twice once `tmt-file-submit` adds submitted files as well. I proposed to drop it from the framework, and rely on beakerlib submitting it.

tmt/frameworks/beakerlib.py
happz: See the discussion above: beakerlib explicitly “submits” XML journal, so that one will appear among logs. It does not submit TXT journal, we need to add that one.

@bajertom please, add a comment here, explaining the situation for future engineers.

psss: Hm, I did a quick check:

```
cd tests/execute/basic/data
tmt run --all plan --name beaker
```

And got this in `results.yaml`:

```yaml
  - name: /test/beakerlib/bad
    result: fail
    original-result: fail
    note: []
    log:
      - data/guest/default-0/test/beakerlib/bad-1/output.txt
      - data/guest/default-0/test/beakerlib/bad-1/journal.txt
      - data/guest/default-0/test/beakerlib/bad-1/failures.yaml
    start-time: '2025-07-01T14:08:27.027681+00:00'
    end-time: '2025-07-01T14:08:28.799318+00:00'
```

Seems that `journal.xml` is no more included among the logs.

happz: Can you check the test log? I can reproduce it, but I also see the following in test output:

```
16:17:01                 out: ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
16:17:01                 out: ::   /test/beakerlib/bad
16:17:01                 out: ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
16:17:01                 out:
16:17:01                 out: :: [ 16:17:01 ] :: [   LOG    ] :: Phases fingerprint:  QlMF4l35
16:17:01                 out: :: [ 16:17:01 ] :: [   LOG    ] :: Asserts fingerprint: QlMF4l35
16:17:01                 out: /usr/share/beakerlib/journal.sh: line 270: /usr/local/bin/tmt-file-submit: No such file or directory
16:17:01                 out: :: [ 16:17:01 ] :: [  ERROR   ] :: rlJournalEnd: Submit wasn't successful
16:17:01                 out: ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
```

I guess it's limited to `local` then, which does not install these scripts, and `journal.xml` is not submitted.

tmt/steps/execute/scripts/tmt-file-submit
psss: > Addressed in [b313406](https://github.com/teemtee/tmt/commit/b31340693d1be146f5506ca8fc4614a26b81dae1).

Seems the `if [[ "$(basename "$FILENAME")" == "journal.xml" ]]; then` part got lost somewhere?

tmt/frameworks/beakerlib.py
happz: With `provision -h virtual`:

```
    log:
      - data/guest/default-0/test/beakerlib/bad-1/output.txt
      - data/guest/default-0/test/beakerlib/bad-1/journal.txt
      - data/guest/default-0/test/beakerlib/bad-1/failures.yaml
      - data/guest/default-0/test/beakerlib/bad-1/data/journal.xml
```

```
::   /test/beakerlib/bad

:: [ 14:22:36 ] :: [   LOG    ] :: Phases fingerprint:  QlMF4l35
:: [ 14:22:36 ] :: [   LOG    ] :: Asserts fingerprint: QlMF4l35
File '/var/tmp/tmt/run-322/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/bad-1/journal.xml' stored to '/var/tmp/tmt/run-322/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/bad-1/data'.
```

tmt/steps/execute/scripts/tmt-file-submit
happz: Yes, that's what we are discussing last two days: `journal.xml` appeared in the list twice, because 1. framework added it, 2. Beakerlib submits `journal.xml` too. The condition here was to undo the latter from happening, but I believe it's better for framework to not repeat what Beakerlib already does, and then the condition is not needed...

tmt/frameworks/beakerlib.py
psss: > I guess it's limited to `local` then, which does not install these scripts, and `journal.xml` is not submitted.

Right, with `tmt run --all plan --name beaker provision --how container` it works. So that's problem of the `local` provision only. Which, btw, we should probably fix by packaging the helper scripts in the `rpm`?

happz: Either that, or there was an idea to install them under `/var/tmp/tmt/scripts` or something like that. Then it would work with localhost too.

tmt/steps/execute/__init__.py
bajertom: Changed to test invocation path instead in 92702e42b8154425ae8e310a55c2546187c4ffe9.

tmt/steps/execute/__init__.py
bajertom: I've changed it to this workflow in 92702e42b8154425ae8e310a55c2546187c4ffe9. Shortened also to `self.relative_test_data_path / line`

tmt/steps/execute/__init__.py
bajertom: Resolved in 92702e42b8154425ae8e310a55c2546187c4ffe9.

docs/releases.rst
```diff
@@ -5,6 +5,14 @@
 ======================
 
 
+tmt-1.52.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Files submitted during a test via
+:ref:`tmt-file-submit</stories/features/report-log>` or beakerlib's
```
```suggestion
:ref:`tmt-file-submit</stories/features/report-log>` or Beakerlib's
```

docs/releases.rst
```diff
@@ -5,6 +5,14 @@
...
+rlFileSubmit are now a part of main test result as logs.
```
```suggestion
``rlFileSubmit`` are now a part of main test result as logs.
```

bajertom:
The thing with journal.xml is that Beakerlib submits this file automatically. Now because of this PR it gets collected as a log and appended to results along with all the other submitted files. However due to https://github.com/teemtee/tmt/pull/3779 which explicitly adds this file to logs (because it was missing before), it is now collected twice. 

So the options are either to drop the addition from https://github.com/teemtee/tmt/pull/3779 or to filter the `journal.xml` in `tmt-file-submit` script itself (which Beakerlib uses for `journal.xml` automatically).

bajertom:
Also - I've used `basename "$FILENAME" >> "$TMT_TEST_SUBMITTED_FILES"` in the `tmt-file-submit` script, because just  `$FILENAME` itself gave different results based on whether it was called via `rlFileSubmit` or `tmt-file-submit`. Now it gives just the basename and the results are unified.

docs/releases.rst
bajertom: Fixed in 9995053e964d476fe6f00ab59c439984325a4cb6.

docs/releases.rst
bajertom: Fixed in 9995053e964d476fe6f00ab59c439984325a4cb6.

tmt/frameworks/beakerlib.py
psss: Yeah, proposing to handle this with (at least) the `should` priority:

* https://github.com/teemtee/tmt/issues/1853

tests/execute/filesubmit/test.sh
```diff
@@ -32,8 +33,14 @@ rlJournalStart
...
-        # FIXME - Present this information somehow to the user
-        # (that they have some files...)
+        # Paths of submitted files were logged
+        rlAssertExists "$submission_log_path"
+        rlAssertGrep "this_file.txt" "$submission_log_path"
+        rlAssertGrep "tmp-bundle_name-$tmt_test_serial_number.tar.gz" "$submission_log_path"
+
+	# Show the submitted files in increased verbosity
+	rlAssertGrep "$FILE_PATH" $rlRun_LOG
+	rlAssertGrep "$BUNDLE_PATH" $rlRun_LOG
```
tcornell-bus, line 48: Should the results.yaml file also be checked for the relative paths?

bajertom: Added in 6b6106b9afa4c4cb90c7160580ec84874754409e.

docs/releases.rst
psss: Sorry for nitpicking, but I believe it's supposed to be spelled either `beakerlib` or `BeakerLib`:

 * https://beakerlib.readthedocs.io/en/latest/manual.html

therazix (approved):

tmt/frameworks/beakerlib.py
thrix: @psss @happz ok, I guess we can resolve this then, and there is a follow up for `local`?

tmt/steps/execute/internal.py
```diff
@@ -421,6 +421,7 @@ def _test_environment(
         )
         environment["TMT_TEST_NAME"] = EnvVarValue(invocation.test.name)
         environment["TMT_TEST_DATA"] = EnvVarValue(invocation.test_data_path)
+        environment["TMT_TEST_SUBMITTED_FILES"] = EnvVarValue(invocation.submission_log_path)
```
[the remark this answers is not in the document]
thrix: lol, graphite is getting better :dancers:

tmt/frameworks/beakerlib.py
thrix: @bajertom would it make sense to mention the issue @psss  somewhere in the code as a TODO item, in case somebody would be wondering ?

thrix (approved):
LGTM, fixed the release notes conflict

tmt/steps/execute/scripts/tmt-file-submit
thrix: @psss all good now? The file is not omited in the framework, so this condition is not needed. Resolving. Please reopen if needed.

tmt/steps/execute/scripts/tmt-file-submit
thrix: @psss @happz @bajertom the last thing what is unclear to me is why then https://github.com/teemtee/tmt/pull/3779/files was needed, if it is submitted by default?? Or was the change completely uneeded??

docs/releases.rst
thrix: @bajertom this seems to be a clear fix, just resolve the comments maybe if no more discussion needed next time?

tmt/steps/execute/internal.py
thrix: seems resolved to me.

tmt/steps/execute/__init__.py
thrix: Thanks, looks good

tmt/steps/execute/__init__.py
thrix: looks good, thx

tmt/steps/execute/__init__.py
thrix: @bajertom @psss so `self.path` is the test invocation path? In that case looks resolved?

docs/releases.rst
```diff
@@ -8,6 +8,10 @@
 tmt-1.52.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Files submitted during a test via
+:ref:`tmt-file-submit</stories/features/report-log>` or Beakerlib's
```
thrix:
```suggestion
:ref:`tmt-file-submit</stories/features/report-log>` or BeakerLib's
```

tmt/frameworks/beakerlib.py
```diff
@@ -133,9 +133,12 @@ def extract_results(
 
         # Initialize data, prepare log paths
         note: list[str] = []
+
+        # Beakerlib automatically submits 'journal.xml' and it gets collected as a log, so
```
thrix:
```suggestion
        # BeakerLib automatically submits 'journal.xml' and it gets collected as a log, so
```

docs/releases.rst
thrix: Changed to BeakerLib, both occurances in the cnange. Resolving

tmt/steps/scripts/tmt-file-submit
```diff
@@ -37,3 +37,6 @@ fi
 [ -d "$TMT_TEST_DATA" ] || mkdir -p "$TMT_TEST_DATA"
 cp -f "$FILENAME" "$TMT_TEST_DATA"
 echo "File '$FILENAME' stored to '$TMT_TEST_DATA'."
+
+# Collecting paths of submitted files into a single log
+basename "$FILENAME" >> "$TMT_TEST_SUBMITTED_FILES"
```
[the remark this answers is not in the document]
thrix: Sorry no, this is in purpose. The file is dropped in the framework code, see the conversation hard rock!

tmt/steps/execute/__init__.py
bajertom: Yes, `self.path` is test invocation path, resolved.

bajertom:
Added Milos's improved path handling + graphite nitpicks + put back @thrix's changes (BeakerLib) which I have accidentaly overwritten.

tmt/steps/execute/scripts/tmt-file-submit
happz: The change was needed, because `tmt-file-submit` did not propagate log paths to `Result.log`. That was the missing piece: beakerlib called `tmt-file-submit`, `tmt-file-submit` copied `journal.xml` into test data directory, and then `beakerlib.py` had a list and added `journal.xml` *from invocation path* to the list of result logs. `tmt-file-submit` did not tell anyone which logs/files were submitted. Now it does, therefore the list can drop files beakerlib submits via `rlFileSubmit` or `tmt-file-submit`.

bajertom: In `main`  the `journal.xml` was present in test invocation path + duplicated into `data` directory by Beakerlib. However none of those files path were attached to resullts.log, so #3779 fixed that. 
Then came this PR, which
a) discovered, that `journal.xml` is duplicated in two directories
b) collected submitted files (including the duplicated `journal.xml`) from `data` dir and stored their path in `submitted-files.log` and then appended them to results.log

So then `journal.xml` was in the results logged twice. Another commit in this PR
1.) removed the first logging from #3779
2.) stopped the duplication of `journal.xml` by BeakerLib itself
3.) kept collecting submitted files paths from the `submitted-files.log`

So #3779 was needed and then became non-needed later.

tmt/frameworks/beakerlib.py
psss: I don't think the issue above is related in a special way to this code area. It's just I run into it again.

tmt/steps/execute/scripts/tmt-file-submit
psss: The `journal.xml` file is submitted by `beakerlib` through the `rlFileSubmit` mechanism at the end of the test. Now, that each `rlFileSubmit` call stores the log we don't have to handle it separately here. @thrix, does it make sense?

thrix: Yep, thanks for the input folks, it does completely. Thanks for enlightening me and providing all the context to understand! :hugs:

thrix (approved):
LGTM

thrix:
@bajertom Howdy, pls check on Monday they failures, they look legitimate and need an update I believe because of the changes.

bajertom:
I've fixed a test `tests/execute/result/subresults.sh` to point to proper log path now.

When it comes to the other failing test - `/tests/execute/basic`, scenario `/Check-beakerlib-results` there is an issue with beakerlib running tmt, but the old tmt - that does not have changes from this PR and therefore the `journal.xml` file submission is done the old way - tmt-file-submit duplicated the journal into `data`:
```
File '/tmp/tmp.FhFTKLIqya/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/need-3/journal.xml' stored to '/tmp/tmp.FhFTKLIqya/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/need-3/data'
```
But the "outer" tmt (with PR changes) expects a different behavior - duplication of `journal.xml` blocked by tmt-file-submit, it's path inserted into `submitted-files.log` and from there appended into results.log. And *not* taken from invocation path (which was removed by this PR). However there is no `submitted-files.log` constructed by the "inner" tmt and therefore `journal.xml` is missing in the logs.  

I don't know what is the way forward here. I suppose that these changes would propagate into the "inner" tmt run by beakerlib only after this PR gets merged into main? 

Tagging the experts @happz @psss

psss:
I believe the problem here is that after the change the `local` provision does not create the `journal.xml` because of:

* #1853

Once we have the `tmt-file-submit` script packaged with `tmt` the journal should be submitted as well. For now, I'd suggest to exercise the `/tests/execute/basic` test against `container` and `virtual` provision methods and enable `local` back together with the fix for #1853. Which I propose to rise the priority to get it fixed soon.

psss:
> For now, I'd suggest to exercise the `/tests/execute/basic` test against `container` and `virtual` provision methods and enable `local` back together with the fix for #1853.

Done in aeec6006.

psss:
Failing job is a minor timing issue. Merging.

PR #3848, opened 2025-06-30
Add note on internal check plugins, and mark their docs as reviewed

happz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/docs/plugins/test-checks-header.inc.rst
+++ b/docs/plugins/test-checks-header.inc.rst
@@ -1,0 +1,8 @@
+.. _/plugins/test-checks/internal:
+
+Internal checks
+---------------
+
+Some checks are internal-only, perform mandatory actions, and cannot be
+selected or disabled by user in test metadata. Names of such checks
+begin with ``internal/`` prefix.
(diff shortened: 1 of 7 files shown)


therazix (commented):
The newly added file `tmt/checks/internal/guest.py` should also be included. Otherwise, LGTM.

happz:
> The newly added file `tmt/checks/internal/guest.py` should also be included. Otherwise, LGTM.

Added in ee27825e

docs/plugins/test-checks-header.inc.rst
```diff
@@ -0,0 +1,8 @@
+.. _/plugins/test-checks/internal:
+
+Internal checks
```
psss:
```suggestion
Internal Checks
```

psss (approved):
Just one nitpick. I remember we've discussed it would be nice to collapse all the internal checks under `internal`, similar as other test checks can be expanded. But probably material for separate pull request? Or easy enough to address here?

happz:
> Just one nitpick. I remember we've discussed it would be nice to collapse all the internal checks under `internal`, similar as other test checks can be expanded. But probably material for separate pull request? Or easy enough to address here?

I don't have a smart idea how to do it nicely and quickly. A future PR is what I would prefer.

docs/plugins/test-checks-header.inc.rst
happz: Fixed in 25597550.

psss:
The red job is an irrelevant infra error, merging.

PR #3851, opened 2025-06-30
Fix plugin key rendering for sequences

psss:
When a sequence is rendered, a new line is missing after the generated list which results in the environment variable to be concatenated with the default.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/docs/templates/plugins.rst.j2
+++ b/docs/templates/plugins.rst.j2
@@ -60,8 +60,7 @@
             {% if not actual_default %}
     Default: *not set*
             {% else %}
-    Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor %}
-
+    Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor +%}
             {% endif %}
         {% elif is_enum(actual_default) %}
     Default: ``{{ actual_default.value }}``
--- a/docs/templates/plugins.rst.j2
+++ b/docs/templates/plugins.rst.j2
@@ -59,8 +59,9 @@
         {% elif actual_default is sequence %}
             {% if not actual_default %}
     Default: *not set*
-        {% else %}
+            {% else %}
     Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor %}
+
             {% endif %}
         {% elif is_enum(actual_default) %}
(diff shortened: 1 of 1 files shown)


psss:
Before the fix:

image

After the fix:

image

psss:
@happz, seems a bit weird that the extra line is needed just for the sequence defaults. Any idea why?

happz:
> @happz, seems a bit weird that the extra line is needed just for the sequence defaults. Any idea why?

No idea, I will check the generated ReST file, as there is an empty line just before `Environment`.

Was it all sequences, or just this one? Maybe it’s something special about this particular sequence.

happz:
Aah, I forgot we have `trim_blocks` enabled. Which means, `{% endfor %}` loses its trailing new-line.

But there seem to be more issues, I just noticed the default for a sequence is rendered, but assuming it's a sequence of patterns, so "sequence default" is more broken than just the whitespace.

happz:
> Aah, I forgot we have `trim_blocks` enabled. Which means, `{% endfor %}` loses its trailing new-line.
> 
> But there seem to be more issues, I just noticed the default for a sequence is rendered, but assuming it's a sequence of patterns, so "sequence default" is more broken than just the whitespace.

Nevermind, it's just the blank line, the rest is fine.

@psss so, I would do the following, instead of the extra empty line, I would use `+`, as per https://jinja.palletsprojects.com/en/stable/templates/#whitespace-control

* that `{% else %}` was definitely incorrectly positioned, AFAICT, so it indeed needs to move
* `{% endfor %}` -> `{% endfor +%}` - adding `+` at the end of block is the official way to override `trim_blocks=True` and preserve all `\n` characters after `{% endfor %}` and before `Environment:`.

```diff
diff --git a/docs/templates/plugins.rst.j2 b/docs/templates/plugins.rst.j2
index 626afe0a..77954f6a 100644
--- a/docs/templates/plugins.rst.j2
+++ b/docs/templates/plugins.rst.j2
@@ -59,8 +59,8 @@
         {% elif actual_default is sequence %}
             {% if not actual_default %}
     Default: *not set*
-        {% else %}
-    Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor %}
+            {% else %}
+    Default: {% for default_item in actual_default %}``{{ default_item.pattern | default(default_item) }}``{% if not loop.last %}, {% endif %}{% endfor +%}
             {% endif %}
         {% elif is_enum(actual_default) %}
     Default: ``{{ actual_default.value }}``
```

psss:
Nice, updated in 60c9214d.

happz:
Documentation-only, merging.

PR #3862, opened 2025-07-08
Implement `ref` and `url` options into `shell` prepare plugin

### LINKED ISSUE #3640: Enable `url` and `ref` in the `shell` prepare step
It would be helpful in cases to allow a remote git repo and ref for prepare shell scripts, which is not currently supported

ex:
```
prepare+:
  - name: Module Setup
    how: shell
    url: $external_url
    ref: $external_ref
    script: /some/dir/runtest.sh
```

### Pull Requests

* #3862 
* #3864

mcasquer:
Fix #3640.

Pull Request Checklist

* [x] implement the feature
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -20,7 +20,7 @@ test names that must be discovered and executed during the run.
 
 The :ref:`/plugins/prepare/shell` plugin now supports ``url`` and
 ``ref`` options, so preparation scripts can be fetched directly
-from remote ``Git`` repositories.
+from remote git repositories.
 
 
 tmt-1.53.0
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -18,9 +18,9 @@ The :ref:`discover </plugins/discover>` plugins now support a new
 ``require-test`` key. This key allows users to specify a list of
 test names that must be discovered and executed during the run.
 
-The :ref:`/plugins/prepare/shell` plugin now supports ``url`` and
-``ref`` options, so preparation scripts can be fetched directly
-from remote git repositories.
+The prepare :ref:`/plugins/prepare/shell` plugin now supports
+``url`` and ``ref`` options, so preparation scripts can be fetched
+directly from remote git repositories.
 
 
 tmt-1.53.0
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -18,6 +18,10 @@ The :ref:`discover </plugins/discover>` plugins now support a new
 ``require-test`` key. This key allows users to specify a list of
 test names that must be discovered and executed during the run.
 
+The :ref:`/plugins/prepare/shell` plugin now supports ``url`` and
+``ref`` options, so preparation scripts can be fetched directly
+from remote ``Git`` repositories.
+
 
 tmt-1.53.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/prepare/shell/test.sh
+++ b/tests/prepare/shell/test.sh
@@ -21,7 +21,8 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Remote Script"
-        rlRun "tmt -vvv run provision prepare finish plan -n url" 0 "Prepare using a remote script"
+        rlRun -s "tmt -vvv run provision prepare finish plan -n url" 0 "Prepare using a remote script"
+        rlAssertGrep "Hello world" "$rlRun_LOG"
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/prepare/shell/test.sh
+++ b/tests/prepare/shell/test.sh
@@ -20,6 +20,10 @@ rlJournalStart
             prepare -h shell -s 'touch /tmp/first' -s 'touch /tmp/second'"
     rlPhaseEnd
 
+    rlPhaseStartTest "Remote Script"
+        rlRun "tmt run -rv plan -n url" 0 "Prepare using a remote script"
+    rlPhaseEnd
+
     rlPhaseStartCleanup
         rlRun "popd"
     rlPhaseEnd
--- a/tests/prepare/shell/test.sh
+++ b/tests/prepare/shell/test.sh
@@ -21,7 +21,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Remote Script"
-        rlRun "tmt run -rv plan -n url" 0 "Prepare using a remote script"
+        rlRun "tmt -vvv run provision prepare finish plan -n url" 0 "Prepare using a remote script"
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/prepare/shell/test.sh
+++ b/tests/prepare/shell/test.sh
@@ -21,7 +21,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Remote Script"
-        rlRun -s "tmt -vvv run provision prepare finish plan -n url" 0 "Prepare using a remote script"
+        rlRun -s "tmt -vvv run provision prepare cleanup plan -n url" 0 "Prepare using a remote script"
         rlAssertGrep "Hello world" "$rlRun_LOG"
     rlPhaseEnd
 
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -140,11 +140,11 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
                                 Command('git', 'checkout', '-f', self.data.ref), cwd=repo_path
                             )
 
-                        guest.push(
-                            source=repo_path,
-                            destination=repo_path,
-                            options=["-s", "-p", "--chmod=755"],
-                        )
+                guest.push(
+                    source=repo_path,
+                    destination=repo_path,
+                    options=["-s", "-p", "--chmod=755"],
+                )
 
         if not self.is_dry_run:
             topology = tmt.steps.Topology(self.step.plan.provision.ready_guests)
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -13,7 +13,7 @@ import tmt.utils.git
 from tmt.container import container, field
 from tmt.steps import safe_filename
 from tmt.steps.provision import Guest
-from tmt.utils import Path, ShellScript
+from tmt.utils import EnvVarValue, ShellScript
 
 PREPARE_WRAPPER_FILENAME = 'tmt-prepare-wrapper.sh'
 env_var = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
@@ -113,22 +113,18 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         scripts = self.data.script
         # Give a short summary
         overview = fmf.utils.listed(scripts, 'script')
-        if self.data.url:
-            overview += f" from '{self.data.url}'"
-            if self.data.ref:
-                overview += f" at '{self.data.ref}'"
         logger.info('overview', f'{overview} found', 'green')
 
         workdir = self.step.plan.worktree
         assert workdir is not None  # narrow type
 
-        if env_var in self.step.plan.environment:
-            repo_path = Path(self.step.plan.environment[env_var])
+        if self.data.url:
+            repo_path = workdir / "repository"
+            self.step.plan.environment[env_var] = EnvVarValue(repo_path.resolve())
             if not self.is_dry_run:
                 with self._url_clone_lock:
                     if not repo_path.exists():
                         repo_path.parent.mkdir(parents=True, exist_ok=True)
-                        assert self.data.url is not None
                         tmt.utils.git.git_clone(
                             url=self.data.url,
                             destination=repo_path,
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -120,7 +120,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         assert workdir is not None  # narrow type
 
         if self.data.url:
-            repo_path = workdir / str(self.data.url)
+            repo_path = workdir / "repository"
             if not self.is_dry_run:
                 tmt.utils.git.git_clone(
                     url=self.data.url,
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -1,5 +1,3 @@
-
-import os
 import threading
 from typing import Any, Optional, cast
 
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -13,7 +13,7 @@ import tmt.utils.git
 from tmt.container import container, field
 from tmt.steps import safe_filename
 from tmt.steps.provision import Guest
-from tmt.utils import EnvVarValue, ShellScript
+from tmt.utils import Command, EnvVarValue, ShellScript
 
 PREPARE_WRAPPER_FILENAME = 'tmt-prepare-wrapper.sh'
 env_var = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
@@ -133,6 +133,12 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
                             env=environment,
                             logger=self._logger,
                         )
+
+                        if self.data.ref:
+                            self.info('ref', self.data.ref, 'green')
+                            self.run(
+                                Command('git', 'checkout', '-f', self.data.ref), cwd=repo_path
+                            )
 
                         guest.push(
                             source=repo_path,
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -16,7 +16,7 @@ from tmt.steps.provision import Guest
 from tmt.utils import Command, EnvVarValue, ShellScript
 
 PREPARE_WRAPPER_FILENAME = 'tmt-prepare-wrapper.sh'
-env_var = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
+ENV_VAR = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
 
 
 @container
@@ -36,7 +36,12 @@ class PrepareShellData(tmt.steps.prepare.PrepareStepData):
         default=None,
         option='--url',
         metavar='REPOSITORY',
-        help='Git repository URL for fetching shell scripts.',
+        help="""
+            URL of a repository to clone. It will be pushed to guests before
+            running any scripts,
+            and environment variable ``TMT_PREPARE_SHELL_URL_REPOSITORY`` will
+            hold its path on the guest.
+            """,
     )
 
     ref: Optional[str] = field(
@@ -44,8 +49,8 @@ class PrepareShellData(tmt.steps.prepare.PrepareStepData):
         option='--ref',
         metavar='REVISION',
         help="""
-            Branch, tag or commit specifying the desired git
-            revision.
+            Branch, tag or commit to checkout in the git repository
+            cloned when ``url`` is specified.
             """,
     )
 
@@ -79,8 +84,8 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
     Scripts can also be fetched from a remote git repository.
     Specify the ``url`` for the repository and optionally ``ref``
     to checkout a specific branch, tag or commit.
-    The ``script`` paths will then be treated as relative to the
-    repository root.
+    ``TMT_PREPARE_SHELL_URL_REPOSITORY`` will hold the value of the
+    repository path.
 
     .. code-block:: yaml
 
@@ -88,8 +93,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
             how: shell
             url: https://github.com/teemtee/tmt.git
             ref: main
-            script:
-              - tmt/steps/prepare/test.sh
+            script: cd $TMT_PREPARE_SHELL_URL_REPOSITORY && make docs
     """
 
     _data_class = PrepareShellData
@@ -120,7 +124,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         if self.data.url:
             repo_path = workdir / "repository"
 
-            environment[env_var] = EnvVarValue(repo_path.resolve())
+            environment[ENV_VAR] = EnvVarValue(repo_path.resolve())
 
             if not self.is_dry_run:
                 with self._url_clone_lock:
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -16,7 +16,6 @@ from tmt.steps.provision import Guest
 from tmt.utils import Command, EnvVarValue, ShellScript
 
 PREPARE_WRAPPER_FILENAME = 'tmt-prepare-wrapper.sh'
-ENV_VAR = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
 
 
 @container
@@ -98,6 +97,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
 
     _data_class = PrepareShellData
     _url_clone_lock = threading.Lock()
+    _cloned_repo_path_envvar_name = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
 
     def go(
         self,
@@ -124,7 +124,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         if self.data.url:
             repo_path = workdir / "repository"
 
-            environment[ENV_VAR] = EnvVarValue(repo_path.resolve())
+            environment[self._cloned_repo_path_envvar_name] = EnvVarValue(repo_path.resolve())
 
             if not self.is_dry_run:
                 with self._url_clone_lock:
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -3,6 +3,7 @@ from typing import Any, Optional, cast
 import fmf
 import fmf.utils
 
+import threading
 import tmt
 import tmt.log
 import tmt.steps
@@ -121,14 +122,18 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
 
         if self.data.url:
             repo_path = workdir / "repository"
+            _url_clone_lock = threading.Lock()
             if not self.is_dry_run:
-                tmt.utils.git.git_clone(
-                    url=self.data.url,
-                    destination=repo_path,
-                    shallow=False,
-                    env=environment,
-                    logger=self._logger,
-                )
+                with _url_clone_lock:
+                    if not repo_path.exists():
+                        repo_path.parent.mkdir(parents=True, exist_ok=True)
+                        tmt.utils.git.git_clone(
+                            url=self.data.url,
+                            destination=repo_path,
+                            shallow=False,
+                            env=environment,
+                            logger=self._logger,
+                        )
 
             scripts = [ShellScript(str(repo_path.joinpath(str(script)))) for script in scripts]
 
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -29,6 +29,23 @@ class PrepareShellData(tmt.steps.prepare.PrepareStepData):
         unserialize=lambda serialized: [ShellScript(script) for script in serialized],
     )
 
+    url: Optional[str] = field(
+        default=None,
+        option='--url',
+        metavar='REPOSITORY',
+        help='Git repository URL for fetching shell scripts.',
+    )
+
+    ref: Optional[str] = field(
+        default=None,
+        option='--ref',
+        metavar='REVISION',
+        help="""
+            Branch, tag or commit specifying the desired git
+            revision.
+            """,
+    )
+
     # ignore[override] & cast: two base classes define to_spec(), with conflicting
     # formal types.
     def to_spec(self) -> dict[str, Any]:  # type: ignore[override]
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -120,7 +120,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         assert workdir is not None  # narrow type
 
         if self.data.url:
-            repo_path = workdir / str(self.data.url.split('/')[-1])
+            repo_path = workdir / str(self.data.url)
             if not self.is_dry_run:
                 tmt.utils.git.git_clone(
                     url=self.data.url,
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -110,9 +110,8 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
 
         environment = environment or tmt.utils.Environment()
 
-        scripts = self.data.script
         # Give a short summary
-        overview = fmf.utils.listed(scripts, 'script')
+        overview = fmf.utils.listed(self.data.script, 'script')
         logger.info('overview', f'{overview} found', 'green')
 
         workdir = self.step.plan.worktree
@@ -120,7 +119,9 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
 
         if self.data.url:
             repo_path = workdir / "repository"
-            self.step.plan.environment[env_var] = EnvVarValue(repo_path.resolve())
+
+            environment[env_var] = EnvVarValue(repo_path.resolve())
+
             if not self.is_dry_run:
                 with self._url_clone_lock:
                     if not repo_path.exists():
@@ -160,7 +161,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         logger.debug('prepare wrapper', prepare_wrapper_path, level=3)
 
         # Execute each script on the guest (with default shell options)
-        for script in scripts:
+        for script in self.data.script:
             logger.verbose('script', script, 'green')
             script_with_options = tmt.utils.ShellScript(f'{tmt.utils.SHELL_OPTIONS}; {script}')
             self.write(prepare_wrapper_path, str(script_with_options), 'w')
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -1,3 +1,5 @@
+
+import os
 import threading
 from typing import Any, Optional, cast
 
@@ -13,9 +15,10 @@ import tmt.utils.git
 from tmt.container import container, field
 from tmt.steps import safe_filename
 from tmt.steps.provision import Guest
-from tmt.utils import ShellScript
+from tmt.utils import Path, ShellScript
 
 PREPARE_WRAPPER_FILENAME = 'tmt-prepare-wrapper.sh'
+env_var = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
 
 
 @container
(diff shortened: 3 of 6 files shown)


mcasquer:
The current implementation is failing when executing the test, perhaps related with the `repo_path` where the scripts are being stored?

```bash
Cause number 1:

        Command '/var/tmp/tmt/run-051/url/tree/tmt-prepare-wrapper.sh-default-0-default-0' returned 127.

        # stderr (1 lines)
        /var/tmp/tmt/run-051/url/tree/tmt-prepare-wrapper.sh-default-0-default-0: line 1: /var/tmp/tmt/run-051/url/tree/tmt.git/tests/prepare/shell/remote/test.sh: No such file or directory
```

mcasquer:
@coderabbitai review

tmt/steps/prepare/shell.py
```diff
@@ -75,13 +108,31 @@ def go(
...
+            overview += f" from '{self.data.url}'"
+            if self.data.ref:
+                overview += f" at '{self.data.ref}'"
         logger.info('overview', f'{overview} found', 'green')
 
         workdir = self.step.plan.worktree
         assert workdir is not None  # narrow type
 
+        if self.data.url:
+            repo_path = workdir / str(self.data.url.split('/')[-1])
```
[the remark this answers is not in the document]
happz: Please, use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

tmt/steps/prepare/shell.py
```diff
@@ -75,13 +108,31 @@ def go(
...
+            if not self.is_dry_run:
+                tmt.utils.git.git_clone(
+                    url=self.data.url,
+                    destination=repo_path,
+                    shallow=False,
+                    env=environment,
+                    logger=self._logger,
+                )
+
+            scripts = [ShellScript(str(repo_path.joinpath(str(script)))) for script in scripts]
```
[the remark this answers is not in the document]
happz: `repo_path` is `Path` instance, I'd say the merge can be simplified a bit: `scripts = [ShellScript(str(repo_path / script)) for script in scripts]`

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
mcasquer: If a repository is used, such as `https://github.com/teemtee/tmt.git` after `urlparse` I should have `/teemtee/tmt.git` at the `path` attribute, is that a valid path? need I to apply the split anyways? Could be this the reason of the failure I commented above?

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
mcasquer: Getting this error when trying that way
`fail: unsupported operand type(s) for /: 'Path' and 'ShellScript'`

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
happz: That is a valid path.

But thinking about it, do we even need that name? There will always be at most one such URL in a given `prepare` phase, there will never be two or more:

```yaml
# Not supported:
prepare:
  - how: shell
    url:
      - foo
      - bar

# Supported:
prepare:
  - how: shell
    url: foo

  - how: shell
    url: bar
```

So you can check out into `workdir / 'repository'` and it should work, the repository path does not need to bear relation to the URL, because there is no need for it to be unique - there will be no `git clone some-other-url` in this phase's `workdir`.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
happz: Hm, true, they are already converted... Shouldn't we add new key, e.g. `path`, for this? because what would happen in this scenario:

```yaml
prepare:
  how: shell
  url: https://foo.bar.baz
  script: ls -al /
```

The assumption is that when `url` is used, `script` is treated as a list of paths. What if they are not? `/local/repo/path/ls -al /` is hardly a valid path to a script.

@psss care to chime in? How about adding e.g. `script-path`? It can work in a way similar to `script` when `url` is not set, and become the only consumed `script*` key when `url` is set, to avoid messy paths from above. And `shell` can emit a warning when both `url` and `script` would be set.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
mcasquer: Updated, using the `url` directly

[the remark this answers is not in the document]
happz: Please, don't. Imagine `url: https://gitlab.com/redhat/centos-stream/tests/binutils.git`, you will end up with a really weird directory structure, `$workdir/https:/gitlab.com/redhat/centos-stream/tests/binutils.git`, with all the levels, including `https:`. Either use a sanitized value of the URL, `tmt.utils.sanitize_name(self.data.url, allow_slash=False)`, or use the same name every time - as mentioned above, there should be no conflict, every phase has its own workdir and every phase can have 1 URL at most, therefore `workdir / 'repository'` would be as safe as URL, but maybe more readable.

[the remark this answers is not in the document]
mcasquer: Ohhh I misunderstood 'repository' intention as a literal, sorry for that.

tmt/steps/prepare/shell.py
```diff
@@ -75,13 +108,31 @@ def go(
...
+            overview += f" from '{self.data.url}'"
+            if self.data.ref:
+                overview += f" at '{self.data.ref}'"
         logger.info('overview', f'{overview} found', 'green')
 
         workdir = self.step.plan.worktree
         assert workdir is not None  # narrow type
 
+        if self.data.url:
+            repo_path = workdir / "repository"
```
LecrisUT, line 125: This part seems fragile. It could break if you have multiple prepare steps that do the git clone. Not sure what alternative convention to suggest though.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
LecrisUT: If we go with `path` keyword, wouldn't it make sense to have a different `how` instead? I am comparing it to how `discover.fmf` got so loaded, and would be nice to avoid that.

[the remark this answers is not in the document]
psss: My expectation was that the `script` field will keep the same functionality, even for the case when `url` is defined, that is `Shell script to be executed.`, not a path to the script inside the fetched repository. But I understand that we have to somehow provide the `path` to the fetched repo so that users can apply it in their scripts. What about one of these options?

* provide a `destination` or `target` key which will point to the directory where the git should be fetched, then it would be up to the user to pick a good name not to collide with possible other `prepare` steps phases and could is when constructing their scripts, e.g. `./target/prepare/script.sh`
* clone the git repository into the prepare phase workdir (which is unique for each phase, so the name collisions would be resolved) and provide an environment variable, let's say `TMT_PREPARE_REPO`, which then could be used inside user scripts, e.g. `$TMT_PREPARE_REPO/prepare/script.sh`
* another option could be to `cd` to the cloned git repo and document that behaviour, then users could use relative path from the git root to get to their desired scripts, e.g. `./prepare/script.sh`

@BeeGrech, any preference or expectations from your side?

tmt/steps/prepare/shell.py
happz: Hmmm, actually, yes, it's not safe. Sorry, @mcasquer, everything I said was true, there is indeed just one `url` for the phase, but the problem is that the phase may run against multiple guests. So, multiple invocations of the single phase will run at the same moment, cloning into the same location. And some of them will definitely fail.

We have two options:

* add guest name to the path, that's common, `workdir / guest.safe_name / 'repository'`. Safe. But we end up with several copies of the same repo. Wasting space and time.
* add a class-level lock, and use it to guard the `git clone`:

```python
# Class variable:
_url_clone_lock = threading.Lock()

# In the code:
if not self.is_dry_run:
    with self._url_clone_lock:
        if not repo_path.exists():
            repo_path.parent.mkdir(parents=True, exists_ok=True)

            tmt.utils.git.git_clone(
                url=self.data.url,
                destination=repo_path,
                shallow=False,
                env=environment,
                logger=self._logger,
            )
```

That should serialize the cloning operation, leaving us with just a single clone of the repo. It's probably not going to be a bottleneck, other invocations of the phase, for other guests, would need to clone anyway, and once the first invocation is done and releases the lock, they can easily continue.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
happz: Do we - or user - need to provide path to the cloned repository? That's not what I asked about, we already know that, we can pick the path when we clone the repo, user does not need to be involved.

My question was about what to do when `url` is set, we clone the repo, and `script` is e.g. `ls -al /var`. The functionality as implemented in this PR will treat `script` as "path to a script to be executed" rather than "shell script to be executed", and will construct `/cloned/repo/path/ls -al /var`, which is, if nothing else, ugly and puzzling. Hence my question, shouldn't we rather add new key, e.g. `script-path`, to be used in tandem with `url`, rather than assuming `script` is a path-like string?

tmt/steps/prepare/shell.py
LecrisUT: We could use a hash of the url+ref. Multi guest access would be tricky to account for though.

happz: For a given `prepare/shell` phase in a plan, there will be only a single repository to clone. Different repository means different phase, which means different workdir. So we will not have any conflict of multiple repositories fighting over `workdir / repository`. What we will have is conflicts of one phase being invoked on multiple guests, each invocation will try to `git clone` on the tmt runner first. After that, the use is read-only, basically just `cd workdir/repository` & run a given script. So all we need to prevent is multiple threads running `git clone $url $workdir/repository` at the same time. I would use the lock, to run it just once, and share the clone across the invocations of one `shell` phase.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
psss: > The functionality as implemented in this PR will treat `script` as "path to a script to be executed" rather than "shell script to be executed"

I believe the current implementation should be changed and keep `script` field behaviour as it was before, that is make `ls -la /var` working nicely, with no change.

> Hence my question, shouldn't we rather add new key, e.g. script-path, to be used in tandem with url, rather than assuming script is a path-like string?

If possible I would prevent adding a new `script-path` key as to me it sounds a bit confusing (having two fields for providing scripts which should be executed) and does not bring much benefit.

Thus I'm rather thinking about how to combine the existing `script` key with the newly cloned repo so that users can do any shell magic they need, e.g. `cd $TMT_PREPARE_REPO/scripts; for script in prepare*; do ./$script; done` or similar.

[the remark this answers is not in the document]
LecrisUT: Rereading the comments, I think this approach fits best?

> Shouldn't we add new key, e.g. `path`, for this?

> * another option could be to `cd` to the cloned git repo and document that behaviour, then users could use relative path from the git root to get to their desired scripts, e.g. `./prepare/script.sh`

We already do that for the tests.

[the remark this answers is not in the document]
happz: @psss again, `cd $TMT_PREPARE_REPO/scripts; for script in prepare*; do ./$script; done` will not save us from `script: ls -al /`. The assumption that *`script` is a path* is incorrect, user is free to put any executable shell script into `script`. I don't insist we should add `script-path`, fine, let's stick with `script` alone, I just care about tmt trying to run `./ls -al /`. If nothing else, there should be a strong check verifying `script` is in the repository, and that it is a valid path, and fail if it's not true.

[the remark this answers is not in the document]
psss: Hmmm, seems there is a misunderstanding?

> The assumption that script is a path is incorrect

I completely agree and suggest to change the implementation in this pull request so that `script` is not considered to be a path but handle it as full shell script as before.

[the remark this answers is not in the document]
happz: So `for script in prepare*; do $script; done`, not `for script in prepare*; do ./$script; done` from your comment?

BTW we need a push, to push the cloned repository to guest, before we `cd` and execute stuff in it.

tmt/steps/prepare/shell.py
```diff
@@ -74,13 +108,35 @@ def go(
...
+            _url_clone_lock = threading.Lock()
```
happz: This must be either module-level or class-level variable. It must be accessible to all invocations of `shell`. An in-function, local variable isn't visible to other threads, isn't shared, and wouldn't help with the synchronization.

mcasquer: My bad, code updated properly

mcasquer:
Still hitting the script not being found issue when running the test
`/var/tmp/tmt/run-061/url/tree/tmt-prepare-wrapper.sh-default-0-default-0: line 1: /var/tmp/tmt/run-061/url/tree/repository/tests/prepare/shell/remote/test.sh: No such file or directory`

happz:
> Still hitting the script not being found issue when running the test `/var/tmp/tmt/run-061/url/tree/tmt-prepare-wrapper.sh-default-0-default-0: line 1: /var/tmp/tmt/run-061/url/tree/repository/tests/prepare/shell/remote/test.sh: No such file or directory`

I believe that's expected: the `git clone` runs on the runner, where tmt is invoked, and therefore it does not exist on the guests. But `shell` tries to execute it there. So we need `guest.push()` and upload the repository to its remote location.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
happz: @mcasquer so, after the discussion with @psss , we came up with the following:

* `scripts = [ShellScript(str(repo_path.joinpath(str(script)))) for script in scripts]` cannot work, as `script` may be and shall remain an actual shell script, e.g. `ls /` - treating it as a path cannot work, i.e. mangling a script into a path with `repo_path` is not the way to go.
* from the options in https://github.com/teemtee/tmt/pull/3862/files#discussion_r2208286683, we picked the environment variable - `cd` to repository before running the given script could make assumptions of the script (maybe it's not even supposed to run in the `repo_path`). Instead, we should expose an environment variable, e.g. `TMT_PREPARE_SHELL_URL_REPOSITORY` (naming is hard...) which we would set to `repo_path`, and then we would simply run the script as we do now. If the script needs to run in the cloned repo, user should add `cd $TMT_PREPARE_SHELL_URL_REPOSITORY` on their own.

TL;DR, `scripts` remains as it is, we do not need to modify it, we just document for users that there is new envvar and if they want their script to run in the repo, the should `cd` first; and we're missing the `guest.push()` to upload the repo to guest before we start running scripts.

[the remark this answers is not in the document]
mcasquer: @happz so the `guest.push` from:
https://github.com/teemtee/tmt/blob/c41959ec4ef86fa9d3bb5a3a43fa987b5fd453a5/tmt/steps/prepare/shell.py#L111
doesn't mean exactly that? I see the execute is 10 lines below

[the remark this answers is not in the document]
happz: Yeah, but read the arguments:

```python
            guest.push(
                source=prepare_wrapper_path,
                destination=prepare_wrapper_path,
            )
```

It's pushing a single file, unrelated to the cloned repository. So, yeah, there is a `guest.push()` call, but it does nothing for the cloned repo.

mcasquer:
There's still something wrong in my code, if the `TMT_PREPARE_SHELL_URL_REPOSITORY` doesn't exist the prepare step finish successfully but the test fails with no traceback, only a message.

```bash
        how: shell
        overview: 1 script from 'https://github.com/teemtee/tmt.git' at 'main' found
        script: cd /var/tmp/prepare_url && echo "Hello world"
    
```
```bash
total: 1 error
:: [ 12:34:51 ] :: [   FAIL   ] :: Prepare using a remote script (Expected 0, got 2)
```
If the directory already exists, I get the following error which could be some permission issue
`/var/tmp/tmt/run-107/url/tree/tmt-prepare-wrapper.sh-default-0-default-0: line 1: cd: /var/tmp/prepare_url: No such file or directory`

happz:
> There's still something wrong in my code, if the `TMT_PREPARE_SHELL_URL_REPOSITORY` doesn't exist the prepare step finish successfully but the test fails with no traceback, only a message.
> 
> ```shell
>  prepare task #2: default-0 on default-0
>         how: shell
>         overview: 1 script from 'https://github.com/teemtee/tmt.git' at 'main' found
>         script: cd /var/tmp/prepare_url && echo "Hello world"
>     
>         queued pull task #1: pull from default-0
>         
>         pull task #1: pull from default-0
>     
>         summary: 2 preparations applied
> ```
> 
> ```shell
> total: 1 error
> :: [ 12:34:51 ] :: [   FAIL   ] :: Prepare using a remote script (Expected 0, got 2)
> ```
> 
> If the directory already exists, I get the following error which could be some permission issue `/var/tmp/tmt/run-107/url/tree/tmt-prepare-wrapper.sh-default-0-default-0: line 1: cd: /var/tmp/prepare_url: No such file or directory`

There is misunderstanding: `TMT_PREPARE_SHELL_URL_REPOSITORY` is not for user to set, it's not up to user to select the location. tmt picks the path - `workdir / 'repository'`, and tmt adds `TMT_PREPARE_SHELL_URL_REPOSITORY` to the environment when running individual scripts.

mcasquer:
> There is misunderstanding: `TMT_PREPARE_SHELL_URL_REPOSITORY` is not for user to set, it's not up to user to select the location. tmt picks the path - `workdir / 'repository'`, and tmt adds `TMT_PREPARE_SHELL_URL_REPOSITORY` to the environment when running individual scripts.

Ok, my fault. How does it look it now? The test keeps failing but now no matter how many executions I just get the message at the end but prepare apparently passes
```bash
total: 1 error
:: [ 13:42:17 ] :: [   FAIL   ] :: Prepare using a remote script (Expected 0, got 2)
```

happz:
> > There is misunderstanding: `TMT_PREPARE_SHELL_URL_REPOSITORY` is not for user to set, it's not up to user to select the location. tmt picks the path - `workdir / 'repository'`, and tmt adds `TMT_PREPARE_SHELL_URL_REPOSITORY` to the environment when running individual scripts.
> 
> Ok, my fault. How does it look it now? The test keeps failing but now no matter how many executions I just get the message at the end but prepare apparently passes
> 
> ```shell
> total: 1 error
> :: [ 13:42:17 ] :: [   FAIL   ] :: Prepare using a remote script (Expected 0, got 2)
> ```

I see now where the problem was. You already have environment to add the variable to, `environment` parameter is exactly that. Modifying plan itself will not help, as that one is pretty much read-only.

So, I would do the following changes:

```
diff --git a/tests/prepare/shell/test.sh b/tests/prepare/shell/test.sh
index d9ae4b35..508d1e69 100755
--- a/tests/prepare/shell/test.sh
+++ b/tests/prepare/shell/test.sh
@@ -21,7 +21,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Remote Script"
-        rlRun "tmt run -rv plan -n url" 0 "Prepare using a remote script"
+        rlRun "tmt -vvv run provision prepare finish plan -n url" 0 "Prepare using a remote script"
     rlPhaseEnd
 
     rlPhaseStartCleanup
diff --git a/tmt/steps/prepare/shell.py b/tmt/steps/prepare/shell.py
index 44d7dd81..741ef57f 100644
--- a/tmt/steps/prepare/shell.py
+++ b/tmt/steps/prepare/shell.py
@@ -110,9 +110,8 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
 
         environment = environment or tmt.utils.Environment()
 
-        scripts = self.data.script
         # Give a short summary
-        overview = fmf.utils.listed(scripts, 'script')
+        overview = fmf.utils.listed(self.data.script, 'script')
         logger.info('overview', f'{overview} found', 'green')
 
         workdir = self.step.plan.worktree
@@ -120,7 +119,9 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
 
         if self.data.url:
             repo_path = workdir / "repository"
-            self.step.plan.environment[env_var] = EnvVarValue(repo_path.resolve())
+
+            environment[env_var] = EnvVarValue(repo_path.resolve())
+
             if not self.is_dry_run:
                 with self._url_clone_lock:
                     if not repo_path.exists():
@@ -160,7 +161,7 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
         logger.debug('prepare wrapper', prepare_wrapper_path, level=3)
 
         # Execute each script on the guest (with default shell options)
-        for script in scripts:
+        for script in self.data.script:
             logger.verbose('script', script, 'green')
             script_with_options = tmt.utils.ShellScript(f'{tmt.utils.SHELL_OPTIONS}; {script}')
             self.write(prepare_wrapper_path, str(script_with_options), 'w')
```

mcasquer:
> So, I would do the following changes:

`:: [ 08:05:08 ] :: [   PASS   ] :: Prepare using a remote script (Expected 0, got 0)`

Definitely that worked, thanks! :grin: :grin:

tmt/steps/prepare/shell.py
```diff
@@ -81,6 +117,29 @@ def go(
...
+                with self._url_clone_lock:
+                    if not repo_path.exists():
+                        repo_path.parent.mkdir(parents=True, exist_ok=True)
+                        tmt.utils.git.git_clone(
+                            url=self.data.url,
+                            destination=repo_path,
+                            shallow=False,
+                            env=environment,
+                            logger=self._logger,
+                        )
```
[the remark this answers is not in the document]
mcasquer: @happz adding the following block from discover shell should handle the reference part right?

```
# Resolve possible dynamic references
try:
    ref = tmt.base.resolve_dynamic_ref(
        logger=self._logger, workdir=repo_path, ref=self.data.ref, plan=self.step.plan
    )
except tmt.utils.FileError as error:
    raise tmt.utils.DiscoverError(str(error))

if ref:
    self.info('ref', ref, 'green')
    self.debug(f"Checkout ref '{ref}'.")
    self.run(Command('git', 'checkout', '-f', ref), cwd=repo_path)
```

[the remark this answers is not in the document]
happz: It's not needed, that one's for *dynamic* references. Here we have a static one, from the user, so if it's set, you just go ahead and `self.run(Command('git', 'checkout', '-f', ref), cwd=repo_path)`.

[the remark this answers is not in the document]
mcasquer: Code updated, thanks !

tmt/steps/prepare/shell.py
```diff
@@ -81,6 +117,35 @@ def go(
...
+                            self.info('ref', self.data.ref, 'green')
+                            self.run(
+                                Command('git', 'checkout', '-f', self.data.ref), cwd=repo_path
+                            )
+
+                        guest.push(
+                            source=repo_path,
+                            destination=repo_path,
+                            options=["-s", "-p", "--chmod=755"],
+                        )
```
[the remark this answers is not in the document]
mcasquer: If I am not wrong, the `workdir` directory should be unique for each `prepare` step, this fact plus the thread lock should avoid any race condition or duplicated directory

tmt/steps/prepare/shell.py
martinhoyer: +1 for the @happz approach, but wouldn't the fact that the `gust.push()` is in `if not repo_path.exists()` block basically only push it to the one guest that would be first? (move outside the `with` block?)

martinhoyer: hmm, maybe even the `checkout` should be moved outside the block? I'm not sure, but perhaps some fringe cases like user modifying the `ref` and running tmt run -l, or something like that. I at least assume git checkout has such a small overhead that we don't really need to try to minimize it's usage?

happz: The checkout shouldn't be moved outside of `with`, it should happen while holding the lock. Otherwise, multiple threads may run the same `git checkout` command - which, probably, is not a genuinely bad thing, so it **can** be moved outside of `with`, but it's a bad habit to alow multiple threads to perform the same singleton task, it's a needles duplication, and who knows what can happen.

`guest.push()`, on the other hand, must happen outside of `with`, exactly because it must happen for all guests, not just for the one responsible for `git clone` and `git checkout`. All must get their own local copy of the repository, and all can do it in parallel as the clone is now present and basically read-only.

tmt/steps/prepare/shell.py
```diff
@@ -75,13 +108,31 @@ def go(
...
+        if self.data.url:
+            repo_path = workdir / str(self.data.url.split('/')[-1])
+            if not self.is_dry_run:
+                tmt.utils.git.git_clone(
+                    url=self.data.url,
+                    destination=repo_path,
+                    shallow=False,
+                    env=environment,
+                    logger=self._logger,
+                )
```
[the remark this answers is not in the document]
martinhoyer: not sure about this. @happz @LecrisUT Could you please point us to some best practices around catching exceptions like this in the tmt codebase?  (which we can also give rabbit&friends so it has better feedback going forward)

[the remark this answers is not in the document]
happz: Wrapping the operation with `try/except` makes sense if we can provide more context. If `git_clone()` fails, it will produce an exception - can we provide a better error message, more details? Will it help user to resolve the problem? `git_clone()` is probably going to raise a `RunError`, URL and ref will be mentioned, command will be there, but it might be puzzling for the user. `raise `PrepareError("Failed to clone script repository '{self.data.url}@{self.data.ref}'.") from error` could be better for user to ingest rather than `git clone ... command failed`.

tmt/steps/prepare/shell.py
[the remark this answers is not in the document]
happz: Correct, there will be just a single url and ref in any given workdir.

tmt/steps/prepare/shell.py
mcasquer: @martinhoyer thanks for the review, I'll move `guest.push` out of the `with` block

mcasquer: Updated !

docs/releases.rst
```diff
@@ -4,6 +4,14 @@
     Releases
 ======================
 
+tmt-1.54.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`/plugins/prepare/shell` plugin now supports ``url`` and
+``ref`` options, so preparation scripts can be fetched directly from
+remote ``Git`` repositories.
```
happz: I believe "git" alone should be enough, no need to wrap with backticks as it's not a literal.

tmt/steps/prepare/shell.py
```diff
@@ -8,12 +9,14 @@
...
 import tmt.utils
+import tmt.utils.git
 from tmt.container import container, field
 from tmt.steps import safe_filename
 from tmt.steps.provision import Guest
-from tmt.utils import ShellScript
+from tmt.utils import Command, EnvVarValue, ShellScript
 
 PREPARE_WRAPPER_FILENAME = 'tmt-prepare-wrapper.sh'
+env_var = 'TMT_PREPARE_SHELL_URL_REPOSITORY'
```
happz: The name should be uppercased, i.e. `ENV_VAR`, as it's a module-level "constant". In any case, I suspect it may become a class-level variable, so we can parameterize it if we decide to have different one in `prepare` and `finish`.

tmt/steps/prepare/shell.py
```diff
@@ -29,6 +32,23 @@ class PrepareShellData(tmt.steps.prepare.PrepareStepData):
...
+        metavar='REPOSITORY',
+        help='Git repository URL for fetching shell scripts.',
+    )
+
+    ref: Optional[str] = field(
+        default=None,
+        option='--ref',
+        metavar='REVISION',
+        help="""
+            Branch, tag or commit specifying the desired git
```
happz: How about "Branch, tag or commit to check out in the git repository cloned when ``url`` is specified." ?

tmt/steps/prepare/shell.py
```diff
@@ -29,6 +32,23 @@ class PrepareShellData(tmt.steps.prepare.PrepareStepData):
         unserialize=lambda serialized: [ShellScript(script) for script in serialized],
     )
 
+    url: Optional[str] = field(
+        default=None,
+        option='--url',
+        metavar='REPOSITORY',
+        help='Git repository URL for fetching shell scripts.',
```
happz: How about "URL of a repository to clone. It will be pushed to guests before running any scripts, and environment variable FOO will hold its path on the guest." ?

tmt/steps/prepare/shell.py
```diff
@@ -55,9 +75,25 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
               - sudo dnf install -y 'dnf-command(copr)'
               - sudo dnf copr enable -y psss/tmt
               - sudo dnf install -y tmt
+
+    Scripts can also be fetched from a remote git repository.
+    Specify the ``url`` for the repository and optionally ``ref``
+    to checkout a specific branch, tag or commit.
+    The ``script`` paths will then be treated as relative to the
+    repository root.
```
happz: No, they will not, scripts being relative is no longer in play.

tmt/steps/prepare/shell.py
```diff
@@ -55,9 +75,25 @@ class PrepareShell(tmt.steps.prepare.PreparePlugin[PrepareShellData]):
...
+
+    .. code-block:: yaml
+
+        prepare:
+            how: shell
+            url: https://github.com/teemtee/tmt.git
+            ref: main
+            script:
+              - tmt/steps/prepare/test.sh
```
happz: As proposed, this example should use the environment variable and `cd` to demonstrate the repository presence.

tmt/steps/prepare/shell.py
mcasquer: So should I set it now as a class variable?

docs/releases.rst
mcasquer: Updated !

tmt/steps/prepare/shell.py
mcasquer: Updated !

tmt/steps/prepare/shell.py
mcasquer: Updated !

tmt/steps/prepare/shell.py
mcasquer: Updated !

tmt/steps/prepare/shell.py
mcasquer: Updated !

tmt/steps/prepare/shell.py
```diff
@@ -81,6 +121,35 @@ def go(
...
+                with self._url_clone_lock:
+                    if not repo_path.exists():
+                        repo_path.parent.mkdir(parents=True, exist_ok=True)
+                        tmt.utils.git.git_clone(
+                            url=self.data.url,
+                            destination=repo_path,
+                            shallow=False,
+                            env=environment,
+                            logger=self._logger,
+                        )
```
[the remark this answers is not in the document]
happz: > threads waiting to acquire the lock could all determine the path doesn't exist before any thread gets the lock.

That should not be possible, threads only check whether the path exists *after* taking the lock. No thread can will check its existence before. The first thread to acquire the lock will see the path does not exist, and it will perform `git clone` and `git checkout`, releasing the lock afterwards. All subsequent thread will acquire the lock, see the path exist and release the lock immediately.

tmt/steps/prepare/shell.py
happz: Either in this PR, or in the one reducing the `finish/shell` plugin to the same implementation - if we stick to the step name being included in the envvar name, it would need to become a class variable so the code could use `self` to access it and use the correct one.

Or you can make it a template, e.g. `environment[f'TMT_{self.step.name.upper()}_SHELL_URL_REPOSITORY'] = ...`.

mcasquer: Sorry for the late reply I was on PTO. Updated to class level variable, anyways I am not sure if `_env_var` is a proper name... what do you think?

tmt/steps/prepare/shell.py
```diff
@@ -81,6 +121,35 @@ def go(
         workdir = self.step.plan.worktree
         assert workdir is not None  # narrow type
 
+        if self.data.url:
+            repo_path = workdir / "repository"
+
+            self._env_var = str(repo_path.resolve())
```
happz: I'm not sure what is the purpose of this line, it would overwrite the environment variable name with repository URL. That does not look like intended action.

tmt/steps/prepare/shell.py
happz: Maybe `_cloned_repo_path_envvar_name`? `_env_var` can mean a lot of envvars...

mcasquer: Updated

tmt/steps/prepare/shell.py
mcasquer: Sorry, indeed I misunderstood the idea, I think now it's correct

tmt/steps/prepare/shell.py
```diff
@@ -81,6 +121,35 @@ def go(
...
+                            shallow=False,
+                            env=environment,
+                            logger=self._logger,
+                        )
+
+                        if self.data.ref:
+                            self.info('ref', self.data.ref, 'green')
+                            self.run(
+                                Command('git', 'checkout', '-f', self.data.ref), cwd=repo_path
+                            )
```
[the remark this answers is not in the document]
happz: There is no sharing of repository with previous runs, there is no chance to use different `ref`.

tmt/steps/prepare/shell.py
```diff
@@ -81,6 +121,35 @@ def go(
...
+
+                guest.push(
+                    source=repo_path,
+                    destination=repo_path,
+                    options=["-s", "-p", "--chmod=755"],
+                )
```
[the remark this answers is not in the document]
happz: Cannot be moved outside of the `if not self.is_dry_run`, there would be no directory to push. Some scripts may indeed fail, but the order was clear: dry run, make no changes.

mcasquer:
Added env variable to step variables in overview guide as well

tests/prepare/shell/test.sh
```diff
@@ -20,6 +20,10 @@ rlJournalStart
             prepare -h shell -s 'touch /tmp/first' -s 'touch /tmp/second'"
     rlPhaseEnd
 
+    rlPhaseStartTest "Remote Script"
+        rlRun "tmt -vvv run provision prepare finish plan -n url" 0 "Prepare using a remote script"
+    rlPhaseEnd
```
tcornell-bus, line 26: Should the log file be checked to make sure `cd $TMT_PREPARE_SHELL_URL_REPOSITORY && echo "Hello world"` was executed how we would expect?

mcasquer: @tcornell-bus good question, I did a quick attempt with `rlAssertGrep "Hello world" "$rlRun_LOG"` after adding the `rlRun "workdir=\$(mktemp -d)" 0 "Create tmp directory"` to the setup but I hit:
```bash
:: [ 08:13:21 ] :: [   FAIL   ] :: rlAssertGrep: failed to find file
```

happz: `rlRun -s` - `-s` will capture the output in a file, and then you can `rlAssertGrep` it.

mcasquer: Thanks ! test updated
`:: [ 09:33:25 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.L2ZeNPC0' should contain 'Hello world' `

skycastlelily:
Here are some suggestions[ from cursor](https://lnie.fedorapeople.org/improvement3862.txt):
looks like at least the  No URL validation or security checks  and No validation of url/ref combination should be addressed

mcasquer:
> Here are some suggestions[ from cursor](https://lnie.fedorapeople.org/improvement3862.txt):
> looks like at least the No URL validation or security checks and No validation of url/ref combination should be addressed

Thanks for the suggestion. The url validation should be already discussed in this PR. Encapsulating the clone in a try block won't provide too much extra information, we'll know it failed during cloning, for the others maybe a negative test scenario would be acceptable, anyways I'll propose it for a future patch.

mcasquer:
Hello @psss could you give this PR a last review?

happz (approved):

psss:
> Hello @psss could you give this PR a last review?

Yeah, looks good thanks! Added just a minor adjustment in ea6a7a06. Note that `cleanup` step now needs to be used to stop and remove provisioned guests.

psss (approved):

psss:
Red jobs are irrelevant issues, merging.

PR #3866, opened 2025-07-09
Improve setting FIPS mode by explicitly setting crypto-policy

ctull99:
Enhances the robustness of FIPS mode implementation by moving from a reliance on the systemd/dracut-configured auto-policy to an explicit configuration of the FIPS crypto-policy.

While the existing approach correctly enables FIPS mode, it depends on the system's automatic policy selection safeguard mechanism that's there to prevent accidental misconfiguration. This change makes the resulting FIPS configuration more resilient to potential changes or inconsistencies in the auto-policy safeguard behavior.

By directly setting the FIPS crypto-policy, we ensure a more predictable and secure state.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/prepare/feature/fips-enable.yaml
+++ b/tmt/steps/prepare/feature/fips-enable.yaml
@@ -65,6 +65,12 @@
           changed_when: output.rc == 0
           when: ansible_architecture == "s390x"
 
+    - name: Enable FIPS Policy
+      ansible.builtin.command: update-crypto-policies --set FIPS
+      register: output
+      changed_when: "'is not sufficient for' in output.stderr"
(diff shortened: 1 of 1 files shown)


therazix (approved):

psss (approved):

psss:
@The-Mule, looks good to you as well?

The-Mule:
Yes, this is a very good change. Even though the previous playbook was correct I agree that this makes it more robust.

psss:
> Yes, this is a very good change. Even though the previous playbook was correct I agree that this makes it more robust.

Great, thanks for the quick confirmation!

PR #3871, opened 2025-07-11
Fix the `make coverage` target to actually test

psss:
Seems there was a typo in the `coverage` target definition. It should be running `test:cov` to actually execute some tests.


psss:
Hm, btw, is this used at all? Perhaps we should also update the [Unit Tests](https://tmt.readthedocs.io/en/stable/contribute.html#unit-tests) section a bit.

happz (approved):

psss:
Trivial stuff, approved, merging.

PR #3873, opened 2025-07-11
Document how to cover an environment matrix

psss:
Took me some time to explain in the chat what are the available options to run a set of tests with different environment variables. Would be probably good to have this documented.

Pull Request Checklist

* [ ] write the documentation
* [ ] include a release note

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -855,6 +855,92 @@ that repository.
         how: fmf
 
 
+.. _environment-matrix:
+
+Environment Matrix
+------------------------------------------------------------------
+
+Test environment:
+
+.. code-block:: yaml
+
+    /plan:
+        discover:
+            how: fmf
+        execute:
+            how: tmt
+
+    /test:
+        /one:
+            test: echo first test on level $LEVEL
+
+            /level-7:
+                environment:
+                    LEVEL: 7
+            /level-8:
+                environment:
+                    LEVEL: 8
+
+        /two:
+            test: echo second test on level $LEVEL
+            /level-7:
+                environment:
+                    LEVEL: 7
+            /level-8:
+                environment:
+                    LEVEL: 8
+
+Plan environment:
+
+.. code-block:: yaml
+
+    /plan:
+        discover:
+            how: fmf
+        provision:
+            how: local
+        execute:
+            how: tmt
+
+        /level-7:
+            environment:
+                LEVEL: 7
+        /level-8:
+            environment:
+                LEVEL: 8
+
+    /test:
+        /one:
+            test: echo first test on level $LEVEL
+        /two:
+            test: echo second test on level $LEVEL
+
+Adjust Tests:
+
+.. code-block:: yaml
+
+    /plan:
+        discover:
+          - how: fmf
+            adjust-tests:
+                environment+:
+                    LEVEL: 7
+          - how: fmf
+            adjust-tests:
+                environment+:
+                    LEVEL: 8
+        provision:
+            how: local
+        execute:
+            how: tmt
+
+    /test:
+        /one:
+            test: echo first test on level $LEVEL
+        /two:
+            test: echo second test on level $LEVEL
+
+
 .. _anchors-aliases:
 
 Anchors and Aliases


psss:
Rather simple documentation improvement, would help users better understand environment options, tentatively proposing for the next sprint.

LecrisUT:
This does not cover a matrix though. For clarity, can it cover more explicitly a `x` and `y` variable and the tests being `/tests/$x/$y`?

psss:
> For clarity, can it cover more explicitly a `x` and `y` variable and the tests being `/tests/$x/$y`?

I'm not completely sure what do you mean by `x` and `y` here? The variable name does not have to be part of the test name.

LecrisUT:
> > For clarity, can it cover more explicitly a `x` and `y` variable and the tests being `/tests/$x/$y`?
> 
> I'm not completely sure what do you mean by `x` and `y` here? The variable name does not have to be part of the test name.

Not necessary to be part of the test name, but anything that would cover an actual matrix. It is covered now by the first example.

docs/guide.rst
```diff
@@ -855,6 +855,111 @@ that repository.
...
 
+.. _parametrize-tests:
+
+Parametrize Tests
+------------------------------------------------------------------
+
+Often it is desired to run a set of tests with different
+configurations, for example with different feature flags enabled
+or against different service levels. There are multiple ways how
+to achieve this in ``tmt``, let's discover some of them.
```
LecrisUT: Could we make the wording be more explicit about matrix testing. Parametrized tests is a bit too vague (could mean `$@{context}` parametrization, which doesn't fit in here)

docs/guide.rst
```diff
@@ -855,6 +855,111 @@ that repository.
...
+Below you can find a couple of simple working examples you can
+experiment with. Feel free to copy and paste them to your
+directory and try out how each of them works.
+
+The first option is to use the concept of :ref:`virtual-tests` and
+define the desired environment variable for each test separately.
+In this case there is only a single plan. Environment combinations
+happen on the test level.
+
+.. code-block:: yaml
```
LecrisUT, line 877: Could add a `.. tip` to combine with "Anchors and Aliases". In principle this is the most flexible approach, but the hardest to maintain. Another tip could be to add the generation of the fmf files based on a template which can be added as a pre-commit [^1]

[^1]: I had a proof-of-concept in https://fmf-jinja.readthedocs.io/latest/usecase/tmt.html, but have not checked up on it in a while

docs/releases/pending/3873.fmf
```diff
@@ -0,0 +1,4 @@
+description: |
+    A new section :ref:`parametrize-tests` describing several ways
+    how to efficiently parametrize tests using environment
```
mcasquer:
```suggestion
    to efficiently parametrize tests using environment
```

docs/guide.rst
```diff
@@ -855,6 +855,111 @@ that repository.
...
 
 
+.. _parametrize-tests:
+
+Parametrize Tests
+------------------------------------------------------------------
+
+Often it is desired to run a set of tests with different
+configurations, for example with different feature flags enabled
+or against different service levels. There are multiple ways how
```
mcasquer:
```suggestion
or against different service levels. There are multiple ways
```

docs/guide.rst
psss: > Could add a `.. tip` to combine with "Anchors and Aliases". In principle this is the most flexible approach, but the hardest to maintain.

Good idea, added in 7e326053.

> Another tip could be to add the generation of the fmf files based on a template which can be added as a pre-commit

That sounds like a very advanced example. We might introduce a dedicated section for experts in the future, for now I'd say this is sufficient.

docs/guide.rst
psss: Actually, using `matrix` was the original name of the section. But after describing the three examples I realized that they are more about base parametrizing rather then exercising all possible environment combinations. I'd vote for having that in a separate section.

Anyway, it's a good idea to mention providing `environment` from the command line as well, added in 107db738 together with a reference to the `Parametrize Plans` section.

psss:

PR #3876, opened 2025-07-14
Add additional checks for events affecting test result

therazix:
This PR adds additional internal checks for events that affect the overall test results. Internal checks are now executed later than normal checks, as we need a way to detect guest errors such as reboot timeouts.

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


therazix:
I got stuck on how to test the reconnect timeout check, so for now, I haven’t included a test for it. If anyone has suggestions on how to test this, I’d really appreciate the help.

tests/test/check/main.fmf
```diff
@@ -40,5 +40,4 @@ tier: 2
 /internal:
     test: ./test-internal.sh
     tag+:
-      - provision-only
```
happz, line 43: You need to keep `provision-only`, otherwise the test would run in the general TF jobs as well, not just in the one dedicated to `virtual` plugin.

tmt/steps/execute/__init__.py
```diff
@@ -1031,12 +1032,15 @@ def _run_checks_for_test(
         *,
         event: CheckEvent,
         invocation: TestInvocation,
+        internal: bool = False,
```
happz: Hm, once the method needs a hint like this, it's usually better to parametrize the input completely, i.e. instead of `internal: bool` to tell it which attribute to use, give it `checks: Sequence[Check]` or something, and call it with the actual list of checks to run. `run_checks_before_test`, `run_checks_after_test` and `run_internal_checks` can all pick the right list and pass it to this helper. It will make its API cleaner, no side-effect hidden inside.

tests/test/check/main.fmf
therazix: Added in 89414bb167fdadfec1f08ad1fc64873e32b9f1e9.

tmt/steps/execute/__init__.py
therazix: Changed in 89414bb167fdadfec1f08ad1fc64873e32b9f1e9.

tmt/steps/execute/internal.py
```diff
@@ -729,18 +725,20 @@ def _run_tests(
...
-                            result.note.append('reboot timeout')
 
-                    else:
+                    except tmt.utils.ReconnectTimeoutError as error:
+                        invocation.exception = error
+                        for result in invocation.results:
+                            result.result = ResultOutcome.ERROR
+
+                    except tmt.utils.RestartMaxAttemptsError as error:
+                        invocation.exception = error
```
psss, line 733: Seems, here we're doing the same exception handling three times. Perhaps we could catch all three exceptions as a list in a single `except` clause to deduplicate?

psss (changes requested):
Looks good! Thanks for implementing these. Just one suggestion. Could you please also add a short release note mentioning the new scenarios covered by the internal check handling? Thanks!

psss:
Tests are failing, seems that `/tests/core/about` compare list should be updated.

tmt/steps/execute/internal.py
therazix: Changed in e37a5a8ecbff3eb35de55aa49bd6816800bed5db.

therazix:
Added release note in bbfb8893ad93ffd60684ec7526b02f6aff24fd0a, and I also renamed the Guest class in 273bb51cf9352a71c1e4b6918e914c1513ca2409 to avoid cross-reference issues in documentation.

tmt/steps/execute/internal.py
psss: Thanks!

therazix:
Tests were failing because internal checks were executed after the results had already been interpreted, so the note about a failing internal check was missing, which caused the tests to fail. This change shouldn't be a problem, because internal checks never influenced the result, they are only informative. So I only fixed the tests in 229778e71fa62abc64655dc3bc6d34f43b5ebfcc.

psss (approved):

martinhoyer (commented):
I was thinking how to reduce the duplication, 
```
diff --git a/tmt/checks/internal/guest.py b/tmt/checks/internal/guest.py
index fe7e216d..539958d5 100644
--- a/tmt/checks/internal/guest.py
+++ b/tmt/checks/internal/guest.py
@@ -4,6 +4,7 @@ import tmt.log
 import tmt.utils
 from tmt.checks import CheckPlugin, provides_check
 from tmt.checks.internal import InternalCheck
+from tmt.checks.mixins import ExceptionCheckMixin
 from tmt.result import CheckResult, ResultOutcome
 from tmt.steps.execute import TestInvocation
@@ -17,7 +18,7 @@ class GuestCheck(InternalCheck):


 @provides_check(CHECK_NAME)
-class GuestFailures(CheckPlugin[GuestCheck]):
+class GuestFailures(ExceptionCheckMixin, CheckPlugin[GuestCheck]):
     """
     Check for guest errors during test execution.

@@ -29,6 +30,11 @@ class GuestFailures(CheckPlugin[GuestCheck]):

     _check_class = GuestCheck

+    _EXCEPTION_ERRORS = (
+        tmt.utils.RebootTimeoutError,
+        tmt.utils.ReconnectTimeoutError,
+        )
+
     @classmethod
     def after_test(
         cls,
@@ -38,22 +44,4 @@ class GuestFailures(CheckPlugin[GuestCheck]):
         logger: tmt.log.Logger,
     ) -> list[CheckResult]:
-        if isinstance(invocation.exception, tmt.utils.RebootTimeoutError):
-            return [
-                CheckResult(
-                    name=CHECK_NAME,
-                    result=ResultOutcome.FAIL,
-                    note=['Test failed due to guest reboot timeout.'],
-                )
-            ]
-
-        if isinstance(invocation.exception, tmt.utils.ReconnectTimeoutError):
-            return [
-                CheckResult(
-                    name=CHECK_NAME,
-                    result=ResultOutcome.FAIL,
-                    note=['Test failed due to guest reconnect timeout.'],
-                )
-            ]
-
-        return []
+        return cls._check_for_exceptions(check, invocation)
diff --git a/tmt/checks/internal/invocation.py b/tmt/checks/internal/invocation.py
index 962634d5..091def2f 100644
--- a/tmt/checks/internal/invocation.py
+++ b/tmt/checks/internal/invocation.py
@@ -4,6 +4,7 @@ import tmt.log
 import tmt.utils
 from tmt.checks import CheckPlugin, provides_check
 from tmt.checks.internal import InternalCheck
+from tmt.checks.mixins import ExceptionCheckMixin
 from tmt.result import CheckResult, ResultOutcome
 from tmt.steps.execute import TestInvocation
@@ -18,7 +19,7 @@ class InvocationCheck(InternalCheck):


 @provides_check(CHECK_NAME)
-class Invocation(CheckPlugin[InvocationCheck]):
+class Invocation(ExceptionCheckMixin, CheckPlugin[InvocationCheck]):
     """
     Check for uncategorized invocation errors during test execution.

@@ -30,6 +31,8 @@ class Invocation(CheckPlugin[InvocationCheck]):

     _check_class = InvocationCheck

+    _EXCEPTION_ERRORS = (tmt.utils.RestartMaxAttemptsError,)
+
     @classmethod
     def after_test(
         cls,
@@ -48,17 +51,4 @@ class Invocation(CheckPlugin[InvocationCheck]):
                 )
             ]

-        if isinstance(invocation.exception, tmt.utils.RestartMaxAttemptsError):
-            return [
-                CheckResult(
-                    name=CHECK_NAME,
-                    result=ResultOutcome.FAIL,
-                    note=[
-                        'Test reached maximum restart attempts '
-                        f'({invocation.test.restart_max_count}). '
-                        'You may want to set restart-max-count larger.'
-                    ],
-                )
-            ]
-
-        return []
+        return cls._check_for_exceptions(check, invocation)
```
Just an idea, mixin seems idea since it's it's in invocation.py and internal.py and re-using the exception text is cool, but would need some more scaffolding to allow "custom" notes, if there is need for having a different text in exception and note. Again - just an idea. 

EDIT, I haven't pulled the latest changes, so the diff might be wrong..

EDIT2, ahh, the mixin.py didn't make it to the diff. Here it is, with 'custom_note' support experiement:
```
from typing import TYPE_CHECKING, ClassVar, List, Tuple, Type

from tmt.result import CheckResult, ResultOutcome
from tmt.steps.execute import TestInvocation

if TYPE_CHECKING:
    from tmt.checks import Check


class ExceptionCheckMixin:
    """
    A mixin for check plugins to streamline checking for exceptions.

    This mixin provides a standardized way to check if a test invocation
    resulted in one of a predefined set of exceptions. To use it, a
    check plugin should inherit from this mixin and define the
    `_EXCEPTION_ERRORS` class attribute with a tuple of exception classes
    to check for.

    The `after_test` method of the consumer class can then call
    `_check_for_exceptions` to perform the check.
    """

    # A tuple of exception classes to be checked for.
    # This must be overridden by the class using the mixin.
    _EXCEPTION_ERRORS: ClassVar[Tuple[Type[Exception], ...]] = ()

    @classmethod
    def _get_exception_note(
            cls,
            check: "Check",
            invocation: "TestInvocation") -> List[str]:
        """
        Return the note for the exception CheckResult.

        This method can be overridden to provide a custom note. By default,
        it returns the string representation of the exception.
        """
        if invocation.exception is None:
            return []
        return [str(invocation.exception)]

    @classmethod
    def _check_for_exceptions(
            cls,
            check: "Check",
            invocation: "TestInvocation") -> List[CheckResult]:
        """
        Check for exceptions defined in _EXCEPTION_ERRORS.

        If the invocation's exception is an instance of any exception
        in _EXCEPTION_ERRORS, this method returns a FAIL CheckResult.
        """
        if isinstance(invocation.exception, cls._EXCEPTION_ERRORS):
            return [
                CheckResult(
                    name=check.name,
                    result=ResultOutcome.FAIL,
                    note=cls._get_exception_note(check, invocation)
                )
            ]
        return []
        ```

psss:
One more test [failed](https://artifacts.osci.redhat.com/testing-farm/c6b0f9bf-10d9-42ca-9a6e-377e6a3be0bf/). The message check updated in ea759c46.

psss:
> Just an idea, mixin seems idea since it's it's in invocation.py and internal.py and re-using the exception text is cool, but would need some more scaffolding to allow "custom" notes, if there is need for having a different text in exception and note. Again - just an idea.

The idea sounds good to me, would deduplicate some stuff. I'd rather handle this in a separate pull request though as it's close to the release. @therazix, @happz, any quick thoughts?

therazix:
> > Just an idea, mixin seems idea since it's it's in invocation.py and internal.py and re-using the exception text is cool, but would need some more scaffolding to allow "custom" notes, if there is need for having a different text in exception and note. Again - just an idea.
> 
> The idea sounds good to me, would deduplicate some stuff. I'd rather handle this in a separate pull request though as it's close to the release. @therazix, @happz, any quick thoughts?

I think that's a great idea, but I would rather do it in a different pull request.

happz:
> > > Just an idea, mixin seems idea since it's it's in invocation.py and internal.py and re-using the exception text is cool, but would need some more scaffolding to allow "custom" notes, if there is need for having a different text in exception and note. Again - just an idea.
> > 
> > 
> > The idea sounds good to me, would deduplicate some stuff. I'd rather handle this in a separate pull request though as it's close to the release. @therazix, @happz, any quick thoughts?
> 
> I think that's a great idea, but I would rather do it in a different pull request.

Same here, worth the simplification, but preferably in another PR, given the release timeline...

psss:
Failed job is an internal infra problem, merging.

PR #3878, opened 2025-07-14
Environment files on command-line do not need to be under fmf root

### LINKED ISSUE #3877: tmt searches for env files under metadata root directory
In this commit:

https://github.com/teemtee/tmt/commit/15ca013be2ae703f51b555cb59451af49fbb4cab

We introduced a breaking change which changes the behavior of tmt which now searches for the environment files under the fmf metadata root, i.e. the directory passed via `tmt --root ...`. This broke Testing Farm as we rely on generating the file in the directory where tmt is executed. We worakrounded so the env file is also generated on the new location, but this should be fixed.

happz:
Fixes a regression introduced in #3799.

Fixes #3877.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/core/env/test.sh
+++ b/tests/core/env/test.sh
@@ -6,6 +6,8 @@ rlJournalStart
     rlPhaseStartSetup
         rlRun 'pushd data'
         rlRun 'set -o pipefail'
+
+        rlRun "tmp=\$(mktemp -d)" 0 "Create tmp directory"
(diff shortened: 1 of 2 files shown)


martinhoyer (approved):

mcasquer (approved):
LGTM

psss (approved):

Issue #3859, opened 2025-07-07
Lazy discovering of available package managers


mkoncek:
In `tmt/steps/provision/__init__.py`: `_query_package_manager` and `_query_bootc_builder`.
I believe tmt can safely reorder the function calls to first sort package managers by their priority and then query their presence, stopping at first success.
This minor optimization is related to provisioning using virtualized environment, where package installation is handled differently, e.g. `mock`.

psss:
@mkoncek, in general sounds good, but more context would help a bit here. Just to clarify, isn't this already covered by the `probe_priority`?

https://github.com/teemtee/tmt/blob/32ca248bd0e14ef8b9fe00f3d0235dfc3c0c2e2e/tmt/package_managers/__init__.py#L226-L231

psss:
Ah, I guess now I see what you thought:

https://github.com/teemtee/tmt/blob/32ca248bd0e14ef8b9fe00f3d0235dfc3c0c2e2e/tmt/steps/provision/__init__.py#L589-L606

So you suggest not to discover all available plugins first, then sort and the pick the first one but just iterate through already sorted plugins and break after the first one is detected, right?

Sounds ok to me. @happz any concerns here?

happz:
> Sounds ok to me. [@happz](https://github.com/happz) any concerns here?

Only a minor one: it’s such an obvious optimization that I wonder why haven’t we done it from the beginning. But I think it’s because it was all added gradually, and through several iterations that we never realized the opportunity exists. So it’s probably no dark, undocumented magic.

mkoncek:
My proof-of-concept provisioning works with the current code. I was afraid of the `execute` steps taking too long but I found a way around it.
So there is no pressure from me to implement this change, it is just something I thought could improve the performance.

AthreyVinay:
Just wondering what test case would hit `_query_package_manager` & `_query_bootc_builder`? Tried `tests/provision/facts/test-guest.sh` but doesnt seem to hit the breakpoint set in those methods. However I see there's a `test_package_managers.py` unit test - would that be the closest?

mkoncek:
I encountered this while developing the plugin. I created a very simple tmt plan and added bunch of prints in the code. I then ran `tmt ... provision` step.

psss:
> Tried `tests/provision/facts/test-guest.sh` but doesnt seem to hit the breakpoint set in those methods.

[...]

PR #3898, opened 2025-07-21
Optimize package manager discovery

AthreyVinay:
Related #3859 Also, a change to `rest.py` to handle `mypy and pyright`

* [X] implement the feature
* [x] extend the test coverage - depends if not already covered


tmt/steps/provision/__init__.py
```diff
@@ -581,68 +581,55 @@ def _query_distro(self, guest: 'Guest') -> Optional[str]:
...
+            debug_label='package manager',
+        )
 
-        return None
+    def _query_bootc_builder(
+        self, guest: 'Guest'
+    ) -> Optional['tmt.package_managers.GuestPackageManager']:
+        return self._discover_package_manager(
+            guest,
+            plugin_classes=list(
```
happz: Same as above, if `Iterable` is accepted, can we use `iter_plugins()` directly, without materializing a list?

tmt/steps/provision/__init__.py
```diff
@@ -581,68 +581,55 @@ def _query_distro(self, guest: 'Guest') -> Optional[str]:
...
-            guest.debug(
-                'Discovered bootc builders',
-                fmf.utils.listed([pm.NAME for pm in discovered_package_managers]),
-                level=4,
-            )
-
-            return discovered_package_managers[0].NAME
+        return self._discover_package_manager(
+            guest,
+            plugin_classes=[
```
happz: Since an interable is accepted, we should be able to sneak in a generator, `(package_manager_class for _, package_manager_class in (tmt.package_managers._PACKAGE_MANAGER_PLUGIN_REGISTRY.items())` instead of a materialized list.

tmt/steps/provision/__init__.py
```diff
@@ -581,68 +581,55 @@ def _query_distro(self, guest: 'Guest') -> Optional[str]:
     def _query_kernel_release(self, guest: 'Guest') -> Optional[str]:
         return self._query(guest, [(Command('uname', '-r'), r'(.+)')])
 
-    def _query_package_manager(
-        self, guest: 'Guest'
+    def _discover_package_manager(
+        self,
+        guest: 'Guest',
+        plugin_classes: Iterable[Any],
```
happz: This `Any` could turn into `'tmt.package_managers.PackageManagerClass'`, or `'tmt.package_managers.PackageManagerClass[Any]'`.

tmt/steps/provision/__init__.py
AthreyVinay: That was my initial change, but `flake8` suggested I re-write using a list.

tmt/steps/provision/__init__.py
AthreyVinay: That was my initial change, but ~flake8~ ruff suggested I re-write using a list.

tmt/steps/provision/__init__.py
happz: `flake8` should not suggest anything, `flake8` is used in our tooling only to check docstrings :) `ruff` maybe? I don't see the benefit, would you have the message it reported?

AthreyVinay: Apologies - it was `ruff`.
`C400 Unnecessary generator (rewrite as a list comprehension)`

happz: Hm, maybe the code was slightly different, like a `list()` call instead of actual generator. I would still give it a try.

AthreyVinay: Sure :+1:  Will give it a try again so.

tmt/steps/provision/__init__.py
AthreyVinay: The inner generic with <Any> will not be declared correctly isnt it?

tmt/steps/provision/__init__.py
AthreyVinay: Fixed.

happz:
TBH, it would be nice to have the pre-commit part as a standalone PR.

tmt/steps/provision/__init__.py
```diff
@@ -581,68 +580,55 @@ def _query_distro(self, guest: 'Guest') -> Optional[str]:
...
-            guest.debug(
-                'Discovered package managers',
-                fmf.utils.listed([pm.NAME for pm in discovered_package_managers]),
-                level=4,
-            )
+            tmt.package_managers.PackageManagerClass[tmt.package_managers.PackageManagerEngine]
+        ] = sorted(plugin_classes, key=lambda pm: pm.probe_priority, reverse=True)
 
-            return discovered_package_managers[0].NAME
+        for package_manager_class in discovered_package_managers:
```
happz: The are not "discovered" yet, right? Only "sorted". How about `for package_manager_class in sorted(plugin_classes, key=lambda pm: pm.probe_priority, reverse=True)):` ?

tmt/steps/provision/__init__.py
```diff
@@ -581,68 +580,55 @@ def _query_distro(self, guest: 'Guest') -> Optional[str]:
...
+            debug_label='package manager',
+        )
 
-        return None
+    def _query_bootc_builder(
+        self, guest: 'Guest'
+    ) -> Optional['tmt.package_managers.GuestPackageManager']:
+        return self._discover_package_manager(
+            guest,
+            plugin_classes=(tmt.package_managers._PACKAGE_MANAGER_PLUGIN_REGISTRY.iter_plugins()),
```
happz: Just out of curiosity, if you drop `(` and `)`, will ruff complain that a generator that's not a single argument needs to be wrapped with `(` and `)`?

AthreyVinay:
> TBH, it would be nice to have the pre-commit part as a standalone PR.

I totally agree. This is the [PR](https://github.com/teemtee/tmt/pull/3901). If you can please review, thanks.

tmt/steps/provision/__init__.py
```diff
@@ -581,68 +580,55 @@ def _query_distro(self, guest: 'Guest') -> Optional[str]:
...
+            debug_label='bootc builder',
+        )
```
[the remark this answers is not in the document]
happz: Sounds correct to me.

mkoncek:
I have some note to ask that is related to code changes you make here.
For my provision plugin my goal is to activate a specific package manager only when my provision plugin is used. I do that by using the query command `false` by default and in the `go` provision method I replace the query command with `true`.
In this regard my package manager is similar to `bootc`. You think there would be a better way?

tmt/steps/provision/__init__.py
AthreyVinay: This now will need to exist as right pointed out in: https://github.com/teemtee/tmt/pull/3898#discussion_r2219575965

AthreyVinay:
> I have some note to ask that is realted to code changes you make here. For my provision plugin my goal is to activate a specific package manager only when my provision plugin is used. I do that by using the query command `false` by default and in the `go` provision method I replace the query command with `true`. In this regard my package manager is similar to `bootc`. You think there would be a better way?

I would have to look at the documentation to understand the ask tbh - which I will now, thanks.

EDIT:
Just barely touched the surface of the documentation, isnt `package_manager` a `cached_property`? Would it not be already set by other callers if any?

martinhoyer:
@coderabbitai full review

happz:
@AthreyVinay please, check the extended testsuite results, tests are failing as expected because the output of package manager probe changed.

AthreyVinay:
@happz I have commented those containers that have not found a `package_manager`. This is one approach, however, the other approach (which doesnt make a lot of sense) would be to leave those 2 containers with `None` as the `package_manager` specified and add a guard clause to wherever `CONTAINER_BASE_MATRIX` is referenced with:
```python
#Skip undefined/null package managers
if not package_manager_class:
   continue
```

happz:
> @happz I have commented those containers that have not found a `package_manager`. This is one approach, however, the other approach (which doesnt make a lot of sense) would be to leave those 2 containers with `None` as the `package_manager` specified and add a guard clause to wherever `CONTAINER_BASE_MATRIX` is referenced with:
> 
> ```python
> #Skip undefined/null package managers
> if not package_manager_class:
>    continue
> ```

Hmmm, how can there be containers where we do not discover a package manager? :O We discover one now, before this patch, how can we not discover any after the patch's applied? The coreos images should contain `rpm-ostree`, if nothing else - if tmt cannot discover a package manager, it will fail to work with coreos distros in the wild. Can you let those tests fail, and find out why they can't discover a package manager? That's really weird, suggests a problem exists.

AthreyVinay:
Yeah,  that was my approach too. Interestingly, the tests fail on the `main` branch too. So don't think the patch itself has caused any regression. However, I am still debugging the cause for not finding 'a' package_manager.
@happz Thoughts?

happz:
This is `main`, no failure:

```
$ pytest -n 4 -vv --showlocals tests/unit/test_package_managers.py::test_discovery --html=report.html
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.13.5, pytest-8.3.4, pluggy-1.5.0 -- /home/happz/.local/share/hatch/env/virtual/tmt/BuhwJjvK/dev/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.13.5', 'Platform': 'Linux-6.15.4-100.fc41.x86_64-x86_64-with-glibc2.40', 'Packages': {'pytest': '8.3.4', 'pluggy': '1.5.0'}, 'Plugins': {'xdist': '3.6.1', 'testinfra': '10.1.1', 'pytest_container': '0.4.3', 'metadata': '3.1.1', 'html': '4.1.1'}}
rootdir: /home/happz/git/tmt
configfile: pyproject.toml
plugins: xdist-3.6.1, testinfra-10.1.1, pytest_container-0.4.3, metadata-3.1.1, html-4.1.1
4 workers [12 items]    
scheduling tests via LoadScheduling

tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/42/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/rawhide/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/centos/stream9/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/ubi/8/upstream:latest]
[gw0] [  8%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/rawhide/upstream:latest]
[gw3] [ 16%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/ubi/8/upstream:latest]
[gw2] [ 25%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/centos/stream9/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/ubuntu/22.04/upstream:latest]
[gw1] [ 33%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/42/upstream:latest]
[gw3] [ 41%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/ubuntu/22.04/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/41/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/centos/7/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/centos/stream10/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/debian/12.7/upstream:latest]
[gw0] [ 50%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/41/upstream:latest]
[gw2] [ 58%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/centos/7/upstream:latest]
[gw1] [ 66%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/centos/stream10/upstream:latest]
[gw3] [ 75%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/debian/12.7/upstream:latest]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/coreos/ostree:stable]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/coreos:stable]
tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/alpine:latest]
[gw1] [ 83%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/alpine:latest]
[gw2] [ 91%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/coreos/ostree:stable]
[gw0] [100%] PASSED tests/unit/test_package_managers.py::test_discovery[localhost/tmt/container/test/fedora/coreos:stable]
```

`localhost/tmt/container/test/fedora/coreos/ostree:stable` reports the following:

```
DEBUG    tmt:log.py:766 Discovered package managers: rpm-ostree and dnf5
```

`localhost/tmt/container/test/fedora/coreos:stable` reports:

```
Discovered package managers: dnf5
```

So the `localhost/tmt/container/test/fedora/coreos:stable` could indeed report no discovered package managers, as we no longer add `dnf5` - unless it's already in the image, then it should be still discovered. `localhost/tmt/container/test/fedora/coreos/ostree:stable` should report `rpm-ostree`, at least - if it doesn't, why? Test log should report all commands executed on the guest, can you see whether all package manager probes were executed?

AthreyVinay:
Strange, as this is `main` for me:
Screenshot 2025-07-30 at 10 08 08

Will look at the logs.

AthreyVinay:
Aha, I see this for `/fedora/coreos/ostree:stable`:
Screenshot 2025-07-30 at 10 21 48

```
athreyvinay@fedora:/tmp/mini$ podman ps -a
5bdc0a63dd43  localhost/tmt/container/test/fedora/coreos/ostree:stable      /sbin/init  5 minutes ago   Exited (255) 5 minutes ago              elegant_shockley
athreyvinay@fedora:/tmp/mini$ podman logs elegant_shockley
systemd 257.6-1.fc42 running in system mode (+PAM +AUDIT +SELINUX -APPARMOR +IMA +IPE +SMACK +SECCOMP -GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN -IPTC +KMOD +LIBCRYPTSETUP +LIBCRYPTSETUP_PLUGINS +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK +BTF +XKBCOMMON +UTMP +SYSVINIT +LIBARCHIVE)
Detected virtualization podman.
Detected architecture arm64.
Detected first boot.

Welcome to Fedora CoreOS 42.20250623.3.1!

Initializing machine ID from container UUID.
Failed to mount /run/machine-id (type n/a) on /etc/machine-id (MS_BIND ""): Operation not permitted
Failed to create control group inotify object: Too many open files
Failed to allocate manager object: Too many open files
[!!!!!!] Failed to allocate manager object.
Exiting PID 1...
```

This suggests its a local(my machine only!) issue. Will try to fix it - however will uncomment those lines :)


___________________________________________________________________________________________________________________
Final EDIT (Solved!):

As suspected, it was my host machine and this setting in particular:
`fs.inotify.max_user_instances` which is, by default `128`, bumped it to `512` and the containers spun successfully!

tests/unit/test_package_managers.py
```diff
@@ -227,22 +227,22 @@ def _test_discovery(expected: str, expected_discovery: str) -> None:
         guest.info(f'{has_dnf5_preinstalled(container)=}')
 
         if has_legacy_dnf(container):
-            _test_discovery(tmt.package_managers.dnf.Dnf.NAME, 'dnf and yum')
+            _test_discovery(tmt.package_managers.dnf.Dnf.NAME, 'dnf')
 
             if not has_dnf5_preinstalled(container):
                 guest.execute(ShellScript('dnf install --nogpgcheck -y dnf5'))
```
psss, line 233: Shouldn't this part be dropped as `dnf5` should now be preinstalled across all containers where it is expected?

happz: Maybe, this is a tricky part. Seems plausible, things should work even without it.

psss:
Seems that during the rebase 745cae70cd6a232c88a11682161564f2e54383b5 got in and `pre-commit` started to failed again. Reverted, should be good now.

tests/unit/test_package_managers.py
psss: Let's not block on this now, and clean up later.

PR #3899, opened 2025-07-21
Implement the `journal` test check

### LINKED ISSUE #3267: Feature: dmesg-like test check plugin using the journal
There are tests out there that as part of their logic clear the kernel message ring buffer. This could render the dmesg check plugin useless, because it only checks the delta between test start and test end. In addition it's not always desired that the ring buffer is cleared when using the dmesg check plugin.

A good way to not be intrusive and get everything that was logged during test execution is getting the information from the journal, if the system has journald running.

An alternative to that is to watch and parse `/dev/kmsg`, but that means re-implementing partially what `dmesg` or `journald` do, so I think it's not the desired way.

qiyuann:
Fix #3267.

Pull Request Checklist

* [x] implement the feature
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] mention the version
* [x] include a release note


martinhoyer:
Hi @qiyuann, I've added some changes that would make the journal check to be more general, not kernel/dmesg specific. Adding optional filters, like `unit`, `priority` and `identifier` seemed like a simple enough addition and replacing timestamps with cursor files would make it more robust, though I forgot a lot of the discussion around that so maybe I'm missing something and @happz will correct me. :)

Note that I haven't had yet time to properly test the patch and I've pushed it to get feedback from you and others if this is a good direction.

qiyuann:
@martinhoyer The improvement to make it more general looks good to me.

tmt/checks/journal.py
```diff
@@ -0,0 +1,399 @@
...
+""")
+)
+
+TEST_SCRIPT = jinja2.Template(
+    textwrap.dedent(
+        """
+set -x
+export LC_ALL=C
+
+journalctl --after-cursor-file={{ MARK_FILEPATH }} {{ OPTIONS }}
```
qiyuann: I don't see a `--after-cursor-file` option in journalctl command, shouldn't `--cursor-file` be used instead?

tmt/checks/journal.py
```diff
@@ -0,0 +1,399 @@
...
+    ) -> tuple[ResultOutcome, list[Path]]:
+        assert invocation.phase.step.workdir is not None  # narrow type
+        assert invocation.start_time is not None  # narrow type
+
+        report_timestamp = datetime.datetime.now(datetime.timezone.utc)
+        report: list[str] = []
+
+        options: list[str] = []
+        if self.dmesg:
+            options.append('--dmesg')
```
qiyuann, line 255: There is a comment in https://github.com/teemtee/tmt/issues/3267#issuecomment-2435429736

> We would also need to make sure we do not run this with -k since some tests reboot and we would only capture information from the final boot.

IIUC, when the tmt-reboot is called during the test, the check will be executed again after reboot. It should be ok to just check dmesg from the final boot.

Raising this issue to confirm that `--dmesg` is ok here.

tmt/checks/journal.py
qiyuann: Fixed by using `--cursor-file`
Also changed `journalctl -n 0 --show-cursor > {{ MARK_FILEPATH }}` to `journalctl -n 0 --show-cursor --cursor-file={{ MARK_FILEPATH }}`
Otherwise the content in MARK_FILEPATH will be like:
```
-- No entries --
-- cursor: s=b98b6d29fcba484b98d9fd06be57ff8f;i=717;b=e67aaa5c59c4447295b4363d0ceaafa8;m=c44fb5;t=63a9773dd311f;x=3c08a656dbafa220
```
it's not valid when running `journalctl --cursor-file={{ MARK_FILEPATH }}`

martinhoyer: nice catch, yeah, that's exactly the mistake I would do :D
```
     --after-cursor=CURSOR   Show entries after the specified cursor
     --cursor-file=FILE Show entries after cursor in FILE and update FILE
```

tmt/checks/journal.py
martinhoyer: @qiyuann Good point, thanks! 
After reading `journald.conf` manpage, my understanding that this would depend on `Storage` option. Which, from what I can see is set to `auto` by default. (`systemd-analyze cat-config systemd/journald.conf`)

`persistent` saves the logs to a disk, `/var/log/journal` and creates the directory if it doesn't yet exist.
`auto` works the as persistent *if* `/var/log/journal` directory exists.

So I'm thinking, we could add the `Storage=persistent` to the config to ensure we have are not loosing any logs with reboots.  

If we are doing that, should we create a `50-tmt.conf` within  `/etc/systemd/journald.conf.d/` (which is how we are ensuring `Storage=external` in coredump test check for similar reason), or do you think tmt users would welcome the ability to set arbitrary journald.conf key-value pairs?  

I'm thinking, for sake of consistency and simplicity, we could just set the Storage=persistent everywhere, at least for now? One thing is that it could make some test executions slower, but I'd imagine those would be a fringe cases and since this is new opt-in test check, it would not create any regressions.

WDYT?

@psss @happz any thoughts?

happz: > @qiyuann Good point, thanks! After reading `journald.conf` manpage, my understanding that this would depend on `Storage` option. Which, from what I can see is set to `auto` by default. (`systemd-analyze cat-config systemd/journald.conf`)
> 
> `persistent` saves the logs to a disk, `/var/log/journal` and creates the directory if it doesn't yet exist. `auto` works the as persistent _if_ `/var/log/journal` directory exists.
> 
> So I'm thinking, we could add the `Storage=persistent` to the config to ensure we have are not loosing any logs with reboots.

Would creating `/var/log/journal` also enable the persistent storage? I worry about system configuration changes, and adding a directory might be a smaller hammer than setting `Storage=persistent`.

> If we are doing that, should we create a `50-tmt.conf` within `/etc/systemd/journald.conf.d/` (which is how we are ensuring `Storage=external` in coredump test check for similar reason), or do you think tmt users would welcome the ability to set arbitrary journald.conf key-value pairs?

Probably our custom file for just that change of ours. If users would be interested in changing `journald.conf` more, they should do so in `prepare`.

> I'm thinking, for sake of consistency and simplicity, we could just set the Storage=persistent everywhere, at least for now? One thing is that it could make some test executions slower, but I'd imagine those would be a fringe cases and since this is new opt-in test check, it would not create any regressions.

What does "everywhere" cover in this sentence? "Everywhere" where the `journal` check is invoked?

qiyuann: I think the issue here is not regarding if the logs can be persisted during reboot, but if we need to read logs which are recorded before current boot. `journalctl --dmesg` only show kernel message log from the current boot, not including previous logs.
When talking about reboot during testing, it means the test calls `tmt-reboot`, right?
What is the flow for the test check when `tmt-reboot` is called?
When tmt-reboot is called during testing, the test check will call `before_test` again to generate a new cursor when it continues the test after reboot. Is this correct?
If yes, the test check will read the log from the new cursor, which is only related to current boot. Then  `journalctl --dmesg`  should be ok.

happz: > I think the issue here is not regarding if the logs can be persisted during reboot, but if we need to read logs which are recorded before current boot. `journalctl --dmesg` only show kernel message log from the current boot, not including previous logs.

I guess `dmesg` check also returns the last reboot only, as the buffer is emptied by the reboot. So it wouldn't be unheard of, although it does not look very polished. 

> When talking about reboot during testing, it means the test calls `tmt-reboot`, right?

Either `tmt-reboot` call, or hard reboot invoked by tmt because the guest becomes unresponsive (kernel panic, for example).

> What is the flow for the test check when `tmt-reboot` is called? When tmt-reboot is called during testing, the test check will call `before_test` again to generate a new cursor when it continues the test after reboot. Is this correct?

Yes, after the reboot - intended or not - the test is invoked once again, including all of its `before` checks. In the case of `tmt-reboot`, before rebooting the guest the `after` check is executed, therefore it should save the content of `dmesg` buffer, then the guest reboots, and `after` is called once more before new test invocation.

Let's say there is `tmt-reboot` call, interrupts the test, `after` check saves `dmesg`-like content, guest is restarted, `before` check runs again and saves new timestamp (or cursors, whatever) with "now". Will the next `after` call see the `dmesg`-like content that was logged by kernel *before* the test started again, i.e. the boot-related stuff emitted before this second cursor file was created? `dmesg` would save it, because there is no cursor or timestamp to speak about, but would `journalctl` give me all logged stuff, or just all logged stuff since that cursor?

If the latter is the case, `journal` would miss this after-boot-but-before-second-test-invocation content which did truly happened during "test execution", even though the execution was split in two by a reboot. Would not saving new cursor if one already exists prevent this? Would `journalctl` then deliver messages since the first started for the first time? Or is that in conflict with "`--dmesg` showing only the last boot" statement?

qiyuann: I updated the code. 
Now it creates cursor only when it doesn't exist. `journalctl --cursor-file` will update the cursor file when `after_test` is called before reboot, after reboot, the `before_test` won't refresh the cursor. So the logs after-boot-but-before-second-test-invocation can be captured.
Besides, the `--dmesg` option is changed to `_TRANSPORT=kernel`, so that the message before current boot can also be captured.

tmt/checks/journal.py
```diff
@@ -0,0 +1,399 @@
...
+            assert output is not None
+            report += _report_success('journal log', output)
+            outcome = ResultOutcome.PASS
+        else:
+            report += _report_failure('journal log', exc)
+            output = exc.output
+            outcome = ResultOutcome.ERROR
+
+        failures = self._extract_failures(output.stdout or '')
+        if failures and outcome == ResultOutcome.PASS:
```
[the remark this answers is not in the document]
happz: > but the code attempts to extract failures from output.stdout on line 282.

No, it doesn't, it's happening on line 283, and by that point `output` is already set, either returned from `_run_script()`, or `exc.output`.

> The assignment output = exc.output on line 280 should be moved before the call to self._extract_failures() to ensure output is properly defined in the error case.

It *is* before the call to `self._extract_failures()`.

tmt/checks/journal.py
martinhoyer: > Would creating /var/log/journal also enable the persistent storage?

@happz Only if Storage=auto is set, making it less deterministic.

> What does "everywhere" cover in this sentence? 

making it deterministic, not relying on OS default config.

> Besides, the --dmesg option is changed to _TRANSPORT=kernel, so that the message before current boot can also be captured.

@qiyuann Why? isn't `--dmesg` an alias for that? Perhaps you meant `--boot=all` ?
If I'm missing something and it's needed, I'd also change the parameter name.

> after the reboot - intended or not - the test is invoked once again, including all of its before checks.

Well that's strange. 
Checking for existing cursor sounds good - the workdir where the file is saved is ephemeral (no chance of hitting a cursor file from previous test), right?

happz: Cursor file is stored in a path dedicated for the single test invocation, i.e. specific for test, serial number and guest. There should be no conflict different test.

qiyuann: > Why? isn't --dmesg an alias for that? Perhaps you meant --boot=all ?

@martinhoyer According to the journalctl help message, `-k --dmesg   Show kernel message log from the current boot` , `journalctl --dmesg` can only show message from current boot.
But you are right, `journalctl --dmesg --boot=all` can collect the entire kernel logs.

tmt/checks/journal.py
```diff
@@ -0,0 +1,399 @@
...
+             and not cause a failure.
+             """,
+        metavar='PATTERN',
+        normalize=tmt.utils.normalize_pattern_list,
+        exporter=lambda patterns: [pattern.pattern for pattern in patterns],
+        serialize=lambda patterns: [pattern.pattern for pattern in patterns],
+        unserialize=lambda serialized: [re.compile(pattern) for pattern in serialized],
+    )
+    dmesg: bool = field(
+        default=False, help='Shorthand for ``--dmesg``, check only kernel messages.'
```
psss: I'm not sure I understand: `dmesg` is a shorthand for `--dmesg`? Ah, on a second thought I got it, so it is a shorthand for `jourtnalctl --dmesg`. What about just omitting the shorthand and stating what it is doing?

qiyuann: Removed the `Shorthand` part.

qiyuann:
> Just noticed, we need to update the `test` schema as well:
> 
@psss There is the ticket https://github.com/teemtee/tmt/issues/3929, which will improve the schema. Do we still need to update the schema here?
Currently the schema is for all checks, will it cause problem to add keys specific to journal, such as `dmesg, unit, identifier, priority`

happz:
> > Just noticed, we need to update the `test` schema as well:
> 
> @psss There is the ticket #3929, which will improve the schema. Do we still need to update the schema here? Currently the schema is for all checks, will it cause problem to add keys specific to journal, such as `dmesg, unit, identifier, priority`

Yes, adding these keys among the "global" list of check keys should work for now. #3929 will improve the situation in the future.

tmt/checks/journal.py
```diff
@@ -0,0 +1,399 @@
...
+    ) -> tuple[ResultOutcome, list[Path]]:
+        assert invocation.phase.step.workdir is not None  # narrow type
+        assert invocation.start_time is not None  # narrow type
+
+        report_timestamp = datetime.datetime.now(datetime.timezone.utc)
+        report: list[str] = []
+
+        options: list[str] = []
+        if self.dmesg:
+            options.append('--dmesg --boot=all')
```
martinhoyer: This shouldn't be dmesg-specific imho. `--boot=all` should be either always present or be a separate parameter. I'd lean towards always adding it to the journalctl command, for consistency.

qiyuann: Thank you @martinhoyer 
Moved the `--boot=all` to the journalclt command line, so it works for all options.

tmt/checks/journal.py
```diff
@@ -0,0 +1,397 @@
...
+    @classmethod
+    def after_test(
+        cls,
+        *,
+        check: 'JournalCheck',
+        invocation: 'TestInvocation',
+        environment: Optional[tmt.utils.Environment] = None,
+        logger: tmt.log.Logger,
+    ) -> list[CheckResult]:
+        if not invocation.guest.facts.has_systemd:
```
martinhoyer, line 390: not sure if we want to replicate/reuse something like what's in coredump, i.e. `is_available` flag (and if https://github.com/teemtee/tmt/pull/3498 will ever get in.)

martinhoyer (approved):
LGTM. Not sure about the `_run_script()`, especially when it looks like duplication from avc.py, but to be honest I don't understand all the things around run, shellscript, to have any opinion.

tmt/schemas/test.yaml
```diff
@@ -170,5 +170,19 @@ definitions:
           - type: array
             items:
               type: string
+      failure_pattern:
```
psss:
```suggestion
      failure-pattern:
```

qiyuann: fixed

tmt/checks/journal.py
psss: Let's cover this in #3934.

tmt/checks/journal.py
psss: Thanks!

tests/test/check/test-journal.sh
```diff
@@ -0,0 +1,136 @@
...
+
+        fi
+    rlPhaseEnd
+
+    if [ "$PROVISION_HOW" = "virtual" ]; then
+        # Segfault test only reproducible with virtual (needs root)
+        rlPhaseStartTest "Test journal check with $PROVISION_HOW with a segfault"
+            rlRun "journal_log=$segfault/checks/journal.txt"
+
+            rlRun "tmt run --id $run --scratch -a -vv provision -h $PROVISION_HOW test -n /journal/segfault" "1"
```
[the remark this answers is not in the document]
psss: I just don't understand...

psss (approved):

psss:
Moved the persistent storage enhancement into a separate pull request: #3934

psss:
Seems that the [Test journal check with unit](https://artifacts.osci.redhat.com/testing-farm/6b4d5785-2b76-49cf-a688-8aeb3554cb49/work-the-rest4v3pqlv6/plans/provision/virtual/the-rest/execute/data/guest/default-0/tests/test/check/journal-18/output.txt) failed, @qiyuann could you please [have a look](https://artifacts.osci.redhat.com/testing-farm/6b4d5785-2b76-49cf-a688-8aeb3554cb49/)?

qiyuann:
The captured journal log is empty in unit-test:
```
## journal log

# exit code: finished successfully

# stdout (1 lines)
-- No entries --
```
Could it be that the journal log is not updated in time?

qiyuann:
It should be ok now. The sleep should be out of systemd-run

tmt/checks/journal.py
```diff
@@ -0,0 +1,397 @@
...
+              # These are default patterns
+              - 'Call Trace:'
+              - '\\ssegfault\\s'
+
+              # More patterns to look for
+              - '\\[Firmware Bug\\]'
+            ignore-pattern:
+              - 'a known harmless error message'
+
+    .. versionadded:: 1.54.0
```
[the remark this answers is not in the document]
psss: It **is** included under the correct section, no point mentioning the version again.

psss:
Red jobs are `/plans/provision/container/the-rest` failing because of timeout. Should be fixed by #3937. Merging.

PR #3904, opened 2025-07-21
Implement the new `cleanup` step

### LINKED ISSUE #1737: Introduce a new step for cleanup tasks
On the hacking session today it was suggested that we should probably introduce a new step which would be dedicated to various cleanup tasks and which would allow to separate these from the guest cleanup. Two options were discussed:

 * Keep `finish` for the guest cleanup and move cleanup tasks to the new step
 * Use `finish` for the cleanup tasks and move the guest cleanup to the new step

Let's gather thoughts about advantages and disadvantages of each above-mentioned approach and choose what what be the best way forward.

psss:
Introduce a separate `cleanup` step to handle guest cleanup and workdir pruning so that the `finish` step can be reserved for user finishing tasks only without having to always stop and remove guests.

Fix #1737.

Pull Request Checklist

* [x] implement the feature
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] mention the version
* [x] include a release note


psss:
@AthreyVinay, @vaibhavdaren, although this one is a bit longer, I think it might be worth if you review this pull request as it goes through the whole step implemtation and touches all relevant places. It could give you some good insight of how the step implementation, schema handling, command line options and docs generation works.

docs/releases.rst
```diff
@@ -4,6 +4,17 @@
     Releases
 ======================
 
+tmt-1.54.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A new :ref:`/spec/plans/cleanup` step has been introduced which
+takes care of removing guests and prunning of the workdir. This
```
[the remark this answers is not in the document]
psss: Fixed in 6ecaa689.

happz:
> Tested it briefly, with virtual provisioning and as a user, I very much appreciate, going forward, not accumulating 50 tmt VMs to clean manually from time to time :)

Shouldn't the current `finish` clean those as well? There are certainly no VMs piling up on my workstation.

tmt/steps/cleanup/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # ignore[assignment]: as a base class, CleanupStepData is not included in
+    # CleanupStepDataT.
+    _data_class = CleanupStepData  # type: ignore[assignment]
+
+    # Methods ("how: ..." implementations) registered for the same step.
+    _supported_methods: PluginRegistry[Method] = PluginRegistry('step.cleanup')
+
+    # Internal cleanup plugin is the default implementation
+    how = 'display'
```
happz:
```suggestion
    how = 'tmt'
```

tmt/steps/cleanup/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        if not self.plan.provision.guests:
+            self.warn("Nothing to cleanup, no guests provisioned.", shift=1)
+            return
+
+        # Prepare guest copies
+        guest_copies: list[Guest] = []
+
+        for guest in self.plan.provision.guests:
+            # Create a guest copy and change its parent so that the
+            # operations inside finish plugins on the guest use the
```
happz: "finish plugins"

psss:
> Shouldn't the current `finish` clean those as well? There are certainly no VMs piling up on my workstation.

Yes, the current `finish` implementation definitely takes care of the guest cleanup. The functionality was only moved.

tmt/steps/cleanup/__init__.py
psss: Fixed in f474d7b7.

tmt/steps/cleanup/__init__.py
psss: Fixed in f474d7b7.

psss:
`/tests/provision/container/toolbox` seems to be [failing](https://artifacts.dev.testing-farm.io/806e966e-7be8-417a-8891-cfcc2968f54a/) repeatedly but the problem does not seem to be related.

therazix (approved):
LGTM

happz (approved):

PR #3905, opened 2025-07-22
Revert changes for `ruamel` in precommit mypy

mcasquer:
Since the [mypy upstream issue](https://github.com/python/mypy/issues/12664) has been fixed, the workaround is no longer needed and the warning could be avoided.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,6 @@ repos:
           - "types-lxml"
 
         pass_filenames: false
-        args: [--config-file=pyproject.toml]
 
   - repo: https://github.com/RobertCraigie/pyright-python
     rev: v1.1.399


.pre-commit-config.yaml
```diff
@@ -64,6 +64,7 @@ repos:
           - "types-lxml"
 
         pass_filenames: false
+        args: [--config-file=pyproject.toml]
```
happz: This revert shouldn't be necessary, it's not related to the warning. It's just there, in the same PR, a dropped line that is not needed, and illustrates why adding unrelated changes in a PR may make things more complicated for the future readers.

mcasquer: Updated !

mcasquer:
I've done three test commits in a row, works fine in my case :)

martinhoyer (approved):
Thanks!

Forgot to mention yesterday that it works for me as well :)

mcasquer:
@psss could you review this PR?

psss:
> @psss could you review this PR?

Looks good. Thanks for the cleanup!

psss (approved):

happz:
Not waiting for more tests, core should be enough.

PR #3914, opened 2025-07-24
Update the list of supported container images

psss:
Fedora 40 support has ended, let's update the list of supported container images accordingly.

### DIFF
--- a/tests/images.sh
+++ b/tests/images.sh
@@ -22,8 +22,8 @@ TEST_CONTAINER_IMAGES="${TEST_CONTAINER_IMAGES:-$TEST_IMAGE_PREFIX/alpine:latest
 $TEST_IMAGE_PREFIX/centos/7/upstream:latest
 $TEST_IMAGE_PREFIX/centos/stream9/upstream:latest
 $TEST_IMAGE_PREFIX/centos/stream10/upstream:latest
-$TEST_IMAGE_PREFIX/fedora/40/upstream:latest
 $TEST_IMAGE_PREFIX/fedora/41/upstream:latest
+$TEST_IMAGE_PREFIX/fedora/42/upstream:latest
 $TEST_IMAGE_PREFIX/fedora/rawhide/upstream:latest
 $TEST_IMAGE_PREFIX/fedora/coreos:stable
 $TEST_IMAGE_PREFIX/fedora/coreos/ostree:stable
@@ -38,8 +38,8 @@ $TEST_IMAGE_PREFIX/debian/12.7/upstream:latest}"
 # TODO: enable centos-7 again with modified repo files
 TEST_VIRTUAL_IMAGES="${TEST_VIRTUAL_IMAGES:-centos-stream-9
 centos-stream-10
-fedora-40
 fedora-41
+fedora-42
 fedora-rawhide
 fedora-coreos}"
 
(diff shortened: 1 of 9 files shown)


tests/images.sh
```diff
@@ -58,9 +58,9 @@ function is_fedora_41 () {
     return 1
 }
 
-function is_fedora_40 () {
```
happz, line 61: Some tests are using this function.

happz:
Check also `tests/unit/__init__.py`

tests/images.sh
psss: Fixed in 0168c4ca.

psss:
> Check also `tests/unit/__init__.py`

Should be addressed in 997961b1.

psss:
Blocker for testing, adding to the release.

happz (approved):

psss:
> Why not fedora:latest instead of manually bumping the versions all the time? (and I guess latest -1)

I think one reason might have been that there was a [special handling](https://github.com/teemtee/tmt/pull/3914/files#diff-73ef46a5e318f561d83a865ab25a3cefe3fce01585c4e49f4b53ae194f84c608L126-L128) needed for `fedora-40`. Now it seems there are no special rules needed. Autodetecting the latest `fedora` sounds ok to me.

On the other hand, having a predictable list of distros which are tested saves us from unexpected surprises when then version bump happens (similar to freezing specific python module versions). In this way we can test the version bump change before merging and possibly negatively affecting other pull requests. I remember there were some cases like that in the past.

@happz, any thoughts here?

therazix (commented):
`/tests/prepare/require/test.sh` should be probably updated as well. And fedora 42 containerfile is missing in `/containers/fedora` directory.

psss:
> `/tests/prepare/require/test.sh` should be probably updated as well. And fedora 42 containerfile is missing in `/containers/fedora` directory.

Should be addressed in 57b021e4.

happz:
> > Why not fedora:latest instead of manually bumping the versions all the time? (and I guess latest -1)
> 
> I think one reason might have been that there was a [special handling](https://github.com/teemtee/tmt/pull/3914/files#diff-73ef46a5e318f561d83a865ab25a3cefe3fce01585c4e49f4b53ae194f84c608L126-L128) needed for `fedora-40`. Now it seems there are no special rules needed. Autodetecting the latest `fedora` sounds ok to me.
> 
> On the other hand, having a predictable list of distros which are tested saves us from unexpected surprises when then version bump happens (similar to freezing specific python module versions). In this way we can test the version bump change before merging and possibly negatively affecting other pull requests. I remember there were some cases like that in the past.

I agree having a predictable list of images (VM or container, doesn't matter) is crucial. I think even `latest - 1` would still need to materialize somewhere, and sometimes it does require special care, e.g. dnf in Fedora 40 and Fedora 41 produced different error messages. Can there be more places where the bump needs to happen? Yes, absolutely. Can there be a single `FEDORA_LATEST = ...` variable, and we'd derive the rest? Yes, I think so. Can we get rid of `is_fedora 41` checks in some tests? I don't think so, not as long as Fedora N and Fedora N+1 produce different output during the same task... So to me, bumping a version seems unavoidable, although I can see that it can be simplified.

Issue #1544, opened 2022-09-29
support guest/hostname for context:dimension


yizhanglinux:
Hello
Could you help support the guest/hostname for context:dimension, I want to pass different environments when the test running on different servers, just like below:

https://tmt.readthedocs.io/en/stable/spec/context.html#dimension
```
adjust:
  - environment:
        TEST_DEVS: nvme0n1
    when: guest == server1
```

Thanks
Yi

happz:
This one is pretty much a duplicate of https://github.com/teemtee/tmt/issues/3852, and will be resolved by https://github.com/teemtee/tmt/pull/3917

AthreyVinay:
@happz Can this be moved to `done` as the PR that resolves this is merged?

PR #3917, opened 2025-07-25
Support environment in the provision phase

### LINKED ISSUE #3852: Support `environment` in the `provision` phase
Please support to set environment in provision phase, so we can add something to test environment based on guest e.g.

```yaml
provision:
  - hardware:
      network:
        driver: ice
    environment:
      NIC_DRIVER: ice

  - hardware:
      network:
        driver: mlx5_core
    environment:
      NIC_DRIVER: mlx5_core
```

Related issues:

* https://issues.redhat.com/browse/TFT-3467
* https://issues.redhat.com/browse/TFT-3468

AthreyVinay:
The provision plugins  now supports setting an environment key. This allows users to define environment variables that are injected directly into the test environment for each guest during the provision phase.

Resolves #3852, #1544

Pull Request Checklist

* [X] implement the feature
* [ ] write the documentation
* [X] adjust plugin docstring
* [X] modify the json schema
* [ ] mention the version
* [X] include a release note


tests/provision/environment/data/plans.fmf
```diff
@@ -0,0 +1,17 @@
...
+    name: default-1
+    environment:
+      SOME_ENVVAR_FROM_GUEST_ENVIRONMENT: bar
+
+execute:
+    how: tmt
+    script: |
+      . "$TMT_TOPOLOGY_BASH"
+
+      echo ${TMT_GUEST[name]}:$SOME_ENVVAR_FROM_GUEST_ENVIRONMENT
```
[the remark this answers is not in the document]
happz: Correct. So nothing to address here.

tmt/steps/provision/__init__.py
```diff
@@ -1197,6 +1205,14 @@ def get_data_class(cls) -> type[GuestData]:
 
     hardware: Optional[tmt.hardware.Hardware]
 
+    environment: tmt.utils.Environment = field(
```
happz: At this place, a declaration with type should be enough: the value would be propagated from step data, i.e. there will be no need for `normalize` or any other callback, not even the default value.

docs/releases.rst
```diff
@@ -4,6 +4,15 @@
     Releases
 ======================
 
+tmt-1.54.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`provision </plugins/provision>` plugins now support
+setting an environment key. This allows users to define
```
happz: @psss how to document this key? It's a new key, it's a plugin-level key, but fairly shared too, i.e. would it deserve a section in Specification?

vaibhavdaren: Discussed in the hacking session : move this to provision headers https://github.com/teemtee/tmt/blob/main/docs/plugins/provision-header.inc.rst 
as this is common to all the plugins.

happz: Yep, a section in `provision-header.inc.rst` documenting "keys supported by all provision plugins", `environment` being the first such documented plugin. And then you can point to that section from this release note.

psss:
Missing reviews, this will not make it into 1.54, moving.

tmt/steps/execute/internal.py
```diff
@@ -407,7 +407,12 @@ def _test_environment(
         extra_environment = extra_environment or Environment()
 
         environment = extra_environment.copy()
-        environment.update(invocation.test.environment)
+        environment.update(
+            {
+                **invocation.guest.environment,
+                **invocation.test.environment,
+            }
+        )
```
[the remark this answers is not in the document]
tcornell-bus: What is the intended precedence here? Since there is the possibility of the same env variable being defined in the plan prepare step and in the test, it would be nice to have a test demonstrating the intended precedence in case of conflict.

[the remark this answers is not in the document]
happz: I believe test variables should take precedence over guest-specific variables. Having a test demonstrating that would indeed be nice.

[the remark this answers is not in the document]
thrix: I will just add that it should be documented clearly, what is the precedence so users can find out without the need to read the code.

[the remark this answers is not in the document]
AthreyVinay: ~Have documented the precedence for now, trying to figure out the suitable test, does this look in the right direction?~
```
discover:
  how: shell
  tests:
    - name: env-test
      test: |
        echo "$(hostname):$SOME_ENVVAR_FROM_GUEST_ENVIRONMENT"

prepare:
  - how: shell
    name: set-global-env
    script: |
      export SOME_ENVVAR_FROM_GUEST_ENVIRONMENT=baz

provision:
  - how: container
    name: default-0
    environment:
      SOME_ENVVAR_FROM_GUEST_ENVIRONMENT: foo
  - how: container
    name: default-1
    environment:
      SOME_ENVVAR_FROM_GUEST_ENVIRONMENT: bar

execute:
  how: tmt
```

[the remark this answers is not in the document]
AthreyVinay: Tests implemented to prove `test` variables override `guest` variables :)

docs/plugins/provision-header.inc.rst
```diff
@@ -53,3 +53,15 @@ default to relevant commands:
...
+
+Keys supported by all provision plugins
+---------------------------------------
+
+The following keys are supported by all provision plugins:
+
+* ``environment`` — Setting the environment key allows users to define
+  environment variables that are injected directly into the test environment
+  for each guest during the provision phase.
+  NB: Test variables take precedence over guest-specific variables.
```
happz: What does "NB" stand for?

AthreyVinay: "NB" stands for nota bene, which is a Latin phrase meaning "note well" or "pay attention".  A test case has been added to support this.

docs/plugins/provision-header.inc.rst
```diff
@@ -53,3 +53,15 @@ default to relevant commands:
 Additional SSH options can be specified either via ``ssh-option`` key
 of respective plugins, or by setting ``TMT_SSH_*`` environment
 variables; see :ref:`command-variables` for details.
+
+.. _/plugins/provision/common-keys:
```
happz: Right, the conflict with existing anchor: how about moving the text below to `help` of `environment` key in `GuestData`? First, it really needs a help text, second, it's going to be included in each plugin's configuration docs. Which is pretty much the documentation we were in need for, it's just harder to link to it from release notes. Maybe a link to "Provision" page instead?

```
See the documentation of :ref:`</plugins/provision>`.
```

docs/plugins/provision-header.inc.rst
happz: I see. I'd recommend expanding the acronym, there are plenty of "Note that..." in our docs, but no "NB".

AthreyVinay: Will do!

docs/plugins/provision-header.inc.rst
AthreyVinay: Is my recent commit is what is expected from your comment @happz 
Just missing on the "linking" bit.

therazix: I believe the link to the provision page was meant to be in the release note. It feels redundant to include it in the help text for the environment key.

happz: Yes, release note. The help text of `environment` does not need, it should focus on describing what the key does. The release note would then point to `/plugins/provision`, because we won't have any lone `environment` paragraph, just several of them, one for each plugin.

AthreyVinay: Thanks folks - this will be added to `1.55` to resonate with the current milestone.

docs/releases.rst
```diff
@@ -10,6 +10,9 @@ tmt-1.55.0
 In verbose mode, the ``execute`` step now prints the duration
 limit for each test.
 
+The provision plugins now support setting an ``environment`` key.
+See the documentation of :ref:`/plugins/provision/html`.
```
therazix: The correct reference should be `/plugins/provision`:
```suggestion
See the documentation of :ref:`/plugins/provision`.
```

AthreyVinay: Yep - thank you. Just fixed it there.

tmt/steps/provision/__init__.py
```diff
@@ -1001,6 +1001,19 @@ class GuestData(SerializableContainer):
...
+        default_factory=tmt.utils.Environment,
+        normalize=tmt.utils.Environment.normalize,
+        serialize=lambda environment: environment.to_fmf_spec(),
+        unserialize=lambda serialized: tmt.utils.Environment.from_fmf_spec(serialized),
+        exporter=lambda environment: environment.to_fmf_spec(),
+        help="""
+            Environment variables to be defined for this guest. These will be available
+            during test execution and can be used to customize behavior on a per-guest basis.
+            Note that variables defined here can be overridden by test-level environment variables.
+            """,
```
tcornell-bus, line 1014: When I do `tmt run provision --how container --help` with this branch checked out and in hatch env I don't see this help info for environment. I see the `--hardware` help though. Looking closer it looks like we only support this option in .fmf files, if so how do I go about seeing this help message?

tmt/steps/provision/__init__.py
```diff
@@ -1001,6 +1001,19 @@ class GuestData(SerializableContainer):
         unserialize=lambda serialized: GuestFacts.from_serialized(serialized),
     )
 
+    environment: tmt.utils.Environment = field(
+        default_factory=tmt.utils.Environment,
```
happz, line 1005:
```suggestion
        default_factory=tmt.utils.Environment,
        option=('-e', '--environment'),
        metavar='KEY=VALUE',
```

tmt/steps/provision/__init__.py
happz: Indeed. See https://github.com/teemtee/tmt/pull/3917/files#r2260857993 for the fix.

AthreyVinay: Should be visible now, thanks.

tcornell-bus (approved):

PR #3920, opened 2025-07-28
Change unit test to pytest and move common fixtures to conftest

### LINKED ISSUE #3769: Replace unittest's TestCase, setUp and tearDown with pytest-style fixtures in tests/unit/*
A couple of unit test files uses unittest's `TestCase` classes: besides the class being an extra namespace, and being implemented in different way tan the rest of codebase, there is an actual functionality issue with the setup/teardown methods since they cannot use pytest fixtures as easily as pytest fixtures:

```python
class Foo(TestCase):
    # Ends up with:
    # E       TypeError: Foo.setUp() missing 1 required positional argument: 'tmppath'
    def setUp(self, tmppath):
```

Pytest can work with them, but the missing native support for fixtures creates a rift between these tests and the rest - a clear example is the "create a temporary directory" task: pytest offers the `tmpdir` fixture, which we wrap with a `tmppath` fixture to produce `Path` instances. Including `tmppath` or `tmppath_factory` as a parameter will not work. This opens the door to duplication of code, implementations in other files (`tempfile` can be found in 5 files under `tests/unit/`...), and different ways in which setup/teardown needs to be reasoned about in general, which I think is a bad idea.

So, I'd like to get rid of test classes. Current `tests/unit/test_id.py`:

```python
class IdEmpty(TestCase):
    def setUp(self):
        self.path = Path(tempfile.mkdtemp()) / "empty"
        shutil.copytree(test_path / "empty", self.path)
        self.original_directory = Path.cwd()
        os.chdir(self.path)
        self.base_tree = fmf.Tree(self.path)

    def tearDown(self):
        os.chdir(self.original_directory)
        shutil.rmtree(self.path)

    def test_base(self):
        node = self.base_tree.find("/some/structure")
        test = tmt.Test(logger=root_logger, node=node)
        assert test.id is None
```

Pytest tends toward something like this:

```python
@pytest.fixture(name='empty_tree')
def fixture_empty_tree(tmppath: Path) -> fmf.Tree:
    path = tmppath / "empty"
    shutil.copytree(test_path / "empty", path)

    original_directory = Path.cwd()
    os.chdir(path)

    try:
        yield fmf.Tree(path)

    finally:
        os.chdir(original_directory)
        shutil.rmtree(path)


def test_base(empty_tree: fmf.Tree):
    node = empty_tree.find("/some/structure")
    test = tmt.Test(logger=root_logger, node=node)
    assert test.id is None
```

vaibhavdaren:
Fix #3769
```/home/vaibhav/.local/share/hatch/env/virtual/tmt/ovVD67rI/dev/bin/python /var/lib/snapd/snap/pycharm-community/503/plugins/python-ce/helpers/pycharm/_jb_pytest_runner.py --path /home/vaibhav/office/tmt/tests/unit/test_id.py 
Testing started at 14:44 ...
Launching pytest with arguments /home/vaibhav/office/tmt/tests/unit/test_id.py --no-header --no-summary -q in /home/vaibhav/office/tmt/tests/unit

============================= test session starts ==============================
collecting ... collected 6 items

test_id.py::test_base 
test_id.py::test_manually_add_id 
test_id.py::test_defined_dry 
test_id.py::test_defined_real 
test_id.py::test_empty_dry 
test_id.py::test_empty_real 

======================= 6 passed, 1572 warnings in 0.42s =======================
PASSED                                             [ 16%]PASSED                                  [ 33%]PASSED                                      [ 50%]PASSED                                     [ 66%]PASSED                                        [ 83%]PASSED                                       [100%]
Process finished with exit code 0
```

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -2,7 +2,7 @@ import os
 import pathlib
 import shutil
 from collections.abc import Iterator
-from typing import Any, Callable
+from typing import Any
 
 import _pytest.logging
 import _pytest.tmpdir
@@ -13,8 +13,6 @@ from pytest_container.container import ContainerData
 from tmt.log import Logger
 from tmt.steps.provision.podman import GuestContainer, PodmanGuestData
 from tmt.utils import Path
-
-PathFixture = Callable[[Path, Path, str], Iterator[Path]]
 
 
 @pytest.fixture(name='root_logger')
@@ -91,29 +89,22 @@ def test_path() -> Path:
     return Path(__file__).parent
 
 
-@pytest.fixture
-def make_path_fixture() -> PathFixture:
+def create_path_helper(tmppath: Path, test_path: Path, name: str) -> Iterator[Path]:
     """
-    Provides a reusable helper function to create fixtures that populate a temporary directory.
-
     The returned function creates a temporary directory, populates it with test
     data from a given subdirectory, and changes the current working directory to it.
-    The original working directory is restored after the test.
+    the original working directory is restored after the test.
     """
+    path = tmppath / name
+    shutil.copytree(test_path / name, path)
 
-    def _helper(tmppath: Path, test_path: Path, name: str) -> Iterator[Path]:
-        path = tmppath / name
-        shutil.copytree(test_path / name, path)
+    original_directory = Path.cwd()
+    os.chdir(path)
 
-        original_directory = Path.cwd()
-        os.chdir(path)
-
-        try:
-            yield path
-        finally:
-            os.chdir(original_directory)
-
-    return _helper
+    try:
+        yield path
+    finally:
+        os.chdir(original_directory)
 
 
 @pytest.fixture(scope='module')
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,12 +1,11 @@
 import os
 import pathlib
 import shutil
-from collections.abc import Generator
+from collections.abc import Iterator
 from typing import Any, Callable
 
 import _pytest.logging
 import _pytest.tmpdir
-import fmf
 import py.path
 import pytest
 from pytest_container.container import ContainerData
@@ -14,6 +13,8 @@ from pytest_container.container import ContainerData
 from tmt.log import Logger
 from tmt.steps.provision.podman import GuestContainer, PodmanGuestData
 from tmt.utils import Path
+
+PathFixture = Callable[[Path, Path, str], Iterator[Path]]
 
 
 @pytest.fixture(name='root_logger')
@@ -69,6 +70,10 @@ try:
 except ImportError:
     # ... and if the import fails, we're wrapping `py.path.local` from `tmpdir` family.
 
+    # ignore[name-defined]: when inspected with our daily Python 3.9 or something,
+    # the pytest is probably way newer than the one in RHEL8, and therefore the
+    # name indeed would not exist. But this whole branch applies only with the old
+    # pytest, therefore things are safe.
     @pytest.fixture(scope='session')
     def tmppath_factory(
         tmpdir_factory: '_pytest.tmpdir.TempdirFactory',  # type: ignore[name-defined]
@@ -87,7 +92,7 @@ def test_path() -> Path:
 
 
 @pytest.fixture
-def make_path_fixture() -> Callable[[Path, Path, str], Generator[Path, None, None]]:
+def make_path_fixture() -> PathFixture:
     """
     Provides a reusable helper function to create fixtures that populate a temporary directory.
 
@@ -96,7 +101,7 @@ def make_path_fixture() -> Callable[[Path, Path, str], Generator[Path, None, Non
     The original working directory is restored after the test.
     """
 
-    def _helper(tmppath: Path, test_path: Path, name: str) -> Generator[Path, None, None]:
+    def _helper(tmppath: Path, test_path: Path, name: str) -> Iterator[Path]:
         path = tmppath / name
         shutil.copytree(test_path / name, path)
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_export_to_nitrate.py
```diff
@@ -1,29 +1,30 @@
...
 
-class NitrateExportAutomated(TestCase):
-    def setUp(self):
-        self.tmp_dir = Path(tempfile.mkdtemp(prefix=str(TEST_DIR)))
-        shutil.copytree(TEST_DIR, self.tmp_dir, dirs_exist_ok=True)
-        self.cwd = os.getcwd()
-        self.dir_name = 'manual_test'
+@pytest.fixture(name="manual_test_path")
+def fixture_manual_test_path(
+    tmppath: Path, make_path_fixture: callable
```
AthreyVinay: ~❌ `callable` is not a valid type annotation.
✅ Use `Callable` from the `typing` module instead =>
`Callable[[Path, Path, str], Generator[Path, None, None]]`~

```
from conftest import PathFixture

@pytest.fixture(name="manual_test_path")
def fixture_manual_test_path(
    tmppath: Path, make_path_fixture: PathFixture
) -> Generator[Path, None, None]:
```

tests/unit/conftest.py
```diff
@@ -81,6 +80,57 @@ def tmppath(tmpdir: py.path.local) -> Path:
...
 
 
+@pytest.fixture
+def test_path() -> Path:
+    """Returns the path to the directory containing the tests."""
+    return Path(__file__).parent
+
+
+@pytest.fixture
+def make_path_fixture() -> Callable[[Path, Path, str], Generator[Path, None, None]]:
```
AthreyVinay: Safe to use a `type` and use it across?

```
type PathFixture = Callable[[Path, Path, str], Generator[Path, None, None]]

@pytest.fixture
def make_path_fixture() -> PathFixture:
    ...
```

tests/unit/conftest.py
```diff
@@ -81,6 +80,57 @@ def tmppath(tmpdir: py.path.local) -> Path:
...
+            os.chdir(original_directory)
+
+    return _helper
+
+
+@pytest.fixture
+def defined_path(
+    tmppath: Path,
+    test_path: Path,
+    make_path_fixture: Callable[[Path, Path, str], Generator[Path, None, None]],
```
AthreyVinay: `make_path_fixture: PathFixture`

tests/unit/conftest.py
```diff
@@ -81,6 +80,57 @@ def tmppath(tmpdir: py.path.local) -> Path:
...
+) -> Generator[Path, None, None]:
+    """Fixture for tests requiring the 'defined' test data."""
+    yield from make_path_fixture(tmppath, test_path, "defined")
+
+
+@pytest.fixture
+def empty_path(
+    tmppath: Path,
+    test_path: Path,
+    make_path_fixture: Callable[[Path, Path, str], Generator[Path, None, None]],
```
AthreyVinay: `make_path_fixture: PathFixture`

tests/unit/conftest.py
happz: It is safe, although 1. you don't need `type ...`, just `PathFixture`, and `Generator[Path, None, None]` is `Iterator[Path]`.

tests/unit/conftest.py
```diff
@@ -81,6 +80,57 @@ def tmppath(tmpdir: py.path.local) -> Path:
...
+        try:
+            yield path
+        finally:
+            os.chdir(original_directory)
+
+    return _helper
+
+
+@pytest.fixture
+def defined_path(
```
happz: I would reconsider whether these three fixtures are generic enough that any test could use them - `tmppath_factory` and `tmppath` are for sure, but `empty_path` and `defined_path`, looking at their code and corresponding unit tests, they look like fixtures more suited to be `test_id`-specific, i.e. defined in that file, with smaller scope.

tests/unit/conftest.py
happz: This does not seem to benefit from being a fixture, `_helper` could be simply a function with a generator inside. ` make_path_fixture()` accepts no arguments after all, all are given by the actual fixtures to `_helper()` when they call it, so why `make_path_fixture()` as a fixture?

tests/unit/conftest.py
happz: BTW check out `fixture_id_tree_empty` and `fixture_id_tree_defined` defined below :) They construct fmf trees from the directories your work with, and looking at test code, they would be candidates for being moved to `test_utils.py` as they do nto seem to have any use in other files.

tests/unit/conftest.py
AthreyVinay: Yes, I agree we can ignore `type`  :+1:

tests/unit/conftest.py
vaibhavdaren: Moved `fixture_id_tree_empty` and `fixture_id_tree_defined` to test_utils.py . They are not used anywhere though.

vaibhavdaren: `empty_path` and `defined_path` moved to `test_id.py` as they are test specfic. Increases locality of code.

AthreyVinay (approved):
Changes LGTM.

tests/unit/test_id.py
```diff
@@ -1,114 +1,121 @@
...
-
-        base_tree = fmf.Tree(self.path)
-        node = base_tree.find("/some/structure")
-        assert len(node.data[ID_KEY]) > 10
+
+
+@pytest.fixture(name='test_path')
+def fixture_test_path() -> Path:
+    """Provides the path to the test data for 'id' tests"""
+    return Path(__file__).parent / "id"
```
happz: Isn't this fixture the same as one moved to `conftest.py`?

vaibhavdaren: The path is different.  Changed this to use `test_path` from conftest.py to get parent directory.

therazix (approved):

PR #3923, opened 2025-07-29
Temporarily disable the `avc` test check

psss:
In order to unblock merging, let's disable the check until https://github.com/containers/container-selinux/issues/389 is fixed.


mcasquer (approved):
LGTM

PR #3930, opened 2025-07-30
change test_fllesystem.py to use pytest like other tests.

### LINKED ISSUE #3769: Replace unittest's TestCase, setUp and tearDown with pytest-style fixtures in tests/unit/*

vaibhavdaren:
Fix #3769


tests/unit/test_filesystem.py
```diff
@@ -1,410 +1,335 @@
...
+    Returns a tuple containing the source path, destination path, and a boolean
+    indicating if symlinks are supported on the platform.
+    """
+    source_dir = tmppath / "source"
+    dest_dir = tmppath / "dest"
+    source_dir.mkdir()
+    dest_dir.mkdir()
+
+    # Create test files and directories in the source directory
+    (source_dir / ".fmf").mkdir()
```
AthreyVinay: Just a suggestion: a dictionary will reduce duplication & improves readability, something like this at the module level (outside the fixture) as this can be referenced in the test further below:
 ```python
 EXPECTED_FILES = {
    ".fmf/version": "1",
    "file1.txt": "content1",
    "file2.txt": "content2", 
    "subdir/file3.txt": "content3"
}

@pytest.fixture(name="copy_tree_paths")
def fixture_copy_tree_paths(tmppath: Path) -> copy_tree_path_config:
    ...
    for path, content in EXPECTED_FILES.items():
        file_path = source_dir / path
        file_path.parent.mkdir(parents=True, exist_ok=True)
        file_path.write_text(content)
```

File: tests/unit/test_filesystem.py [hunk 2]
```diff
@@ -1,410 +1,335 @@
...
-        side_effect=OSError("Simulated shutil.copytree failure"),
+
+
+def test_copy_tree_basic(copy_tree_paths: copy_tree_path_config, root_logger: tmt.log.Logger):
+    """Test basic copy operation with default parameters."""
+    source_dir, dest_dir, symlinks_supported = copy_tree_paths
+    tmt.utils.filesystem.copy_tree(source_dir, dest_dir, root_logger)
+
+    # Check if all files were copied
+    assert (dest_dir / ".fmf" / "version").exists()
```
AthreyVinay: Something similar can be done here too:
```python
for file_path in EXPECTED_FILES:
    assert (dest_dir / file_path).exists(), f"Missing file: {file_path}"
```

tests/unit/test_filesystem.py
```diff
@@ -1,410 +1,335 @@
...
+    assert (dest_dir / "file1.txt").exists()
+    assert (dest_dir / "file2.txt").exists()
+    assert (dest_dir / "subdir" / "file3.txt").exists()
+
+    # Check file contents
+    assert (dest_dir / ".fmf" / "version").read_text() == "1"
```
AthreyVinay: This can be refactored too :)

tests/unit/test_filesystem.py
```diff
@@ -1,410 +1,335 @@
...
+):
+    """Test fallback to shutil.copytree when _copy_tree_cp fails."""
+    source_dir, dest_dir, _ = copy_tree_paths
+    tmt.utils.filesystem.copy_tree(source_dir, dest_dir, root_logger)
+
+    mock_copy_tree_cp.assert_called_once_with(source_dir, dest_dir, root_logger)
+    mock_copy_tree_shutil.assert_called_once_with(source_dir, dest_dir, root_logger)
+
+    # Verify files were copied using the fallback approach (shutil.copytree)
+    assert (dest_dir / ".fmf" / "version").exists()
```
AthreyVinay: This can be refactored too :)

tests/unit/test_filesystem.py
vaibhavdaren: Make sense.  As the intent of the CR was to move it to pytest. I kept the code which was already reviewed as it is so that its easy for the reviewers.

tests/unit/test_filesystem.py
```diff
@@ -1,410 +1,313 @@
...
+# A dictionary mapping file paths to their content for test setup.
+# This reduces duplication and improves readability in tests.
+_EXPECTED_TEST_FILES = {
+    ".fmf/version": "1",
+    "file1.txt": "content1",
+    "file2.txt": "content2",
+    "subdir/file3.txt": "content3",
+}
+
+copy_tree_path_config = tuple[Path, Path, bool]
```
happz: Usually a camel case is used for type aliases, therefore `CopyTreePathConfig` would be aligned with other similar objects.

therazix (approved):
LGTM

PR #3933, opened 2025-07-30
Add the new `cleanup` step to `tmt try` as well

psss:
The `tmt try` command needs to be updated as well to include the new `cleanup` command. Without this change provisioned guests would not be stopped and removed upon quit from the `tmt try` session.

Related: #3904

Pull Request Checklist

* [x] implement the feature


psss:
Thanks much for the fast review! :+1:

PR #3940, opened 2025-08-01
Add support for skipping tests before and/or after system upgrade

happz:
New keys, ``skip-tests-before`` and ``skip-tests-after``, were added to
the `execute/upgrade` plugin, allowing users to skip running discovered
tests on the old system or new system, respectively.

Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -12,6 +12,10 @@ plugin, the reboot command now properly applies ``sudo`` when
 necessary, ensuring correct privilege handling during system
 restarts.
 
+New keys, ``skip-tests-before`` and ``skip-tests-after``, were added to
+the :ref:`/plugins/execute/upgrade` plugin, allowing users to skip
+running discovered tests on the old system or new system, respectively.
+
 
 tmt-1.56.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/execute/upgrade/main.fmf
+++ b/tests/execute/upgrade/main.fmf
@@ -30,3 +30,6 @@ tier: 4
 /skip-tests:
     summary: Skip tests before and/or after upgrade
     test: ./skip-tests.sh
+    tag+:
+      - provision-only
+      - provision-virtual
--- a/tests/execute/upgrade/main.fmf
+++ b/tests/execute/upgrade/main.fmf
@@ -29,4 +29,4 @@ tier: 4
 
 /skip-tests:
     summary: Skip tests before and/or after upgrade
-    test: ./skip-test.sh
+    test: ./skip-tests.sh
--- a/tests/execute/upgrade/main.fmf
+++ b/tests/execute/upgrade/main.fmf
@@ -30,6 +30,7 @@ tier: 4
 /skip-tests:
     summary: Skip tests before and/or after upgrade
     test: ./skip-tests.sh
+    duration: 1h
     tag+:
       - provision-only
       - provision-virtual
--- a/tests/execute/upgrade/main.fmf
+++ b/tests/execute/upgrade/main.fmf
@@ -26,3 +26,7 @@ tier: 4
     summary: Plan and upgrade path in the local repo
     framework: shell
     test: cd local && tmt run -vvv --remove plan --name plan
+
+/skip-tests:
+    summary: Skip tests before and/or after upgrade
+    test: ./skip-test.sh
--- a/tests/execute/upgrade/skip-tests.sh
+++ b/tests/execute/upgrade/skip-tests.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "source fedora-version.sh"
+        rlRun "pushd data"
+        rlRun "run=/var/tmp/tmt/run-upgrade"
+    rlPhaseEnd
+
+    rlPhaseStartTest
+        rlRun -s "tmt \
+            --context distro=fedora-${PREVIOUS_VERSION} \
+            --context upgrade-path="${UPGRADE_PATH}" \
+            run --id $run --scratch --rm -vvv --before finish \
+            plan --name /plan/skip-tests/before" 0 "Run the upgrade test"
+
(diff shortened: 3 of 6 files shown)


therazix (approved):

thrix:
Some flakes in downloading copr builds.

docs/releases.rst
```diff
@@ -4,6 +4,14 @@
     Releases
 ======================
 
+tmt-1.55.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+New keys, ``skip-tests-before`` and ``skip-tests-after``, were added to
+:ref:`/plugins/execute/upgrade` plugin, to allow not running discovered
```
thrix: Can we rephrase the `to allow not running`? Something like:

```sugestion
New keys, ``skip-tests-before`` and ``skip-tests-after``, were
added to the :ref:`/plugins/execute/upgrade` plugin, allowing
users to skip running discovered tests on the old system or
new system, respectively.
```

thrix:
@happz I am a bit surprised, I do not see an issue referenced here, was this just something you found useful, or the issue was not connected to this PR?

happz:
> @happz I am a bit surprised, I do not see an issue referenced here, was this just something you found useful, or the issue was not connected to this PR?

No issue as far as I know. I was tasked with running the component upgrade testing for binutils, my tests depend on context, and I needed this to prevent pointless failures.

thrix:
> > @happz I am a bit surprised, I do not see an issue referenced here, was this just something you found useful, or the issue was not connected to this PR?
> 
> No issue as far as I know. I was tasked with running the component upgrade testing for binutils, my tests depend on context, and I needed this to prevent pointless failures.

Thanks, I would not be afraid to fix the motivation next time, it helps people like me, if possible of course.

skycastlelily (approved):
here are two possible valid improvement suggestions: 
1) Issue: No reporting of skipped test phases
Recommendation: Add result metadata:
```
if self.data.skip_tests_before:
    self.verbose('upgrade', 'skipping tests on the old system', color='blue', shift=1)
    # Add metadata about skipped phase
    self._results.append(Result(
        name='skip-tests-before',
        result=ResultOutcome.SKIP,
        log=['Tests skipped before upgrade as requested']
    ))
```
2) you may want to check the senario where both "skip-tests-before" and "skip-test-after" are both set to be True

and Miro's comment,otherwise,LGTM^^

happz:
> here are two possible valid improvement suggestions:
> 
> 1. Issue: No reporting of skipped test phases
>    Recommendation: Add result metadata:
> 
> ```
> if self.data.skip_tests_before:
>     self.verbose('upgrade', 'skipping tests on the old system', color='blue', shift=1)
>     # Add metadata about skipped phase
>     self._results.append(Result(
>         name='skip-tests-before',
>         result=ResultOutcome.SKIP,
>         log=['Tests skipped before upgrade as requested']
>     ))
> ```

That's taken care of by `execute` itself, as it adds `PENDING` results for all discovered tests.

> 2. you may want to check the senario where both "skip-tests-before" and "skip-test-after" are both set to be True

I'm fine with both being set, running just the upgrade.

docs/releases.rst
happz: Sure, changed in 3d18fecb.

happz:
> > > @happz I am a bit surprised, I do not see an issue referenced here, was this just something you found useful, or the issue was not connected to this PR?
> > 
> > 
> > No issue as far as I know. I was tasked with running the component upgrade testing for binutils, my tests depend on context, and I needed this to prevent pointless failures.
> 
> Thanks, I would not be afraid to fix the motivation next time, it helps people like me, if possible of course.

Interesting, I'd swear I made this a commit message. I guess I just wrote it as a release note. Fixed.

tests/execute/upgrade/main.fmf
```diff
@@ -26,3 +26,7 @@ tier: 4
     summary: Plan and upgrade path in the local repo
     framework: shell
     test: cd local && tmt run -vvv --remove plan --name plan
+
+/skip-tests:
+    summary: Skip tests before and/or after upgrade
+    test: ./skip-test.sh
```
happz:
```suggestion
    test: ./skip-tests.sh
```

docs/releases.rst
happz: @thrix please, rereview.

tests/execute/upgrade/main.fmf
```diff
@@ -26,3 +26,7 @@ tier: 4
     summary: Plan and upgrade path in the local repo
     framework: shell
     test: cd local && tmt run -vvv --remove plan --name plan
+
+/skip-tests:
+    summary: Skip tests before and/or after upgrade
+    test: ./skip-tests.sh
```
psss, line 32: Should be tagged with `provision-only` and `provision-virtual` to be run in the `provision` which should fix tests failing because `libvirtd` is not started.

tests/execute/upgrade/skip-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+            plan --name /plan/skip-tests/after" 0 "Run the upgrade test"
+
+        # 1 test before + 3 upgrade tasks + 0 test after
+        rlAssertGrep "4 tests passed" $rlRun_LOG
+        rlAssertGrep    "pass /old/test" $rlRun_LOG
+        rlAssertNotGrep "pass /new/test" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartCleanup
+        rlRun "tmt run --id $run finish" 0 "Stop the guest and remove the workdir"
```
psss:
```suggestion
        rlRun "tmt run --id $run cleanup" 0 "Stop the guest and remove the workdir"
```

Perhaps `--before cleanup` should used above as well.

psss (changes requested):
Nice improvement! Found two issues.

tests/execute/upgrade/skip-tests.sh
happz: Addressed in ee7f7f94a28a1b970e41ed4c68832b3a3b39f68c

tests/execute/upgrade/main.fmf
happz: Addressed in ee7f7f94a28a1b970e41ed4c68832b3a3b39f68c

PR #3941, opened 2025-08-04
Support multiple threads provisioning for `Mrack`

skycastlelily:
The test dosen't cover the guest.start()  part for multi host provisioning as it's impossible without beaker config provided. 

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -1166,7 +1166,9 @@ class BeakerAPI:
         try:
             config_data = tmt.utils.yaml_to_dict(provisioning_config.read_text())
         except Exception as config_err:
-            raise ProvisionError(f"Failed to load mrack provisioning configuration: {config_err}")
+            raise ProvisionError(
+                "Failed to load mrack provisioning configuration."
+            ) from config_err
 
         # Create transformer with thread-safe context
         self._mrack_transformer = TmtBeakerTransformer()
@@ -1188,7 +1190,7 @@ class BeakerAPI:
                 f"Configuration file missing: {missing_conf_err.filename}"
             ) from missing_conf_err
         except Exception as e:
-            raise ProvisionError(f"Failed to initialize mrack transformer: {e}")
+            raise ProvisionError("Failed to initialize mrack transformer.") from e
 
         self._mrack_provider = self._mrack_transformer._provider
         self._mrack_provider.poll_sleep = DEFAULT_PROVISION_TICK
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -893,7 +893,8 @@ def import_and_load_mrack_deps(workdir: Any, name: str, logger: tmt.log.Logger) 
                 if Path("mrack.log").exists():
                     Path("mrack.log").unlink()
 
-            logging.FileHandler(str(log_filename))
+            file_handler = logging.FileHandler(str(log_filename))
+            mrack.logger.addHandler(file_handler)
 
             providers.register(BEAKER, BeakerProvider)
 
@@ -1208,17 +1209,18 @@ class BeakerAPI:
         """
         Create provider configuration from parsed YAML config (thread-safe)
         """
+
         try:
             # The YAML config should already be in the correct format for mrack
             # Just validate that it has the necessary structure
             if not isinstance(config_data, dict):
-                raise ProvisionError(
+                raise tmt.utils.SpecificationError(
                     "Invalid mrack provisioning configuration format - expected dictionary"
                 )
 
             # Check for required provider configurations
             if 'beaker' not in config_data:
-                raise ProvisionError(
+                raise tmt.utils.SpecificationError(
                     "No 'beaker' section found in mrack provisioning configuration"
                 )
 
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -844,7 +844,7 @@ def constraint_to_beaker_filter(
 _MRACK_IMPORT_LOCK = threading.Lock()
 
 
-def import_and_load_mrack_deps(workdir: Any, name: str, logger: tmt.log.Logger) -> None:
+def import_and_load_mrack_deps(mrack_log: str, logger: tmt.log.Logger) -> None:
     """
     Import mrack module only when needed (thread-safe)
     """
@@ -877,22 +877,15 @@ def import_and_load_mrack_deps(workdir: Any, name: str, logger: tmt.log.Logger) 
             MRACK_VERSION = importlib.metadata.version('mrack')
 
             # hack: remove mrack stdout and move the logfile to /tmp
-            # Make log file unique per thread to avoid conflicts
             if hasattr(mrack, 'console_handler'):
                 mrack.logger.removeHandler(mrack.console_handler)
             if hasattr(mrack, 'file_handler'):
                 mrack.logger.removeHandler(mrack.file_handler)
 
-            # Use thread-safe approach for log file cleanup
-            thread_id = threading.get_ident()
-            log_filename = f"{workdir}/{name}-mrack-{thread_id}.log"
-
             with suppress(OSError):
-                # Only try to remove the generic log file, not thread-specific ones
                 if Path("mrack.log").exists():
                     Path("mrack.log").unlink()
-
-            mrack.logger.addHandler(logging.FileHandler(str(log_filename)))
+            mrack.logger.addHandler(logging.FileHandler(mrack_log))
 
             providers.register(BEAKER, BeakerProvider)
 
@@ -1300,7 +1293,8 @@ class GuestBeaker(tmt.steps.provision.GuestSsh):
         super().__init__(*args, **kwargs)
 
         assert isinstance(self.parent, tmt.steps.provision.Provision)
-        import_and_load_mrack_deps(self.parent.workdir, self.parent.name, self._logger)
+        self.mrack_log = f"{self.workdir}/{self.parent.name}-mrack.log"
+        import_and_load_mrack_deps(self.mrack_log, self._logger)
 
     @property
     def api(self) -> BeakerAPI:
@@ -1640,6 +1634,18 @@ class ProvisionBeaker(tmt.steps.provision.ProvisionPlugin[ProvisionBeakerData]):
                 parent=self.step,
                 logger=self._logger,
             )
+
+    @property
+    def _preserved_workdir_members(self) -> set[str]:
+        """
+        A set of members of the step workdir that should not be removed.
+        """
+        assert self._guest
+
+        # Extract just the filename from the full path since
+        # prune_directory compares against member.name
+        mrack_log_filename = Path(self._guest.mrack_log).name
+        return {*super()._preserved_workdir_members, mrack_log_filename}
 
     def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         """
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -6,7 +6,6 @@ import importlib.metadata
 import logging
 import re
 import threading
-from collections.abc import Mapping
 from contextlib import suppress
 from functools import wraps
 from typing import Any, Callable, Optional, TypedDict, TypeVar, Union, cast
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -1295,7 +1295,7 @@ class GuestBeaker(tmt.steps.provision.GuestSsh):
         super().__init__(*args, **kwargs)
 
         assert isinstance(self.parent, tmt.steps.provision.Provision)
-        self.mrack_log = f"{self.workdir}/{self.parent.name}-mrack.log"
+        self.mrack_log = f"{self.parent.workdir}/{self.parent.name}-mrack.log"
         import_and_load_mrack_deps(self.mrack_log, self._logger)
 
     @property
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -893,8 +893,7 @@ def import_and_load_mrack_deps(workdir: Any, name: str, logger: tmt.log.Logger) 
                 if Path("mrack.log").exists():
                     Path("mrack.log").unlink()
 
-            file_handler = logging.FileHandler(str(log_filename))
-            mrack.logger.addHandler(file_handler)
+            mrack.logger.addHandler(logging.FileHandler(str(log_filename)))
 
             providers.register(BEAKER, BeakerProvider)
 
@@ -1214,13 +1213,13 @@ class BeakerAPI:
             # The YAML config should already be in the correct format for mrack
             # Just validate that it has the necessary structure
             if not isinstance(config_data, dict):
-                raise tmt.utils.SpecificationError(
+                raise ProvisionError(
                     "Invalid mrack provisioning configuration format - expected dictionary"
                 )
 
             # Check for required provider configurations
             if 'beaker' not in config_data:
-                raise tmt.utils.SpecificationError(
+                raise ProvisionError(
                     "No 'beaker' section found in mrack provisioning configuration"
                 )
 
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -1226,7 +1226,7 @@ class BeakerAPI:
             return config_data
 
         except Exception as e:
-            raise ProvisionError(f"Failed to parse mrack provisioning configuration: {e}")
+            raise ProvisionError("Failed to parse mrack provisioning configuration.") from e
 
     @async_run
     async def create(self, data: CreateJobParameters) -> Any:
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -46,6 +46,7 @@ BeakerTransformer: Any
 TmtBeakerTransformer: Any
 
 _MRACK_IMPORTED: bool = False
+_MRACK_CONTEXT_INITIALIZED: bool = False
 
 DEFAULT_ARCH = 'x86_64'
 DEFAULT_IMAGE = 'fedora'
@@ -842,6 +843,8 @@ def constraint_to_beaker_filter(
 
 # Thread-safe lock for mrack imports and initialization
 _MRACK_IMPORT_LOCK = threading.Lock()
+_MRACK_CONTEXT_LOCK = threading.Lock()
+_MRACK_CONTEXT_CONFIG_PATH: Optional[str] = None
 
 
 def import_and_load_mrack_deps(mrack_log: str, logger: tmt.log.Logger) -> None:
@@ -954,6 +957,25 @@ def import_and_load_mrack_deps(mrack_log: str, logger: tmt.log.Logger) -> None:
             return req
 
     _MRACK_IMPORTED = True
+
+
+def init_mrack_global_context(config_path: str) -> None:
+    """
+    Initialize mrack global context in a thread-safe manner
+    """
+    global _MRACK_CONTEXT_INITIALIZED, _MRACK_CONTEXT_CONFIG_PATH
+
+    with _MRACK_CONTEXT_LOCK:
+        # If already initialized with the same config, return
+        if _MRACK_CONTEXT_INITIALIZED and config_path == _MRACK_CONTEXT_CONFIG_PATH:
+            return
+
+        # If initialized with different config, reinitialize
+        global_context = mrack.context.global_context
+        global_context.init(config_path)
+
+        _MRACK_CONTEXT_INITIALIZED = True
+        _MRACK_CONTEXT_CONFIG_PATH = config_path
 
 
 def async_run(func: Any) -> Any:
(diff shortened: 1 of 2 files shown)


skycastlelily:
Please note,  most of the test code is generated by cursor :)

tmt/steps/provision/mrack.py
```diff
@@ -1123,13 +1143,39 @@ async def __init__(self, guest: 'GuestBeaker') -> None:  # type: ignore[misc]
...
+            if 'beaker' not in config_data:
+                raise ProvisionError(
+                    "No 'beaker' section found in mrack provisioning configuration"
+                )
+
+            # Return the configuration data as-is since it's already in the correct format
+            return config_data
+
+        except Exception as e:
+            raise ProvisionError(f"Failed to parse mrack provisioning configuration: {e}")
```
happz: `raise ... from e` instead of embedding `e` into the message. It preserves the sequence of exceptions, with their tracebacks, and it is reported correctly by tmt.

skycastlelily: updated^^: https://github.com/teemtee/tmt/pull/3941/commits/e49a362aa8379831f95f0f1384130e3080effb40

thrix:
As the code was generated, I am adding some review from Claude Code :) Hopefully it is not a garbage. Btw the unit tests fail in the tests.

### 🚨 Critical Issues

**1. Exception Type Mismatch**

Location: tmt/steps/provision/mrack.py:1163,1169

```
  # Current (WRONG):
  raise ProvisionError("Invalid mrack provisioning configuration format...")
  raise ProvisionError("No 'beaker' section found...")
```

```
  # Should be:
  raise tmt.utils.SpecificationError("Invalid mrack provisioning configuration format...")
  raise tmt.utils.SpecificationError("No 'beaker' section found...")
```

Tests expect SpecificationError but code raises ProvisionError. The private `_create_provider_config()` method should raise validation-specific exceptions that get caught and re-wrapped by the public interface.

**2. FileHandler Created But Not Used**

Location: tmt/steps/provision/mrack.py:839

```
  # Current (WRONG):
  logging.FileHandler(str(log_filename))  # Creates object but doesn't use it
```

```
  # Should be:
  file_handler = logging.FileHandler(str(log_filename))
  mrack.logger.addHandler(file_handler)
```

Issue: Log file creation doesn't work - FileHandler object is created but never attached to any logger.

**3. Test Doesn't Test What It Claims**

Location: tests/unit/provision/mrack/test_thread_safety.py:17-18

```
# This completely mocks BeakerAPI, so no real thread safety is tested:
patch('tmt.steps.provision.mrack.BeakerAPI') as mock_beaker_api,
```

The `test_multiple_beaker_api_instances_thread_safety()` patches away the actual BeakerAPI class, so the thread-safe import logic is never exercised. The test name is misleading.

### 🔧 Logic Issues

**4. Incorrect Object Initialization in Tests**

Location: tests/unit/provision/mrack/test_thread_safety.py:136,170

```
  # Current (PROBLEMATIC):
  guest = GuestBeaker.__new__(GuestBeaker)
  guest._guest = sample_data  # Wrong attribute, bypasses __init__
```

```
  # Should properly initialize or mock:
  guest = Mock(spec=GuestBeaker)
  # or use proper constructor
```  
 
Using `__new__()` bypasses initialization and sets potentially wrong attributes, could cause runtime errors.

**5. Global Logger State Contamination**

Location: tmt/steps/provision/mrack.py:825-828

```
  # Modifies global mrack logger for all threads/processes:
  if hasattr(mrack, 'console_handler'):
      mrack.logger.removeHandler(mrack.console_handler)
```

Global logger modifications affect the entire process, not just current thread.  Could interfere with other mrack usage.

### ⚠️ Minor Issues

6. Thread ID Log File Collision Risk

Location: tmt/steps/provision/mrack.py:831-832

````
thread_id = threading.get_ident()
log_filename = f"{workdir}/{name}-mrack-{thread_id}.log"
````

Thread IDs can be recycled in long-running processes, potentially causing log file conflicts. Consider using threading.current_thread().name or UUID instead.

✅ What's Actually Good

  - Threading lock implementation is correct
  - Configuration file search logic is solid
  - Error propagation preserves original exceptions with from e
  - Thread-safe approach for configuration loading
  - _thread_safe = True flag properly set

📋 Priority Fix Order

1. Exception types - Tests will fail until fixed
2. FileHandler attachment - Logging doesn't work currently
3. Test logic - Misleading test that doesn't test thread safety
4. Object initialization - Potential runtime errors
5. Global state handling - Could cause side effects
6. Thread ID collision - Minor, long-term concern

**Comment by happz** [comment]
> As the code was generated, I am adding some review from Claude Code :) Hopefully it is not a garbage. Btw the unit tests fail in the tests.
> 
> ### 🚨 Critical Issues
> **1. Exception Type Mismatch**
> 
> Location: tmt/steps/provision/mrack.py:1163,1169
> 
> ```
>   # Current (WRONG):
>   raise ProvisionError("Invalid mrack provisioning configuration format...")
>   raise ProvisionError("No 'beaker' section found...")
> ```
> 
> ```
>   # Should be:
>   raise tmt.utils.SpecificationError("Invalid mrack provisioning configuration format...")
>   raise tmt.utils.SpecificationError("No 'beaker' section found...")
> ```
>
> Tests expect SpecificationError but code raises ProvisionError. The private `_create_provider_config()` method should raise validation-specific exceptions that get caught and re-wrapped by the public interface.

`tmt.utils.SpecificationError` is raised when test, plan, story, or other user-provided inputs are invalid. This is a configuration issue of a plugin rather than a mangled envvar in an fmf file, `ProvisionError` seems more fitting to me.

File: tmt/steps/provision/mrack.py [hunk 2]
```diff
@@ -1123,13 +1144,40 @@ async def __init__(self, guest: 'GuestBeaker') -> None:  # type: ignore[misc]
...
+    def _create_provider_config(self, config_data: dict[str, Any]) -> dict[str, Any]:
+        """
+        Create provider configuration from parsed YAML config (thread-safe)
+        """
+
+        try:
+            # The YAML config should already be in the correct format for mrack
+            # Just validate that it has the necessary structure
+            if not isinstance(config_data, dict):
+                raise tmt.utils.SpecificationError(
```
happz: Please, see https://github.com/teemtee/tmt/pull/3941#issuecomment-3244031495

tmt/steps/provision/mrack.py
```diff
@@ -784,49 +784,65 @@ def constraint_to_beaker_filter(
...
-        with suppress(OSError):
-            os.remove("mrack.log")
+            with suppress(OSError):
+                # Only try to remove the generic log file, not thread-specific ones
+                if Path("mrack.log").exists():
+                    Path("mrack.log").unlink()
 
-        logging.FileHandler(str(f"{workdir}/{name}-mrack.log"))
+            file_handler = logging.FileHandler(str(log_filename))
+            mrack.logger.addHandler(file_handler)
```
happz: How about `mrack.logger.addHandler(logging.FileHandler(str(log_filename)))`? `file_handler` has no purpose other than being passed to `addHandler()` call...

tmt/steps/provision/mrack.py
skycastlelily: addressed: https://github.com/teemtee/tmt/pull/3941/commits/1552984e5376cd187347cc125d368b3e5f4239a7 :sweat_smile:

tmt/steps/provision/mrack.py
skycastlelily: addressed: https://github.com/teemtee/tmt/pull/3941/commits/1552984e5376cd187347cc125d368b3e5f4239a7 ^^

skycastlelily:
> Btw the unit tests fail in the tests.

Forgot to say, this is addressed  in: https://github.com/teemtee/tmt/pull/3941/commits/1552984e5376cd187347cc125d368b3e5f4239a7 :)

skycastlelily:
I asked cursor yesterday  to cover tests for guest.start() and guest.setup() part where the multi host provisioning happens,but it failed( I wondered if it's possible even before I asked, and yeah, it can't:()

❌ What It DOESN'T Test
The current solution cannot test:
❌ Real BeakerAPI.__init__() async operations
❌ Real guest.start() provisioning flow
❌ Real guest.setup() SSH initialization
❌ Actual mrack transformer/provider thread safety
🎯 Conclusion
Your original concern was 100% valid - the previous test didn't test real BeakerAPI thread safety.
My attempt to fix it properly failed due to the complexity of:
Async operations in threading contexts
External service dependencies
Complex mrack initialization chains
The current solution is a reasonable compromise that tests the thread-safe components we can realistically test, with clear documentation about its limitations.
You were right to question the approach, and I've learned that sometimes the "perfect" test isn't achievable due to infrastructure complexity. The current test does provide meaningful thread safety verification for the components that matter most. 🎯

tcornell-bus:
I was going to try to test this manually, but it looks like there is a merge conflict in `tmt/steps/provision/mrack.py` during build.

skycastlelily:
> I was going to try to test this manually, but it looks like there is a merge conflict in `tmt/steps/provision/mrack.py` during build.

resolved : )

skycastlelily:
For this commit:
https://github.com/teemtee/tmt/pull/3941/commits/8c3daa9702dfab0079b67432ccd114ef9a64d8e1
I removed the failure test,because gemini said:)

Why It Passed Locally but Fails on CI
Your local machine and the CI environment likely have different CPU architectures, core counts, and scheduling policies.

Thread Scheduling: On your local machine, the OS scheduler might be distributing the three threads across different cores or giving them enough time between context switches that threading.get_ident() returns a unique identifier for each thread's execution.

CI Environment: In contrast, the CI environment, which might be a virtual machine with fewer cores or a highly optimized scheduler, could be running the threads in such tight succession that two or more threads get scheduled on the same core before the others, leading to a non-unique thread ID. The thread IDs are a low-level OS construct and can be reused.

Instead of tracking and asserting the uniqueness of thread IDs, you should:
Remove the assertion: Delete or comment out the line assert len(unique_config_threads) == 3. This assertion is the source of the failure and doesn't actually confirm the core concept of thread safety—that multiple threads can access a shared resource without corrupting it.

Focus on the core test: The real thread safety is tested by the other assertions, such as:

assert len(exceptions) == 0: Ensures no exceptions were raised due to race conditions.

assert len(results) == 3: Confirms that all threads completed their execution and produced a result.

assert len(config_load_calls) == 3: Verifies that the config loading function was called the correct number of times (once per thread), which is a valid indicator of independent execution.

happz:
> For this commit: [8c3daa9](https://github.com/teemtee/tmt/commit/8c3daa9702dfab0079b67432ccd114ef9a64d8e1) I removed the failure test,because gemini said:)

Please, don't do things just because an LLM said you should do them. They hallucinate, and they do not know what they say.

The explanation below is nonsense. The number of CPU cores is absolutely irrelevant when it comes to Python and CPython, its multithreading support does not use multiple CPUs, not even multiple CPU cores, all threads running Python bytecode are stuck on the same processors. There is no running of threads on multiple CPUs, no scheduling policy can change this.

> Why It Passed Locally but Fails on CI Your local machine and the CI environment likely have different CPU architectures, core counts, and scheduling policies.
> 
> Thread Scheduling: On your local machine, the OS scheduler might be distributing the three threads across different cores or giving them enough time between context switches that threading.get_ident() returns a unique identifier for each thread's execution.
> 
> CI Environment: In contrast, the CI environment, which might be a virtual machine with fewer cores or a highly optimized scheduler, could be running the threads in such tight succession that two or more threads get scheduled on the same core before the others, leading to a non-unique thread ID. The thread IDs are a low-level OS construct and can be reused.

skycastlelily:
> They hallucinate, and they do not know what they say.

I know, I didn't look into the tech detail, one reason is I think what it said below make sense,and I really do think cursor created too many tests :sweat_smile: 

> Focus on the core test: The real thread safety is tested by the other assertions, such as:
> assert len(exceptions) == 0: Ensures no exceptions were raised due to race conditions.
> 
> assert len(results) == 3: Confirms that all threads completed their execution and produced a result.
> assert len(config_load_calls) == 3: Verifies that the config loading function was called the correct number of times (once per thread), which is a valid indicator of independent execution.
> 
>  The failure is a symptom of a race condition in your test itself, not in the code being tested. 

so we agree remove the fragile test as I did in https://github.com/teemtee/tmt/commit/8c3daa9702dfab0079b67432ccd114ef9a64d8e1 , right?^^

skycastlelily:
The https://github.com/teemtee/tmt/commit/7c1bf056daf3e6e41ce065e18a562188ffcdf0ce commit make the status update show more detail.
without that change, there is only one line shows status for multiple jobs
for instance,
status: queue
the jobs may have different status, and we don't know which job is in queue status,
with this change it will show:
status (default-1[J:11616271]): Installing

However,when multiple jobs are submitted, it shows the status of the job which was last submitted ,for most of the time.
I asked cursor to fix it, the patch is more than 100 lines, should we adopt it? :)

tmt/steps/provision/mrack.py
```diff
@@ -840,49 +840,64 @@ def constraint_to_beaker_filter(
...
-        from mrack.errors import NotAuthenticatedError, ProvisioningError
-        from mrack.providers import providers
-        from mrack.providers.beaker import PROVISIONER_KEY as BEAKER
-        from mrack.providers.beaker import BeakerProvider
-        from mrack.transformers.beaker import BeakerTransformer
+            MRACK_VERSION = importlib.metadata.version('mrack')
 
-        MRACK_VERSION = importlib.metadata.version('mrack')
+            # hack: remove mrack stdout and move the logfile to /tmp
+            # Make log file unique per thread to avoid conflicts
```
tcornell-bus: I think this comment has more to do with the code just below this.

tmt/steps/provision/mrack.py
```diff
@@ -840,49 +840,64 @@ def constraint_to_beaker_filter(
...
+                mrack.logger.removeHandler(mrack.console_handler)
+            if hasattr(mrack, 'file_handler'):
+                mrack.logger.removeHandler(mrack.file_handler)
 
-        # hack: remove mrack stdout and move the logfile to /tmp
-        mrack.logger.removeHandler(mrack.console_handler)
-        mrack.logger.removeHandler(mrack.file_handler)
+            # Use thread-safe approach for log file cleanup
+            thread_id = threading.get_ident()
+            log_filename = f"{workdir}/{name}-mrack-{thread_id}.log"
```
tcornell-bus: Do we know why this log file gets removed? While testing with a simple test requiring 3 hosts, I see in the `log.txt` this line:
```
Remove '/var/tmp/tmt/run-015/plans/multihost-beaker/provision/provision-mrack-140137121973952.log'.
```

In addition to the question above, only see 1 `mrack-\d+.log` getting removed, when 3 guests were provisioned. I would have expected to see it 3 times with different thread identifiers. Or maybe I am misunderstanding what I am reading here.

tcornell-bus (commented):
Tested with a 3 host simple multihost test. Provisions much faster :)

tmt/steps/provision/mrack.py
happz: My two cents: these files are pruned by tmt when the step's done. See https://github.com/teemtee/tmt/blob/main/tmt/steps/provision/__init__.py#L3256, it's not aware of plugin-specific files being put into step's workdir, and so they are removed.

happz: And you see only one removed, because there is probably just one such file. `import_and_load_mrack_deps()` might be called from multiple threads, but it runs really just once, under `_MRACK_IMPORT_LOCK` and sets `_MRACK_IMPORTED`. So the first thread wins, we get just one file.

happz: The "thread-safe" dance is pointless, there will never be multiple logs, no conflict there.

skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/3941/commits/28cac843f9cdf43e0d957410e4d3816e0a255e9a ^^

tmt/steps/provision/mrack.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/28cac843f9cdf43e0d957410e4d3816e0a255e9a ^^

tmt/steps/provision/mrack.py
```diff
@@ -1132,38 +1140,43 @@ class BeakerAPI:
...
+            raise ProvisionError("Configuration file 'mrack-provisioning-config.yaml' not found.")
 
+        # Load configuration file
         try:
-            global_context.init(str(mrack_config))
-        except mrack.errors.ConfigError as mrack_conf_err:
-            raise ProvisionError(mrack_conf_err)
+            config_data = tmt.utils.yaml_to_dict(provisioning_config.read_text())
+        except Exception as config_err:
+            raise ProvisionError(f"Failed to load mrack provisioning configuration: {config_err}")
```
happz: Please, review your code, and resolve all occurences of exceptions being embedded into error messages. This shall be `raise ... from ...`.

skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/3941/commits/c136766be952ae83d409e4a9019e38e8a9ae7668 ^^

tmt/steps/provision/mrack.py
```diff
@@ -1253,7 +1295,8 @@ def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
 
         assert isinstance(self.parent, tmt.steps.provision.Provision)
-        import_and_load_mrack_deps(self.parent.workdir, self.parent.name, self._logger)
+        self.mrack_log = f"{self.workdir}/{self.parent.name}-mrack.log"
+        import_and_load_mrack_deps(self.mrack_log, self._logger)
```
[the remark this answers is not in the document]
happz: Right, `self.workdir` should probably be `self.parent.workdir`, i.e. workdir of the `provision` step rather than a workdir of a single guest. We can't have more than one file anyway.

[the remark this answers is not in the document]
skycastlelily: Addressed : https://github.com/teemtee/tmt/pull/3941/commits/4fd0b4cf4c8562dc2947614b75519656a7180938 ^^

tmt/steps/provision/mrack.py
```diff
@@ -1132,38 +1140,45 @@ class BeakerAPI:
...
-            Path(__file__).parent / "mrack/mrack.conf",
-            Path("/etc/tmt/mrack.conf"),
-            Path("~/.mrack/mrack.conf").expanduser(),
-            Path.cwd() / "mrack.conf",
+        # Find mrack provisioning configuration file
+        provisioning_config_locations = [
+            Path(__file__).parent / "mrack/mrack-provisioning-config.yaml",
+            Path("/etc/tmt/mrack-provisioning-config.yaml"),
+            Path("~/.mrack/mrack-provisioning-config.yaml").expanduser(),
+            Path.cwd() / "mrack-provisioning-config.yaml",
```
happz: Hmmm, is this change intended? Is e.g. `/etc/tmt/mrack.conf` no longer needed?

tmt/steps/provision/mrack.py
```diff
@@ -1174,13 +1189,40 @@ async def __init__(self, guest: 'GuestBeaker') -> None:  # type: ignore[misc]
...
+        except Exception as e:
+            raise ProvisionError("Failed to initialize mrack transformer.") from e
 
         self._mrack_provider = self._mrack_transformer._provider
         self._mrack_provider.poll_sleep = DEFAULT_PROVISION_TICK
 
         if guest.job_id:
             self._bkr_job_id = guest.job_id
 
+    def _create_provider_config(self, config_data: dict[str, Any]) -> dict[str, Any]:
```
happz: TBH, I wouldn't bother. If I drop `strategy` key from `config_data['beaker']`, or if I set it to invalid value, we would not notice anyway. This is only a superficial check of some basic keys, not a serious validation - let's not pretend we validate the file.

Can mrack do it? Is there e.g. a JSON schema we could use for actual validation? That would be very useful.

skycastlelily: mrack will validate the configure, but put the keys needed to be validated in the _required_config_attrs list : 

```
class BeakerTransformer(Transformer):
    """Beaker transformer."""

    _config_key = CONFIG_KEY
    _required_config_attrs = [
        "distros",
        "reserve_duration",
        "timeout",
    ]  # List[str]
```
so, removed the _create_provider_config: https://github.com/teemtee/tmt/pull/3941/commits/3cea7fd9d920105c6dc68b049509bb259f1b389b ^^

tmt/steps/provision/mrack.py
tcornell-bus: I think the goal here is to skip over mrack's `global_context` because it's a singleton.

In the old version, `mrack.conf` needed to be fed into `global_context.init()` to get the `global_context.PROV_CONFIG`, and here she's skipping straight to processing the `mrack-provisioning-config.yaml` the same way mrack's global_context would have done it.

I might be misunderstanding something, so please correct me if I am wrong.

thrix: I am looking forward for the time we will remove mrack for this codebase :)

skycastlelily: > Hmmm, is this change intended? Is e.g. /etc/tmt/mrack.conf no longer needed? 

np, really sorry, this patch is my first AI patch, I should not trust it that much,I only checked it could provision multihosts in parallel, and the test passed, I didn't look into the code deeply :sweat_smile: 
Addressed : https://github.com/teemtee/tmt/commit/5eed423d02fc9de2ea3fddeb5e753fba1c517f6f ^^ 

And sorry for failing to notice your comment:sweat_smile: , I'm also really very busy with f43 blockers stuff,as the beta go/not-go meeting is this Thursday

skycastlelily:
> not this patch fault, in general that mrack plugin seems quite messy 

yeah,maybe, kind of, as we don't have an Milos^^ python expert ,but  there is no need to start a server, if users only want to use beaker, I guess that's why it's here in tmt:)

tmt/steps/provision/mrack.py
happz: I see, thanks.

> np, really sorry, this patch is my first AI patch, I should not trust it that much

Exactly. This. Hundred times this. Those tools do not understand the problem. The code they produce must be reviewed thoroughly **before** becoming a PR.

Issue #3927, opened 2025-07-29
provision - virtual - does not show proper memory value


jscotka:
I have HW requirements inside TMT plan and virtual plugin work well with `memory` but output of command is very confusing.

```
provision:
  hardware:
      virtualization:
          is-supported: true
      memory: '>= 8 GB'
```

and output when excute the command `tmt run -vv --all provision --how=virtual --image=https://cloud.centos.org/centos/10-stream/x86_64/images/CentOS-Stream-GenericCloud-x86_64-10-20250722.0.x86_64.qcow2 plan --name granular test --name /tests/keyboard/x/x` is:
```
/plans/granular
summary: Run anaconda tests granurally on CentOS stream images
    discover
        how: fmf
        directory: /home/jscotka/git/anaconda/kickstart-tests
        hash: 2305f5d
        filter: tag: -knownfailure
        summary: 1 test selected
            /tests/keyboard/x/x
    provision
        queued provision.provision task #1: default-0
        
        provision.provision task #1: default-0
        how: virtual
        hardware:
            virtualization:
                is-supported: true
            memory: '>= 8 GB'
        image: https://cloud.centos.org/centos/10-stream/x86_64/images/CentOS-Stream-GenericCloud-x86_64-10-20250722.0.x86_64.qcow2
        memory: 2048 MB
        disk: 40 GB
        warn: Hardware requirement 'virtualization.is-supported' is not supported.
        qcow: CentOS-Stream-GenericCloud-x86_64-10-20250722.0.x86_64.qcow2
        progress: downloading...
        effective hardware:
            virtualization:
                is-supported: true
            memory: >= 8 GB
        name: tmt-024-HFkNAwmq
        key: /var/tmp/tmt/run-024/plans/granular/provision/default-0/id_ecdsa
        progress: booting...
        console: /var/tmp/tmt/run-024/plans/granular/provision/default-0/logs/console.txt
        primary address: 127.0.0.1
        topology address: 127.0.0.1
        port: 10023
            cmd: mkdir -p /usr/local/bin
        multihost name: default-0
        arch: x86_64
        distro: CentOS Stream 10 (Coughlan)
        kernel: 6.12.0-109.el10.x86_64
        package manager: dnf
        selinux: yes
        systemd: yes
        is superuser: yes
        is_container: no

```

what is confusing if the machine will use proper memory setup or not. from these output it seems not used, but in log.txt there is more clear that there is used `8Gb ` not `memory: 2048 MB` as written in this output

[...]

PR #3944, opened 2025-08-05
Memory key in-sync with hardware and provision

AthreyVinay:
Fixes: #3927 

* [X] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


tmt/steps/provision/testcloud.py
```diff
@@ -1073,11 +1057,18 @@ def start(self) -> None:
...
+        )
+        self.info(
+            'disk',
+            f'{tmt.hardware.UNITS(f"{self._domain.storage_devices[0].size} GB").to("GB")}',
+            'green',
+        )
 
-        for i, device in enumerate(self._domain.storage_devices):
-            self.debug(f'final domain disk #{i} size', str(device.size))
+        for i, device in enumerate(self._domain.storage_devices, start=1):
```
AthreyVinay: I felt it made sense to start from 1-based index :)

happz: There are 2 kinds of people: those who start indexing from 0, and those who are wrong :)

I would recommend starting with 0 - 0 is the index of the first element in Python, it's a debugging message and reports content of the whole iterable. Starting with `1` will just confuse readers.

tmt/steps/provision/testcloud.py
```diff
@@ -1073,11 +1057,18 @@ def start(self) -> None:
...
+            size_gb = tmt.hardware.UNITS(f"{device.size} GB").to("GB")
+            self.debug(f'Domain disk {i} size: {size_gb}')
```
happz: `debug()` can also log key/value type of value, like `info()` does above, therefore `self.debug(f'Domain disk {i} size', tmt.hardware.UNITS(f"{device.size} GB").to("GB"))`. And you can save one extra variable.

tmt/steps/provision/testcloud.py
```diff
@@ -1023,23 +1018,12 @@ def start(self) -> None:
             self.image_url = self._guess_image_url(self.image_url)
             self.debug(f"Guessed image url: '{self.image_url}'", level=3)
 
-        # Initialize and prepare testcloud image
+        # Initialize testcloud image
         assert testcloud is not None
         self._image = testcloud.image.Image(self.image_url)
         self.verbose('qcow', self._image.name, 'green')
         if not Path(self._image.local_path).exists():
             self.info('progress', 'downloading...', 'cyan')
```
happz: You should probably move the lines above as well. For example, if the image is not downloaded yet, `downloading...` would be printed, followed by `memory` and `disk`. The actual download would happen - and log - after... So the `self._image = ...` should probably stay attached to the followup code.

AthreyVinay: Did you mean move it down? Sorry, I dont follow. Based on the understanding of your comment, the order needs to be:
`memory` -> `disk` -> `'progress', 'downloading...'` correct?

Currently, my patch would do:
`'progress', 'downloading...'` -> `memory` -> `disk`

happz: Yeah, but the downloading happens in `self._image.prepare()`, so the message announcing it is detached from the point in time when it actually happens. With your change, you announce `downloading...` and start doing something unrelated - sorting out the HW requirements - and then you start downloading. So a lot of stuff in log between `downloading...` and actual download will be unrelated to the download.

The order can be (`downloading...` -> actual download -> sorting out HW requirements -> `memory`, `disk`), or (sorting out HW requirements -> `memory`, `disk` -> `downloading...` -> actual download), but the current implementation is neither of them. Instead, it's an announcement, followed by a step aside, then returning back to the announced action.

happz: In other words, why is it

```python
        # Initialize testcloud image
        self._image = testcloud.image.Image(self.image_url)
        self.verbose('qcow', self._image.name, 'green')
        if not Path(self._image.local_path).exists():
            self.info('progress', 'downloading...', 'cyan')

        # Now do something completely different for many lines
        ...

        # Prepare testcloud image
        try:
            self._image.prepare()
        except FileNotFoundError as error:
            raise ProvisionError(f"Image '{self._image.local_path}' not found.") from error
        except (testcloud.exceptions.TestcloudPermissionsError, PermissionError) as error:
            raise ProvisionError(
                f"Failed to prepare the image. Check the '{self.testcloud_image_dirpath}' "
                f"directory permissions."
            ) from error
        except KeyError as error:
            raise ProvisionError(f"Failed to prepare image '{self.image_url}'.") from error
```

instead of

```python
        # Initialize testcloud image
        self._image = testcloud.image.Image(self.image_url)
        self.verbose('qcow', self._image.name, 'green')
        if not Path(self._image.local_path).exists():
            self.info('progress', 'downloading...', 'cyan')
        # Prepare testcloud image
        try:
            self._image.prepare()
        except FileNotFoundError as error:
            raise ProvisionError(f"Image '{self._image.local_path}' not found.") from error
        except (testcloud.exceptions.TestcloudPermissionsError, PermissionError) as error:
            raise ProvisionError(
                f"Failed to prepare the image. Check the '{self.testcloud_image_dirpath}' "
                f"directory permissions."
            ) from error
        except KeyError as error:
            raise ProvisionError(f"Failed to prepare image '{self.image_url}'.") from error

        # Now do something completely different for many lines
        ...
```

or

```python
        # Now do something completely different for many lines
        ...

        # Initialize testcloud image
        self._image = testcloud.image.Image(self.image_url)
        self.verbose('qcow', self._image.name, 'green')
        if not Path(self._image.local_path).exists():
            self.info('progress', 'downloading...', 'cyan')
        # Prepare testcloud image
        try:
            self._image.prepare()
        except FileNotFoundError as error:
            raise ProvisionError(f"Image '{self._image.local_path}' not found.") from error
        except (testcloud.exceptions.TestcloudPermissionsError, PermissionError) as error:
            raise ProvisionError(
                f"Failed to prepare the image. Check the '{self.testcloud_image_dirpath}' "
                f"directory permissions."
            ) from error
        except KeyError as error:
            raise ProvisionError(f"Failed to prepare image '{self.image_url}'.") from error
```

AthreyVinay: Thanks - the idea suggested makes absolute sense.
The thought process was to keep the heavier tasks to the end which is currently whats happening but display eagerly. But as pointed out, the timing would not be accurate as yeah, we do "other stuff" after the `downloading` is done which is un-related. Feedback taken - will be fixed.

tmt/steps/provision/testcloud.py
AthreyVinay: I understand that python has 0 based index, but from a logging standpoint, I thought(debatable) this looks more consumable given 0 does not hold significant value in this context:
`Domain disk 1 size: 40 GB`

happz: I understand that. Yet, it's listing a sequence, and the message targets developers rather than mere users, I believe starting with 0 would make it clear the first message is indeed the first item of the sequence as well.

tests/provision/memory/test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        rlRun "pushd data"
+        rlRun "set -o pipefail"
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
```
happz: The default value, `container`, will probably never happen under tmt, since test tags mark it as `virtual`-only. Since it's targeting `virtual` plugin only, making the default `virtual` here would be better.

tmt/steps/provision/testcloud.py
AthreyVinay: Reverted.

tests/provision/memory/test.sh
AthreyVinay: Yep, nice catch!
Fixed: [a4284b](https://github.com/teemtee/tmt/pull/3944/commits/a4284bd3c695c9fb2b143dc4a8eb592e2822de88)

happz:
@AthreyVinay See https://artifacts.osci.redhat.com/testing-farm/f20b2194-8191-4753-be86-9a4ecf3d1079/, `/tests/provision/virtual.testcloud` needs attention as well.

PR #3947, opened 2025-08-07
add bootc support for prepare shell step

ukulekek:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


happz:
I'm afraid bootc guest implementation, its internals, are leaking into plugins. This way, the plugin would have to be aware to which guest it talks to. Let's say there would be a 3rd-party `prepare` plugin, it would need to be updated to recognize `bootc` and realize it needs to be treated differently than every other plugin.

Instead, what we need is a way to "hide" the different behaviour behind the `Guest` class API, so it would be transparent of its users, which would then allow all the plugins to use it without taking extra steps for one guest class.

IIUIC, the goal is for `prepare/shell` to not run on the live guest, but instead collect commands, rebuild the image, and reboot the guest from it? Like the installation of requirements? With package manager, we managed to separate the "command construction" and "command execution", which lets the `bootc` package manager to hook into the process, and instead of running commands are collected. Would it make sense to change implementation got `bootc`'s guest class, to become aware of the different mode of operations? Regular guests executing things immediately, `bootc` guest collecting them for the future image. I would investigate this path instead.

Actually, what commands are supposed to be executed on the live bootc-powered guest? The more I think about it, isn't it just the test itself? Or are there more types of command we do not want to happen on the live guest?

ukulekek:
@happz, hi, thanks for looking at it. Hiding it in the `Guest` class sounds like a good idea. I'll take a look at it. 
> Actually, what commands are supposed to be executed on the live bootc-powered guest? The more I think about it, isn't it just the test itself? Or are there more types of command we do not want to happen on the live guest?

Yeah, I believe we want to run only test commands live on the guest, the rest should go through the build, reboot cycle.

happz:
> @happz, hi, thanks for looking at it. Hiding it in the `Guest` class sounds like a good idea. I'll take a look at it.
> 
> > Actually, what commands are supposed to be executed on the live bootc-powered guest? The more I think about it, isn't it just the test itself? Or are there more types of command we do not want to happen on the live guest?
> 
> Yeah, I believe we want to run only test commands live on the guest, the rest should go through the build, reboot cycle.

There is a `test_session` parameter accepted by `Guest.execute()` methods, if it's set, the command is **the** test itself. So, naively, I would try to inspect this parameter, and if it's not set, just collect the commands instead of executing them.

What I don't know right now is how to perform the "commit" of these collected commands, when to turn them into an image. Let's say you have `prepare` phase which would run a couple of scripts. Guest would collect them, `prepare` ends - and the guest is not notified it's time to flush the buffer. There is no explicit marker or callback telling guests the current phase is completed. We might need one.

It would be possible to just collect and build the image once you see `test_session=True` - we're going to run the test, so let's turn all collected commands into an image, reboot, and then resume and run the test command like nothing happened. Delayed, lazy approach. But this won't work for `finish`, for example, there's no test after `finish` step, so commands from `finish` would remain in the buffer. Adding an explicit "on_start"/"on_finish" events to a phase/step workflow would be probably better. But it's an immediate idea.

ukulekek:
@happz thanks. Do you think `Guest` can use `Bootc` and `BootEngine` classes, or should they be necessarily merged?

ukulekek:
My bad, we still need to execute some commands live on boot, for example, the probe_command of package managers is one case. Also, checking the presence of packages and retrieving the current running image from the bootc command are executed live as well.

ukulekek:
Notes:
* While executing Guest.execute(), the function can't determine if it needs to add the command to the Containerfile or run it immediately. 
    * The idea is to add the `make_changes=True` option to the Guest.execute(), go through the code and mark every command which doesn't make changes with `make_changes=False`. For example, probe commands and checks of package presence should be marked.
* `tmt` can't determine when the constructed Containerfile should be built.
    * See happz's comment above: https://github.com/teemtee/tmt/pull/3947#issuecomment-3173688800 

Also, the Guest.push() should have the ability to add `COPY ...` lines to the Containerfile. At first glance, it's not necessary as the `/var` folder is mounted after the reboot to the built container, so the content remains. However, it is needed to support the prepare shell with URL + script case, such as:
```
prepare:
    how: shell
    url: https://github.com/teemtee/tmt.git
    ref: main
    script: cd $TMT_PREPARE_SHELL_URL_REPOSITORY && make docs
```
In this case, the script should be executed during the build process, but the content isn't available yet without adding `COPY ...`.

psss:
@ukulekek, hi! What is the status / plans with this one?
@thrix, are you going to finish it to cover #4495 from the current sprint?
Or shall it be dropped as outdated?

thrix:
Closing in favor of a new implemention, draft to be opened today.

Issue #2995, opened 2024-06-06
Add 


lukaszachy:

AthreyVinay:
Adding a menu option to recognize `!` would need a code change to `choose_action()` too to accommodate this. Hence, I went with `host -> h` option.  Open to suggestions :)

PR #3963, opened 2025-08-13
Implement running command on host as a menu option

AthreyVinay:
Resolves: #2995 

Pull Request Checklist

* [X] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -7,6 +7,12 @@
 tmt-1.56.0
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+The :ref:`tmt try</stories/cli/try>` command's interactive session
+now includes a new ``h`` option to run shell commands directly
+on the host (not on provisioned guests). Users can exit cleanly
+by typing ``'\q'``, pressing Ctrl-C, or pressing Ctrl-D, and failed commands
+are reported with clear error messages.
+
 When interrupted while running a test, tmt will now interrupt the test
 and quit immediately after, skipping all remaining tests. This is now
 aligned with a test abort or :ref:`exit-first </plugins/execute/tmt>`
--- a/tmt/trying.py
+++ b/tmt/trying.py
@@ -4,6 +4,7 @@ Easily try tests and experiment with guests
 
 import enum
 import re
+import shlex
 import textwrap
 from collections.abc import Iterator
 from typing import Any, cast
@@ -25,7 +26,7 @@ import tmt.utils
 from tmt import Plan
 from tmt.base import RunData
 from tmt.steps.prepare import PreparePlugin
-from tmt.utils import GeneralError, MetadataError, Path
+from tmt.utils import Command, GeneralError, MetadataError, Path
 from tmt.utils.themes import style
 
 USER_PLAN_NAME = "/user/plan"
@@ -38,6 +39,7 @@ class Action(enum.Enum):
 
     TEST = "t", "rediscover tests and execute them again"
     LOGIN = "l", "log into the guest for experimenting"
+    HOST = "h", "run command on the local host"
     VERBOSE = "v", "set the desired level of verbosity"
     DEBUG = "b", "choose a different debugging level"
 
@@ -313,6 +315,7 @@ class Try(tmt.utils.Common):
 
                         {Action.TEST.menu}
                         {Action.LOGIN.menu}
+                        {Action.HOST.menu}
                         {Action.VERBOSE.menu}
                         {Action.DEBUG.menu}
 
@@ -527,6 +530,36 @@ class Try(tmt.utils.Common):
         run_id = style(str(plan.my_run.workdir), fg="magenta")
         self.print(f"Run {run_id} successfully finished. Bye for now!")
 
+    def action_host(self, plan: Plan) -> None:
+        """
+        Run command on the host
+        """
+
+        while True:
+            self.print(style(f"Enter command (or '\\{Action.QUIT.key}' to quit): ", fg="green"))
+            try:
+                raw_command = input("> ")
+            except (KeyboardInterrupt, EOFError):
+                self.print("Exiting host command mode. Bye for now!")
+                break
+
+            if not raw_command or raw_command == f'\\{Action.QUIT.key}':
+                self.print("Exiting host command mode. Bye for now!")
+                break
+
+            # Execute the command on the host
+            try:
+                Command(*shlex.split(raw_command)).run(
+                    cwd=plan.workdir, logger=self._logger, interactive=True
+                )
+            except tmt.utils.RunError as error:
+                tmt.utils.show_exception_as_warning(
+                    exception=error,
+                    message=f"'{raw_command}' command failed to run.",
+                    include_logfiles=True,
+                    logger=self._logger,
+                )
+
     def handle_options(self, plan: Plan) -> None:
         """
         Choose requested cli option
@@ -647,14 +680,13 @@ class Try(tmt.utils.Common):
             self.action_verbose(plan)
 
         # Choose the initial action
+        action = Action.START_ASK
         if self.opt("login"):
             action = Action.START_LOGIN
         elif self.opt("ask"):
-            action = Action.START_ASK
+            pass  # already START_ASK
         elif self.tests:
             action = Action.START_TEST
-        else:
-            action = Action.START_ASK
 
         # Loop over the actions
         try:
--- a/tests/unit/test_cli_actions.py
+++ b/tests/unit/test_cli_actions.py
@@ -1,0 +1,92 @@
+from unittest import mock
+
+import pytest
+
+from tmt._compat.pathlib import Path
+from tmt.trying import Try
+from tmt.utils import RunError, style
+
+
+class TestHostCommandExecution:
+    @pytest.fixture
+    def mock_try_instance(self):
+        return mock.MagicMock()
+
+    @pytest.fixture
+    def mock_plan(self):
+        plan = mock.MagicMock()
+        plan.workdir = Path("/test/workdir")
+        return plan
+
+    @pytest.mark.parametrize(
+        ("inputs", "effects", "expected_outputs"),
+        [
+            (
+                ["pwd", r"\q"],
+                [None],
+                [
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    "Exiting host command mode. Bye for now!",
+                ],
+            ),
+            (
+                ["abcde", r"\q"],
+                [RunError("abcde command failed to run.", "abcde", 127)],
+                [
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    "Exiting host command mode. Bye for now!",
+                ],
+            ),
+            (
+                ["pwd", "ls -la", ""],
+                [None, None],
+                [
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    style(r"Enter command (or '\q' to quit): ", fg="green"),
+                    "Exiting host command mode. Bye for now!",
+                ],
+            ),
+            *[
+                (
+                    [exc],
+                    [],
+                    [
+                        style(r"Enter command (or '\q' to quit): ", fg="green"),
+                        "Exiting host command mode. Bye for now!",
+                    ],
+                )
+                for exc in (KeyboardInterrupt, EOFError)
+            ],
+        ],
+        ids=(
+            'Single command success',
+            'Command failure',
+            'Multiple commands success',
+            'KeyboardInterrupt',
+            'EOFError',
+        ),
+    )
+    @mock.patch("tmt.utils.show_exception_as_warning")
+    @mock.patch("tmt.utils.Command.run")
+    @mock.patch("builtins.input")
+    def test_action_host(
+        self,
+        mock_input,
+        mock_command_run,
+        mock_show_exception,
+        mock_try_instance,
+        mock_plan,
+        inputs,
+        effects,
+        expected_outputs,
+    ):
+        mock_input.side_effect = inputs
+        mock_command_run.side_effect = effects
+
+        Try.action_host(mock_try_instance, mock_plan)
+
+        printed = [args[0] for args, _ in mock_try_instance.print.call_args_list]
+        assert printed == expected_outputs


tmt/trying.py
```diff
@@ -527,6 +532,28 @@ def action_quit(self, plan: Plan) -> None:
...
+            try:
+                command = input("> ")
+            except (KeyboardInterrupt, EOFError):
+                break
+
+            if not command or command == Action.QUIT.key:
+                break
+
+            # Execute the command on the host
+            # TODO: Maybe we should capture the output, timeout etc?
```
happz: Yep, check out `Command.run()` from `tmt.utils`. We should use it instead of subprocess, for its API, consistent logging, timeout support, etc.

AthreyVinay: Cool, tks.

AthreyVinay: Refactored to use `Command.run()` here: [03399f2](https://github.com/teemtee/tmt/pull/3963/commits/03399f2262a5e78562882af5ef001477b1e534e6)

tmt/trying.py
```diff
@@ -527,6 +532,32 @@ def action_quit(self, plan: Plan) -> None:
...
+                break
+
+            # Execute the command on the host
+            try:
+                command = Command(*shlex.split(raw_command))
+                if result := command.run(cwd=plan.workdir, logger=self._logger):
+                    output = result.stdout.strip() if result.stdout is not None else ""
+                    self.print(f"{raw_command}: {output}")
+            except tmt.utils.RunError:
+                self.print(f"Failed to run command '{raw_command}' on the host: command not found")
```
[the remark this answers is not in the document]
AthreyVinay: The exception is handled in the higher stack(`run()`) and would raise a `FileNotFoundError` which is not desired in this case.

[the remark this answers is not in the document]
happz: What if the reason was different? `Command.run()` will raise the exception for non-zero exit code too, `ls / && exit 1` will lead to the same exception, but with very different cause.

tmt/trying.py
```diff
@@ -527,6 +532,32 @@ def action_quit(self, plan: Plan) -> None:
...
+            if not raw_command or raw_command == Action.QUIT.key:
+                self.print("Exiting host command mode. Bye for now!")
+                break
+
+            # Execute the command on the host
+            try:
+                command = Command(*shlex.split(raw_command))
+                if result := command.run(cwd=plan.workdir, logger=self._logger):
+                    output = result.stdout.strip() if result.stdout is not None else ""
+                    self.print(f"{raw_command}: {output}")
```
happz: Isn't the first line of multiline output going to be mangled, by being prefixed with the command?

```
ls -al /: total 24
dr-xr-xr-x.   1 root root  158 Dec 29  2024 .
dr-xr-xr-x.   1 root root  158 Dec 29  2024 ..
```

tmt/trying.py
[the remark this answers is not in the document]
AthreyVinay: It still doesn't warrant as a `FileNotFoundError` isnt it? Maybe a `BaseException` to handle different cause?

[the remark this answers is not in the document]
happz:
```
>>> from tmt.utils import Command, ShellScript, Path
>>> from tmt.log import Logger
>>> ShellScript('ls -al / && exit 1').to_shell_command().run(cwd=Path.cwd(), logger=Logger.get_bootstrap_logger())
  File "<python-input-12>", line 1, in <module>
    ShellScript('ls -al / && exit 1').to_shell_command().run(cwd=Path.cwd(), logger=Logger.get_bootstrap_logger())
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/happz/git/tmt/tmt/utils/__init__.py", line 1404, in run
    raise RunError(
    ...<6 lines>...
    )
tmt.utils.RunError: Command '/bin/bash -c 'ls -al / && exit 1'' returned 1.
>>>
```

As you can see, `RunError` does not mean "command not found", the command was found and failed. Capturing the `RunError` and telling user "command not found" is not correct handling of this exception, as it may have many different causes. It simply means "command failed", and only sometimes it's because "command not found".

tmt/trying.py
AthreyVinay: Nice one!
Questioned the need to prefix the output with the `raw_command` and hence removed it. Much cleaner now: [d7d7603](https://github.com/teemtee/tmt/pull/3963/commits/d7d76031b2a1ec6032e82f55f2dd73461b2104c0)
Thanks!

happz: Can you try running the command with `interactive=True`? That should stream the output directly.

tmt/trying.py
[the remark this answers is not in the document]
AthreyVinay: AHA! That makes sense..... sorry @graphite :D
Fixed here: [518ac23](https://github.com/teemtee/tmt/pull/3963/commits/518ac2370c3dd734de2691deeede2d6cd2c39651)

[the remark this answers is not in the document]
happz: See `tmt.utils.show_exception_as_warning()` instead of `print()`, more verbose, true, but includes traceback if needed, and saves logs in files as well.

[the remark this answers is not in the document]
AthreyVinay: Cool - more learning for me, thanks for the quick feedback @happz !

tmt/trying.py
AthreyVinay: Fixed here: [3636c4d](https://github.com/teemtee/tmt/pull/3963/commits/3636c4dc60c34e0ed17e3b72e8f6280765e5b9ad)

tmt/trying.py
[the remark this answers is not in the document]
AthreyVinay: Fixed here: [3636c4d](https://github.com/teemtee/tmt/pull/3963/commits/3636c4dc60c34e0ed17e3b72e8f6280765e5b9ad)

docs/releases.rst
```diff
@@ -4,6 +4,16 @@
     Releases
 ======================
 
+tmt-1.56.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :ref:`tmt try</stories/cli/try>` command's interactive session
+now includes a new ``-h`` option to run shell commands directly
```
happz: Is it really `-h`, or just `h`? Is it an option, or command in the interactive session?

docs/releases.rst
```diff
@@ -4,6 +4,16 @@
...
+on the host (not on provisioned guests). Users can exit cleanly
+with ``'q'``, KeyboardInterrupt, or EOFError, and failed commands
```
happz: Users don't really deal with `EOFError`s or `KeyboardInterrupt`s. Instead, they deal with "pressing `Ctrl-D`" or similar actions.

Also, I would propose using different command rather than simple `q`, users may easily have a script called `q`. `\q` is pretty common, backslash making it more special and less prone to conflicts with actual commands in the live Linux system.

tmt/trying.py
```diff
@@ -527,6 +532,35 @@ def action_quit(self, plan: Plan) -> None:
...
+                break
+
+            if not raw_command or raw_command == Action.QUIT.key:
+                self.print("Exiting host command mode. Bye for now!")
+                break
+
+            # Execute the command on the host
+            try:
+                command = Command(*shlex.split(raw_command))
+                command.run(cwd=plan.workdir, logger=self._logger, interactive=True)
```
happz: `command` exists only to be used in `command.run()`. Does it have to exist? The following would work as well:

```python
Command(*shlex.split(raw_command)).run(cwd=plan.workdir, logger=self._logger, interactive=True)
```

AthreyVinay: I indeed thought of this, but left it for better debugging.

happz: If the debugging of `Command(...).run(...)` is subpar, we should improve it. Please, file an issue of your findings, what's missing e.g. when reading a traceback - fixing it in `Command` would improve situation for all call sites, `command = ...` here would make it better just for the single one.

tmt/trying.py
```diff
@@ -527,6 +532,35 @@ def action_quit(self, plan: Plan) -> None:
...
+            except tmt.utils.RunError as error:
+                tmt.utils.show_exception_as_warning(
+                    exception=error,
+                    message=error.message,
```
happz: Exception message would be logged as part of the traceback by default. We can use this parameter to provide more contextual message than possibly a raw exception output. E.g. "Failed to run command on the host" or something similar.

tmt/trying.py
AthreyVinay: Makes sense @happz !

tmt/trying.py
AthreyVinay: Addressed here: [6b0fcd65](https://github.com/teemtee/tmt/pull/3963/commits/6b0fcd65044ece38c0a55cc75fc4ce104d67d9a3)

tmt/trying.py
AthreyVinay: Addressed here: [6b0fcd65](https://github.com/teemtee/tmt/pull/3963/commits/6b0fcd65044ece38c0a55cc75fc4ce104d67d9a3)

docs/releases.rst
AthreyVinay: Addressed here: [6b0fcd65](https://github.com/teemtee/tmt/pull/3963/commits/6b0fcd65044ece38c0a55cc75fc4ce104d67d9a3)

docs/releases.rst
AthreyVinay: Good catch, must be just `h`.
Addressed here: [6b0fcd65](https://github.com/teemtee/tmt/pull/3963/commits/6b0fcd65044ece38c0a55cc75fc4ce104d67d9a3)

tmt/trying.py
```diff
@@ -25,7 +26,7 @@
 from tmt import Plan
 from tmt.base import RunData
 from tmt.steps.prepare import PreparePlugin
-from tmt.utils import GeneralError, MetadataError, Path
+from tmt.utils import Command, GeneralError, MetadataError, Path
```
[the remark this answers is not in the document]
AthreyVinay: It's fully qualified(`tmt.utils.RunError`) and will not raise `NameError`.

tmt/trying.py
```diff
@@ -527,6 +532,36 @@ def action_quit(self, plan: Plan) -> None:
...
+                break
+
+            if not raw_command or raw_command == f'\\{Action.QUIT.key}':
+                self.print("Exiting host command mode. Bye for now!")
+                break
+
+            # Execute the command on the host
+            try:
+                Command(*shlex.split(raw_command)).run(
+                    cwd=plan.workdir, logger=self._logger, interactive=True
```
AthreyVinay, line 553: @happz Just wondering about the ask on rediscovering tests, the `cwd` should probably point to `os.getcwd()` rather than `plan.workdir`. WDYT?

AthreyVinay:
Moving it back to implement as [this](https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1755252226235369) needs to be supported.

happz:
> Moving it back to implement as [this](https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1755252226235369) needs to be supported.

Does it need to be implemented in *this* PR? I for one would be perfectly fine with second PR, building on top of this one, adding the capability of running a test after `cd` on the runner.

tmt/trying.py
happz: Maybe. I would leave it for the "yeah, I can `cd`, can I also run a test from that new path?" patch.

AthreyVinay:
> > Moving it back to implement as [this](https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1755252226235369) needs to be supported.
> 
> Does it need to be implemented in _this_ PR? I for one would be perfectly fine with second PR, building on top of this one, adding the capability of running a test after `cd` on the runner.

Makes sense, I will leave this PR to just `run_command_on_host` and will try to work on a follow-up PR when more details are gathered.

tmt/trying.py
AthreyVinay: Leaving this as-is i.e. `cwd=plan.workdir`

AthreyVinay:
>Makes sense, I will leave this PR to just run_command_on_host and will try to work on a follow-up PR when more details are gathered.

@happz I have created a new PR in favor of 2 new menu options, that's more towards what's needed. So maybe just close this and focus on: #3976 ?

happz:
> > Makes sense, I will leave this PR to just run_command_on_host and will try to work on a follow-up PR when more details are gathered.
> 
> @happz I have created a new PR in favor of 2 new menu options, that's more towards what's needed. So maybe just close this and focus on: #3976 ?

My two cents: I would finish this first. It’s something that almost done, it works, it does stuff. Running command on the host is worth on its own, running test from changed directory is different feature - related, but different.

AthreyVinay:
> > > Makes sense, I will leave this PR to just run_command_on_host and will try to work on a follow-up PR when more details are gathered.
> > 
> > 
> > @happz I have created a new PR in favor of 2 new menu options, that's more towards what's needed. So maybe just close this and focus on: #3976 ?
> 
> My two cents: I would finish this first. It’s something that almost done, it works, it does stuff. Running command on the host is worth on its own, running test from changed directory is different feature - related, but different.

SGTM :) At this point, I think the feature is finished(apart from failing tests obviously)., thanks.

happz (approved):

therazix (approved):

tmt/trying.py
```diff
@@ -51,6 +52,8 @@ class Action(enum.Enum):
     KEEP = "k", "exit the session but keep the run for later use"
     QUIT = "q", "clean up the run and quit the session"
 
+    HOST = "h", "run command on host"
```
psss:
```suggestion
    HOST = "h", "run command on the local host"
```

What about making the description a bit more specific?

tmt/trying.py
```diff
@@ -325,6 +328,8 @@ def choose_action(self) -> Action:
 
                         {Action.KEEP.menu}
                         {Action.QUIT.menu}
+
+                        {Action.HOST.menu}
```
psss: What about moving the command to the first section just under the `login`? Seems those two are kind of related. And here, at the very end, `host` seems quite lonely :-)

tmt/trying.py
AthreyVinay: Refactored here: [1ecbe99](https://github.com/teemtee/tmt/pull/3963/commits/1ecbe99c80487cd2b7c24ada822f1715f54386a5), thanks

tmt/trying.py
AthreyVinay: Refactored here: [1ecbe99](https://github.com/teemtee/tmt/pull/3963/commits/1ecbe99c80487cd2b7c24ada822f1715f54386a5), thanks

tmt/trying.py
psss: Thanks.

tmt/trying.py
psss: Thanks.

tmt/trying.py
```diff
@@ -527,6 +530,36 @@ def action_quit(self, plan: Plan) -> None:
...
+
+        while True:
+            self.print(style(f"Enter command (or '\\{Action.QUIT.key}' to quit): ", fg="green"))
+            try:
+                raw_command = input("> ")
+            except (KeyboardInterrupt, EOFError):
+                self.print("Exiting host command mode. Bye for now!")
+                break
+
+            if not raw_command or raw_command == f'\\{Action.QUIT.key}':
```
psss, line 546: Just realized one more thing: Wouldn't it be a bit better to keep the quit action consistent with the main menu? We could use the `Action.find(raw_command)` here to quickly check for both `q` and `quit`. I find the `\q` a little bit confusing and I don't expect there would be collisions with other existing commands.

happz: https://github.com/teemtee/tmt/pull/3963#discussion_r2284396561

psss: Ok, I see, different people, different expectations :sunglasses:

thrix:
squashed and rebased on top of main

thrix:

* for F43 see https://issues.redhat.com/browse/TFT-3621
* a possible problem from #3963 - see https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1756327635069479?thread_ts=1756295426.357349&cid=C04LRPNDZ4Y

PR #3939, opened 2025-08-01
Release 1.55.0

bajertom:
### Follow the steps below to create a new major or minor release:

- [x] Update overview.rst with new contributors since the last release
- [x] Review the release notes in releases.rst, update as needed
- [x] Add a Release x.y.z commit, empty if needed: git commit --allow-empty -m "Release x.y.z"
- [x] Create a pull request with the commit, ensure tests pass, merge it
- [x] Move the fedora branch to point to the new release
- [x] Tag the commit with x.y.z, push tags git push --tags

### Create a new github release based on the tag above

- [x] Mention the most important changes in the name, do not include version
- [x] Use ; as a delimiter, when multiple items are mentioned in the name
- [x] Push the “Generate release notes” button to create the content
- [x] Prepend the “See the release notes for the list of interesting changes.” line
- [x] Publish the release, check Fedora pull requests, make sure tests pass and merge

### Finally, if everything went well

- [x] Close the corresponding release milestone
- [x] Once the non development [copr build](https://copr.fedorainfracloud.org/coprs/g/teemtee/stable/builds/) is completed, run the [publish-images](https://github.com/teemtee/tmt/actions/workflows/publish-images.yml) workflow to build fresh container image.

### Handle manually what did not went well

- [x] If the automation triggered by publishing the new github release was not successful, publish the fresh code to the pypi repository manually using make wheel && make upload
- [x] If there was a problem with creating Fedora pull requests, you can trigger them manually using /packit propose-downstream in any open issue.

PR #3966, opened 2025-08-18
Silence mypy/pyright reporting issues in ReST rendering

happz:
Docutils marks some classes as deprecated, and suddenly, a `cast()` is no longer needed.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,7 @@ Homepage = "https://github.com/teemtee/tmt"
 [tool.hatch.version]
 source = "vcs"
 raw-options.version_scheme = "release-branch-semver"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -109,6 +109,7 @@ Homepage = "https://github.com/teemtee/tmt"
 [tool.hatch.version]
 source = "vcs"
 raw-options.version_scheme = "release-branch-semver"
+raw-options.git_describe_command = ["git", "describe", "--dirty", "--tags", "--long", "--abbrev=9", "--match", "*[0-9]*"]
```
psss, line 112: Why is the explicit `git describe` command needed?

psss: Ah, have just noticed the commit description. Sorry for the noise.

psss:
Failing jobs were just rawhide failing to boot and a known `beakerlib` issue. Those will be rerun in #3939.

Issue #3949, opened 2025-08-08
Do not create `/usr/local/bin` if it already exists


frantisekz:
I have a container built by Containerfile that chmods /usr/local/bin to 777. I have the following tmt plan:

```
summary: Run test in container
execute:
  how: tmt
discover:
  how: fmf
provision:
  how: container
  user: default
  become: false
  image: localhost/ai-test:latest
report:
  - how: display
  - how: html
```

Yet, tmt still attempts to execute `sudo mkdir -p /usr/local/bin` during provision. The folder is already there, with sufficient permissions.

psss:
Relevant code is located here:

https://github.com/teemtee/tmt/blob/0aef85ca59930937171ea3fb8bc147aaff019ba4/tmt/steps/provision/__init__.py#L1388-L1394

@AthreyVinay, @vaibhavdaren, anyone interested to fix this?

happz:
Is the `mkdir -p` command failing somehow? The directory exists, `-p` should be smart enough to do nothing and move on. What kind of error this produces? In other words: "The folder is already there, with sufficient permissions." - ok, but what's the harm of `mkdir -p`?

AthreyVinay:
>@AthreyVinay, @vaibhavdaren, anyone interested to fix this?

Sure - I can take a look at this. 

Interesting to know what @happz has asked for: `What kind of error this produces?`

AthreyVinay:
Test environment:
1. `Dockerfile`:
```yaml
FROM fedora:latest

RUN dnf install -y python3 python3-pip sudo bash

RUN mkdir -p /usr/local/bin
RUN chmod 777 /usr/local/bin

WORKDIR /test
USER default
```
2. `plan`: same as  mentioned in the issue description.

Test run:
`stat /usr/local/bin` -> `tmt run plan --name samplePlan` -> `stat /usr/local/bin`
The before and after `inode` and `timestamp` of the directory are the same (hence just pasting one):
```bash
/usr/local/bin
  Size: 0               Blocks: 0          IO Block: 4096   directory
Device: 0,39    Inode: 53907       Links: 1
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:bin_t:s0
Access: 2025-08-18 08:24:24.156253020 +0100
Modify: 2025-07-30 01:00:00.000000000 +0100
Change: 2025-08-18 08:23:58.932253008 +0100
[...]

PR #3978, opened 2025-08-21
Add helper scripts installation logic to `local` provision

### LINKED ISSUE #1853: Make `tmt-*` helper scripts available for the `local` provision

mcasquer:
Copy tmt helper scripts under a new `scripts` directory that will be created under the running workdir. Opportunistically, raises a exception if `TMT_SCRIPTS_DIR` env is set for `local` provisioning which is reflected in the documentation ass well.

Closes #1853 

* [x] implement the feature
* [x] write the documentation
* [x] include a release note


mcasquer:
When running, e.g. `tmt --feeling-safe run --all plan --name beaker -v` (see the importance of verbose here)

We can observe how the scripts are installed:
```bash
provision
        
        how: local
        order: 50
        primary address: localhost
        Installing script 'tmt-abort' to '/usr/local/bin/tmt-abort'.
        Installing script 'tmt-abort' to '/usr/local/bin/rstrnt-abort'.
        Installing script 'tmt-abort' to '/usr/local/bin/rhts-abort'.
        Installing script 'tmt-file-submit' to '/usr/local/bin/tmt-file-submit'.
        Installing script 'tmt-file-submit' to '/usr/local/bin/rstrnt-report-log'.
        Installing script 'tmt-file-submit' to '/usr/local/bin/rhts-submit-log'.
        Installing script 'tmt-file-submit' to '/usr/local/bin/rhts_submit_log'.
        Installing script 'tmt-reboot' to '/usr/local/bin/tmt-reboot'.
        Installing script 'tmt-reboot' to '/usr/local/bin/rstrnt-reboot'.
        Installing script 'tmt-reboot' to '/usr/local/bin/rhts-reboot'.
        Installing script 'tmt-reboot-core' to '/usr/local/bin/tmt-reboot-core'.
        Installing script 'tmt-report-result' to '/usr/local/bin/tmt-report-result'.
        Installing script 'tmt-report-result' to '/usr/local/bin/rstrnt-report-result'.
        Installing script 'tmt-report-result' to '/usr/local/bin/rhts-report-result'.
```
And verify the existence of the journal.xml
```bash
journal.txt (/var/tmp/tmt/run-012/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/good-2/journal.txt)
                    failures.yaml (/var/tmp/tmt/run-012/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/good-2/failures.yaml)
                    journal.xml (/var/tmp/tmt/run-012/plan/beakerlib/execute/data/guest/default-0/test/beakerlib/good-2/journal.xml)
                ..:..:.. pass /Test (subresult)
```

happz:
Related issue: https://github.com/teemtee/tmt/issues/3949

tmt/steps/provision/local.py
```diff
@@ -136,7 +136,31 @@ def execute(
...
+
+        # Install all scripts on local guest
+        for script in scripts:
+            if not script.enabled(self):
+                continue
+
+            with script as source:
+                for filename in [script.source_filename, *script.aliases]:
+                    destination = script.destination_path or self.scripts_path / filename
+                    self.verbose(f"Installing script '{source.name}' to '{destination}'.")
```
happz: I'd drop this verbose logging. No other plugin shares info on scripts in this way.

thrix:
From discussion form hacking meeting:

* we agreed that we will go forward with this implementation
* `tmt` scripts dir should be set to non-standard path in case of the local provisioner, to mitigate overwriting files from other packages (TMT_SCRIPTS_DIR)

thrix:
@mcasquer howdy, will you have time to update this change for the release?

mcasquer:
> @mcasquer howdy, will you have time to update this change for the release?

Hey @thrix I asked you yesterday internally:

1. Which is the point with the `TMT_SCRIPTS_DIR`, in my case I am using `self.scripts_path`, should I switch to this variable?
2. Is there any concerns regarding the `sudo` usage in `local` provision and the `/usr/local/bin` directory creation that I need to keep in mind?

cc @psss

happz:
> > @mcasquer howdy, will you have time to update this change for the release?
> 
> Hey @thrix I asked you yesterday internally:
> 
> 1. Which is the point with the `TMT_SCRIPTS_DIR`, in my case I am using `self.scripts_path`, should I switch to this variable?
> 2. Is there any concerns regarding the `sudo` usage in `local` provision and the `/usr/local/bin` directory creation that I need to keep in mind?
> 
> cc @psss

From the implementation point of view, leaving the theory to @thrix and @psss, a different directory on localhost would be added by overriding `scripts_path` property in the `local` plugin/guest. See the default implementation which already uses different default when ostree-ish system is used; `local` would simply provide its own implementation, calling `tmt.steps.scripts.effective_scripts_dest_dir` with its own special default.

mcasquer:
> From the implementation point of view, leaving the theory to @thrix and @psss, a different directory on localhost would be added by overriding `scripts_path` property in the `local` plugin/guest. See the default implementation which already uses different default when ostree-ish system is used; `local` would simply provide its own implementation, calling `tmt.steps.scripts.effective_scripts_dest_dir` with its own special default.

Ok, so @psss @thrix the implementation is something like?:
```python
class GuestLocal(tmt.Guest):
...
    def scripts_path(self) -> Path:
        return tmt.steps.scripts.effective_scripts_dest_dir(
            default=os.environ.get(TMT_SCRIPTS_DIR, default))
```

happz:
> > From the implementation point of view, leaving the theory to @thrix and @psss, a different directory on localhost would be added by overriding `scripts_path` property in the `local` plugin/guest. See the default implementation which already uses different default when ostree-ish system is used; `local` would simply provide its own implementation, calling `tmt.steps.scripts.effective_scripts_dest_dir` with its own special default.
> 
> Ok, so @psss @thrix the implementation is something like?:
> 
> ```python
> class GuestLocal(tmt.Guest):
> ...
>     def scripts_path(self) -> Path:
>         return tmt.steps.scripts.effective_scripts_dest_dir(
>             default=os.environ.get(TMT_SCRIPTS_DIR, default))
> ```

No, do not read `TMT_SCRIPTS_DIR` envvar. That is not your job here. You job is to provide `effective_scripts_dest_dir()` wit the default you want, leaving the rest to `effective_scripts_dest_dir()`. `effective_scripts_dest_dir(default=MY_VERY_SPECIAL_LOCAL_PATH)`.

tmt/steps/provision/local.py
mcasquer: verbose trace removed !

mcasquer:
@psss @thrix @happz let me know if this looks good, I wasn't sure about which path use so I chose `/var/lib/tmt/scripts` feel free to propose a new one

tmt/steps/provision/local.py
```diff
@@ -27,6 +27,13 @@ class GuestLocal(tmt.Guest):
     localhost = True
     parent: Optional[tmt.steps.Step]
 
+    @property
+    def scripts_path(self) -> Path:
+        """
+        Absolute path to tmt scripts directory
+        """
+        return tmt.steps.scripts.effective_scripts_dest_dir(default=Path("/var/lib/tmt/scripts"))
```
happz: Right, and the path itself should be a module-level constant in this plugin, just like https://github.com/teemtee/tmt/blob/main/tmt/steps/scripts.py#L24 and https://github.com/teemtee/tmt/blob/main/tmt/steps/scripts.py#L27 are defined in `tmt.steps.scripts` for the general usage.

mcasquer: Yeah sorry for that creepy literal, constant added!

mcasquer:
> so I chose `/var/lib/tmt/scripts` feel free to propose a new one

Finally decided to `/var/lib/tmt/helper/scripts` :)

tmt/steps/provision/local.py
```diff
@@ -136,7 +148,30 @@ def execute(
...
-        self.debug("No installation of tmt scripts is needed on localhost.")
+        """Install scripts required by tmt"""
+
+        # Make sure scripts directory exists
+        command = Command("mkdir", "-p", f"{self.scripts_path}")
+
+        if not self.facts.is_superuser:
+            command = Command("sudo") + command
+
+        self.execute(command, silent=True)
```
LecrisUT: Implementation wise, I think it should follow https://github.com/teemtee/tmt/pull/3983

tmt/steps/provision/local.py
```diff
@@ -136,7 +148,30 @@ def execute(
         )
 
     def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
-        self.debug("No installation of tmt scripts is needed on localhost.")
+        """Install scripts required by tmt"""
```
LecrisUT: Somewhere we would need to expand `PATH` to include the calculated `self.scripts_path`. I think it's something like `self.plan.environment`

tmt/steps/provision/local.py
mcasquer: I everybody agrees I will send an update to follow the same logic

tmt/steps/provision/local.py
mcasquer: Could you provide more context? I am not getting your point here :)

tmt/steps/provision/local.py
thrix: @mcasquer I believe this is what we agreed upon yes

tmt/steps/provision/local.py
```diff
@@ -13,6 +13,9 @@
 from tmt.utils.hints import get_hint
 from tmt.utils.wait import Waiting
 
+#: The default helper scripts destination directory
+DEFAULT_HELPER_SCRIPTS_DEST_DIR = Path("/var/lib/tmt/helper/scripts")
```
thrix: @mcasquer I would move this default to the other ones, so they are on the same place maybe? and just call it `DEFAULT_SCRIPTS_DEST_DIR_LOCAL` ?

see tmt/steps/scripts.py

tmt/steps/provision/local.py
```diff
@@ -136,7 +148,30 @@ def execute(
...
+        for script in scripts:
+            if not script.enabled(self):
+                continue
+
+            with script as source:
+                for filename in [script.source_filename, *script.aliases]:
+                    destination = script.destination_path or self.scripts_path / filename
+
+                    # Copy file and set permissions
+                    install_cmd = Command('install', '-p', '-m', '755', source, destination)
```
thrix: there should be assertions somewhere for this

tmt/steps/provision/local.py
mcasquer: So finally do we rename the constant or as it's going to be in local.py the naming doesn't longer matter

happz: Renaming can't hurt, but I would keep it in `local.py`.

tests/provision/local/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "export TMT_SCRIPTS_DIR=/var/tmp/helper/scripts"
```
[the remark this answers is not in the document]
mcasquer: Doesn't apply to this case

tmt/steps/provision/local.py
mcasquer: Added a new test !

mcasquer:
While trying to align this PR with the changes from https://github.com/teemtee/tmt/pull/3983, I've attempted to directly call the parent function, with something like:

```python
def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
        """Install scripts required by tmt"""
        super().install_scripts(scripts)
```

Then, I found the `self.push` is not really "pushing" the files to the `destination`, which in this case is `self.scripts_path / filename`, which in turn it takes the value defined in `TMT_SCRIPTS_DIR` env (the expected behavior).

Perhaps I am misconceiving the idea but the `push` defined in `Guest` class is not implemented
https://github.com/teemtee/tmt/blob/8ce648a73429dbc45ba9725d114567a74c09c1b9/tmt/steps/provision/__init__.py#L1792-L1803

cc: @thrix @LecrisUT

happz:
> While trying to align this PR with the changes from #3983, I've attempted to directly call the parent function, with something like:
> 
> ```python
> def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
>         """Install scripts required by tmt"""
>         super().install_scripts(scripts)
> ```
> 
> Then, I found the `self.push` is not really "pushing" the files to the `destination`, which in this case is `self.scripts_path / filename`, which in turn it takes the value defined in `TMT_SCRIPTS_DIR` env (the expected behavior).
> 
> Perhaps I am misconceiving the idea but the `push` defined in `Guest` class is not implemented

`Guest.push()` is indeed an abstract method, since the implementation depends on the infrastructure. SSH-based guests have their own, localhost or podman need to implement it as well.

I think you are hitting this code in `local.py`:

```python
    def push(
        self,
        source: Optional[Path] = None,
        destination: Optional[Path] = None,
        options: Optional[list[str]] = None,
        superuser: bool = False,
    ) -> None:
        """
        Nothing to be done to push workdir
        """
```

The method is implemented - `self.push()` is really this one, not `Guest.push()` - but does nothing, assuming that whatever tmt might want to "push" to localhost, it's already there, so why bother? That's true, but only for workdir - and you push files to completely different location, and the method ding nothing is behind your observation of "`self.push` is not really "pushing" the files to the `destination`". It's true, it does not push, but it's not fault of `Guest.push()`, `GuestLocal.push()` is to blame.

I think you can use the parent implementation of `install_scripts()`, but you would need to add more code to `GuestLocal.push()` for it to work, e.g. test `source` and `destination` and if they are *not* workdir, then actually do copy/install/push the given content to the given destination.

mcasquer:
> I think you can use the parent implementation of `install_scripts()`, but you would need to add more code to `GuestLocal.push()` for it to work, e.g. test `source` and `destination` and if they are _not_ workdir, then actually do copy/install/push the given content to the given destination.

Indeed, my bad again :cry:, push code updated !

Ps: I took `GuestSsh` as reference so I am using `rsync` which shouldn't be totally necessary? In addition, raising errors and no log traces added to `source`, `destination` and `workdir` checks, feel free to propose a different approach :)

mcasquer:
```
(dev) [root@local local]# rm -rf /var/tmp/helper
(dev) [root@local local]# ./test.sh 

:: [ 05:23:05 ] :: [   PASS   ] :: Command 'export TMT_SCRIPTS_DIR=/var/tmp/helper/scripts' (Expected 0, got 0)
...
:: [ 05:23:09 ] :: [   PASS   ] :: File /var/tmp/helper/scripts/tmt-abort should exist
...
:: [ 05:23:09 ] :: [   PASS   ] :: Check if script is executable (Expected 0, got 0)
...
(dev) [root@local local]# ls /var/tmp/helper/scripts
rhts-abort   rhts-report-result  rhts_submit_log  rstrnt-reboot      rstrnt-report-result  tmt-file-submit  tmt-reboot-core
rhts-reboot  rhts-submit-log     rstrnt-abort     rstrnt-report-log  tmt-abort             tmt-reboot       tmt-report-result
```

tests/provision/local/test.sh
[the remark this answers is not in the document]
mcasquer: Doesn't apply

happz: Isn't it the default path? I believe we spoke about using a temporary directory, to avoid messing with scripts other tmt invocations might be using.

tmt/steps/provision/local.py
happz: `/usr/local/bin` is usually in `$PATH`, `/whatever/special/path/local/plugin/uses` most likely isn't. We install them, but they can be reached only directly, and only if the test is aware of the path picked.

* we need to connect the dots, so to speak, e.g. by adding this scripts path to `$PATH`, possibly in `TEST_OUTER_WRAPPER_TEMPLATE` in `tmt.steps.execute.internal`. That will make `tmt-report-result ...` calls successful.
* we should probably also expose the scripts path, so tests and scripts can reach them directly, e.g. `$TMT_SCRIPTS_DIR/tmt-report-result ...` - and this is already done, `$TMT_SCRIPTS_DIR` is already exposed by tmt.
* and we also need to connect the dots for `prepare` or `finish`, not just for tests - even `prepare/script` may need to call `tmt-report-result ...`, and it won't be able to. But that I think we can leave for later, because I don't think we have anything in place for them yet, and I have a work-in-progress patch enabling `tmt-reboot` in `prepare` and `finish`, so I will be poking the area, possibly resolving this eventually.

happz:
> > I think you can use the parent implementation of `install_scripts()`, but you would need to add more code to `GuestLocal.push()` for it to work, e.g. test `source` and `destination` and if they are _not_ workdir, then actually do copy/install/push the given content to the given destination.
> 
> Indeed, my bad again 😢, push code updated !
> 
> Ps: I took `GuestSsh` as reference so I am using `rsync` which shouldn't be totally necessary? In addition, raising errors and no log traces added to `source`, `destination` and `workdir` checks, feel free to propose a different approach :)

Yeah, `rsync` does not seem to be the best choice. You can check whether source/destination are directories or not, check `tmt.utils.filesystem.copy_tree` which works for copying directory content, and you can also use the `install` or simply `cp` if it's individual files. Which is the case of scripts.

tests/provision/local/test.sh
mcasquer: No, the default path defined in this PR is `/var/lib/tmt/helper/scripts`
If you prefer I can place them directly under `/tmp` so they get cleared after reboot

mcasquer:
> Yeah, `rsync` does not seem to be the best choice. You can check whether source/destination are directories or not, check `tmt.utils.filesystem.copy_tree` which works for copying directory content, and you can also use the `install` or simply `cp` if it's individual files. Which is the case of scripts.

Using install command now !

tests/provision/local/test.sh
happz: I'd recommend using a directory created by `mktemp -d`, it really prevents any possible collision. It's very unlikely, but still possible that two people would run the test and clash over this path... I think it's a good habit to isolate one's work.

mcasquer: Updated !

tmt/steps/provision/local.py
mcasquer: So from my side I should only meet the first point?

> we need to connect the dots, so to speak, e.g. by adding this scripts path to $PATH, possibly in TEST_OUTER_WRAPPER_TEMPLATE in tmt.steps.execute.internal. That will make tmt-report-result ... calls successful.

happz: I think so, yes.

mcasquer: Then, there's something I am not understanding... Indeed the customized path is not being included in the `$PATH` or being accessible to tmt, neither the default one: `/var/lib/tmt/helper/scripts` nor the one defined in `$TMT_SCRIPTS_DIR`. I've tried both locally and scripts placed one place or another are not accessible.

The point is, looking to the `TEST_OUTER_WRAPPER_TEMPLATE` I already see `scripts_path` being included to the `$PATH`, but this `INVOCATION.guest.scripts_path` is not the same we're defining in `local.py` or in `$TMT_SCRIPTS_DIR`?

```
# Make sure guest scripts path is searched by shell
if ! grep -q "{{ INVOCATION.guest.scripts_path }}" <<< "${PATH}"; then
    export PATH={{ INVOCATION.guest.scripts_path }}:${PATH}
fi
```
Sorry if I ask somethign stupid, this is beyond my knowledge... :confused:

happz: > Then, there's something I am not understanding... Indeed the customized path is not being included in the `$PATH` or being accessible to tmt, neither the default one: `/var/lib/tmt/helper/scripts` nor the one defined in `$TMT_SCRIPTS_DIR`. I've tried both locally and scripts placed one place or another are not accessible.
> 
> The point is, looking to the `TEST_OUTER_WRAPPER_TEMPLATE` I already see `scripts_path` being included to the `$PATH`, but this `INVOCATION.guest.scripts_path` is not the same we're defining in `local.py` or in `$TMT_SCRIPTS_DIR`?
> 
> ```
> # Make sure guest scripts path is searched by shell
> if ! grep -q "{{ INVOCATION.guest.scripts_path }}" <<< "${PATH}"; then
>     export PATH={{ INVOCATION.guest.scripts_path }}:${PATH}
> fi
> ```
> 
> Sorry if I ask somethign stupid, this is beyond my knowledge... 😕

Ouch, it's really there, I missed that. It should be the same path, it's the same attribute, `Guest.scripts_path`. So that's one problem solved.

The second issue: you say it doesn't work, i.e. you don't see your custom `local` path being part of `$PATH`? I briefly checked, with `TMT_SCRIPTS_DIR=/tmp/foo`, and it seems to work for me:

```
++ grep -q /tmp/foo
++ export PATH=/tmp/foo:/home/happz/.local/share/hatch/env/virtual/tmt/BuhwJjvK/dev/bin:...
++ PATH=/tmp/foo:/home/happz/.local/share/hatch/env/virtual/tmt/BuhwJjvK/dev/bin:...
```

mcasquer: > The second issue: you say it doesn't work, i.e. you don't see your custom `local` path being part of `$PATH`? I briefly checked, with `TMT_SCRIPTS_DIR=/tmp/foo`, and it seems to work for me:

That's it, IIUC, local `PATH` must get modified along with scripts allocation, I am directly trying by executing the plan of this PR and no luck:
```bash
(dev) [root@local data]# tmt --feeling-safe run provision plan -n script -vvv
...
topology address
scripts_path: /var/lib/tmt/helper/scripts --> A verbose trace I added
multihost name: default-0
...
```
If I export `TMT_SCRIPTS_DIR`:
```bash
export TMT_SCRIPTS_DIR=/var/tmp/helper/scripts/
```
In both cases PATH is not modified:
```bash
(dev) [root@local data]# echo $PATH | grep helper
(dev) [root@local data]# tmt-abort
bash: tmt-abort: command not found
(dev) [root@local data]# echo $TMT_SCRIPTS_DIR
/var/tmp/helper/scripts/
```

happz: What is the context of the last block? What shell session is it?

mcasquer: My own SSH session to a host, so is `$PATH` modification not visible outside the tmt execution? Not persistent?

happz: Correct, it's visible to the test only:

```
if ! grep -q "{{ INVOCATION.guest.scripts_path }}" <<< "${PATH}"; then
    export PATH={{ INVOCATION.guest.scripts_path }}:${PATH}
fi
```

It's not persistent.

mcasquer: Ok, so that part should be covered, then remains to look into GuestLocal.push, which is making all CI jobs to fail, seeing the following:
```
Cause number 1:

    prepare step failed

    Cause number 1:

        The source is not defined.
```
Which indeed is a exception added in this PR, but my understanding is that calling this function means you need a source and a destination... not sure what's happening here :confused:

happz: See the documentation of the method you override in `local`:

```
        :param source: if set, this path will be uploaded to the guest.
            If not set, plan workdir is uploaded.
```

`source` can and often is not set, and it means "plan workdir". Same goes for `destination`:

```
        :param destination: if set, content will be uploaded to this
            path. If not set, root (``/``) is used.
```

So when unset, it's the original no-op action, sync the workdir to guest, and it's already there because the guest is localhost.

mcasquer:
From what I can see, the CI failures are because of scripts installation using `sudo` [1] and some other ones to a bad connection during certificates installation[2]:

[1]
```
Cause number 1:

        Command 'sudo install -p -m 755 /usr/lib/python3.13/site-packages/tmt/steps/scripts/tmt-abort /var/lib/tmt/helper/scripts/tmt-abort' returned 1.

        # stderr (11/11 lines)
        We trust you have received the usual lecture from the local System
        Administrator. It usually boils down to these three things:

            #1) Respect the privacy of others.
            #2) Think before you type.
            #3) With great power comes great responsibility.

        For security reasons, the password you type will not be visible.

        sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
        sudo: a password is required

```
[2]   
`Failed to download metadata for repo 'beaker-client': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried `

tests/provision/local/data/script.fmf
```diff
@@ -0,0 +1,7 @@
+discover:
```
psss: Let's call the file `plan.fmf` as it is a `tmt` plan, not a script, right?

tests/provision/local/data/script.fmf
```diff
@@ -0,0 +1,7 @@
...
+    how: fmf
+    path: .
```
psss: This line is not necessary. Let's drop it.

tests/provision/local/test.sh
```diff
@@ -0,0 +1,28 @@
...
+    rlPhaseStartTest "Test script installation"
+        # Run tmt with local provision
+        rlRun "tmt --feeling-safe run provision plan -n script"
+
+        # Check if the script was installed in the default location
+        script_path="$tmp/tmt-abort"
+        rlAssertExists "$script_path"
+
+        # Check if the script is executable
+        rlRun "test -x $script_path" 0 "Check if script is executable"
```
psss: Would be nice to extend the test to cover the original error reported in the issue #1853 as well:

    bash: /usr/local/bin/tmt-file-submit: No such file or directory

These/similar errors should not appear in the test output anymore.

tmt/steps/provision/local.py
```diff
@@ -9,11 +9,14 @@
...
 import tmt.utils
 from tmt.container import container
-from tmt.steps.provision import TransferOptions
+from tmt.steps.provision import Provision, TransferOptions
 from tmt.utils import Command, OnProcessEndCallback, OnProcessStartCallback, Path, ShellScript
 from tmt.utils.hints import get_hint
 from tmt.utils.wait import Waiting
 
+#: The default helper scripts destination directory
+DEFAULT_HELPER_SCRIPTS_DEST_DIR = Path("/var/lib/tmt/helper/scripts")
```
psss:
```suggestion
DEFAULT_HELPER_SCRIPTS_DEST_DIR = Path("/var/tmp/tmt/scripts")
```

I would suggest to use a path which is writeable by a regular user as the default, to make the default scenario more fluent.

mcasquer: After the long internal discussion... accepting it :)

mcasquer: Code updated !

tests/provision/local/test.sh
mcasquer: Changed to `tmt-file-submit` and added an extra grep assertion that checks the log file

tests/provision/local/data/script.fmf
mcasquer: Removed !

tests/provision/local/data/script.fmf
mcasquer: Fair enough, updated !

tests/provision/local/test.sh
psss: The error would appear during the test execution, but above `tmt run provision` is used which means only the `provision` step is executed. We should run all steps and add a simple `test.fmf` file as well.

mcasquer: @psss so, for example, create a `test.fmf` that would call a `test.sh` and inside the latter use `tmt-file-submit` with a dummy file?

psss: Yes, a very small, minimal `test.sh`, probably with a single `beakerlib` phase. Just to exercise the end of phase reporting which actually calls the helper script. This would be better rather then to call the helper script directly.

tests/provision/local/data/script.fmf
psss: Thanks!

tmt/steps/provision/local.py
psss: Thanks.

psss:
Seems that the following two tests are [failing](https://artifacts.dev.testing-farm.io/a145a11e-9d6e-4186-a135-45e01f927611/):

* `/tests/run/permissions`
* `/tests/run/shell`

Both with this error:

    sudo: a password is required

Please, have a look.

mcasquer:
> Seems that the following two tests are failing:
> 
> * `/tests/run/permissions`
> * `/tests/run/shell`
> 
> Both with this error:
> 
> ```
> sudo: a password is required
> ```
> 
> Please, have a look.

Ongoing :)

psss:
As discussed on the chat, let's store the scripts under the run workdir, for example:

    /var/tmp/tmt/run-001/scripts/tmt-abort
    /var/tmp/tmt/run-001/scripts/tmt-reboot
    ...

This should prevent the permissions problem when multiple users on the same test runner are executing tests.

mcasquer:
> This should prevent the permissions problem when multiple users on the same test runner are executing tests.

So, no `DEFAULT_HELPER_SCRIPTS_DEST_DIR` and no `scripts_path` property definition, just setting `destination` as `workdir/scripts` ?

happz:

tmt/steps/provision/local.py
```diff
@@ -177,9 +190,37 @@ def push(
...
+
+        # no-op case
+        if source_resolved == destination_resolved or workdir_resolved in (
+            source_resolved,
+            destination_resolved,
+        ):
+            return
+
+        # Copy scripts and set permissions
+        install_cmd = Command('install', '-p', '-m', '755', source_resolved, destination_resolved)
```
psss: I'm not sure about this part: For example, what if the `prepare` step is copying local rpm files to be installed to it's directory:

    .../run/plan/prepare/default-0/packages/*

This would set the execute bit for each package which is copied there. Should/could we just preserve the permissions using `cp -a` instead?

mcasquer: Copying should be fine as there are merely scripts... so I can change it :)

tmt/base.py
```diff
@@ -4670,6 +4670,13 @@ def go(self) -> None:
 
         self.show_runner(self._logger)
 
+        # Create scripts directory and copy TMT scripts there
```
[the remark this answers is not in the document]
mcasquer: Usual tmt uppercase suggestion... ignoring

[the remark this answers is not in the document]
mcasquer: Well, in this case it could be applied, my bad :)

mcasquer:
Now scripts are copied under the running workdir:
```
# tmt --feeling-safe run provision plan -vvv
/var/tmp/tmt/run-037
warn: User is feeling safe.

/plan
    provision
...
dev) [root@local data]# ls /var/tmp/tmt/run-037/scripts/
tmt-abort  tmt-file-submit  tmt-reboot  tmt-reboot-core  tmt-report-result  tmt.sh.j2
```

With `TMT_SCRIPTS_DIR` set as env, warning is raised in local provisioning:
```
(dev) [root@local data]# env | grep TMT_SCRIPTS
TMT_SCRIPTS_DIR=/var/tmp/helper
(dev) [root@local data]# tmt --feeling-safe run provision plan -vvv
/var/tmp/tmt/run-036
warn: User is feeling safe.

/plan
    provision
        
        how: local
        order: 50
        primary address: localhost
        warn: TMT_SCRIPTS_DIR is defined for the local provision.
        primary address: localhost
```
@psss @happz @thrix once you confirm the approach I will re-add the test coverage :)
Ps: with `copytree`, are we sure that the aliases (if any) are copied as well?

tmt/base.py
```diff
@@ -4670,6 +4670,13 @@ def go(self) -> None:
 
         self.show_runner(self._logger)
 
+        # Create scripts directory and copy tmt scripts there
+        tmt_parent = Path(tmt.__file__).resolve().parent
+        source = tmt_parent / "steps" / "scripts"
+        destination = self.workdir / "scripts"
+        destination.mkdir(exist_ok=True)
+        shutil.copytree(source, destination, dirs_exist_ok=True)
+
```
psss, line 4711: Let's create a new method which would handle copying scripts to make the `go()` method more readable/maintainable.

tmt/steps/provision/local.py
```diff
@@ -268,6 +284,9 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         if data.hardware and data.hardware.constraint:
             self.warn("The 'local' provision plugin does not support hardware requirements.")
 
+        if os.environ.get("TMT_SCRIPTS_DIR"):
+            self.warn("TMT_SCRIPTS_DIR is defined for the local provision.")
```
psss: Let's be a bit more specific? Perhaps something like:

    The ... variable is not supported, the default scripts path ... is used instead

mcasquer: Warning extended!
`warn: The TMT_SCRIPTS_DIR env is not supported in 'local' provision, the default scripts path /var/tmp/tmt/run-007/scripts is used instead.
`

tmt/base.py
mcasquer: Done !

mcasquer:
Test coverage added
```bash
:: [ 05:22:12 ] :: [   PASS   ] :: Command 'tmt --feeling-safe run --workdir-root /tmp/tmp.pbq1AqSlSG provision plan' (Expected 0, got 0)
:: [ 05:22:12 ] :: [   PASS   ] :: File /tmp/tmp.pbq1AqSlSG/run-001/scripts/tmt-file-submit should exist 
:: [ 05:22:12 ] :: [  BEGIN   ] :: Check if script is executable :: actually running 'test -x /tmp/tmp.pbq1AqSlSG/run-001/scripts/tmt-file-submit'
:: [ 05:22:12 ] :: [   PASS   ] :: Check if script is executable (Expected 0, got 0)
:: [ 05:22:12 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.Rp7P2iB5' should contain 'The TMT_SCRIPTS_DIR env is not supported' 
```
Please @psss @happz @thrix consider moving this PR to **review** status. I still have no permissions to do it :)

mcasquer:
And one more thing, do we need a release note for this? Please let me know :)

tmt/base.py
```diff
@@ -4635,6 +4635,18 @@ def show_runner(self, logger: tmt.log.Logger) -> None:
...
+        """
+        Copy the tmt helper scripts under the running workdir
+        into a new scripts directory.
+        """
+        tmt_parent = Path(tmt.__file__).resolve().parent
+        source = tmt_parent / "steps" / "scripts"
+        assert self.workdir is not None
+        destination = self.workdir / "scripts"
+        destination.mkdir(exist_ok=True)
+        shutil.copytree(source, destination, dirs_exist_ok=True)
```
tcornell-bus: Is `tmt/steps/scripts/tmt.sh.j2` supposed to be copied too? Currently it's getting copied.

tcornell-bus:
> Ps: with copytree, are we sure that the aliases (if any) are copied as well?

By default, it should copy the contents of any symlinks. If you want the symlinks copied as symlinks, I think you need to add `symlinks=True`

tmt/steps/provision/local.py
```diff
@@ -28,6 +29,19 @@ class GuestLocal(tmt.Guest):
...
+        """
+        Absolute path to tmt scripts directory
+        """
+        parent = cast(Provision, self.parent)
+        assert parent.plan.workdir is not None
+
+        workdir_resolved = parent.plan.workdir.resolve()
+        default_path = workdir_resolved / Path("scripts")
+
+        return tmt.steps.scripts.effective_scripts_dest_dir(default=default_path)
```
happz: Both `workdir_resolved` and `default_path` exist to be used just once; `workdir` probably does not need to be resolved, it is not compared to anything; `/` works with `Path` and strings nicely, no need for `Path(...)`. How about this?

```python
return tmt.steps.scripts.effective_scripts_dest_dir(
    default=parent.plan.workdir.absolute() / "scripts"
)
```

tmt/steps/provision/local.py
```diff
@@ -268,6 +284,17 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
 
+        assert isinstance(self.parent, tmt.steps.provision.Provision)
+        assert self.parent.plan.my_run is not None
+        workdir = self.parent.plan.my_run.workdir
+        assert workdir is not None
+
+        if os.environ.get("TMT_SCRIPTS_DIR"):
+            self.warn(
+                "The TMT_SCRIPTS_DIR env is not supported in 'local' provision, "
+                f"the default scripts path {workdir / Path('scripts')} is used instead."
```
happz:
```suggestion
                f"the default scripts path {workdir.absolute() / 'scripts'} is used instead."
```

tmt/base.py
```diff
@@ -4635,6 +4635,18 @@ def show_runner(self, logger: tmt.log.Logger) -> None:
         for _, key_formatted, value_formatted in self.runner.facts.format():
             log(key_formatted, value_formatted, shift=1)
 
+    def copy_scripts(self) -> None:
+        """
+        Copy the tmt helper scripts under the running workdir
+        into a new scripts directory.
+        """
+        tmt_parent = Path(tmt.__file__).resolve().parent
+        source = tmt_parent / "steps" / "scripts"
```
happz: Use `tmt.steps.scripts.SCRIPTS_SRC_DIR` instead, that is the correct way to reach into tmt package.

tmt/base.py
happz: > Is `tmt/steps/scripts/tmt.sh.j2` supposed to be copied too? Currently it's getting copied.

Probably not, but I wouldn't complicate it even more...

@mcasquer use `tmt.utils.filesystem.copy_tree()` instead, you can drop `mkdir()` call as well. `tmt.utils.filesystem.copy_tree(tmt.steps.scripts.SCRIPTS_SRC_DIR, self.workdir / "scripts", self._logger)` could work.

tmt/base.py

tmt/steps/provision/local.py
mcasquer: Updated ! Works nice as well
`warn: The TMT_SCRIPTS_DIR env is not supported in 'local' provision, the default scripts path /var/tmp/tmt/run-012/scripts is used instead.`

tmt/steps/provision/local.py
mcasquer: Updated !

tmt/base.py

mcasquer:
> By default, it should copy the contents of any symlinks. If you want the symlinks copied as symlinks, I think you need to add `symlinks=True`

Yeah I've missed that, thanks ! Anyways now using `tmt.utils.filesystem.copy_tree` should included as mentioned in the docstring: "`Symlinks are always preserved."`

tests/provision/local/test.sh
tcornell-bus: Will this test be extended to address this? :slightly_smiling_face:

psss:
> Opportunistically, raises a exception if `TMT_SCRIPTS_DIR` env is set for local provisioning.

What about rather explaining that using `TMT_SCRIPTS_DIR` with the `local` provision is currently not supported and point to an issue which will track the fix for this?

tests/provision/local/test.sh
psss: Yes, this is needed to verify the original scenarios has been fixed. @mcasquer, could you please add a simple beakerlib test under the `data` directory?

tests/provision/local/test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "tmp=\$(mktemp -d)" 0 "Create tmp directory"
+        rlRun "export TMT_SCRIPTS_DIR=$tmp"
```
psss: Would be good to also exercise scenario with `TMT_SCRIPTS_DIR` not set (no warning).

tmt/base.py
psss: Thanks!

tmt/steps/provision/local.py
```diff
@@ -28,6 +29,18 @@ class GuestLocal(tmt.Guest):
...
+    def scripts_path(self) -> Path:
+        """
+        Absolute path to tmt scripts directory
+        """
+        parent = cast(Provision, self.parent)
+        assert parent.plan.workdir is not None
+
+        return tmt.steps.scripts.effective_scripts_dest_dir(
+            default=parent.plan.workdir.absolute() / "scripts"
+        )
```
psss: This is pointing into the plan workdir:

```
(Pdb) parent.plan.workdir.absolute()
Path('/var/tmp/tmt/run-089/plan')
```

We need to go to the run, one level up: `parent.plan.my_run.workdir.absolute()`

You would find out, if you would actually test it :-)

tmt/steps/provision/local.py
psss: The `PATH` extension should be covered.

tmt/steps/provision/local.py
```diff
@@ -179,6 +193,7 @@ def push(
         """
         Nothing to be done to push workdir
         """
+        # no-op function
```
psss: No need  for any extra comment, docstring say it all.

tmt/steps/provision/local.py
```diff
@@ -268,6 +283,17 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
 
+        assert isinstance(self.parent, tmt.steps.provision.Provision)
+        assert self.parent.plan.my_run is not None
+        workdir = self.parent.plan.my_run.workdir
+        assert workdir is not None
+
+        if os.environ.get("TMT_SCRIPTS_DIR"):
+            self.warn(
+                "The TMT_SCRIPTS_DIR env is not supported in 'local' provision, "
+                f"the default scripts path {workdir.absolute() / 'scripts'} is used instead."
```
psss:
```suggestion
                "The 'TMT_SCRIPTS_DIR' env is not supported in 'local' provision, "
                f"the default scripts path '{workdir.absolute() / 'scripts'}' is used instead."
```

For consistent user messages as documented [here](https://tmt.readthedocs.io/en/stable/contribute.html#introduction).

tmt/steps/provision/local.py
```diff
@@ -176,9 +190,41 @@ def push(
...
+        # no-op case
+        if (
+            source.resolve() == parent.plan.workdir.resolve()
+            and destination.resolve() == Path('/').resolve()
+        ):
+            return
+
+        # For local guest, check source and destination are not workdir
+        if parent.plan.workdir.resolve() in (source.resolve(), destination.resolve()):
+            return
```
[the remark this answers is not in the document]
mcasquer: Outdated, resolving...

tmt/steps/provision/local.py
```diff
@@ -176,9 +190,41 @@ def push(
...
+
+        # Copy scripts and set permissions
+        install_cmd = Command('install', '-p', '-m', '755', source, destination)
```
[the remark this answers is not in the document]
mcasquer: Outdated, resolving...

tmt/steps/provision/local.py
```diff
@@ -176,9 +190,41 @@ def push(
...
+        # Files identical, skipping
+        if source.resolve() == destination.resolve():
+            return
+
+        # no-op case
+        if (
+            source.resolve() == parent.plan.workdir.resolve()
+            and destination.resolve() == Path('/').resolve()
+        ):
+            return
```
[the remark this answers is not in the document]
mcasquer: Outdated, resolving...

mcasquer:
> What about rather explaining that using `TMT_SCRIPTS_DIR` with the `local` provision is currently not supported and point to an issue which will track the fix for this?

Ok, I can point to the issue this PR tries to close, but current message is what [you actually proposed](https://github.com/teemtee/tmt/pull/3978#discussion_r2351348676) 2 days ago :sweat_smile:

psss:
> Ok, I can point to the issue this PR tries to close, but current message is what [you actually proposed](https://github.com/teemtee/tmt/pull/3978#discussion_r2351348676) 2 days ago 😅

Yes, the message is fine and should stay. The suggestion was just to extend the documentation as well.

tmt/steps/provision/local.py
mcasquer: Done!

tmt/steps/provision/local.py
mcasquer: Done !

tmt/steps/provision/local.py
mcasquer: Good catch, sorry for that... damn copy-pasting. Updated now !

tests/provision/local/test.sh
mcasquer: Scenario added

tmt/base.py
```diff
@@ -4635,6 +4636,16 @@ def show_runner(self, logger: tmt.log.Logger) -> None:
...
 
+    def copy_scripts(self) -> None:
+        """
+        Copy the tmt helper scripts under the running workdir
+        into a new scripts directory.
+        """
+        assert self.workdir is not None
+        tmt.utils.filesystem.copy_tree(
+            tmt.steps.scripts.SCRIPTS_SRC_DIR, self.workdir / "scripts", self._logger
+        )
```
psss: Hmmm, I think we need to create the aliases like `rhts-report-result` and `rstrnt-report-result` as well. Perhaps doing something like this?

https://github.com/teemtee/tmt/blob/83654f5fe7c89441f3bad6bea239d85c1038e494/tmt/steps/provision/__init__.py#L1494-L1495

mcasquer: Code updated !

mcasquer:
> Please, add a release note and extend the `ProvisionLocal` docstring with a short mention about the `TMT_SCRIPTS_DIR` variable currently not being supported:

Done !

tests/provision/local/test.sh
mcasquer: Added, please feel free to review it

tests/provision/local/test.sh
tcornell-bus, line 1: This file is missing executable permissions

tmt/base.py
mcasquer:
```bash
# ls /var/tmp/tmt/run-020/scripts/
rhts-abort   rhts-report-result  rhts_submit_log  rstrnt-reboot      rstrnt-report-result  tmt-file-submit  tmt-reboot-core    tmt.sh
rhts-reboot  rhts-submit-log     rstrnt-abort     rstrnt-report-log  tmt-abort             tmt-reboot       tmt-report-result
```

tests/provision/local/data/test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartTest "Test tmt-file-submit"
+        rlRun "echo Hello"
```
tcornell-bus, line 6: I think to test that a helper script is usable, doing this instead would work.
```suggestion
        rlRun "tmt-file-submit XXXX plan.fmf"
```

tests/provision/local/data/test.sh
tcornell-bus, line 1: This file is missing executable permissions

tests/provision/local/test.sh
```diff
@@ -0,0 +1,34 @@
...
+        script_path="$tmp_root/run-001/scripts/tmt-file-submit"
+        rlAssertExists "$script_path"
+
+        # Check if the script is executable
+        rlRun "test -x $script_path" 0 "Check if script is executable"
+
+        # Check warning is raised
+        rlRun "export TMT_SCRIPTS_DIR=$tmp"
+        rlRun -s "tmt --feeling-safe run --workdir-root $tmp_root provision plan"
+        rlAssertGrep "The TMT_SCRIPTS_DIR env is not supported" $rlRun_LOG
```
tcornell-bus:
```suggestion
        rlAssertGrep "The 'TMT_SCRIPTS_DIR' env is not supported" $rlRun_LOG
```

tmt/steps/provision/local.py
```diff
@@ -268,6 +288,17 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
 
+        assert isinstance(self.parent, tmt.steps.provision.Provision)
+        assert self.parent.plan.my_run is not None
+        workdir = self.parent.plan.my_run.workdir
+        assert workdir is not None
+
+        if os.environ.get("TMT_SCRIPTS_DIR"):
+            self.warn(
+                "The 'TMT_SCRIPTS_DIR' env is not supported in 'local' provision, "
+                f"the default scripts path '{workdir.absolute() / 'scripts'}' is used instead."
```
tcornell-bus:
```suggestion
                f"the default scripts path '{workdir.absolute() / 'scripts'}' will be used instead."
```
nit-pick feel free to ignore

tests/provision/local/data/test.sh
mcasquer: Yeah, that would make more sense, the point is the `tmt-file-submit` is only available under `$PATH` during the tmt execution, so that will report a command not found. So @psss proposed to just add a dummy test here... anyways I might misunderstood the intention :)

tests/provision/local/test.sh
mcasquer: Updated, thanks !

tests/provision/local/data/test.sh
mcasquer: Updated, thanks !

tests/provision/local/test.sh
mcasquer: Good catch, updated !

tmt/steps/provision/local.py
mcasquer: Updated ! :)

tmt/steps/provision/local.py
```diff
@@ -268,6 +288,18 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
...
+        assert self.parent.plan.my_run is not None
+        workdir = self.parent.plan.my_run.workdir
+        assert workdir is not None
+
+        if os.environ.get("TMT_SCRIPTS_DIR"):
+            self.warn(
+                "The 'TMT_SCRIPTS_DIR' env is not supported in 'local' provision, "
+                f"the default scripts path '{workdir.absolute() / 'scripts'}'"
+                "will be used instead."
+            )
```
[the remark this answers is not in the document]
mcasquer: Good one

tests/provision/local/data/test.sh
psss: Well, I suggested that the test exercises & verifies the original scenario reported in #1853, which is still not covered. We should run the test in verbose mode and check the output for any `No such file or directory` errors. Please, use `tmt run -vvv` and check the output using `rlAssertGrep`.

Second thing: The helper scripts are called by `beakerlib`, e.g. `rhts-report-result` is called at the end of the `beakerlib` phase. So even with no explicit call to `tmt-file-submit` we get some coverage. You might want to add `rlBundleLog` as mentioned in the original issue to cover that one as well.

tests/provision/local/data/test.fmf
```diff
@@ -0,0 +1,2 @@
+test: ./test.sh
+framework: beakerlib
```
psss:
```suggestion
/beakerlib:
    summary: Run helper scripts through beakerlib
    test: ./test.sh
    framework: beakerlib
/directly:
    summary: Run helper scripts directly
    test: ./directly.sh
```

I think @tcornell-bus has a good idea to exercise the helper scripts directly as well. Let's add one more test `directly.sh` which will call `tmt-report-result` directly, without `beakerlib`.

tests/provision/local/test.sh
psss: Thanks!

tests/provision/local/test.sh
```diff
@@ -0,0 +1,33 @@
...
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "tmp=\$(mktemp -d)" 0 "Create tmp directory"
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test script installation"
+        # Run tmt with local provision
+        rlRun "tmp_root=\$(mktemp -d)"
+        rlRun -s "tmt --feeling-safe run --workdir-root $tmp_root"
```
psss:
```suggestion
        rlRun -s "tmt --feeling-safe run --id $run -vvv"
```

The workdir was needed before because we installed scripts directly there. Now that we put them under run workdir we don't need a dedicated workdir. Let's:

        rlRun "run=\$(mktemp -d)" 0 "Create run directory"

... in the setup phase instead and use it directly.

tests/provision/local/test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+        # Check if the script was copied to the default location
+        script_path="$tmp_root/run-001/scripts/tmt-file-submit"
```
psss:
```suggestion
        script_path="$run/scripts/tmt-file-submit"
```

psss (changes requested):

tmt/base.py
```diff
@@ -59,8 +59,10 @@
 import tmt.steps.prepare
 import tmt.steps.provision
 import tmt.steps.report
+import tmt.steps.scripts
 import tmt.templates
 import tmt.utils
+import tmt.utils.filesystem
```
psss: I guess this should be dropped now?

tmt/base.py
```diff
@@ -4658,6 +4660,21 @@ def show_runner(self, logger: tmt.log.Logger) -> None:
...
+        """
+        Copy the tmt helper scripts under the running workdir
+        into a new scripts directory.
+        """
+        assert self.workdir is not None
+        scripts = tmt.steps.scripts.SCRIPTS
+        destination = self.workdir / "scripts"
+        destination.mkdir(exist_ok=True)
+
+        for script in scripts:
```
psss:
```suggestion
        for script in tmt.steps.scripts.SCRIPTS:
```

The variable is used just once, no need for the extra line.

tmt/base.py
mcasquer: Removed ! :)

tests/provision/local/test.sh
mcasquer: Updated ! :)

tests/provision/local/test.sh
mcasquer: Updated ! :)

tmt/base.py
```diff
@@ -4658,6 +4659,21 @@ def show_runner(self, logger: tmt.log.Logger) -> None:
...
             log(key_formatted, value_formatted, shift=1)
 
+    def copy_scripts(self) -> None:
+        """
+        Copy the tmt helper scripts under the running workdir
+        into a new scripts directory.
+        """
+        assert self.workdir is not None
+        scripts = tmt.steps.scripts.SCRIPTS
+        destination = self.workdir / "scripts"
```
psss:
```suggestion
        destination = self.workdir / tmt.steps.scripts.SCRIPTS_DIR_NAME
```

The `scripts` constant is used across multiple places, let's define it once, in the `scripts` module and use it here and in the `local.py` as well.

tmt/steps/provision/local.py
```diff
@@ -137,7 +151,8 @@ def execute(
         )
 
     def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
-        self.debug("No installation of tmt scripts is needed on localhost.")
+        """Install scripts required by tmt"""
+        # no-op function
```
psss: Let's keep the previous debug message. It's still relevant, we do no install scripts on the local guest.

tmt/steps/provision/local.py
```diff
@@ -240,6 +255,11 @@ class ProvisionLocal(tmt.steps.provision.ProvisionPlugin[ProvisionLocalData]):
     .. note::
 
         Neither hard nor soft reboot is supported.
+
+    .. note::
+        The ``TMT_SCRIPTS_DIR`` variable is not supported in the
+        ``local`` provision. This solution has been provided as part
+        of issue #1853.
```
psss: We need a new issue for that as this will be closed with the pull request.
Filed here: https://github.com/teemtee/tmt/issues/4081

psss (changes requested):
Couple more comments.

tmt/steps/provision/local.py
mcasquer: Updated !

tmt/steps/provision/local.py
mcasquer: Recovered !

tmt/base.py

tmt/base.py

tests/provision/local/data/test.fmf
```diff
@@ -0,0 +1,7 @@
+/beakerlib:
+    summary: Run helper scripts through beakerlib
+    test: ./test.sh
+    framework: beakerlib
+/directly:
+    summary: Run helper scripts directly
+    test: ./directly.sh
```
tcornell-bus, line 7: woops, looks like this file is missing from the PR

psss: Oh! Should be fixed by 6f1164f7.

tmt/steps/provision/local.py
```diff
@@ -28,6 +29,19 @@ class GuestLocal(tmt.Guest):
     localhost = True
     parent: Optional[tmt.steps.Step]
 
+    @property
+    def scripts_path(self) -> Path:
+        """
+        Absolute path to tmt scripts directory
```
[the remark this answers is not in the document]
psss: What??? It **is** lower case.

tests/provision/local/data/test.fmf
psss: Covered in 6b98484d.

tmt/steps/provision/local.py
psss: I've updated to use `self._guest.scripts_path` instead, less duplication.

tmt/base.py
psss: This part of the code has been completely replaced.

psss (approved):

tmt/steps/provision/local.py
```diff
@@ -28,6 +29,17 @@ class GuestLocal(tmt.Guest):
...
+        """
+
+        # TODO: For now we ignore the 'TMT_SCRIPTS_DIR' variable.
+        # To be fixed in https://github.com/teemtee/tmt/issues/4081
+        parent = cast(Provision, self.parent)
```
happz: It's generally safer to use `assert isinstance()` instead of `cast()`. Both provide hint for a type checker, but `isinstance()` will fail in runtime if the assumption is incorrect. With `cast()`, you are basically saying `self.parent` is `Provision`, and it can never ever change, it's like a law of physics. If you are wrong, nothing will let you know you were wrong. On the other hand, if you were wrong about this "it's always `Provision` statement", `isinstance()` will raise an error, and it will let you know what the Universe thinks about your absolute statements.

psss: Understood, makes sense, thanks for the clarification, improved in 94113b7e.

PR #3979, opened 2025-08-21
Rewrite remote url for plans as well

lukaszachy:
Allows injection of git credentials

Pull Request Checklist

* [X] implement the feature

### DIFF
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -3124,6 +3124,9 @@ class Plan(
         Discover plan-like nodes matching the given reference in fmf cache.
         """
 
+        # TODO: similar situation as in _resolve_import_from_git
+        assert reference.url is not None
+
         if str(reference.ref).startswith('@'):
             self.debug(
                 f"Not enough data to evaluate dynamic ref '{reference.ref}', "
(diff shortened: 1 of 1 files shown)


tmt/base.py
```diff
@@ -3094,10 +3094,11 @@ def _resolve_import_from_fmf_cache(self, reference: RemotePlanReference) -> Iter
 
                 reference.resolve_dynamic_ref(tmpdirname, self)
 
+        assert reference.url
```
happz: Please, move the assert at the beginning of the method - see `_resolve_import_from_git()` and line 3040, it's the exact same case, narrowing the type we know is no longer `None` but linter can't spot it.

lukaszachy: done - df49d75e

happz:
@lukaszachy do you plan to add test coverage?

happz:
@lukaszachy ping - do you plan to add test coverage?

lukaszachy:
> @lukaszachy ping - do you plan to add test coverage?

I was thinking about it but it for a proper e2e it would require working gitlab/github token. Do we have one for CI / want to use it? 
Frankly I'm not sure whether it is worth it.

happz:
Rawhide issues being tackled in https://github.com/teemtee/tmt/pull/4017

PR #3983, opened 2025-08-26
Check if scripts directory exists and do nothing if it does

### LINKED ISSUE #3949: Do not create `/usr/local/bin` if it already exists

AthreyVinay:
Fixes #3949 
* [X] implement the feature


tmt/steps/provision/__init__.py
```diff
@@ -1394,13 +1394,15 @@ def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
         Install scripts required by tmt.
         """
 
-        # Make sure scripts directory exists
-        command = Command("mkdir", "-p", f"{self.scripts_path}")
+        # Check for scripts directory existence before creation
+        if not self.scripts_path.exists():
```
happz: This is going to tell you whether the path exists _on the runner_. Which it most likely will. The test must happen on the guest, e.g.

```python
output = guest.execute(ShellScript(f'[ -d {self.scripts_path} ] && echo "foo" || echo "bar"').to_shell_command())
# inspect output.stdout, and decide
```

AthreyVinay: Yes, my bad, it is indeed checking on the runner host and not on the guest where the script lands. Thanks.
Have fixed it here: [78bb77c](https://github.com/teemtee/tmt/pull/3983/commits/78bb77c44a52a2afe920867dd7cda762841a237c)

happz: And how about this step of the evolution :)

```python
self.execute(
    ShellScript(
        f'[ ! -d {self.scripts_path} ] && {"sudo" if not self.facts.is_superuser else ""} mkdir -p {self.scripts_path}'
    ).to_shell_command(),
    silent=True
)
```

Feel free to reformat, but I think we can squeeze the check and followup action into one script & one invocation.

AthreyVinay: I 100% thought of this - but  I’ wanted to keep the check and the command separate for readability. If still not convinced, I can refactor :)

happz: 1 remote command instead of 2? :)

AthreyVinay: > I’ wanted to keep the check and the command separate for readability.

Unfortunately yes due to the aforementioned reason.

happz: The combined shell script is not complicated, it will remain fairly readable. It's just a test and an action.

AthreyVinay: Refactored here: [033ef7a](https://github.com/teemtee/tmt/pull/3983/commits/033ef7a1834e1459822281445b787a5f194b3a64)

tmt/steps/provision/__init__.py
```diff
@@ -1394,13 +1394,15 @@ def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
...
-        if not self.facts.is_superuser:
-            command = Command("sudo") + command
-
-        self.execute(command, silent=True)
+        # Ensure scripts directory exists on guest (create only if missing)
+        self.execute(
+            ShellScript(
+                f"[ ! -d {self.scripts_path} ] && "
+                f'{"sudo" if not self.facts.is_superuser else ""} '
+                f"mkdir -p {self.scripts_path}"
```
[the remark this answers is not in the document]
LecrisUT: The splitting of the lines and the inconsistency of the quotation marks is a bit jarring, but not a blocking issue. The `ShellScript` should allow for proper multi-lined input, and you could move the `sudo` string construction outside.

LecrisUT (approved):

tmt/steps/provision/__init__.py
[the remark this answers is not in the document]
mcasquer: If this is true we have a problem with this approach, otherwise this code LGTM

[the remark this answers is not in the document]
LecrisUT: About the graphite app, not really, it is a `Path` property and I don't see that it is exposed for the user to edit.

Edit: Ah, no it is exposed. But still the issue is existing, and it seems too much workaround to satisfy it

tmt/steps/provision/__init__.py
```diff
@@ -1394,13 +1394,15 @@ def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
...
-        command = Command("mkdir", "-p", f"{self.scripts_path}")
-
-        if not self.facts.is_superuser:
-            command = Command("sudo") + command
-
-        self.execute(command, silent=True)
+        # Ensure scripts directory exists on guest (create only if missing)
+        self.execute(
+            ShellScript(
+                f"[ ! -d {self.scripts_path} ] && "
```
therazix: Tests are failing because when the directory already exists, the directory check returns false, which causes the whole command to be treated as a failure. This should fix the tests:
```suggestion
                f"[ -d {self.scripts_path} ] || "
```

tmt/steps/provision/__init__.py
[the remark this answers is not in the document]
psss: This is executed on the guest, so probably not a bit deal. But, `shlex.quote` should not hurt? Plus it would cover for example spaces in the path, e.g. `TMT_SCRIPTS_DIR=/my/super cool/directory`. Don't know who would do that but still... just in case, hm?

tmt/steps/provision/__init__.py
```diff
@@ -1394,13 +1394,15 @@ def install_scripts(self, scripts: Sequence[tmt.steps.scripts.Script]) -> None:
...
-
-        if not self.facts.is_superuser:
-            command = Command("sudo") + command
-
-        self.execute(command, silent=True)
+        # Ensure scripts directory exists on guest (create only if missing)
+        self.execute(
+            ShellScript(
+                f"[ -d {self.scripts_path} ] || "
+                f'{"sudo" if not self.facts.is_superuser else ""} '
```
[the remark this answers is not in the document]
psss: That one extra space does not hurt that much. This sounds also ok.

tmt/steps/provision/__init__.py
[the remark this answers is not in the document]
thrix: +1

vaibhavdaren (approved):

thrix:

* for Rawhide and F43 see https://issues.redhat.com/browse/TFT-3621

PR #3987, opened 2025-08-27
Drop `fedora-rawhide` from package manager testing

psss:
We're repeatedly running into various failures caused by rawhide instabilities, this often blocks pull request merging. It's not clear the benefit overweights the effort. Let's drop rawhide from the list of container images used for package manager testing. We can always revert if this does not prove to be a better approach.


happz:
It's dropped from many other places, the test images are removed completely, and `prepare/install` and `prepare/feature` tests are also affected. I'd say the subject and description doesn't cover the full scope of the change.

psss:
Title updated.

psss:
Hmmm, we've completely forgotten about this one. I remember that at that time it was a major annoyance but since then, it seems, things around rawhide images were not painful. So... shall we drop it? Or let's simplify this anyway? @happz, @thrix, @LecrisUT, what do you say?

LecrisUT:
Imo we should keep some basic testings of the fedora-rawhide. As for simplification, we could split the container/vm plan based on the image that they test :thinking:

happz:
Some degree of testing should be performed. We should dedicate a couple of days just for dissecting and reworking the testsuite, what it does, where, and why...

psss:
More stable recently, let's keep it.

Issue #3791, opened 2025-06-03
tmt attempts to pull/rebuild the container on RHEL AI 1.5 instances while awk/flock are present


frantisekz:
I am trying to run tmt against remote Azure VM with RHEL AI 1.5 deployed (bootc based system on RHEL 9.4), and it seems to be stuck on "Waiting for sysroot lock..."

Host OS: tmt version: 1.49.0 , Fedora 41

Plan:
```
summary: Basic smoke test
execute:
    how: tmt
    script: bash --version
provision:
    how: connect
    guest: 20.245.170.158
    user: azureuser
    become: true
    key: /home/fanys/.ssh/rhel_ai_azure_west
```

Relevant log:

```
/plans/example
summary: Basic smoke test
info
environment:
    TMT_PLAN_DATA: /var/tmp/tmt/run-010/plans/example/data
    TMT_PLAN_ENVIRONMENT_FILE: /var/tmp/tmt/run-010/plans/example/data/variables.env
    TMT_TREE: /var/tmp/tmt/run-010/plans/example/tree
    TMT_VERSION: 1.49.0
context: {}
wake
discover
    Workdir '/var/tmp/tmt/run-010/plans/example/discover' created.
    Read file '/var/tmp/tmt/run-010/plans/example/discover/step.yaml'.
    Step data not found.
    Read file '/var/tmp/tmt/run-010/plans/example/discover/tests.yaml'.
    Discovered tests not found.
    ExecutePlugin.delegate(step=execute, data=None, raw_data={'how': 'tmt', 'script': 'bash --version', 'name': 'default-0'})
    Using the 'ExecuteInternal' plugin for the 'tmt' method.
    ExecuteInternalData: original raw data: {'how': 'tmt', 'script': 'bash --version', 'name': 'default-0'}
    DiscoverPlugin.delegate(step=discover, data=None, raw_data={'name': 'default-0', 'how': 'shell'})
    Using the 'DiscoverShell' plugin for the 'shell' method.
    DiscoverShellData: original raw data: {'name': 'default-0', 'how': 'shell'}
[...]

PR #3991, opened 2025-08-27
Apply `sudo` for reboot command when `become` set

thrix:
While working on #3791 I was connecting to host via the `connect` provisioner and I found out that the `reboot` fails because `sudo` is missing when running it.

Add tests to cover the fix.

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -2151,7 +2151,7 @@ class Reboot(Action):
             default=None,
             help=f"""
             A command to run on the guest to trigger the reboot.
-            Default is '{DEFAULT_REBOOT_COMMAND}'.
+            Default is ``{DEFAULT_REBOOT_COMMAND}``.
             """,
         )
         def reboot(context: 'tmt.cli.Context', **kwargs: Any) -> None:
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -2147,7 +2147,7 @@ class Reboot(Action):
         )
         @option(
             '--command',
-            type=Optional[str],
+            type=str,
             default=None,
             help=f"""
             A command to run on the guest to trigger the reboot.
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -2147,9 +2147,12 @@ class Reboot(Action):
         )
         @option(
             '--command',
-            type=str,
-            default=str(DEFAULT_REBOOT_COMMAND),
-            help='A command to run on the guest to trigger the reboot.',
+            type=Optional[str],
+            default=None,
+            help=f"""
+            A command to run on the guest to trigger the reboot.
+            Default is '{DEFAULT_REBOOT_COMMAND}'.
+            """,
         )
         def reboot(context: 'tmt.cli.Context', **kwargs: Any) -> None:
             """
@@ -2184,7 +2187,12 @@ class Reboot(Action):
         assert hasattr(self.parent, 'plan')
         assert self.parent.plan is not None
         for guest in self.parent.plan.provision.ready_guests:
-            guest.reboot(hard=self.opt('hard'), command=tmt.utils.ShellScript(self.opt('command')))
+            guest.reboot(
+                hard=self.opt('hard'),
+                command=tmt.utils.ShellScript(self.opt('command'))
+                if self.opt('command')
+                else None,
+            )
         self.info('reboot', 'Reboot finished', color='yellow')
 
 
--- a/tmt/steps/provision/connect.py
+++ b/tmt/steps/provision/connect.py
@@ -132,14 +132,9 @@ class GuestConnect(tmt.steps.provision.GuestSsh):
             )
 
         if command is not None:
-            reboot_command = command
-
-            if str(command) == str(tmt.steps.DEFAULT_REBOOT_COMMAND) and self.become:
-                reboot_command = ShellScript(f'sudo {reboot_command}')
-
             return super().reboot(
                 hard=False,
-                command=reboot_command,
+                command=command,
                 waiting=waiting,
             )
 
--- a/tmt/steps/provision/connect.py
+++ b/tmt/steps/provision/connect.py
@@ -132,19 +132,30 @@ class GuestConnect(tmt.steps.provision.GuestSsh):
             )
 
         if command is not None:
+            reboot_command = command
+
+            if str(command) == str(tmt.steps.DEFAULT_REBOOT_COMMAND) and self.become:
+                reboot_command = ShellScript(f'sudo {reboot_command}')
+
             return super().reboot(
                 hard=False,
-                command=command,
+                command=reboot_command,
                 waiting=waiting,
             )
 
         if self.soft_reboot is not None:
-            self.debug(f"Soft reboot using the soft reboot command '{self.soft_reboot}'.")
+            reboot_command = self.soft_reboot
+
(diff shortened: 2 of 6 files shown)


tmt/steps/provision/connect.py
```diff
@@ -132,19 +132,30 @@ def reboot(
...
             return super().reboot(
                 hard=False,
-                command=command,
+                command=reboot_command,
                 waiting=waiting,
             )
 
         if self.soft_reboot is not None:
-            self.debug(f"Soft reboot using the soft reboot command '{self.soft_reboot}'.")
+            reboot_command = self.soft_reboot
```
happz: This does not look correct to me: user provided a reboot command, and tmt should blindly use it. If it fails without `sudo`, it fails and reboot would not happen, but that's user fault. tmt should take the provided command and execute it. It may be a custom script containing complex command, a script already whitelisted with sudoers, for example.

tmt/steps/provision/connect.py
```diff
@@ -132,19 +132,30 @@ def reboot(
             )
 
         if command is not None:
+            reboot_command = command
```
happz: Isn't this branch handler by `super().reboot()` already? Passing `command` as it is would be intercepted there, and modified if necessary.

tmt/steps/provision/connect.py
thrix: yeah, I believe this came from some experiments, will roll this back

tmt/steps/provision/connect.py
thrix: I will check on it

tmt/steps/provision/connect.py
```diff
@@ -132,19 +132,30 @@ def reboot(
...
                 waiting=waiting,
             )
 
         if self.soft_reboot is not None:
-            self.debug(f"Soft reboot using the soft reboot command '{self.soft_reboot}'.")
+            reboot_command = self.soft_reboot
+
+            # Handle default reboot command with becom
+            if command == tmt.steps.DEFAULT_REBOOT_COMMAND and self.become:
+                reboot_command = ShellScript(f'sudo {reboot_command.to_shell_command()}')
```
skycastlelily: I guess there is no need to check command, if command is not none,  it will return on line 140,
I guess the comment should be moved before " if str(command) == str(tmt.steps.DEFAULT_REBOOT_COMMAND) and self.become", plus there is no typo in the comment:s/ becom /become

tmt/steps/provision/connect.py
thrix: indeed, in the case of my reproducer the `self.soft_reboot` is not set. Reverted

tmt/steps/provision/connect.py
thrix: @skycastlelily thanks, for the comment, I reverted this part of the code per the comment from @happz,

tmt/steps/provision/connect.py
thrix: @happz it is not handled in this case. The `sudo` is added only if no `command` is passed, but in this case `command` was passed. I am assuming that `command` comes from the `reboot` run command:

```
  0         @option(                                                                                                                                                                           
  1             '--command',                                                                                                                                                                   
  2             type=str,                                                                                                                                                                      
  3             default=str(DEFAULT_REBOOT_COMMAND),                                                                                                                                           
  4             help='A command to run on the guest to trigger the reboot.',                                                                                                                   
  5         )    
```

We have 3 options:

* do not set in this option the to the default reboot command, leave it to `None` and then default reboot command will be used and `sudo` added
* fix the code in the parent to handle the case where the command if it is the same as `DEFAULT_REBOOT_COMMAND`
* leave it as I have it

happz: I'm not sure this has a nice solution. If I run tmt with `reboot --command reboot`, which is, by accident, the default command, tmt will add `sudo` to it. Which is not what I told it to do, I asked tmt to run `reboot`, not `sudo reboot`. As long as there is a comparison against the default value, the default value itself is affected and possibly spoiled by the action.

Which points towards the first point on your list, leave `default` unset here, and apply the default (ith optional `sudo`) only when command is not set in guest reboot internals. We can update the help text here easily, adding e.g. "If not set, ``{DEFAULT_REBOOT_COMMAND}`` is used." Would it work?

thrix: @happz yep, seems the cleanest solution, will update today

thrix: @happz please recheck, it is `None` now and `DEFAULT_REBOOT_COMMAND` is used. Also this part of the code was reverted, as it does not hit the path again, only if the user forces the `command`.

tmt/steps/provision/connect.py
```diff
@@ -132,15 +132,18 @@ def reboot(
             )
 
         if command is not None:
+            reboot_command = command
+
+            if str(command) == str(tmt.steps.DEFAULT_REBOOT_COMMAND) and self.become:
+                reboot_command = ShellScript(f'sudo {reboot_command}')
```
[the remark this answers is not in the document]
thrix: outdated!

tmt/steps/provision/connect.py
```diff
@@ -139,8 +139,6 @@ def reboot(
             )
 
         if self.soft_reboot is not None:
-            self.debug(f"Soft reboot using the soft reboot command '{self.soft_reboot}'.")
```
happz: Why this change?

thrix: chm, I guess I messed up something, sry

thrix: Added back, removed by an error.

tmt/steps/__init__.py
```diff
@@ -2147,9 +2147,12 @@ def command(
         )
         @option(
             '--command',
-            type=str,
-            default=str(DEFAULT_REBOOT_COMMAND),
-            help='A command to run on the guest to trigger the reboot.',
+            type=Optional[str],
```
happz: `type` as understood by Click is a callable accepting a single argument. I believe `type=str` should remain - if the option is set, it would be a string, as compared to e.g. `type=int` or `type=float`.

thrix: trying

thrix: fixed

tmt/steps/__init__.py
```diff
@@ -2148,8 +2148,11 @@ def command(
         @option(
             '--command',
             type=str,
-            default=str(DEFAULT_REBOOT_COMMAND),
-            help='A command to run on the guest to trigger the reboot.',
+            default=None,
+            help=f"""
+            A command to run on the guest to trigger the reboot.
+            Default is '{DEFAULT_REBOOT_COMMAND}'.
```
happz:
```suggestion
            Default is ``{DEFAULT_REBOOT_COMMAND}``.
```

thrix: yeah, damn it, thanks, accepted

thrix: @happz btw there is in other part of the code:

```
       @option(
            '-c',
            '--command',
            metavar='COMMAND',
            multiple=True,
            default=[DEFAULT_LOGIN_COMMAND],
            help="Run given command(s). Default is 'bash'.",
        )
```

I took it from there :)

happz (approved):

psss (approved):

Issue #3089, opened 2024-07-17
tmt run failed  if no "keyname" is specified for artemis 


skycastlelily:
(dev) lnie@fedora:~/tmt$ tmt run plans --name plans/artemis$
/var/tmp/tmt/run-001
*****

            Connection: close
            content-length: '37'
            
            message: No such SSH key exists
    finish
        warn: Nothing to finish, no guests provisioned.

skycastlelily:
We should change DEFAULT_KEYNAME from 'default' to 'master-key', which  is the default key name for artemis server,and without it artemis server won't be able to start,so users won't see the error above forever:)

psss:
Could you please provide the example plan to reproduce this?

skycastlelily:
sure, here is the plan

```
summary: Basic smoke test
provision:
    how: artemis
    # Specify the Artemis URL where the service is running.
    # Here is an example of a local Artemis instance
    api-url: "http://127.0.0.1:8001/v0.0.58"
    api-version: "0.0.58"
    image: fedora-42
    arch: x86_64
    pool: beaker
    #keyname: master-key
```

PR #3999, opened 2025-08-28
Change `DEFAULT_KEYNAME` for `artemis` plugin

skycastlelily:
We should change it to 'master-key', which is the default key name for artemis server, without a right key name provided, artemis server will refuse provisioning requests.

Fixes: #3089

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


happz:
"artemis server won't be able to start." - Artemis server will be doing just fine ;) Instead, Artemis will fail to connect to the freshly provisioned guest, eventually failing to complete the provisioning.

tmt/steps/provision/artemis.py
```diff
@@ -82,7 +82,7 @@
 DEFAULT_API_URL = 'http://127.0.0.1:8001'
 DEFAULT_ARCH = 'x86_64'
 DEFAULT_PRIORITY_GROUP = 'default-priority'
-DEFAULT_KEYNAME = 'default'
+DEFAULT_KEYNAME = 'master-key'
```
happz, line 86: I would add a comment that this value is adopted from Artemis project so we'd be aligned with it.

skycastlelily: Addressed : https://github.com/teemtee/tmt/commit/57f24c24aa6c2ac45e7f366e838b7e4c04b2643a ^^

happz (approved):

therazix (approved):

happz:
Fedora 43 instability in TF, the rest is green. Merging.

PR #4005, opened 2025-08-28
Exclude inactive test cases from Polarion Test Case search

KwisatzHaderach:
We don't want to search among inactive test cases especially for Test Runs. If the test cases are moved between projects this can sometimes lead to having duplicate test cases and common practice is disabling the ones in the old project. Until now tmt was preferring returning any Test Case matching so excluding old cases meant removing a lot of data from those old ones.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/export/polarion.py
+++ b/tmt/export/polarion.py
@@ -49,20 +49,22 @@ def import_polarion() -> None:
 def get_polarion_ids(
     query_result: list[Any],
     preferred_project: Optional[str] = None,
-) -> tuple[str, Optional[str]]:
+) -> tuple[Optional[str], Optional[str]]:
     """
     Return case and project ids from query results
     """
 
     if not query_result:
-        return 'None', None
-    if len(query_result) == 1:
+        return None, None
+    if len(query_result) == 1 and query_result[0].status != 'inactive':
         return query_result[0].work_item_id, query_result[0].project_id
 
     if preferred_project:
         try:
             return next(
-                item.work_item_id for item in query_result if item.project_id == preferred_project
+                item.work_item_id
+                for item in query_result
+                if item.project_id == preferred_project and item.status != 'inactive'
             ), preferred_project
         except StopIteration:
             pass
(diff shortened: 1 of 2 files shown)


happz:
@KwisatzHaderach was this PR cleared for 1.57 milestone during the today's planning meeting?

KwisatzHaderach:
Guess not, not following the latest (latest being last 6 months) changes in the processes. This is not critical though.

happz:
> Guess not, not following the latest (latest being last 6 months) changes in the processes. This is not critical though.

Understood, no problem. We kind of abandoned the practice of adding PRs to milestones outside of planning meetings, module warranted exceptions. But we try to flush the ready-for-review buffer anyway, so it may easily end up in 1.57 eventually.

tmt/export/polarion.py
```diff
@@ -56,24 +56,26 @@ def get_polarion_ids(
...
 
     for result in query_result:
         # If multiple cases are found prefer cases from other projects
         # than these legacy ones
-        if str(result.project_id) not in LEGACY_POLARION_PROJECTS:
+        if str(result.project_id) not in LEGACY_POLARION_PROJECTS and result.status != 'inactive':
             return result.work_item_id, result.project_id
 
-    return query_result[0].work_item_id, query_result[0].project_id
+    return 'None', None
```
happz: Does `'None'`, a string `'None'`, have any special meaning for Polarion or the import? If it's used here just to signal "no case ID", then I'd strongly recommend making it an actual `None`, plus changing the return type to `Optional[str]` instead of `str`. Encoding "nothingness" into strings makes it terribly hard for linters to follow, it's easy to use `'None'` where a meaningful string is expected - `None` would be reported by a type checker, `'None'` wouldn't.

KwisatzHaderach: yeah, it's a name that "surely" wouldn't match anything and makes the code simpler (although it might look weird), I could add another if there instead
since the "fun" part of getting a Polarion Test Case None is it returns something
```
>>> tt = TestCase(work_item_id=None, project_id=None)
>>> tt.project_id
PROJECT
>>> tt.work_item_id
>>> tt.title
>>>
```

happz (approved):

tcornell-bus:
It seems like this warrants a release note. Otherwise LGTM

psss:
@KwisatzHaderach, seems the test data will need update as well. Could you please have a look?

KwisatzHaderach:
Just mentioned the change for those https://github.com/teemtee/tests/pull/12, but on one of the PRs we need to merge without passing pipeline since it's impossible to use a branch ad-hoc. Should be safe to merge the change on tests first.

psss:
Great, thanks for the quick update! Let's just make sure the tests will be green before we do the merge: bb8ad5dc

psss:
Hmmm, seems to be still [failing](https://artifacts.dev.testing-farm.io/c7f962d2-5d80-4a04-8da0-e7eac75bb708/) with:

    >       self.assertIn(ID_KEY, fmf_node.data)
    E       AssertionError: 'id' not found in {'test': '/bin/true', 'summary': 'This is new testcase inside polarion', 'description': 'Please remove this case before data regeneration'}

psss:
Good, now [remote polarion](https://artifacts.dev.testing-farm.io/87bcf3cf-e93f-4f7c-b3fa-7a6c587116ff/#work-polarion2ktvy42x_plans-remote-polarion) plan is green. Going to merge the tests pull request.

psss:
Dropped the temporary commit and rebased on the latest `main`. Should now be green and ready to merge.

psss:
The only failure is a network connection glitch:

    Failure downloading https://dl.fedoraproject.org/pub/epel/epel-next-release-latest-9.noarch.rpm, Request failed: <urlopen error _ssl.c:1130: The handshake operation timed out>

Everything else green, merging.

psss (approved):

PR #4008, opened 2025-08-28
Drop snapshot "support" from Artemis provision plugin

happz:
It was not completed anyway. The feature is being removed from Artemis.

Pull Request Checklist

* [x] implement the feature


happz:
Test results are good enough, no plugin touched but `artemis` which does not have any real tests in our checks.

PR #4014, opened 2025-09-02
Skip `chcon` if selinux not used by host

tweksteen:
If the host does not have SELinux enabled, skip the chcon call on the console log file.

### DIFF
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -1473,9 +1473,10 @@ class ConsoleLog(tmt.steps.provision.GuestLog):
         logger.debug(f"Created console log directory '{self.exchange_directory}'.", level=3)
 
         self.exchange_directory.chmod(0o755)
-        self.guest._run_guest_command(
-            Command("chcon", "--type", "virt_log_t", self.exchange_directory), silent=True
-        )
+        if self.guest.parent.plan.my_run.runner.facts.has_selinux:
+            self.guest._run_guest_command(
+                Command("chcon", "--type", "virt_log_t", self.exchange_directory), silent=True
+            )
 
     def cleanup(self, logger: tmt.log.Logger) -> None:
         """
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -1473,6 +1473,10 @@ class ConsoleLog(tmt.steps.provision.GuestLog):
         logger.debug(f"Created console log directory '{self.exchange_directory}'.", level=3)
 
         self.exchange_directory.chmod(0o755)
+
+        assert isinstance(self.guest.parent, tmt.steps.provision.Provision)  # narrow type
+        assert self.guest.parent.plan.my_run is not None  # narrow type
+
         if self.guest.parent.plan.my_run.runner.facts.has_selinux:
             self.guest._run_guest_command(
                 Command("chcon", "--type", "virt_log_t", self.exchange_directory), silent=True


tmt/steps/provision/testcloud.py
```diff
@@ -1473,9 +1473,14 @@ def prepare(self, logger: tmt.log.Logger) -> None:
...
 
         self.exchange_directory.chmod(0o755)
-        self.guest._run_guest_command(
-            Command("chcon", "--type", "virt_log_t", self.exchange_directory), silent=True
-        )
+
+        output = self.guest._run_guest_command(Command('cat', '/proc/filesystems'), silent=True)
+        if output is None or output.stdout is None:
+            return None
+        if 'selinux' in output.stdout:
```
happz: tmt should already know whether selinux is available, see & test `self.guest.facts.has_selinux`.

tweksteen: From my understanding, here we are checking if the *host* has SELinux or not; while `self.guest.facts.has_selinux` is specific to the *guest*?

happz: Correct, `self.guest.facts.has_selinux` test whether the guest has selinux - and the `chcon` command executed below is executed *on the guest*. Checking whether the guest has selinux does make sense then, doesn't it?

tweksteen: > the chcon command executed below is executed on the guest

I don't think this is correct. If you look at `ConsoleLog`, it is instantiated on line [992](https://github.com/teemtee/tmt/blob/55b50b904dc3ba7440362eab4a7ea47a6fd804e1/tmt/steps/provision/testcloud.py#L992) within GuestTestCloud.start() (prepare is called just after). The guest is spin up later on, line [1103](https://github.com/teemtee/tmt/blob/55b50b904dc3ba7440362eab4a7ea47a6fd804e1/tmt/steps/provision/testcloud.py#L1103).

When ConsoleLog is instantiated, the guest has not been started yet. I can reproduce experimentally by running `hostname` here and confirm that my host machine is executing this. (Setup: host is Debian testing, guest is Fedora 42, test suite is selinux-testsuite).

I think a parallel issue here is the name `_run_guest_command`, which doesn't quite reflect the behaviour of that method.

happz: Dammit, you're correct, I misread that. Yet even in such a dire situation of me being wrong about something I will fight back! :)

You can test `self.parent.plan.my_run.runner.facts.has_selinux` instead, it describes the runner itself.

tweksteen: No worries. That worked. Please take a look. Thanks!

tmt/steps/provision/testcloud.py
```diff
@@ -1473,9 +1473,10 @@ def prepare(self, logger: tmt.log.Logger) -> None:
         logger.debug(f"Created console log directory '{self.exchange_directory}'.", level=3)
 
         self.exchange_directory.chmod(0o755)
-        self.guest._run_guest_command(
-            Command("chcon", "--type", "virt_log_t", self.exchange_directory), silent=True
-        )
+        if self.guest.parent.plan.my_run.runner.facts.has_selinux:
```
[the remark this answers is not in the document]
psss: All of them need to be assigned, otherwise things would not be working at all.

happz, line 1480:
```suggestion

        assert isinstance(self.guest.parent, tmt.steps.provision.Provision)  # narrow type
        assert self.guest.parent.plan.my_run is not None  # narrow type

        if self.guest.parent.plan.my_run.runner.facts.has_selinux:
```

PR #4025, opened 2025-09-03
Prevent building container images if Containerfile is empty in Image Mode

ukulekek:
There is a hack which TF uses to exclude packages from installation:
```
prepare:
  - how: install
    exclude:
      - curl-minimal
      - libcurl-minimal
```
While it is correctly skipped in Package Mode:
```
        how: install
        
        prepare task #3...
```
In Image Mode it throws the next error:
```
        how: install
        package: building container image with dependencies
        warn: Installation failed, trying again after metadata refresh.
        package: building container image with dependencies
        package: switching to new image localhost/tmt/bootc/98d6fbdf-87c5-45d9-9bf2-d57a7ad7a065
        package: rebooting to apply new image
        package: building container image with dependencies
        fail: Command 'podman build -t localhost/tmt/bootc/71da0589-8b12-4642-a548-8ef063ba2524 -f tmp.sYftMCdBvD/Containerfile /var/ARTIFACTS/work-all-no-buildrootj4t_5bzu/plans/all-no-buildroot/prepare' returned 125.
```
The reason is that the prepare step above doesn't trigger the `install` function of the `BootcEngine`, which leads to an empty `containerfile_directives` of the engine. The patch adds a check to see if `containerfile_directives` is empty and skips the build process completely in that case.  

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


tests/prepare/install/data/empty-exclude.fmf
```diff
@@ -0,0 +1,7 @@
+summary: Empty prepare install with exclude
```
happz: I'd recommend renaming the test, IIUIC it's not "empty exclude", it's "empty with exclude". It took me a while to figure that out.

ukulekek: Good point. Thanks, fixed

thrix (approved):
LGTM

PR #4029, opened 2025-09-04
Replace rsync options in Guest push/pull API

### LINKED ISSUE #3841: Change Guest.push/pull API to not assume everyone uses rsync
Both `push` and `pull` accept `options` parameter, as options to replace default rsync options, `pull` accepts also `extend_options` as options to add to the default ones. And these parameters are completely ignored by any guest implementation that is not based on SSH, i.e. only subclasses of `GuestSsh` consume them while `container` and `local` silently ignore them.

```python
# From `tmt/steps/execute/internal.py`:

        def pull_from_guest() -> None:
            if not invocation.is_guest_healthy:
                return

            try:
                guest.pull(
                    source=invocation.path,
                    extend_options=[
                        *test.test_framework.get_pull_options(invocation, logger),
                        '--exclude',
                        str(invocation.path / TEST_OUTPUT_FILENAME),
                    ],
                )
# And:
            guest.push(
                source=test_wrapper_filepath,
                destination=test_wrapper_filepath,
            )


# From Topology class:

            guest.push(
                source=filepath,
                destination=filepath,
            )


prepare/shell:

            guest.push(
                source=finish_wrapper_path,
                destination=finish_wrapper_path,
            )
```

And so on. With `container` plugin, `options` are worthless and may lead to incorrect or unexpected behavior.

I propose checking all the call sites, extracting the meaning of the options tmt uses, and turn them into parameters: `"--chmod=755"` => `chmod=0755`, `"--exclude"` => `exclude: list[Path]`, `"-p"` => `archive=True`, and so on. If tmt pushes data to a container and must preserve attributes, tmt shall set `archive=True` or `preserve_permissions=True` rather than 1. adding "-p" that rsync would consume and 2. relying on `podman cp --archive` being enabled by default.

therazix:
Remove raw `rsync` options from the Guest push/pull API and replace them with a dedicated `TransferOptions` class.

Resolves #3841 

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [x] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note


tmt/frameworks/__init__.py
```diff
@@ -101,18 +102,23 @@ def get_test_command(
...
-        :returns: additional options for the ``rsync`` tmt would use to pull
-            the test data directory from the guest.
+        :returns: options the tmt would use to pull the test data
+            directory from the guest.
         """
 
-        return []
+        if existing_options:
+            return existing_options
+        return None
```
happz: `return existing_options or None` for a single-liner.

happz: Although, would it make sense to return an actual `TransferOptions` object instead, e.g. `TransferOptions()`? It should have reasonable defaults by default.

tmt/frameworks/beakerlib.py
```diff
@@ -115,8 +116,15 @@ def get_environment_variables(
...
-    def get_pull_options(cls, invocation: 'TestInvocation', logger: tmt.log.Logger) -> list[str]:
-        return ['--exclude', str(invocation.path / "backup*")]
+    def get_pull_options(
+        cls,
+        invocation: 'TestInvocation',
+        existing_options: Optional[TransferOptions],
+        logger: tmt.log.Logger,
+    ) -> Optional[TransferOptions]:
+        options = existing_options or TransferOptions()
+        options.exclude.append(str(invocation.path / "backup*"))
```
happz:
```
            options = test.test_framework.get_pull_options(
                invocation, DEFAULT_PULL_OPTIONS, logger
            )
```

and `options.exclude.append(str(invocation.path / "backup*"))` will be adding "backup" to the default options every time a test runs.

Make a copy of `existing_options`, modify and return that copy. Treat `existing_options` as read-only. Which aligns with my comment above, about returning `TransferOptions()` instead of `None`.

tmt/steps/execute/internal.py
```diff
@@ -621,15 +621,16 @@ def pull_from_guest() -> None:
             if not invocation.is_guest_healthy:
                 return
 
+            options = test.test_framework.get_pull_options(
```
happz, line 624: Could be `options = test.test_framework.get_pull_options(...) or DEFAULT_PULL_OPTIONS`, i.e. one less `if ...` below.

tmt/frameworks/__init__.py
```diff
@@ -101,18 +102,23 @@ def get_test_command(
     def get_pull_options(
         cls,
         invocation: 'TestInvocation',
+        existing_options: Optional[TransferOptions],
```
happz: I'd probably pick slightly different name, those options exist in any case, maybe "current" options? Or just "options"?

tmt/steps/provision/__init__.py
```diff
@@ -316,16 +316,93 @@ def _socket_path_hash(
...
+        options = []
+
+        if self.chmod:
+            options.append(f'--chmod={oct(self.chmod)[2:]}')
+        if self.compress:
+            options.append('-z')
+        if self.delete:
+            options.append('--delete')
+        if self.exclude:
+            options += [f'--exclude {pattern}' for pattern in self.exclude]
```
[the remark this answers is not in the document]
happz: Or `--exclude={pattern}`, that should also help tieing option and its value into a single argument.

tmt/steps/provision/__init__.py
```diff
@@ -316,16 +316,93 @@ def _socket_path_hash(
...
+)
+
+DEFAULT_PULL_OPTIONS = TransferOptions(
+    protect_args=True,
+    relative=True,
+    recursive=True,
+    compress=True,
+    links=True,
+    safe_links=True,
+)
```
happz, line 408: Quite a few `=True` - would it make sense to make `True` the default for these options?

therazix:
@happz In the issue, you mention the following:

> With container plugin, options are worthless and may lead to incorrect or unexpected behavior.

and

> tmt shall set archive=True or preserve_permissions=True rather than 1. adding "-p" that rsync would consume and 2. relying on podman cp --archive being enabled by default.

If I understand it correctly, should I also add `--archive` option to [this command](https://github.com/teemtee/tmt/blob/1934590a04106a35d4c22ad2aaf82c7c5500c25c/tmt/steps/provision/podman.py#L504) based on the TransferOptions passed into the function? If yes, would it make sense to log a warning if some options are provided that will have no effect?

tmt/steps/provision/__init__.py
```diff
@@ -316,16 +316,93 @@ def _socket_path_hash(
...
+    delete: bool = False
+
+    #: Exclude files matching any of these patterns
+    exclude: list[str] = field(default_factory=list, normalize=tmt.utils.normalize_string_list)
+
+    #: Copy symlinks as symlinks
+    links: bool = False
+
+    #: Preserve file permissions
+    preserve_perms: bool = False
```
happz, line 343: I'd argue this should be enabled by default.

happz:
> @happz In the issue, you mention the following:
> 
> > With container plugin, options are worthless and may lead to incorrect or unexpected behavior.
> 
> and
> 
> > tmt shall set archive=True or preserve_permissions=True rather than 1. adding "-p" that rsync would consume and 2. relying on podman cp --archive being enabled by default.
> 
> If I understand it correctly, should I also add `--archive` option to [this command](https://github.com/teemtee/tmt/blob/1934590a04106a35d4c22ad2aaf82c7c5500c25c/tmt/steps/provision/podman.py#L504) based on the TransferOptions passed into the function?

I believe so, yes, that's how it should work. Plugin-agnostic options, each plugin translating them into their own domain.

> If yes, would it make sense to log a warning if some options are provided that will have no effect?

Probably, but it's tricky: if a plugin does not support e.g. `chmod`, and pushed files always get `ugo=rwx` permissions, the plugin should run extra `chmod` to "fix" the permissions so they reflect the given `options`. Not necessarily in this PR, `container` is ignoring them for some time already, no rush, but I think it's what should happen eventually.

On the other hand, some options may be impossible to support, then a warning would be a good start, but eventually we should resolve that too: a caller sets an options because it tries to reach some state on the guest, e.g. executable test script. The plugin cannot deliver that, ever, it has no way to set executable bits, it's really limited. We need to fix that somehow, otherwise the caller would expect something that wouldn't be true, possibly causing seemingly unrelated issues elsewhere.

tmt/steps/provision/__init__.py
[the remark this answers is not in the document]
therazix: Changed in 7cce4affef13f5fc4334ab90224d1d312fac70ab.

tmt/steps/execute/internal.py
```diff
@@ -621,15 +621,16 @@ def pull_from_guest() -> None:
...
+                invocation, DEFAULT_PULL_OPTIONS, logger
+            )
+            if options is None:
+                options = DEFAULT_PULL_OPTIONS
+
+            options.exclude.append(str(invocation.path / TEST_OUTPUT_FILENAME))
```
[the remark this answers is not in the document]
therazix: Changed in 7cce4affef13f5fc4334ab90224d1d312fac70ab.

tmt/frameworks/beakerlib.py
therazix: Changed in 7cce4affef13f5fc4334ab90224d1d312fac70ab.

tmt/steps/execute/internal.py
therazix: Changed in 7cce4affef13f5fc4334ab90224d1d312fac70ab.

tmt/frameworks/__init__.py
therazix: Renamed to `options` in 7cce4affef13f5fc4334ab90224d1d312fac70ab.

tmt/steps/provision/__init__.py
therazix: I thought a better approach would be to keep all options set to `False` by default, and set them to `True` only when needed. There are some cases where we only set a few options, for example [here](https://github.com/teemtee/tmt/blob/fdb098c3d2ff8e568211ff155aec18041d08224c/tmt/steps/__init__.py#L2582), and would need to explicitly set the other options to `False`.

tmt/frameworks/__init__.py
therazix: Changed in 7cce4affef13f5fc4334ab90224d1d312fac70ab.

tmt/steps/provision/podman.py
```diff
@@ -503,6 +503,7 @@ def push(
             self.podman(
                 Command(
                     "cp",
+                    f"--archive={'false' if options and options.preserve_ownership else 'true'}",
```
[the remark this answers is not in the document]
therazix: From the documentation:

> Archive mode (copy all UID/GID information). When set to true, files copied to a container have changed ownership to the primary UID/GID of the container. When set to false, maintain UID/GID from archive sources instead of changing them to the primary UID/GID of the destination container. The default is true.

If I understand it correctly, `--archive=false` will preserve the ownership.

tmt/frameworks/__init__.py
```diff
@@ -101,18 +103,21 @@ def get_test_command(
...
+            can be extended or overridden by the framework.
         :param logger: to use for logging.
-        :returns: additional options for the ``rsync`` tmt would use to pull
-            the test data directory from the guest.
+        :returns: options the tmt would use to pull the test data
+            directory from the guest.
         """
 
-        return []
+        return copy.deepcopy(options) if options else TransferOptions()
```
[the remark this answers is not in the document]
happz: `dataclasses.replace(options)` is easier way to copy a "container" instance.

tmt/frameworks/beakerlib.py
happz: You can use `dataclasses.replace(...)` for copying, it's closer to containers.

tmt/steps/provision/__init__.py
happz: Ok, let's revisit in the future if it becomes unbearable verbose.

tmt/steps/provision/__init__.py
```diff
@@ -2734,9 +2813,7 @@ def pull(
         self._assert_rsync()
 
         # Prepare options and the pull command
-        options = options or DEFAULT_RSYNC_PULL_OPTIONS
-        if extend_options is not None:
-            options.extend(extend_options)
+        options = copy.deepcopy(options or DEFAULT_PULL_OPTIONS)
```
happz: Is the copy needed here? `options` seems to remain unmodified.

tmt/steps/provision/__init__.py
```diff
@@ -316,16 +317,98 @@ def _socket_path_hash(
...
+    delete: bool = False
+
+    #: Exclude files matching any of these patterns
+    exclude: list[str] = field(default_factory=list, normalize=tmt.utils.normalize_string_list)
+
+    #: Copy symlinks as symlinks
+    links: bool = False
+
+    #: Preserve file ownership (user and group)
+    preserve_ownership: bool = False
```
happz: Is this one used anywhere? I see comment below about podman, but the default is `False`, and I can't find where would we enable this particular feature.

tmt/frameworks/__init__.py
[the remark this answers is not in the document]
therazix: Using it here would keep the reference to the `exclude` list. If we then modify this list in the copied object, the original would be modified as well. Is there some easy way to ensure that it creates a separate list?

tmt/steps/provision/__init__.py
therazix: It isn’t needed. I just wanted to make sure we don’t accidentally modify the original options in the future. But I removed the copy in b1a6a97066ac10a193cf99af51505f035839c5f3.

tmt/steps/provision/__init__.py
therazix: This option was added so that we can control the `archive` option for `podman cp`. That’s what I understood from the following:

> > If I understand it correctly, should I also add --archive option to [this command](https://github.com/teemtee/tmt/blob/1934590a04106a35d4c22ad2aaf82c7c5500c25c/tmt/steps/provision/podman.py#L504) based on the TransferOptions passed into the function?
> 
> I believe so, yes, that's how it should work. Plugin-agnostic options, each plugin translating them into their own domain.

I initially wanted to use `preserve_perms`, but the archive option in podman has a different meaning. It preserves ownership, not permissions, so I added this new option for that. It is set to `False` by default, so the behavior remains the same as it is now. But I am not sure in which cases we would want to enable this option.

tmt/frameworks/__init__.py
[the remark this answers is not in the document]
happz: Eh, mutable attributes...

Hm, how about adding a method, e.g. `copy()`, that would do `return dataclasses.replace(self, exclude=self.exclude[:])`?

tmt/steps/provision/__init__.py
happz: I see. Yeah, that's what I said, based on the assumption `preserve_ownership` is needed, but it seems noone is asking for this feature in code. If I'm right, I'd drop the field and `archive` and call it a day.

tmt/frameworks/__init__.py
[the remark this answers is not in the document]
therazix: Great idea. Changed in 0af2764b7e4619e59436cfd9bc262b7b5a81e1d3.

tmt/frameworks/beakerlib.py
therazix: Changed in 0af2764b7e4619e59436cfd9bc262b7b5a81e1d3.

tmt/steps/provision/__init__.py
therazix: Understood. Dropped in 0af2764b7e4619e59436cfd9bc262b7b5a81e1d3.

tmt/steps/provision/__init__.py
```diff
@@ -316,16 +316,98 @@ def _socket_path_hash(
...
+
+    #: Use relative paths
+    relative: bool = False
+
+    #: Ignore symlinks that point outside the source tree
+    safe_links: bool = False
+
+    def copy(self) -> 'TransferOptions':
+        """Create a copy of the options."""
+
```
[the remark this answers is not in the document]
therazix: It is imported.

tmt/steps/provision/__init__.py
therazix: Same as https://github.com/teemtee/tmt/pull/4029#discussion_r2321702005. Unless you mean that `preserve_perms` should also be enabled in `DEFAULT_PUSH_OPTIONS` and `DEFAULT_PULL_OPTIONS`.

tmt/frameworks/__init__.py
```diff
@@ -101,18 +102,21 @@ def get_test_command(
...
         """
         Provide additional options for pulling test data directory.
 
         :param invocation: test invocation to which the check belongs to.
+        :param options: options already specified by tmt, which
+            can be extended or overridden by the framework.
         :param logger: to use for logging.
-        :returns: additional options for the ``rsync`` tmt would use to pull
-            the test data directory from the guest.
+        :returns: options the tmt would use to pull the test data
```
tcornell-bus:
```suggestion
        :returns: options tmt would use to pull the test data
```
Minor nitpick

tmt/frameworks/beakerlib.py
```diff
@@ -115,8 +117,19 @@ def get_environment_variables(
...
-        return ['--exclude', str(invocation.path / "backup*")]
+    def get_pull_options(
+        cls,
+        invocation: 'TestInvocation',
+        options: Optional['TransferOptions'],
+        logger: tmt.log.Logger,
+    ) -> 'TransferOptions':
+        return (
+            dataclasses.replace(
+                options, exclude=[*options.exclude, str(invocation.path / "backup*")]
```
thrix, line 129: @therazix if we are here, would be to good to document this specialty here with the excludes?

tmt/steps/execute/internal.py
```diff
@@ -621,15 +621,13 @@ def pull_from_guest() -> None:
             if not invocation.is_guest_healthy:
                 return
 
+            options = test.test_framework.get_pull_options(
+                invocation, DEFAULT_PULL_OPTIONS, logger
+            )
+            options.exclude.append(str(invocation.path / TEST_OUTPUT_FILENAME))
```
thrix, line 628: also a reason for this exclude would be great

thrix (changes requested):
LGTM, just ask for a bit of docs to be added if we are changing these parts

tmt/frameworks/beakerlib.py
therazix: Added comment in 92db5dff666b67bb744261d39f011ffc61779905.

tmt/steps/execute/internal.py
therazix: Added comment in 92db5dff666b67bb744261d39f011ffc61779905.

therazix:
Added missing exclude in 92db5dff666b67bb744261d39f011ffc61779905.

tmt/frameworks/__init__.py
therazix: Changed in bfe313666eeaf28b1de02d290e4d0a87c6c2f817.

Issue #255, opened 2020-05-22
Add support for running tests under mock


psss:
In the original fedora [issue](https://pagure.io/fedora-ci/general/issue/4) it has been proposed to support running tests in `mock` as well. Let's create a new provision plugin for it then! :)

sopos:
what's the advantage over container?

psss:
Perhaps @hroncok can provide some advantages or his motivation as he proposed `mock` in his original feature request.

hroncok:
I proposed mock, because as a packager I am very used to working with it.

As a benefit over containers, it has dnf cache on host. Hence should be much faster to fetch and install all dependencies.

psss:
@mkoncek, @mizdebsk, fyi.

mizdebsk:
I've been successfully running tmt tests via mock for a couple of days and I'm very pleased with performance and ease of use of this way of running tests. It looks like it's going to become my primary way of running tests locally after every package build. Below are a few important details how I'm doing that.

Inside mock container I am running sshd on a custom port and use the tmt connect provision plugin to connect over SSH to localhost on that port. I decided to go for the SSH way due to Ansible prepare steps, that are most straightforward with SSH. Ansible can't manage mock containers out-of-the-box and would need a custom plugin to be able to do so. SSH could be avoided as long as Ansible prepare steps are not used.

Besides sshd, mock container needs to have dnf installed because tmt will install extra dependencies by ssh-ing into container and calling dnf. This could be avoided by writing a custom tmt provision plugin that would install directly dendencies via mock, without requiring dnf inside container.

mkoncek:
Thanks to @xsuchy, I discovered that the mock plugin `pm_request` is probably ideal for handling package installation into the environment (from the `prepare` step).

mkoncek:
After some more experimenting, I found the following.
## ❌ `pm_request`
It works and is possible to use for implementing custom package managers and the `prepare::install` step. However there is no clean way to implement `prepare::shell` and allow the user to call the package manager from inside the shell.

## ✅ `overlayfs`
We can use `mount -t overlay ...` to overlay the read-only bootstrap mock chroot (which contains the package manager binary) and allow tmt to proceed with normal PM discovery and usage. Mock plugins are already running with escalated privileges so we don't need to use `sudo`.

## ❌ Using mock plugin `mount_bind` for the tmt work directory
[...]

PR #4034, opened 2025-09-05
Implement the `mock` provision plugin

mkoncek:
This pull request implements a guest provisioning using the `mock` tool. Mock provides isolation similar to containers but also easy access to the file system similar to local provisioning. Moreover, it provides resource reuse, and it should be easily possible to run tests on newly built packages.

This plugin has minimal custom configuration and leaves it to mock's configuration.

See #4097 for follow-up issues to be addressed later.

Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema
* [x] include a release note

### DIFF
--- a/tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/main.fmf
+++ b/tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/main.fmf
@@ -2,3 +2,7 @@ summary: Test Fedora Rawhide mock container
 description:
     A simple test to see if the provision does at least provide us with a
     working shell.
+
+# TODO: Failing on `assert self.mock_shell is not None`
+# Disabling for now, not to block the proof-of-concept
+enabled: false
--- a/tests/provision/mock/main.fmf
+++ b/tests/provision/mock/main.fmf
@@ -1,3 +1,3 @@
 tag+:
   - provision-only
-  - provision-mock
+  - provision-virtual


mkoncek:
This plugin will require importing the `mockbuild` module.
Moreover I have not yet written tests.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,410 @@
...
+        prepare_data_class = cast(  # type: ignore[redundant-cast]
+            type[tmt.steps.prepare.shell.PrepareShellData],
+            tmt.steps.prepare.shell.PrepareShell.get_data_class(),
+        )
+
+        data = prepare_data_class(
+            name="tmt-unmount-mock-bootstrap-overlay",
+            how='shell',
+            script=[
+                TODO
```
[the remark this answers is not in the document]
happz: No.

mkoncek:
Some comments and important notes are in the related Issue: https://github.com/teemtee/tmt/issues/255

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,504 @@
...
+        self.root_path = Path(root_path.rstrip())
+
+    def __del__(self):
+        self.exit_shell()
+
+    def exit_shell(self) -> None:
+        if self.mock_shell is not None:
+            self.parent.verbose("Exiting mock shell")
+
+            self.mock_shell.stdin.write("")
```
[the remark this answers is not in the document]
mkoncek: Sending an empty string acts as sending EOF to the shell.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,505 @@
...
+        shell_command += " 1>/tmp/stdout 2>/tmp/stderr; echo $?>/tmp/returncode"
+
+        self.parent.verbose("Executing inside mock shell", shell_command)
+        shell_command += "\n"
+
+        stdout_file = self.root_path / "tmp/stdout"
+        stderr_file = self.root_path / "tmp/stderr"
+        returncode_file = self.root_path / "tmp/returncode"
+
+        with io.FileIO(os.open(str(stdout_file), os.O_RDONLY | os.O_NONBLOCK)) as stdout_io, io.FileIO(os.open(str(stderr_file), os.O_RDONLY | os.O_NONBLOCK)) as stderr_io, io.FileIO(os.open(str(returncode_file), os.O_RDONLY | os.O_NONBLOCK)) as returncode_io:
```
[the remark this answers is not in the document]
mkoncek: Seems like a valid concern. Empirically in a normal program flow all fds are unregistered in the main loop. If there is an error, then we indeed have some leaks.
Fixing this would require some more complex object wrapping in context managers.

mcasquer:
@mkoncek could you update the PR description including some information regarding this new plugin?, the purpose of the PR is clear but it'd nice having more background e.g. of use cases, invocation..

mcasquer:
@mkoncek moreover, is there a plan to include documentation along with this new plugin? I cannot find anything other than `_run_ansible` docstring

mkoncek:
>@mkoncek moreover, is there a plan to include documentation along with this new plugin? I cannot find anything other than _run_ansible docstring

Right, of course, it would take some time though. I usually write more detailed documentation after having some prototype, but I don't expect many further changes.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,505 @@
...
+        self.mock_shell = MockShell(parent=self, config=self.config, rootdir=self.rootdir)
+        self.mock_config = data.mock_config
+
+    def save(self) -> tmt.steps.provision.GuestData:
+        # inherit
+        return super().save()
+
+    def wake(self) -> None:
+        # noop
+        return super().wake()
```
happz: Hm, can these bodies be dropped then?

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,505 @@
...
+
+        # NOTE this may be unnecessary, tmt package manager detection should work fine
+        # but mock already knows what the package manager is...
+        mock_config = mockbuild.config.simple_load_config(data.config)
+        package_manager = "mock-" + mock_config['package_manager']
+
+        # If this provisioning is selected, then we force the use of `mock-` package manager.
+        # NOTE use a global variable instead?
+        tmt.package_managers.find_package_manager(
+            package_manager
```
happz:
```suggestion
            "mock-" + mock_config['package_manager']
```

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,505 @@
...
+
+        # NOTE any better ideas?
+        data.primary_address = (data.config or '<default>') + (
+            '@' + data.rootdir if data.rootdir is not None else ''
+        )
+        data.mock_config = mock_config
+
+        data.show(verbose=self.verbosity_level, logger=self._logger)
+
+        self.assert_feeling_safe("1.38", "The 'mock' provision plugin")
```
happz: `1.38` is probably incorrectly copied. I think you can use `1.58` immediately.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,505 @@
...
+        shell_command = ""
+        if cwd is not None:
+            shell_command += "(cd " + shlex.quote(str(cwd)) + " && "
+        if env is not None:
+            for key, value in env.items():
+                shell_command += f"{key}={shlex.quote(value)} "
+        shell_command += str(command)
+        if cwd is not None:
+            shell_command += ")"
+        shell_command += " 1>/tmp/stdout 2>/tmp/stderr; echo $?>/tmp/returncode"
```
happz: Too many leading and trailing spaces for my taste. And round brackets. Easy to forget one here and there. How about moving the spaces and round bracket handling to one place? Something like this.

```python
shell_command_components: list[str] = [
    command
]

if env is not None:
    shell_command_components = [
        *(f"{key}={shlex.quote(value)}" for key, value in env.items()),
        *shell_command_components
    ]

if cwd is not None:
    shell_command_components = [
        "(",
        f"cd {shlex.quote(str(cwd))}",
        "&&",
        *shell_command_components,
        ")"
    ]

shell_command_components = [
    *shell_command_components,
    "1>/tmp/stdout 2>/tmp/stderr; echo $?"
]

shell_command = ' '.join(shell_command_components)
```

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,505 @@
...
+                if pos == -1:
+                    break
+                string = self.output[:pos].decode('utf-8', errors='replace')
+                self.output = self.output[pos + 1:]
+                self.logger(string)
+                self.string += string
+                self.string += '\n'
+            return self
+
+    def __init__(self, parent: 'GuestMock', config: Optional[str], rootdir: Optional[str] = None):
```
happz: Both `config` and `rootdir` specified in the single instantiation of `MockShell`, is it necessary to have default value for one of them? `MockShell(parent=self, config=self.config, rootdir=self.rootdir)`.

happz:
One followup note: I miss the goodies we have in `Command.run()` - friendly command, logging, duration, exit code logging, stdout/stderr, timing events. I don't think it's a blocker, but I strongly believe the "run command on a guest" situation in log file should be plugin-agnostic - extra messages don't hurt as much as missing ones, makes the investigation of issues ridden with exceptions based on which provision plugin was used. It should be fixed ASAP.

tmt/steps/provision/mock.py
mkoncek: I kept it separated because at some point i wanted to print the value...

tmt/steps/provision/mock.py
mkoncek: Yes, looks good.

tmt/steps/provision/mock.py
happz: I guess it fits into the "align the logging with the rest of tmt" bucket I'd like to sort out as a follow up. tmt showing guest facts should print out package manager as well as other bits, I wouldn't want to see individual plugins invent their logging of shared properties and processes.

mkoncek:
1. I would start mirroring the functionality found in `Command.run`. There are a few questions related to `dry` run for example. Whether we should actually enter mock and run `echo $command` or just print the cmd.
2. I agree that the logging as written by me was quite ad-hoc and more for development. It does need to be unified with the rest of the sources.
3. I noticed that I had to write:
```
stdout = MockShell.Stream(lambda text: output_logger("out", text, 'yellow', level = 0))
stderr = MockShell.Stream(lambda text: output_logger("err", text, 'yellow', level = 0))
```
to get any output, I know that for example the `container` provision uses `level=2`, but I could not see the logs even when running `tmt -vvv`.

mkoncek:
As i mentioned elsewhere, for this plugin to work, it needs to install a mock plugin into the `mockbuild` directory. Where should such file be placed in the source base of this project?

LecrisUT:
> As i mentioned elsewhere, for this plugin to work, it needs to install a mock plugin into the `mockbuild` directory. Where should such file be placed in the source base of this project?

We don't have one, but you could make one like `contrib` or `third_party` and others will comment if they have a specific preference.

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,43 @@
...
+@provides_package_manager('mock-dnf')
+class MockDnf(BaseMock, PackageManager[MockDnfEngine]):
+    NAME = 'mock-dnf'
+    _engine_class = MockDnfEngine
+
+
+@provides_package_manager('mock-dnf5')
+class MockDnf5(BaseMock, PackageManager[MockDnf5Engine]):
+    NAME = 'mock-dnf5'
+    _engine_class = MockDnf5Engine
```
LecrisUT: When we discussed needing to install `dnf` directly into `mock`, I mentioned that I recalled being able to run `mock --install` even with an active `mock --shell`. I just tested it and confirmed it worked. Is there more context that I did not understand?

happz:
> 1. I would start mirroring the functionality found in `Command.run`. There are a few questions related to `dry` run for example. Whether we should actually enter mock and run `echo $command` or just print the cmd.
> 2. I agree that the logging as written by me was quite ad-hoc and more for development. It does need to be unified with the rest of the sources.
> 3. I noticed that I had to write:
> 
> ```
> stdout = MockShell.Stream(lambda text: output_logger("out", text, 'yellow', level = 0))
> stderr = MockShell.Stream(lambda text: output_logger("err", text, 'yellow', level = 0))
> ```
> 
> to get any output, I know that for example the `container` provision uses `level=2`, but I could not see the logs even when running `tmt -vvv`.

It would help collecting these improvements in an issue.

tmt/package_managers/mock.py
mkoncek: I remember now. The reason why I wanted to always have the package manager executable present in some form in the mock environment is for implementation of the `prepare::shell` step. Even our packages do something like that. For example we needed to execute `dnf swap` on time.

`mock --install` could be theoretically used for implementing things like file / URL / directory installation though.

LecrisUT: > For example we needed to execute `dnf swap` on time.

Is that in the mock provision client context though? It seems like it would be dependent on specific test implementations. Could the enabling of dnf be an opt-in feature then?

psss:
@mkoncek, could you please add a short release note, include the corresponding [schema](https://github.com/teemtee/tmt/tree/main/tmt/schemas/provision) and at least a simple test?

psss:
> It would help collecting these improvements in an issue.

Filed here:

* #4097

psss:
Package building seems to be failing because of:

    ModuleNotFoundError: No module named 'mockbuild'

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,497 @@
...
+import select
+import shlex
+import subprocess
+from pathlib import Path
+from typing import Any, Optional, Union
+
+import mockbuild
+import mockbuild.config
+import mockbuild.plugins
+import mockbuild.plugins.pm_request
```
LecrisUT: See `tmt/steps/prepare/artifact/providers/koji_provider.py` for a recent example of how we deal with optional imports (yeah it is not pretty).

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,121 @@
...
+    We use `mock --pm-cmd ...` to execute the package manager commands inside
+    the mock. Such scripts need to be executed locally and not inside the mock
+    shell. To differentiate these scripts we set a special attribute `_local`
+    of the script object.
+    """
+
+    def _prepare_local_script(self, script: str):
+        result = ShellScript(f'{self.command} {self.options.to_script()} {script}')
+        setattr(result, "_local", True)
+        return result
```
LecrisUT: `Command.run`, `Guest.run`, `Guest._run_guest_command`  would always be executed locally, while `guest.execute` are the commands that will be executed on the guest. Let's just use those intead of injecting a `_local`

tmt/steps/provision/mock.py
mkoncek: Thanks. `mockbuild` is only needed inside of `go`. So I may not need a global variable and simply use the import inside a function.

LecrisUT: That part is for printing import errors and it gets in the way of mypy and stuff, it needs to be revisited, but for now let's keep using it

tmt/package_managers/mock.py
mkoncek: Right, I believe I could add the conditionals into the `install.py` file. Originally I copied it from the existing ones.

mkoncek: Actually I don't understand the point of package manager engines.

happz: Bootc and image mode. In some situations, we do not want the `dnf install foo` to actually run on the guest, but we still want that command to be put together. Instead of running it, which is the job of a package manager class, we have decoupled the composition of commands into engines, and `bootc` then call e.g. `dnf` package manager engine, collect the commands, and shove them into a Containerfile. The rest of the world would use the `dnf` package manager to compose *and* execute commands on guests directly. Hence the separation.

mkoncek: I believe I addressed it more-less in the latest commit. It would be still a good idea to revisit the command building for the mock package manager.

mcasquer:
@mkoncek regarding the failures in the RPM building, in Fedora 42 I glimpsed:
```
Processing files: tmt+provision-mock-1.58.0.dev55+g7183d2b9d-1.20250924105001229033.pr4034.55.g7183d2b9.fc42.noarch

Error: The package name contains an extras name `provision-mock` that was not found in the metadata.
Check if the extras were removed from the project. If so, consider removing the subpackage and obsoleting it from another.

error: Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_EXTRAS_NOT_FOUND_ERROR___SEE_STDERR ***

Error: The package name contains an extras name `provision-mock` that was not found in the metadata.
Check if the extras were removed from the project. If so, consider removing the subpackage and obsoleting it from another.

error: Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_EXTRAS_NOT_FOUND_ERROR___SEE_STDERR ***
Provides: tmt+provision-mock = 1.58.0.dev55+g7183d2b9d-1.20250924105001229033.pr4034.55.g7183d2b9.fc42 tmt-provision-mock = 1.58.0.dev55+g7183d2b9d-1.20250924105001229033.pr4034.55.g7183d2b9.fc42
Requires(rpmlib): rpmlib(CompressedFileNames) <= 3.0.4-1 rpmlib(FileDigests) <= 4.6.0-1 rpmlib(PayloadFilesHavePrefix) <= 4.0-1

RPM build errors:
    Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_EXTRAS_NOT_FOUND_ERROR___SEE_STDERR ***
    Dependency tokens must begin with alpha-numeric, '_' or '/': *** PYTHON_EXTRAS_NOT_FOUND_ERROR___SEE_STDERR ***
```

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,144 @@
...
+class BaseMockEngine(PackageManagerEngine):
+    """
+    We use `mock --pm-cmd ...` to execute the package manager commands inside
+    the mock. Such scripts need to be executed locally and not inside the mock
+    shell. To differentiate these scripts we set a special attribute `_local`
+    of the script object.
+    """
+
+    def _prepare_local_script(self, script: str):
+        return ShellScript(f'{self.command} {self.options.to_script()} {script}')
```
LecrisUT: The naming of the function is confusing. Try `_mock_script`, `_make_mock_script`, etc. It took me a few readings to realize `mock` is always added there.

Otherwise I would say better use a `Command` as an input to make sure it is always a single command being run (compared to scripts which can have multiple lines ans such).

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,144 @@
...
+        if self.guest.config is not None:
+            options += Command('-r')
+            options += Command(self.guest.config)
+        options += Command('--pm-cmd')
+        return (Command('mock'), options)
+
+    def check_presence(self, *installables: Installable) -> ShellScript:
+        return ShellScript(
+            f'rpm -q --whatprovides {" ".join(escape_installables(*installables))}'
+        )
```
LecrisUT: I am inclined to make this a `/usr/bin/false` just so that it always does the installation (we always assume `mock` is a clean environment so we don't need this check). But probably something to be addressed separately.

One thing I didn't check is if this thing is always run on the same "guest" like in the case `Local` provision, or does the Guest decide how to consume this.

mkoncek:
I am not sure about the RPM issue. I copied the .spec declarations from other plugins.

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,144 @@
...
+            f'{installword} {extra_options} '
+            f'{" ".join(escape_installables(*installables))}'
+        )
+
+    def prepare_command(self) -> tuple[Command, Command]:
+        options = Command()
+        assert isinstance(self.guest, GuestMock)
+        if self.guest.config is not None:
+            options += Command('-r')
+            options += Command(self.guest.config)
```
LecrisUT:
```suggestion
            options += Command('-r', self.guest.config)
```
More readable?

LecrisUT:
> I am not sure about the RPM issue. I copied the .spec declarations from other plugins.

It is missing an equivalent entry in `pyproject.toml`

psss:
There's still a bunch of open comments, red pre-commit and tests failing as well. Let's finish this in the next sprint.

mkoncek:
A lot of pre-commit mypy warnings are from `package_managers/mock.py` about that `GuestMock` has no attribute `config` and so. Those attributes are assigned dynamically. Is there a standard way to pass these warnings?

LecrisUT:
> A lot of pre-commit mypy warnings are from `package_managers/mock.py` about that `GuestMock` has no attribute `config` and so. Those attributes are assigned dynamically. Is there a standard way to pass these warnings?

Add it as a class attribute.
```python
class GuestMock(tmt.Guest):
    _data_class = MockGuestData
    config: str
```

mkoncek:
I fixed many issues. I will take care of the `execute` function later. I will need help with the `package_managers` errors and then I need to know about the `OnProcessStartCallback`-s second parameter (`Popen[bytes]`). I don't know if it fits into the current execute model, where we do not spawn processes on the guest machine but send commands to the running shell and read outputs.

tmt/package_managers/mock.py
mkoncek: Whether `mock` is a clean environment is very much for debate. The current implementation does not clean it. And users may want to reuse the environments just after their builds.

This script is intended to be executed on the guest (i.e. inside the mock shell) because `rpm` is available there and it is easier to read the command results.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,504 @@
...
+
+    def exit_shell(self) -> None:
+        if self.mock_shell is not None:
+            self.parent.verbose("Exiting mock shell")
+
+            self.mock_shell.stdin.write("")
+            self.mock_shell.stdin.flush()
+            self.mock_shell.communicate()
+            self.mock_shell = None
+            self.epoll.close()
```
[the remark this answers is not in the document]
mkoncek: Enter / Exit shell needs a complete revisit.

mkoncek:
There are currently the following lint failures:
```
check for merge conflicts................................................Passed
check for broken symlinks............................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
detect destroyed symlinks................................................Passed
detect private key.......................................................Passed
fix end of files.........................................................Passed
mixed line ending........................................................Passed
don't commit to branch...................................................Passed
trim trailing whitespace.................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tmt/steps/provision/mock.py:415: error: Argument 2 has incompatible type "Optional[Popen[str]]"; expected "Popen[bytes]"  [arg-type]
tmt/steps/provision/mock.py:472: error: "MockGuestData" has no attribute "mock_config"  [attr-defined]
tmt/steps/provision/mock.py:544: error: "MockGuestData" has no attribute "mock_config"  [attr-defined]
tmt/package_managers/mock.py:106: error: Item "None" of "Optional[Options]" has no attribute "check_first"  [union-attr]
tmt/package_managers/mock.py:108: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:108: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:108: error: "BaseMock" has no attribute "engine"  [attr-defined]
tmt/package_managers/mock.py:111: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:111: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:112: error: "BaseMock" has no attribute "engine"  [attr-defined]
tmt/package_managers/mock.py:122: error: Item "None" of "Optional[Options]" has no attribute "check_first"  [union-attr]
tmt/package_managers/mock.py:124: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:124: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:124: error: "BaseMock" has no attribute "engine"  [attr-defined]
tmt/package_managers/mock.py:127: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:127: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:128: error: "BaseMock" has no attribute "engine"  [attr-defined]
tmt/package_managers/mock.py:136: error: Item "None" of "Optional[Options]" has no attribute "check_first"  [union-attr]
tmt/package_managers/mock.py:138: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:138: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:138: error: "BaseMock" has no attribute "engine"  [attr-defined]
tmt/package_managers/mock.py:141: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:141: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:142: error: "BaseMock" has no attribute "engine"  [attr-defined]
tmt/package_managers/mock.py:146: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
tmt/package_managers/mock.py:146: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:146: error: "BaseMock" has no attribute "engine"  [attr-defined]
Found 27 errors in 2 files (checked 129 source files)

pyright..................................................................Failed
- hook id: pyright
- exit code: 1
- files were modified by this hook

WARNING: there is a new pyright version available (v1.1.399 -> v1.1.405).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

/home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:106:20 - error: "check_first" is not a known attribute of "None" (reportOptionalMemberAccess)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:108:24 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:108:29 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:108:48 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:111:16 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:111:21 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:112:18 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:122:20 - error: "check_first" is not a known attribute of "None" (reportOptionalMemberAccess)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:124:24 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:124:29 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:124:48 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:127:16 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:127:21 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:128:18 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:136:20 - error: "check_first" is not a known attribute of "None" (reportOptionalMemberAccess)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:138:24 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:138:29 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:138:48 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:141:16 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:141:21 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:142:18 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:146:16 - error: Return type is unknown (reportUnknownVariableType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:146:21 - error: Cannot access attribute "guest" for class "BaseMock*"
    Attribute "guest" is unknown (reportAttributeAccessIssue)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:146:36 - error: Cannot access attribute "engine" for class "BaseMock*"
    Attribute "engine" is unknown (reportAttributeAccessIssue)
24 errors, 0 warnings, 0 informations

Check JSON schemas validity..........................(no files to check)Skipped
yamllint.............................................(no files to check)Skipped
Ansible-lint.............................................................Passed
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1
- files were modified by this hook

tmt/steps/provision/mock.py:83:14: RUF005 Consider `[*self.command_prefix, '--print-root-path']` instead of concatenation
   |
82 |         root_path = (
83 |             (self.command_prefix + ['--print-root-path'])
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ RUF005
84 |             .run(cwd=None, logger=self.parent._logger)
85 |             .stdout
   |
   = help: Replace with `[*self.command_prefix, '--print-root-path']`
```
There are a few categories of issues:
* `Popen` issue: `OnProcessStartCallback` has a parameter `Popen[bytes]` which is usually the spawned process. In our case the spawned process is the whole shell and I don't know how to handle this.
* `mock_config` - I dynamically assign the expanded dictionary after load so I don't have to read the configuration files multiple times. But I cannot add this as a field to `MockGuestData` because it deos not belong there and linter will complain too.
* `package_managers/mock.py`: Python inheritance and code reuse - I am not too experienced with Python.
* `RUF005` - this is bogus, `Command` supports `+` operator but not array expansion.

happz:
> There are currently the following lint failures:
> 
> ```
> check for merge conflicts................................................Passed
> check for broken symlinks............................(no files to check)Skipped
> check toml...........................................(no files to check)Skipped
> detect destroyed symlinks................................................Passed
> detect private key.......................................................Passed
> fix end of files.........................................................Passed
> mixed line ending........................................................Passed
> don't commit to branch...................................................Passed
> trim trailing whitespace.................................................Passed
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 1
> 
> tmt/steps/provision/mock.py:415: error: Argument 2 has incompatible type "Optional[Popen[str]]"; expected "Popen[bytes]"  [arg-type]
> tmt/steps/provision/mock.py:472: error: "MockGuestData" has no attribute "mock_config"  [attr-defined]
> tmt/steps/provision/mock.py:544: error: "MockGuestData" has no attribute "mock_config"  [attr-defined]
> tmt/package_managers/mock.py:106: error: Item "None" of "Optional[Options]" has no attribute "check_first"  [union-attr]
> tmt/package_managers/mock.py:108: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:108: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:108: error: "BaseMock" has no attribute "engine"  [attr-defined]
> tmt/package_managers/mock.py:111: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:111: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:112: error: "BaseMock" has no attribute "engine"  [attr-defined]
> tmt/package_managers/mock.py:122: error: Item "None" of "Optional[Options]" has no attribute "check_first"  [union-attr]
> tmt/package_managers/mock.py:124: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:124: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:124: error: "BaseMock" has no attribute "engine"  [attr-defined]
> tmt/package_managers/mock.py:127: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:127: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:128: error: "BaseMock" has no attribute "engine"  [attr-defined]
> tmt/package_managers/mock.py:136: error: Item "None" of "Optional[Options]" has no attribute "check_first"  [union-attr]
> tmt/package_managers/mock.py:138: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:138: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:138: error: "BaseMock" has no attribute "engine"  [attr-defined]
> tmt/package_managers/mock.py:141: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:141: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:142: error: "BaseMock" has no attribute "engine"  [attr-defined]
> tmt/package_managers/mock.py:146: error: Returning Any from function declared to return "CommandOutput"  [no-any-return]
> tmt/package_managers/mock.py:146: error: "BaseMock" has no attribute "guest"  [attr-defined]
> tmt/package_managers/mock.py:146: error: "BaseMock" has no attribute "engine"  [attr-defined]
> Found 27 errors in 2 files (checked 129 source files)
> 
> pyright..................................................................Failed
> - hook id: pyright
> - exit code: 1
> - files were modified by this hook
> 
> WARNING: there is a new pyright version available (v1.1.399 -> v1.1.405).
> Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`
> 
> /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:106:20 - error: "check_first" is not a known attribute of "None" (reportOptionalMemberAccess)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:108:24 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:108:29 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:108:48 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:111:16 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:111:21 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:112:18 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:122:20 - error: "check_first" is not a known attribute of "None" (reportOptionalMemberAccess)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:124:24 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:124:29 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:124:48 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:127:16 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:127:21 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:128:18 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:136:20 - error: "check_first" is not a known attribute of "None" (reportOptionalMemberAccess)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:138:24 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:138:29 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:138:48 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:141:16 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:141:21 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:142:18 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:146:16 - error: Return type is unknown (reportUnknownVariableType)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:146:21 - error: Cannot access attribute "guest" for class "BaseMock*"
>     Attribute "guest" is unknown (reportAttributeAccessIssue)
>   /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:146:36 - error: Cannot access attribute "engine" for class "BaseMock*"
>     Attribute "engine" is unknown (reportAttributeAccessIssue)
> 24 errors, 0 warnings, 0 informations
> 
> Check JSON schemas validity..........................(no files to check)Skipped
> yamllint.............................................(no files to check)Skipped
> Ansible-lint.............................................................Passed
> ruff.....................................................................Failed
> - hook id: ruff
> - exit code: 1
> - files were modified by this hook
> 
> tmt/steps/provision/mock.py:83:14: RUF005 Consider `[*self.command_prefix, '--print-root-path']` instead of concatenation
>    |
> 82 |         root_path = (
> 83 |             (self.command_prefix + ['--print-root-path'])
>    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ RUF005
> 84 |             .run(cwd=None, logger=self.parent._logger)
> 85 |             .stdout
>    |
>    = help: Replace with `[*self.command_prefix, '--print-root-path']`
> ```
> 
> There are a few categories of issues:
> 
> * `Popen` issue: `OnProcessStartCallback` has a parameter `Popen[bytes]` which is usually the spawned process. In our case the spawned process is the whole shell and I don't know how to handle this.
> * `mock_config` - I dynamically assign the expanded dictionary after load so I don't have to read the configuration files multiple times. But I cannot add this as a field to `MockGuestData` because it deos not belong there and linter will complain too.
> * `package_managers/mock.py`: Python inheritance and code reuse - I am not too experienced with Python.
> * `RUF005` - this is bogus, `Command` supports `+` operator but not array expansion.

Shall I try to tackle these? Last time I offered you said you'd try your luck, I can still give it a try.

mkoncek:
>Shall I try to tackle these? Last time I offered you said you'd try your luck, I can still give it a try.

Yes please, my luck has run out.

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,162 @@
...
+    probe_priority = 130
+
+    def install(
+        self,
+        *installables: Installable,
+        options: Optional[Options] = None,
+    ) -> CommandOutput:
+        if options.check_first:
+            try:
+                return self.guest.execute(self.engine.check_presence(*installables))
```
LecrisUT, line 96: Re
```
tmt/package_managers/mock.py:108: error: "BaseMock" has no attribute "guest"  [attr-defined]
tmt/package_managers/mock.py:108: error: "BaseMock" has no attribute "engine"  [attr-defined]
```
It's quite straighforward where they fail. The class `BaseMock` does not have `guest` or `engine` either defined in `__init__` or inherited from somewhere. I believe your intent here is to be a mixin? in which case you would want
```python
class BaseMock:
    probe_command = Command('/usr/bin/false')
    probe_priority = 130
    guest: tmt.steps.provision.Guest
    engine: tmt.package_managers.PackageManagerEngine
```

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,552 @@
...
+        self.pid: Optional[int] = None
+
+        self.command_prefix = Command('mock')
+        if self.root is not None:
+            self.command_prefix += ['-r', self.root]
+        if self.rootdir is not None:
+            self.command_prefix += ['--rootdir', self.rootdir]
+
+        root_path = (
+            (self.command_prefix + ['--print-root-path'])
```
```suggestion
            (self.command_prefix + Command('--print-root-path'))
```
I believe this is the common standard all across. Similarly
```python
self.command_prefix += Command('-r', self.root)
```

LecrisUT:
> * `mock_config` - I dynamically assign the expanded dictionary after load so I don't have to read the configuration files multiple times. But I cannot add this as a field to `MockGuestData` because it deos not belong there and linter will complain too.

Add it to `GuestMock` instead. You would need to override the `__init__` to initialize a new dict for it at construction.

> * `Popen` issue: `OnProcessStartCallback` has a parameter `Popen[bytes]` which is usually the spawned process. In our case the spawned process is the whole shell and I don't know how to handle this.

Best way would be inherit `Popen`. Unfortunately this seems to be the first case where we have an active terminal so this is quite new territory for us too (it would be really cool to learn from this though).

tmt/package_managers/mock.py
mkoncek: There are additional errors regarding those classes:
```
/home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:150:2 - error: Argument of type "type[MockYum]" cannot be assigned to parameter of type "PackageManagerClass[PackageManagerEngine]"
    "type[MockYum]" is not assignable to "PackageManagerClass[PackageManagerEngine]"
    Type "type[MockYum]" is not assignable to type "PackageManagerClass[PackageManagerEngine]"
      Type parameter "PackageManagerEngineT@PackageManager" is invariant, but "MockYumEngine" is not the same as "PackageManagerEngine" (reportArgumentType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:151:7 - error: Base classes for class "MockYum" define variable "engine" in incompatible way (reportIncompatibleVariableOverride)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:156:2 - error: Argument of type "type[MockDnf]" cannot be assigned to parameter of type "PackageManagerClass[PackageManagerEngine]"
    "type[MockDnf]" is not assignable to "PackageManagerClass[PackageManagerEngine]"
    Type "type[MockDnf]" is not assignable to type "PackageManagerClass[PackageManagerEngine]"
      Type parameter "PackageManagerEngineT@PackageManager" is invariant, but "MockDnfEngine" is not the same as "PackageManagerEngine" (reportArgumentType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:157:7 - error: Base classes for class "MockDnf" define variable "engine" in incompatible way (reportIncompatibleVariableOverride)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:162:2 - error: Argument of type "type[MockDnf5]" cannot be assigned to parameter of type "PackageManagerClass[PackageManagerEngine]"
    "type[MockDnf5]" is not assignable to "PackageManagerClass[PackageManagerEngine]"
    Type "type[MockDnf5]" is not assignable to type "PackageManagerClass[PackageManagerEngine]"
      Type parameter "PackageManagerEngineT@PackageManager" is invariant, but "MockDnf5Engine" is not the same as "PackageManagerEngine" (reportArgumentType)
  /home/mkoncek/Upstream/tmt/tmt/package_managers/mock.py:163:7 - error: Base classes for class "MockDnf5" define variable "engine" in incompatible way (reportIncompatibleVariableOverride)
6 errors, 0 warnings, 0 informations
```
Though I was thinking the whole package manager module should be refactored with support for DNF 4 and YUM.

mkoncek:
> Add it to `GuestMock` instead. You would need to override the `__init__` to initialize a new dict for it at construction.

@LecrisUT
Will the dictionary be serialized and saved between tmt invocations? I think that is handled via the `Data` class.

tmt/package_managers/mock.py
LecrisUT: Yes those were newly introduced in https://github.com/teemtee/tmt/pull/4109. Currently we skip that check so try to use a similar comment as in that PR

LecrisUT:
> > Add it to `GuestMock` instead. You would need to override the `__init__` to initialize a new dict for it at construction.
> 
> @LecrisUT Will the dictionary be serialized and saved between tmt invocations? I think that is handled via the `Data` class.

Good question, I do not know, probably @happz and @psss know it. But workaround that is to keep it in `Data` but make it a `field(internal=True)`. I do not like that design if it can be avoided, but around this serealization/desrealization, I do not have the experience.

happz:
> > * `Popen` issue: `OnProcessStartCallback` has a parameter `Popen[bytes]` which is usually the spawned process. In our case the spawned process is the whole shell and I don't know how to handle this.
> 
> Best way would be inherit `Popen`. Unfortunately this seems to be the first case where we have an active terminal so this is quite new territory for us too (it would be really cool to learn from this though).

I'd try to relax the `bytes` parameter, or make it a `TypeVar`. We don't really care whether it's `bytes` or `str` in that callback, it just needs to be the right `Popen` object. Probably beyond the scope of this PR, I'd add a `type: ignore` - still need to sort out it's `Optional`, i.e. `assert ... is not None  # narrow type` is needed - with a not in the tracking issue, and be done with it here and now.

happz:
> > > Add it to `GuestMock` instead. You would need to override the `__init__` to initialize a new dict for it at construction.
> > 
> > 
> > @LecrisUT Will the dictionary be serialized and saved between tmt invocations? I think that is handled via the `Data` class.
> 
> Good question, I do not know, probably @happz and @psss know it. But workaround that is to keep it in `Data` but make it a `field(internal=True)`. I do not like that design if it can be avoided, but around this serealization/desrealization, I do not have the experience.

`field(internal=True)` shoudl work. Also, may it be a `cached_property` of `GuestMock` class? It seems to be constructed from `data.root`, that should be `self.root` inside `GuestMock`, does it need to be stored in `guests.yaml`?

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
thrix: What to do with this?

tmt/steps/provision/mock.py
thrix: @mkoncek ping

tmt/steps/provision/mock.py
thrix: @mkoncek what about to create an issue tracking these agreed-upon follow ups?

tmt/steps/provision/mock.py
thrix: seems not present anymore

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,410 @@
...
+        self.root_path = Path(root_path.rstrip())
+
+    def __del__(self):
+        self.exit_shell()
+
+    def exit_shell(self) -> None:
+        if self.mock_shell is not None:
+            self.parent.verbose("Exiting mock shell")
+
+            self.mock_shell.stdin.write("")
```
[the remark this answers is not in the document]
thrix: outdated, closing

tmt/steps/provision/mock.py
thrix: resolving, looks addressed

thrix:
@mkoncek the tests are failing can you look at them?

tmt/steps/provision/mock.py
mkoncek: Already fixed.

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
thrix: @mkoncek is this part of the follow up items, or something we need to address NOW?

tmt/steps/provision/mock.py
mkoncek: Followup issue created: https://github.com/teemtee/tmt/issues/4097

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
thrix: @mkoncek file it as a follow up item?

tmt/steps/provision/mock.py
mkoncek: Already fixed.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,504 @@
...
+        # NOTE this may be unnecessary, tmt package manager detection should work fine
+        # but mock already knows what the package manager is...
+        mock_config = mockbuild.config.simple_load_config(data.config)
+        package_manager = "mock-" + mock_config['package_manager']
+
+        # If this provisioning is selected, then we force the use of `mock-` package manager.
+        # NOTE use a global variable instead?
+        tmt.package_managers.find_package_manager(
+            package_manager
+        ).probe_command = tmt.utils.Command('true')
```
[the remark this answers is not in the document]
mkoncek: It is not elegant but I don't see a better way to do it.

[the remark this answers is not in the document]
thrix: outadated

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
mkoncek: I am going to apply the suggestion.

thrix:
> @mkoncek could you update the PR description including some information regarding this new plugin?, the purpose of the PR is clear but it'd nice having more background e.g. of use cases, invocation..

@mkoncek this stands, please update the MR description which will be used for the commit message to add more detail about this change.

thrix:
> > 1. I would start mirroring the functionality found in `Command.run`. There are a few questions related to `dry` run for example. Whether we should actually enter mock and run `echo $command` or just print the cmd.
> > 2. I agree that the logging as written by me was quite ad-hoc and more for development. It does need to be unified with the rest of the sources.
> > 3. I noticed that I had to write:
> > 
> > ```
> > stdout = MockShell.Stream(lambda text: output_logger("out", text, 'yellow', level = 0))
> > stderr = MockShell.Stream(lambda text: output_logger("err", text, 'yellow', level = 0))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to get any output, I know that for example the `container` provision uses `level=2`, but I could not see the logs even when running `tmt -vvv`.
> 
> It would help collecting these improvements in an issue.

@mkoncek please create a new issue for these follow-ups, the original issue # 255 will be closed with this MR

happz:
> > > 1. I would start mirroring the functionality found in `Command.run`. There are a few questions related to `dry` run for example. Whether we should actually enter mock and run `echo $command` or just print the cmd.
> > > 2. I agree that the logging as written by me was quite ad-hoc and more for development. It does need to be unified with the rest of the sources.
> > > 3. I noticed that I had to write:
> > > 
> > > ```
> > > stdout = MockShell.Stream(lambda text: output_logger("out", text, 'yellow', level = 0))
> > > stderr = MockShell.Stream(lambda text: output_logger("err", text, 'yellow', level = 0))
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > to get any output, I know that for example the `container` provision uses `level=2`, but I could not see the logs even when running `tmt -vvv`.
> > 
> > 
> > It would help collecting these improvements in an issue.
> 
> @mkoncek please create a new issue for these follow-ups, the original issue # 255 will be closed with this MR

https://github.com/teemtee/tmt/issues/4097

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,569 @@
...
+
+            assert self.mock_shell.stdin is not None
+            self.mock_shell.stdin.write('')
+            self.mock_shell.stdin.flush()
+            self.mock_shell.communicate()
+            self.mock_shell = None
+            assert self.epoll is not None
+            self.epoll.close()
+            self.epoll = None
+            self.pid = None
```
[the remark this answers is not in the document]
mkoncek: Addressed.

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
mkoncek: I wrapped them in a context manager.

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
mkoncek: Addressed.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,608 @@
...
+            )
+            returncode = None
+
+            while self.mock_shell.poll() is None:
+                events = self.epoll.poll(timeout=timeout)
+
+                if len(events) == 0:
+                    # TODO
+                    # kill the process spawned inside the mock shell
+                    pass
```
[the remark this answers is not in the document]
mkoncek: Will implement later.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,608 @@
...
+        while True:
+            mock_shell_result = self.mock_shell.poll()
+            if mock_shell_result is not None:
+                break
+            events = self.epoll.poll()
+            if len(events) == 1 and events[0][0] == self.mock_shell_stdout_fd:
+                if not self.mock_shell.stdout.read():
+                    mock_shell_result = self.mock_shell.wait()
+                # shell is ready
+                break
```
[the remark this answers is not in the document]
mkoncek: Shell readiness should work fine.

tmt/steps/provision/mock.py
LecrisUT: Still relevant, see the following snippets for how to print the import hints
https://github.com/teemtee/tmt/blob/c91dd027733725c7ae341131ec981b40bf82dd83/tmt/steps/provision/testcloud.py#L1248-L1260
https://github.com/teemtee/tmt/blob/c91dd027733725c7ae341131ec981b40bf82dd83/tmt/steps/provision/testcloud.py#L78-L101

mkoncek:
> @mkoncek could you update the PR description including some information regarding this new plugin?, the purpose of the PR is clear but it'd nice having more background e.g. of use cases, invocation..

Updated the description, is it good enough?

mkoncek:
> @mkoncek the tests are failing can you look at them?

`rpm-build` or some other ones? Can you point me?

LecrisUT:
> > @mkoncek the tests are failing can you look at them?
> 
> `rpm-build` or some other ones? Can you point me?

For starters this needs to be resolved so the others can be viewed. This is blocked by merge conflicts

mkoncek:
sounds like...
`Automatic merge failed; fix conflicts and then commit the result.'`

tmt.spec
```diff
@@ -115,6 +115,14 @@ Requires:       python3-mrack-beaker
...
+%package -n     tmt+provision-mock
+Summary:        Dependencies required for tmt mock provisioner
+Provides:       tmt-provision-mock == %{version}-%{release}
+Requires:       tmt == %{version}-%{release}
+Requires:       mock
+
+%description -n tmt+provision-mock %_metapackage_description
+
 # Replace with pyproject_extras_subpkg at some point
 %package -n     tmt+all
```
LecrisUT, line 127: This package (`tmt+all`) is missing `Requires:       tmt+provision-mock == %{version}-%{release}`

mkoncek: Should it be a `Requires` though?

mkoncek: Right, `tmt+all`, makes sense.

tmt/package_managers/mock.py
LecrisUT: Will probably be covered by the artifact plugin, so can postpone this.

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,153 @@
...
+    def reinstall(
+        self,
+        *installables: Installable,
+        options: Optional[Options] = None,
+    ) -> CommandOutput:
+        if options is not None and options.check_first:
+            try:
+                return self.guest.execute(self.engine.check_presence(*installables))
+            except RunError:
+                pass
```
LecrisUT: The logic when it does `reinstall` is different than the `install`. If you check the `DnfEngine` one of them is guled by `&` (effectively `&&`) the other by `|` (effectively `||`).

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,153 @@
...
+    def install_debuginfo(
+        self,
+        *installables: Installable,
+        options: Optional[Options] = None,
+    ) -> CommandOutput:
+        if options is not None and options.check_first:
+            try:
+                return self.guest.execute(self.engine.check_presence(*installables))
+            except RunError:
+                pass
```
LecrisUT: This one does not seem to do a check in the `DnfEngine`, maybe also skip it here.

tmt/steps/prepare/install.py
```diff
@@ -671,6 +671,65 @@ def _install(self) -> None:
...
 
+class InstallMock(InstallBase):
+    def install_from_repository(self) -> None:
+        self.guest.package_manager.install(
+            *self.list_installables("package", *self.packages),
+            options=Options(
+                excluded_packages=self.exclude,
+                skip_missing=self.skip_missing,
+            ),
+        )
```
LecrisUT, line 683: These would be in common with `InstallDnf`. Can you comment on why it is not subclassed from there and maybe track that?

tmt/steps/prepare/install.py
```diff
@@ -671,6 +671,65 @@ def _install(self) -> None:
...
+
+        # Use both dnf install/reinstall to get all packages refreshed
+        # FIXME Simplify this once BZ#1831022 is fixed/implemented.
+
+        # mock's package manager mounts the buildroot directory, so we need to
+        # prefix the path.
+
+        filelist = [
+            PackagePath(
+                self.guest.root_path / self.package_directory.relative_to("/") / filename.name
```
LecrisUT, line 699: Why the `guest.root_path` here, wouldn't it work with arbitrary file locations? Should you instead override `prepare_install_local`?

tmt/steps/prepare/install.py
```diff
@@ -909,6 +968,18 @@ def go(
                 guest=guest,
             )
 
+        elif guest.facts.package_manager is not None and guest.facts.package_manager.startswith(
+            'mock-'
+        ):
```
LecrisUT: Could you instead move the line
```python
        elif guest.facts.package_manager is None:
            raise tmt.utils.PrepareError('Unrecognized package manager.')
```
to be the first check so that you don't have to add the `is not None` check in there?

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,741 @@
...
+                self.epoll.close()
+                self.epoll = None
+            self.mock_shell.communicate()
+            self.mock_shell = None
+            self.parent.verbose('mock', 'Exited shell.', color='blue', level=2)
+
+    def enter_shell(self) -> None:
+        command = self.parent.mock_command_prefix.to_popen()
+        command.append('-q')
+        command.append('--enable-network')
```
LecrisUT, line 130: I believe this one was desired to be configurable. Can you track it for later TODO?

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,741 @@
...
+    ) -> bool:
+        # Nothing to do.
+        self.debug(f"Doing nothing to reboot guest '{self.primary_address}'.")
+        return False
+
+    def remove(self) -> None:
+        """
+        Currently do not prune the mock chroot, that may be undesirable.
+        """
+        pass
```
LecrisUT, line 550: This should be handled by tmt itself since you could have similar issues with the other provision. Would track this.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,741 @@
...
+
+    def reboot(
+        self,
+        hard: bool = False,
+        command: Optional[Union[Command, ShellScript]] = None,
+        waiting: Optional[Waiting] = None,
+    ) -> bool:
+        # Nothing to do.
+        self.debug(f"Doing nothing to reboot guest '{self.primary_address}'.")
+        return False
```
LecrisUT: What about refreshing the shell? Would at least help clear the env vars that are accumulating there?

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,741 @@
...
+        excludes = Command()
+        permissions = Command('-m', f'{(options.chmod or 0o755):03o}')
+        for exclude in options.exclude:
+            excludes += Command(f'--exclude={exclude}')
+        source = source or self.plan_workdir
+        destination = destination or source
+        if not (self.root_path / source.relative_to('/')).exists():
+            if options.create_destination:
+                Command('mkdir', '-p', str(destination)).run(cwd=None, logger=self._logger)
+        elif source.is_dir():
```
LecrisUT: I don't think you will know `source` to query `is_dir` here.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,741 @@
...
+    ) -> None:
+        """
+        Pull content from the mock chroot via a pipe at /srv/tmt-mock/filesync.
+        For directories we use tar.
+        For files we use cp / install.
+        Compress option is ignored, it only slows down the execution.
+        """
+        options = options or tmt.steps.provision.DEFAULT_PULL_OPTIONS
+        excludes = Command()
+        permissions = Command('-m', f'{(options.chmod or 0o755):03o}')
```
LecrisUT: Not sure about always adding `-m`. Anything that could break if it is not added?

tmt/steps/provision/mock.py
mkoncek: That sounds good. But will need some time to test, will add a TODO.

tmt/steps/provision/mock.py
mkoncek: That's a good catch.

tmt/steps/provision/mock.py
mkoncek: Short answer is: I don't think a guest without a network access is useful. Does any other provisioner do anything similar? I didn't want to add another configuration flag so I just enabled it by default.
We can discuss it of course.

tmt/steps/provision/mock.py
mkoncek: These are for `install` command, i.e. when pulling a single file only. That should be a simple use-case.

tmt/steps/provision/mock.py
mkoncek: Mock environment is usually ready in the `--init` -ed state inbetween package builds. For that reason it is questionable what we really want.
I will add a TODO.

tmt/steps/prepare/install.py
mkoncek: Perhaps, `mock --pm-cmd ...` only has visibility inside the mock chroot, so we need to get the local files inside. Adding a TODO to revisit.

tmt/steps/prepare/install.py
mkoncek: I think you are right and it would ease the implementation but this requires more testing and I am not willing to do large changes now.
Adding a TODO.

tmt/steps/provision/mock.py
mkoncek: You are right, adding a condition would be better.

tmt/package_managers/mock.py
mkoncek: Will do.

tmt/package_managers/mock.py
mkoncek: Good point, will fix.

mkoncek:
Hmmm, i don't know if i should have pushed any more.

mkoncek:
I can at least say that the Maven matrix suite passed with the latest version.

psss:
@mkoncek, could you please add a short release note? I think the formulation from he pull request description nicely summarizes it.

psss:
Ah, sorry, scratch it, I got confused by the checklist.

thrix (approved):

tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/test.sh
```diff
@@ -0,0 +1,22 @@
...
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create a run directory"
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    # Split provision and prepare steps into separate processes
+    rlPhaseStartTest
+        rlRun -s "TMT_SHOW_TRACEBACK=1 tmt --feeling-safe run --id $run -vvv discover provision"
+        rlAssertGrep "NAME.*Fedora Linux" $rlRun_LOG
+        rlRun -s "TMT_SHOW_TRACEBACK=1 tmt --feeling-safe run --id $run -vvv prepare execute report"
```
psss, line 14: Why not just running all steps? Or at least including `cleanup` as well? Omitted because it currently does nothing?

tests/provision/mock/main.fmf
```diff
@@ -0,0 +1,3 @@
+tag+:
+  - provision-only
+  - provision-mock
```
psss: As we don't have a separate plan for the `mock` provision, this means that the test is not run in CI. Is that intentional? Or shall we change to `provision-virtual` until a dedicated plan is created?

psss (approved):
Seems like a good kick off! Thanks for implementing this. Quickly tested, works for me, added just two questions/suggestions, not blocking the PoC.

tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/test.sh
mkoncek: `cleanup` should have been there too. The purpose was to test things like `load` When tmt is executed multiple times. So basically all steps up to `provision` then exit then all steps following `provision`.

tests/provision/mock/main.fmf
mkoncek: I would like to have it working as part of CI. But I don't know if it will actually manage to `import mockbuild`. I would like to create the plan for mock some time.

tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/test.sh
psss: Test does not seem to be ready, fails with a traceback, disabled for now in cdfe5b21.

tests/provision/mock/main.fmf
psss: Added to the virtual provision plan in cdfe5b21.

tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/main.fmf
```diff
@@ -2,3 +2,7 @@ summary: Test Fedora Rawhide mock container
 description:
     A simple test to see if the provision does at least provide us with a
     working shell.
+
+# TODO: Failing on `assert self.mock_shell is not None`
+# Disabling for now, not to block the proof-of-concept
```
mkoncek, line 7: Ah, that's a shame. That mock shell part definitely needs a revisit.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,756 @@
...
+        source_in_chroot = self.root_path / source.relative_to('/')
+
+        if not source_in_chroot.exists():
+            if options.create_destination:
+                Command('mkdir', '-p', str(destination)).run(cwd=None, logger=self._logger)
+        elif source_in_chroot.is_dir():
+            if options.create_destination:
+                Command('mkdir', '-p', str(destination)).run(cwd=None, logger=self._logger)
+            p = self.mock_shell._spawn_command(
+                Command('tar', '-C', str(source), '-cf', '/srv/tmt-mock/filesync', '.'),
```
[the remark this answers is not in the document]
mkoncek: No, source is a path on the guest.

tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/main.fmf
psss: Yeah, but let's handle that later. The basic use case of executing all steps works fine.

mkoncek: Can you try?
```diff
diff --git a/tmt/steps/provision/mock.py b/tmt/steps/provision/mock.py
index 3cc9b476..0ef50252 100644
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -242,7 +242,8 @@ class MockShell:
         Execute the command in a running mock shell for increased speed.
         """
 
-        assert self.mock_shell is not None
+        if self.mock_shell is None:
+            self.enter_shell()
 
         if kwargs is not None and len(kwargs) != 0:
             logger.debug(
@@ -422,7 +423,7 @@ class GuestMock(tmt.Guest):
         if self.root is not None:
             self.mock_command_prefix += Command('-r', self.root)
         if self.rootdir is not None:
-            self.mock_command_prefix += Command('--rootdir', self.rootdir)
+            self.mock_command_prefix += Command('--rootdir', str(self.rootdir))
         root_path = (
             (self.mock_command_prefix + Command('--print-root-path'))
             .run(cwd=None, logger=self._logger)
```

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,756 @@
...
+            if len(events) == 1 and events[0][0] == self.mock_shell_stdout_fd:
+                if not self.mock_shell.stdout.read():
+                    mock_shell_result = self.mock_shell.wait()
+                # shell is ready
+                break
+            for fileno, _ in events:
+                if fileno == self.mock_shell_stderr_fd:
+                    for line in self.mock_shell.stderr.readlines():
+                        self.parent._logger.debug('err', line.rstrip(), 'yellow')
+
```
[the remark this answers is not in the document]
thrix: I would leave this now

tests/provision/mock/fedora-rawhide-x86_64/provision-prepare/main.fmf
thrix: @mkoncek is it needed, can it be done in #4097?

mkoncek: @thrix 
Both fixes are "nice to have", can wait for 2 weeks. I thought the mock shell fix would enable tests.

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
mkoncek: Communicating with a running shell is a whole lot of race conditions.
Even executing commands sequentially is a lot of race conditions.

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,756 @@
...
+    root: Optional[str] = None
+    rootdir: Optional[Path] = None
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.mock_command_prefix = Command('mock')
+        if self.root is not None:
+            self.mock_command_prefix += Command('-r', self.root)
+        if self.rootdir is not None:
```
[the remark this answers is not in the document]
mkoncek: Command constructor:
```
def __init__(self, *elements: RawCommandElement) -> None:
        self._command = [str(element) for element in elements]
```
Should be ok as is.

mkoncek:
Test failures seem unrelated.

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,163 @@
...
+        options: Optional[Options] = None,
+    ) -> ShellScript:
+        return self._prepare_mock_install_script('install', *installables, options=options)
+
+    def reinstall(
+        self,
+        *installables: Installable,
+        options: Optional[Options] = None,
+    ) -> ShellScript:
+        print(options)
```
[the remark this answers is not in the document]
thrix: outdated

tmt/package_managers/mock.py
```diff
@@ -0,0 +1,162 @@
...
+from tmt.utils import Command, CommandOutput, RunError, ShellScript
+
+
+class BaseMockEngine(PackageManagerEngine):
+    """
+    We use `mock --pm-cmd ...` to execute the package manager commands inside
+    the mock. Such scripts need to be executed locally and not inside the mock
+    shell. To differentiate these scripts we set a special attribute `_local`
+    of the script object.
+    """
```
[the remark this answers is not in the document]
thrix: outdated

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,552 @@
...
+    def __del__(self) -> None:
+        self.exit_shell()
+
+    def exit_shell(self) -> None:
+        if self.mock_shell is not None:
+            self.parent.verbose('Exiting mock shell')
+
+            assert self.mock_shell.stdin is not None
+            self.mock_shell.stdin.write('')
+            self.mock_shell.stdin.flush()
```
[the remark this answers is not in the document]
thrix: outadated

tmt/steps/provision/mock.py
thrix: Added to follow up issue

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,522 @@
...
+
+        loop = 2
+        while loop != 0 and self.mock_shell.poll() is None:
+            events = self.epoll.poll()
+            for fileno, _ in events:
+                if fileno == self.mock_shell_stdout_fd:
+                    loop -= 1
+                    self.mock_shell.stdout.read()
+                    break
+        self.mock_shell.stderr.read()
```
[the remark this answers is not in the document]
thrix: outdated

tmt/steps/provision/mock.py
```diff
@@ -0,0 +1,552 @@
...
+                        content = os.read(stderr_fd, 128)
+                        stream_err += content
+                        if not content:
+                            self.epoll.unregister(stderr_fd)
+                    elif fileno == returncode_fd:
+                        content = os.read(returncode_fd, 16)
+                        if not content:
+                            self.epoll.unregister(returncode_fd)
+                        else:
+                            returncode = int(content.decode('utf-8').strip())
```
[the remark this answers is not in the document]
thrix: outdated

PR #4038, opened 2025-09-08
Align `Beaker` and `Artemis` with other plugins for status update

skycastlelily:
Assisted-by: cursor

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/artemis.py
+++ b/tmt/steps/provision/artemis.py
@@ -17,7 +17,6 @@ from tmt.utils import (
     Command,
     ProvisionError,
     ShellScript,
-    UpdatableMessage,
     dict_to_yaml,
     normalize_string_dict,
     retry_session,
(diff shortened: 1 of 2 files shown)


happz:
@skycastlelily it would be nice to be more verbose in the commit message. "Align" is nice, but what does it mean? What do `artemis` and `beaker` do differently that we decided to "align" them with the rest? This kind of stuff would help reviewers to judge the intent, the functionality, whether _those_ are aligned as well...

LecrisUT (approved):
So the main change here is dropping the `UpdatableMessage`. Some better explanation as to why would be needed in the description or as a PR comment.

PR #4040, opened 2025-09-08
Add simple helpers for accessing run, plan, phase and guest workdirs

happz:
tmt codebase is plagued with `assert foo.workdir` hints for type checkers. Classes derived from `tmt.utils.Common` provide "workdir" functionality, but this workdir is `Optional`, can - and sometimes is - unset, and current type system is not enough to express guarantees about `workdir` lifetime.

On top of that, it is often uneasy to navigate the tree of plans, steps and phases, and telling what `self.parent.workdir` is is a task for stronghearted.

Patch adds a set of mixin classes, with properties that provide direct access to run, plan, step, phase, and guest workdirs. Each interesting class in our tree can include one or more of the mixins, and implement necessary access. A plugin can then reach run, plan, step and phase wordir as attributes of `self`.

And since the type hints are silent assumptions that *now* workdir will exist, and are performed by `assert` which can be disabled, these helpers shall raise an exception, to explain that *now* we expected a workdir to exist, and there was none, instead of a stingy assert exception.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -557,7 +557,8 @@ class Step(
     def step_workdir(self) -> Path:
         if self.workdir is None:
             raise GeneralError(
-                "Existence of a step workdir was presumed but the workdir does not exist."
+                f"Existence of a step '{self.step_name}' workdir"
+                " was presumed but the workdir does not exist."
             )
 
         return self.workdir
@@ -1565,7 +1566,8 @@ class BasePlugin(
     def phase_workdir(self) -> Path:
         if self.workdir is None:
             raise GeneralError(
-                "Existence of a phase workdir was presumed but the workdir does not exist."
+                f"Existence of a phase '{self.name}' workdir"
+                " was presumed but the workdir does not exist."
             )
 
         return self.workdir
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -1350,7 +1350,8 @@ class Guest(
     def guest_workdir(self) -> Path:
         if self.workdir is None:
             raise GeneralError(
-                'Existence of a guest workdir was presumed but the workdir does not exist.'
+                f"Existence of a guest '{self.name}' workdir"
+                " was presumed but the workdir does not exist."
             )
 
         return self.workdir
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -2163,7 +2163,8 @@ class Plan(
     def plan_workdir(self) -> Path:
         if self.workdir is None:
             raise GeneralError(
-                "Existence of a plan workdir was presumed but the workdir does not exist."
+                f"Existence of a plan '{self.name}' workdir"
+                " was presumed but the workdir does not exist."
             )
 
         return self.workdir


tmt/steps/__init__.py
```diff
@@ -539,6 +549,23 @@ def __init__(
...
 
+    @property
+    def plan_workdir(self) -> Path:
+        return self.plan.plan_workdir
+
+    @property
+    def step_workdir(self) -> Path:
+        if self.workdir is None:
+            raise GeneralError(
+                "Existence of a step workdir was presumed but the workdir does not exist."
```
skycastlelily: The error messages could be more specific about context:

Existence step '{self.name}' workdir was presumed but the workdir does not exist.

skycastlelily (approved):
You might also want to change 
  run_workdir
  plan_workdir
  step_workdir
  phase_workdir

to
   
 workdir_run
  workdir_plan
  workdir_step
  workdir_phase

for consistent naming with --workdir-root/_workdir_path,etc.

and Christian's comment, Otherwise,LGTM^^

happz:
> You might also want to change run_workdir plan_workdir step_workdir phase_workdir
> 
> to
> 
> workdir_run workdir_plan workdir_step workdir_phase
> 
> for consistent naming with --workdir-root/_workdir_path,etc.

I prefer the current state, i.e. `foo_workdir`, as it feels ore natural.

tmt/libraries/beakerlib.py
```diff
@@ -225,8 +225,7 @@ def fetch(self) -> None:
         try:
             library = self._library_cache[str(self)]
             # Check in case "tmt try retest" deleted the libs
-            assert self.parent.workdir
-            if not (self.parent.workdir / self.dest / self.repo).exists():
+            if not (self.step_workdir / self.dest / self.repo).exists():
```
psss:
```suggestion
            if not (self.phase_workdir / self.dest / self.repo).exists():
```

I believe this should be phase workdir as we're storing the library into the discover phase. Quick check confirms:

Before:

```
├── discover
│   ├── default-0
│   │   ├── libs
│   │   │   ├── certgen
│   │   │   │   └── certgen
│   │   │   │       ├── lib.sh
│   │   │   │       ├── main.fmf
│   │   │   │       ├── Makefile
│   │   │   │       └── runtest.sh
│   │   │   ├── httpd
│   │   │   │   └── http
│   │   │   │       ├── lib.sh
│   │   │   │       ├── main.fmf
│   │   │   │       └── runtest.sh
│   │   │   └── openssl
│   │   │       └── certgen
│   │   │           ├── lib.sh
│   │   │           └── main.fmf
│   │   └── tests
│   │       ├── main.fmf
│   │       └── test.sh
```

After:

```
├── discover
│   ├── default-0
│   │   └── tests
│   │       ├── main.fmf
│   │       └── test.sh
│   ├── libs
│   │   ├── certgen
│   │   │   └── certgen
│   │   │       ├── lib.sh
│   │   │       ├── main.fmf
│   │   │       ├── Makefile
│   │   │       └── runtest.sh
│   │   ├── httpd
│   │   │   └── http
│   │   │       ├── lib.sh
│   │   │       ├── main.fmf
│   │   │       └── runtest.sh
│   │   └── openssl
│   │       └── certgen
│   │           ├── lib.sh
│   │           └── main.fmf
```

tmt/libraries/beakerlib.py
```diff
@@ -276,8 +275,7 @@ def fetch(self) -> None:
         except (KeyError, FileNotFoundError):
             self.parent.debug(f"Fetch library '{self}'.", level=3)
             # Prepare path, clone the repository, checkout ref
-            assert self.parent.workdir
-            directory = self.parent.workdir / self.dest / self.repo
+            directory = self.step_workdir / self.dest / self.repo
```
psss: Similar here.

tmt/libraries/__init__.py
```diff
@@ -33,7 +35,7 @@ class LibraryError(Exception):
     """
 
 
-class Library:
+class Library(HasStepWorkdir):
```
psss: The library code should be adjusted as well.

psss (changes requested):
Very nice, thanks much for getting us rid of all those asserts. Much appreciated! I just spotted one unexpected change in the code which is handling libraries, otherwise looks good.

tmt/libraries/beakerlib.py
happz: Hmhm, correct, that's a mistake. I traced `parent` argument, it does lead into a plugin, but I mistook it for step itself.

tmt/steps/provision/__init__.py
```diff
@@ -1324,6 +1331,35 @@ def __init__(
...
+    def step_workdir(self) -> Path:
+        return cast(Provision, self.parent).step_workdir
+
+    # TODO: this will not work: `self.parent` is not the owning phase,
+    # but the `provision` step itself. We do not have access to the
+    # original phase.
+    #
+    # @property
+    # def phase_workdir(self) -> Path:
+    #     return cast(Provision, self.parent).phase_workdir
```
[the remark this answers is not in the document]
happz: It should be `guest_workdir`, actually. Changed in 96d2d37a.

tmt/libraries/__init__.py
happz: Removed in c5b1be5a802b35f19e03939d9ebfa52a392caa60

tmt/libraries/beakerlib.py
happz: Removed in c5b1be5a802b35f19e03939d9ebfa52a392caa60

tmt/libraries/beakerlib.py
happz: Removed in c5b1be5a802b35f19e03939d9ebfa52a392caa60

tmt/steps/__init__.py
happz: Added names in 9b5f532d

PR #4042, opened 2025-09-09
Onboard Fedora ELN containers

yselkowitz:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/images.sh
+++ b/tests/images.sh
@@ -45,8 +45,7 @@ centos-stream-10
 fedora-42
 fedora-43
 fedora-rawhide
-fedora-coreos
-fedora-eln}"
+fedora-coreos}"
 
 # For the following images we do not exercise all possible feature
 # combinations, just make sure the basic functionality works.


LecrisUT:
These would need to be hooked up to the tests, but in principle it should already support it. Is there anything we should keep an eye out when maintaining eln in those tests?

One thing that comes to mind is how should `fmf` handle the distro context matching? https://fmf.readthedocs.io/en/stable/context.html#comparison. I think we do not do any explicit handling of the aliases and the `fedora-rawhide` ordering is just accidental and would apply the same for eln

LecrisUT:
Sorry for the late response here. You are only missing adding elen to https://github.com/teemtee/tmt/blob/01d7ff0e4ba84e6ad6666cb4761f53cc33588ac5/tests/images.sh#L21-L32

I think there might be some issues like figuring what is the correct package manager on ELN since I've heard that there are still dnf4 remnants there.

happz (approved):

LecrisUT (approved):

tests/images.sh
```diff
@@ -44,7 +45,8 @@ centos-stream-10
 fedora-42
 fedora-43
 fedora-rawhide
-fedora-coreos}"
+fedora-coreos
+fedora-eln}"
```
happz: This needs to be removed, it seems our `virtual` plugin does not support `fedora-eln` VM images: https://artifacts.osci.redhat.com/testing-farm/25d94337-ebfe-48b2-b9df-0bd715545bcc/work-install7srbxzt9/plans/provision/virtual/install/execute/data/guest/default-0/tests/prepare/install-1/output.txt

thrix: Removed in https://github.com/teemtee/tmt/pull/4042/commits/9839e6639132c1a586aa09d8caf8d437db9054f2

Filed an issue for testcloud here: https://github.com/teemtee/testcloud/issues/21

thrix (approved):
LGTM

PR #4044, opened 2025-09-10
Transform `x86_64` to `x8664` for Polarion report compatibility

### LINKED ISSUE #3588: Invalid `x86_64` arch mapping in the `polarion` report plugin
--use-facts / --no-use-facts    Use hostname and arch from guest facts,
                                  also uses environment variable
                                  TMT_PLUGIN_REPORT_POLARION_USE_FACTS.

use_facts does not work as expected for arch.

(Pdb) guests[0].facts.arch
'x86_64'

Transform 'x86_64' to ''x8664'', then the arch field can be written into the polarion test run, or else it could not.

guoguojenna:
Fixes #3588

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/report/polarion.py
+++ b/tmt/steps/report/polarion.py
@@ -357,7 +357,6 @@ class ReportPolarion(tmt.steps.report.ReportPlugin[ReportPolarionData]):
                 if arch == 'x86_64':
                     arch = 'x8664'
                 testsuites_properties['polarion-custom-arch'] = arch
-
             except IndexError as error:
                 raise tmt.utils.ReportError(
                     'Failed to retrieve facts from the guest environment. '
--- a/tmt/steps/report/polarion.py
+++ b/tmt/steps/report/polarion.py
@@ -344,13 +344,20 @@ class ReportPolarion(tmt.steps.report.ReportPlugin[ReportPolarionData]):
             # TODO: remove the os.getenv when envvars in click work with steps in plans as well
             # as with steps on cmdline
             if param:
+                # Transform x86_64 to x8664 for Polarion compatibility
+                if tr_field == 'arch' and param == 'x86_64':
+                    param = 'x8664'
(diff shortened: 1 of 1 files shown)


psss:
While the change sounds to be straightforward, we should ensure we do not break other users by this. Copy pasting a valid [comment](https://github.com/KwisatzHaderach) from @KwisatzHaderach from the original issue:

> I don't really mind changing, but just to be aware that this is not without the risk of someone else creating another issue later on, asking for exactly the opposite.

Hmmm, this is actually a good point. So... it is not a Polarion limitation that `_` cannot be used in the enum value but just an arbitrary custom project setting?

psss:
> Hmmm, this is actually a good point. So... it is not a Polarion limitation that `_` cannot be used in the enum value but just an arbitrary custom project setting?

@guoguojenna, is this just a single project setting? Or a common way how `arch` is configured across Polarion projects?

guoguojenna:
> > Hmmm, this is actually a good point. So... it is not a Polarion limitation that `_` cannot be used in the enum value but just an arbitrary custom project setting?
> 
> @guoguojenna, is this just a single project setting? Or a common way how `arch` is configured across Polarion projects?

Hi @psss, I check 3 projects, all use x8664.  I think Polarion should maintain a consistent definition for fields like architecture across all projects. Maybe we can probably merge this fix now, and then handle any special cases that arise in the future with a more robust solution. Thank you!

psss:
Thanks for checking the other projects.

> Maybe we can probably merge this fix now, and then handle any special cases that arise in the future with a more robust solution. Thank you!

Yeah, makes sense to me.

psss:
The only error is a random connection issue:

    503 Server Error: Service Temporarily Unavailable for url: https://src.fedoraproject.org/repo/pkgs/rpms/tmt

Everything else green, merging.

Issue #4033, opened 2025-09-05
beaker plugin: include key, public-key and order to schema


bgoncalv:
```
[{'name': 'server', 'how': 'beaker', 'image': '$@{distro}', 'key': '~/.ssh/id_rsa', 'public-key': '~/.ssh/id_rsa.pub', 'role': 'server', 'hardware': {'hostname': 'dell-per750-56.rhts.eng.pek2.redhat.com'}, 'order': 1}] is not valid under any of the given schemas
```

vaibhavdaren:
[Discussion] I've noticed an increase in lint issues recently. What’s the best approach for handling changes like this in the future, particularly before elevating warnings to errors? Do we have access to user TMT plans to run lint checks on all of them (e.g., focusing on top users) and identify failures? Alternatively, should we consider double-verifying the schemas?

happz:
> [Discussion] I've noticed an increase in lint issues recently. What’s the best approach for handling changes like this in the future, particularly before elevating warnings to errors? Do we have access to user TMT plans to run lint checks on all of them (e.g., focusing on top users) and identify failures? Alternatively, should we consider double-verifying the schemas?

The long-term goal is to generate schema from plugin data classes, which would make them up-to-date all the time.

We do not have easy access to user plans, there is quite a lot of them. Double-checking all changes certainly makes sense in any case.

PR #4048, opened 2025-09-11
Include `key`, `order`, and `public-key` in beaker provision schema

tcornell-bus:
- Add `key` and `order`.
- Update `public-key` to allow a single string.
- Add `tmt lint` test for beaker provision schema. 

Fixes: #4033 

Pull Request Checklist

* [x] modify the json schema

### DIFF
--- a/tests/provision/beaker/data/main.fmf
+++ b/tests/provision/beaker/data/main.fmf
@@ -14,13 +14,13 @@
                  processors: 6
             - or:
                - device:
-                   driver: ahci
+                   - driver: ahci
                - device:
-                   driver: uhci
+                   - driver: uhci
                - disk:
-                   driver: foo
+                   - driver: foo
                - disk:
-                   driver: bar
+                   - driver: bar
 
 /plan/bootc:
     provision+:
@@ -28,3 +28,29 @@
         bootc-image-url: quay.io/fedora/custom-bootc:latest
         bootc-registry-secret: dummysecret
         image: fedora-42
+
+/plan/schema1:
+    provision+:
+      - name: server1
+        image: $@{distro}
+        key: ~/.ssh/key
+        public-key: ~/.ssh/key.pub
+        role: server
+        hardware:
+            hostname: my_host2
+        order: 1
+
+/plan/schema2:
+    provision+:
+      - name: server2
+        image: $@{distro}
+        key:
+            - ~/.ssh/key1
+            - ~/.ssh/key2
+        public-key:
+            - ~/.ssh/key1.pub
+            - ~/.ssh/key2.pub
+        role: server
+        hardware:
+            hostname: my_host2
+        order: 2
--- a/tests/provision/beaker/data/main.fmf
+++ b/tests/provision/beaker/data/main.fmf
@@ -14,13 +14,13 @@
                  processors: 6
             - or:
                - device:
-                   - driver: ahci
+                   driver: ahci
                - device:
-                   - driver: uhci
+                   driver: uhci
                - disk:
-                   - driver: foo
+                   driver: foo
                - disk:
-                   - driver: bar
+                   driver: bar
 
 /plan/bootc:
     provision+:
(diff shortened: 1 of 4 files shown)


tests/provision/beaker/data/main.fmf
```diff
@@ -14,17 +14,43 @@
                  processors: 6
             - or:
                - device:
-                   driver: ahci
+                   - driver: ahci
```
happz: I don't believe this is correct, `device` is a mapping, `driver` a key in that mapping. `driver: ahci` was correct, `- driver: ahci` would be an item in a sequence labeled `driver`, which is not expected.

https://tmt.readthedocs.io/en/stable/spec/hardware.html#device

AthreyVinay: +1

tcornell-bus: Hmm, the linter is not happy with how it was. I could do one of two things. 

1. I can update the `tmt/schemas/provision/hardware.yaml` too. A change to allow for single elements seems to fix the issue. Does this seem like the right way to fix?
```
--- a/tmt/schemas/provision/hardware.yaml
+++ b/tmt/schemas/provision/hardware.yaml
@@ -414,10 +414,14 @@ definitions:
         "$ref": "#/definitions/cpu"
 
       device:
-        "$ref": "#/definitions/devices"
+        oneOf:
+          - "$ref": "#/definitions/device"
+          - "$ref": "#/definitions/devices"
 
       disk:
-        "$ref": "#/definitions/disks"
+        oneOf:
+          - "$ref": "#/definitions/disk"
+          - "$ref": "#/definitions/disks"
 
       gpu:
```
2. Or I can target only the `schema` plans, so the linter doesn't complain about the `hardware` plan.

What do you think @happz @AthreyVinay ?

happz: Hmmmm, `devices` seems like an unfinished idea: it's implemented as a mapping, not a list, and does not expect a list of devices, but I think it's supposed to be also a list, similarly to `disk` which is a list for sure. We certainly do want to keep `disk` a list, i.e. no `"#/definitions/disk"`, no alternatives.

Allowing a list will not work, now, because the implementation would not accept it. That can be fixed, of course.

tcornell-bus: Ahhh, ok. It sounds like the hardware schema updates should be in their own PR. I vote to do option 2, and I can make an issue to investigate/fix these hardware schema definitions and their implementations. How does that sound?

happz: Sounds good to me. There is more to fix than just key, order and public-key in Beaker schema.

AthreyVinay: Agree needs to be a list in the schema and "for me" (from what I have understood until now) option1 looks ok.

Longer term fix: Schema changes to handle list of devices and testing it :)

happz: > Agree needs to be a list in the schema and "for me" (from what I have understood until now) option1 looks ok.
> 
> Longer term fix: Schema changes to handle list of devices and testing it :)

The problem is it's already a list in schema, which is not accepted by implementation. Or my mind, TBH, I forgot there for the idea for `device` to contain multiple devices...

psss:
@tcornell-bus, could you please rebase on the latest `main`? Seems there is now also included commit from #4035 which has been already merged.

tcornell-bus:
Trying to fix now, I tried rebasing, which resulted in this, apparently I did it wrong :sweat_smile:

tests/provision/beaker/schema.sh
```diff
@@ -0,0 +1,19 @@
...
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "pushd data"
+        rlRun "set -o pipefail"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Check that beaker schema contains correct keys"
+        rlRun -s "tmt lint /plan/schema"
```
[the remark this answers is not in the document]
tcornell-bus: It finds both plans just fine

tests/provision/beaker/data/main.fmf
tcornell-bus: Addressed option 2 in bea6511cc177f18e6d57bb775ef52dac2571e505 and created an [issue](https://github.com/teemtee/tmt/issues/4054) for hardware schema.

PR #4053, opened 2025-09-12
Enable the `avc` check for `centos-stream` as well

psss:
The fresh `container-selinux` with the `avc` fix should now be available for both `centos-stream-9` and `centos-stream-10`:

* https://gitlab.com/redhat/centos-stream/rpms/container-selinux/-/merge_requests/151
* https://gitlab.com/redhat/centos-stream/rpms/container-selinux/-/merge_requests/150

Let's enable the `avc` check across all supported distros.

### DIFF
--- a/tests/main.fmf
+++ b/tests/main.fmf
@@ -10,13 +10,6 @@ environment:
 
 adjust:
   - check: []
-    when: distro == centos-stream
-    because:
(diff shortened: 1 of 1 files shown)


psss:
Simple, nobrainer, re-enabling disabled test coverage, adding to `1.58`.

PR #4059, opened 2025-09-12
Get rid of the `awk` dependency on guests

psss:
Currently we're using `awk` for parsing text on just a couple of places. Seems that we could replace them quite easily with bash built-ins and standard utilities. Let's get rid of one unnecessary dependency.

Generated-by: Cursor

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -15,6 +15,10 @@ The Guide has been extended with a new :ref:`guest-preparation`
 section which covers :ref:`minimal-requirements` for guests and
 describes :ref:`helper-scripts` used for special test actions.
 
+Searching a Test Case in Polarion during test export/import or
+test run export will no longer find test cases that are set as
+inactive.
+
 :ref:`/plugins/execute` no longer require ``awk`` to be installed
 on the guest, as all ``awk`` usage has been replaced with Bash
 built-ins and standard utilities.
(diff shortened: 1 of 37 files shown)


docs/releases.rst
```diff
@@ -4,6 +4,14 @@
     Releases
 ======================
 
+tmt-1.58.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`/plugins/execute` no longer require ``awk`` to be installed
```
happz, line 22:
```suggestion
:ref:`/plugins/execute` no longer requires ``awk`` to be installed
```

docs/releases.rst
```diff
@@ -4,6 +4,14 @@
...
+on the guest, as all awk usage has been replaced with bash
```
```suggestion
on the guest, as all ``awk`` usage has been replaced with Bash
```

docs/releases.rst
psss: This actually renders into:

    Execute Plugins no longer require awk to be installed 

So should be good.

docs/releases.rst
psss: Fixed in 2d8e3066.

vaibhavdaren (approved):

psss:
Also dropped mention about `awk` included in the new `Guest Preparation` section: b6216d21

happz (approved):

PR #4111, opened 2025-09-25
Rename artifact provider and artifact info sources

happz:
* The `_provider` bit in `providers/*_provider.py` seems superfluous, we already know what we have here are artifact providers.
* `ArtifactInfo` base class belongs to `__init__.py`, `RpmArtifactInfo` belongs to one of the plugins - picking `koji` for now. Might be moved later, but `info.py` defining provider-specific classes is not right.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/unit/artifact/test_base.py
+++ b/tests/unit/artifact/test_base.py
@@ -1,8 +1,7 @@
 import re
 from unittest.mock import MagicMock
 
-from tmt.steps.prepare.artifact.providers import ArtifactProvider
-from tmt.steps.prepare.artifact.providers.info import ArtifactInfo
+from tmt.steps.prepare.artifact.providers import ArtifactInfo, ArtifactProvider
 
 
 class MockArtifactInfo(ArtifactInfo):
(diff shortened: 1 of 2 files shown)


tmt/steps/prepare/artifact/providers/koji.py
```diff
@@ -54,6 +55,32 @@ def import_koji(logger: tmt.log.Logger) -> None:
...
+    @property
+    def id(self) -> str:
+        """A koji rpm identifier"""
+        return f"{self._raw_artifact['nvr']}.{self._raw_artifact['arch']}.rpm"
+
+    @property
+    def location(self) -> str:
+        """Get the download URL for the given RPM metadata."""
+        return (
+            f"{self.PKG_URL}{self._raw_artifact['name']}/"
```
AthreyVinay, line 76: There's a missing `/` after `f"{self.PKG_URL}` that I have recently added. Otherwise LGTM

AthreyVinay:
Sorry, just realized the names in tests need to be refactored to pass, thanks.

vaibhavdaren (approved):

psss (approved):

PR #4116, opened 2025-09-26
Add 'Local' distgit handler

lukaszachy:
When sources are not yet uploaded to lookaside but one wants to run dist-git-source test locally.

I missed this feature, but it might be a niche use case specific to our team workflow.  Feel free to object that one can just push sources to lookaside and be happy.

If makes sense as feature:
- Is `shutil` usage OK or should I use `tmt.utils.filesystem.copy_tree` ?
- Should plugin return absolute path to the source file instead of `./` ? 
- Running `TMT_PLUGIN_DISCOVER_SHELL_DIST_GIT_TYPE=local tmt run` didn't enforce this plugin, I had to write it into plan.fmf. Is that expected limitation (IIRC `how`  was a bit special) or should I look into this as bug?
```
 discover:
  how: shell
  dist-git-source: true
  dist-git-type: local
```
- Do you foresee another usage for "file://" protocol? If so it might be better to add support directly into `retry_session`. 
 

Pull Request Checklist

* [] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/utils/git.py
+++ b/tmt/utils/git.py
@@ -5,6 +5,7 @@ Test Metadata Utilities
 import functools
 import os
 import re
+import shutil
 import subprocess
 import urllib.parse
 from re import Pattern
@@ -640,6 +641,24 @@ class RedHatDistGit(DistGitHandler):
     remote_substring = re.compile(r'redhat/rhel/|pkgs\.devel\.redhat\.com')
 
 
+class LocalDistGit(DistGitHandler):
+    """
+    Local source files
+    """
+
+    usage_name = "local"
+    re_source = re.compile(r"^(\w+) \(([^)]+)\) = ([0-9a-fA-F]+)$")
+    lookaside_server = "file://"
+    uri = "./{filename}"
+
+    def its_me(self, remotes: list[str]) -> bool:
+        """
+        Has to be an explicit request for this type
+        """
+
+        return False
+
+
 def get_distgit_handler(
     remotes: Optional[list[str]] = None,
     usage_name: Optional[str] = None,
@@ -696,11 +715,15 @@ def distgit_download(
 
     for url, source_name in handler.url_and_name(distgit_dir):
         logger.debug(f"Download sources from '{url}'.")
-        with tmt.utils.retry_session() as session:
-            response = session.get(url)
-        response.raise_for_status()
-        target_dir.mkdir(exist_ok=True, parents=True)
-        (target_dir / source_name).write_bytes(response.content)
+        if url.startswith('file://'):
+            target_dir.mkdir(exist_ok=True, parents=True)
+            shutil.copy(url[7:], target_dir / source_name)
+        else:
+            with tmt.utils.retry_session() as session:
+                response = session.get(url)
+            response.raise_for_status()
+            target_dir.mkdir(exist_ok=True, parents=True)
+            (target_dir / source_name).write_bytes(response.content)
 
 
 def git_clone(


tmt/utils/git.py
```diff
@@ -640,6 +641,24 @@ class RedHatDistGit(DistGitHandler):
...
 
+class LocalDistGit(DistGitHandler):
+    """
+    Local source files
+    """
+
+    usage_name = "local"
+    re_source = re.compile(r"^(\w+) \(([^)]+)\) = ([0-9a-fA-F]+)$")
+    lookaside_server = "file://"
+    uri = "./{filename}"
```
LecrisUT: Hmm, this seems ill-defined. Can it be more specific navigating through the fmf tree / plan_workdir (+ path)?

LecrisUT (commented):
Looks fine to me to add, just would like it to be more concrete what path it uses there.

tmt/utils/git.py
lukaszachy: plugin doesn't seem to have access to tree/workdr, but one level above it detects path to the 'spec' file. I'm using that in 806dca54 and storing absolute path here.

lukaszachy:
added docs, schema. IMO good for a real review.

lukaszachy:
hm, so apparently precommit doesn't run when one rebases and resolves conflicts :/

tmt/utils/git.py
```diff
@@ -697,7 +717,11 @@ def distgit_download(
 
     for url, source_name in handler.url_and_name(distgit_dir):
         target_dir.mkdir(exist_ok=True, parents=True)
-        tmt.utils.url.download(url, target_dir / source_name, logger=logger)
+        if url.startswith('file://'):
+            # Cannot use copy_tree src is a file, not a dir
+            shutil.copy(url[7:], target_dir / source_name)
```
LecrisUT: For a future implementation, would like to avoid running these with bare `shutil` or equivalents. It would be better to go through the guest/runner to make it more clear where it is supposed to be run from/at.

LecrisUT: Can we constify the `file://` and avoid the magic number?

LecrisUT (approved):
Lgtm, only 1 minor suggestion

tmt/utils/git.py
lukaszachy: I'm not sure what you mean by  go through the guest/runner. The above ` tmt.utils.url.download` doesn't do such thing neither, does it?

lukaszachy: Done in a1c7378e

LecrisUT: Yeah, it's it's unrelated to the current implementations, opened #4292 to better track that.

tmt/utils/git.py
```diff
@@ -30,6 +31,9 @@
     import tmt.base
 
 
+FILE_PROTOCOL = 'file://'
```
happz: In URI terms, `file` bit is called "schema".

lukaszachy: d7d3f10a

happz (approved):

psss:
Time and unrelated failure:

    Error: unable to copy from source docker://registry.access.redhat.com/ubi9-init:latest: copying
    system image from manifest list: Source image rejected: None of the signatures were accepted,
    reasons: missing dev.sigstore.cosign/bundle annotation; missing dev.sigstore.cosign/bundle
    annotation

PR #4117, opened 2025-09-26
Add ids all tests, not just finished

### LINKED ISSUE #3784: Polarion report missing properties for incomplete tests
Tests can now be reported as incomplete, this is problematic for Polarion report plugin which will be missing properties which are the only thing parsing test cases executed to the ones in Polarion, resulting in not having these results in Polarion at all and run appearing as full passed.

This is most likely due to a missed part of execute that is filling necessary data into the test case objects for polarion report plugin to use for searching in Polarion.
These test cases also need to be marked as blocked in the report xml.

KwisatzHaderach:
Add ids for all results upon creation to have identifiable information even for skipped/pending tests.

Fixes #3784 

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -15,6 +15,7 @@ import fmf.utils
 
 import tmt
 import tmt.base
+import tmt.identifier
 import tmt.log
 import tmt.steps
 import tmt.steps.scripts
@@ -25,7 +26,9 @@ from tmt.checks import Check, CheckEvent, CheckPlugin
 from tmt.container import container, field, simple_field
 from tmt.options import option
 from tmt.plugins import PluginRegistry
-from tmt.result import CheckResult, Result, ResultGuestData, ResultInterpret, ResultOutcome
+from tmt.result import (
+    EXTRA_RESULT_IDENTIFICATION_KEYS, CheckResult, Result, ResultGuestData,
+    ResultIds, ResultInterpret, ResultOutcome)
 from tmt.steps import Action, ActionTask, PhaseQueue, PluginTask, Step
 from tmt.steps.abort import AbortContext, AbortStep
 from tmt.steps.discover import Discover, DiscoverPlugin, DiscoverStepData
@@ -1246,6 +1249,14 @@ class Execute(tmt.steps.Step):
             if not test.where:
                 result_guests = guests
 
+            # Saving identifiable information for each test case so we can match them
+            # to Polarion/Nitrate/other cases and report run results there
+            # TODO: would an exception be better? Can test.id be None?
+            ids: ResultIds = {tmt.identifier.ID_KEY: test.id}
+            for key in EXTRA_RESULT_IDENTIFICATION_KEYS:
+                value: Any = cast(Any, test.node.get(key))
+                ids[key] = None if value is None else str(value)
+
             results.extend(
                 [
                     Result(
@@ -1253,6 +1264,7 @@ class Execute(tmt.steps.Step):
                         serial_number=test.serial_number,
                         result=tmt.result.ResultOutcome.PENDING,
                         fmf_id=test.fmf_id,
+                        ids=ids,
                         guest=ResultGuestData(
                             name=guest[0],
                             role=guest[1],


tcornell-bus:
I am not seeing the added test in the [testing-farm run](https://artifacts.dev.testing-farm.io/052b26d5-8226-4c0a-b113-8fb3947f6967/#work-executeqkq0qz68_plans-features-steps-execute), shouldn't it be run in the full suite?

tests/execute/pending-ids/test.sh
```diff
@@ -0,0 +1,23 @@
...
+        rlAssertGrep 'pass /test/1-passed' $rlRun_LOG '-F'
+        rlAssertGrep 'errr /test/2-error' $rlRun_LOG '-F'
+        rlAssertGrep 'pending /test/3-pending' $rlRun_LOG '-F'
+        rlAssertGrep '86ef8a3d-bb6c-4c6e-86bb-1751ab8da302' ${run}/plan/execute/results.yaml
+        rlAssertGrep '2af20892-8474-4cf9-ba8c-b9bd1cca5078' ${run}/plan/execute/results.yaml
+        rlAssertGrep '5435a332-3b68-4bd4-a614-ac5b605ae346' ${run}/plan/execute/results.yaml
+    rlPhaseEnd
+
+    rlPhaseStartCleanup
+        rlRun "popd"
```
tcornell-bus, line 25:
```suggestion
        rlRun "popd"
        rlRun "rm -r $run" 0 "Remove the run directory"
```

It would be nice if the run dir was removed at the end.

tests/execute/pending-ids/data/main.fmf
```diff
@@ -0,0 +1,15 @@
+/test:
+    test: shutdown -h now && sleep 60
+    duration: 5s
+    /1-passed:
+        test: 'true'
+        id: 86ef8a3d-bb6c-4c6e-86bb-1751ab8da302
+    /2-error:
+        id: 2af20892-8474-4cf9-ba8c-b9bd1cca5078
+    /3-pending:
+        id: 5435a332-3b68-4bd4-a614-ac5b605ae346
```
tcornell-bus, line 10: Could we extend to have one of the tests include the two EXTRA_RESULT_IDENTIFICATION_KEYS? And grep for them in the test as well?

```suggestion
        id: 5435a332-3b68-4bd4-a614-ac5b605ae346
        extra-nitrate: TC#0000000
        extra-task: /my/test/path
```

tcornell-bus, line 10: A test where there is no `id` defined in the fmf would be great as well. Just to make sure the output results.yaml is what you are expecting.

KwisatzHaderach:
The test is part of the provision suite only, since it needs to be run on virtual. Thanks for the suggestions.

docs/releases.rst
```diff
@@ -19,6 +19,9 @@ New :ref:`sprints` section has been added to the documentation for
 contributors. It describes the overall sprint process, the key
 sync points and provides links to progress tracking boards.
 
+Pending tests are now also exported to Polarion thanks to no
+longer missing test case ids.
```
tcornell-bus:
```suggestion
longer missing test case ids in the ``results.yaml``.
The ``results.yaml`` file now contains test ``ids``, such
as ``id``, ``extra-nitrate``, and ``extra-task``.
```

KwisatzHaderach: The fact that it contains those values is not something new, that was there for couple years now. I can add the mention of location of those ids, that part makes sense to add.

tcornell-bus: My mistake, thanks for clarifying :slightly_smiling_face:

tmt/steps/execute/__init__.py
```diff
@@ -1246,13 +1255,22 @@ def create_results(self, tests: list['tmt.steps.discover.TestOrigin']) -> list['
             if not test.where:
                 result_guests = guests
 
+            # Saving identifiable information for each test case so we can match them
+            # to Polarion/Nitrate/other cases and report run results there
+            # TODO: would an exception be better? Can test.id be None?
```
happz: Yes, `Test.id` can absolutely be `None`.

KwisatzHaderach: this is just copied over from the other place this code is at, exception would break it for everyone not having ids, I think this is mostly a rhetorical question, not sure we still need the TODO :/

happz: Ah, I see, it's a copy & paste, just with removed empty lines... Hm, would it make sense to refactor the code, to move these four lines into a function or method or property, e.g. `Test.result_ids`, or `Test.default_result_ids`, returning `ResultIds` mapping, so this would happen at one place other places can call? Because now we have two copies of an action, not same, just similar, but with the same comments... I'd argue it'd harder to maintain now.

KwisatzHaderach: guess we can implement it in Test object and just call that with `TestInvocation.test.get_ids()` in result and `Test.get_ids()` here

tmt/steps/execute/__init__.py
```diff
@@ -1248,13 +1255,18 @@ def create_results(self, tests: list['tmt.steps.discover.TestOrigin']) -> list['
             if not test.where:
                 result_guests = guests
 
+            # Saving identifiable information for each test case so we can match them
+            # to Polarion/Nitrate/other cases and report run results there
+            ids: ResultIds = test.get_ids()
```
happz: `ids` as a name does not need to exist, `ids=test.get_ids()` below would work as well. It's also possible to add a comment if you wish:

```python
                        # Saving identifiable information for each test case so we can match them
                        # to Polarion/Nitrate/other cases and report run results there
                        ids=ids,
```

tmt/base.py
```diff
@@ -1574,6 +1577,17 @@ def show(self) -> None:
                 if value not in [None, [], {}]:
                     echo(tmt.utils.format(key, value, key_color='blue'))
 
+    def get_ids(self) -> tmt.result.ResultIds:
```
happz: And with my inclination to properties, would you mind making it a `property`?

```python
@property
def ids(self) -> tmt.result.ResultIds:
    ...
```

Might be also a `cached_property`, as it's established once and stays as it is for the duration of tmt run.

tmt/result.py
```diff
@@ -376,14 +372,8 @@ def from_test_invocation(
...
-        default_ids: ResultIds = {tmt.identifier.ID_KEY: invocation.test.id}
-
-        for key in EXTRA_RESULT_IDENTIFICATION_KEYS:
-            value: Any = cast(Any, invocation.test.node.get(key))
-
-            default_ids[key] = None if value is None else str(value)
+        default_ids: ResultIds = invocation.test.get_ids()
 
         default_ids.update(ids)
         ids = default_ids
```
happz: We can do something nice here:

```python
ids = {
  **invocation.test.ids,
  **(ids or {})
}
```

The update of `default_ids` and then `ids = default_ids` is really scary, and since you are in the area...

happz: Oh, and we don't even need to combine it here :)

```python
            ...
            ids={
                **invocation.test.ids,
                **(ids or {})
            },
            ...
```

PR #4131, opened 2025-09-29
Check for package existence early during prepare `install`

cgwalters:
In our test flow we have the "essential-requires" of `awk` and `flock`, but basically in the bootc
"package manager" backend we're only testing package existence in the container build which means we end up doing a no-op build.

### DIFF
--- a/tests/provision/bootc/test.sh
+++ b/tests/provision/bootc/test.sh
@@ -31,7 +31,7 @@ rlJournalStart
         rlRun -s "tmt -vvv run --scratch -i $run plan --name /plans/image/needs-deps"
         # Testing the output of the bootc package manager
         rlAssertGrep "building container image with dependencies" $rlRun_LOG
-        rlAssertGrep "STEP 1/2: FROM containers-storage:localhost/tmt/bootc" $rlRun_LOG
+        rlAssertGrep "STEP 1/2: FROM containers-storage:localhost/tmtmodified" $rlRun_LOG
         rlAssertGrep "Successfully tagged localhost/tmt/bootc/" $rlRun_LOG
         rlAssertGrep "switching to new image localhost/tmt/bootc/" $rlRun_LOG
         rlAssertGrep "rebooting to apply new image" $rlRun_LOG


psss:
Thanks for the patch! Just to double check: The fix in #3986 does not cover your scenario fully? @thrix, I thought the main intention of that pull request was exactly fixing this problem, including installing essential requires.

cgwalters:
Yes, it didn't work for me. The reproducer is booting from a disk image that starts out in image mode, which is how our CI works: https://github.com/bootc-dev/bootc/blob/8cf097171969f6fbee9a82225789c041b29221ff/tmt/plans/integration.fmf#L4
(And it's how RHELAI's works)

cgwalters:
The bootc base images we ship all have `awk` and `flock` (the latter is part of `util-linux-core` which would be very unusual not to include).

Now bigger picture of course it would align better with image mode if we didn't include lots of runtime requirements (especially for example python) as this makes it easier to test smaller systems. Ultimately the way to do that is to move to having tests be executed as (possibly privileged) container images, or low-depenency binaries (e.g. go/rust, but also python can be bundled with pyinstaller and tools like that).

I would probably recommend trying to drop the hard requirement on `awk` in that vein and make it something that tests have to pull in if they need.'

But anyways let's try to fix this unnecessary install+reboot as it's an extremely large hit to edit-compile-test iteration speed.

happz:
> I would probably recommend trying to drop the hard requirement on `awk` in that vein and make it something that tests have to pull in if they need.'

Dropped in tmt 1.58, https://github.com/teemtee/tmt/pull/4059

thrix:
Looks like really an incomplete fix to me :face_exhaling:  I do not see the check for essential requirements done on the booted host, what is a surprise, `rsync` obviously is checked. So yeah ....

And that means also the test not checking for this correctly ...

thrix:
Added to next release, but maybe even this will be possible.

thrix:
@cgwalters right, the #3986 was not at all fixing this, it was only addressing the unexpected container rebuild in case of `tmt run ... --skip prepare ...` what RHEL AI is using to make sure not preparations are applied, including essential requirements. I will extend this MR with tests covering your use case and recheck if this is the only place which needs fixing.

cgwalters:
I will say I think significant cleanup is possible in all of this code.

> I do not see the check for essential requirements done on the booted host, what is a surprise, rsync obviously is checked.

Right I think moving `rsync` into working the same as an essential requires would make sense and ensure there's just *one* bootstrap install run once very early on would be good.

thrix:
> I will say I think significant cleanup is possible in all of this code.

Agreed, I will add that as a discussion topic and move this forward as a bugfix as it resolves the problem for me.
I have updated the test and it verifies no container rebuild happening anymore.

thrix (approved):
Confirmed the changes fix the problem and no container rebuild is happening if the essential requirements are installed on the machine.

I updated the tests and should be passing now and test that the changes have the effect they should have.

psss:
Seems there are unexpected test failures, e.g. installing an [invalid package does not fail](https://artifacts.osci.redhat.com/testing-farm/355437cd-edd7-4db3-9eab-786136fae8e4/work-preparehv74lpli/plans/provision/virtual/prepare/execute/data/guest/default-0/tests/prepare/install-3/output.txt).

thrix:
damn it, tests need more fixes obviously, 1.60 it is :(

psss:
> damn it, tests need more fixes obviously, 1.60 it is :(

Just to make it clear: This is not just about adjusting tests, there seems to be something wrong with the logic, because when an invalid package is requested the `prepare` step happily reports everything is ok:

```yaml
provision:
    how: virtual
    image: fedora-coreos
prepare:
    how: install
    package: something-weird
execute:
    how: tmt
    script: echo hi
```

Here's the output:

```
prepare
    how: install
    name: essential-requires
    order: 30
    where: default-0
    package: 1 package requested
        /usr/bin/flock
        cmd: rpm -q --whatprovides /usr/bin/flock
        out: util-linux-core-2.40.4-7.fc42.x86_64
    how: install
    order: 50
    package: 1 package requested
        something-weird
        cmd: rpm -q --whatprovides something-weird
        out: no package provides something-weird
```

The problem seems to be limited to `fedora-coreos` though, for regular `fedora` everything works as expected.

thrix:
Finally found the issues after debugging the failed tests -  `installables` is a generator, the original code tries to iterate
   through the generator twice, the second try will have an empty list.

With that also fix these issues [discussed on slack](https://redhat-internal.slack.com/archives/C08J5QPSSKH/p1761034842364259):
    
1. checking `installables` emptiness is not needed (not a bug, but an improvement)
   
2. use presence dict to iterate through to mitigate the need to convert the generator or a list
    
3. use a `set` instead of a `list`, it is fine that the `installables` are unique 
    
4. fix the original code from #3986 which has the same problem with the generator

I am waiting for the tests and it is expected some additional adjustments will be needed

thrix:
@cgwalters can you try it out? it should be in latest tmt copr repo now:

https://copr.fedorainfracloud.org/coprs/g/teemtee/latest

PR #4144, opened 2025-10-02
Use `to_yaml` filter when changing complex objects in policies

happz:
Without the filter, some special Python values may produce unexpected values when modified by a policy. Namelt this has been seen with `None` which gets stringified and becomes `"None"`.

Reported in https://issues.redhat.com/browse/TFT-3928

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/spec/policy.fmf
+++ b/spec/policy.fmf
@@ -102,7 +102,7 @@ description: |
           # Incorrect outcome when `unit` key is not set
           - {{ check }}
 
-          # [{"how": journal", ..., "unit": "None"}]
+          # [{"how": "journal", ..., "unit": "None"}]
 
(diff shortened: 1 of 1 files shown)


skycastlelily (approved):
LGTM^^, you might also want to update contact.yaml and environment.yaml to use consistent to_yaml rendering

happz:
> LGTM^^, you might also want to update contact.yaml and environment.yaml to use consistent to_yaml rendering

Those two are different, their "items" are trivial objects, strings and string key/value pairs, while `check` items are complex objects. It's easy to enforce the right indentation for trivial objects.

happz:
Unrelated Fedora CoreOS image issues, merging.

PR #4164, opened 2025-10-08
Use `BeakerLib Libraries` for the library examples

### LINKED ISSUE #1884: Better name for the test libraries examples section
A test can specify an external GIT URL in https://tmt.readthedocs.io/en/stable/spec/tests.html#require such as
```yaml
require:
  - url: https://github.com/redhat-qe-security/certgen
    ref: devel
    nick: openssl
    name: /certgen
```
Since the documentation doesn't state this explicitly, I'm assuming that
* `name` is a subdirectory inside the cloned `url` (FMF object name)
* `nick` is the destination name of this subdirectory, as seen by the test

The problem is that the test has no idea where to find this FMF object retrieved by TMT.

The current approach is to assume a specific internal TMT implementation and search for the required object in parent directories, which is likely not a good idea.

Please provide some language-independent way, ideally an environment variable, so the test could easily access files and directories inside the `require`.

As an example,
* `TMT_TEST_REQUIRE` could be a path to a directory, which contains directories (or symlinks) named after `nick`, or `name` where `nick` is not specified. A bash test could source like
  ```bash
  . "$TMT_TEST_REQUIRE/openssl/somefile.sh"
  ```
* Alternatively, if you want to avoid this "middle layer" of symlinks, you could hardcode variable names after `nick` (or `name`) like
  ```bash
  . "$TMT_TEST_REQUIRE_openssl/somefile.sh"
  ```

For the purposes of #507, you could further state that a FMF object (path) provided in this way must be used as read-only, which would allow you to optimize file transfers per #507, sharing requires across tests, ie. using symlinks as the first option above suggests.

---
This is probably not specific to GIT URLs as a test can `require` a local directory (FMF object) as well, which might need to be symlinked.

psss:
As suggested in #1884 the generic name seems to be a bit confusing. Let's explicitly include `BeakerLib` in the section name to make the scope obvious.

Fix #1884.

Pull Request Checklist

* [x] write the documentation

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -416,7 +416,8 @@ at https://github.com/psss/python-nitrate
 Configuration and guide for setting up pylero can be found
 at https://github.com/RedHatQE/pylero
 
-Test Libraries
+
+BeakerLib Libraries
 ------------------------------------------------------------------
 
 In order to prevent unnecessary test code duplication it makes


psss:
Documenation only, core tests passed, merging.

PR #4173, opened 2025-10-09
Refactor for mock plugin

mkoncek:
This among others fixes `Fix the /tests/provision/mock/fedora-rawhide-x86_64/provision-prepare traceback` mentioned in: https://github.com/teemtee/tmt/issues/4097

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -773,7 +773,7 @@ class ProvisionMock(tmt.steps.provision.ProvisionPlugin[ProvisionMockData]):
         This plugin requires the ``--feeling-safe`` option or
         the ``TMT_FEELING_SAFE=1`` environment variable defined.
         While it is roughly as safe as ``container`` provisioning,
-        the plugin still bind-mounts the test directory.
+        it has access to local filesystem.
 
     Using the plugin:
 
@@ -781,11 +781,11 @@ class ProvisionMock(tmt.steps.provision.ProvisionPlugin[ProvisionMockData]):
 
         provision:
             how: mock
-            config: fedora-rawhide-x86_64
+            root: fedora-rawhide-x86_64
 
     .. code-block:: shell
 
-        provision --how mock --config fedora-rawhide-x86_64
+        provision --how mock --root fedora-rawhide-x86_64
 
     .. note::
 


tmt/steps/provision/mock.py
```diff
@@ -57,55 +57,65 @@ class ProvisionMockData(MockGuestData, tmt.steps.provision.ProvisionStepData):
...
+        super().__exit__(exc_type, exc_value, exc_traceback)
+
+
+class _DecodingStream:
+    """
+    Lazy object that collects chunks of bytes and decodes them when a
+    newline is encountered.
+    """
+
+    def __init__(self, logger: Callable[[str], None]):
```
LecrisUT: Could you document and type-check the attributes (and similarly for the other class)? Particularly `logger` has a clashing meaning, maybe you can rename it to `log_func`

tmt/steps/provision/mock.py
```diff
@@ -57,55 +57,65 @@ class ProvisionMockData(MockGuestData, tmt.steps.provision.ProvisionStepData):
     pass
 
 
+class _ManagedEpollIo(io.FileIO):
+    """
+    Context manager for file descriptors registered with epoll.
+    """
```
LecrisUT, line 80: Could you add a reference for epoll documentation. This part is quite niche that a reference would help.

tmt/steps/provision/mock.py
```diff
@@ -556,10 +567,14 @@ def _do_scrub(self) -> None:
...
     def suspend(self) -> None:
-        self.mock_shell.exit_shell()
+        self.stop()
+
+    def start(self) -> None:
+        self.mock_shell.enter_shell()
 
     def stop(self) -> None:
         self.mock_shell.exit_shell()
+        self.run(Command('rm', '-rf', str(self.root_path / 'srv/tmt-mock/*')))
```
LecrisUT: Could you parametrize/constify `srv/tmt-mock` and add a comment about why the rm command is run here?

tmt/steps/provision/mock.py
```diff
@@ -239,7 +256,9 @@ def _spawn_command(
         **kwargs: Any,
     ) -> Generator[tuple[str, str]]:
         """
-        Execute the command in a running mock shell for increased speed.
+        Spawn a command in the shell. The first call to `next` will start
```
```suggestion
        Spawn a command in the shell. The first call to ``next`` will start
```
It's RST formatted. I think this is where the CI has failed.

tmt/steps/provision/mock.py
LecrisUT: Close, here is the preferred way of documentation:
```python
class _DecodingStream:
    """
    Lazy object that collects chunks of bytes and decodes them when a
    newline is encountered.
    """
    #: Function receiving the decoded text line-by-line soon as it is available.
    log_func: Callable[[str], None]
    #: ...
    output: bytes
    #: ...
    string: str

    def __init__(self, log_func: Callable[[str], None]):
        """
        :param log_func: Function receiving the decoded text line-by-line as
            soon as it is available.
        """

        self.log_func: Callable[[str], None] = log_func
        self.output: bytes = b''
        self.string: str = ''
```
Documenting the constructor is usually not needed, but for non-container/non-dataclass I am fine with it. The docstring for `log_func` could use more clarity: it receives the stdout/stderr? from inside the mock and does ...?

tmt/steps/provision/mock.py
```diff
@@ -203,17 +235,17 @@ def enter_shell(self) -> None:
...
-            ' /srv/tmt-mock/filesync'
+            f' /{MOCK_PIPE_STDOUT_STEM}'
+            f' /{MOCK_PIPE_STDERR_STEM}'
+            f' /{MOCK_PIPE_RETURNCODE_STEM}'
+            f' /{MOCK_PIPE_FILESYNC_STEM}'
             '\n'
         )
-        self.mock_shell.stdin.write('chmod a+rw /srv/tmt-mock/filesync\n')
+        self.mock_shell.stdin.write(f'chmod -R a+rw /{MOCK_PIPE_STEM}\n')
         self.mock_shell.stdin.flush()
```
LecrisUT: Fine to deffer, but do you think you could generalize these, similar to how `ShellScript` is handled (or maybe reuse that completely)? It is unintuitive how the stdin translates to command execution.

tmt/steps/provision/mock.py
```diff
@@ -671,17 +705,24 @@ def pull(
...
                 + permissions
             ).run(cwd=None, logger=self._logger)
             next(p)
 
 
-@tmt.steps.provides_method('mock')
+@tmt.steps.provides_method(
+    'mock',
+    installation_hint="""
+        Make sure `mock` package is installed.
```
```suggestion
        Make sure ``mock`` package is installed.
```

tmt/steps/provision/mock.py
mkoncek: Most of the command execution is handled via `execute`. There are only 2 (?) places where we need a more primitive way of executing commands and this is one of them.
That's why i wasn't bothered with abstracting the logic. Maybe I can put more comments on it?

LecrisUT: > Maybe I can put more comments on it?

Yes please :slightly_smiling_face:

tmt/steps/provision/mock.py
LecrisUT: @mkoncek Still relevant :) (and only remaining comment from me)

mkoncek: Ok, working on it. I would prefer to have the attribue documentation on a single place, so just attributes.

LecrisUT: I am curious, does your IDE not display those, and are you using vscode? For me on PyCharm it displays them, but only if it is in the other format
Screenshot_20251112_133331

mkoncek: I am using KDE's Kate with `pylsp` language server, it usually shows types of variables.

LecrisUT: Found the relevant issue on it I think https://github.com/python-lsp/python-lsp-server/issues/372

mkoncek: Yeah, that would be nice. Anyway, is there something for me to fix further?

LecrisUT: No, I don't think so, just going to do one more final look-through

tmt/steps/provision/mock.py
```diff
@@ -57,55 +63,85 @@ class ProvisionMockData(MockGuestData, tmt.steps.provision.ProvisionStepData):
...
+    log_func: Callable[[str], None]
+    #: The content of the byte buffer accumulated by adding content to this object.
+    output: bytes
+    #: The decoded ``output``.
+    string: str
+
+    def __init__(self, log_func: Callable[[str], None]):
+        self.log_func: Callable[[str], None] = log_func
+        self.output: bytes = b''
+        self.string: str = ''
```
```suggestion
        self.log_func = log_func
        self.output = b''
        self.string = ''
```
Type hints are covered in the part above. Should not duplicate to avoid having to patch multiple places later on.

tmt/steps/provision/mock.py
```diff
@@ -57,55 +63,85 @@ class ProvisionMockData(MockGuestData, tmt.steps.provision.ProvisionStepData):
...
+    """
+
+    def __init__(self, *args: Any, epoll: 'select.epoll', **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        # `select.epoll` is not available on non-Linux platforms.
+        # The `ruff` linter complains but for no good reason, so we silence it.
+        self.epoll: 'select.epoll' = epoll  # noqa: UP037
+
+        self.registered: bool = False
```
LecrisUT: Can move type hints for this in the class level. Will leave it up to you if you want to document the attributes purpose also, seems fairly straightforward these ones imo.

tmt/steps/provision/mock.py
```diff
@@ -200,32 +266,26 @@ def enter_shell(self) -> None:
...
             'mkfifo'
-            ' /srv/tmt-mock/stdout'
-            ' /srv/tmt-mock/stderr'
-            ' /srv/tmt-mock/returncode'
-            ' /srv/tmt-mock/filesync'
-            '\n'
+            f' /{MOCK_PIPE_STDOUT_STEM}'
+            f' /{MOCK_PIPE_STDERR_STEM}'
+            f' /{MOCK_PIPE_RETURNCODE_STEM}'
+            f' /{MOCK_PIPE_FILESYNC_STEM}',
```
LecrisUT: Took me a few reads to see that this is actually only one string. Please make them 1 line or construct it outside, maybe like a 
```python
mkfifo_cmd = "mkfifo"
for mock_dir in [MOCK_PIPE_STDOUT_STEM, MOCK_PIPE_STDERR_STEM, MOCK_PIPE_RETURNCODE_STEM, MOCK_PIPE_FILESYNC_STEM]:
    mkfifo_cmd += " /{mock_dir}"
```

tmt/steps/provision/mock.py
```diff
@@ -20,6 +20,12 @@
 from tmt.utils import Command, OnProcessEndCallback, OnProcessStartCallback, Path, ShellScript
 from tmt.utils.wait import Waiting
 
+MOCK_PIPE_STEM: str = 'srv/tmt-mock'
+MOCK_PIPE_STDOUT_STEM: str = f'{MOCK_PIPE_STEM}/stdout'
+MOCK_PIPE_STDERR_STEM: str = f'{MOCK_PIPE_STEM}/stderr'
+MOCK_PIPE_RETURNCODE_STEM: str = f'{MOCK_PIPE_STEM}/returncode'
+MOCK_PIPE_FILESYNC_STEM: str = f'{MOCK_PIPE_STEM}/filesync'
```
LecrisUT: I think you should make these `tmt._compat.pathlib.Path` and then selectively use `Path.unrooted` to get rid of the `/`. Afaict the unrooted version is much less common then the rooted one, so you can avoid a rooting construction in most places. Could then also drop the `STEM` part to make it more compact

tmt/steps/provision/mock.py
```python
        from tmt.utils.hints import print_hints

        print_hints('provision/mock', logger=logger)

        raise tmt.utils.ProvisionError('Could not import mockbuild.config module.') from error
```
This should come together with the `installation_hint` change

LecrisUT (approved):
Awesome, thanks for the work on this 🙂

happz:
Unrelated failures, merging.

PR #4177, opened 2025-10-10
Add a Fedora integration plan, require `yq` in tests

psss:
Add a new friends plan `/fedora` to be used for verifying koji builds before releasing them in Fedora. Instead of maintaining it in the dist-git branches, let's store it in the main `tmt` repo and import it as a remote plan from there.

Also move the `yq` package requirement directly to the test metadata as it simplifies the `prepare` step and allows to more easily develop individual tests in a `tmt try` session (each test brings the required packages automatically).

Pull Request Checklist

### DIFF
--- a/plans/friends/fedora.fmf
+++ b/plans/friends/fedora.fmf
@@ -1,0 +1,18 @@
+summary: Fedora integration plan
+
+description:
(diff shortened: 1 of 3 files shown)


psss:
Obsoleted by #4253.

PR #4184, opened 2025-10-13
Special operator for `boot.method` HW requirement

happz:
`boot.method: != foo` cannot use the simple `not contains` operator, because `foo` *can* be allowed on the list of supported boot methods, as long as other boot methods are supported as well. Provisioning can than chose the right method:

```
!= bios + [uefi]       -> allowed, obviously
!= bios + [bios]       -> not allowed, there is no other option
!= bios + [bios, uefi] -> allowed, plugin can chose uefi
```

Therefor adding new operator and teaching `boot.method` parsing to use it.

Related https://issues.redhat.com/browse/TFT-3558, and https://gitlab.com/testing-farm/artemis/-/merge_requests/1466/

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/hardware.py
+++ b/tmt/hardware.py
@@ -82,14 +82,18 @@ class Operator(enum.Enum):
     LTE = '<='
     LT = '<'
 
+    #: Needle must be present in the haystack.
     CONTAINS = 'contains'
+    #: Needle must not be present in the haystack.
     NOTCONTAINS = 'not contains'
+    #: Needle may be present in the haystack as long it is not the only item.
+    NOTCONTAINS_EXCLUSIVE = 'not contains exclusive'
 
 
 INPUTABLE_OPERATORS = [
     operator
     for operator in Operator.__members__.values()
-    if operator not in (Operator.CONTAINS, Operator.NOTCONTAINS)
+    if operator not in (Operator.CONTAINS, Operator.NOTCONTAINS, Operator.NOTCONTAINS_EXCLUSIVE)
 ]
 
 
@@ -275,6 +279,24 @@ def not_contains(haystack: list[str], needle: str) -> bool:
     return needle not in haystack
 
 
+def not_contains_exclusive(haystack: list[str], needle: str) -> bool:
+    """
+    Find out whether an item is in the given list.
+
+    .. note::
+
+       A variant of :py:func:`not_contains`: an item may be present,
+       as long as other items are present.
+
+    :param haystack: container to examine.
+    :param needle: item to look for in ``haystack``.
+    :returns: ``True`` if ``needle`` is **not** in ``haystack`` or if
+        ``needle`` is in ``haystack`` but other items are there as well.
+    """
+
+    return haystack != [needle]
+
+
 OPERATOR_SIGN_TO_OPERATOR = {
     '=': Operator.EQ,
     '==': Operator.EQ,
@@ -303,6 +325,7 @@ OPERATOR_TO_HANDLER: dict[Operator, OperatorHandlerType] = {
     Operator.NOTMATCH: not_match,
     Operator.CONTAINS: operator.contains,
     Operator.NOTCONTAINS: not_contains,
+    Operator.NOTCONTAINS_EXCLUSIVE: not_contains_exclusive,
 }
 
 
@@ -1169,7 +1192,7 @@ def _parse_boot(spec: Spec) -> BaseConstraint:
             constraint.change_operator(Operator.CONTAINS)
 
         elif constraint.operator == Operator.NEQ:
-            constraint.change_operator(Operator.NOTCONTAINS)
+            constraint.change_operator(Operator.NOTCONTAINS_EXCLUSIVE)
 
         group.constraints += [constraint]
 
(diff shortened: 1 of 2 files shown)


tmt/hardware.py
```diff
@@ -82,14 +82,18 @@ class Operator(enum.Enum):
     LTE = '<='
     LT = '<'
 
+    #: Needle must be present in the haystack.
     CONTAINS = 'contains'
+    #: Needle must not be present in the haystack.
     NOTCONTAINS = 'not contains'
+    #: Needle may be present in the haystack as long it is not the only item.
+    NOTCONTAINS_EXCLUSIVE = 'not contains exclusive'
```
psss, line 90: One question: Wouldn't it be a bit better (from the readability point of view) to provide the positive version of the operator instead? Something like:

```yaml
boot.method: contains exclusive bios
```

or

```yaml
boot.method: contains only bios
```

Or, perhaps, we should add both positive & negative version of the operator?

happz: There are three options: the needle is not in the list, the needle is in the list, alone, or the needle is in the list and there are other items. We need an operator that will return positive value for the first and the third option, negative for the second. So it's something like "does not contain, or does contain but it's not the single item".

I see a negation in that in any case: both "contains exclusive bios" and "contains only bios" suggest the needle would be in the list alone, which would be the second option for which the operator must produce negative outcome.

psss: Hmmm... trying to understand use cases for the individual operators: What is actually the difference between `==` and `contains`? Is it supposed to be something like this?

```
                                [bios]          [bios,uefi]     [uefi]
method: == bios                 true            false           false
method: != bios                 false           true            true
method: contains bios           true            true            false
method: not contains bios       false           false           true
```

The use case described in TFT-3558 should be covered by the last one, right? Which seems to be the current implementation:

https://github.com/teemtee/tmt/blob/ffc89078ae11e5a4de828191db0e6ee1a66d1e7c/tmt/hardware.py#L262-L275

Perhaps, I'm missing something?

happz: > Hmmm... trying to understand use cases for the individual operators: What is actually the difference between `==` and `contains`? Is it supposed to be something like this?
> 
> ```
>                                 [bios]          [bios,uefi]     [uefi]
> method: == bios                 true            false           false
> method: != bios                 false           true            true
> method: contains bios           true            true            false
> method: not contains bios       false           false           true
> ```

Incorrect:

> ```
>                                 [bios]          [bios,uefi]     [uefi]
> method: == bios                 false           false           false
> method: != bios                 false           false           false
> method: contains bios           true            true            false
> method: not contains bios       false           false           true
> ```

Note the change in first two lines, `==` and `!=`. `==` as "equal" must produce `false` for all three inputs as `bios` is not equal to neither `[bios]`, `[bios,uefi]` nor `[uefi]`. These are lists, `bios` is a string, they will never be equal. Hence, the existence of `contains` and `not contains` operators for constrains working with lists of values.

> The use case described in TFT-3558 should be covered by the last one, right? Which seems to be the current implementation:

I'm afraid not, what we need is not strict "not contains". Our hypothetical operator:

> ```
>                                 [bios]          [bios,uefi]     [uefi]
> method: == bios                 false           false           false
> method: != bios                 false           false           false
> method: contains bios           true            true            false
> method: not contains bios       false           false           true
> method: X bios                  false           true            true
> ```

We can allow `bios` to be in the list _as long as it's not the only value there_ - drivers (and plugins) can specialize later. For example, an instance type that supports both BIOS and UEFI does not need to be disqualified when testing for `!= bios` (aka `not contains bios) - UEFI is supported too, such an instance type is acceptable. Strict `not contains` is too strict in this case.

The `==` operator could be used only if we overwrite the constraint itself, from `!= bios` to something like `!= [bios]`, and that is complicated, and it would affect how the value of constraint would be presented to the world.

psss: > > The use case described in TFT-3558 should be covered by the last one, right? Which seems to be the current implementation:
> 
> I'm afraid not, what we need is not strict "not contains". Our hypothetical operator:

I'm a bit confused. Let's get in sync on that tangible use case:

> Yes, so in this case the machine selected should not support `uefi` to make sure `bios` is used, because all our images use `uefi-preferred` and this will default to UEFI if booted and machine and image supports them.

If user specifies:

    method: not contains `uefi`

It would report `true` only for machines with `[bios]` which is exactly what is needed to make sure that `bios` is used. What am I missing?

happz: User cannot specify `not contains uefi`, use can only specify `boot.method: != uefi`, as in "not uefi, please". `[bios]` is certainly a suitable answer, but that `[bios, uefi]` is also acceptable, as long as the provision plugin takes steps to prevent the `uefi` to happen.

psss: Ok, clarified on the call. Here we want to only make sure that possibly fitting instances are not filtered out by too strict `not contains` filter.

The other part of the problem (to make sure that if an instance with `[bios, uefi]` is picked, `bios` will be picked from the two instead of preferred `uefi` as user requested by `method: != uefi`) is to be handled on the artemis side.

happz:
Unrelated Fedora CoreOS image issues, merging.

PR #4190, opened 2025-10-13
Fix the forgotten template in `/tests/discover/link`

psss:
Add a reasonable summary and description about the test. This slipped through in #4146.


psss:
Full tests passed, just docs change, merging.

PR #4202, opened 2025-10-16
Implement Brew artifact provider

AthreyVinay:
resolves: #4174

Pull Request Checklist

* [X] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tests/unit/artifact/test_brew.py
+++ b/tests/unit/artifact/test_brew.py
@@ -1,0 +1,60 @@
+import pytest
+
+from tmt.steps.prepare.artifact.providers.brew import BrewArtifactProvider
+
+
+@pytest.mark.skip(reason="would be replaced by mocks")
+@pytest.mark.integration
+def test_brew_valid_build(root_logger):
+    provider = BrewArtifactProvider("brew.build:3866328", root_logger)
+    assert len(provider.artifacts) == 21
+
+
+@pytest.mark.skip(reason="would be replaced by mocks")
+@pytest.mark.integration
+def test_brew_valid_draft_build(root_logger):
+    provider = BrewArtifactProvider("brew.build:3525300", root_logger)
+    assert len(provider.artifacts) == 2
+    assert any(
+        "draft_3525300" in artifact._raw_artifact['url'] for artifact in provider.artifacts
+    ), "No artifact URL contains 'draft_3525300'"
+
+
+@pytest.mark.skip(reason="would be replaced by mocks")
+@pytest.mark.integration
+def test_brew_valid_nvr(root_logger):
+    provider = BrewArtifactProvider("brew.nvr:unixODBC-2.3.12-1.el9", root_logger)
+    assert len(provider.artifacts) == 21
+    assert provider.build_id == 3866328  # Known build ID for this NVR
+
+
+@pytest.mark.skip(reason="would be replaced by mocks")
+@pytest.mark.integration
+def test_brew_invalid_nvr(root_logger):
+    from tmt.utils import GeneralError
+
+    provider = BrewArtifactProvider("brew.nvr:nonexistent-1.0-1.fc43", root_logger)
+
+    with pytest.raises(GeneralError, match=r"No build found for NVR 'nonexistent-1\.0-1\.fc43'\."):
+        _ = provider.build_id
+
+
+@pytest.mark.skip(reason="would be replaced by mocks")
+@pytest.mark.integration
+def test_brew_valid_task_id_actual_build(root_logger):
+    provider = BrewArtifactProvider("brew.task:69098388", root_logger)
+    assert provider.build_id == 3866328  # Known build ID for this task
+    assert len(provider.artifacts) == 21
+
+
+@pytest.mark.skip(reason="would be replaced by mocks")
+@pytest.mark.integration
+def test_brew_valid_task_id_scratch_build(root_logger):
+    task_id = 69111304
+    provider = BrewArtifactProvider(f"brew.task:{task_id}", root_logger)
+    tasks = list(provider._get_task_children(task_id))
+
+    assert len(tasks) == 6
+    assert task_id in tasks
+    assert provider.build_id is None
+    assert len(provider.artifacts) == 12
--- a/tmt/steps/prepare/artifact/providers/brew.py
+++ b/tmt/steps/prepare/artifact/providers/brew.py
@@ -2,14 +2,103 @@
 Brew Artifact Provider
 """
 
+from collections.abc import Sequence
+from functools import cached_property
+from typing import Any, ClassVar, Optional
+from urllib.parse import urljoin
+
+import tmt.log
 from tmt.steps.prepare.artifact.providers import provides_artifact_provider
-from tmt.steps.prepare.artifact.providers.koji import KojiArtifactProvider
+from tmt.steps.prepare.artifact.providers.koji import (
+    KojiArtifactProvider,
+    KojiBuild,
+    KojiNvr,
+    KojiTask,
+)
 
 
 # ignore[type-arg]: TypeVar in provider registry annotations is
 # puzzling for type checkers. And not a good idea in general, probably.
 @provides_artifact_provider('brew')  # type: ignore[arg-type]
-class BrewProvider(KojiArtifactProvider):
+class BrewArtifactProvider(KojiArtifactProvider):
     """
+    Provider for downloading artifacts from Brew builds.
+
     Brew builds are just a special case of Koji builds
+
+    .. note::
+
+        Only RPMs are supported currently.
+
+    .. code-block:: python
+
+        provider = BrewArtifactProvider("brew.build:123456", logger)
+        artifacts = provider.fetch_contents(guest, Path("/tmp"))
     """
+
+    SUPPORTED_PREFIXES: ClassVar[tuple[str, ...]] = ()
+
+    def __new__(cls, raw_provider_id: str, logger: tmt.log.Logger) -> Any:
+        """
+        Create a specific Brew provider based on the ``raw_provider_id`` prefix.
+
+        The supported providers are:
+        :py:class:`BrewBuild`,
+        :py:class:`BrewTask`,
+        :py:class:`BrewNvr`.
+
+        :raises ValueError: If the prefix is not supported
+        """
+        return cls._dispatch_subclass(raw_provider_id, cls._REGISTRY)
+
+    def __init__(self, raw_provider_id: str, logger: tmt.log.Logger):
+        super().__init__(raw_provider_id, logger)
+        self._session = self._initialize_session(
+            api_url="https://brewhub.engineering.redhat.com/brewhub",
+            top_url="https://download.eng.bos.redhat.com/brew",
+        )
+
+    @cached_property
+    def build_provider(self) -> Optional['BrewBuild']:
+        return self._make_build_provider(BrewBuild, "brew.build")
+
+    def _rpm_url(self, rpm_meta: dict[str, str]) -> str:
+        """Construct Brew RPM URL."""
+        name = rpm_meta["name"]
+        version = rpm_meta["version"]
+        release = rpm_meta["release"]
+        arch = rpm_meta["arch"]
+        assert self.build_info is not None
+        volume = self.build_info["volume_name"]
+        draft_suffix = f",draft_{self.build_id}" if self.build_info["draft"] else ""
+
+        path = (
+            f"vol/{volume}/packages/{name}/{version}/"
+            f"{release}{draft_suffix}/{arch}/"
+            f"{name}-{version}-{release}.{arch}.rpm"
+        )
+
+        return urljoin(self._top_url, path)
+
+
+@provides_artifact_provider("brew.build")  # type: ignore[arg-type]
+class BrewBuild(BrewArtifactProvider, KojiBuild):
+    pass
+
+
+@provides_artifact_provider("brew.task")  # type: ignore[arg-type]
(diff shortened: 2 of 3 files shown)


tmt/steps/prepare/artifact/providers/brew.py
```diff
@@ -2,14 +2,103 @@
...
+
+        :raises ValueError: If the prefix is not supported
+        """
+        return cls._dispatch_subclass(raw_provider_id, cls._REGISTRY)
+
+    def __init__(self, raw_provider_id: str, logger: tmt.log.Logger):
+        super().__init__(raw_provider_id, logger)
+        self._session = self._initialize_session(
+            api_url="https://brewhub.engineering.redhat.com/brewhub",
+            top_url="https://download.eng.bos.redhat.com/brew",
```
tcornell-bus, line 58: `download.eng.bos.redhat.com` is already retired or will be very soon. Is there somewhere in particular you got this URL? If so, maybe we can get the new URL from there.

Other options:
```
download.engineering.redhat.com
download.devel.redhat.com
```

tcornell-bus (commented):
When I attempt to run the unit tests, I get a certificate error. Is there a certain way the unit tests need to be run?

AthreyVinay:
> When I attempt to run the unit tests, I get a certificate error. Is there a certain way the unit tests need to be run?

Yes, you would need the certs on your system:
```rst
cd /etc/pki/ca-trust/source/anchors
sudo wget https://certs.corp.redhat.com/certs/Current-IT-Root-CAs.pem --no-check-certificate
sudo update-ca-trust extract
```
and then setting the `REQUESTS_CA_BUNDLE` (https://tmt.readthedocs.io/en/stable/contribute.html#develop). I did something like this:
```rst
export REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/source/anchors/Current-IT-Root-CAs.pem
```

Also make sure the vpn is connected :) Lemme know if this still doesnt work / complains of cert errors. 

Thanks for testing btw :) & thanks @psss for the quick guide(https://redhat-internal.slack.com/archives/C08J5QPSSKH/p1760347815746409?thread_ts=1760335209.852519&cid=C08J5QPSSKH)

tmt/steps/prepare/artifact/providers/brew.py
AthreyVinay: I dont remember exactly now - but must have been from one of the links from various slack threads that I was reading / referencing to get brew to download rpms :rofl: 

The real fix should be to read the url's from some sort of a config (like koji) rather than hardcoding. WDYT? BUt I can always replace the url with one of the above and test if it works :)

tcornell-bus: I think getting it from a config file would be best, like you did with koji. I have a `/etc/koji.conf.d/brewkoji.conf` file that has the [brew] profile.

Are we expecting the test runner machine to already have brew setup? If so, then the config file should be there, and you should be able to do the same with brew that you did with koji:
`config = koji.read_config("brew")`.

tcornell-bus: Also the topurl in my config is `https://download.devel.redhat.com/brewroot` in case you want it for troubleshooting.

tests/unit/artifact/test_brew.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+
+from tmt.steps.prepare.artifact.providers.brew import BrewArtifactProvider
+
+
+@pytest.mark.integration
+def test_brew_valid_build(root_logger):
+    provider = BrewArtifactProvider("brew.build:3866328", root_logger)
+    assert len(provider.artifacts) == 21
```
tcornell-bus, line 10: There are 21 rpms for that build, but that includes all the different arches. Don't we only want the arch we are testing on plus any noarch rpms? For example, with x86_64 there are only 5 rpms, which includes the src.rpm. I am not sure if the src.rpm is necessary for install.

tmt/steps/prepare/artifact/providers/brew.py
AthreyVinay: >  I have a /etc/koji.conf.d/brewkoji.conf file that has the [brew] profile.

This is what I would expect the runners to have.

tests/unit/artifact/test_brew.py
AthreyVinay: Perfect, point noted - this calls for a wider audience discussion - the idea was to initially gather all that ends with `.rpm`.

tmt/steps/prepare/artifact/providers/brew.py
happz: They most likely won't have any pre-existing Brew configuration. On top of the setup, we will need to split these unit tests from the main test coverage, and run them in the private ranch, because the public one has no access to Brew.

Well, also one thing to consider: if the test needs an actual URL and live API, it becomes an integration test rather than a unit-level test. Mocking the API and API responses would be a way forward, we would focus on testing our side of the contract without worrying whether Brew API is up or not.

tests/unit/artifact/test_brew.py
tcornell-bus: 🫡 Sounds good to me, the rpms downloaded can always be narrowed later

tmt/steps/prepare/artifact/providers/brew.py
happz: BTW we can add an actual integration test, e.g. running `tmt run artifact --brew-build ...` command and checking the outcomes, and make it part of the set of tests we run in the private ranch. It would complement the unit testing nicely, and provide the safe check of against a live API. But for checking whether our code translates API responses into the expected Brew provider behavior - or Koji provider, same case - we don't really need Brew or Koji services, because we can mock the hell out of the session objects & pretend they return the expected data structures.

happz:
> > When I attempt to run the unit tests, I get a certificate error. Is there a certain way the unit tests need to be run?
> 
> Yes, you would need the certs on your system:
> 
> ```
> cd /etc/pki/ca-trust/source/anchors
> sudo wget https://certs.corp.redhat.com/certs/Current-IT-Root-CAs.pem --no-check-certificate
> sudo update-ca-trust extract
> ```
> 
> and then setting the `REQUESTS_CA_BUNDLE` (https://tmt.readthedocs.io/en/stable/contribute.html#develop). I did something like this:
> 
> ```
> export REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/source/anchors/Current-IT-Root-CAs.pem
> ```
> 
> Also make sure the vpn is connected :) Lemme know if this still doesnt work / complains of cert errors.
> 
> Thanks for testing btw :) & thanks @psss for the quick guide(https://redhat-internal.slack.com/archives/C08J5QPSSKH/p1760347815746409?thread_ts=1760335209.852519&cid=C08J5QPSSKH)

```
export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-bundle.crt
```

Unlike `/etc/pki/ca-trust/source/anchors/Current-IT-Root-CAs.pe`, this makes sure all CAs are honored, not just RH ones. You might have issues with public services, for example (not here and now when testing Brew, but it's a good habit).

tests/unit/artifact/test_brew.py
AthreyVinay: Yes, this exists in the base solely for this reason:
https://github.com/teemtee/tmt/blob/64be40c6ad99e16e029c23fce2d5befe348e0da0/tmt/steps/prepare/artifact/providers/__init__.py#L184

tmt/steps/prepare/artifact/providers/brew.py
AthreyVinay: It just felt nice to call the live API's :) rather than having to "mock the hell out" ! (not a fan of mocking something that can otherwise be called in realtime)

AthreyVinay: @happz Do you want me to replace the integration tests with mocks or let this be for the time being and revisit?

happz: If those tests depend in live Brew API, they will most likely fail because they run in the public ranch with no access to Brew. So disabling them and a follow-up PR changing them to use mock would work for me, or dropping them for now and a follow-up PR re-adding them but based on mocks would work as well. Whichever test depends on live Brew API will fail, and moving them into private ranch seems like a lot of work when mocking would be more natural way to approach the problem anyway.

AthreyVinay: Tests disabled here: [dc2ba93](https://github.com/teemtee/tmt/pull/4202/commits/dc2ba9324b875e5af7bf6b544564e5bdbdf89519)

LecrisUT: > I think getting it from a config file would be best, like you did with koji. I have a `/etc/koji.conf.d/brewkoji.conf` file that has the [brew] profile.

Could you `dnf whatprovides` that file? I couldn't find the brew packages either on fedora or centos stream. If we get it from a file that is not packaged on Fedora, we might not be able to add it by default, but we could make it as an internal plugin, which might not be a bad way around the testings and everything.

> It just felt nice to call the live API's :) rather than having to "mock the hell out" ! (not a fan of mocking something that can otherwise be called in realtime)

Depends on what is being tested. For the koji tests it would be good to have at least some functional tests because there is just too much magic behind those. But if for brew you do not have access to the internal api we could just do it as best-effort. We could probably make a specific set of tests to run on the internal, but that would be part of a more general refactor (it was discussed, but can't find the tracking issue)

tcornell-bus: It's provided by `brewkoji` from the internal RCMTOOLS repo. I think it will have to be an internal plugin

happz: I believe we will have mock-based tests in one of the future PRs.

LecrisUT: Still the mocked tests do not help if we didn't tests those manually before or have some other integration tests that cover it. But we can improve on this gradually.

tmt/steps/prepare/artifact/providers/brew.py
```diff
@@ -2,14 +2,103 @@
...
+        arch = rpm_meta["arch"]
+        assert self.build_info is not None
+        volume = self.build_info["volume_name"]
+        draft_suffix = f",draft_{self.build_id}" if self.build_info["draft"] else ""
+
+        path = (
+            f"vol/{volume}/packages/{name}/{version}/"
+            f"{release}{draft_suffix}/{arch}/"
+            f"{name}-{version}-{release}.{arch}.rpm"
+        )
```
LecrisUT, line 79: The fact that this path is constructed so much more differently from the koji one scares me. I am guessing that if you use `brew` module instead of `koji` it could be constructed more sanely?

tmt/steps/prepare/artifact/providers/brew.py
```diff
@@ -2,14 +2,103 @@
...
+
+        return urljoin(self._top_url, path)
+
+
+@provides_artifact_provider("brew.build")  # type: ignore[arg-type]
+class BrewBuild(BrewArtifactProvider, KojiBuild):
+    pass
```
LecrisUT, line 86: Diamond inheritance issues are always fun to review. I suspect it is fine because `_rpm_url` and `build_provider` are not overwritten by the `Koji*`, but just to double-check, you've tested these parts localy?

tmt/steps/prepare/artifact/providers/brew.py
```diff
@@ -2,14 +2,103 @@
...
+
+    def _rpm_url(self, rpm_meta: dict[str, str]) -> str:
+        """Construct Brew RPM URL."""
+        name = rpm_meta["name"]
+        version = rpm_meta["version"]
+        release = rpm_meta["release"]
+        arch = rpm_meta["arch"]
+        assert self.build_info is not None
+        volume = self.build_info["volume_name"]
+        draft_suffix = f",draft_{self.build_id}" if self.build_info["draft"] else ""
```
LecrisUT, line 73: Does this really use `,` in the url?

tmt/steps/prepare/artifact/providers/brew.py
AthreyVinay: I have - yes. The integration tests were succeeding until replaced.

LecrisUT (approved):

tmt/steps/prepare/artifact/providers/brew.py
AthreyVinay: It does - I was lucky/unlucky  to have stumbled upon a draft.
For example:
```
osci-internal-test-package-0.2.0-22.el10_0,draft_3525300
```

LecrisUT: :scream:

happz:
@AthreyVinay please, rebase once you have a moment. A simple fast-forward is not possible because of conflicts.

AthreyVinay:
> rebase once you have a moment.

Done :+1:

tmt/steps/prepare/artifact/providers/brew.py
happz: Scary enough to do something about it before merging, or just low-level scary it would be good enough to put it on the scratch list of things to get back to later?

LecrisUT: In the current design there is nothing we can do. Scratch list for now I suppose

AthreyVinay: Updated the scratch list with a link to this comment.

PR #4205, opened 2025-10-16
Add Repository Class with Tests

vaibhavdaren:
This commit introduces the core functionality for creating and identifying `Repository` objects. The primary goal is to provide flexible initialization options while ensuring that identical repositories always resolve to the same unique ID.

**Repository Creation Methods:**
1.  **From URL:**
    -   The last segment of the URL path is treated as the filename.
    -   The repository `name` defaults to the filename but can be overridden by an explicit parameter.
    -   Handles URLs that end with or without the `.repo` extension.
2.  **From File Path:**
        -   The repository `name` is derived from the file path.
        -   An explicit `name` parameter will take precedence over the derived name.
3.  **From Content:**
        -   Requires the raw content and a `file_name` to be provided directly.

### DIFF
--- a/tests/unit/artifact/test_repository.py
+++ b/tests/unit/artifact/test_repository.py
@@ -58,25 +58,27 @@ def temp_repo_file_no_ext(tmp_path):
     return repo_file
 
 
-def test_init_from_content():
+def test_init_from_content(root_logger):
     """Test successful initialization from a content string"""
-    repo = Repository.from_content(name="from-content", content=VALID_REPO_CONTENT)
+    repo = Repository.from_content(
+        name="from-content", content=VALID_REPO_CONTENT, logger=root_logger
+    )
     assert repo.name == "from-content"
     assert repo.content == VALID_REPO_CONTENT
     assert repo.repo_ids == EXPECTED_REPO_IDS
     assert repo.filename == "from-content.repo"
 
 
-def test_init_from_file(temp_repo_file):
+def test_init_from_file(temp_repo_file, root_logger):
     """Test successful initialization from a local file path"""
-    repo = Repository.from_file_path(file_path=temp_repo_file)
+    repo = Repository.from_file_path(file_path=temp_repo_file, logger=root_logger)
     assert repo.name == "docker-ce"
     assert repo.content == VALID_REPO_CONTENT
     assert repo.repo_ids == EXPECTED_REPO_IDS
 
 
 @patch('tmt.steps.prepare.artifact.providers.tmt.utils.retry_session')
-def test_init_from_url(mock_retry_session):
+def test_init_from_url(mock_retry_session, root_logger):
     """Test successful initialization from a URL"""
     mock_session = MagicMock()
     mock_response = MagicMock()
@@ -86,7 +88,7 @@ def test_init_from_url(mock_retry_session):
     mock_retry_session.return_value.__enter__.return_value = mock_session
 
     repo_url = "https://download.docker.com/linux/centos/docker-ce.repo"
-    repo = Repository.from_url(url=repo_url)
+    repo = Repository.from_url(url=repo_url, logger=root_logger)
 
     assert repo.name == "docker-ce"
     assert repo.content == VALID_REPO_CONTENT
@@ -95,10 +97,10 @@ def test_init_from_url(mock_retry_session):
 
 
 @patch('tmt.steps.prepare.artifact.providers.tmt.utils.retry_session')
-def test_name_derivation(mock_retry_session, temp_repo_file):
+def test_name_derivation(mock_retry_session, temp_repo_file, root_logger):
     """Test the logic for deriving the repository name"""
     # Provided name takes precedence
-    repo = Repository.from_content(name="explicit-name", content="[foo]")
+    repo = Repository.from_content(name="explicit-name", content="[foo]", logger=root_logger)
     assert repo.name == "explicit-name"
 
     # Mock the session for the URL test
@@ -110,52 +112,54 @@ def test_name_derivation(mock_retry_session, temp_repo_file):
     mock_retry_session.return_value.__enter__.return_value = mock_session
 
     # Name derived from URL
-    repo = Repository.from_url(url="http://a.b/c/docker-ce.repo")
+    repo = Repository.from_url(url="http://a.b/c/docker-ce.repo", logger=root_logger)
     assert repo.name == "docker-ce"
 
     # Name derived from file path
-    repo = Repository.from_file_path(file_path=temp_repo_file)
+    repo = Repository.from_file_path(file_path=temp_repo_file, logger=root_logger)
     assert repo.name == "docker-ce"
 
 
-def test_repo_id_parsing():
+def test_repo_id_parsing(root_logger):
     """Test the parsing of repo IDs from content"""
     # Valid content with multiple sections
-    repo = Repository.from_content(name="valid", content=VALID_REPO_CONTENT)
+    repo = Repository.from_content(name="valid", content=VALID_REPO_CONTENT, logger=root_logger)
     assert repo.repo_ids == EXPECTED_REPO_IDS
 
     # Content with no sections (empty)
     with pytest.raises(GeneralError, match="No repository sections found"):
-        Repository.from_content(name="no-sections", content="")
+        Repository.from_content(name="no-sections", content="", logger=root_logger)
 
     # Content with options but no section header
     with pytest.raises(GeneralError, match="No repository sections found"):
-        Repository.from_content(name="no-sections", content=NO_SECTION_CONTENT)
+        Repository.from_content(name="no-sections", content=NO_SECTION_CONTENT, logger=root_logger)
 
     # Malformed content should also raise an error
     with pytest.raises(GeneralError, match=r"The .repo file may be malformed"):
-        Repository.from_content(name="malformed", content=MALFORMED_REPO_CONTENT)
+        Repository.from_content(
+            name="malformed", content=MALFORMED_REPO_CONTENT, logger=root_logger
+        )
 
 
 @patch('tmt.steps.prepare.artifact.providers.tmt.utils.retry_session')
-def test_failed_url_fetch(mock_retry_session):
+def test_failed_url_fetch(mock_retry_session, root_logger):
     """Test that a failed URL fetch raises an error"""
     mock_session = MagicMock()
     mock_session.get.side_effect = requests.RequestException("Connection failed")
     mock_retry_session.return_value.__enter__.return_value = mock_session
 
     with pytest.raises(GeneralError, match="Failed to fetch repository content"):
-        Repository.from_url(url="http://example.com/invalid.repo")
+        Repository.from_url(url="http://example.com/invalid.repo", logger=root_logger)
 
 
-def test_nonexistent_file():
+def test_nonexistent_file(root_logger):
     """Test that a non-existent file path raises an error"""
     with pytest.raises(GeneralError, match="Failed to read repository file"):
-        Repository.from_file_path(file_path=Path("/no/such/file.repo"))
+        Repository.from_file_path(file_path=Path("/no/such/file.repo"), logger=root_logger)
 
 
 @patch('tmt.steps.prepare.artifact.providers.tmt.utils.retry_session')
-def test_url_trailing_slash(mock_retry_session):
+def test_url_trailing_slash(mock_retry_session, root_logger):
     """Test URL with trailing slash"""
     mock_session = MagicMock()
     mock_response = MagicMock()
@@ -164,12 +168,12 @@ def test_url_trailing_slash(mock_retry_session):
     mock_session.get.return_value = mock_response
     mock_retry_session.return_value.__enter__.return_value = mock_session
 
-    repo = Repository.from_url(url="https://example.com/repo/")
+    repo = Repository.from_url(url="https://example.com/repo/", logger=root_logger)
     assert repo.name == "repo"
 
 
 @patch('tmt.steps.prepare.artifact.providers.tmt.utils.retry_session')
-def test_url_no_repo_ext(mock_retry_session):
+def test_url_no_repo_ext(mock_retry_session, root_logger):
     """Test URL without .repo extension"""
     mock_session = MagicMock()
     mock_response = MagicMock()
@@ -178,34 +182,34 @@ def test_url_no_repo_ext(mock_retry_session):
     mock_session.get.return_value = mock_response
     mock_retry_session.return_value.__enter__.return_value = mock_session
 
-    repo = Repository.from_url(url="https://example.com/docker-ce")
+    repo = Repository.from_url(url="https://example.com/docker-ce", logger=root_logger)
     assert repo.name == "docker-ce"
 
 
-def test_file_no_repo_ext(temp_repo_file_no_ext):
+def test_file_no_repo_ext(temp_repo_file_no_ext, root_logger):
     """Test file without .repo extension"""
-    repo = Repository.from_file_path(file_path=temp_repo_file_no_ext)
+    repo = Repository.from_file_path(file_path=temp_repo_file_no_ext, logger=root_logger)
     assert repo.name == "docker-ce"
 
 
-def test_content_no_name():
+def test_content_no_name(root_logger):
     """Test content without name raises error"""
     with pytest.raises(GeneralError, match="Repository name cannot be empty"):
-        Repository.from_content(content=VALID_REPO_CONTENT, name="")
+        Repository.from_content(content=VALID_REPO_CONTENT, name="", logger=root_logger)
 
 
 @patch('tmt.steps.prepare.artifact.providers.tmt.utils.retry_session')
-def test_invalid_url_format(mock_retry_session):
+def test_invalid_url_format(mock_retry_session, root_logger):
     """Test initialization with invalid URL format"""
     mock_session = MagicMock()
     mock_session.get.side_effect = requests.exceptions.InvalidURL("Invalid URL")
     mock_retry_session.return_value.__enter__.return_value = mock_session
 
     with pytest.raises(GeneralError, match="Failed to fetch repository content"):
-        Repository.from_url(url="invalid_url")
+        Repository.from_url(url="invalid_url", logger=root_logger)
 
 
-def test_url_no_path():
+def test_url_no_path(root_logger):
     """Test URL with no path raises error"""
     with patch(
         'tmt.steps.prepare.artifact.providers.tmt.utils.retry_session'
@@ -218,4 +222,4 @@ def test_url_no_path():
         mock_retry_session.return_value.__enter__.return_value = mock_session
 
         with pytest.raises(GeneralError, match="Could not derive repository name from URL"):
-            Repository.from_url(url="https://example.com/")
+            Repository.from_url(url="https://example.com/", logger=root_logger)
--- a/tmt/steps/prepare/artifact/providers/__init__.py
+++ b/tmt/steps/prepare/artifact/providers/__init__.py
@@ -238,18 +238,21 @@ class Repository:
             ) from error
 
     @classmethod
-    def from_url(cls, url: str, name: Optional[str] = None) -> "Repository":
+    def from_url(
+        cls, url: str, logger: tmt.log.Logger, name: Optional[str] = None
+    ) -> "Repository":
         """
         Create a Repository instance by fetching content from a URL.
 
         :param url: The URL to fetch the repository content from.
+        :param logger: Logger to use for the operation.
         :param name: Optional name for the repository. If not provided,
             derived from the URL.
         :returns: A Repository instance.
         :raises GeneralError: If fetching or parsing fails.
         """
         try:
-            with tmt.utils.retry_session() as session:
+            with tmt.utils.retry_session(logger=logger) as session:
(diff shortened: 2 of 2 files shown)


tmt/steps/prepare/artifact/providers/repository.py
```diff
@@ -35,80 +35,81 @@ class RepositoryFileProvider(ArtifactProvider[RpmArtifactInfo]):
...
+    #     """
+    #     # Initialize and install the repository on the guest
+    #     repository = Repository(url=self.id, filename=self.repo_filename)
+    #     repository.install(guest=guest, logger=self.logger)
+    #
+    #     # Cache the list of discovered RPMs
+    #     self._rpm_list = repository.rpms
+    #
+    #     # Return empty list as no artifacts are downloaded
+    #     return []
```
tcornell-bus: Should all this just be deleted?

vaibhavdaren: ^_^  I have a very small PR to make changes for the Provider to use the new format https://github.com/teemtee/tmt/pull/4207.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +201,114 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+        name: Optional[str] = None,
+        url: Optional[str] = None,
+        file_path: Optional[Path] = None,
+        content: Optional[str] = None,
+    ):
+        self.logger = logger
+        self._provided_name = name
+        self._provided_url = url
+        self._provided_file_path = file_path
+        self._provided_content = content
```
LecrisUT: Dunno about having all options being saved. Would prefer something like using `__new__` to dispatch specific sub classes or saving it into a more generic name.

As for the options, I do not like it that in 2 cases this is not tied to a physical file even though you have a `filename` property. Would prefer for the constructor to explicitly request a path where the physical repo will be created which we can then sync with the guests.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +201,114 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+            return self._provided_name
+        if self._provided_url:
+            # Use the last path segment from the URL
+            parsed_path = urlparse(self._provided_url).path
+            return parsed_path.rstrip('/').split('/')[-1].replace('.repo', '')
+        if self._provided_file_path:
+            # Derive from local filename
+            return self._provided_file_path.name.replace('.repo', '')
+        # Fallback to a name derived from the unique ID
+        return f"repo-{self.id[:8]}"
```
LecrisUT: This handling is quite arbitrary, would rather enforce the caller to provide a name. Let's skip the hashing logic until we know if/how we need it.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +201,114 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+
+    @functools.cached_property
+    def repo_ids(self) -> list[str]:
+        """Parses the .repo content to extract repository IDs using configparser."""
+        if not self.content:
+            return []
+        config = configparser.ConfigParser()
+        try:
+            config.read_string(self.content)
+            return config.sections()
```
LecrisUT: Check that `sections` is non-empty?

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +201,114 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+                raise tmt.utils.GeneralError(
+                    f"Failed to read repository file '{self._provided_file_path}'."
+                ) from error
+        return None
+
+    @functools.cached_property
+    def repo_ids(self) -> list[str]:
+        """Parses the .repo content to extract repository IDs using configparser."""
+        if not self.content:
+            return []
```
LecrisUT: Would error on this. If we need to allow users to include an empty repo, let's first figure out their context of needing it, then we can try to add it in whichever way fits best.

tests/unit/artifact/test_repository.py
```diff
@@ -0,0 +1,187 @@
...
+[my-repo
+name=Missing closing bracket
+"""
+
+
+# Mock logger fixture for all tests
+@pytest.fixture
+def logger():
+    """Provides a tmt logger for tests"""
+    return tmt.log.Logger.create()
```
LecrisUT: We already have `root_logger` fixture.

tests/unit/artifact/test_repository.py
```diff
@@ -0,0 +1,187 @@
...
+    assert repo.name == "docker-ce"
+    assert repo.content == VALID_REPO_CONTENT
+    assert repo.repo_ids == EXPECTED_REPO_IDS
+    mock_session.get.assert_called_once_with(repo_url)
+
+
+def test_init_no_source_fails(logger):
+    """Test that initialization fails if no source is provided"""
+    with pytest.raises(GeneralError, match="Repository content could not be loaded"):
+        Repository(logger=logger)
```
LecrisUT: This should be enforced via type-checking and not just via runtime error.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +201,114 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+        config = configparser.ConfigParser()
+        try:
+            config.read_string(self.content)
+            return config.sections()
+        except configparser.Error as e:
```
psss:
```suggestion
        except configparser.Error as error:
```

Let's not use one-letter variables.

tests/unit/artifact/test_repository.py
psss: You might want to reuse the common logger fixture:

https://github.com/teemtee/tmt/blob/e37496938089567a6c564825e36d7318112d1ddd/tests/unit/conftest.py#L18-L20

psss (changes requested):
Thank for implementing the class. Looks like a good kick-off! Just two minor suggestions.

tmt/steps/prepare/artifact/providers/__init__.py
LecrisUT: Reviewed in the wrong order, I see why you didn't tie it to a file with the usage in the engine, but this would make debugging when this goes wrong quite tricky. Having a local copy would still be preferable even with that implementation.

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: Addressed in 36fdcb40cdeb25e484fb83e844c750019ffbbbb6

tests/unit/artifact/test_repository.py
vaibhavdaren: Addressed in 36fdcb40cdeb25e484fb83e844c750019ffbbbb6

tmt/steps/prepare/artifact/providers/__init__.py
psss: Thanks!

tests/unit/artifact/test_repository.py
psss: Fixed in 36fdcb40cdeb25e484fb83e844c750019ffbbbb6.

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: Removed in next commit. Repo is now identified by name.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +200,121 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+
+    @classmethod
+    def create(
+        cls,
+        logger: tmt.log.Logger,
+        name: Optional[str] = None,
+        url: Optional[str] = None,
+        file_path: Optional[Path] = None,
+        content: Optional[str] = None,
+    ) -> "Repository":
```
LecrisUT: Either drop this one, or add `@typing.overload` type hints to enforce the `provided == 1` rule you have below. I don't expect we would need  `create` over the individual `from_url` since we have to specialize caller to pass the appropriate `url`/`file_path`/`content` keyword anyway.

Also, general tip, you should use `*` to enforce these are passed as `kwargs`.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +200,121 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+    def from_content(cls, logger: tmt.log.Logger, content: str, name: str) -> "Repository":
+        return cls(logger=logger, name=name, content=content)
+
+    @property
+    def content(self) -> str:
+        return self._content
+
+    @property
+    def name(self) -> str:
+        return self._name
```
LecrisUT: Drop the properties if they don't do anything fancy.

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: Added typing overload type hints.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,201 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+class Repository:
+    """A class to represent a dnf/yum software repository."""
+
+    def __init__(self, logger: tmt.log.Logger, name: str, content: str):
+        self.logger = logger
+        self.name = name
+        self.content = content
+        self._repo_ids = self._get_repo_ids()
+
+    # The @overload decorator is used here to provide type hinting for multiple possible
```
LecrisUT: Remove or be brief:
```
@overload to enforce the runtime check at type-checking
```

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,201 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+    @overload
+    @classmethod
+    def create(
+        cls,
+        logger: tmt.log.Logger,
+        *,
+        name: Optional[str] = None,
+        url: str,
+        file_path: None = None,
+        content: None = None,
```
LecrisUT: Can you not remove `file_path` and `content`? 

Also need to verify that if mypy does not match any of it it would error out. It might require stripping the annotations of the implementation, not familiar with the nuances of that, but some light experimentation should uncover the behavior

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,201 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+        logger: tmt.log.Logger,
+        *,
+        name: str,
+        content: str,
+        url: None = None,
+        file_path: None = None,
+    ) -> "Repository": ...
+
+    @classmethod
+    def create(
```
happz: I'm not sure I see the point of `create()` with so many parameters. If I have a repository file URL, I would simply call `from_url()` to create a `Repository` instance. What is the use case of `create()`?

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,201 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
                 yield artifact
 
 
+class Repository:
+    """A class to represent a dnf/yum software repository."""
+
+    def __init__(self, logger: tmt.log.Logger, name: str, content: str):
+        self.logger = logger
```
LecrisUT: No need to save the logger

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: #### Direct calls (what you're suggesting)
```
repo1 = Repository.from_url(logger=logger, url="https://example.com/repo.repo")
repo2 = Repository.from_file_path(logger=logger, file_path=Path("/local/repo.repo"))
repo3 = Repository.from_content(logger=logger, name="myrepo", content="...")
```
#### Unified via create()
```
repo1 = Repository.create(logger=logger, url="https://example.com/repo.repo")
repo2 = Repository.create(logger=logger, file_path=Path("/local/repo.repo"))
repo3 = Repository.create(logger=logger, name="myrepo", content="...")
```

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,201 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
                 yield artifact
 
 
+class Repository:
+    """A class to represent a dnf/yum software repository."""
```
LecrisUT: I don't like the wording, but it's hard to articulate why. Probably the main thing is:
- `A class`: self-evident. There should be something better to use
- `dnf/yum software repository ` there does not exist a distinction between dnf and yum on repo level, and there's no need to document it as we will only support those
- it misses the fact that this is primarily to contain the contents of the repo file vs its physical file or whatever

Can't find a meaningful docstring to replace it with, but sometimes no docstring is fine, particularly if the class is lightweight enough to understand at a glance.

Maybe you can focus on documenting the attributes and something would come up from there.


```suggestion
    """
    Thin wrapper/holder for .repo file content
    """
    #: Content of the repository
    content: str
    #: Uniquely identifiable name
    name: str
```
(My class docstring is by no means acceptable either but it is closer to what I would like to see in documentation: what's the purpose of it? Just a thin wrapper class around some data. What is the main content of it? Dnf repo, its text representation. Does it do any special actions or controls? Not that I can see. Not necessary to answer all or any of these questions, but anything that considers who would read such a documentation and what would be useful for them to see there)

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,201 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+            raise GeneralError(f"No repository sections found in the content for '{self.name}'.")
+        except configparser.Error as error:
+            raise GeneralError(
+                f"Failed to parse the content of repository '{self.name}'. "
+                "The .repo file may be malformed."
+            ) from error
+
+    @property
+    def repo_ids(self) -> list[str]:
+        return self._repo_ids
```
LecrisUT: Also not needed, same as the previous.

tmt/steps/prepare/artifact/providers/__init__.py
LecrisUT: Usually the designs go the other way around having an almighty creator and then specialized creators with more context and nuance.

There could be cases where it is reversed e.g. if you want to ignore the input that you pass to it like when you forward the contents of a yaml file.

Overall the unified create does not help us because in the context that we are going to use, we already know what variants we want to use in each case. Think of it from the perspective of the user, if you could use both of them, when would you choose one or the other. I would chose the explicit `from_url` because the interface there looks more narrow, and I already know that I want to construct it from the url explicitly.

vaibhavdaren:
**[Note]** The test failures are unrelated to this change.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -195,6 +199,117 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+        :raises GeneralError: If the name is empty.
+        """
+        if not name:
+            raise GeneralError("Repository name cannot be empty.")
+        return cls(name=name, content=content)
+
+    @property
+    def filename(self) -> str:
+        """The name of the .repo file (e.g., 'my-repo.repo')."""
+        return f"{self.name}.repo"
```
happz: Risky, name can be something really ugly, or even contain characters like `/`. Use `tmt.utils.sanitize_name()`.

vaibhavdaren: Thankyou for pointing me to the fancy util function. Addressed in  3c4944700b6b0f13e7601639e7a866d6c718a334.

vaibhavdaren:
Checking: 
```
tmt/steps/prepare/artifact/providers/__init__.py:252: error: Missing named argument "logger" for "retry_session"  [call-arg]
Found 1 error in 1 file (checked 131 source files)
````

Started failing after : https://github.com/teemtee/tmt/pull/4240

File: tmt/steps/prepare/artifact/providers/__init__.py [hunk 18]
```diff
@@ -195,6 +199,117 @@ def _filter_artifacts(self, exclude_patterns: list[Pattern[str]]) -> Iterator[Ar
...
+        Create a Repository instance by fetching content from a URL.
+
+        :param url: The URL to fetch the repository content from.
+        :param name: Optional name for the repository. If not provided,
+            derived from the URL.
+        :returns: A Repository instance.
+        :raises GeneralError: If fetching or parsing fails.
+        """
+        try:
+            with tmt.utils.retry_session() as session:
```
```suggestion
            with tmt.utils.retry_session(logger=logger) as session:
```
And add it to the required variables in `from_url`

vaibhavdaren: Added logger as a parameter in https://github.com/teemtee/tmt/pull/4205/commits/b4903d19e174acce5eebbac70718889b7c17416d.  The tests should pass.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -209,6 +209,8 @@ class Repository:
     content: str
     #: Uniquely identifiable name
     name: str
+    #: Logger
+    logger: tmt.log.Logger
```
LecrisUT: Nitpick: if a class is not action orientated it should not have a logger. If it is data oriented and primarily a wrapper it should be easier to read what it is contained in it. Would be better to log the specific methods instead as needed.

happz: A point for the ticket collecting assorted bits to address later, I guess?

LecrisUT: Don't see why it cannot be addressed quickly here. But if we want to get it quickly out of the gate, that is fine too

vaibhavdaren: Added  logger to creator methods. https://github.com/teemtee/tmt/pull/4205/commits/500c5a3c39209fe21d303a7810dc3d18fa16e458
I have added to all of the `from_*` functions for now. Even though they are not used at the moment.  It makes repository creation uniform.

LecrisUT: Thnx

happz: > Don't see why it cannot be addressed quickly here.

It surely can. I just don't think it will be quick :) @vaibhavdaren would gladly add a commit addressing your comment, a commit that just landed, but I could raise a comment about the order of parameters, for example, as the current way is "logger last" because that's what @psss preferred and we stick to it, so @vaibhavdaren would add yet another commit, reordering parameters, maybe change some methods to keyword only, and the wheel would roll and roll and roll while the Koji & Brew providers suffer the similar issue, and a PR that finally stabilized gets into another round of back and forth exchanges. Or I would not comment on the order of parameters, leaving it to some later moment, adding the topic to the TODO list, and the we would head into the code anyway...

> But if we want to get it quickly out of the gate, that is fine too

Yeah, I was hoping that would be the case. It was finally ready, silent, calm. Now there's new code, new changes, and new review :(

LecrisUT: If it is something blocker for me I would change the review to "request changes", if not take it as "it's fine, but let's keep track of this and address it later", and I leave it up to author to decide if they want to do a quick fix or let the CI flow.

As there was no scheduled merge, this didn't affect the merge flow (by the time we wake up tomorrow the CI would have finished), just the review flow. But if I should be more explicit about these intents, I will hapily adjust as needed.

happz: Please, be more explicit about what you expect from a developer. I will be more precise too; next time, instead of suggesting a nitpick, at this stage of PR development, could be just another point for the TODO list, I will clearly state it should be a TODO list item.

> As there was no scheduled merge, this didn't affect the merge flow (by the time we wake up tomorrow the CI would have finished), just the review flow.

Yes and no. This PR is the second PR in our little merge "queue", the first one - https://github.com/teemtee/tmt/pull/4250 - is running the last check. Once merged, this PR would have been rebased, auto-merge enabled, and by the time we wake up tomorrow, it would have been merged. Instead, it will require the rebase, because https://github.com/teemtee/tmt/pull/4250 will get in first, plus review, and then we will wait for its CI to complete.

This is the context you most likely did not have, but I did, so, yes, I will be more explicit about my intentions in the future. Because now there won't be two PRs off my table when I get to bed, just one. Added two more points to https://github.com/teemtee/tmt/issues/4221.

bajertom:
epel fails unrelated to this PR, required checks passed, merged

Issue #2681, opened 2024-02-12
Pull `TMT_PLAN_DATA` at the end of the `finish` step


lukaszachy:
Use case: run script in `finish` which writes data to `$TMT_PLAN_DATA` and use `login` (`tmt run -a login`)

We need to 'pull' data before login 'push' or keep existing data on the guest. Currently the 'login' deletes everything produced by finish (as no pull happens before)

psss:
Sounds like we just need to add `guest.pull()` at the end of the `finish` step. Reproducible easily with the following plan:

```yaml
provision:
    how: virtual
execute:
    how: tmt
    script: /bin/true
finish:
    script: echo hi > $TMT_PLAN_DATA/message
```

And this command:

```
tmt run -vvv login --command 'cat $TMT_PLAN_DATA/message'
```

Adding to the backlog.

psss:
This seems to be one of the scenarios to be covered in:

* #4067 

Adding as a subissue.

PR #4212, opened 2025-10-17
Fix: Prevent race condition between plugins and actions

vaibhavdaren:
The finish step was executing all its phases (plugins and actions like 'login') concurrently in a single queue.

This caused a race condition where an action (e.g., 'tmt run login --command "cat file"') could execute before the plugin (e.g., 'finish: how: shell, script: "echo > file"') had finished creating the file, leading to "No such file or directory" errors.

Refactor the 'finish.go()' method to use two separate queues:
1. Run all finish plugins first.
2. Wait for plugins to complete.
3. Pull artifacts from the guest ('sync_with_guests').
4. Run all finish actions (like 'login') last.

This ensures that when actions are executed, all plugin tasks have completed and their resulting files have been pulled to the host.

Try to fix: #2681

### DIFF
--- a/tmt/steps/finish/__init__.py
+++ b/tmt/steps/finish/__init__.py
@@ -1,10 +1,11 @@
 import copy
-from typing import TYPE_CHECKING, Any, Optional, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union, cast
 
 import click
 import fmf
 
 import tmt
+import tmt.queue
 import tmt.steps
 from tmt.container import container
 from tmt.options import option
@@ -12,6 +13,7 @@ from tmt.plugins import PluginRegistry
 from tmt.result import PhaseResult, ResultOutcome
 from tmt.steps import (
     Action,
+    ActionTask,
     Method,
     PhaseQueue,
     PluginOutcome,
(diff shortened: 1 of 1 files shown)


happz:
> The finish step was executing all its phases (plugins and actions like 'login') concurrently in a single queue.

True, but no steps runs *all* its phases - plugins and actions - concurrently *at the same time*. 

> This caused a race condition where an action (e.g., 'tmt run login --command "cat file"') could execute before the plugin (e.g., 'finish: how: shell, script: "echo > file"') had finished creating the file, leading to "No such file or directory" errors.

That is a user-issue: user can and should tell tmt when to run the `login` action, before or after a particular step, or even before and after a particular phase. See https://tmt.readthedocs.io/en/stable/examples.html#guest-login. Running a `login` to print a file created by `prepare` phase that did not run yet is user's fault.

At no point `prepare` step runs a `shell` phase **and** `login` action at the same time, they always run in sequence: first the `shell` phase runs on all guests as defined by its `when` key, then `login` runs - or the other way around, but never together.

```python
        for prepare_phase in self.phases(classes=(Action, PreparePlugin)):
            if isinstance(prepare_phase, Action):
                queue.enqueue_action(phase=prepare_phase)

            elif prepare_phase.enabled_by_when:
                queue.enqueue_plugin(
                    phase=prepare_phase,  # type: ignore[arg-type]
                    guests=[
                        guest for guest in guest_copies if prepare_phase.enabled_on_guest(guest)
                    ],
                )
```

Actions and plugins are enqueued, each as a distinct, separate task, and queue runs them in this order. A task's `go()` method is called, the method may yield zero or more task results, and when `go()` of a task finishes, `go()` of the next task is invoked. There is no concurrency at this level, no multithreading, pure sequence of tasks. The concurrency happens in the `MultiGuestTask.go()` - which is *plugin* task, never *action* task, and only this method uses a thread executor with a pool of worker threads, one for each given guest.

So this seems like a misunderstanding of how `login` and other actions fit into `tmt run` workflow rather than a race condition. Phases and actions never run at the same time, and their order is specified by user, by `order` keys and `login` subcommand's `--step` and `--when` options. It is absolutely possible to get the order wrong, and try to print out a file that has not been created yet. It's easy to to with `prepare` itself, just a question of ordering:

```yaml
# This won't work:
prepare:
  - how: shell
    script: echo /tmp/foo

  - how: shell
    script: touch /tmp/foo

# This will:
prepare:
  - how: shell
    order: 100
    script: echo /tmp/foo

  - how: shell
    order: 99
    script: touch /tmp/foo

# Or this:
prepare:
  - how: shell
    script: touch /tmp/foo

  - how: shell
    script: echo /tmp/foo

# And this is broken again:
prepare:
  - how: shell
    order: 100
    script: touch /tmp/foo

  - how: shell
    order: 99
    script: echo /tmp/foo
```

psss:
Alternative fix: #4223

vaibhavdaren:
Thanks for the detailed comment. @happz. 
Taken care by Petr in a separate PR.

PR #4218, opened 2025-10-20
Mock koji & brew tests

AthreyVinay:
Converting integration to mocks. This mostly started here: https://github.com/teemtee/tmt/pull/4202#discussion_r2439557024


tests/unit/artifact/test_brew.py
```diff
@@ -1,60 +1,107 @@
...
+            side_effect=lambda rpm: f"http://brew.example.com/{rpm['name']}.rpm",
+        ),
+    ):
+        yield
+
 
+@pytest.fixture
+def mock_call_api():
+    with patch.object(BrewArtifactProvider, "_call_api") as mock:
+        yield mock
```
LecrisUT, line 18: Please move to `conftest` and make the callers fixtures instead.

tests/unit/artifact/test_koji.py
```diff
@@ -1,44 +1,94 @@
...
+
+    with patch.object(koji_module, "koji", mock_koji):
+        provider = KojiArtifactProvider(f"koji.task:{task_id}", root_logger)
+        provider._top_url = "http://koji.example.com"
+        tasks = list(provider._get_task_children(task_id))
+
+        assert len(tasks) == 2
+        assert task_id in tasks
+        assert provider.build_id is None
+        assert len(provider.artifacts) == 2
```
LecrisUT: Technically ok to mock this part for unit tests, but the original will need to be properly reviewed.

But it does make me wonder, what exactly are we testing and trying to catch with these? Would it be more useful to make this an integration test (at least for the koji). We can use the cli to get a test build, although scratch-build would be tricky.

tests/unit/artifact/test_brew.py
AthreyVinay: Sure, updated, tks.

tests/unit/artifact/test_koji.py
AthreyVinay: Please see: https://github.com/teemtee/tmt/pull/4202#discussion_r2437233046

tests/unit/artifact/test_koji.py
```diff
@@ -1,49 +1,80 @@
...
 import pytest
 
+from tests.unit.artifact.conftest import (
+    MOCK_BUILD_ID,
+    MOCK_RPMS_KOJI,
+    create_mock_pathinfo,
+    mock_build_api_responses,
+    mock_call_api_for,
+    mock_task_api_responses,
+)
```
LecrisUT: Not needed. pytest automagically includes them.

tests/unit/artifact/test_koji.py
```diff
@@ -1,49 +1,80 @@
...
+from tmt.steps.prepare.artifact.providers import koji as koji_module
```
LecrisUT: Better to import `koji` directly. Also noticed that these tests should be gated by presence of `koji` and `brew` respectively: https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-on-a-missing-import-dependency. Fine if deferred.

tests/unit/artifact/test_koji.py
```diff
@@ -1,49 +1,80 @@
...
 
-@pytest.mark.skip(reason="would be replaced by mocks")
-@pytest.mark.integration
-def test_koji_valid_task_id_scratch_build(root_logger):
+def test_koji_valid_task_id_scratch_build(mock_koji, mock_call_api, root_logger):
     task_id = 137705547
-    provider = KojiArtifactProvider(f"koji.task:{task_id}", root_logger)
-    tasks = list(provider._get_task_children(task_id))
+    mock_pathinfo = create_mock_pathinfo()
+    mock_koji = MagicMock()
```
LecrisUT: Don't like how this one overwrites the previous parameter, but I also do not have enough experience with `MagicMock` (I usually use `flexmock`). Can't it be moved to the fixture though?

tests/unit/artifact/test_koji.py
LecrisUT: > But it does make me wonder, what exactly are we testing and trying to catch with these?

I mean more generally to make sure the mocking is as minimal as possible. So far as I am able to tell, all seem to be needed, except for maybe `PathInfo` which can go either way depending on if we allow to do the real `import koji` or not (if we do, we do not gain much from mocking it, or does it try to connect to the servers when it is called?)

As for having a proper integration test, everything under `_get_task_children` is not tested for any divergences, for which I guess we will rely on the user feedback for if anything blows up. But fine to look into it in the future as things evolve.

tests/unit/artifact/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+            task_id = args[0] if args else kwargs.get('taskID')
+            return {str(task_id): None, str(task_id + 1): None}
+        if method == "listTaskOutput":
+            return ["foo.rpm", "bar.rpm"]
+        return None
+
+    mock_call_api.side_effect = mock_api
+
+
+def create_mock_pathinfo():
```
LecrisUT: Can be a fixture.

tests/unit/artifact/test_koji.py
AthreyVinay: Sure, added a note in the TODO issue #4221

tests/unit/artifact/test_koji.py
AthreyVinay: AFAIK, pytest automatically imports `fixtures` ONLY and not constants or helper functions.

LecrisUT: You are right. Importing conftest is still not ideal though. I wanted to avoid moving the `mock_call_api` fixture, but probably not a simple way to do that. Can you move it and make 2 separate fixtures for those? It should be fine to have them depend on yet undefined fixtures `mock_brew` that are in the individual files. You could make some more magic with paremetrized fixtures or by reading what other fixtures are available alongside, but not sure how deep you want to research to make that happen.

AthreyVinay: I feel for something thats transient (with the topic of bringing back integration tests), the effort in this PR is sufficient. More efforts and time could be spent while actually refactoring this (again just what I feel) - WDYT?

LecrisUT: My main concern is with having to import `conftest` and navigating through the test as if it werw a top-level package. It feels counter to pytest typical design. It feels like it can break against refactoring either by us or by pytest.

Another approach with less work is to move these in a `utils.py` module and import it as `import utils` or `from .utils import`

tests/unit/artifact/test_koji.py
AthreyVinay: Absolutely - will be moved into the fixture!

tests/unit/artifact/test_koji.py
AthreyVinay: Yeah - at first that was my instinct to not to have imported from `conftest` simply for the above reason. But when I did a quick search in the codebase, the pattern was already present. But yeah - no reason to duplicate. Will move it to `utils` :)

happz: Util-like helpers are located in respective `__init__.py` files under `tests/unit`, you can spot several `from . import ...`. We can move them into new package, but let's do that in a standalone patch - if we want to add new helpers today, let's put them into one of the `__init__.py` we already use for that.

AthreyVinay: Cool - so used the best of both worlds: [fc097fd](https://github.com/teemtee/tmt/pull/4218/commits/fc097fda13402ea85c145a11c5fcc97773ed5870)

tests/unit/artifact/test_koji.py
AthreyVinay: Addressed here: [fc097fd](https://github.com/teemtee/tmt/pull/4218/commits/fc097fda13402ea85c145a11c5fcc97773ed5870)

tests/unit/artifact/conftest.py
AthreyVinay: Addressed here: [fc097fd](https://github.com/teemtee/tmt/pull/4218/commits/fc097fda13402ea85c145a11c5fcc97773ed5870)

tests/unit/artifact/test_koji.py
AthreyVinay: Happy to defer it as its noted in the todo.

tests/unit/artifact/test_koji.py
AthreyVinay: > let's put them into one of the __init__.py we already use for that.

Done.

Issue #4274, opened 2025-11-03
mrack: implement device-name from device specification


juk:
Please consider implementing a `device-name` field from the `device` specification [1] that maps to Beaker’s `Devices/Description`.

For example, to find a system with the correct CCP device, it’s necessary to specify a device name:
`device.device-name`: 'Genoa CCP'

[1] https://tmt.readthedocs.io/en/stable/spec/hardware.html#device

PR #4276, opened 2025-11-04
Add support `device.device_name` for `mrack`

skycastlelily:
Fixes:#4274

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/mrack.py
+++ b/tmt/steps/provision/mrack.py
@@ -198,6 +198,19 @@ class MrackHWKeyValue(MrackHWElement):
         self.attributes = {'_key': name, '_op': operator, '_value': value}
 
 
+@container(init=False)
+class MrackHWDeviceElement(MrackHWElement):
+    """
+    An element for device with op and description attributes
+    """
+
+    def __init__(self, operator: str, description: str) -> None:
+        super().__init__('device')
+
+        # Use underscore prefix for attributes in mrack
+        self.attributes = {'_op': operator, '_description': description}
+
+
 BeakerizedConstraint = Union[MrackBaseHWElement, dict[str, Any]]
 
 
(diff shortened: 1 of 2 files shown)


happz (approved):

tmt/steps/provision/mrack.py
```diff
@@ -548,6 +561,28 @@ def _transform_device_driver(
     return _transform_driver_helper("MODULE", constraint, logger)
 
 
+@transforms
+def _transform_device_device_name(
```
AthreyVinay, line 565: Nitptick: The naming is a bit awkward with `device` repeating itself (literally translating the dot notation). Why not just `_transform_device_name` instead or `transform_device_description` to match beaker's field?

happz: See the docstring for `@transforms`, the function name must reflect the HW requirement.

tmt/steps/provision/mrack.py
```diff
@@ -198,6 +198,19 @@ def __init__(self, name: str, operator: str, value: str) -> None:
         self.attributes = {'_key': name, '_op': operator, '_value': value}
 
 
+@container(init=False)
+class MrackHWDeviceElement(MrackHWElement):
```
AthreyVinay, line 202: While you are here, all three sub-classes basically set `self.attributes` - can this be handled by the constructor of the base class(`MrackHWElement`) and reduce redundancy?

tmt/steps/provision/mrack.py
AthreyVinay: Ack, guessed there would be a reason for this :)

tmt/steps/provision/mrack.py
skycastlelily: You mean remove MrackHWBinOp,  MrackHWKeyValue and the newly added MrackHWDeviceElement?

AthreyVinay: You could go aggressive and create a single configurable class - however for this, all I meant was setting the attributes could be handled by the constructir of base class  just making it more `DRY`er,  like so,
```rst
@container
class MrackHWElement(MrackBaseHWElement):
    ...

    def __init__(self, name: str, **attrs: str) -> None:
        self.name = name
        # Prefix keys with underscore
        self.attributes = {f"_{k}": v for k, v in attrs.items()}

    ...


@container(init=False)
class MrackHWBinOp(MrackHWElement):
    ...

    def __init__(self, name: str, operator: str, value: str) -> None:
        super().__init__(name, op=operator, value=value)


@container(init=False)
class MrackHWKeyValue(MrackHWElement):
    ...

    def __init__(self, key: str, operator: str, value: str) -> None:
        super().__init__('key_value', key=key, op=operator, value=value)


@container(init=False)
class MrackHWDeviceElement(MrackHWElement):
    ...

    def __init__(self, operator: str, description: str) -> None:
        super().__init__('device', op=operator, description=description)
```

skycastlelily: TBO, I don't think it's a good idea to remove that classes, that's why I asked for confirmation   and I don't think it's a good idea to modify the code like above either: The amount of "duplicate" code is tiny( one line each class),while it may loss code clarity and type safety, plus, it would mean we only have the init lines for that three classes which looks a little ugly.so I prefer keep the current code.

AthreyVinay: I agree the current duplication is small, but my goal with this change is to centralize how attributes are defined and keep key naming consistent across all HW elements. This makes it easier to extend later and change the prefix logic in one place if needed. The clarity of each subclass’s constructor remains high, and type safety isn’t impacted since each subclass still defines its own parameters.

> so I prefer keep the current code.

Sure :)

PR #4280, opened 2025-11-04
Add Polarion story export with test case linking and custom fields - features inside polarion

jscotka:
Features:
- Export tmt stories to Polarion as requirement work items
- Prevent duplicate features with automatic lookup using tmtid or extra-polarion
- Link verified-by test cases with 'verifies' relationship
- Automatically export missing test cases to Polarion
- Support for custom Polarion fields via extra-polarion-* prefix
- Canonical URL generation for test script hyperlinks
- Fallback mechanism for Polarion ID storage in FMF metadata

CLI Options:
- --project-id: Specify Polarion project
- --polarion-feature-id: Link to existing feature
- --duplicate/--no-duplicate: Allow/prevent duplicate features
- --export-linked-tests: Auto-export test cases (default: true)
- --link-polarion: Add back-references to FMF metadata
- --append-summary: Include in-story test links

Examples:
- Added examples/polarion-story/ with story and test examples
- Includes README with usage instructions

Documentation:
- Added polarion-inspect/ directory with diagnostic tools
- Example story with verified-by links
- Backward compatibility removal documentation

Testing:
- All features tested and working
- Duplicate prevention verified
- Token authentication documented

Generated-by: Cursor


tmt/export/polarion.py
```diff
@@ -17,30 +18,140 @@
...
+    canonical_repos = {
+        'tmt': 'https://github.com/teemtee/tmt',
+    }
+    
+    # Detect if this is a tmt repository (any fork)
+    if 'tmt.git' in url or '/tmt/' in url or url.endswith('/tmt'):
+        return canonical_repos['tmt']
+    
+    # If no canonical mapping found, return original
+    return url
```
KwisatzHaderach: unused and also not doing anything for any test repo (just tmt)

tmt/export/polarion.py
```diff
@@ -17,30 +18,140 @@
...
+            canonical_link
+        )
+        
+        # Replace tree/branch_name with blob/main for default branch
+        # This handles local feature branches like stories_polarion
+        canonical_link = re.sub(
+            r'/tree/[^/]+/',
+            '/blob/main/',
+            canonical_link
+        )
```
KwisatzHaderach: I wouldn't do this unless you make it universal (getting the fork source instead of just substituting hardcoded strings), right now this would do something only for this one specific repo, wouldn't work for package test repos

tmt/export/polarion.py
```diff
@@ -236,27 +346,19 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
-                    f"Can't add ID because {polarion_case.project_id} project "
-                    "doesn't have the 'tmtid' field defined.",
-                    fg='yellow',
-                )
-            )
-    if dry_mode or (polarion_case is not None and polarion_case.tmtid):
-        echo(style(f"Append the ID {uuid}.", fg='green'))
+        polarion_work_item_id = str(polarion_case.work_item_id)
+        with test.node as data:
+            data['extra-polarion'] = polarion_work_item_id
```
KwisatzHaderach: this field was something @psss never wanted to mention in tmt metadata and this was the base for the whole tmt id thing

tmt/export/polarion.py
```diff
@@ -414,9 +521,466 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+            f'<hr/>'
+        )
+    
+    if story.story:
+        description += markdown_to_html(story.story)
+    if story.description:
+        if description and not web_link:  # Only add spacing if no header was added
+            description += '<br/><br/>'
+        elif description:  # Header was added
+            description += '<br/>'
```
KwisatzHaderach: shouldn't one line be enough for both of these cases? even where there would be a web_link, then that part is before the story, so IMO doesn't matter here

tmt/export/polarion.py
```diff
@@ -414,9 +521,466 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+            elif isinstance(link.target, tmt.base.FmfId):
+                test_path = link.target.name
+                polarion_id = test_case_map.get(link.target.name)
+            
+            if polarion_id:
+                try:
+                    if not dry_mode:
+                        assert polarion_feature  # Narrow type
+                        # Use "verifies" role - the test case verifies the requirement
+                        polarion_feature.add_linked_item(polarion_id, 'verifies')
```
KwisatzHaderach: I think the relation is reversed, IMO it should be `verified by`, you are not verifying each test case by the feature

tmt/export/polarion.py
```diff
@@ -414,9 +521,466 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+                                    }
+                                    polarion_test_case = export_to_polarion(test, options=test_options)
+                                    if polarion_test_case:
+                                        logger.info(f'Test case exported: {polarion_test_case.work_item_id}')
+                                except Exception as export_err:
+                                    logger.debug(f"Failed to export test case: {export_err}")
+                                    logger.warning(f'Failed to export test case: {export_err}')
+                            
+                            if polarion_test_case:
+                                test_case_map[link.target.name] = str(polarion_test_case.work_item_id)
```
KwisatzHaderach: this looks about the same as the previous section, can we have helper function or just handle these together?

KwisatzHaderach (commented):
just couple small comments, but overall it looks good and might be useful for some teams even though the features/requirements should already be imported from JIRA into Polarion directly so import feature will be a bit more interesting

tmt/export/polarion.py
jscotka: I understand, but currently `tmtid` is not set as custom field for features inside polarion projects, so it can be removed as it will be updated to all projects.
So I suggest to add there text, it is workaround :-), what do you think @KwisatzHaderach

KwisatzHaderach: Sure, but IMO it's easy to ask Jenny for this field to be added into requirements as well.

pkis: Jenny left Red Hat. Just ask bmurray on the #forum-polarion-users what it takes adding the field to Polarion? I don't think it takes too much effort.

psss:
Thanks for working on this! Looks like a very nice enhancement for the story / requirements / traceability area. Tentatively proposing for Sprint 10.

jscotka:
Hi @psss @KwisatzHaderach I've fixed your PR's review comments. Now it should be fixed.

Especially

- verified-by for testcases to features (needed to chage the logic little bit (set it vice versa))
- using html markup function from utils 
- removed some duplications
- extra-polarion should be used just in case it is not set in project. (not tested well)



Also rebased the PR to main.

LecrisUT:
Please break down this PR into its pure essential, this is a huge review burden in the current state.
- The example does not fit well in the current structure and it is too explicit to polarion
- The `polarion-inspect` cli appears out of nowhere and it does not seem to be related to tmt itself
- The real changes in the `tmt/**` files in this PR are already a big enough burden to review. Just at a glance I see issues with how it use `echo` which we do not use anywhere else in the code. Please try to condense the real implementation as well

jscotka:
> Please break down this PR into its pure essential, this is a huge review burden in the current state.
> 
>     * The example does not fit well in the current structure and it is too explicit to polarion
> 
>     * The `polarion-inspect` cli appears out of nowhere and it does not seem to be related to tmt itself
> 
>     * The real changes in the `tmt/**` files in this PR are already a big enough burden to review. Just at a glance I see issues with how it use `echo` which we do not use anywhere else in the code. Please try to condense the real implementation as well

@LecrisUT you are completely right. 

I've expected to remove all these debugging stuff when review is done, part of squashing commits.
 - examples can be removed, as unnecessary, then we add external testst using requre inside tests repo.
 - polarion-inspect is more debugging tool when used cursor, not anyhow relevant TMT directly. But I expect to be useful for PR review for someone.

So in case just removing this code and change `click.echo` to something else, I can imagine partially move it into self.info and debug, b, I can do it.
What is preferred way how to display text to output, you mean using something like: `echo(style('title: ', fg='green') + text)`? instead of click.echo?

LecrisUT:
> * But I expect to be useful for PR review for someone.

This assumes that the reviewers know how and what you have defined there and for what reason. We do not and it is just noise. Even regarding polarion itself, we rely on the polarion code owner to verify the intent and functionality as we do not have the enough background on it.

> So in case just removing this code and change `click.echo` to something else,

That was not a task list, that was just the immediate thing that I can see. The thing that I requested is plainly

> Please break down this PR into its pure essential, this is a huge review burden in the current state.

Make it reviewable, make it simple, do not put noise, use what is available, and most importantly, ask US if you have issues, are not sure how to implement, etc. We want to avoid hundreds of back-and-forth while the "AI" is "learning" what we are saying

jscotka:
> > * But I expect to be useful for PR review for someone.
> 
> This assumes that the reviewers know how and what you have defined there and for what reason. We do not and it is just noise. Even regarding polarion itself, we rely on the polarion code owner to verify the intent and functionality as we do not have the enough background on it.
> 
> > So in case just removing this code and change `click.echo` to something else,
> 
> That was not a task list, that was just the immediate thing that I can see. The thing that I requested is plainly
> 
> > Please break down this PR into its pure essential, this is a huge review burden in the current state.
> 
> Make it reviewable, make it simple, do not put noise, use what is available, and most importantly, ask US if you have issues, are not sure how to implement, etc. We want to avoid hundreds of back-and-forth while the "AI" is "learning" what we are saying

Hi @LecrisUT ,
I've tried to fix it. remove debugging part, and cleanup logging and verbose output.

I understand your PoV from perspective of PR review. But my whole initialive at the beginning were PoC and demonstration how it works, so It needed to have some real useful example and something to be able to debug polarion and also for revirevers e.g. to be able to find part of polarion/pylero what should be changed to something else.

Now it should be ready for review I hope what do you think @psss?

     Regards
    Honza

LecrisUT:
> Hi @LecrisUT , I've tried to fix it. remove debugging part, and cleanup logging and verbose output.
> 
> I understand your PoV from perspective of PR review.

Thank you it is in a more reviewable state

> But my whole initialive at the beginning were PoC and demonstration how it works, so It needed to have some real useful example and something to be able to debug polarion and also for revirevers e.g. to be able to find part of polarion/pylero what should be changed to something else.

Understandable, but that was not a correct way to do so. Next time, if needed, please make a an example repo, to show case what you need and move auxiliary things in there.

tmt/cli/_root.py
```diff
@@ -1544,11 +1544,55 @@ def headfoot(text: str) -> None:
     choices=_story_export_formats,
 )
 @option(
-    '-d',
-    '--debug',
```
LecrisUT: You have removed the `--debug` flag. Please revert

tmt/cli/_root.py
```diff
@@ -1544,11 +1544,55 @@ def headfoot(text: str) -> None:
...
+    '--project-id',
+    help='Use specific Polarion project ID.',
+)
+@option(
+    '--polarion-feature-id',
+    help='Polarion feature/requirement work item ID to link the story to.',
+)
+@option(
+    '--link-polarion / --no-link-polarion',
+    default=False,
```
LecrisUT: Do not add the options directly via `click`. We have a mechanism as shown in `tmt run` to define the options in something like `StepData` which are gated by the `--how`. What needs to be done instead is to replicate that design to `tmt.export`.

tmt/cli/_root.py
```diff
@@ -1570,6 +1614,10 @@ def stories_export(
     unverified: bool,
     undocumented: bool,
     uncovered: bool,
+    project_id: Optional[str],
+    link_polarion: bool,
+    append_summary: bool,
+    create: bool,
```
LecrisUT: Unused

tmt/export/polarion.py
```diff
@@ -17,30 +17,100 @@
...
+        relative_test_path = test.fmf_id.path / relative_test_path
+    
+    # Construct the web URL using the git utilities
+    return tmt.utils.git.web_git_url(test.fmf_id.url, test.fmf_id.ref, Path('/') / relative_test_path)
+
+
 def import_polarion() -> None:
     """
     Import polarion python api - pylero
     """
```
LecrisUT, line 36: Please keep this at the top for visibility

tmt/export/polarion.py
```diff
@@ -17,30 +17,100 @@
...
+    - ./test.sh → same directory as FMF file
+    - ../test.sh → one directory up
+    - ../../runtest.sh → two directories up
+    """
+    if not test.test or not test.fmf_id.url:
+        return None
+    
+    # Get the actual metadata file path from node.sources (last source is most specific)
+    if not test.node.sources:
+        return None
```
LecrisUT: Why the `return none`? When are each of these expected outcomes?

tmt/export/polarion.py
```diff
@@ -17,30 +17,100 @@
...
 RE_POLARION_URL = r'.*/polarion/#/project/.*/workitem\?id=(.*)'
-LEGACY_POLARION_PROJECTS = {'RedHatEnterpriseLinux7'}
 
 # TODO: why this exists?
 log = fmf.utils.Logging('tmt').logger
 
 
+def get_test_script_link(test: tmt.base.Test) -> Optional[str]:
+    """
+    Generate a web link to the actual test script (not the metadata file)
```
LecrisUT: What if `test` is an in-line script, here are two cases where this can fail
```yaml
/test_in_line:
  test: |
    cd test_dir
    pytest

/test_as_file:
  test: $TEST_CASE/test.sh
```

There is an issue about how/if we can make special cases for detecting if a test is a script, and there were too many edge-cases to figure out it didn't seem feasible.

tmt/export/polarion.py
```diff
@@ -17,30 +17,100 @@
...
+    test_script_path = (metadata_dir / test_command).resolve()
+    
+    # Check if the resolved test script exists locally
+    if not test_script_path.exists():
+        # If file doesn't exist, fall back to original web_link behavior
+        return test.web_link()
+    
+    # Get the actual git repository root (not FMF root, which may be nested)
+    # FMF root is where .fmf/ directory is, git root is where .git/ directory is
+    git_root_path = tmt.utils.git.git_root(fmf_root=Path(test.node.root), logger=test._logger)
```
LecrisUT: The way the git root is taken does not feel right, but I can't put my finger on what it should be. For example, what if the test metadata was given from a `DiscoverShell` or in the `execute` step?

tmt/export/polarion.py
```diff
@@ -429,9 +523,595 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+    except (AttributeError, PolarionException) as exc:
+        logger.debug(f"Failed to update feature: {exc}")
+        raise ConvertError(f"Failed to update Polarion feature: {exc}") from exc
+    
+    # Final success message - keep this as user-facing output
+    echo(style(f"Story '{summary}' successfully exported to Polarion.", fg='magenta'))
 
 
 @tmt.base.Test.provides_export('polarion')
+@tmt.base.Story.provides_export('polarion')
```
LecrisUT, line 1120: Iiuc much of this PR is to handled the story -> test linkage. Is it possible to split this into just exporting story and then adding the story -> test linkage? It would make the review quicker.

tmt/utils/__init__.py
```diff
@@ -3645,26 +3645,43 @@ def json_to_list(data: Any) -> list[Any]:
...
 
-    Expects: Markdown document as a file.
+    Expects: Markdown document as a file path or text string.
     Returns: An HTML document as a string.
+    
+    Args:
+        source: Either a Path to a markdown file or a markdown text string
+        extensions: Optional list of markdown extensions to enable
+                   (e.g., ['extra', 'nl2br', 'sane_lists'])
     """
```
LecrisUT, line 3659: The docstring format does not follow the `sphinx` format that we use.

tmt/utils/__init__.py
```diff
@@ -3645,26 +3645,43 @@ def json_to_list(data: Any) -> list[Any]:
     return loaded_data
 
 
-def markdown_to_html(filename: Path) -> str:
+def markdown_to_html(
+        source: Path | str,
```
LecrisUT: I do not like the change to allow `str`, especially if those are just snippets as it's done here because the html produced would not be consistent depending on the context around the snippet. Instead it would be preferred to finalize a whole markdown file (even if it a `mktemp` file) and convert that into html

tmt/export/polarion.py
```diff
@@ -429,9 +523,595 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+
+    # Update Polarion feature with all changes
+    try:
+        polarion_feature.update()
+    except (AttributeError, PolarionException) as exc:
+        logger.debug(f"Failed to update feature: {exc}")
+        raise ConvertError(f"Failed to update Polarion feature: {exc}") from exc
+    
+    # Final success message - keep this as user-facing output
+    echo(style(f"Story '{summary}' successfully exported to Polarion.", fg='magenta'))
```
LecrisUT: We don't do `echo`. Use `logger.info`. See other calls to `logger.info` or its implementation for more context on how to convert this.

tmt/export/polarion.py
```diff
@@ -429,9 +523,595 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+    web_link = story.web_link()
+    
+    # Add header indicating TMT autogeneration with link to metadata
+    if web_link:
+        # Use the actual origin repository URL without any hardcoded modifications
+        description = (
+            f'<p><em>ℹ️ This feature is autogenerated by TMT from metadata: '
+            f'<a href="{web_link}">{story.name}</a></em></p>'
+            f'<hr/>'
+        )
```
LecrisUT, line 667: -1 to these. You are using markdown_to_html a bit lower down. Construct the whole thing as a markdown format with either jinja or pythong string format and make that into an html (unless you cannot have polarion do the conversion instead)

tmt/export/polarion.py
```diff
@@ -429,9 +523,595 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+    
+    # Markdown extensions for better formatting in Polarion
+    markdown_extensions = [
+        'extra',           # Adds tables, fenced code blocks, etc.
+        'nl2br',           # Converts newlines to <br/>
+        'sane_lists',      # Better list handling
+    ]
```
LecrisUT: Please elaborate each of these extensions with examples on how enabling them changes the output. The question is why not make all of these extensions the default.

LecrisUT (changes requested):
It is still a lot to go through, so I only did a preliminary review

tmt/cli/_root.py
jscotka: should be replaced by adding @verbosity_options instead of direct debug option, shouldn't

tmt/cli/_root.py
jscotka: I'll look at this mechanism. Thank for pinpointing.

tmt/export/polarion.py
jscotka: yes, this is hard part, but at least current mechanism were very fragile, as it does not work for my more less easy cases. as I think I were using `../runtest.sh` and it was not able correctly resolve this path.

I understand you PoV, question is if some improvement is better than current situation, what resolved it in bad way.

tmt/cli/_root.py
LecrisUT: Hmm indeed so. Can you make a separate PR for this. It seems to be lacking a
```python
    logger = context.obj.logger.descend(logger_name='export', extra_shift=0)
    logger.apply_verbosity_options(**kwargs)
```
to be properly picked up.

tmt/export/polarion.py
jscotka: I have no idea, this is little bit tricky part, and question is how we can deal with these dynamically defined "tests" , what do you think?

tmt/cli/_root.py
LecrisUT: Note that this would not be easy, and please make a dedicated PR for this. It should not have any functional change initially. You can also deffer it to the team to do that part if it turns out too difficult.

tmt/export/polarion.py
jscotka: yep, I can create some template instead of direct formatting, this is about top level parts, not about markdown. After experimenting with polarion, there were many issues when I tried to do formatting. So maybe some template there will be  better, sure.

tmt/export/polarion.py
jscotka: Not sure if make it default  could be better. This is more less backward compatible and improve just part what I affected.

tmt/export/polarion.py
LecrisUT: I would step one step backwards. Do you really need the link to the test file, or can you instead link to the `.fmf` file that defines the test and provide the raw `test` that was provided. The test file by itself would not be useful if for example there are `environment` that it depends on being set.

tmt/export/polarion.py
jscotka: okay, I can use logger.info directly, I though I saw somewhere echoing, expecially in these export parts. But probably my fault there.

LecrisUT: You are correct and it should be fixed

<details>

```console
$ grep -R "echo(" tmt/
tmt/export/__init__.py:                echo(style(f"Failed to set qe_test_coverage+ flag for BZ#{bug_id}", fg='red'))
tmt/export/__init__.py:                echo(style(f"Failed to link to BZ#{bug_id}", fg='red'))
tmt/export/__init__.py:    echo(
tmt/export/nitrate.py:                        echo(
tmt/export/nitrate.py:        echo(style(f'Markdown file doesn\'t exist {fail_message}', fg='yellow'))
tmt/export/nitrate.py:    echo(style(f'{len(md_files)} Markdown files found {fail_message}', fg='yellow'))
tmt/export/nitrate.py:    echo(style(f"Test case '{testcase.identifier}' created.", fg='blue'))
tmt/export/nitrate.py:                echo(style(f"Link to plan '{child_plan}'.", fg='magenta'))
tmt/export/nitrate.py:                echo(style(f"Link to run '{testrun}'.", fg='magenta'))
tmt/export/nitrate.py:            echo(style(f"Exporting regardless '{error_msg}'.", fg='red'))
tmt/export/nitrate.py:        echo(style(f"Test case '{nitrate_case.identifier}' found.", fg='blue'))
tmt/export/nitrate.py:                    echo(style(f"Test case '{extra_summary}' created.", fg='blue'))
tmt/export/nitrate.py:                echo(style("Append the nitrate test case id.", fg='green'))
tmt/export/nitrate.py:    echo(style('summary: ', fg='green') + summary)
tmt/export/nitrate.py:        echo(style('script: ', fg='green') + test.node.get('extra-task'))
tmt/export/nitrate.py:        echo(style('components: ', fg='green') + ' '.join(test.component))
tmt/export/nitrate.py:                echo(style(f"Failed to add component '{component}'.", fg='red'))
tmt/export/nitrate.py:                    echo(style(f"Linked to general plan '{general_plan}'.", fg='magenta'))
tmt/export/nitrate.py:                    echo(style(f"Failed to find general test plan for '{component}'.", fg='red'))
tmt/export/nitrate.py:                echo(style(f"Removed general plan '{nitrate_plan}'.", fg='red'))
tmt/export/nitrate.py:    echo(style('tags: ', fg='green') + ' '.join(set(test.tag)))
tmt/export/nitrate.py:            echo(style('default tester: ', fg='green') + email_address)
tmt/export/nitrate.py:    echo(style('estimated time: ', fg='green') + test.duration)
tmt/export/nitrate.py:    echo(style('automated: ', fg='green') + ['auto', 'manual'][test.manual])
tmt/export/nitrate.py:        echo(style('status: ', fg='green') + 'CONFIRMED')
tmt/export/nitrate.py:        echo(style('status: ', fg='green') + 'DISABLED')
tmt/export/nitrate.py:        echo(style('status: ', fg='green') + str(current_status))
tmt/export/nitrate.py:        echo(style('arguments: ', fg='green') + environment)
tmt/export/nitrate.py:        echo(style('arguments: ', fg='green') + "' '")
tmt/export/nitrate.py:    echo(style('Structured Field: ', fg='green'))
tmt/export/nitrate.py:            echo(style(section + ': ', fg='green') + attribute.strip())
tmt/export/nitrate.py:    echo(style('fmf id:\n', fg='green') + fmf_id.strip())
tmt/export/nitrate.py:        echo(style('Add migration warning to the test case notes.', fg='green'))
tmt/export/nitrate.py:    echo(style(f"Append the ID {uuid}.", fg='green'))
tmt/export/nitrate.py:        echo(style("manual steps:", fg='green') + f" found in {md_path}")
tmt/export/nitrate.py:        echo(
tmt/export/nitrate.py:        echo(
tmt/export/polarion.py:        echo(style(f"Test case '{polarion_case.work_item_id!s}' found.", fg='blue'))
tmt/export/polarion.py:    echo(style(f"Test case '{testcase.work_item_id}' created.", fg='blue'))
tmt/export/polarion.py:            echo(style(f"Exporting regardless '{error_msg}'.", fg='red'))
tmt/export/polarion.py:                echo(style(f"Test case '{summary}' created.", fg='blue'))
tmt/export/polarion.py:        echo(style('title: ', fg='green') + test.summary)
tmt/export/polarion.py:            echo(
tmt/export/polarion.py:        echo(style(f"Append the ID {uuid}.", fg='green'))
tmt/export/polarion.py:    echo(style('description: ', fg='green') + description)
tmt/export/polarion.py:    echo(style('script: ', fg='green') + automation_script)
tmt/export/polarion.py:    echo(style('components: ', fg='green') + ' '.join(test.component))
tmt/export/polarion.py:    echo(style('tags: ', fg='green') + ' '.join(set(test.tag)))
tmt/export/polarion.py:            echo(style('default tester: ', fg='green') + login_name)
tmt/export/polarion.py:    echo(style('enabled: ', fg='green') + str(test.enabled))
tmt/export/polarion.py:    echo(style("Append the Polarion test case link.", fg='green'))
tmt/export/polarion.py:    echo(style(f"Test case '{summary}' successfully exported to Polarion.", fg='magenta'))
tmt/steps/discover/__init__.py:                click.echo(''.join(export_fmf_ids), nl=False)
tmt/steps/discover/shell.py:            click.echo(tmt.utils.format('tests', [test.name for test in self.data.tests]))
tmt/steps/provision/__init__.py:                echo(tmt.utils.format('hardware', tmt.utils.to_yaml(hardware.to_spec())))
tmt/steps/provision/bootc.py:            click.echo(tmt.log.indent(data.image, level=2))
tmt/steps/provision/testcloud.py:            click.echo(f"{store_dir / filename}")
tmt/steps/__init__.py:        echo(
tmt/steps/__init__.py:            echo(tmt.utils.format(key_to_option(key), value))
tmt/utils/__init__.py:    echo(text, **kwargs)
tmt/utils/__init__.py:    echo(message)
tmt/utils/__init__.py:    echo(message)
tmt/base.py:            echo(tmt.utils.format('id', self.id, key_color='magenta'))
tmt/base.py:        echo(tmt.utils.format('sources', self.fmf_sources, key_color='magenta'))
tmt/base.py:            echo(tmt.utils.format('web', web_link, key_color='magenta', wrap=False))
tmt/base.py:        echo(
tmt/base.py:        echo(style(self.name, fg='red'))
tmt/base.py:            echo(tmt.utils.format('summary', self.summary))
tmt/base.py:        echo(
tmt/base.py:            echo(tmt.utils.format('instructions', instructions))
tmt/base.py:                echo(
tmt/base.py:                echo(
tmt/base.py:                echo(tmt.utils.format(key, value.value))
tmt/base.py:                echo(tmt.utils.format(key, value))
tmt/base.py:                    echo(tmt.utils.format(key, value, key_color='blue'))
tmt/base.py:        echo(style("Lint checks on all", fg='red'))
tmt/base.py:        echo(
tmt/base.py:            echo(tmt.utils.format('description', self.description, key_color='green'))
tmt/base.py:            echo(tmt.utils.format('author', self.author, key_color='green'))
tmt/base.py:            echo(tmt.utils.format('contact', self.contact, key_color='green'))
tmt/base.py:            echo(tmt.utils.format('environment', self.environment, key_color='blue'))
tmt/base.py:            echo(
tmt/base.py:        echo(tmt.utils.format('enabled', self.enabled, key_color='cyan'))
tmt/base.py:            echo(tmt.utils.format('order', self.order, key_color='cyan'))
tmt/base.py:            echo(tmt.utils.format('id', self.id, key_color='cyan'))
tmt/base.py:            echo(tmt.utils.format('tag', self.tag, key_color='cyan'))
tmt/base.py:            echo(tmt.utils.format('tier', self.tier, key_color='cyan'))
tmt/base.py:                echo(tmt.utils.format('import', '', key_color='blue'))
tmt/base.py:                    echo(tmt.utils.format(key, value, key_color='green'))
tmt/base.py:        echo(
tmt/base.py:                echo(tmt.utils.format(key, value, wrap=False))
tmt/base.py:                echo(tmt.utils.format(key, value, wrap=wrap))
tmt/base.py:        echo(self)
tmt/base.py:                echo(f"The fmf tree root '{tree.root}/.fmf' already exists.")
tmt/base.py:                echo(f"Path '{path}' already has a parent fmf tree root '{tree.root}/.fmf'.")
tmt/base.py:                echo(f"Would initialize the fmf tree root '{fmf_dir}'.")
tmt/base.py:                echo(f"Initialized the fmf tree root '{fmf_dir}'.")
tmt/base.py:                echo(f"Path '{fmf_dir}' would be added to git index.")
tmt/base.py:                    echo(f"Path '{fmf_dir}' added to git index.")
tmt/base.py:                    echo(message)
tmt/base.py:            echo(f"Use 'tmt init --template' with {choices} to create example content.")
tmt/base.py:            echo(
tmt/base.py:            echo(f"Applying template '{template}'.")
tmt/base.py:        echo(run_status, nl=False)
tmt/base.py:        echo(run.run_workdir)
tmt/base.py:                echo(self.colorize_column(self.pad_with_spaces(plan_status)), nl=False)
tmt/base.py:                echo(f'{run.run_workdir}  {plan.name}')
tmt/base.py:                    echo(self.colorize_column(column), nl=False)
tmt/base.py:                echo(f' {run.run_workdir}  {plan.name}')
tmt/base.py:        echo(style(header, fg='blue'))
tmt/base.py:            echo(
tmt/convert.py:        echo(f"Importing the '{directory}' test case.")
tmt/convert.py:        echo(style(f"Test case successfully stored into '{path}'.", fg='magenta'))
tmt/convert.py:    echo(style(f'{type_}: ', fg='green') + new_link[type_])
tmt/convert.py:    echo(style('task: ', fg='green') + beaker_task)
tmt/convert.py:        echo(style('summary: ', fg='green') + data['summary'])
tmt/convert.py:        echo(style('test: ', fg='green') + data['test'])
tmt/convert.py:        echo(style("framework: ", fg="green") + data["framework"])
tmt/convert.py:        echo(style('contact: ', fg='green') + data['contact'])
tmt/convert.py:            echo(style('component: ', fg='green') + ' '.join(data['component']))
tmt/convert.py:        echo(style('duration: ', fg='green') + data['duration'])
tmt/convert.py:            echo(style('environment:', fg='green'))
tmt/convert.py:            echo(format_value(data['environment']))
tmt/convert.py:        echo(style('require: ', fg='green') + ' '.join(data['require']))
tmt/convert.py:        echo(style('recommend: ', fg='green') + ' '.join(data['recommend']))
tmt/convert.py:                echo(style("tag: ", fg="green") + " ".join(tags))
tmt/convert.py:    echo(f"Checking the '{path}' directory.")
tmt/convert.py:            echo(style('Restraint file ', fg='blue'), nl=False)
tmt/convert.py:            echo(style('Makefile ', fg='blue'), nl=False)
tmt/convert.py:        echo(style('Makefile ', fg='blue'), nl=False)
tmt/convert.py:        echo(style('Restraint ', fg='blue'), nl=False)
tmt/convert.py:        echo(f"found in '{datafile_path}'.")
tmt/convert.py:            echo(style("warn: Extra lines detected in the 'run' target:", fg="yellow"))
tmt/convert.py:                echo(f"    {line}")
tmt/convert.py:            echo(style("warn: Multiple lines detected in the 'build' target:", fg="yellow"))
tmt/convert.py:                echo(f"    {line}")
tmt/convert.py:        echo(style('Purpose ', fg='blue'), nl=False)
tmt/convert.py:            echo(f"found in '{purpose_path}'.")
tmt/convert.py:            echo(style('description:', fg='green'))
tmt/convert.py:            echo(data['description'].rstrip('\n'))
tmt/convert.py:            echo("not found.")
tmt/convert.py:    echo(style('Nitrate ', fg='blue'), nl=False)
tmt/convert.py:        echo(
tmt/convert.py:        echo(f"Multiple test cases found for '{beaker_task}'.")
tmt/convert.py:            echo(style(f"Test case would be stored into '{md_path}'.", fg='magenta'))
tmt/convert.py:            echo(style(f"Test case file '{md_path}' successfully removed.", fg='magenta'))
tmt/convert.py:        echo(style('tier: ', fg='green') + str(data['tier']))
tmt/convert.py:            echo(style('test: ', fg='green') + common_data['test'])
tmt/convert.py:        echo(
tmt/convert.py:    echo(style('Polarion ', fg='blue'), nl=False)
tmt/convert.py:        echo(style('summary: ', fg='green') + current_data['summary'])
tmt/convert.py:            echo(style('environment variables:', fg='green'))
tmt/convert.py:            echo(format_value(current_data['environment']))
tmt/convert.py:        echo(style('description: ', fg='green') + current_data['description'])
tmt/convert.py:        echo(style('enabled: ', fg='green') + str(current_data['enabled']))
tmt/convert.py:        echo(style('contact: ', fg='green') + current_data['contact'])
tmt/convert.py:                    echo(
tmt/convert.py:    echo(style('ID: ', fg='green') + uuid)
tmt/convert.py:        echo(style('component: ', fg='green') + ' '.join(current_data['component']))
tmt/convert.py:        echo(style('tag: ', fg='green') + ' '.join(current_data['tag']))
tmt/convert.py:    echo(f"test case found '{testcase.identifier}'.")
tmt/convert.py:        echo(style('extra-summary: ', fg='green') + data['extra-summary'])
tmt/convert.py:        echo(style('contact: ', fg='green') + data['contact'])
tmt/convert.py:            echo(style('environment:', fg='green'))
tmt/convert.py:            echo(format_value(data['environment']))
tmt/convert.py:        echo(style('tag: ', fg='green') + str(data['tag']))
tmt/convert.py:                        echo(f"Adding component '{component}' from the linked general plan.")
tmt/convert.py:    echo(style('component: ', fg='green') + ' '.join(data['component']))
tmt/convert.py:    echo(style('enabled: ', fg='green') + str(data['enabled']))
tmt/convert.py:        echo(style('adjust:', fg='green'))
tmt/convert.py:        echo(tmt.utils.to_yaml(data['adjust']).strip())
tmt/convert.py:            echo(style('extra-pepa:', fg='green'))
tmt/convert.py:            echo(data['extra-pepa'].rstrip('\n'))
tmt/convert.py:            echo(style('extra-hardware:', fg='green'))
tmt/convert.py:            echo(data['extra-hardware'].rstrip('\n'))
tmt/convert.py:                    echo(
tmt/convert.py:                    echo(
tmt/convert.py:        echo(style(f"Metadata successfully stored into '{path}'.", fg='magenta'))
```

</details>

We do not want any new ones to be used moving forward though

tmt/export/polarion.py
jscotka: Theoretically yes, practically I'm not sure if I'll able to split it. But this is good point. ad create story and default linking of testcases when exporting could be splitted. What do you think @psss or @therazix .

My usage and reason were also that I need something working and fully implementing stories with testcase coverage, as just story is empty part. 
Not sure If I have enough time to be able to do this splitting.

tmt/export/polarion.py
LecrisUT: Please try to make the template a markdown file. If you have issues with doing that, please show where and how that would be the case so we can understand the situation and provide assistence if needed. E.g. the above can be rewritten as
```markdown
ℹ️ _This feature is autogenerated by TMT from metadata:_ [{{ story.name }}]({{ story.web_link() }})
```

tmt/export/polarion.py
LecrisUT: > Please elaborate each of these extensions with examples on how enabling them changes the output.

Still need :point_up: to properly review and understand the why. If it can be avoided altogether by better construction of markdown text then all the better, but can't properly review without the context of why these were added.

tmt/export/polarion.py
jscotka: > I would step one step backwards. Do you really need the link to the test file, or can you instead link to the `.fmf` file that defines the test and provide the raw `test` that was provided. The test file by itself would not be useful if for example there are `environment` that it depends on being set.

It is harder to reply, link to fmf file is fine, but when I've talked with Jenny Severance as polarion expert for customer focus. She said, it is bad, and there should be link to test, not to metadata, to be able to show test to stakeholeders if necessary. And In case stakeholders are not familiar with FMF/TMT, as it is not so relevant for them, it is hard to understand what to search inside FMF files and how to construct link to test.

tmt/cli/_root.py
jscotka: > Note that this would not be easy, and please make a dedicated PR for this. It should not have any functional change initially. You can also deffer it to the team to do that part if it turns out too difficult.

That was also my idea at all, that someone from team help to finalize this PR, as I have not enough time to clean it very well, as you said it is big, as this was for me PoC that it is possible and very useful inside TMT, and also I do not have sometimes so deep knowledge of the code, how to do it well, or what is preferred way inside code. So I can help with TMT development, as much as possible, as I seen TMT as crucial and important project and we inside cockpit team using TMT as much as possible and I'm adding features into FMF or TMT projects when necessary, but still I cannot fully invest my time into TMT.

tmt/export/polarion.py
```diff
@@ -33,19 +37,43 @@ def import_polarion() -> None:
...
+    consistently for all Markdown-to-HTML conversions in this module.
+
+    :param text: Markdown source text.
+    :returns: an HTML string.
+    """
+    with tempfile.NamedTemporaryFile(
+        mode='w', suffix='.md', delete=False, encoding='utf-8'
+    ) as tmp:
+        tmp_path = tmt.utils.Path(tmp.name)
+        tmp_path.write_text(text)
```
thrix: This writes a string to a temp file only to have `markdown_to_html()` read it back. Since `markdown.markdown()` accepts strings directly, the temp file round-trip is unnecessary.

Simpler approach:

```python
import markdown

def _markdown_str_to_html(text: str) -> str:
    return markdown.markdown(text, extensions=tmt.utils.DEFAULT_MD_HTML_EXTENSIONS)
```

Or if you prefer keeping the shared code path, add a string-accepting helper to `tmt/utils/__init__.py` alongside the existing file-based `markdown_to_html()`.

tmt/export/polarion.py
```diff
@@ -93,34 +125,34 @@ def find_polarion_case_ids(
...
-        query_result = PolarionWorkItem.query(
-            f"tcmscaseid:{nitrate_case_id}", fields=wanted_fields
-        )
-        case_id, project_id = get_polarion_ids(query_result, preferred_project)
-
-    # Search by extra task
-    if not project_id and data.get('extra-task'):
-        query_result = PolarionWorkItem.query(data.get('extra-task'), fields=wanted_fields)
-        case_id, project_id = get_polarion_ids(query_result, preferred_project)
+    # Search by TMT ID (mapped to Polarion's tmtid custom field)
```
thrix, line 115: The old code searched by UUID (plain query), TCMS Case ID (`extra-nitrate`), and `extra-task`. These are all removed, which breaks backward compatibility for users who relied on them to find existing Polarion cases.

Consider keeping the old search methods as additional fallbacks after the new `tmtid` and `extra-polarion` searches:

```python
# 1. Direct polarion_case_id (explicit override)
# 2. TMT ID via tmtid custom field (new, preferred)
# 3. extra-polarion field (new fallback)
# 4. UUID plain query (legacy, kept for backward compat)
# 5. extra-nitrate TCMS Case ID (legacy)
# 6. extra-task (legacy)
```

If removal is intentional, it should be documented in the PR description and release notes as a breaking change.

tmt/export/polarion.py
```diff
@@ -430,9 +479,643 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+        logger.debug(f"tags: {' '.join(set(story.tag))}")
+    if login_name:
+        logger.debug(f"assignee: {login_name}")
+
+    # Early return if dry mode - don't actually update Polarion
+    if dry_mode:
+        return
+
+    # Apply all changes to Polarion (only if not dry mode)
+    try:
```
thrix, line 740: `story.tag.append('fmf-export')` mutates the original story's tag list in place. This side effect persists on the story object and could affect other code using it after this function returns.

Fix by working on a copy:

```python
tags = list(story.tag) if story.tag else []
tags.append('fmf-export')
if not dry_mode:
    assert polarion_feature
    try:
        polarion_feature.tags = ' '.join(tags)
    except (AttributeError, PolarionException) as exc:
        logger.debug(f"Failed to set tags: {exc}")
logger.debug(f"tags: {' '.join(set(tags))}")
```

tmt/export/polarion.py
```diff
@@ -430,9 +479,643 @@ def export_to_polarion(test: tmt.base.Test) -> None:
...
+            try:
+                polarion_feature.add_assignee(login_name)
+            except (AttributeError, PolarionException) as err:
+                logger.debug(f"Failed to set assignee: {err}")
+
+        # Custom Polarion fields from extra-polarion-* metadata
+        logger.debug('Setting custom Polarion fields')
+        set_polarion_custom_fields(polarion_feature, story.node, dry_mode=False)
+
+    except Exception as exc:
```
thrix: This `except Exception` wraps the entire field-setting block (title, description, priority, tags, assignee, custom fields) and logs at DEBUG level only. If, say, setting the title fails due to a programming error, the user gets no indication.

The individual try/except blocks inside (for priority, tags, assignee) already handle Polarion-specific errors gracefully. This outer catch-all is redundant and hides real bugs.

Suggestion: remove the outer `try/except Exception` and let the individual field handlers manage their own errors. If you want a safety net, at minimum log at WARNING:

```python
except Exception as exc:
    logger.warning(f"Failed to set some Polarion fields: {exc}")
```

tmt/export/polarion.py
```diff
@@ -9,19 +10,22 @@
...
 import tmt.utils
+from tmt.container import container, field
 from tmt.identifier import ID_KEY, add_uuid_if_not_defined
 from tmt.utils import ConvertError, Path
 from tmt.utils.themes import style
 
 PolarionException: Any = None
 PolarionTestCase: Any = None
 PolarionWorkItem: Any = None
+PolarionRequirement: Any = None
```
thrix, line 24: Nit: `LEGACY_POLARION_PROJECTS` and its filtering logic were removed. Previously, results from `RedHatEnterpriseLinux7` were deprioritized when multiple cases were found. If this is intentional, worth a note in the PR description so the change is visible to existing users.

thrix (changes requested):
Nice work on bringing story export to Polarion - this is a valuable addition for the traceability story. A few things to address before merge:

**Must fix:**
1. **`story.tag` mutation** - `story.tag.append('fmf-export')` modifies the original list in place, side effect that leaks outside the function
2. **`_markdown_str_to_html` temp file** - writing a string to a temp file just to read it back is unnecessary, `markdown.markdown()` accepts strings directly
3. **Broad `except Exception` in `_set_polarion_feature_fields`** - silently swallows all errors at DEBUG level, hiding real bugs from the user

**Should address:**
4. **Backward-incompatible search removal** - `extra-nitrate`, `extra-task`, and UUID plain query searches were dropped without documentation. Existing users relying on these to find Polarion cases will break. Either preserve as fallbacks or document as a breaking change.
5. **`LEGACY_POLARION_PROJECTS` removal** - undocumented behavior change, worth a note in the PR description
6. **No automated tests** - 823 lines of new code with zero test coverage. Unit tests for the core functions (`find_polarion_case_ids`, `get_polarion_feature`, `set_polarion_custom_fields`) would add confidence.

The dynamic CLI option collection via `_data_class` on `ExportPlugin` is a clean pattern. The `prepare_extra_summary` type broadening to `Core` is correct.

tmt/export/polarion.py
jscotka: I'll fix it, it was not so intentional, but part of cleanup of using extra-* fields.

jscotka:
> Nice work on bringing story export to Polarion - this is a valuable addition for the traceability story. A few things to address before merge:
> 
> **Must fix:**
> 
>     1. **`story.tag` mutation** - `story.tag.append('fmf-export')` modifies the original list in place, side effect that leaks outside the function
> 

Should be fixed by copy

>     2. **`_markdown_str_to_html` temp file** - writing a string to a temp file just to read it back is unnecessary, `markdown.markdown()` accepts strings directly
> 

created separate str conversion function inside utils

>     3. **Broad `except Exception` in `_set_polarion_feature_fields`** - silently swallows all errors at DEBUG level, hiding real bugs from the user
> 

Removed broad exception

> 
> **Should address:** 4. **Backward-incompatible search removal** - `extra-nitrate`, `extra-task`, and UUID plain query searches were dropped without documentation. Existing users relying on these to find Polarion cases will break. Either preserve as fallbacks or document as a breaking change. 5. **`LEGACY_POLARION_PROJECTS` removal** - undocumented behavior change, worth a note in the PR description 6. **No automated tests** - 823 lines of new code with zero test coverage. Unit tests for the core functions (`find_polarion_case_ids`, `get_polarion_feature`, `set_polarion_custom_fields`) would add confidence.
> 
> The dynamic CLI option collection via `_data_class` on `ExportPlugin` is a clean pattern. The `prepare_extra_summary` type broadening to `Core` is correct.

thrix:
Quick follow-up on my March 10 review: the main issues I flagged have been addressed in the latest commits — tag mutation fixed, temp file round-trip replaced with `markdown_to_html_str`, exception types narrowed, and legacy search methods preserved. Nice work.

The PR needs a rebase (merge conflicts) before another full review pass. Once that's done I'll take another look.

Minor items still open from my original review:
- Outer exception handler in `_set_polarion_feature_fields` still logs at DEBUG — consider WARNING so users see failures
- `LEGACY_POLARION_PROJECTS` removal worth a note in the PR description

Assisted-by: Claude Code

tmt/export/polarion.py
thrix: Outdated — the temp file round-trip has been replaced with `markdown_to_html_str` which converts strings directly.

tmt/export/polarion.py
thrix: Author acknowledged and confirmed fixing — legacy search methods are being preserved for backward compatibility.

tmt/export/polarion.py
thrix: Addressed — tag list is now copied with `list()` before mutation to avoid side effects on the original story object.

tmt/export/polarion.py
```diff
@@ -93,17 +107,41 @@ def find_polarion_case_ids(
...
+                except PolarionException as err:
+                    log.debug(f"Query '{query_format}' failed: {err}")
+                    continue
+
+    # Search by extra-polarion field
+    if not project_id:
+        extra_polarion = data.get('extra-polarion')
+        if extra_polarion:
+            query_result = PolarionWorkItem.query(f'id:{extra_polarion}', fields=wanted_fields)
+            case_id, project_id = get_polarion_ids(query_result, preferred_project)
```
LecrisUT, line 137: This was never introduced and per the other conversation `tmtid` is preferred. Please remove this as dead-code

tmt/cli/_root.py
```diff
@@ -1523,6 +1523,20 @@ def headfoot(text: str) -> None:
...
+    opts: list[tmt.options.ClickOptionDecoratorType] = []
+    registry = tmt.Story.get_export_plugin_registry()
+    for plugin_id in registry.iter_plugin_ids():
+        plugin_cls = registry.get_plugin(plugin_id)
+        if plugin_cls is not None:
+            opts.extend(plugin_cls.options())
+    return opts
+
+
+_story_export_plugin_options = create_options_decorator(_collect_story_export_plugin_options())
```
LecrisUT, line 1537: Please separate this to a smaller PR for better review

tmt/export/__init__.py
```diff
@@ -200,6 +202,26 @@ class ExportPlugin(abc.ABC):
...
 
+    _data_class: ClassVar[type] = type('_EmptyExportData', (), {})
+
+    @classmethod
+    def get_data_class(cls) -> type:
+        """Return the data class holding plugin-specific options"""
+        return cls._data_class
+
+    @classmethod
+    def options(cls, how: Optional[str] = None) -> list['tmt.options.ClickOptionDecoratorType']:
```
LecrisUT, line 213: Note these as originating from `tmt.step.Plugin` so we can figure out how to consolidate them

tmt/export/polarion.py
```diff
@@ -9,19 +9,22 @@
...
 PolarionException: Any = None
 PolarionTestCase: Any = None
 PolarionWorkItem: Any = None
+PolarionRequirement: Any = None
 
 
 POLARION_TRACKER_ID = 117  # ID of polarion in RH's bugzilla
 RE_POLARION_URL = r'.*/polarion/#/project/.*/workitem\?id=(.*)'
-LEGACY_POLARION_PROJECTS = {'RedHatEnterpriseLinux7'}
+
```
LecrisUT, line 27: Why the whitespace?

tmt/export/polarion.py
LecrisUT: Conflicts with the previous comment on the usage of `markdown_to_html`. I do not like the unnecessary repetition. Decide on if it is Path or str and only use that.

tmt/export/polarion.py
```diff
@@ -33,19 +36,23 @@ def import_polarion() -> None:
...
         from pylero.exceptions import PyleroLibException as PolarionException
     except ImportError as error:
         raise ConvertError("Install 'tmt+export-polarion' to use Polarion API") from error
 
     try:
         from pylero.work_item import TestCase as PolarionTestCase
         from pylero.work_item import _WorkItem as PolarionWorkItem
+
+        # PolarionRequirement is just an alias for _WorkItem with type='requirement'
+        PolarionRequirement = PolarionWorkItem
```
LecrisUT, line 49: Unused remove

tmt/export/polarion.py
```diff
@@ -33,19 +36,23 @@ def import_polarion() -> None:
...
     except PolarionException as exc:
         log.debug(traceback.format_exc())
         raise ConvertError("Failed to login with pylero") from exc
 
 
+
```
LecrisUT, line 55: Whitespace that breaks pre-commit

tmt/export/polarion.py
```diff
@@ -84,7 +90,15 @@ def find_polarion_case_ids(
     polarion_case_id: Optional[str] = None,
 ) -> tuple[Optional[str], Optional[str]]:
     """
-    Find IDs for Polarion case from data dictionary
+    Find IDs for Polarion case from data dictionary.
+
+    Searches in order:
+    1. Direct ``polarion_case_id`` parameter (explicit override)
+    2. TMT ID via ``tmtid`` custom field (preferred)
+    3. ``extra-polarion`` FMF field (new fallback)
```
LecrisUT, line 98: Do not use words like `new`. They will be outdated in minutes

LecrisUT (commented):
This is still too much unrelated changes for a human to review. And the AI does a super lousy job at catching the issues that the human is trying to raise. Again please split this PR into smaller components. We can get smaller PRs merged and gradually unblock this.

To give you some proper number, it would take me 2 weeks to just finish giving comments on everything in this PR, and this is not including commenting on the next changes. If you split the change of `markdown_to_html_str` that would be less than an hour worth of review and discussions.

The more you are able to reuse the previous logic the faster it would be to review

happz:
The PR is in a dire need of splitting, despite being reduced already, at least the following two points stand out:

* "markdown to HTML" change - a standalone change, adds support for returning the content, which should be a primitive for `markdown_to_html` which writes into a file.
* options for export plugins - a feature that could be very useful on its own, perfectly standalone addition. There must be some duplication between export plugins and step plugins in this area, it would be nice to reuse existing code, or librate it to serve both families.

Other comments:

* markdown composed by Python rather than using a template makes it really hard to follow its construction
* `log` should go away, use `logger`, propagate it to functions if needed.
* pyright should be enabled for both report plugin and export plugin, the amount of changes in modules not yet covered by both type linters is just asking for trouble.
* various minor issues - `TMT` => `tmt`, incorrect annotation here and there, etc.

thrix:
Btw, there is an effort to move away from Polarion to Jira (Atlassian Cloud) in RH:

https://docs.google.com/document/d/1XJ1bfYeNVue-zPIvMMja0giXbWWB76YHR4wHelctgBc/ (RH link only)

Just if folks would not be aware

thrix:
This did not get to sprint, initial review was done for 1.70 but needs substantial refactorin to move forward. See comments.

PR #4316, opened 2025-11-12
Add additional converter methods to FmfContext, StepData and Discover/shell

### LINKED ISSUE #4494: Add additional converter methods to FmfContext, StepData, and Discover/shell
Introduce additional converter methods to FmfContext, StepData and Discover/shell to support recipe loading feature.

therazix:
Adds several `to_minimal_spec` methods and one `from_serialized` method. It also makes `WhereableStepData` a serializable container. This is needed for the recipe loading feature (#4208).

Resolves #4494

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/discover/shell.py
+++ b/tmt/steps/discover/shell.py
@@ -224,9 +224,8 @@ class DiscoverShellData(tmt.steps.discover.DiscoverStepData):
 
     def to_minimal_spec(self) -> _RawDiscoverShellData:
         spec = {**super().to_minimal_spec()}
-        tests = [test.to_minimal_spec() for test in self.tests]
-        if tests:
-            spec['tests'] = tests
+        if self.tests:
+            spec['tests'] = [test.to_minimal_spec() for test in self.tests]
         return cast(_RawDiscoverShellData, spec)
 
 
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -419,17 +419,12 @@ class FmfContext(dict[str, list[str]]):
         return dict(self)
 
     @classmethod
-    def from_serialized(cls, spec: dict[str, list[str]]) -> 'FmfContext':
+    def from_serialized(cls, serialized: dict[str, list[str]]) -> 'FmfContext':
         """
         Convert from a serialized form.
         """
 
-        return FmfContext(
-            {
-                str(dimension): [str(value) for value in values]
-                for dimension, values in spec.items()
-            }
-        )
+        return FmfContext(serialized)
 
 
 #: A type of environment variable name.


tmt/utils/__init__.py
```diff
@@ -387,6 +379,22 @@ def to_spec(self) -> dict[str, Any]:
 
         return dict(self)
 
+    @classmethod
+    def from_serialized(cls, spec: dict[str, Union[str, list[str]]]) -> 'FmfContext':
```
happz: This seems unexpected: when would the serialized form not use list even for a single value? Serialized form should always be a `list[str]`.

tmt/steps/__init__.py
```diff
@@ -354,6 +354,9 @@ class _RawStepData(TypedDict, total=False):
     summary: Optional[str]
     order: Optional[int]
 
+    # Used by DiscoverShellData
```
happz: Shouldn't we have a subclass for `DiscoverShellData` then? I mean, this is leaking one key of one plugin, and the obvious question would be "why not other keys and other plugins?"

happz: Hmm, I see the comment, https://github.com/teemtee/tmt/pull/4208#discussion_r2518239538

I'm not sure I like this approach. Well, no, I'm sure I don't like it, let's try to find a better way tomorrow.

tmt/utils/__init__.py
therazix: Changed in a8155213895d51a87a0da35c32f01178980f4a2e.

tmt/steps/__init__.py
therazix: Changed in a8155213895d51a87a0da35c32f01178980f4a2e.

LecrisUT (approved):

tmt/steps/discover/shell.py
```diff
@@ -204,19 +212,22 @@ class DiscoverShellData(tmt.steps.discover.DiscoverStepData):
...
+            _RawDiscoverShellData,
+            {**super().to_spec(), 'tests': [test.to_spec() for test in self.tests]},
+        )
 
-        return data
+    def to_minimal_spec(self) -> _RawDiscoverShellData:
+        spec = {**super().to_minimal_spec()}
+        tests = [test.to_minimal_spec() for test in self.tests]
+        if tests:
+            spec['tests'] = tests
```
happz:
```
if self.tests:
    spec['tests'] = [...]
```

One `tests` variable gone.

tmt/utils/__init__.py
```diff
@@ -418,6 +418,19 @@ def to_spec(self) -> dict[str, Any]:
...
 
+    @classmethod
+    def from_serialized(cls, spec: dict[str, list[str]]) -> 'FmfContext':
+        """
+        Convert from a serialized form.
+        """
+
+        return FmfContext(
+            {
+                str(dimension): [str(value) for value in values]
```
happz: * `spec` is `dict[str, list[str]]` therefore `str()` calls should not be needed
* better to rename `spec` to `serialized`, to align with the rest of code.

tmt/steps/discover/shell.py
therazix: Changed in 1fc6d75111df78a7655ee231658044328aa34812.

tmt/utils/__init__.py
therazix: Changed in 1fc6d75111df78a7655ee231658044328aa34812.

Issue #4176, opened 2025-10-10
Included TMT examples don't pass tmt tests lint


vkadlcik:
https://tmt.readthedocs.io/en/stable/spec/tests.html#manual says that this is a minimal example of a manual test document:
```
# Test

## Step
Do this and that.

## Expect
Check this and that.
```
Similarly, https://github.com/teemtee/tmt/blob/main/examples/manual/mini.md offers
```
# Test

## Step
Description of what has to be done

## Expect
Description of the desired outcome
```

They don't pass tmt `tests lint` though:
```
warn T008 Heading "<h2>Step</h2>" from the section "Step" is used outside of Test sections.
```

Took me some fiddling to find out they miss a `# Cleanup` section at the end. Please fix the examples to be `lint`-clean. And maybe make the warning message clearer?

I'm using tmt-1.58.0-1.fc42.noarch.

AthreyVinay:
~Would re-wording the warning like below make it more clearer?~

~~```
warn_outside_test_section = (
    'Heading "{}" appears outside of a valid Test block. '
    'Each test should include "# Test", "## Step", "## Expect"  '
    '(and optionally "## Cleanup").'
)~~

**EDIT**
Ok, now I'm in a fix. So the `Cleanup` section is intended to be `optional` according to the test specification, but the linter currently requires it to terminate a Test block. So should be `required`?

PR #4322, opened 2025-11-14
Fix heading detection logic and enforce correct Test/Step/Expect structure

AthreyVinay:
Probable fix for: #4176

Updated https://tmt.readthedocs.io/en/stable/spec/tests.html#manual & also improves the `T008` linter warning.

This might be completely incorrect as noted here: https://github.com/teemtee/tmt/issues/4176#issuecomment-3524544662

### DIFF
--- a/tests/lint/test/test.sh
+++ b/tests/lint/test/test.sh
@@ -95,6 +95,10 @@ rlJournalStart
         rlAssertGrep "$fail unknown html heading "<h2>Unknown heading end</h2>" is used" $rlRun_LOG
         rlAssertGrep "$fail unknown html heading "<h3>Unknown heading begin</h3>" is used" $rlRun_LOG
 
+        # Warn if 2 or more # Setup or # Cleanup are used
+        rlAssertGrep "$fail 2 headings \"<h1>Setup</h1>\" are used" $rlRun_LOG
+        rlAssertGrep "$fail 3 headings \"<h1>Cleanup</h1>\" are used" $rlRun_LOG
+
         # Step is used outside of test sections.
         rlAssertGrep "$fail Heading \"<h2>Step</h2>\" from the section \"Step\" is used outside of Test sections." $rlRun_LOG
 
--- a/tmt/export/__init__.py
+++ b/tmt/export/__init__.py
@@ -345,6 +345,17 @@ class TestSection:
     expects: list[str] = simple_field(default_factory=list)
 
 
+@container
+class MarkdownFileSection:
+    """
+    A container for all sections in a markdown file
+    """
+
+    tests: list[TestSection] = simple_field(default_factory=list)
+    setup: list[str] = simple_field(default_factory=list)
+    cleanup: list[str] = simple_field(default_factory=list)
+
+
 def get_bz_instance() -> BugzillaInstance:
     """
     Import the bugzilla module and return BZ instance
@@ -465,7 +476,7 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
         for match in HEADING_PATTERN.finditer(md_to_html)
     ]
     warnings = []
-    test_sections: list[TestSection] = []
+    file_section = MarkdownFileSection()
     current_test: Optional[TestSection] = None
 
     for level, heading in headings:
@@ -475,11 +486,17 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
         if not section_type:
             warnings.append(f'unknown html heading "{heading}" is used')
 
+        # Collect Setup/Cleanup occurrences
+        if section_type == "Setup":
+            file_section.setup.append(heading)
+        elif section_type == "Cleanup":
+            file_section.cleanup.append(heading)
+
         # Start new test section on h1 heading
         if level == 1:
             current_test = TestSection(name=heading) if section_type == "Test" else None
             if current_test:
-                test_sections.append(current_test)
+                file_section.tests.append(current_test)
             continue
 
         # Inside an open test section
@@ -499,13 +516,20 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
                 f'used outside of Test sections.'
             )
 
+    # Warn if more than one Setup or Cleanup
+    warnings.extend(
+        f'{len(h)} headings "{h[0]}" are used'
+        for h in (file_section.setup, file_section.cleanup)
+        if len(h) > 1
+    )
+
     # At least one test section must exist
-    if not test_sections:
+    if not file_section.tests:
         warnings.append('"Test" section doesn\'t exist in the Markdown file')
         return warnings
 
     # # Step isn't in pair with # Expect
-    for test in test_sections:
+    for test in file_section.tests:
         steps_count = len(test.steps)
         expects_count = len(test.expects)
         if steps_count != expects_count:


examples/manual/mini.md
```diff
@@ -5,3 +5,6 @@ Description of what has to be done
 
 ## Expect
 Description of the desired outcome
+
+# Cleanup
+(Optional) Description of what needs to be performed after the test execution has been completed
```
LecrisUT: This is the mini example, and if the cleanup is optional, this should not be needed?

AthreyVinay: Exactly and hence the comment in the description. 
Pasting it here again: 
> This might be completely incorrect as noted here: https://github.com/teemtee/tmt/issues/4176#issuecomment-3524544662

LecrisUT: Yes, and the answer for that is seemingly no (it doesn't seem to be needed indeed). But of course the source of truth here should be https://github.com/teemtee/tmt/blob/main/spec/tests/manual.fmf

AthreyVinay: Cheers - I think this helps!

AthreyVinay: Does this look any better? [55ddb4d](https://github.com/teemtee/tmt/pull/4322/commits/55ddb4d4c593e0f93e541f0337e66541df09b7cf)

tmt/export/__init__.py
```diff
@@ -469,7 +469,9 @@ def count_html_headings(heading: str) -> None:
     count_html_headings(sections_headings['Cleanup'][0])
 
     warn_outside_test_section = (
-        'Heading "{}" from the section "{}" is used \noutside of Test sections.'
+        'Heading "{}" from the section "{}" is used \noutside of Test sections. '
+        'Each test should include "# Test", "## Step", "## Expect", '
+        '(and optionally "## Cleanup").'
     )
```
LecrisUT: Judging by the usage down below, this warning only appears when checking `Step` and `Expect` section and it is checking if it belongs to a `# Test`-like section. The current warning seems fine since `Setup` and `Cleanup` seem to be only `h1` level headers

tmt/export/__init__.py
```diff
@@ -491,10 +493,13 @@ def required_section_exists(
...
-        html_headings_from_file.pop()
-        # Add # Test heading to close the file
+    # If the last heading isn't Test or Cleanup, append a Test heading
+    # to serve as a closing boundary for proper section detection
+    if (
+        html_headings_from_file[-1] != '<h1>Test</h1>'
+        and html_headings_from_file[-1] != '<h1>Cleanup</h1>'
+    ):
+        # acts like a closing boundary
         html_headings_from_file.append(sections_headings['Test'][0])
```
LecrisUT: Trying to wrap my head on what the previous and current one is doing. At most I would think that that the "closing boundary" should be added unconditionally.

tmt/export/__init__.py
LecrisUT: Ok, I think the key here is this check where it either parses the last entry or not depending on if a "closing boundary" was added.

tmt/export/__init__.py
AthreyVinay: **Previously**:
The parser used  the presence of the last heading to infer test-block boundaries:
```rst
# Remove Optional heading #Cleanup if it's in the end of document
if html_headings_from_file[-1] == '<h1>Cleanup</h1>':
    html_headings_from_file.pop()
    # Add # Test heading to close the file
    html_headings_from_file.append(sections_headings['Test'][0])
```
This meant:
1. `Cleanup` at end of doc is treated as optional.
2. But if it's missing entirely, then the final Test block never ends →
Step / Expect headings after the first block are interpreted as “outside of test sections”.
Result: false T008 warnings.
So the manual test spec intended Cleanup to be optional, but the linter logic accidentally forced ti.

**Currently**:
Append a synthetic closing marker for the last Test block

tmt/export/__init__.py
AthreyVinay: +1 - makes sense
Done: [0d99b09](https://github.com/teemtee/tmt/pull/4322/commits/0d99b092fa6cf66321d592e3f13911b197ea526c)

tmt/export/__init__.py
LecrisUT: Yeah, but I don't see why this should have checked explicitly for `Cleanup` only. It should have done it in all cases afaict. Regardless the approach is still too flaky. Reworking it would be better than whack-a-mole until the next edge-case comes up don't you think?

AthreyVinay: All in for refactoring if reduces complexity and improves readability / maintenance. I'll draft something up.

AthreyVinay:
@LecrisUT: [refactor: a humble effort](https://github.com/teemtee/tmt/pull/4322/commits/01d06dc2ed55b3b01bbcb1d377c0301f5d6c785d) :)

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
 
-    warn_outside_test_section = (
-        'Heading "{}" from the section "{}" is used \noutside of Test sections.'
-    )
-    warn_headings_not_in_pairs = (
-        'The number of headings from the section "Step" - {}\ndoesn\'t equal to the '
-        'number of headings from the section \n"Expect" - {} in the test section "{}"'
+    # Identify test-section headings
+    test_heading_pattern = re.compile(r'^<h1>(Test .*|Test)</h1>$')
+    all_valid_headings = {heading for section in sections_headings.values() for heading in section}
```
LecrisUT: Let's move all of the `*_headings` in one part

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+    expect_headings = set(sections_headings['Expect'])
 
+    # Extract headings
     md_to_html = tmt.utils.markdown_to_html(md_path)
-    html_headings_from_file = [
-        i[0] for i in re.findall('(^<h[1-4]>(.+?)</h[1-4]>$)', md_to_html, re.MULTILINE)
+    headings = [
+        heading
+        for heading, _ in re.findall(r'(^<h[1-4]>(.+?)</h[1-4]>$)', md_to_html, re.MULTILINE)
     ]
```
LecrisUT, line 477: - Can you extract the heading-level and content into separate groups and make this something like a list of tuples?
- Also just for sake of correctness, can you match the heading levels on the opening and closing tag?
- Please make the regex patterns compiled once at the module level

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
-                warn_outside_test_section.format(html_headings_from_file[index], 'Step')
+
+    for heading in headings:
+        # Check if valid heading
+        is_valid_heading = heading in all_valid_headings
+        is_test_heading = test_heading_pattern.match(heading)
+
+        if not is_valid_heading and not is_test_heading:
+            warnings.append(f'unknown html heading "{heading}" is used')
+            continue
```
LecrisUT: I don't think we are restricting the user to have only the defined headings, or at least we should make that check change a distinct PR for discussion

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+
+        # Start new test section
+        if is_test_heading:
+            if current_test is not None:
+                test_sections.append(current_test)
+            current_test = {'name': heading, 'steps': [], 'expects': []}
+            continue
```
LecrisUT: - You should first check if you are encountering a `h1` heading and then either clear the `current_test` or create a new instance
- You can add `currrent_test` to the `test_sections` immediately. When you later update the dictionary it updates its contents as well
  ```python
  >>> my_list = []
  ... a_dict = {}
  ... my_list.append(a_dict)
  ... a_dict["something"] = "value"
  ... print(my_list)
  [{'something': 'value'}]
  ```

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+            elif section_type == "Expect":
+                current_test['expects'].append(heading)
+            continue
+
+        # Outside test section — detect orphan Step/Expect
+        section_type = heading_to_type.get(heading)
+        if section_type:
+            warnings.append(
+                f'Heading "{heading}" from "{section_type}" is used outside of Test sections'
             )
```
LecrisUT, line 517: - Just for logic readability, put this in an `else` statement to link it to the above
- Also let's check explicitly if the `section_type` is `Step`/`Expect`

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+
+        # Start new test section
+        if is_test_heading:
+            if current_test is not None:
+                test_sections.append(current_test)
+            current_test = {'name': heading, 'steps': [], 'expects': []}
+            continue
+
+        # Inside an open test section
+        if current_test is not None:
```
LecrisUT:
```suggestion
        if current_test:
```

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
-    )
-    warn_headings_not_in_pairs = (
-        'The number of headings from the section "Step" - {}\ndoesn\'t equal to the '
-        'number of headings from the section \n"Expect" - {} in the test section "{}"'
+    # Identify test-section headings
+    test_heading_pattern = re.compile(r'^<h1>(Test .*|Test)</h1>$')
+    all_valid_headings = {heading for section in sections_headings.values() for heading in section}
+    heading_to_type = dict.fromkeys(step_headings, "Step") | dict.fromkeys(
+        expect_headings, "Expect"
     )
```
LecrisUT: I think would be fine to do it for all of them, just to make it simple

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+            continue
+
+        # Inside an open test section
+        if current_test is not None:
+            section_type = heading_to_type.get(heading)
+            if section_type == "Step":
+                current_test['steps'].append(heading)
+            elif section_type == "Expect":
+                current_test['expects'].append(heading)
+            continue
```
LecrisUT: `if not section_type` -> `warn_unexpected_headings` (warn if there are other `h2` headings in the test section). Let's keep that old behavior and make a new PR to discuss if we change the spec about that.

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
 
     import tmt.base
 
-    warnings_list = []
     sections_headings = tmt.base.SECTIONS_HEADINGS
```
LecrisUT: I wonder if you should drop the heading tags from these and instead put that logic in the code here.

tmt/export/__init__.py
```diff
@@ -430,131 +430,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
-            html_headings_from_file[index] == sections_headings['Step'][0]
-            or html_headings_from_file[index] == sections_headings['Step'][1]
-        ):
-            warnings_list.append(
-                warn_outside_test_section.format(html_headings_from_file[index], 'Step')
+
+    for heading in headings:
+        # Check if valid heading
+        is_valid_heading = heading in all_valid_headings
+        is_test_heading = test_heading_pattern.match(heading)
```
LecrisUT: Could make a small helper that would
```python
for section, allowed_values in sections_headings.items():
    for value in allowed_values:
        if isinstance(value, re.Pattern):
            if value.match(heading):
                return section
        else:
            if heading == value:
                return section
return None
```

AthreyVinay: I feel it's unnecessary complexity. The simple dictionary lookup with `heading_to_type.get(full_heading)` is cleaner, & easier to understand.

LecrisUT: Was thinking in order to make `tmt.base.SECTIONS_HEADINGS`  as the only source of truth in there, and eliminating `test_heading_pattern`, `all_valid_headings`, etc. since you have already checked what heading type it was once.

AthreyVinay: AHA! Fair enough - makes sense.

AthreyVinay: Ok I think I have mostly addressed the comments here: [refactor: a humble effort, round II](https://github.com/teemtee/tmt/pull/4322/commits/905cbe910e976c4adc50807c5a1a14fe5707a211)

tmt/export/__init__.py
```diff
@@ -430,131 +445,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
-    for _ in list(sections_headings.values()):
-        values += _
 
+    def get_heading_section(heading: str) -> Optional[str]:
+        """Determine the section type for a heading."""
+        for section, allowed_values in sections_headings.items():
+            for value in allowed_values:
+                if isinstance(value, re.Pattern):
+                    if value.match(heading):
+                        return section
```
LecrisUT: You need to compile the pattern in `SECTIONS_HEADINGS`
https://github.com/teemtee/tmt/blob/5d95b6249c4793f422b6fd90221f12a78b8d8648/tmt/base.py#L141-L147
and adjust the only other instance where that is also used
https://github.com/teemtee/tmt/blob/5d95b6249c4793f422b6fd90221f12a78b8d8648/tmt/export/nitrate.py#L128-L137
Something like
```python
    for key in sections_headings:
        if isinstance(key, re.Pattern):
            key = key.pattern
        assert isinstance(key, str)
```
The code there is equally "a joy to navigate" but an issue for another day.

tmt/export/__init__.py
```diff
@@ -54,6 +55,9 @@
 BUGZILLA_XMLRPC_URL = "https://bugzilla.redhat.com/xmlrpc.cgi"
 RE_BUGZILLA_URL = r'bugzilla.redhat.com/show_bug.cgi\?id=(\d+)'
 
+# Used to extract <h1>-<h4> headings and their text from HTML
+HEADING_PATTERN = re.compile(r'^<h([1-4])>(.+?)</h\1>$', re.MULTILINE)
```
LecrisUT:
```suggestion
HEADING_PATTERN = re.compile(r'^<h(?P<level>[1-4])>(?P<title>.+?)</h(?P=level)>$', re.MULTILINE)
```
Equivalent but less confusing if you can access the groups by names.

tmt/export/__init__.py
```diff
@@ -430,131 +445,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+    test_sections: list[TestSection] = []
+    current_test: Optional[TestSection] = None
+
+    for level, heading in headings:
+        section_type = get_heading_section(heading)
+
+        # invalid section
+        if not section_type:
+            warnings.append(f'Unknown html heading "{heading}" is used')
+            continue
```
LecrisUT: This continue clashes with the logic under it. Afaict, you can just drop this one and the rest should be fine

tmt/export/__init__.py
```diff
@@ -430,131 +445,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+
+        # Start new test section on h1 heading
+        if level == 1:
+            if current_test:
+                test_sections.append(current_test)
+
+            current_test = TestSection(name=heading) if section_type == "Test" else None
+            continue
```
LecrisUT, line 500:
```suggestion
            current_test = TestSection(name=heading) if section_type == "Test" else None
            if current_test:
                test_sections.append(current_test)
            continue
```
Rearaging and you can drop the line in `# Add final test section`

tmt/export/__init__.py
```diff
@@ -430,131 +445,78 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+            if section_type == "Step":
+                current_test.steps.append(heading)
+            elif section_type == "Expect":
+                current_test.expects.append(heading)
+            else:
+                warnings.append(
+                    f'Heading "{heading}" isn\'t expected in the section "{current_test.name}"'
+                )
+        # Outside test section — detect orphan Step/Expect
+        elif section_type in ("Step", "Expect"):
```
LecrisUT:
```suggestion
        elif section_type in {"Step", "Expect"}:
```
Eventually ruff will catch these, but it is not enabled yet.

tmt/export/__init__.py
LecrisUT: Yep, looks great indeed.

tmt/export/__init__.py
LecrisUT: Nope, I was wrong at this one, the warning was present before.

tmt/export/__init__.py
AthreyVinay: Sure - for this PR - Ill leave it as-is and submit a subsequent PR that:
1. Compiles the pattern in `SECTIONS_HEADINGS`
2. Refactor the usages(2 places)

tmt/export/__init__.py
AthreyVinay: Addressed [6d13772](https://github.com/teemtee/tmt/pull/4322/commits/6d137722b05f3901e57fe3829c4a423f57f0eff0)

tmt/export/__init__.py
AthreyVinay: Addressed [6d13772](https://github.com/teemtee/tmt/pull/4322/commits/6d137722b05f3901e57fe3829c4a423f57f0eff0)

tmt/export/__init__.py
AthreyVinay: Addressed [6d13772](https://github.com/teemtee/tmt/pull/4322/commits/6d137722b05f3901e57fe3829c4a423f57f0eff0)

tmt/export/__init__.py
AthreyVinay: Addressed [6d13772](https://github.com/teemtee/tmt/pull/4322/commits/6d137722b05f3901e57fe3829c4a423f57f0eff0)

tmt/export/__init__.py
```diff
@@ -430,131 +445,70 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
-    required_headings = set(sections_headings['Step'] + sections_headings['Expect'])
-    values = []
-    for _ in list(sections_headings.values()):
-        values += _
-
+    def get_heading_section(heading: str) -> Optional[str]:
+        """Determine the section type for a heading."""
+        for section, allowed_values in tmt.base.SECTIONS_HEADINGS.items():
+            for value in allowed_values:
+                if isinstance(value, re.Pattern):
```
happz: IIUIC, according to https://github.com/teemtee/tmt/pull/4322/files#r2551066460 `SECTIONS_HEADINGS` remains `dict[str, list[str]]`, therefore checking whether it contains `Pattern` instances is puzzling, and belongs to the aforementioned PR, correct?

AthreyVinay: Spot on :+1: 
For this pr, I will ccompile at runtime until pre-compiled in the next one.

Done here: https://github.com/teemtee/tmt/pull/4322/commits/50acabdeb9c0c4846f112a121f73714208e34877

LecrisUT: Hackish solution, but I am ok with it if it is short-lived anyway.

LecrisUT (approved):

AthreyVinay:
Subsequent PR: #4362

LecrisUT:
@AthreyVinay please check the test failures

tmt/export/__init__.py
AthreyVinay: Okay - so the failing tests revealed a probable bug here.... we will need the `continue` here, reason being how else shall we reject unknown headings? A simple example:

if heading is:
`(level=1, heading='FooBar')   # unknown heading`

we end up with:
`test_sections = [None, ...]` as it enters `current_test = TestSection(name=heading) if section_type == "Test" else None` and then eventually appending `None` to `test_sections`. Thoughts?

LecrisUT: hooray for test coverage.
- 👍 for including the warning, it seems it was the previous behavior as well?
- Appending `None` is sus since everything was guarded by `if current_test` (there was even a duplicate `append`. Can you elaborate on how that happened?
- The `continue` should not be added, otherwise it would not end the test section when the next header is encountered.

AthreyVinay: Fixed here: https://github.com/teemtee/tmt/pull/4322/commits/a3794954ac831e5c1bbc2a79bd32a16b10e83893 thanks!

tests/lint/test/test.sh
```diff
@@ -93,14 +93,14 @@ rlJournalStart
...
+        # rlAssertGrep "$fail 2 headings \"<h1>Setup</h1>\" are used" $rlRun_LOG
+        # rlAssertGrep "$fail 3 headings \"<h1>Cleanup</h1>\" are used" $rlRun_LOG
 
         # Step is used outside of test sections.
         rlAssertGrep "$fail Heading \"<h2>Step</h2>\" from the section \"Step\" is used outside of Test sections." $rlRun_LOG
 
-        # Unexpected headings
-        rlAssertGrep "$fail Headings \".*\" aren't expected in the section \"<h1>Test</h1>\"" $rlRun_LOG
+        # Unexpected headings - empty headings are silently ignored
+        # rlAssertGrep "$fail Headings \".*\" aren't expected in the section \"<h1>Test</h1>\"" $rlRun_LOG
```
LecrisUT: Is the commenting of the tests intentional?

AthreyVinay: Actually just left it there - will be removed in the next one.

tmt/export/__init__.py
```diff
@@ -430,131 +445,74 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+    warnings = []
+    test_sections: list[TestSection] = []
+    current_test: Optional[TestSection] = None
+
+    for level, heading in headings:
+        section_type = get_heading_section(heading)
+
+        # Ignore unknown headings
+        if not section_type:
+            warnings.append(f'unknown html heading "{heading}" is used.')
```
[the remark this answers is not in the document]
AthreyVinay: Not true.

tests/lint/test/test.sh
LecrisUT: I mean, will you revisit the tests and make sure the checks make sense? The checks under there seem reasonable to be covered, but maybe the current implementation doesn't cover it?

AthreyVinay: Do we need to cover it given the current implementation? We ignore empty headings and the test doesnt make sense to check for it given it is never going to hit that. Are we on the same page?

LecrisUT: We do not ignore empty headings do we? Those simply evaluate to the `section_type is None` logic. Might have been the case for the previous implementation. The `rlAssertGrep` also doesn't assume empty headings would be silently ignored (and it might as well shouldn't do that for now).

At least for one round of review, it would be good to re-enable the tests and see specifically how/why it fails so that we can properly check the logic.

AthreyVinay: Yikes that took some time to fix the tests: [revert: tests are back !](https://github.com/teemtee/tmt/pull/4322/commits/830718317d9be04919c0a510ed417d27d7aa3b1f)

tests/lint/test/test.sh
```diff
@@ -102,9 +102,6 @@ rlJournalStart
         # Step is used outside of test sections.
         rlAssertGrep "$fail Heading \"<h2>Step</h2>\" from the section \"Step\" is used outside of Test sections." $rlRun_LOG
 
-        # Unexpected headings
-        rlAssertGrep "$fail Headings \".*\" aren't expected in the section \"<h1>Test</h1>\"" $rlRun_LOG
-
```
LecrisUT, line 107: Just for the review record, the `Headings` that were being caught here was `<h1>Cleansup</h1>`, which does not make sense since you would not have an `<h1>` heading inside a `<h1>`. The previous logic and this test were flawed in that sense.

tmt/export/__init__.py
```diff
@@ -430,131 +456,87 @@ def check_md_file_respects_spec(md_path: Path) -> list[str]:
...
+            file_section.setup.append(heading)
+        elif section_type == "Cleanup":
+            file_section.cleanup.append(heading)
+
+        # Start new test section on h1 heading
+        if level == 1:
+            current_test = TestSection(name=heading) if section_type == "Test" else None
+            if current_test:
+                file_section.tests.append(current_test)
+            continue
```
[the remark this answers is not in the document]
LecrisUT: That is very much intentional!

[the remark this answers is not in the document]
AthreyVinay: Not true:
Example test.md
```rst
# Test two
## Step
description for step 2-1

## Unknown heading end

## Result
description for result 2-1


# Cleanup
Optionally remove temporary directory created in the first step
2 line of cleanup
3 line of cleanup
```

output:
```rst
pass C001 summary key is set and is reasonably long
pass T001 correct keys are used
pass T002 test script is defined
pass T003 directory path is absolute
pass T004 test path '/home/athreyvinay/Documents/tmt/examples/manual' does exist
skip T005 legacy relevancy not detected
skip T006 legacy 'coverage' field not detected
pass T007 manual test path "/home/athreyvinay/Documents/tmt/examples/manual/mini.md" does exist
warn T008 unknown html heading "<h2>Unknown heading end</h2>" is used
warn T008 Heading "<h2>Unknown heading end</h2>" isn't expected in the section "<h1>Test two</h1>"
skip T009 library/file requirements not used
```

LecrisUT:
Hah, tests are failing because it was using git merged reference which has the newer tests, but the packit build used the PR source which did not have the changes related to the tests. That is a fun combo to be aware of

AthreyVinay:
> That is a fun combo to be aware of

Thanks for following up on this!

happz (approved):

Issue #4327, opened 2025-11-17
Add support for passing host devices (specifically /dev/kvm) into container


kraxel:
Use case: tests which use qemu, being able to use kvm speeds up things alot.
Podman just needs `--device=/dev/kvm` to make that work.

Complete podman command line (use `Ctrl-A x` to stop qemu and quit container):
```
podman run \
       --tty --interactive \
       --device=/dev/kvm \
       registry.fedoraproject.org/fedora:latest \
       sh -c 'dnf install -y qemu-system-x86-core; qemu-system-x86_64 -nographic -accel kvm'
```

PR #4333, opened 2025-11-18
Allow runner devices to be mounted in the `provision/container`

skycastlelily:
Introduce `expose-device` key for the `provision/container` allowing to forward the devices to be mounted on the container. This key is gated by the top-level `exposable-runner-devices`.

Fixes: #4327

---

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] include a release note

### DIFF
--- a/docs/ext/tmt_domain/base.py
+++ b/docs/ext/tmt_domain/base.py
@@ -13,10 +13,12 @@ This split is similar to the python sphinx domain workflow.
 import abc
 import typing
 from functools import cached_property
-from typing import Generic
+from typing import Generic, Optional
 
 from docutils.parsers.rst import directives
+from sphinx import addnodes
 from sphinx.directives import ObjDescT, ObjectDescription
+from sphinx.roles import XRefRole
 from sphinx.util import logging
 from sphinx.util.docutils import SphinxDirective
 
@@ -26,6 +28,8 @@ from tmt._compat.pathlib import Path
 from .autodoc import AutodocDirectiveBase
 
 if typing.TYPE_CHECKING:
+    from docutils.nodes import Element, Node, TextElement, system_message
+
     from .domain import TmtDomain
 
 
@@ -129,3 +133,45 @@ class TmtAutodocDirective(
         Get the :py:attr:`tmt_object`
         """
         raise NotImplementedError
+
+
+class TmtXRefRole(XRefRole):
+    """
+    Base tmt role for cross-referencing to a tmt object.
+
+    This saves some additional options indicating how to expand the role during the
+    call to :py:meth:`sphinx.domains.Domain.resolve_xref`.
+    """
+
+    #: Whether to use the tmt object's name/title key instead of the address
+    use_obj_name: bool = False
+
+    def __init__(
+        self,
+        fix_parens: bool = False,
+        lowercase: bool = False,
+        nodeclass: Optional[type["Element"]] = None,
+        innernodeclass: Optional[type["TextElement"]] = None,
+        warn_dangling: bool = False,
+        use_obj_name: Optional[bool] = None,
+    ) -> None:
+        if use_obj_name is not None:
+            self.use_obj_name = use_obj_name
+        super().__init__(
+            fix_parens=fix_parens,
+            lowercase=lowercase,
+            nodeclass=nodeclass,
+            innernodeclass=innernodeclass,
+            warn_dangling=warn_dangling,
+        )
+
+    def create_xref_node(self) -> tuple[list["Node"], list["system_message"]]:
+        # We do not have access to the actual tmt object until `Domain.resolve_xref` is
+        # being executed. We can only save the intent in the pending_xref node's data
+        # and handle it accordingly in the `resolve_xref`. See similar logic used with
+        # `refexplicit` (`sphinx.domains.std`)
+        nodes, messages = super().create_xref_node()
+        ref_node = nodes[0]
+        assert isinstance(ref_node, addnodes.pending_xref)
+        ref_node["tmtrefuseobjname"] = self.use_obj_name
+        return nodes, messages
--- a/docs/ext/tmt_domain/domain.py
+++ b/docs/ext/tmt_domain/domain.py
@@ -8,14 +8,13 @@ from functools import cached_property
 
 from docutils import nodes
 from sphinx.domains import Domain, ObjType
-from sphinx.roles import XRefRole
 from sphinx.util import logging
 from sphinx.util.nodes import make_refnode
 
 import tmt
 import tmt.log
 
-from .story import AutoStoryDirective, StoryDirective, StoryIndex
+from .story import AutoStoryDirective, StoryDirective, StoryIndex, StoryRole
 
 if typing.TYPE_CHECKING:
     from sphinx.addnodes import pending_xref
@@ -42,7 +41,7 @@ class TmtDomain(Domain):
     name = "tmt"
     label = "Internal tmt sphinx domain"
     roles = {
-        "story": XRefRole(),
+        "story": StoryRole(),
     }
     directives = {
         "autostory": AutoStoryDirective,
@@ -137,9 +136,12 @@ class TmtDomain(Domain):
         typ_objects = self.objects.get(obj_types[0], {})
         if not (obj := typ_objects.get(target)):
             return None
-        # Fix title from the actual object's title.
-        # Check `refexplicit` aka `has_explicit_title` if we actually need to do it.
-        if not node.get("refexplicit", False):
+        # Fix title from the actual object's title if needed.
+        # - `tmtrefuseobjname` aka `use_obj_name` is defined for each `TmtXRefRole` and
+        #   specifies if we want to use the tmt object's name/title (saved as `obj.name`)
+        # - `refexplicit` aka `has_explicit_title` is `True` if the role was defined like
+        #   :tmt:story:`Other title</spec/plans>` (we do not need to change the title then)
+        if node.get("tmtrefuseobjname", False) and not node.get("refexplicit", False):
             contnode.pop(0)
             contnode += nodes.Text(obj.name)
         return make_refnode(
--- a/docs/ext/tmt_domain/story.py
+++ b/docs/ext/tmt_domain/story.py
@@ -15,7 +15,7 @@ from tmt.base.core import Link, Story
 from tmt.utils.git import web_git_url
 
 from .autodoc import Content
-from .base import TmtAutodocDirective, TmtDirective
+from .base import TmtAutodocDirective, TmtDirective, TmtXRefRole
 
 if typing.TYPE_CHECKING:
     from .domain import TmtDomain
@@ -258,3 +258,12 @@ class StoryIndex(Index):
             entries = content.setdefault(index_key, [])
             entries.append(index_item)
         return sorted(content.items()), True
+
+
+class StoryRole(TmtXRefRole):
+    # The default literal node formats the string as a code.
+    # Using inline instead, same as sphinx's `ref` role (`sphinx.domains.std`)
+    innernodeclass = nodes.inline
+    # We want to use the story titles here. See `tmt_domain.note_object` call in
+    # `StoryDirective.run` for the index entry that we created.
+    use_obj_name = True


happz:
I would recommend making this feature more generic.

```python
if self.kvm_device_access:
    additional_args.extend(['--device=/dev/kvm'])
```

Let's say I want to also expose `/dev/ttyS3` and `/dev/drm_dp_aux5`, what would the tmt answer be in that case, `--ttyS3-device-access`, `--drm_dp_aux5-device-access`?

> Use case: tests which use qemu, being able to use kvm speeds up things alot.
Podman just needs --device=/dev/kvm to make that work.
>
> Complete podman command line (use Ctrl-A x to stop qemu and quit container):

We have several options how to proceed:

1. Add `--kvm-device-access`, solve that single use case, letting podman use `/dev/kvm`, and nothing else.
2. We could treat the request more generic, can we maybe enable `/dev/kvm` every time when it's available? If it's safe and beneficial, maybe it could be new default, and we can always offer `--no-kvm` option to disable the feature in case of trouble.
3. Podman's `--device` is quite generic, should we offer something similar? E.g. `device` key that would propagate its values to podman's `--device` - now that would probably unblock all devices, including `/dev/kvm`.

From these options, 1. seems like the least pleasing one, as it leaves every other device unavailable, and treats `/dev/kvm` in an exceptional way.

skycastlelily:
> I would recommend making this feature more generic.
> 
> ```python
> if self.kvm_device_access:
>     additional_args.extend(['--device=/dev/kvm'])
> ```
> 
> Let's say I want to also expose `/dev/ttyS3` and `/dev/drm_dp_aux5`, what would the tmt answer be in that case, `--ttyS3-device-access`, `--drm_dp_aux5-device-access`?
> 
> > Use case: tests which use qemu, being able to use kvm speeds up things alot.
> > Podman just needs --device=/dev/kvm to make that work.
> > Complete podman command line (use Ctrl-A x to stop qemu and quit container):
> 
> We have several options how to proceed:
> 
> 1. Add `--kvm-device-access`, solve that single use case, letting podman use `/dev/kvm`, and nothing else.
> 2. We could treat the request more generic, can we maybe enable `/dev/kvm` every time when it's available? If it's safe and beneficial, maybe it could be new default, and we can always offer `--no-kvm` option to disable the feature in case of trouble.
> 3. Podman's `--device` is quite generic, should we offer something similar? E.g. `device` key that would propagate its values to podman's `--device` - now that would probably unblock all devices, including `/dev/kvm`.
> 
> From these options, 1. seems like the least pleasing one, as it leaves every other device unavailable, and treats `/dev/kvm` in an exceptional way.

go with 2: https://github.com/teemtee/tmt/pull/4333/commits/6d553318063df9ce1c6d3078e195d67f53ce4450 :saluting_face:

tmt/steps/provision/podman.py
```diff
@@ -230,6 +240,10 @@ def start(self) -> None:
 
         additional_args.extend(self._setup_network())
 
+        # Add device access if requested
+        for device_path in self.device:
+            additional_args.extend(['--device=' + device_path])
```
happz: * Either use an f-string, `f'--device={device_path}'`, or use the 2-item form, `['--device', device_path]`. The latter would seem best to me, but both would work better than `+`.
* how about `device` instead of `device_path`? It's an item of `self.device`, not `self.device_path`, `device` would be closer.

tmt/steps/provision/podman.py
```diff
@@ -99,6 +99,15 @@ class PodmanGuestData(tmt.steps.provision.GuestData):
         normalize=tmt.utils.normalize_int,
     )
 
+    device: list[str] = field(
+        default_factory=list,
+        option='--device',
+        multiple=True,
+        metavar='DEVICE',
+        help='Device to expose to the container (e.g., /dev/kvm, /dev/ttyS3). '
+        'Can be specified multiple times.',
```
happz: A genuine multiline string can be used:

```python
help="""
     Device to expose to the container (e.g., /dev/kvm, /dev/ttyS3).
     Can be specified multiple times.
     """
```

tmt/steps/provision/podman.py
skycastlelily: Just back from long  vacation, and  addressed: https://github.com/teemtee/tmt/pull/4333/commits/76aaef211124d727e5f54fb966e037dd9e163b55 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/76aaef211124d727e5f54fb966e037dd9e163b55 :)

happz:
Hmhm, can this be dangerous? User adding `device: /dev/sda` into their plan, then run it in Testing Farm? Maybe we should have an allowlist, similar to one we have for `how`, `--allowed-how`, to prevent using every possible provision plugin in TF.

@psss @thrix @LecrisUT ^^

LecrisUT:
> Hmhm, can this be dangerous? User adding `device: /dev/sda` into their plan, then run it in Testing Farm? Maybe we should have an allowlist, similar to one we have for `how`, `--allowed-how`, to prevent using every possible provision plugin in TF.
> 
> @psss @thrix @LecrisUT ^^

At the same time, stuff like gpu loading might be useful in here as well. Hard to tell if a whitelist or blacklist would be best in these cases. Whitelist could be the more secure choice because if the parsing fails at least it fails safely.

Selinux might prevent the most outrageous attempts, but better to be double-protected.

happz:
> > Hmhm, can this be dangerous? User adding `device: /dev/sda` into their plan, then run it in Testing Farm? Maybe we should have an allowlist, similar to one we have for `how`, `--allowed-how`, to prevent using every possible provision plugin in TF.
> > @psss @thrix @LecrisUT ^^
> 
> At the same time, stuff like gpu loading might be useful in here as well. Hard to tell if a whitelist or blacklist would be best in these cases. Whitelist could be the more secure choice because if the parsing fails at least it fails safely.

I'd prefer allowlist rather than denylist, just to not miss any obvious security hole by not forbidding it.

I wouldn't mind adding it, but right now I don't have clear idea which part of tmt should own it. Making it a `provision/container` seems natural, but it's tricky when it comes to how TF would enforce it. As a CLI option it may have a side-effect of turning all `provision` phases into `how: container`... Well, something to think about.

LecrisUT:
Why not a primarily environment variable interface? Using the cli seems like it would get complicated quickly when deciding when to inject it or not, while a env variable would be loaded only when relevant right?

happz:
> Why not a primarily environment variable interface? Using the cli seems like it would get complicated quickly when deciding when to inject it or not, while a env variable would be loaded only when relevant right?

I simply like options, and I'm not done thinking. I haven't ruled out envvar at all.

tests/provision/container/basic/data/main.fmf
```diff
@@ -7,5 +7,63 @@
...
+    discover:
+        how: fmf
+    provision:
+        how: container
+        image: fedora
+        device:
+            - /dev/kvm
+            - /dev/ttyS3
+    execute:
+        how: tmt
```
LecrisUT: These are beyond the scope of `/tests/provision/container/basic`. Please split it out to something like `/tests/provision/container/device`. Also please use some inheritence

tests/provision/container/basic/test.sh
```diff
@@ -4,17 +4,50 @@
 rlJournalStart
     rlPhaseStartSetup
         rlRun "run=\$(mktemp -d)" 0 "Create a run directory"
+        rlRun "run_kvm=\$(mktemp -d)" 0 "Create a run directory for KVM test"
+        rlRun "run_multi=\$(mktemp -d)" 0 "Create a run directory for multiple devices test"
```
LecrisUT: We reuse runs whenever possible

tmt/steps/provision/podman.py
```diff
@@ -681,11 +697,34 @@ class ProvisionPodman(tmt.steps.provision.ProvisionPlugin[ProvisionPodmanData]):
...
+    .. code-block:: yaml
+
+        # Expose multiple devices to the container
+        provision:
+            how: container
+            image: fedora:latest
+            device:
+                - /dev/kvm
+                - /dev/ttyS3
+                - /dev/drm_dp_aux5
```
LecrisUT: One example is sufficient

tmt/steps/provision/podman.py
```diff
@@ -681,11 +697,34 @@ class ProvisionPodman(tmt.steps.provision.ProvisionPlugin[ProvisionPodmanData]):
...
+
     In order to always pull the fresh container image use ``pull: true``.
 
     In order to run the container with different user as the default ``root``,
     use ``user: USER``.
 
+    Use ``device`` to expose host devices to the container. This is useful
+    for cases like KVM acceleration (``device: /dev/kvm``) or accessing
+    serial devices. Multiple devices can be specified as a list.
```
LecrisUT: Since we cannot use `.. versionadded` can you make a release note?

happz:
> Looks good, but @happz's point about being able to whitelist is important. We might even want to gate this by `feeling-safe` locally as well (at least if the user did not provide the hardware list as well).

`--feeling-safe` could make the allowlist allow everything, but we will need extra control for the allowlist as `--feeling-safe` would enable other feature we do not want to enabled in a TF run.

LecrisUT:
> > Looks good, but @happz's point about being able to whitelist is important. We might even want to gate this by `feeling-safe` locally as well (at least if the user did not provide the hardware list as well).
> 
> `--feeling-safe` could make the allowlist allow everything, but we will need extra control for the allowlist as `--feeling-safe` would enable other feature we do not want to enabled in a TF run.

Indeed, that's why I was proposing to require it only if the hardware whitelist was not also provided

thrix (changes requested):
## Code Review: Add KVM device access support to Podman

Reviewed against commit `76aaef2`.

### Summary

This PR adds a generic `device` option to the Podman provision plugin, allowing users to expose host devices to containers. The implementation is straightforward — a `list[str]` field on `PodmanGuestData` that maps to `podman run --device` flags. However, there are several issues that need attention.

### Issues

**1. Test assertions will fail — format mismatch (bug)**

The code passes devices as two separate arguments:
```python
additional_args.extend(['--device', device])
```

This produces the command `podman run --device /dev/kvm ...` (space-separated). But the test asserts:
```bash
rlAssertGrep ".*--device=/dev/kvm.*" $rlRun_LOG
```

This expects `--device=/dev/kvm` with an `=` sign. `Command.__str__()` calls `to_element()` which joins args with spaces via `shlex.quote`. The log will contain `--device /dev/kvm`, not `--device=/dev/kvm`. This is confirmed by how the existing environment test works — `tests/provision/container/environment/test.sh` greps for `podman run.*--env-file $ENV_FILE` (space-separated), matching how `_setup_environment` uses `['--env-file', str(env_file)]`.

Fix: either change the code to `additional_args.append(f'--device={device}')`, or change the test patterns to match the space-separated form.

**2. Security: no input validation on device paths (raised in discussion)**

User-supplied device paths are passed directly to `podman --device` without any validation or sanitization. As @happz and @LecrisUT discussed, this is a security concern in shared environments like Testing Farm — a user could specify `device: /dev/sda` in their plan. The discussion about allowlists is still ongoing but this should be resolved before merging.

**3. Test organization (per @LecrisUT's review)**

Agree with @LecrisUT's unresolved comments:
- Device tests should be in a separate directory (`/tests/provision/container/device`), not added to `/tests/provision/container/basic`
- The test creates separate run directories (`run_kvm`, `run_multi`) instead of reusing runs as is the convention

**4. Weak test for multiple devices**

`/test-multiple-devices` passes if *any one* of the devices is available (`success > 0`). The test comment acknowledges `/dev/ttyS3` may not exist, but then why test with it at all? Either test with devices that are reliably available, or make the test only verify the `--device` flags are passed to podman (which is what `test.sh` already does via grep).

**5. Missing PR checklist items**

Several items are unchecked: documentation, spec update, JSON schema, version mention, release note. @LecrisUT's comment about adding a release note since `.. versionadded` can't be used is valid.

### Minor

- One docstring example is sufficient (per @LecrisUT) — the two separate code blocks showing single and multiple devices could be consolidated into one.

### Verdict

The test assertion mismatch (issue #1) is a bug that needs fixing. The security discussion (issue #2) needs resolution before merge. Test organization (issue #3) should be addressed per @LecrisUT's feedback.

Generated-by: Claude Code

tests/provision/container/basic/data/main.fmf
```diff
@@ -7,5 +7,63 @@
...
+
+/test-multiple-devices:
+    test: |
+        # Check if multiple devices are available
+        success=0
+
+        if [ -c /dev/kvm ]; then
+            echo "SUCCESS: /dev/kvm device is available"
+            success=$((success + 1))
+        else
```
thrix: This test passes if *any one* device is available (`success > 0`), but `/dev/ttyS3` is never actually tested — it's just listed. Either test with reliably available devices, or simplify this to only verify the `--device` flags are passed to podman (which `test.sh` already does via grep).

Generated-by: Claude Code

tests/provision/container/basic/test.sh
```diff
@@ -4,17 +4,50 @@
...
+    rlPhaseStartTest "Basic container test"
+        rlRun -s "tmt run --id $run -vvv plan --name /plan"
         rlAssertGrep "NAME.*Fedora Linux" $rlRun_LOG
     rlPhaseEnd
 
+    rlPhaseStartTest "Single device access test"
+        # Test the plan with single device access enabled
+        rlRun -s "tmt run --id $run_kvm -dddvvv plan --name /plan-with-kvm test --name /test-kvm-device"
+        # Check if device access was configured (look for --device=/dev/kvm in logs)
+        rlAssertGrep ".*--device=/dev/kvm.*" $rlRun_LOG
```
thrix: This grep pattern expects `--device=/dev/kvm` (with `=`), but the code uses `additional_args.extend(['--device', device])` which produces two separate command elements. `Command.__str__()` joins them with spaces via `shlex.quote`, so tmt's debug log will contain `--device /dev/kvm` (space-separated), not `--device=/dev/kvm`.

This is consistent with how the existing environment test works — `tests/provision/container/environment/test.sh` greps for `podman run.*--env-file $ENV_FILE` (space), matching the `['--env-file', str(env_file)]` pattern in `_setup_environment`.

The same issue applies to the assertions on lines 35-36 for the multiple devices test.

Fix: either change the code to use `f'--device={device}'` or update these grep patterns to match the space-separated form.

Generated-by: Claude Code

tmt/steps/provision/podman.py
```diff
@@ -311,6 +323,10 @@ def start(self) -> None:
         additional_args.extend(self._setup_network())
         additional_args.extend(self._setup_environment())
 
+        # Add device access if requested
+        for device in self.device:
+            additional_args.extend(['--device', device])
```
thrix, line 330: No input validation on device paths. As discussed by @happz and @LecrisUT, user-supplied values go straight to `podman --device` without any sanitization or allowlist check — e.g. a user could specify `device: /dev/sda` in a plan run on Testing Farm.

Generated-by: Claude Code

tests/provision/container/basic/data/main.fmf
```diff
@@ -7,5 +7,63 @@
...
+            echo "FAIL: /dev/kvm device not found"
+        fi
+
+        # /dev/ttyS3 may not exist on all systems, so just check if it was attempted to be mounted
+        # The important part is that the --device flag was passed to podman
+        echo "Checking for serial device access..."
+        ls -la /dev/tty* | head -5
+
+        if [ $success -gt 0 ]; then
```
thrix: This test passes if *any one* device is available (`success > 0`), but `/dev/ttyS3` is never actually checked — only listed. The test effectively only validates `/dev/kvm`. Either use reliably available devices for a real multi-device test, or simplify this to only verify the `--device` flags are passed to podman (which `test.sh` already does via grep).

Generated-by: Claude Code

happz:
Proposed a safeguard in https://github.com/teemtee/tmt/pull/4705. No tests, as nothing uses it yet. The idea is the plugin would make sure the values of the `device` key match at least one of the patterns.

tmt/steps/provision/podman.py
```diff
@@ -114,6 +114,17 @@ class PodmanGuestData(tmt.guest.GuestData):
         normalize=tmt.utils.normalize_int,
     )
 
+    device: list[str] = field(
```
happz: One late idea, did we discuss the name of the key? `expose_device`/`--expose-device` might be more descriptive than mere `device`.

tmt/steps/provision/podman.py
```diff
@@ -114,6 +114,17 @@ class PodmanGuestData(tmt.guest.GuestData):
...
+        default_factory=list,
+        option='--device',
+        multiple=True,
+        metavar='DEVICE',
+        help="""
+        Device to expose to the container (e.g., /dev/kvm, /dev/ttyS3).
+        Can be specified multiple times.
+        """,
```
happz:
```suggestion
        help="""
             Device to expose to the container (e.g., ``/dev/kvm``, ``/dev/ttyS3``).
             Can be specified multiple times.
             """, 
```

skycastlelily:
>  On the technical note, does this work on rootless podman?

 Sorry,I was on pto,in most scenario,no, unless the @user is under kvm group,   because /dev/kvm is a hardware device node typically owned by root and the kvm group :)

tests/provision/container/basic/data/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tests/provision/container/basic/test.sh
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tests/provision/container/basic/test.sh
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tests/provision/container/basic/data/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 ^^

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/b2badcceb5eece1581bf79eef8a9fc1c9a19baf7 ^^

tests/provision/container/device/test.sh
```diff
@@ -0,0 +1,30 @@
...
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create a run directory"
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Single device access test"
+        # Test the plan with single device access enabled
+        rlRun -s "tmt run -vvvddd --id $run  plan --name /plan-with-random test --name /test-random-device"
```
LecrisUT: Include a `--scratch` flag for these commands.

tests/provision/container/device/data/main.fmf
```diff
@@ -0,0 +1,51 @@
...
+# 3. Tests kept in a separate tree to avoid schema warnings
+/tests:
+    /test-random-device:
+        framework: shell
+        test: |
+            if [ -c /dev/random ]; then
+                echo "SUCCESS"
+            else
+                exit 1
+            fi
```
LecrisUT: Please incorporate these tests in the plan themselves

tests/provision/container/device/data/main.fmf
```diff
@@ -0,0 +1,51 @@
+# 1. The Base Plan (contains common settings)
+/plan-base:
```
LecrisUT: Naming structure should be `/plan/with-random`, etc. Otherwise current name is `/plan-base/plan-with-random` which is redundant.

tests/provision/container/device/test.sh
```diff
@@ -0,0 +1,30 @@
...
+        # Check if device access was configured (look for --device /dev/random in logs)
+        rlAssertGrep "--device /dev/random" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartTest "Multiple devices access test"
+        # Test the plan with multiple devices access enabled
+	rlRun -s "tmt run -vvvddd --id $run  plan --name /plan-with-multiple-devices test --name /test-multiple-devices"
+        # Check if multiple device access was configured
+	rlAssertGrep "--device /dev/random" $rlRun_LOG
+	rlAssertGrep "--device /dev/urandom" $rlRun_LOG
```
LecrisUT: Indentation

tests/provision/container/device/test.sh
```diff
@@ -0,0 +1,30 @@
...
+        rlRun "run=\$(mktemp -d)" 0 "Create a run directory"
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Single device access test"
+        # Test the plan with single device access enabled
+        rlRun -s "tmt run -vvvddd --id $run  plan --name /plan-with-random test --name /test-random-device"
+        # Check if device access was configured (look for --device /dev/random in logs)
+        rlAssertGrep "--device /dev/random" $rlRun_LOG
+    rlPhaseEnd
```
LecrisUT, line 16: Can you include the failing case when `TMT_EXPOSABLE_RUNNER_DEVICES` is not specified?

tmt/steps/provision/podman.py
```diff
@@ -37,6 +37,82 @@
...
+    # Graphics auxiliary devices (not primary graphics devices)
+    '/dev/drm_dp_aux*',
+    # Random number devices
+    '/dev/random',
+    '/dev/urandom',
+    '/dev/hwrng',
+    # Specific input devices
+    '/dev/input/event*',
+    '/dev/input/js*',
+]
```
LecrisUT: Let's drop this, maintaining this list ourselves seems out of our scope. Instead document that `TMT_EXPOSABLE_RUNNER_DEVICES` must be specified alongside.

tmt/steps/provision/podman.py
```diff
@@ -37,6 +37,82 @@
...
+    Only devices matching allowed patterns are permitted to prevent exposure
+    of dangerous devices like storage devices in shared environments like
+    Testing Farm.
+    """
+    import fnmatch
+
+    # Check if device matches any allowed pattern
+    for pattern in ALLOWED_DEVICE_PATTERNS:
+        if fnmatch.fnmatch(device, pattern):
+            return device
```
LecrisUT: The `return device` here is meaningless if the other loop is just a `raise`

tmt/steps/provision/podman.py
```diff
@@ -37,6 +37,82 @@
...
+
+    validated_devices = []
+    for device in devices:
+        if not isinstance(device, str):
+            raise tmt.utils.SpecificationError(
+                f"Each device must be a string, got {type(device).__name__}: {device}"
+            )
+        validated_devices.append(_validate_device_against_allowlist(device.strip()))
+
+    return validated_devices
```
LecrisUT: Not needed this is exactly the same as `tmt.utils.normalize_string_list`

tmt/steps/provision/podman.py
happz: Yes, this list is not supposed to be baked into the plugin, it's supposed to be set by the entity who invokes tmt - TF, a user on their workstation, this kind of entities.

tmt/steps/provision/podman.py
happz: `validated_devices` is needles, either they are all fine, or there was one that did not pass the match test, and all are refused and exception gets raised. You end up with either `validated_devices == devices`, or leave because of an exception.

tmt/steps/provision/podman.py
```diff
@@ -37,6 +37,82 @@
...
+    Validate device path against allowlist for security.
+
+    Only devices matching allowed patterns are permitted to prevent exposure
+    of dangerous devices like storage devices in shared environments like
+    Testing Farm.
+    """
+    import fnmatch
+
+    # Check if device matches any allowed pattern
+    for pattern in ALLOWED_DEVICE_PATTERNS:
```
happz: Right, the check must happen against the `exposable_runner_devices` attribute - the list of patterns is defined by whoever runs tmt, not by us. It's empty by default, we do not force `/dev/kvm` being exposable by default.

happz: BTW `exposable_runner_devices` is a list of compiled patterns, `all()` should be enough to handle it, `fnmatch` should not be necessary.

docs/releases/1.70.0/4333.fmf
```diff
@@ -0,0 +1,5 @@
+description: |
+    The :ref:`/plugins/provision/container` plugin now supports
+    ``expose-device`` parameter to expose host devices to the container.
```
happz:
```suggestion
    ``expose-device`` key to expose runner devices to the guest.
```

tmt/steps/provision/podman.py
```diff
@@ -38,6 +38,62 @@
...
+    The TMT_EXPOSABLE_RUNNER_DEVICES environment variable should contain a
+    colon-separated list of allowed device paths.
+
+    Example:
+        export TMT_EXPOSABLE_RUNNER_DEVICES="/dev/random:/dev/urandom:/dev/kvm"
+    """
+    import os
+
+    # Get allowlist from environment variable
+    allowlist_env = os.environ.get('TMT_EXPOSABLE_RUNNER_DEVICES')
```
happz: Please, read the patch you build on. It clearly adds `Common.exposable_runner_device_patterns`, there is no need for any special extractions or conversions.

> "must be explicitly enabled by setting TMT_EXPOSABLE_RUNNER_DEVICES to "
> "a colon-separated list of allowed device paths. "

No, it really doesn't have to be.

1. user can use `--exposable-runner-devices` otpion *or* the environment variable, it's not mandatory to use the envvar
2. there is no colon separation, just regular expressions. User can specify them multiple times via command line option, adding multiple patterns, or use the envvar with just one pattern using alternatives, `^/dev/(kvm|sda)$`
3. no paths, patterns, `^/dev/.*`. No exact matching.

Plus, the exception message is also quite long - consider adding a hint that would be printed out, followed by a more focused exception.

skycastlelily: yeah, polishing :sweat_smile:  please wait until I comment, thanks for your quick response anyway ^^

happz: Ok, I moved it to `implement` state then. Flip it back once you're done.

skycastlelily: done ^^

tests/provision/container/device/test.sh
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tests/provision/container/device/data/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tests/provision/container/device/test.sh
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tests/provision/container/device/test.sh
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 :)

docs/releases/1.70.0/4333.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/0d65e4e56f4934bda0fdfd420c7f19b536d34715 ^^

tests/provision/container/device/data/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/83bb871de41316669792ba3cf42dde5a346df982 :)

tmt/steps/provision/podman.py
```diff
@@ -684,11 +702,41 @@ class ProvisionPodman(tmt.steps.provision.ProvisionPlugin[ProvisionPodmanData]):
...
+
+    For security, device access must be explicitly enabled by the environment
+    running tmt (Testing Farm, user workstation, etc.) rather than maintaining
+    a hardcoded allowlist in tmt itself. This puts the security responsibility
+    on the infrastructure provider.
+
+    Permitted devices are defined using the ``--exposable-runner-devices``
+    CLI option (which can be passed multiple times) or the
+    ``TMT_EXPOSABLE_RUNNER_DEVICES`` environment variable (formatted as a
+    space-separated list). Both exact paths and regular expressions are supported.
```
happz: >    ``TMT_EXPOSABLE_RUNNER_DEVICES`` environment variable (formatted as a
>    space-separated list). Both exact paths and regular expressions are supported.

There is not space-separated list. Please, read the documentation, this is completely fabulated behavior.

I would cut out the whole detail paragraphs about exposable runner devices options, replacing it with just a mention for user to read about it elsewhere. This is duplication of the text, we would need to maintain two places documenting the option.

```python
Use ``expose-device`` to expose host devices to the container. This is useful
for cases like KVM acceleration (``expose-device: /dev/kvm``) or accessing
serial devices. Multiple devices can be specified as a list.

For security reasons, exposable devices need to be explicitly allowed by tmt
runner, either via ``--exposable-runner-devices`` CLI option
``TMT_EXPOSABLE_RUNNER_DEVICES`` environment variable.
```

And that's it. How the option behaves, what's the format - please, go and read the documentation and help for that option.

tmt/steps/provision/podman.py
```diff
@@ -727,6 +794,10 @@ def go(self, *, logger: Optional[tmt.log.Logger] = None) -> None:
         if data.hardware and data.hardware.constraint:
             self.warn("The 'container' provision plugin does not support hardware requirements.")
 
+        # Validate device access configuration before creating guest
+        for device in data.expose_device:
+            self._validate_device_against_allowlist(device.strip())
```
happz: No `strip()` please, you have a list of devices and stick to it. If `strip()` was needed or not, that's not up to you at this moment, such a cleanup, normalization step should have been already performed.

tests/provision/container/device/main.fmf
```diff
@@ -0,0 +1,7 @@
+tag+:
+  - provision-container-device
```
psss: What is the purpose/intention of this new `tag`? I don't see it used anywhere.

docs/releases/1.70.0/4333.fmf
```diff
@@ -0,0 +1,5 @@
+description: |
+    The :ref:`/plugins/provision/container` plugin now supports
```
psss:
```suggestion
    The :ref:`/plugins/provision/container` plugin now supports the
```

docs/releases/pending/4333.fmf
```diff
@@ -0,0 +1,5 @@
+description: |
```
psss, line 1: As a good practice, please, move the release note under the `pending` directory. Pull requests sometime overlap releases and it's not necessary to handle the right version here, it will be handled at the release time.

tests/provision/container/device/test.sh
```diff
@@ -0,0 +1,39 @@
...
+        # Check if multiple device access was configured
+        rlAssertGrep "--device /dev/random" $rlRun_LOG
+        rlAssertGrep "--device /dev/urandom" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartTest "Device access failure test when allowlist not configured"
+        # Test that device access fails when TMT_EXPOSABLE_RUNNER_DEVICES is not set
+        # Temporarily unset the environment variable to test security restriction
+        rlRun "unset TMT_EXPOSABLE_RUNNER_DEVICES"
+        rlRun "tmt run -vvvddd --scratch --id $run  plan --name /plan/with-random test --name /test-random-device" 2 "Device access should fail without allowlist"
```
LecrisUT: Please grep the failure

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/520aa3a3eb6f443a1631cc36d99b49727878be5b ^^

tmt/steps/provision/podman.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/520aa3a3eb6f443a1631cc36d99b49727878be5b ^^

tests/provision/container/device/test.sh
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4333/commits/40a6916fa30dbfa33f7940f8bfee652f0550300d :)

tests/provision/container/device/test.sh
```diff
@@ -0,0 +1,41 @@
...
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create a run directory"
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Single device access test"
+        # Test the plan with single device access enabled
+        rlRun "export TMT_EXPOSABLE_RUNNER_DEVICES='/dev/random'" 0 "Set device allowlist"
+        rlRun -s "tmt run -vvvddd --scratch --id $run  plan --name /plan/with-random test --name /test-random-device"
```
psss: The `--name /test-random-device` part has no effect as it is the name of the test but this option is under the `plan` filter command so it tries to find plans with such a name.

tests/provision/container/device/test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
psss, line 1: One more: The script does not have the execution permission set:

```
00:00:00 errr /tests/provision/container/device
    Note: beakerlib: TestResults FileError
    logs (with content):
        output.txt (/var/tmp/tmt/run-009/plans/provision/container/the-rest/execute/data/guest/default-0/tests/provision/container/device-1/output.txt)
            content: ./tmt-test-wrapper-inner.sh-tests-provision-container-device-1-default-0-default-0: line 1: ./test.sh: Permission denied
        failures.yaml (/var/tmp/tmt/run-009/plans/provision/container/the-rest/execute/data/guest/default-0/tests/provision/container/device-1/failures.yaml)
    00:00:00 fail internal/permission (after-test check)
        Note: Test failed due to denied permissions
```

tests/provision/container/device/main.fmf
skycastlelily: Addressed : https://github.com/teemtee/tmt/pull/4333/commits/87542eb4f2a63ea5929079287eb3d83b28fb2cb0 :)

docs/releases/1.70.0/4333.fmf
skycastlelily: Addressed : https://github.com/teemtee/tmt/commit/87542eb4f2a63ea5929079287eb3d83b28fb2cb0 :)

docs/releases/pending/4333.fmf
skycastlelily: Addressed : https://github.com/teemtee/tmt/commit/87542eb4f2a63ea5929079287eb3d83b28fb2cb0 :)

tests/provision/container/device/test.sh
skycastlelily: Addressed : https://github.com/teemtee/tmt/commit/87542eb4f2a63ea5929079287eb3d83b28fb2cb0 :)

tests/provision/container/device/test.sh
skycastlelily: Addressed : https://github.com/teemtee/tmt/commit/87542eb4f2a63ea5929079287eb3d83b28fb2cb0 :)

psss:
Tests are failing:

```
:: [ 15:08:36 ] :: [  BEGIN   ] :: Running 'diff /tmp/tmp.nuhbIWn9Cm/actual_normalized_recipe.yaml /tmp/tmp.nuhbIWn9Cm/expected_normalized_recipe.yaml'
286d285
<           expose-device: []
:: [ 15:08:36 ] :: [   FAIL   ] :: Command 'diff /tmp/tmp.nuhbIWn9Cm/actual_normalized_recipe.yaml /tmp/tmp.nuhbIWn9Cm/expected_normalized_recipe.yaml' (Expected 0, got 1)
```

Seems that `/tests/recipe/generate` needs to be adjusted to include the newly introduced key.

skycastlelily:
> Tests are failing:
> 
> ```
> :: [ 15:08:36 ] :: [  BEGIN   ] :: Running 'diff /tmp/tmp.nuhbIWn9Cm/actual_normalized_recipe.yaml /tmp/tmp.nuhbIWn9Cm/expected_normalized_recipe.yaml'
> 286d285
> <           expose-device: []
> :: [ 15:08:36 ] :: [   FAIL   ] :: Command 'diff /tmp/tmp.nuhbIWn9Cm/actual_normalized_recipe.yaml /tmp/tmp.nuhbIWn9Cm/expected_normalized_recipe.yaml' (Expected 0, got 1)
> ```
> 
> Seems that `/tests/recipe/generate` needs to be adjusted to include the newly introduced key.

Should be addressed: https://github.com/teemtee/tmt/pull/4333/commits/eff812b937d40a0e8b4deeac75d93ef2f071596f :)

psss:
@thrix, seems that your comments have been addressed. Could you please re-review?

tmt/steps/provision/podman.py
```diff
@@ -712,6 +748,25 @@ def default(self, option: str, default: Any = None) -> Any:
...
+        """
+        Validate a requested device path against the configured security allowlist.
+
+        Raises a SpecificationError if the device does not match any of the
+        patterns allowed by the TMT_EXPOSABLE_RUNNER_DEVICES configuration.
+        """
+        allowed_patterns = self.exposable_runner_device_patterns
+
+        # Now you can iterate over allowed_patterns and use pattern.match(device)
+        if not any(pattern.match(device) for pattern in allowed_patterns):
```
LecrisUT: `allowed_patterns` is unnecessary anymore

tmt/steps/provision/podman.py
```diff
@@ -712,6 +748,25 @@ def default(self, option: str, default: Any = None) -> Any:
...
+            raise tmt.utils.SpecificationError(
+                f"Device '{device}' cannot be exposed. The device is not in the security "
+                "allowlist. Please configure the '--exposable-runner-devices' option or the "
+                "'TMT_EXPOSABLE_RUNNER_DEVICES' environment variable."
+            )
+
+        return device
```
LecrisUT: The return is unused anywhere else, can drop it

tmt/steps/provision/podman.py
```diff
@@ -712,6 +748,25 @@ def default(self, option: str, default: Any = None) -> Any:
 
         return super().default(option, default=default)
 
+    def _validate_device_against_allowlist(self, device: str) -> str:
+        """
+        Validate a requested device path against the configured security allowlist.
+
+        Raises a SpecificationError if the device does not match any of the
+        patterns allowed by the TMT_EXPOSABLE_RUNNER_DEVICES configuration.
```
LecrisUT:
```suggestion
        patterns allowed by the ``TMT_EXPOSABLE_RUNNER_DEVICES`` configuration.
```

happz: Please, drop the environment variable name entitely, and use “exposable devices”. Mentioning just the envvar ignores the command line option, and suggests this is the only way. This piece of code must not be concerned how the valuebis delivered, it’d simply consuming the list of patterns husrfing exposable devices, and that’s all.

happz:
```suggestion
        exposable device patterns.
```

tmt/steps/provision/podman.py
skycastlelily: Addressed:
https://github.com/teemtee/tmt/pull/4333/commits/48311873b53f5de0c00eca0d9e4ca250d43782ee:)

tmt/steps/provision/podman.py
skycastlelily: Addressed:
https://github.com/teemtee/tmt/commit/48311873b53f5de0c00eca0d9e4ca250d43782ee:)

tmt/steps/provision/podman.py
skycastlelily: Addressed:
https://github.com/teemtee/tmt/commit/48311873b53f5de0c00eca0d9e4ca250d43782ee^^

skycastlelily: Addressed:
https://github.com/teemtee/tmt/commit/48311873b53f5de0c00eca0d9e4ca250d43782ee^^

tmt/steps/provision/podman.py
```diff
@@ -723,7 +723,7 @@ class ProvisionPodman(tmt.steps.provision.ProvisionPlugin[ProvisionPodmanData]):
 
     For security reasons, exposable devices need to be explicitly allowed by tmt
     runner, either via ``--exposable-runner-devices`` CLI option or
-    ``TMT_EXPOSABLE_RUNNER_DEVICES`` environment variable.
+    exposable device patterns.
```
LecrisUT: The change here does not fit well

skycastlelily: ah, right, thanks, talented and also hard-working breed ^^

docs/releases/pending/4705.fmf
```diff
@@ -2,6 +2,6 @@ description: |
   Adding new option of tmt command, ``--exposable-runner-devices``, and
   a corresponding environment variable, ``TMT_EXPOSABLE_RUNNER_DEVICES``.
   When a guest asks for a runner device to be exposed to said guest, e.g.
-  via podman's ``--device`` option or as a libvirt "node device", only
+  via podman's ``--expose-device`` option or as a libvirt "node device", only
```
thrix: @skycastlelily podman's option is really `--device`

thrix: this one is our new option

tmt/steps/provision/podman.py
```diff
@@ -684,11 +702,29 @@ class ProvisionPodman(tmt.steps.provision.ProvisionPlugin[ProvisionPodmanData]):
...
     In order to run the container with different user as the default ``root``,
     use ``user: USER``.
 
+    Use ``expose-device`` to expose host devices to the container. This is useful
+    for cases like KVM acceleration (``expose-device: /dev/kvm``) or accessing
+    serial devices. Multiple devices can be specified as a list.
+
+    For security reasons, exposable devices need to be explicitly allowed by tmt
+    runner, either via ``--exposable-runner-devices`` CLI option or the
+    'TMT_EXPOSABLE_RUNNER_DEVICES' environment variable.
```
thrix:
```suggestion
    ``TMT_EXPOSABLE_RUNNER_DEVICES`` environment variable.
```

thrix: let's be consistent here?

docs/releases/pending/4705.fmf
skycastlelily: Addressed:https://github.com/teemtee/tmt/pull/4333/commits/9f873b8eccc3eba27ed0949aa1dacd3669439eee :)

tmt/steps/provision/podman.py
skycastlelily: Addressed:https://github.com/teemtee/tmt/commit/9f873b8eccc3eba27ed0949aa1dacd3669439eee :)

LecrisUT:
rawhide test failure is a known issue fixed with a rebase

Issue #3805, opened 2025-06-09
Polarion Report creation failing


BeeGrech:
Something in the report generation of the log is being interpreted literally  it seams 

```
Cause number 1:

    unacceptable character #x001b: control characters are not allowed
      in "<unicode string>", position 390

        Traceback (most recent call last):

        File /usr/lib/python3.13/site-packages/tmt/base.py, line 4504, in go
          plan.go()

          self = <tmt.base.Run object at 0x7f90ca3d4ad0>
          enabled_steps = {'execute', 'finish', 'discover', 'prepare', 'report', 'provision'}
          all_steps = True
          since = None
          until = None
          after = None
          before = None
          skip = ()
          first = 0
          last = 5
          index = 5
          step = 'finish'
          crashed_plans = []
          plan = <tmt.base.Plan object at 0x7f90ca3d57f0>

        File /usr/lib/python3.13/site-packages/tmt/base.py, line 2876, in go
          self.report.go()

          self = <tmt.base.Plan object at 0x7f90ca3d57f0>
          step = <tmt.steps.execute.Execute object at 0x7f90cabc3a10>
          standalone = set()
          standalone_plugins = 0
          abort = False
          tests = [TestOrigin(phase='selinux_state_check', test=<tmt.base.Test object at 0x7f90cabc2e40>), TestOrigin(phase='Save tests/all last commit hash', test=<tmt.base.Test object at 0x7f90ca3fd1d0>), TestOrigin(phase='default-0', test=<tmt.base.Test object at 0x7f90ca98b4d0>), TestOrigin(phase='Check stack tracing results', test=<tmt.base.Test object at 0x7f90c8adf750>)]

        File /usr/lib/python3.13/site-packages/tmt/steps/report/__init__.py, line 144, in go
          phase.go()

          self = <tmt.steps.report.Report object at 0x7f90cabc2120>
          force = False
          phase = <tmt.steps.report.polarion.ReportPolarion object at 0x7f90cac5b750>
          __class__ = <class 'tmt.steps.report.Report'>

        File /usr/lib/python3.13/site-packages/tmt/steps/report/polarion.py, line 420, in go
          xml_data = make_junit_xml(

          self = <tmt.steps.report.polarion.ReportPolarion object at 0x7f90cac5b750>
          logger = None
          find_polarion_case_ids = <function find_polarion_case_ids at 0x7f90ca7c68e0>
          import_polarion = <function import_polarion at 0x7f90ca7c6700>
[...]

PR #4334, opened 2025-11-18
Refactor "to YAML" and "from YAML" helpers

happz:
* naming: use `to_yaml` and `from_yaml` first, to be aligned with the rest of codebase (e.g. `to_spec`).
* `to_yaml` and `from_yaml` are the primary work horses: most of the time, caller does not care about the data types at all.
* `yaml_to_list` and `yaml_to_dict` survive, but may eventually become replaced with some Pydantic magic, as they are not just about dict or list, but also types of keys, values, and items, and our code cannot validate that.
* YAML loader/dumper creation separated from its use. Eventually we may cache the object, but it's unknown whether the object itself can be used in multiple threads safely.
* some extra documentation, keyword-only arguments, etc.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/export/nitrate.py
+++ b/tmt/export/nitrate.py
@@ -100,14 +100,11 @@ def _nitrate_find_fmf_testcases(test: 'tmt.Test') -> Iterator[Any]:
                 struct_field = StructuredField(testcase.notes)
                 try:
                     fmf_field = struct_field.get('fmf')
+                    if isinstance(fmf_field, list):
+                        fmf_field = '\n'.join(fmf_field)
 
                     fmf_id = tmt.base.FmfId.from_spec(
-                        cast(
-                            tmt.base._RawFmfId,
-                            tmt.utils.yaml_to_dict(
-                                fmf_field if isinstance(fmf_field, str) else '\n'.join(fmf_field)
-                            ),
-                        )
+                        cast(tmt.base._RawFmfId, tmt.utils.yaml_to_dict(fmf_field))
                     )
                     if fmf_id == test.fmf_id:
                         echo(
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -3361,7 +3361,8 @@ def filter_paths(directory: Path, searching: list[str], files_only: bool = False
 
 
 #: Which type of YAML loader/dumper implementation to use when creating
-#: a YAML loader/dumper instance. See :py:class:`YAML` for details.
+#: a YAML loader/dumper instance. See :py:class:`!ruamel.yaml.YAML` for
+#: details.
 YamlTypType = Literal['rt', 'safe', 'unsafe', 'base']
 
 
@@ -3428,6 +3429,75 @@ def _yaml(
     return yaml
 
 
+def _sanitize_yaml_string(s: str) -> str:
+    """
+    Convert multiline strings, sanitize invalid characters.
+
+    Prevents saving non-printable characters a YAML parser might later
+    reject - see https://github.com/teemtee/tmt/issues/3805.
+
+    Based on :py:meth:`!ruamel.yaml.scalarstring.walk_tree` which does
+    not support any other test than "is this character in that string?".
+    """
+
+    pattern = ruamel.yaml.reader.Reader.NON_PRINTABLE
+
+    if '\n' in s:
+        s = ruamel.yaml.scalarstring.preserve_literal(s)
+
+    return ''.join(rf'#{{{ord(c):x}}}' if pattern.match(c) else c for c in s)
+
+
+def _sanitize_yaml_tree(value: Any, sort_keys: bool) -> Any:
+    """
+    Convert multiline strings, sanitize invalid characters.
+
+    Prevents saving non-printable characters a YAML parser might later
+    reject - see https://github.com/teemtee/tmt/issues/3805.
+
+    Based on :py:meth:`!ruamel.yaml.scalarstring.walk_tree` which does
+    not support any other test than "is this character in that string?".
+
+    :param sort_keys: if set, sort mapping keys.
+    """
+
+    from collections.abc import MutableMapping, MutableSequence
+
+    if isinstance(value, MutableMapping):
+        if sort_keys:
+            # Sort the data https://stackoverflow.com/a/40227545
+            sorted_value = CommentedMap()
+
+            for key in sorted(value):
+                sorted_value[key] = value[key]
+
+            value = sorted_value
+
+        for k, v in value.items():
+            if isinstance(v, str):
+                value[k] = _sanitize_yaml_string(v)
+
+            else:
+                value[k] = _sanitize_yaml_tree(v, sort_keys)
+
+        return value
+
+    if isinstance(value, MutableSequence):
+        for k, v in enumerate(value):
+            if isinstance(v, str):
+                value[k] = _sanitize_yaml_string(v)
+
+            else:
+                value[k] = _sanitize_yaml_tree(v, sort_keys)
+
+        return value
+
+    if isinstance(value, str):
+        return _sanitize_yaml_string(value)
+
+    return value
+
+
 def to_yaml(
     data: Any,
     *,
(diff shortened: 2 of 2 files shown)


tmt/utils/__init__.py
```diff
@@ -3334,62 +3390,87 @@ def walk_tree(value: Any) -> Any:
...
-    return YAML(typ=yaml_type).load(data)
+    try:
+        return _yaml(yaml_type=yaml_type).load(data)
+
+    except ParserError as error:
+        raise GeneralError('Invalid YAML syntax.') from error
 
 
-def yaml_to_dict(data: Any, yaml_type: Optional[YamlTypType] = None) -> dict[Any, Any]:
+def yaml_to_dict(data: str, *, yaml_type: Optional[YamlTypType] = None) -> dict[T, S]:
```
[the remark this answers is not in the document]
LecrisUT: mypy seems fine with it, although this is not covered by pyright

vaibhavdaren (approved):

tmt/utils/__init__.py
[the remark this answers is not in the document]
happz: Well, it's a fair point, Python type annotations do not work this way after all, and `dict[Any, Any]` would be a better approach.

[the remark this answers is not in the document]
happz: Dropped in ee1cd5cf.

tmt/export/nitrate.py
```diff
@@ -99,8 +99,15 @@ def _nitrate_find_fmf_testcases(test: 'tmt.Test') -> Iterator[Any]:
...
                 struct_field = StructuredField(testcase.notes)
                 try:
+                    fmf_field = struct_field.get('fmf')
+
                     fmf_id = tmt.base.FmfId.from_spec(
-                        cast(tmt.base._RawFmfId, tmt.utils.yaml_to_dict(struct_field.get('fmf')))
+                        cast(
+                            tmt.base._RawFmfId,
+                            tmt.utils.yaml_to_dict(
+                                fmf_field if isinstance(fmf_field, str) else '\n'.join(fmf_field)
```
LecrisUT: Why not move this `if else` near to where `fmf_field` was defined?

tmt/utils/__init__.py
```diff
@@ -3358,22 +3360,56 @@ def filter_paths(directory: Path, searching: list[str], files_only: bool = False
     return [Path(path) for path in set(found_paths)]  # return all matching unique paths as Path's
 
 
-def dict_to_yaml(
-    data: Union[dict[str, Any], list[Any], 'tmt.base._RawFmfId'],
+#: Which type of YAML loader/dumper implementation to use when creating
+#: a YAML loader/dumper instance. See :py:class:`YAML` for details.
```
```suggestion
#: a YAML loader/dumper instance. See :py:class:`!ruamel.yaml.YAML` for details.
```
The `!` is to surpress the linkage not existing since https://yaml.dev/doc/ruamel.yaml/ does not use sphinx so we would not be able to use `intersphinx`.

tmt/utils/__init__.py
```diff
@@ -3358,22 +3360,56 @@ def filter_paths(directory: Path, searching: list[str], files_only: bool = False
...
+        See :py:class:`YAML` for details.
+    :param width: if set, enforce this as the maximal length of lines.
+    :param start: if set, a document start marker, ``---``, would be
+        emitted before each dumped object.
+    :returns: a loader/dumper instance.
     """
 
-    output = io.StringIO()
-    yaml = YAML()
+    yaml = YAML(typ=yaml_type)
```
LecrisUT, line 3407: Would prefer a
```python
yaml_type = yaml_type or "rt"
```
Although not sure if we should be using `rt` as default everywhere, particularly if we do not save the yaml object somewhere to be reused.

tmt/utils/__init__.py
```diff
@@ -3386,17 +3422,37 @@ def dict_to_yaml(
...
+    return yaml
+
+
+def to_yaml(
+    data: Any,
+    *,
+    yaml_type: Optional[YamlTypType] = None,
+    width: Optional[int] = None,
+    sort: bool = False,
+    start: bool = False,
```
LecrisUT, line 3507: How about adding the optional `yaml` here to make room for reusing the yaml object for round-trip `to_yaml`? Although in that case we would not be outputting to string 🤔

tmt/utils/__init__.py
LecrisUT: Not necessary to be in the function right?

PS: this seems like it would be more efficient if done in the `Dumper` class. But do we have a reference of why this is not done by default in ruamel?

tmt/utils/__init__.py
```diff
@@ -3441,62 +3497,87 @@ def walk_tree(value: Any) -> Any:
...
         # Sort the data https://stackoverflow.com/a/40227545
         sorted_data = CommentedMap()
+
         for key in sorted(data):
             # ignore[literal-required]: `data` may be either a generic
             # dictionary, or _RawFmfId which allows only a limited set
             # of keys. That spooks mypy, but we do not add any keys,
             # therefore we will not escape TypedDict constraints.
-            sorted_data[key] = data[key]  # type: ignore[literal-required]
+            sorted_data[key] = data[key]
```
LecrisUT: Do we actually still need this according to https://stackoverflow.com/a/69433777?

happz: I think we do, exactly because `data` is ordered in the order in which keys were added, while `sorted_data` should be ordered alphabeticaly.

tmt/utils/__init__.py
happz: Once we find the use case for it, we can add the argument.

tmt/utils/__init__.py
happz: I'd prefer doing this by changing the default value. I let it be on purpose, leaving it not addressed in this PR, as it may affect more bits besides just refactoring the API.

tmt/utils/__init__.py
LecrisUT: Yep, confirmed that is the case. Weird that my IDE is showing it in the ordered form until I actually open it :shrug:
```python
>>> from ruamel.yaml import YAML
... from io import StringIO
... yaml = YAML()
... stream = StringIO()
... data = {}
... data["xyz"] = 1
... data["abc"] = 2
... print(data)
... yaml.dump(data, stream)
... print(stream.getvalue())
{'xyz': 1, 'abc': 2}
xyz: 1
abc: 2
```

tmt/utils/__init__.py
LecrisUT: Raised them as follow-up tasks in https://github.com/teemtee/tmt/issues/4467

tmt/export/nitrate.py
happz: Can you propose how? I don't see the obvious way.

tmt/utils/__init__.py
happz: > Not necessary to be in the function right?

I don't follow here, what does not need to be in the function?

> PS: this seems like it would be more efficient if done in the `Dumper` class. But do we have a reference of why this is not done by default in ruamel?

Not really, I only found a couple of people with a similar issue, i.e. saving weird characters in YAML and failing to read them back, this was the recommended solution.

tmt/export/nitrate.py
LecrisUT: I mean on line 102
```python
                    fmf_field = struct_field.get('fmf')
                    if isinstance(fmf_field, list):
                        fmf_field = '\n'.join(fmf_field)
```

tmt/utils/__init__.py
LecrisUT: > > Not necessary to be in the function right?
> 
> I don't follow here, what does not need to be in the function?

I mean `_sanitize_yaml_string` and `walk_tree`. The functions do not appear to use anything defined within `to_yaml`, but they are redefined every time in there.

> > PS: this seems like it would be more efficient if done in the `Dumper` class. But do we have a reference of why this is not done by default in ruamel?
> 
> Not really, I only found a couple of people with a similar issue, i.e. saving weird characters in YAML and failing to read them back, this was the recommended solution.

Ok, still seems like a bug on ruamel side. Was it reported there?

tmt/export/nitrate.py
happz: Addressed in 0e901133

tmt/utils/__init__.py
happz: Addressed in 0e901133

tmt/utils/__init__.py
happz: > > > Not necessary to be in the function right?
> > 
> > 
> > I don't follow here, what does not need to be in the function?
> 
> I mean `_sanitize_yaml_string` and `walk_tree`. The functions do not appear to use anything defined within `to_yaml`, but they are redefined every time in there.

True. Moved in 0e901133, and I also move the key sorting bit, as it can be done while traversing the tree.

> > > PS: this seems like it would be more efficient if done in the `Dumper` class. But do we have a reference of why this is not done by default in ruamel?
> > 
> > 
> > Not really, I only found a couple of people with a similar issue, i.e. saving weird characters in YAML and failing to read them back, this was the recommended solution.
> 
> Ok, still seems like a bug on ruamel side. Was it reported there?

Not by me, and I  don't recall encountering an issue. That's one of the reasons I don't feel confident that 1. it's a bug or that 2. it's a bug in ruamel.yaml. Not my forte, I'm afraid. This is the original tmt issue: https://github.com/teemtee/tmt/issues/3805

LecrisUT: > Not by me, and I don't recall encountering an issue. That's one of the reasons I don't feel confident that 1. it's a bug or that 2. it's a bug in ruamel.yaml. Not my forte, I'm afraid. This is the original tmt issue: #3805

If only they did not use sourceforge :sob: 

> True. Moved in [0e90113](https://github.com/teemtee/tmt/commit/0e901133c8d6fd4098d6f3ced216b87a3d714fb5), and I also move the key sorting bit, as it can be done while traversing the tree.

:+1:

psss:
Seems that `tests/test_qe.py::test_missing_tmt` from the internal test suite is not skipped as expected when `tmt` is present.

```python
@pytest.mark.tmt
@pytest.mark.skipif(has_tmt(), reason="Cannot work with tmt available")
def test_missing_tmt():
```

Probably some import problem when detecting `used_tmt_version` in `qe.py`?

psss:
Yeah, this one will be the culprit:

```yaml
    from tmt.utils import dict_to_yaml
```

Btw, shall the `dict_to_yaml` reference be removed from here as well?

https://github.com/teemtee/tmt/blob/7ba640109197d46e094c38fbf1fd42bbcf64f6ee/docs/code/classes.rst?plain=1#L283

psss:
Ah, and I've found the internal change, merged. Sorry for the noise.

PR #4338, opened 2025-11-18
Change the default formatting of YAML emitted by tmt

### LINKED ISSUE #3432: Multiline attributes fails to interpret |2
tmt plan export does not interpret spaces scalar if attribute content starts with new line

How to reproduce

1. `tmt init --template mini`
2. Change plans/example.fmf so it contains
```
summary: |2

  Basic smoke test
  Next line to interpret
execute:
    how: tmt
    script: tmt --help
```
3. Parse  ```tmt plan export /plans/example | yq '.[0].summary'```

Expected 
```
Basic smoke test
Next line to interpret
```

Actual 
```
  Basic smoke test
  Next line to interpret
```
Note: If there is no new line between "summary: |2"
and "Basic smoke test"  - then export behaves as expected

Use case: if I need to pass a kickstart script where I use a file defined by heredoc (the script would not be able to catch the second EOF)

happz:
By setting `mapping`, `sequence`, and `offset` formatting hints for the YAML dumper, the produced YAML incorrectly handles multiline strings starting with an empty line, and export of collection items under `tmt * export`.

After some experimenting, there seem to be no silver bullet ticking all boxes WRT YAML formatting, but resorting to setting just `sequence=2` seems to produce a reasonably formatted YAML.

Note that this does not affect YAML - or fmf - accepted by tmt, only YAML/fmf files written by tmt.

Fixes https://github.com/teemtee/tmt/issues/3432

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -3524,7 +3524,15 @@ def _yaml(
 
     yaml = YAML(typ=yaml_type)
 
-    yaml.indent(mapping=4, sequence=4, offset=2)
+    # Setting `mapping` and `offset` may lead to a subpar-looking YAML
+    # formatting. With some settings, it's even possible to create invalid
+    # YAML, with keys incorrectly indented. Setting `sequence` only seems
+    # to produce a usable YAML (YMMV), and `sequence=2` seems to be acceptably
+    # formatted.
+    #
+    # Discussed in https://github.com/teemtee/tmt/issues/3432
+    yaml.indent(sequence=2)
+
     yaml.default_flow_style = False
     yaml.allow_unicode = True
     yaml.encoding = 'utf-8'
--- a/tests/unit/test_hardware.py
+++ b/tests/unit/test_hardware.py
@@ -305,82 +305,82 @@ Detailed validation error:
 def test_parse_maximal_constraint() -> None:
     hw_spec_out = """
         and:
+        - and:
+          - beaker.pool: '!= foo.*'
+          - beaker.panic-watchdog: == True
+        - boot.method: not contains exclusive bios
+        - and:
+          - compatible.distro: contains rhel-7
+          - compatible.distro: contains rhel-8
+        - and:
+          - cpu.processors: '> 8'
+          - cpu.sockets: <= 1
+          - cpu.cores: == 2
+          - cpu.threads: '>= 8'
+          - cpu.cores-per-socket: == 2
+          - cpu.threads-per-core: == 4
+          - cpu.model: == 62
+          - cpu.family: < 6
+          - cpu.vendor: == 32902
+          - cpu.stepping: '!= 10'
+          - cpu.frequency: '>= 2300.0 MHz'
+          - cpu.family-name: == Skylake
+          - cpu.model-name: '!~ Haswell'
+          - cpu.vendor-name: ~ Intel.*
           - and:
-              - beaker.pool: '!= foo.*'
-              - beaker.panic-watchdog: == True
-          - boot.method: not contains exclusive bios
+            - cpu.flag: contains avx
+            - cpu.flag: contains avx2
+            - cpu.flag: not contains smep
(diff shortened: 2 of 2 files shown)


happz:
Blocked by https://github.com/teemtee/tmt/pull/4334

tmt/utils/__init__.py
```diff
@@ -3524,7 +3524,15 @@ def _yaml(
...
 
-    yaml.indent(mapping=4, sequence=4, offset=2)
+    # Setting `mapping` and `offset` may lead to a subpar-looking YAML
+    # formatting. With some settings, it's even possible to create invalid
+    # YAML, with keys incorrectly indented. Setting `sequence` only seems
+    # to produce a usable YAML (YMMV), and `sequence=2` seems to be acceptably
+    # formatted.
+    #
+    # Discussed in https://github.com/teemtee/tmt/issues/3432
+    yaml.indent(sequence=2)
```
psss, line 3534: I checked the reproducer before and after the change and I see the same output for both:

    > cat plan.fmf
    summary: |2

      Basic smoke test
      Next line to interpret
    execute:
        how: tmt
        script: tmt --help

    > tmt plan export | yq '.[0].summary'

    Basic smoke test
    Next line to interpret

Is the change really needed? If so, what about adding the reproducer to the test coverage?

psss (commented):
I'm not able to reproduce the original issue.

tmt/utils/__init__.py
happz: Indeed, it is no longer reproducible. It has been fixed in 4ef550eaed4f73af5c86d936edf7e290bb69da53, according to `git bisect`, by accident.

> Is the change really needed? If so, what about adding the reproducer to the test coverage?

I'd say yes, it's still worth the effort. Compare the actual YAML output before and after the patch:

```yaml
(dev) happz@multivac [main] ~/git/tmt/tmp $ tmt plan export
  - name: /
    summary: "\nBasic smoke test\nNext line to interpret\n"
    description:
    author: []
    contact: []
    enabled: true
    order: 50
```

```yaml
(dev) happz@multivac [yaml-indent-bug] ~/git/tmt/tmp $ tmt plan export
- name: /
  summary: "\nBasic smoke test\nNext line to interpret\n"
  description:
  author: []
  contact: []
  enabled: true
  order: 50
```

Note the indentation of items of the top-level list, `  - name: /` vs `- name: /`. The former is valid YAML, and acceptable by various tools like `yq`, but I've seen a parser being fooled by the extra indentation. I don't have the said tool at hand, I'm afraid.

psss: > The former is valid YAML, and acceptable by various tools like `yq`, but I've seen a parser being fooled by the extra indentation.

Hm, that's surprising for me, I thought that the indented list is a perfectly valid `yaml`. We use it basically in all our `fmf` examples, right?

happz: Yes, it's valid YAML. But we are talking about the top-level entity living in the file and how that one is indented, not about lists in general, only the very first one. Apparently it may fool some less precise code when the YAML entity not starting at column 0, but at column N.

```yaml
# foo.yaml - valid YAML
---

- 1
- 2

# bar.yaml - also valid YAML, just unexpected
---

  - 1
  - 2

# baz.yaml - both valid, but not the lists in question
baz:
  quux:
    - 1
    - 2
  qux:
  - 1
  - 2
```

LecrisUT (approved):
👍 for me, it is at least consistent with #4712. I just realized that

> The C based emitter doesn’t have the fine control, distinguishing between block mappings and sequences. Do only use the pure Python versions of the dumper if you want to have that sort of control.

🙃

therazix (approved):

tmt/utils/__init__.py
psss: Ok, I see. Thanks for clarification.

psss (approved):

PR #4346, opened 2025-11-20
Deprecate and internalize discover/fmf fields

### LINKED ISSUE #4493: Deprecate and internalize discover/fmf fields
* `upgrade_path` should be made internal
* `repository` and `revision` should be deprecated

therazix:
Makes the `upgrade_path` field internal and deprecates the `repository` and `revision` fields. Also adds a new `check_deprecated_keys` method to the `DataContainer` class that emits a warning if any deprecated field is specified.

Resolves #4493

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/container/__init__.py
+++ b/tmt/container/__init__.py
@@ -345,6 +345,29 @@ class DataContainer:
 
         return {key: value for key, value in self.items() if value is not None}
 
+    def check_deprecated_keys(self, raw_data: dict[str, Any], logger: 'tmt.log.Logger') -> None:
+        """
+        Check for deprecated keys and emit warnings if they are used.
+
+        This method iterates through all fields in the container and checks their
+        metadata for deprecation information. If a deprecated field is present in
+        the raw data, a warning is emitted.
+        """
+
+        for field in container_fields(self):
+            _, option, _, _, metadata = container_field(self, field.name)
+
+            if metadata.deprecated is None:
+                continue
+
+            if option not in raw_data:
+                continue
+
+            logger.warning(
+                f"Field '{option}' is deprecated since {metadata.deprecated.since}"
+                f"{', ' + metadata.deprecated.hint if metadata.deprecated.hint else ''}."
+            )
+
     # This method should remain a class-method: 1. list of keys is known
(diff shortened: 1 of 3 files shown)


tmt/container/__init__.py
```diff
@@ -345,6 +345,29 @@ def to_minimal_dict(self) -> dict[str, Any]:
...
+            if metadata.deprecated is None:
+                continue
+
+            if option not in raw_data:
+                continue
+
+            logger.warning(
+                f"Field '{option}' is deprecated since {metadata.deprecated.since}"
+                f"{', ' + metadata.deprecated.hint if metadata.deprecated.hint else ''}."
+            )
```
LecrisUT, line 369: Reckon you could squeeze a change to either add a class to the warning like `DeprecationWarning` or/and a `topic`? Would be nice to actually start having a way to filter and categorize these warnings.

tmt/steps/__init__.py
```diff
@@ -435,6 +435,7 @@ def post_normalization(self, raw_data: _RawStepData, logger: tmt.log.Logger) ->
         """
         Called after normalization, useful for tweaking normalized data
         """
+        self.check_deprecated_keys(cast(dict[str, Any], raw_data), logger)
```
happz, line 438: Just to be sure, check whether there's anything defining its own `post_normalization` method, and whether it calls `super().post_normalization(...)` correctly.

tmt/container/__init__.py
happz: That could be useful, although I would leave that for another PR, because we do already have plenty of `warning()` calls, and it would be nice to tackle the larger group with such an enhancement.

tmt/steps/__init__.py
LecrisUT: I checked. It is DiscoverFmf and it does

https://github.com/teemtee/tmt/blob/56751d43901ae398900e0c6e73cf671147a6a24a/tmt/steps/discover/fmf.py#L226-L231

PR #4354, opened 2025-11-21
Fix `/tests/execute/unresponsive` expected message

psss:
The main test purpose is to verify that error is correctly reported. There can be different error messages when the guest becomes unresponsive, let's check for both to prevent random flakes.

Pull Request Checklist

### DIFF
--- a/tests/execute/unresponsive/test.sh
+++ b/tests/execute/unresponsive/test.sh
@@ -4,11 +4,12 @@
 rlJournalStart
     rlPhaseStartSetup
         rlRun "pushd data"
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-virtual}"
     rlPhaseEnd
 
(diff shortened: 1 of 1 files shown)


psss:
Adding to the current sprint to prevent random test failures. I've seen that one several times this week.

happz:
@psss would you have an example where an unexpected message appeared? I'm curious about the "error is correctly reported" statement - if different error appears, is it the correct way? Plus, rsync and pull are notoriously complicated, I'd like to make sure we understand why a different error message appears.

psss:
Sure, here's a recent job:

* https://artifacts.osci.redhat.com/testing-farm/9ad06be2-820b-4a22-98c1-d96944f3fc89/

happz:
> Sure, here's a recent job:
> 
> * https://artifacts.osci.redhat.com/testing-farm/9ad06be2-820b-4a22-98c1-d96944f3fc89/

Ok, that looks acceptable.

PR #4355, opened 2025-11-21
Implement configurable additional git locations for libraries

martinky82:
Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/libraries/beakerlib.py
+++ b/tmt/libraries/beakerlib.py
@@ -27,6 +27,11 @@ LIBRARY_REGEXP = re.compile(r'^library\(([^/]+)(/[^)]+)\)$')
 DEFAULT_REPOSITORY_TEMPLATE = 'https://github.com/beakerlib/{repository}'
 DEFAULT_DESTINATION = 'libs'
 
+# Additional beakerlib libraries repository (from environment variable)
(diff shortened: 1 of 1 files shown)


thrix:
@martinky82 @psss suggested to talk this through again on a separate meeting, the most of the team is not very fond supporting this.

psss:
Summary from the sync meeting:

There is a proposal to completely remove beaker task library from the image mode composes

 * Seems this could handle about 90% of the cases
 * Expect that just a few leftover will have to be handled
 * Definitely a good choice, as anything there breaks IM jobs
 * Let’s enable this and see what the impact will be

If it helps and there is not a large number of tests which would have to be modified, it would be ok to drop this pull request.

Long-term we would like to deprecate the beaker taks library usage completely. To be added to the high-level roadmap.

martinky82:
@psss After disabling the tasks repo, fixing the epel library and tweaks to the tests here and there, I reckon the testing OK(ish) with tmt as is in this regard, hence I am ok with dropping this PR.

psss:
Ok, good, thanks!

PR #4356, opened 2025-11-22
Add Pydantic validation for reboot configuration

### LINKED ISSUE #4349: Validate reboot request file has the right format
Validate the data by defining class and check the data are matching the expected format.

see: https://github.com/teemtee/tmt/pull/4232#discussion_r2477930433

AthreyVinay:
Resolves #4349 

Changes made:
- Define `RebootData` Pydantic model to validate config structure
- Simplify downstream logic since all fields are now guaranteed valid

~A few `TODOs` to be reviewed.~

### DIFF
--- a/tmt/_compat/pydantic.py
+++ b/tmt/_compat/pydantic.py
@@ -30,6 +30,15 @@ if PYDANTIC_V1:
                 )
             return cls.parse_obj(obj)
 
+        @classmethod
+        def model_validate_json(cls, json_data: str, **kwargs: Any) -> Self:  # type: ignore[override]
+            if kwargs:
+                raise NotImplementedError(
+                    "Backport of model_validate_json to parse_raw does not include kwargs."
+                )
+
+            return cls.parse_raw(json_data, content_type="application/json")
+
         def model_dump(self, **kwargs: Any) -> dict[str, Any]:
             return self.dict(**kwargs)
 


tmt/steps/context/reboot.py
```diff
@@ -17,6 +17,21 @@
...
+class RebootData(BaseModel):
+    """
+    Data structure representing reboot request details.
+    """
+
+    model_config = ConfigDict(extra='forbid')  # or maybe be less strict?
+
+    command: Optional[str] = None
+    timeout: int = tmt.steps.provision.REBOOT_TIMEOUT
+    # TODO: Do we need more fields here - not used in the code but exists in the documentation?
```
happz: As of now, no other fields are supported.

tmt/steps/context/reboot.py
```diff
@@ -121,21 +136,17 @@ def handle_reboot(self, restart: Optional['RestartContext'] = None) -> bool:
 
         elif self.soft_requested:
             # Extract custom hints from the file, and reset it.
-            reboot_data = json.loads(self.request_path.read_text())
+            reboot_data = RebootData.model_validate_json(
+                self.request_path.read_text()
+            )  # TODO: Do we fail hard or handle more gracefully?
```
happz: Hard. There is no recovery from this: a script asked the guest to be rebooted, by calling `tmt-reboot`, and we get a broken reboot request file.

AthreyVinay: Thanks, `TODO's` removed so.

tmt/steps/context/reboot.py
```diff
@@ -17,6 +17,17 @@
     from tmt.steps.context.restart import RestartContext
 
 
+class RebootData(BaseModel):
+    """
+    Data structure representing reboot request details.
+    """
+
+    model_config = ConfigDict(extra='forbid')
```
LecrisUT:
```suggestion
    if PYDANTIC_V1:
        class Config:
            extra = "forbid"
    else:
        model_config = ConfigDict(
            extra="forbid",
        )
```
Welcome to dealing with backwards-compatibility

tmt/steps/context/reboot.py
```diff
@@ -121,21 +132,15 @@ def handle_reboot(self, restart: Optional['RestartContext'] = None) -> bool:
 
         elif self.soft_requested:
             # Extract custom hints from the file, and reset it.
-            reboot_data = json.loads(self.request_path.read_text())
+            reboot_data = RebootData.model_validate_json(self.request_path.read_text())
```
LecrisUT: Do you need to pass it through `model_validate_json`? Isn't there a direct constructor that could be used (slowly wrapping my head around pydantic)

tmt/steps/context/reboot.py
```diff
@@ -121,21 +132,15 @@ def handle_reboot(self, restart: Optional['RestartContext'] = None) -> bool:
...
 
             reboot_command: Optional[ShellScript] = None
 
-            if reboot_data.get('command'):
+            if reboot_data.command:
                 with suppress(TypeError):
```
LecrisUT: Is the `surpress` still needed here?

tmt/steps/context/reboot.py
happz: `MetadataContainer`, used to hold fmf data (configuration for now, more later) uses `cls.model_validate(tmt.utils.yaml_to_dict(yaml))`: https://github.com/teemtee/tmt/blob/main/tmt/container/__init__.py#L950

It would be good to use the same approach here, and think about the change that would cover `MetadataContainer` as well, if such a change is needed.

BTW we could use `MetadataContainer` here as well, it would bring the `Config` setup, add a new `from_json()` class method there, edit the docstring to speak about "user-provided structured data, like fmf nodes" rather than "fmf nodes" alone, and call it a day. If we decide using a direct constructor is better, we would change that class, fixing it for everyone.

tmt/steps/context/reboot.py
AthreyVinay: Always a JOY!

tmt/steps/context/reboot.py
AthreyVinay: Sorry - are we discussing about the consistency across the codebase or just the usage of `model_validate_json` so that it could be extended ?

tmt/steps/context/reboot.py
AthreyVinay: Nice catch - we do not need this as pydantic guarantees a `str` or `None`.

tmt/steps/context/reboot.py
happz: > Sorry - are we discussing about the consistency across the codebase or just the usage of `model_validate_json` so that it could be extended ?

I guess both :)

Your code reads structured data from a user-provided file, which is exactly what `MetadataContainer` already does, only with fmf files. Therefore, it would be good if we had one answer to "what do I need to take care of to fill this class with data from that JSON/YAML file?". Preferably if the answer would be "call that class' `from_yaml()` or `from_json()` method. Clean, understandable API, with no chance of misunderstanding. So I would propose adding `from_json()` method to the `MetadataContainer` class, derive your new class from the `MetadataContainer`, and call `from_json()` here. This is the "consistency across codebase" bit.

What happens inside these methods is basically the original question, the one about the usage of `model_validate_json`. `from_yaml()` uses `model_validate()` which sounds like the same kind of Pydantic API in question. I don't know whether the yare the best or simplest, `model_validate()` is something I got from Pydantic docs. We can decide against using these methods, or in favor of using them, but that will affect only the internals of these simple API methods. For a common developer, the simple API would remain, without the need to worry what Pydantic method to use..

AthreyVinay: Cheers @happz - thanks for your thoughts.

Adding a `from_json()` method to `MetadataContainer` class makes sense just to stay consistent - however the usage the of `model_validate()` for this feels incorrect simply because Pydantic exposes `model_validate_json()` https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_validate_json for the sole purpose of dealing with raw json data. Thoughts?

tmt/steps/context/reboot.py
AthreyVinay: Addressed here: https://github.com/teemtee/tmt/pull/4356/commits/2a53f2fffb0c4cf2aab905682c18d0cc6f48c1ba

tmt/steps/context/reboot.py
AthreyVinay: Addressed here: https://github.com/teemtee/tmt/pull/4356/commits/7f3f53836a912e0a70bd70f8df88acdd2609844f

tmt/steps/context/reboot.py
AthreyVinay: This has now been refactored / removed due to the inheritance from `MetadataContainer`

vaibhavdaren:
The comments on this PR are nice to read. They give a clear understanding of base models. If our tests are failing probably they will fail for other users too. I think we should add a release note for this as well.

tmt/steps/scripts/tmt-reboot-core
```diff
@@ -19,7 +19,15 @@ read -r test_pid test_reboot_file < "$TMT_TEST_PIDFILE"
...
-printf '{"command": "%s", "timeout": "%s"}' "$command" "$timeout" > "$test_reboot_file"
+# Build reboot-request JSON, including only non-empty command/timeout to satisfy
+# strict Pydantic validation.
+{
+    printf '{'
+    first=true
+    [ -n "$command" ] && { printf '"command": "%s"' "$command"; first=false; }
+    [ -n "$timeout" ] && { $first || printf ', '; printf '"timeout": %s' "$timeout"; }
+    printf '}'
+} > "$test_reboot_file"
```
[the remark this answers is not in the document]
AthreyVinay: Default is handled by pydantic.

AthreyVinay:
> I think we should add a release note for this as well.

I think its a valid comment - since the contract is now changed i.e. empty strings which were allowed will **NOT** be anymore. Anybody with more thoughts / opinions?

happz:
> > I think we should add a release note for this as well.
> 
> I think its a valid comment - since the contract is now changed i.e. empty strings which were allowed will **NOT** be anymore. Anybody with more thoughts / opinions?

Yes, a release note can't hurt.

tmt/steps/scripts/tmt-reboot-core
```diff
@@ -19,7 +19,15 @@ read -r test_pid test_reboot_file < "$TMT_TEST_PIDFILE"
 
 mkdir -p "$(dirname "$test_reboot_file")"
 
-printf '{"command": "%s", "timeout": "%s"}' "$command" "$timeout" > "$test_reboot_file"
+# Build reboot-request JSON, including only non-empty command/timeout to satisfy
+# strict Pydantic validation.
+{
+    printf '{'
+    first=true
+    [ -n "$command" ] && { printf '"command": "%s"' "$command"; first=false; }
```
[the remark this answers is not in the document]
AthreyVinay: I did apply the change @LecrisUT using `jq` as I feel it is a cleaner approach. No?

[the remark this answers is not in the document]
LecrisUT: No implied dependencies in scripts. `jq` would not be available on out-of-the-box containers

tmt/steps/scripts/tmt-reboot-core
```diff
@@ -19,7 +19,12 @@ read -r test_pid test_reboot_file < "$TMT_TEST_PIDFILE"
 
 mkdir -p "$(dirname "$test_reboot_file")"
 
-printf '{"command": "%s", "timeout": "%s"}' "$command" "$timeout" > "$test_reboot_file"
+# Build reboot-request JSON, including only non-empty command/timeout to satisfy
+# strict Pydantic validation.
+jq -n --arg cmd "$command" --argjson timeout "${timeout:-null}" '
```
happz: This would require `jq` to be predemt on every guest, every distro, every architecture. I am afraid you need to stick to bash.

tmt/steps/scripts/tmt-reboot-core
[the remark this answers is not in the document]
AthreyVinay: Reverted.

tmt/steps/scripts/tmt-reboot-core
AthreyVinay: Reverted.

tmt/container/__init__.py
```diff
@@ -967,3 +967,15 @@ def from_yaml(cls, yaml: str) -> Self:
...
+
+    @classmethod
+    def from_json(cls, json_data: str) -> Self:
+        import tmt.utils
+
+        try:
+            return cls.model_validate_json(json_data)
+
+        except ValidationError as error:
+            import tmt.utils
```
therazix: Is this import needed when there is the same one just a couple of lines above?

AthreyVinay: You're right in that import isn't strictly necessary. I'm keeping it because other methods in this module follow the same import pattern, and i wanted to follow consitency. Although it doesnt get re-imported since it is cached.

LecrisUT: Only one case above it, seems to have slipped in a similar manner. Not a big deal, but unfortunate that we do not have a PLC0415 that could catch some of these

AthreyVinay: @therazix Thanks & addressed here: https://github.com/teemtee/tmt/pull/4356/commits/ffb8cc01645d1e7f668372040d5532dcaec86a93

tmt/container/__init__.py
```diff
@@ -964,6 +964,14 @@ def from_yaml(cls, yaml: str) -> Self:
...
-            import tmt.utils
-
             raise tmt.utils.SpecificationError("Invalid metadata in YAML data.") from error
+
+    @classmethod
+    def from_json(cls, json_data: str) -> Self:
+        import tmt.utils
+
+        try:
+            return cls.model_validate_json(json_data)
```
happz, line 974: It turns out that `model_validate_json()` is not available with Pydantic 1.x. We have two options: provide a shim in `tmt._compat.pydantic.BaseModel`, or modify this method to use `json.loads()` + `model_validate()`, just like `from_yaml()` does. The former might be better for the overall usability of our models.

AthreyVinay: Yep,, saw the fail, I have added the shim  - [Add Pydantic v1 shim: implement model_validate_json fallbackc](https://github.com/teemtee/tmt/pull/4356/commits/253c55f16c09c25b64f71ad9e6e59ed1b6f966c6)

happz:
Unrelated failures, merging.

PR #4365, opened 2025-11-26
Prevent execute action in tmt try from overwriting guest modifications

### LINKED ISSUE #4046: 'execute' in tm-try rediscovers the test from the host
Hi,

```
$ rpm -q tmt
tmt-1.56.0-1.fc42.noarch
```

if I use:

```
$ tmt try fedora42
```

to get the machine and run test, choose 'l' to login, update the test on the guest machine, ctrl+d to get tmt-try menu and rerun the updated test by choosing 'e' as 'execute' without reuploading test, the original  test from host is used.

IIUC 'e' should not rediscover and not reupload the test from host, and the test from guest should be used - or is it incorrect?

If that is not right approach, how can I run the test which is updated on the guest via tmt-try commands? Except for running `./test.sh` on the guest.

Thank you in advance!

skycastlelily:
Fixes: #4046

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -519,6 +519,10 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
     #: If set, plugin should run tests only from this discover phase.
     discover_phase: Optional[str] = None
 
+    #: Flag to indicate whether :py:meth:`Guest.push` should be invoked before each tests.
+    #: As it will overwrite remote content, some scenarios, namely ``tmt try``, need to avoid it.
+    skip_guest_push: bool = False
+
     def __init__(
         self,
         *,
@@ -529,8 +533,6 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
     ) -> None:
         super().__init__(logger=logger, step=step, data=data, workdir=workdir)
         self._results: list[tmt.Result] = []
-        # Flag to indicate if this is a tmt try execute action that should not push files
-        self._tmt_try_no_push = False
         if tmt.steps.Login._opt('test'):
             self._login_after_test = tmt.steps.Login(logger=logger, step=self.step, order=90)
 
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -529,6 +529,8 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
     ) -> None:
         super().__init__(logger=logger, step=step, data=data, workdir=workdir)
         self._results: list[tmt.Result] = []
+        # Flag to indicate if this is a tmt try execute action that should not push files
+        self._tmt_try_no_push = False
         if tmt.steps.Login._opt('test'):
             self._login_after_test = tmt.steps.Login(logger=logger, step=self.step, order=90)
 
--- a/tmt/steps/execute/internal.py
+++ b/tmt/steps/execute/internal.py
@@ -560,7 +560,11 @@ class ExecuteInternal(tmt.steps.execute.ExecutePlugin[ExecuteInternalData]):
         test_invocations = self.prepare_tests(guest, logger)
 
         # Push workdir to guest and execute tests
-        guest.push()
+        # Skip push if this is an interactive execute-only action from tmt try
+        if not self._tmt_try_no_push:
+            guest.push()
+        else:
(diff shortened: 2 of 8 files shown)


tmt/trying.py
```diff
@@ -567,6 +567,13 @@ def action_execute(self, plan: Plan) -> None:
         Run tests using the specified executor
         """
 
+        # Set a flag to indicate this is an interactive execute-only operation
+        # that should not push files to the guest (preserving manual changes)
+        for phase in plan.execute.phases():
```
psss: But when the `test` action is used in `tmt try` the fresh test code should still be pushed from user laptop to the guest. This seems to disable the pushing altogether for any future `execute` step invocations? @skycastlelily, would it work? Could you please extend the test coverage to make sure we do not break the main functionality?

skycastlelily: Right, and addressed: https://github.com/teemtee/tmt/pull/4365/commits/ea7edda3f6a4af556e603855df66b725d47c3f22 :)

tmt/trying.py
```diff
@@ -567,7 +568,21 @@ def action_execute(self, plan: Plan) -> None:
...
 
-        plan.execute.go(force=True)
+        # Set a flag to indicate this is an interactive execute-only operation
+        # that should not push files to the guest (preserving manual changes)
+        phases_to_reset: list[Phase] = []
+        for phase in plan.execute.phases():
+            # Set the flag for execute plugins that support it
+            if hasattr(phase, '_tmt_try_no_push'):
+                setattr(phase, '_tmt_try_no_push', True)  # noqa: B010
+                phases_to_reset.append(phase)
```
happz: `phases()` accepts classes whose instances to return, therefore it should eb possible to look for execute plugins rather than fairly generic `Phase` instances, which would then give you the flag for free without using `hasattr` and `setattr` where type annotations do not matter.

```python
for phase in plan.execute.phases(classes=ExecutePlugin):
    ...

Also, when undoing, you presume the original value was `False`, which may not be correct. You should save the original value together with the phase:

```python
phases.append((phase, phase.skip_guest_push))
phase.skip_guest_push = True

...
for phase, original_skip_guest_push in phases_to_reset:
    phase.skip_guest_push = original_skip_guest_push
```

tmt/steps/execute/__init__.py
```diff
@@ -528,6 +528,8 @@ def __init__(
     ) -> None:
         super().__init__(logger=logger, step=step, data=data, workdir=workdir)
         self._results: list[tmt.Result] = []
+        # Flag to indicate if this is a tmt try execute action that should not push files
+        self._tmt_try_no_push = False
```
happz: 1. the use case does not need to be spelled in the attribute name is not necessary, that belongs into the documentation.
2. the flag should be declared on class-levels, where it can be nicely documented.
3. it does not need to be private, as it is expected to be set from outside.

Together:

```python
#: Flag to indicate whether :py:meth:`Guest.push` should be invoked before each tests.
#: As it will overwrite remote content, some scenarios, namely ``tmt try``, need to avoid it.
skip_guest_push: bool = False
```

tmt/steps/execute/internal.py
```diff
@@ -560,7 +560,11 @@ def _run_tests(
         test_invocations = self.prepare_tests(guest, logger)
 
         # Push workdir to guest and execute tests
-        guest.push()
+        # Skip push if this is an interactive execute-only action from tmt try
+        if not self._tmt_try_no_push:
+            guest.push()
+        else:
+            logger.debug("Skipping guest.push() to preserve manual changes from tmt try login.")
```
happz:
```suggestion
            logger.debug("Not pushing to guest before the test as requested.")
```

skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4365/commits/d2c54c9d13c024f0fa84d1c072f727b0c367854a ^^

tmt/steps/execute/__init__.py
skycastlelily: Addressed:https://github.com/teemtee/tmt/pull/4365/commits/d2c54c9d13c024f0fa84d1c072f727b0c367854a ^^

tmt/trying.py
skycastlelily: Thanks mentor, addressed: https://github.com/teemtee/tmt/pull/4365/commits/d2c54c9d13c024f0fa84d1c072f727b0c367854a ^^

PR #4366, opened 2025-11-26
Include `type: library` in the libraries example

psss:
The `type: library` part should be included when referencing `beakerlib` libraries. Include that in the example as well.

Pull Request Checklist

* [x] write the documentation


therazix:

Issue #4377, opened 2025-11-27
Inherit (some) phases from the importing plan


happz:
My use case: I have plans for a component, imported from plans owned and executed by a Tier-running group. Their jobs correctly import and execute my plans, but the results are not reported to RP as imported plans do not have any special `report` phases, and they do not inherit `report` phases from the importing plan.

It would be helpful if imported plan could inherit some of the phases of the importing plan.

psss:
Suggested in https://github.com/teemtee/tmt/issues/4496 to be merged before the module refactoring.
Proposing to finish before the housekeeping sprint.

PR #4379, opened 2025-12-01
Add support for step inheritance in plan imports

skycastlelily:
Fixes: #4377

 Introduces step inheritance functionality for plan imports, allowing plans to
 inherit step configurations from imported plans while maintaining the ability
 to override specific settings.
  How it works:
  1. The imported/remote plan remains unchanged 
  2. The importing/local plan uses adjust-plans to override or add steps to what it inherits
  3. Two modes of inheritance:
    - Replace mode (stepname:): Completely replaces the step from imported plan
    - Append mode (stepname+:): Adds to existing steps from imported plan
  If both remote and local has the same key,local one take priority.
  In conclusion, it works just  like adjust-tests for test adjustment

      Core Implementation
    
      - New adjust-plans field: Added to plan import schema supporting all adjust rule syntax
      - Step inheritance: Import remote plans and selectively modify/add steps locally
      - Full feature parity: Supports all adjust-tests features including conditional rules, pattern matching
    
      Supported Operations
    
      - Step addition (report+:): Append local steps to imported plan steps
      - Step replacement (report:): Replace imported plan steps completely
      - Multiple plugins: Add multiple report plugins with list syntax
      - Conditional rules: Use when/because clauses with context evaluation
      - Pattern matching: Support == operator in conditional expressions

 Step Inheritance Examples:

```
  1. Append Mode - Add Local Steps to Remote Plan

  plan:
      import:
          name: /plans/remote
          adjust-plans:
              report+:                    # '+' means append
                  how: junit
                  file: /tmp/local-results.xml
              finish+:                    # Add additional finish step
                  how: shell
                  script: echo "Local cleanup after remote plan"

  2. Replace Mode - Override Remote Steps

  plan:
      import:
          name: /plans/remote
          adjust-plans:
              report:                     # No '+' means replace
                  how: html
                  file: /tmp/replacement.html
              provision:                  # Replace remote provision
                  how: local

  3. Multiple Step Adjustments

  plan:
      import:
          name: /plans/remote
          inherit-context: false          # Don't inherit context
          inherit-environment: false      # Don't inherit environment
          adjust-plans:
              discover+:                  # Append discovery step
                  how: fmf
                  where: /additional/tests
              prepare:                    # Replace prepare step
                  how: ansible
                  playbook: local-setup.yml
              report+:                    # Add local reporting
                  how: junit
                  file: /tmp/junit-results.xml
              finish+:                    # Add cleanup
                  how: shell
                  script: |
                      echo "Cleaning up local resources"
                      rm -rf /tmp/test-data

      Multiple report plugins:
    
      plan:
          import:
              name: /plans/remote
              adjust-plans:
                - report+:
                    - how: junit
                      file: /tmp/junit-results.xml
                    - how: html
                      file: /tmp/html-report.html
                    - how: display
                      verbosity: 2
    
      Conditional adjustments:
    
      plan:
          context:
              distro: fedora
          import:
              name: /plans/remote
              adjust-plans:
                - report+:
                      how: junit
                      file: /tmp/fedora-results.xml
                  when: distro == fedora
                  because: Fedora supports JUnit reporting

```

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/docs/releases/pending/4379.fmf
+++ b/docs/releases/pending/4379.fmf
@@ -1,17 +1,5 @@
 description: |
-    Added ``adjust-plans`` support for plan imports, enabling flexible step
+    Added ``adjust-plans`` support for :ref:`/spec/plans/import`, enabling flexible step
     inheritance and modification of imported plans using the same powerful
-    rule system as ``adjust-tests``. This allows selective modification and
-    addition of steps when importing remote plans.
-
-    **New Features:**
-
-    * **Step inheritance**: Import remote plans and modify/add steps locally
-    * **Step operations**: Support both append (``+`` suffix) and replace modes for all step types
-    * **Multiple plugins**: Add multiple report plugins with list syntax
-    * **Conditional rules**: Full support for ``when`` and ``because`` clauses
-    * **Context integration**: Works with context inheritance settings
-    * **All step types**: Supports discover, provision, prepare, execute, report, finish, and cleanup steps
-
-    **Schema and validation**: Added ``adjust-plans`` field to plan import schema
-    with full validation and normalization support.
+    rule system as :ref:`/spec/core/adjust` and :ref:`/plugins/discover/fmf`.
+    This allows selective modification and addition of steps when importing remote plans.
--- a/spec/plans/import.fmf
+++ b/spec/plans/import.fmf
@@ -127,10 +127,10 @@ description: |
 
     .. versionadded:: 1.19
 
-    The second way to modify imported plans is via the ``adjust-plans``
-    feature, which allows modification of imported plan steps using
-    the same powerful rule system as :ref:`/spec/core/adjust`. This
-    provides flexible plan inheritance and step customization:
+    The imported plans could also be modified via the ``adjust-plans``
+    key, which allows modification of imported plans using
+    the same powerful rule system as :ref:`/spec/core/adjust` and :ref:`/plugins/discover/fmf`.
+    This provides flexible plan inheritance and customization:
 
     .. code-block:: yaml
 
@@ -144,14 +144,10 @@ description: |
                            how: junit
                            file: /tmp/results.xml
                      - execute:              # Replace execute step
+                           how: tmt
                            timeout: 30m
                        when: arch == x86_64
                        because: x86_64 systems need longer timeout
-
-    The ``adjust-plans`` rules support all step types (discover, provision,
-    prepare, execute, report, finish, cleanup) and follow the same syntax
-    as regular adjust rules with ``when``, ``continue``, and ``because``
-    keys. Use the ``+`` suffix to append steps or no suffix to replace them.
 
     .. versionadded:: 1.67
 
--- a/tests/plan/import/step-inheritance.sh
+++ b/tests/plan/import/step-inheritance.sh
@@ -6,72 +6,43 @@ rlJournalStart
         rlRun "pushd step-inheritance"
     rlPhaseEnd
 
-    rlPhaseStartTest "Test report step append mode"
-        rlRun -s "tmt plan show /test-report-append"
-        # Should show both remote and local report configurations
-        rlAssertGrep "how junit" $rlRun_LOG     # local report
-        rlAssertGrep "/tmp/local-results.xml" $rlRun_LOG
+    rlPhaseStartTest "Test discover step append mode"
+        rlRun -s "tmt plan show /discover-append"
+        # Should show both remote and local discover configurations
+        rlAssertGrep "how shell" $rlRun_LOG     # local discover
+        rlAssertGrep "Additional test from local plan" $rlRun_LOG
 
         # Export and verify inheritance
-        rlRun "tmt plan export /test-report-append > append-export.yaml"
-        rlAssertGrep "how: junit" append-export.yaml
+        rlRun -s "tmt plan export /discover-append"
+        rlAssertGrep "how: shell" $rlRun_LOG
     rlPhaseEnd
 
-    rlPhaseStartTest "Test report step replace mode"
-        rlRun -s "tmt plan show /test-report-replace"
-        # Should show only local report configuration (replacing remote)
-        rlAssertGrep "how html" $rlRun_LOG      # local report
-        rlAssertGrep "/tmp/replacement.html" $rlRun_LOG
+    rlPhaseStartTest "Test discover step replace mode"
+        rlRun -s "tmt plan show /discover-replace"
+        # Should show only local discover configuration (replacing remote)
+        rlAssertGrep "how shell" $rlRun_LOG      # local discover
+        rlAssertGrep "Replacement test from local plan" $rlRun_LOG
 
         # Export and verify inheritance
-        rlRun "tmt plan export /test-report-replace > replace-export.yaml"
-        rlAssertGrep "how: html" replace-export.yaml
-    rlPhaseEnd
-
-    rlPhaseStartTest "Test no step inheritance (control)"
-        rlRun -s "tmt plan show /test-no-inheritance"
-        # Should show only remote plan configuration without local step inheritance
-        rlAssertNotGrep "/tmp/ignored.xml" $rlRun_LOG
-    rlPhaseEnd
-
-    rlPhaseStartTest "Test alternative enum values"
-        rlRun -s "tmt plan show /test-alternative-enum"
-        # Test alternative "append" enum value (should work same as "+")
-        rlAssertGrep "Local finish with append enum" $rlRun_LOG
+        rlRun -s "tmt plan export /discover-replace"
+        rlAssertGrep "how: shell" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "Test multiple adjust fields"
-        rlRun -s "tmt plan show /test-multiple-adjust"
+        rlRun -s "tmt plan show /multiple-adjust"
         # Should show local finish (append) and report (replace)
         rlAssertGrep "Local finish" $rlRun_LOG
         rlAssertGrep "/tmp/local-finish.xml" $rlRun_LOG
     rlPhaseEnd
 
-    rlPhaseStartTest "Test multiple report plugins with list syntax"
-        rlRun -s "tmt plan show -vv /test-multiple-report-plugins"
-        # Should show multiple local report configurations
-        rlAssertGrep "how junit" $rlRun_LOG     # local junit report
-        rlAssertGrep "/tmp/local-results.xml" $rlRun_LOG
-        rlAssertGrep "how html" $rlRun_LOG      # local html report
-        rlAssertGrep "/tmp/local-results.html" $rlRun_LOG
-        rlAssertGrep "how display" $rlRun_LOG   # local display report
-
-        # Export and verify inheritance
-        rlRun "tmt plan export /test-multiple-report-plugins > multiple-export.yaml"
-        rlAssertGrep "how: junit" multiple-export.yaml
-        rlAssertGrep "how: html" multiple-export.yaml
-        rlAssertGrep "how: display" multiple-export.yaml
-    rlPhaseEnd
-
     rlPhaseStartTest "Test conditional adjust-plans with when/because"
-        rlRun -s "tmt plan show /test-conditional-adjust"
+        rlRun -s "tmt plan show -vv /conditional-adjust"
         # Should show conditional adjustments applied
         rlAssertGrep "/tmp/conditional-results.xml" $rlRun_LOG
-        rlAssertGrep "timeout.*30m" $rlRun_LOG
 
         # Export and check for when/because clauses (in original plan data)
-        rlRun "tmt plan export /test-conditional-adjust > conditional-export.yaml"
-        rlAssertGrep "how: junit" conditional-export.yaml
+        rlRun -s "tmt plan export /conditional-adjust"
+        rlAssertGrep "how: junit" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "Schema validation"
@@ -80,8 +51,6 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartCleanup
-        rlRun "rm -f append-export.yaml replace-export.yaml multiple-export.yaml"
-        rlRun "rm -f conditional-export.yaml conditional-patterns-export.yaml"
         rlRun "popd"
     rlPhaseEnd
 rlJournalEnd
--- a/tests/plan/import/step-inheritance/main.fmf
+++ b/tests/plan/import/step-inheritance/main.fmf
@@ -1,67 +1,36 @@
-# Test step inheritance with adjust-plans fields
+# Common configuration defined at the top level (root)
+plan:
+    import:
+        url: https://github.com/teemtee/tests
+        name: /plans/remote
+        # Default for most tests in this group
+        inherit-context: false
+        inherit-environment: false
 
-# Test append mode for report step
-/test-report-append:
-    summary: Test report step inheritance in append mode
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
+# Child plans inherit the configuration above
+/discover-append:
+    summary: Test discover step inheritance in append mode
+    # Use '+' to merge into the inherited 'plan' -> 'import' structure
+    plan+:
+        import+:
             adjust-plans:
-              - report+:
-                    how: junit
-                    file: /tmp/local-results.xml
-
-# Test replace mode for report step
-/test-report-replace:
-    summary: Test report step inheritance in replace mode
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
-            adjust-plans:
-              - report:
-                    how: html
-                    file: /tmp/replacement.html
-
-# Test no step inheritance (control)
-/test-no-inheritance:
-    summary: Test plan with no step inheritance
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
-            # No adjust-plans fields, so only remote plan steps are used
-
-# Test alternative enum values
-/test-alternative-enum:
-    summary: Test alternative enum values (append/replace)
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
-            adjust-plans:
-              - finish+:  # append mode
+              - discover+:
                     how: shell
-                    script: echo "Local finish with append enum"
+                    test: echo "Additional test from local plan"
 
-# Test multiple adjust fields
-/test-multiple-adjust:
+/discover-replace:
+    summary: Test discover step inheritance in replace mode
+    plan+:
+        import+:
+            adjust-plans:
+              - discover:
+                    how: shell
+                    test: echo "Replacement test from local plan"
+
+/multiple-adjust:
     summary: Test multiple adjust fields
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
+    plan+:
+        import+:
             adjust-plans:
               - finish+:
                     name: local-additional-finish
@@ -71,48 +40,18 @@
                     how: junit
                     file: /tmp/local-finish.xml
 
-# Test multiple report plugins with list syntax
-/test-multiple-report-plugins:
-    summary: Test multiple report plugins using adjust-plans
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
-            adjust-plans:
-              - report+:
-                  - how: junit
-                    file: /tmp/local-results.xml
-                  - how: html
-                    file: /tmp/local-results.html
-                  - how: display
-                    verbosity: 2
-
-# Test conditional adjust-plans with when/because like adjust-tests
-/test-conditional-adjust:
+/conditional-adjust:
     summary: Test conditional adjust-plans with when/because clauses
     context:
         arch: x86_64
         distro: fedora
-    plan:
-        import:
-            url: https://github.com/teemtee/tests
-            name: /plans/remote
-            inherit-context: false
-            inherit-environment: false
+    plan+:
+        import+:
+            # Override the inherited setting for this specific plan
+            inherit-context: true
             adjust-plans:
               - report+:
                     how: junit
                     file: /tmp/conditional-results.xml
                 when: distro == fedora
                 because: Fedora supports JUnit reporting
-              - execute:
-                    timeout: 30m
-                when: arch == x86_64
-                because: x86_64 systems need longer timeout
-              - finish+:
-                    how: shell
-                    script: echo "Cleanup for slow systems"
-                when: arch == i386
-                because: Slow system cleanup needed
--- a/tmt/base.py
+++ b/tmt/base.py
@@ -2088,9 +2088,7 @@ class RemotePlanReference(
         )
         reference.inherit_context = bool(raw.get('inherit-context', True))
         reference.inherit_environment = bool(raw.get('inherit-environment', True))
-        reference.adjust_plans = tmt.utils.normalize_adjust(
-            'adjust-plans', raw.get('adjust-plans'), tmt.log.Logger.get_bootstrap_logger()
-        )
+        reference.adjust_plans = cast(list[_RawAdjustRule], raw.get('adjust-plans', []))
 
         return reference
 
@@ -3240,16 +3238,6 @@ class Plan(
             f"Looking for plans in '{tree.root}' matching '{reference.name_pattern}'", level=3
         )
 
-        # Apply adjust-plans rules if specified, similar to adjust-tests
-        if reference.adjust_plans:
-            tmt_tree = tmt.Tree(
-                logger=self._logger,
-                path=tree.root,
-                fmf_context=self.fmf_context,
-                additional_rules=reference.adjust_plans,
-            )
-            tree = tmt_tree.tree
-
         for node in tree.prune(keys=['execute']):
             if reference.name_pattern.match(node.name) is not None:
                 yield node
@@ -3368,6 +3356,7 @@ class Plan(
                 node.name, node.data.get('context', {}), self._logger
             )
 
+            # For final context inheritance, respect inherit_context setting
             if reference.inherit_context:
                 alteration_fmf_context = FmfContext(
                     {
@@ -3382,10 +3371,11 @@ class Plan(
                 )
 
             # Adjust the imported tree, to let any `adjust` rules defined in it take
-            # action.
+            # action, including the adjust-plans rules.
             node.adjust(
                 fmf.context.Context(**alteration_fmf_context),
                 case_sensitive=False,
+                additional_rules=reference.adjust_plans,
             )
 
             # If the local plan is disabled, disable the imported plan as well


tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,62 @@
...
+
+/inherit-report-only:
+    summary: Local plan that inherits report step only
+    report:
+        how: junit
+        file: /tmp/results.xml
+    plan:
+        import:
+            url: file://${PWD}
+            name: /remote-plan-no-report
```
LecrisUT: If only it were that easy. We have https://github.com/teemtee/tests to place the tests that we can import

tmt/base.py
```diff
@@ -3517,6 +3532,37 @@ def reshape(self, tests: list['tmt.steps.discover.TestOrigin']) -> bool:
...
+            # If remote plan has no step configuration, use the local one
+            node.data[step] = local_step_data
+        else:
+            # If remote plan has step configuration, append local configuration
+            if not isinstance(remote_step_data, list):
+                remote_step_data = [remote_step_data]
+
+            # Merge by appending local step configurations to remote ones
+            merged_steps = remote_step_data + local_step_data
+            node.data[step] = merged_steps
```
LecrisUT: It is ambiguous if the user wants to overwrite or append the data if it is a list. If the interface allowed for `{step}+`-like interface, then the intent is more clear (no `+`, always overwrite). I think we do that in the `adjust-tests` under discover, we could try to mimic the interface in there.

tmt/base.py
```diff
@@ -1979,12 +1981,16 @@ class RemotePlanReference(
...
+        'inherit-report',
+        'inherit-finish',
     ]
 
     importing: RemotePlanReferenceImporting = RemotePlanReferenceImporting.REPLACE
     scope: RemotePlanReferenceImportScope = RemotePlanReferenceImportScope.FIRST_PLAN_ONLY
     inherit_context: bool = True
     inherit_environment: bool = True
+    inherit_report: bool = False
+    inherit_finish: bool = False
```
LecrisUT: Do we need explicit `inherit_*` here? Could it instead be either `report` directly or `adjust_*` and assume it is always an adjust rule.

tmt/base.py
```diff
@@ -3325,6 +3333,13 @@ def _convert_node(node: fmf.Tree) -> 'Plan':
                     {**imported_fmf_context, **self._noninheritable_fmf_context}
                 )
 
+            # Merge step configurations from the importing plan if inherit flags are set
+            if reference.inherit_report and self.node.get('report'):
+                self._merge_step_configurations(node, 'report')
+
+            if reference.inherit_finish and self.node.get('finish'):
+                self._merge_step_configurations(node, 'finish')
+
```
LecrisUT: Why not generalize to everything right away?

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/6c408c83e97dda6b3c62c0ff89da9bef13d55e27 :)

tmt/base.py
happz: I believe this is the crucial issue: what did user mean when they asked to "inherit" importing plan's `report` step? Is it supposed to be merged with the imported plan's `report` phases, or is it supposed to truly overwrite the imported plan's `report` step?

While I appreciate the enthusiasm, this topic deserves a discussion first to find out and agree upon what **exactly** should tmt do, why, and how.

therazix:
We need to discuss this first and decide on the overall direction before diving into implementation.

mcasquer:
From the hacking session, we need to decide what to do with this PR, either merge it or overwrite it

mcasquer:
Lili is also going to take a look to the `adjust` test as it should be similar

tmt/base.py
skycastlelily: I take reference from adjust-test, both overwrite and merge are supported on users' desire^^ :
https://github.com/teemtee/tmt/pull/4379/commits/640d577c6c0a89e405b4f7f4748a6ced3caccab6

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4379/commits/640d577c6c0a89e405b4f7f4748a6ced3caccab6 :)

tmt/schemas/plan.yaml
```diff
@@ -172,6 +186,30 @@ properties:
           inherit-environment:
             type: boolean
 
+          adjust-discover:
+            type: string
+            enum: ["+", "append", "=", "replace"]
```
LecrisUT: Right now we use
https://github.com/teemtee/tmt/blob/e35ba6dd9679ee8661b32424b1f20490a367ce91/tmt/schemas/discover/fmf.yaml#L116-L117
I don't think we have a good way to do the schema validation for fmf patterns, so I think it would be better to skip it like in the part above. The explicit `Step inheritance operators` should also not be needed.

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,75 @@
...
+        how: shell
+        script: echo "Local finish"
+    plan:
+        import:
+            url: https://github.com/teemtee/tests
+            name: /plans/remote
+            inherit-context: false
+            inherit-environment: false
+            adjust-prepare: "+"
+            adjust-finish: "="
```
LecrisUT: The interface in `adjust-tests` is like
```yaml
/plan:
  discover:
    how: fmf
    adjust-tests:
      - check+:
          - how: avc
/test:
  test: /usr/bin/true
  check: dmesg
```
So translating the interface to the plan.import feature for the case above it should look like:
```yaml
plan:
  import:
    name: /plans/remote
    adjust-prepare:
      prepare+:
        - how: shell
          script: echo "Local prepare"
    adjust-finish:
      finish:
        how: shell
        script: echo "Local finish"
```
I am realizing that the `adjust-*` is excessive in this case, so it should probably instead be merged into a single key like:
```yaml
plan:
  import:
    name: /plans/remote
    adjust-plans:
      prepare+:
        - how: shell
          script: echo "Local prepare"
      finish:
        how: shell
        script: echo "Local finish"
```

tmt/base.py
```diff
@@ -3546,6 +3584,141 @@ def reshape(self, tests: list['tmt.steps.discover.TestOrigin']) -> bool:
...
+                )
+
+                if not already_handled:
+                    local_step_data = self.node.data[plain_step_key]
+                    self._apply_step_inheritance(node, base_step_name, local_step_data, '+')
+                    self.debug(f"Applied {plain_step_key} with default append mode.", level=3)
+
+    def _apply_step_inheritance(
+        self, node: fmf.Tree, step_name: str, local_step_data: Any, mode: str
+    ) -> None:
```
LecrisUT: Afaict, we do not need to re-invent the wheel here. See reference for how `adjust-tests` is done here
https://github.com/teemtee/tmt/blob/e35ba6dd9679ee8661b32424b1f20490a367ce91/tmt/steps/discover/fmf.py#L821-L826
and you could do it somewhere in
https://github.com/teemtee/tmt/blob/e35ba6dd9679ee8661b32424b1f20490a367ce91/tmt/base.py#L3304-L3310
around the `self._resolve_import_from_git` if you want to use the same logic.

Even if we do not account for the parent `inherit_context` it would be a good enough improvement, but bonus points if you have some ideas on how to handle that.

tmt/schemas/plan.yaml
skycastlelily: Sorry for the late response, was on year-end party,and addressed: https://github.com/teemtee/tmt/pull/4379/commits/cc9d55784a30d043b0bee366f18d1aaa10149ba3 :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4379/commits/cc9d55784a30d043b0bee366f18d1aaa10149ba3 :)

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/cc9d55784a30d043b0bee366f18d1aaa10149ba3 :)

tmt/base.py
```diff
@@ -3354,6 +3368,16 @@ def _convert_node(node: fmf.Tree) -> 'Plan':
                     {**imported_fmf_context, **self._noninheritable_fmf_context}
                 )
 
+            # Step inheritance is now handled by adjust-plans rules via additional_rules
+            # in the Tree constructor, similar to how adjust-tests works
+
+            # For backward compatibility: if no adjust-plans is specified but local steps exist,
+            # merge them as before
```
LecrisUT: This has not been a feature before, so you do not need this backwards compatibility.

tmt/schemas/plan.yaml
```diff
@@ -108,6 +108,7 @@ properties:
   # for details.
   report: true
 
+
```
LecrisUT: Leftover change I suppose?

tests/plan/import/step-inheritance.sh
```diff
@@ -0,0 +1,63 @@
...
+        rlAssertGrep "Local finish" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartTest "Schema validation"
+        # Ensure all adjust-* fields are properly recognized
+        rlRun -s "tmt plan show /test-report-append"
+        rlRun -s "tmt plan show /test-report-replace"
+        rlRun -s "tmt plan show /test-alternative-enum"
+        rlRun -s "tmt plan show /test-multiple-adjust"
+        # If any fail, schema validation has issues
```
LecrisUT: Hmm, does `tmt lint` work here? Or at the very least `rlAssertNotGrep` for the schema error
https://github.com/teemtee/tmt/blob/10953ca85e8f0da4a15a8d623fe7ad370d7b9ef8/tests/provision/beaker/schema.sh#L13
(just realized how incredibly fragile that test is 😬)

tmt/base.py
```diff
@@ -3223,6 +3227,16 @@ def _resolve_import_to_nodes(
             f"Looking for plans in '{tree.root}' matching '{reference.name_pattern}'", level=3
         )
 
+        # Apply adjust-plans rules if specified, similar to adjust-tests
+        if hasattr(reference, 'adjust_plans') and reference.adjust_plans is not None:
```
LecrisUT:
```suggestion
        if reference.adjust_plans:
```
afaict from `inherit_context` we do not need to `hasattr` this.

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4379/commits/b1967601184a9894b693d9f4840b7b143250af2c :)

tmt/schemas/plan.yaml
skycastlelily: Yeah, and addressed: https://github.com/teemtee/tmt/pull/4379/commits/b1967601184a9894b693d9f4840b7b143250af2c :)

tests/plan/import/step-inheritance.sh
skycastlelily: Yeah, and addressed: https://github.com/teemtee/tmt/pull/4379/commits/b1967601184a9894b693d9f4840b7b143250af2c :)

tmt/base.py
skycastlelily: Yeah, and addressed: https://github.com/teemtee/tmt/pull/4379/commits/b1967601184a9894b693d9f4840b7b143250af2c :)

skycastlelily:
> Awesome work on this looks neat. Just a few minor comments

Thanks for your review,you are supposed to be enjoying your holidays :christmas_tree:  :star: ,btw:)

psss:
Summary from the hacking session: @skycastlelily to update the description with the proposed functionality. We'll discuss the approach next week.

psss:
Suggested in #4496 to be merged before the module refactoring.
Proposing to finish before the housekeeping sprint.

tmt/base.py
```diff
@@ -2081,6 +2084,7 @@ def from_spec(cls, raw: _RemotePlanReference) -> 'RemotePlanReference':  # type:
         )
         reference.inherit_context = bool(raw.get('inherit-context', True))
         reference.inherit_environment = bool(raw.get('inherit-environment', True))
+        reference.adjust_plans = raw.get('adjust-plans', None)
```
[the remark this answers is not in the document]
skycastlelily: It could do better: https://github.com/teemtee/tmt/commit/7c86bea978e640795f14299dbb464edb6eb7c549

tmt/base.py
```diff
@@ -1991,12 +1992,14 @@ class RemotePlanReference(
...
         'inherit-context',
         'inherit-environment',
+        'adjust-plans',
     ]
 
     importing: RemotePlanReferenceImporting = RemotePlanReferenceImporting.REPLACE
     scope: RemotePlanReferenceImportScope = RemotePlanReferenceImportScope.FIRST_PLAN_ONLY
     inherit_context: bool = True
     inherit_environment: bool = True
+    adjust_plans: Optional[Any] = None
```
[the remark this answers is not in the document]
skycastlelily: Addressed in: https://github.com/teemtee/tmt/pull/4379/commits/7c86bea978e640795f14299dbb464edb6eb7c549

tmt/base.py
```diff
@@ -1948,6 +1948,7 @@ class _RemotePlanReference(_RawFmfId):
     scope: Optional[str]
     inherit_context: Optional[bool]
     inherit_environment: Optional[bool]
+    adjust_plans: Optional[Any]
```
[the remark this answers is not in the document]
skycastlelily: Addressed in: https://github.com/teemtee/tmt/commit/7c86bea978e640795f14299dbb464edb6eb7c549

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,72 @@
...
+        import:
+            url: https://github.com/teemtee/tests
+            name: /plans/remote
+            inherit-context: false
+            inherit-environment: false
+            adjust-plans:
+                finish+:
+                    name: local-additional-finish
+                    how: shell
+                    script: echo "Local prepare"
```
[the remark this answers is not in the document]
skycastlelily: Addressed in: https://github.com/teemtee/tmt/commit/7c86bea978e640795f14299dbb464edb6eb7c549

tmt/base.py
```diff
@@ -1948,6 +1948,7 @@ class _RemotePlanReference(_RawFmfId):
     scope: Optional[str]
     inherit_context: Optional[bool]
     inherit_environment: Optional[bool]
+    adjust_plans: Optional[list[Any]]
```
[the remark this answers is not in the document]
skycastlelily: https://github.com/teemtee/tmt/pull/4379/commits/3fbef7f194f219ad043b74dabd4f75efdbf75964

tmt/base.py
```diff
@@ -1991,12 +1992,17 @@ class RemotePlanReference(
...
     ]
 
     importing: RemotePlanReferenceImporting = RemotePlanReferenceImporting.REPLACE
     scope: RemotePlanReferenceImportScope = RemotePlanReferenceImportScope.FIRST_PLAN_ONLY
     inherit_context: bool = True
     inherit_environment: bool = True
+    adjust_plans: Optional[list[Any]] = field(
+        default_factory=list,
+        normalize=tmt.utils.normalize_adjust,
+    )
```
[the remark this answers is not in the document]
happz: Seems legit to me. When can the field be `None`?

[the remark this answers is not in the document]
skycastlelily: hmmm,  the field won't be none, but the field uses tmt.utils.normalize_adjust to process its input. normalize_adjust explicitly type-hinted to return Optional[...]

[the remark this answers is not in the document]
happz: 1. it would be nice to see the explanation as a comment here, so we would all know why the advice was rejected.
2. true, `normalize_adjust()` returns `Optional`, but that seems like a bug to me. I wouldn't expect it to return `None`, instead an empty list should be returned from this kind of functions, and that's exactly what `normalize_adjust()` does. I'd fix that annotation, it's unnecessary broad.

psss:
Suggestion from the hacking session:
* Implement `adjust-plans` in the very same way as [adjust-tests](https://tmt.readthedocs.io/en/stable/plugins/discover.html#adjust-tests)
* Would give more flexibility to the user and would be consistent
* Extend the [Import Plans](https://tmt.readthedocs.io/en/stable/spec/plans.html#import-plans) documentation with the new examples

tmt/base.py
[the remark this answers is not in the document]
skycastlelily: should we have the fix involved in this mr,or?^^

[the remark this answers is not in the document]
happz: Seems related and small, I'd include it.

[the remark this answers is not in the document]
skycastlelily: sure, and done https://github.com/teemtee/tmt/pull/4379/commits/bc6bfa7e4a2a82965991fcc0c7756f92d6a193fc ^^

tmt/base.py
```diff
@@ -1991,12 +1992,19 @@ class RemotePlanReference(
...
         'inherit-environment',
+        'adjust-plans',
     ]
 
     importing: RemotePlanReferenceImporting = RemotePlanReferenceImporting.REPLACE
     scope: RemotePlanReferenceImportScope = RemotePlanReferenceImportScope.FIRST_PLAN_ONLY
     inherit_context: bool = True
     inherit_environment: bool = True
+    # Note: normalize_adjust always returns a list, never None,
+    # so Optional is not needed here despite the function's type hint
```
happz: Type hint is now longer "incorrect".

tmt/base.py
```diff
@@ -3361,9 +3382,16 @@ def _convert_node(node: fmf.Tree) -> 'Plan':
                     {**imported_fmf_context, **self._noninheritable_fmf_context}
                 )
 
+            # Step inheritance is now handled by adjust-plans rules via additional_rules
+            # in the Tree constructor, similar to how adjust-tests works
+
             # Adjust the imported tree, to let any `adjust` rules defined in it take
-            # action.
-            node.adjust(fmf.context.Context(**alteration_fmf_context), case_sensitive=False)
+            # action, including the adjust-plans rules.
```
happz, line 3374: Isn't this applying `adjust-plans` twice? It's applied once in `_resolve_import_to_nodes`, where it modifies the imported fmf tree before we scan it for plan-like nodes. Then we have our plan-like nodes, and we start turning them into actual plans, and `node.adjust` is called here once again. It seems like the `adjust-plans` - maybe even the whole `adjust` - runs twice on plan-like nodes.

Is it necessary to inject `adjust_plans` into the imported tree in `_resolve_import_to_nodes`?

happz:
What's missing: documentation, i.e. spec update in `spec/plans/import.fmf`, and a release note.

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4379/commits/288ddf34c570139eef26b25a9ae16782edf7db80 ^^

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4379/commits/288ddf34c570139eef26b25a9ae16782edf7db80 ^^

skycastlelily:
> What's missing: documentation, i.e. spec update in `spec/plans/import.fmf`, and a release note.

Addressed: https://github.com/teemtee/tmt/pull/4379/commits/288ddf34c570139eef26b25a9ae16782edf7db80 ^^

tmt/base.py
happz: No, it's not addressed.

First, I'm genuinely asking whether "is it necessary to inject adjust_plans into the imported tree in _resolve_import_to_nodes?" If I could chose, I would drop that bit, because it's using completely different context than what I'd expect, `self.fmf_context` instead of `alteration_fmf_context` - `reference.inherit_context` is ignored by that tree adjust in `_resolve_import_to_nodes`.

Second, the code as it is now still runs `adjust` twice, once in `_resolve_import_to_nodes`, once here. What I don't understand is why - I know why it's happening here, but I don't understand why it's needed in `_resolve_import_to_nodes`.

docs/releases/pending/4379.fmf
```diff
@@ -0,0 +1,17 @@
+description: |
+    Added ``adjust-plans`` support for plan imports, enabling flexible step
+    inheritance and modification of imported plans using the same powerful
+    rule system as ``adjust-tests``. This allows selective modification and
```
happz: "... as `adjust` and `adjust-tests`."

Where `adjust` and `adjust-tests` could point to "Adjust" docs and `discover/fmf` plugin, respectively.

docs/releases/pending/4379.fmf
```diff
@@ -0,0 +1,17 @@
...
+    addition of steps when importing remote plans.
+
+    **New Features:**
```
happz: I'd drop everything below, seems too much verbose and repetitive for a simple release note.

spec/plans/import.fmf
```diff
@@ -127,6 +127,34 @@ description: |
 
     .. versionadded:: 1.19
 
+    The second way to modify imported plans is via the ``adjust-plans``
+    feature, which allows modification of imported plan steps using
```
happz:
```suggestion
    key, which allows modification of imported plan steps using
```

spec/plans/import.fmf
```diff
@@ -127,6 +127,34 @@ description: |
...
+                   adjust-plans:
+                     - report+:              # Add additional report step
+                           how: junit
+                           file: /tmp/results.xml
+                     - execute:              # Replace execute step
+                           timeout: 30m
+                       when: arch == x86_64
+                       because: x86_64 systems need longer timeout
+
+    The ``adjust-plans`` rules support all step types (discover, provision,
```
happz:
```suggestion
    The ``adjust-plans`` rules support all steps...
```

spec/plans/import.fmf
```diff
@@ -127,6 +127,34 @@ description: |
...
+    prepare, execute, report, finish, cleanup) and follow the same syntax
+    as regular adjust rules with ``when``, ``continue``, and ``because``
```
happz:
```suggestion
    as regular ``adjust`` rules with ``when``, ``continue``, and ``because``
```

docs/releases/pending/4379.fmf
```diff
@@ -0,0 +1,17 @@
...
+
+    * **Step inheritance**: Import remote plans and modify/add steps locally
+    * **Step operations**: Support both append (``+`` suffix) and replace modes for all step types
+    * **Multiple plugins**: Add multiple report plugins with list syntax
+    * **Conditional rules**: Full support for ``when`` and ``because`` clauses
+    * **Context integration**: Works with context inheritance settings
+    * **All step types**: Supports discover, provision, prepare, execute, report, finish, and cleanup steps
+
+    **Schema and validation**: Added ``adjust-plans`` field to plan import schema
+    with full validation and normalization support.
```
LecrisUT: This does not fit in well with the format of the release notes

spec/plans/import.fmf
```diff
@@ -127,6 +127,34 @@ description: |
...
+    keys. Use the ``+`` suffix to append steps or no suffix to replace them.
```
happz:
```suggestion
    keys.
```

They will have links to full `adjust` docs above, which would cover way more topics, including removal of items.

spec/plans/import.fmf
```diff
@@ -127,6 +127,34 @@ description: |
 
     .. versionadded:: 1.19
 
+    The second way to modify imported plans is via the ``adjust-plans``
+    feature, which allows modification of imported plan steps using
+    the same powerful rule system as :ref:`/spec/core/adjust`. This
+    provides flexible plan inheritance and step customization:
```
LecrisUT: This is getting quite cramped with features. Would be better to split it, but maybe for a different PR.

docs/releases/pending/4379.fmf
```diff
@@ -0,0 +1,17 @@
+description: |
+    Added ``adjust-plans`` support for plan imports, enabling flexible step
+    inheritance and modification of imported plans using the same powerful
+    rule system as ``adjust-tests``. This allows selective modification and
+    addition of steps when importing remote plans.
```
LecrisUT: Can you `:ref:` to `/spec/plans/import` somewhere? (note that it will be `:tmt:story:` after https://github.com/teemtee/tmt/pull/4288)

`adjust-tests` should already have a `ref`-able object also?

spec/plans/import.fmf
```diff
@@ -127,6 +127,34 @@ description: |
...
+           plan:
+               import:
+                   url: https://github.com/teemtee/tests
+                   name: /plans/remote
+                   adjust-plans:
+                     - report+:              # Add additional report step
+                           how: junit
+                           file: /tmp/results.xml
+                     - execute:              # Replace execute step
+                           timeout: 30m
```
LecrisUT, line 148:
```suggestion
                     - execute:              # Replace execute step
                           how: tmt
                           timeout: 30m
```

spec/plans/import.fmf
LecrisUT: It is actually independent of the steps. Stuff like `ansible` key should also be supported. Better way to visualize it is that it follows the same rules as a normal fmf file adjustment, i.e. as if you have inherited the plan and you added the contents in there in the child node like
```yaml
summary: imported plan
discover:
  how: fmf

# The contents of `adjust-plan` behave as if you added the content like this
/imported_plan:
  adjust+:
    - prepare+:
        - how: shell
          script: echo hello
```

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,118 @@
...
+
+# Test append mode for report step
+/test-report-append:
+    summary: Test report step inheritance in append mode
+    plan:
+        import:
+            url: https://github.com/teemtee/tests
+            name: /plans/remote
+            inherit-context: false
+            inherit-environment: false
```
LecrisUT: Please use inheritance to de-duplicate this part in the other plans

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,118 @@
...
+            adjust-plans:
+              - report+:
+                    how: junit
+                    file: /tmp/local-results.xml
```
LecrisUT: report is not a good example for this because the original plan does not have a report (well it has the built-in one, but that's a different test case). Try to use `discover` instead.

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,118 @@
...
+    plan:
+        import:
+            url: https://github.com/teemtee/tests
+            name: /plans/remote
+            inherit-context: false
+            inherit-environment: false
+            adjust-plans:
+              - report:
+                    how: html
+                    file: /tmp/replacement.html
```
LecrisUT: Same, maybe use `discover`

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,118 @@
...
+# Test no step inheritance (control)
+/test-no-inheritance:
+    summary: Test plan with no step inheritance
+    plan:
+        import:
+            url: https://github.com/teemtee/tests
+            name: /plans/remote
+            inherit-context: false
+            inherit-environment: false
+            # No adjust-plans fields, so only remote plan steps are used
```
LecrisUT: Hmm, not sure if this should be tested. Not sure what failure it's meant to pick up.

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,118 @@
...
+    plan:
+        import:
+            url: https://github.com/teemtee/tests
+            name: /plans/remote
+            inherit-context: false
+            inherit-environment: false
+            adjust-plans:
+              - finish+:  # append mode
+                    how: shell
+                    script: echo "Local finish with append enum"
```
LecrisUT: Uhm, this is leftover from last approach?

tests/plan/import/step-inheritance/main.fmf
```diff
@@ -0,0 +1,118 @@
...
+            inherit-context: false
+            inherit-environment: false
+            adjust-plans:
+              - report+:
+                  - how: junit
+                    file: /tmp/local-results.xml
+                  - how: html
+                    file: /tmp/local-results.html
+                  - how: display
+                    verbosity: 2
```
LecrisUT: Don't think this one is picking up on anything special. It might pick up on issues like
```yaml
discover:
  how: fmf
  test: /original

/imported_plan:
  discover+:
    - how: fmf
      test: /something
    - how: fmf
      test: /other
```
but not sure if this is the job of `plan.import` to test or fmf inheritance in general. At the very least change it to `discover` or equivalent.

tests/plan/import/step-inheritance.sh
```diff
@@ -0,0 +1,87 @@
...
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test report step append mode"
+        rlRun -s "tmt plan show /test-report-append"
+        # Should show both remote and local report configurations
+        rlAssertGrep "how junit" $rlRun_LOG     # local report
+        rlAssertGrep "/tmp/local-results.xml" $rlRun_LOG
+
+        # Export and verify inheritance
+        rlRun "tmt plan export /test-report-append > append-export.yaml"
```
LecrisUT:
```suggestion
        rlRun -s "tmt plan export /test-report-append"
```
Use the `$rlRun_LOG` to check directly.

tmt/base.py
```diff
@@ -3363,7 +3383,10 @@ def _convert_node(node: fmf.Tree) -> 'Plan':
 
             # Adjust the imported tree, to let any `adjust` rules defined in it take
             # action.
-            node.adjust(fmf.context.Context(**alteration_fmf_context), case_sensitive=False)
+            node.adjust(
+                fmf.context.Context(**alteration_fmf_context),
+                case_sensitive=False,
+            )
```
LecrisUT, line 3379: Refactor leftover

tmt/base.py
```diff
@@ -2081,6 +2088,9 @@ def from_spec(cls, raw: _RemotePlanReference) -> 'RemotePlanReference':  # type:
         )
         reference.inherit_context = bool(raw.get('inherit-context', True))
         reference.inherit_environment = bool(raw.get('inherit-environment', True))
+        reference.adjust_plans = tmt.utils.normalize_adjust(
+            'adjust-plans', raw.get('adjust-plans'), tmt.log.Logger.get_bootstrap_logger()
+        )
```
LecrisUT: I believe we do not do the normalization in here. Can someone double-check?

tmt/base.py
skycastlelily: should really be addressed now :https://github.com/teemtee/tmt/pull/4379/commits/230461437c28d3d282fe43120d6631a2891c46ed :sweat_smile:

docs/releases/pending/4379.fmf
skycastlelily: Addressed:https://github.com/teemtee/tmt/pull/4379/commits/230461437c28d3d282fe43120d6631a2891c46ed ^^

docs/releases/pending/4379.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/pull/4379/commits/230461437c28d3d282fe43120d6631a2891c46ed ^^

spec/plans/import.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed ^^

tmt/base.py
```diff
@@ -3361,9 +3370,16 @@ def _convert_node(node: fmf.Tree) -> 'Plan':
                     {**imported_fmf_context, **self._noninheritable_fmf_context}
                 )
 
+            # Step inheritance is now handled by adjust-plans rules via additional_rules
+            # in the Tree constructor, similar to how adjust-tests works
```
happz: This comment is now incorrect, the one below is more fitting.

spec/plans/import.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed ^^

spec/plans/import.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed ^^

spec/plans/import.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed ^^

docs/releases/pending/4379.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

docs/releases/pending/4379.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

spec/plans/import.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

spec/plans/import.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tests/plan/import/step-inheritance/main.fmf
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tmt/base.py
skycastlelily: Yeah, and addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

tmt/base.py
skycastlelily: Addressed: https://github.com/teemtee/tmt/commit/230461437c28d3d282fe43120d6631a2891c46ed :)

psss:
Red job is an irrelevant failure, merging.

PR #4416, opened 2025-12-10
Enforce test duration across reboots and restarts

happz:
Tests that triggered reboots and restarts were given their full duration on every invocation. This behavior is most likely incorrect, a duration should apply to the whole test, no matter how many times it reboots a guest.

Patch adds `Duration` object which poses a strict deadline, derived from the test `duration` key. Unlike a timeout, it is not reset after guest reboot or test restart.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note

### DIFF
--- a/docs/releases/pending/4416.fmf
+++ b/docs/releases/pending/4416.fmf
@@ -1,5 +1,5 @@
 description: |
-  Test :ref:`/spec/tests/duration` is now measured and enforced across
+  Test :tmt:story:`/spec/tests/duration` is now measured and enforced across
   all test restarts and guest reboots. With previous tmt versions,
   a guest reboot or test restart cleared the duration timer, and every
   test invocation was therefore given full budget. This could lead to
--- a/spec/tests/duration.fmf
+++ b/spec/tests/duration.fmf
@@ -22,7 +22,7 @@ description: |
 
        Test duration is measured and enforced for all its invocations
        combined. Every time a test is invoked, after a
-       :ref:`/spec/tests/restart` or guest reboot, it gets only the
+       :tmt:story:`/spec/tests/restart` or guest reboot, it gets only the
        remainder of its ``duration`` limit. Limit is not refreshed by
        these events.
 


vaibhavdaren (approved):

tmt/steps/execute/internal.py
```diff
@@ -428,32 +428,31 @@ def _test_output_logger(
...
 
         # Pick the proper timeout for the test
-        timeout: Optional[int]
+        deadline: Optional[tmt.utils.wait.Deadline]
 
         if self.data.interactive:
-            if test.duration:
-                logger.warning('Ignoring requested duration, not supported in interactive mode.')
+            logger.warning('Ignoring requested duration, not supported in interactive mode.')
+            logger.info('duration limit', 'unlimited')
```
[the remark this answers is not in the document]
happz: `test.duration` has a default value, therefore it is always set to a meaningful value. `if test.duration` should be always true-ish.

tmt/steps/execute/__init__.py
```diff
@@ -386,6 +396,8 @@ def _reset_process(
                 if self.on_interrupt_callback_token is not None:
                     tmt.utils.signals.remove_callback(self.on_interrupt_callback_token)
 
+        timeout = int(deadline.time_left.total_seconds()) if deadline is not None else None
```
psss: If I understand this correctly, this should nicely work with `tmt-reboot` as well, right? I did a small test with the following config:

```yaml
/plan:
    provision:
        how: virtual
    discover:
        how: fmf
    execute:
        how: tmt

/test:
    duration: 1m
    test: |
        sleep 10
        tmt-reboot
```

But the test seems to be running indefinitely. I debugged this line and `timeout` here seems to have the same initial value of 60 seconds in all reboot iterations.

psss: During the individual iterations I see this:

    Deadline: now=183807.127422535 deadline=183867.127422535
    Deadline: now=183807.127422535 deadline=183867.127422535
    Deadline: now=183807.127422535 deadline=183867.127422535
    ...

Seems that `now` is not updated properly?

happz: Interesting, it worked for me. But it's a good thing, I realized how can I test it, I couldn't come up with a test. So I'm going to add one.

happz: I have a question about `duration`, added it to the hacking session tomorrow. I suspect this will need more work to do things properly.

happz:
Blocked on resolving the question of true meaning of Christmas^W`duration`. What exactly does it enforce, and does tmt "paperwork" counts?

tmt/steps/execute/__init__.py
thrix: hmm, ok ...

psss: Seems to be working fine now. One question: When observing the log I've noticed this:

    duration deadline: 60 seconds, at 07:55:09 2026-01-22 UTC
    duration deadline: 60 seconds, at 07:55:09 2026-01-22 UTC
    duration deadline: 38 seconds, at 07:55:09 2026-01-22 UTC
    duration deadline: 9 seconds, at 07:55:09 2026-01-22 UTC

The `time_left` behaves in a bit weird way for the second invocation. Any thoughts here?

tests/execute/duration/test.sh
```diff
@@ -20,20 +20,20 @@ rlJournalStart
                 rlPhaseStartTest "Test provision $PROVISION_HOW, execute $execute_method, $tty, $interactive short tests"
                     rlRun -s "tmt $context --log-topic=command-events run --scratch -vfi $tmp -a provision -h $PROVISION_HOW execute -h $execute_method $interactive test --name short" 0
 
-                    rlRun "grep 'duration \"5\" exceeded' $tmp/log.txt" 1
+                    rlRun "grep 'duration \"29\" exceeded' $tmp/log.txt" 1
```
psss: Hmmm, where does the `29` duration come from? These are short tests which should finish without a problem. Shouldn't it be just checking that there was not duration exceeded?

```suggestion
                    rlAssertNotGrep "Maximum test time .* exceeded' $tmp/log.txt"
```

tmt/steps/execute/__init__.py
```diff
@@ -479,43 +491,69 @@ def _reset_process(
...
+                else:
+                    raise error
 
+            elif output is not None:
                 self.return_code = tmt.utils.ProcessExitCodes.SUCCESS
 
-            except tmt.utils.RunError as error:
-                self.exceptions.append(error)
+            else:
+                raise GeneralError('Command produced no output but raised no exception')
```
psss:
```suggestion
                raise GeneralError('Command produced no output but raised no exception.')
```

tests/execute/duration/test.sh
psss: Or this:

```suggestion
                    rlRun "grep 'duration .* exceeded' $tmp/log.txt" 1
```

tmt/steps/execute/__init__.py
happz: Please, try it now, Logging did not happen within the right context, 8b490cb5 improves that.

tmt/steps/execute/__init__.py
```diff
@@ -479,43 +491,67 @@ def _reset_process(
...
-                    self.logger.warning('Test failed to manage its pidfile.')
+        if self.return_code == tmt.utils.ProcessExitCodes.TIMEOUT:
+            self.logger.debug(f"Test duration '{self.test.duration}' exceeded.")
 
-        self.end_time = timer.end_time_formatted
-        self.real_duration = timer.duration_formatted
+        elif tmt.utils.ProcessExitCodes.is_pidfile(self.return_code):
+            self.logger.warning('Test failed to manage its pidfile.')
 
         return output
```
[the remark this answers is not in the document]
happz: Should be addressed, mypy does not care.

tmt/steps/execute/__init__.py
happz: Addressed in f0a96ec4

tests/execute/duration/test.sh
happz: Addressed in f0a96ec4

tmt/steps/execute/__init__.py
psss: Great! Now works nicely, thanks also for the `silence` fixes, made it much more readable:

    test: /test
        duration deadline: 60 seconds, at 12:16:35 2026-01-22 UTC
        cmd:
            sleep 10
            tmt-reboot
        out: Shared connection to 127.0.0.1 closed.
        00:00:10 /test [1/1]
    test: /test
        duration deadline: 31 seconds, at 12:16:35 2026-01-22 UTC
        cmd:
            sleep 10
            tmt-reboot
        out: Warning: Permanently added '[127.0.0.1]:10091' (ED25519) to the list of known hosts.
        out: Connection to 127.0.0.1 closed.
        00:00:10 /test [1/1]
    test: /test
        duration deadline: -4 seconds, at 12:16:35 2026-01-22 UTC
        00:00:10 errr /test (on default-0) [1/1]
            00:00:00 fail internal/timeout (after-test check)
                Note: Test exceeded maximum duration of 1m

happz:
Some reboot tests seem to be affected, although the reason why remains a mystery to me. Cannot waive the failures, fix needed.

happz:
Blocked on https://github.com/teemtee/tmt/issues/4519 - it's not an actual blocker, but this PR reveals the bug, so it makes sense to fit the bug first, then tests from this PR should pass.

psss:
Seems that this will be soon unblocked as fix for #4519 is already on review

* #4643

Shall we consider adding this (already approved) pull request to the sprint? @bajertom?

PR #4428, opened 2025-12-12
Enable basic functionality tests for secondary images

### LINKED ISSUE #4393: Add a lightweight test coverage for secondary distros
In order to speed up test execution `fedora-42` was removed from the list of fully exercised distros in #4390. It would be good to perform at least basic test coverage for package operation to make sure the essential functionality is not broken. On the review meeting we've agreed to explicitly define a list of secondary distros for which we will not perform the full test coverage (e.g. verify each and every package operation) but still make sure the basic features work. This should be documented in the [Guest Compatibility](https://tmt.readthedocs.io/en/stable/guide.html#guest-compatibility) section.

psss:
Add `fedora-42` back to the list of tested images. Just make sure we don't exercise each and every feature combination again and again. For images listed in `TEST_CONTAINER_IMAGES_SECONDARY` and `TEST_VIRTUAL_IMAGES_SECONDARY` variables we just execute basic package installation check in `/tests/prepare/install`. In tests like `/tests/prepare/ansible` or `/tests/finish/ansible`, which are relatively fast, `fedora-42` is tested again fully.

Fix #4393.

Pull Request Checklist

### DIFF
--- a/tests/images.sh
+++ b/tests/images.sh
@@ -22,11 +22,17 @@ TEST_CONTAINER_IMAGES="${TEST_CONTAINER_IMAGES:-$TEST_IMAGE_PREFIX/alpine:latest
 $TEST_IMAGE_PREFIX/centos/7/upstream:latest
 $TEST_IMAGE_PREFIX/centos/stream9/upstream:latest
 $TEST_IMAGE_PREFIX/centos/stream10/upstream:latest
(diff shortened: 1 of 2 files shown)


psss:
Makes sense, description updated.

vaibhavdaren (approved):

PR #4430, opened 2025-12-12
Improve logging of requires

bajertom:
Pull Request Checklist
* [x] implement the feature
* [x] include a release note

`dependencies_to_test` gathers both requires + recommends and tmt is going to fail when something could not be installed in the require phase already. If all required packages are available to install, but a recommended one is not, then it will fail also, but the message `{package}: required by {test_name}` might be a bit misleading, because it is not required, only recommended. Should the messaging be handled separately or is simply "needed by" enough to cover both requires+recommend in general?

Output:
```
---------------------------✄-------------------------------------
        prepare task #3: requires on default-0
        how: install
        summary: Install required packages
        name: requires
        where: default-0
        package: sed, grep, coreutils and 5 more
        warn: Installation failed, trying again after metadata refresh.
        package: sed, grep, coreutils and 5 more
        fail: Command 'rpm -q --whatprovides sed grep coreutils bash absent wrong nonexistent missing || dnf5 install -y  sed grep coreutils bash absent wrong nonexistent ridiculous' returned 1.
    
        failed packages: 
            missing: required by /tests/FIRST
            absent: required by /tests/FIRST, /tests/SECOND
            wrong: required by /tests/FIRST, /tests/SECOND
            nonexistent: required by /tests/FIRST
    report
        how: display
        summary: 2 pending
    cleanup
---------------------------✄-------------------------------------
```

Test coverage needed?

Fixes [#4302](https://github.com/teemtee/tmt/issues/4302)

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -10,6 +10,19 @@ prepare+:
       - yq
       - make
       - python3-pip
+  - how: shell
+    summary: Make sure we run koji commands with automatic retry to avoid 50x errors
+    script: |
+      mkdir -p ~/.koji
+      if [ -f ~/.koji/config ]; then
+        echo "~/.koji/config was already set:"
+        cat ~/.koji/config
+      else
+        cat <<EOF > ~/.koji/config
+      [koji]
+      anon_retry = true
+      EOF
+      fi
 
 # Use the internal executor
 execute:


happz:
> Pull Request Checklist
> 
> * [x]  implement the feature
> * [ ]  extend the test coverage
> * [ ]  include a release note
> 
> `dependencies_to_test` gathers both requires + recommends and tmt is going to fail when something could not be installed in the require phase already. If all required packages are available to install, but a recommended one is not, then it will fail also, but the message `{package}: required by {test_name}` might be a bit misleading, because it is not required, only recommended. Should the messaging be handled separately or is simply "needed by" enough to cover both requires+recommend in general?

I'd say it should be separated, wth proper messaging, e.g. "failed to install required packages, see the details below, we're done here" vs "failed to install recommended packages, see the details below, and let's move on".

> 
> Output:
> 
> ```
> ---------------------------✄-------------------------------------
>         prepare task #3: requires on default-0
>         how: install
>         summary: Install required packages
>         name: requires
>         where: default-0
>         package: sed, grep, coreutils and 5 more
>         warn: Installation failed, trying again after metadata refresh.
>         package: sed, grep, coreutils and 5 more
>         fail: Command 'rpm -q --whatprovides sed grep coreutils bash absent wrong nonexistent missing || dnf5 install -y  sed grep coreutils bash absent wrong nonexistent ridiculous' returned 1.
>     
>         failed packages: 
>             missing: required by /tests/FIRST
>             absent: required by /tests/FIRST, /tests/SECOND
>             wrong: required by /tests/FIRST, /tests/SECOND
>             nonexistent: required by /tests/FIRST
>     report
>         how: display
>         summary: 2 pending
>     cleanup
> ---------------------------✄-------------------------------------
> ```
> 
> Test coverage needed?

Yes. We should have a test somewhere checking whether a non-existent package failed to install, it should be enough to extend it with an assert or two to check for new messages.

tmt/steps/prepare/__init__.py
```diff
@@ -221,6 +221,61 @@ def summary(self) -> None:
...
+                if error_text:
+                    # dnf, dnf5, yum, apt, apk specific patterns
+                    patterns = [
+                        r'no package provides\s+([^\s\n]+)',
+                        r'Unable to find a match:\s+([^\s\n]+)',
+                        r'No match for argument:\s+([^\s\n]+)',
+                        r'No package\s+([^\s\n]+)\s+available',
+                        r'Unable to locate package\s+([^\s]+)',
+                        r'E:\s+Unable to locate package\s+([^\s]+)',
+                    ]
```
happz: Would it make sense to make these patterns package manager-specific? Let's say I'd add a plugin to support Gentoo package manager, `emerge`: I would need to update this list for `emerge` messages being decoded. Instead, we can add something like `extract_package_name_from_ugly_package_manager_output()` method, we can reach guest from here, and we can ask it to give us packages that somehow failed by calling the method with command output. Which patterns, stdout or stderr, that would all be owned by the package manager implementation. Plus, we don't need to run `apk` patterns for `dnf` output, and vice versa.

Second point: those patterns should be compiled, see various `re.compile()` calls on module level.

tmt/steps/prepare/__init__.py
```diff
@@ -498,7 +553,9 @@ def _record_exception(
         self._save_results(self.results)
 
         if exceptions:
-            # TODO: needs a better message...
+            failed_packages = self._extract_failed_packages(exceptions)
+            if failed_packages:
+                self._show_failed_packages_with_tests(failed_packages)
```
LecrisUT: Why not move this into the PrepareInstall plugin? Here it seems fragile because it is catching any exceptions.

tmt/steps/prepare/__init__.py
LecrisUT: :+1: for having the package managers owning these.

tmt/steps/prepare/__init__.py
```diff
@@ -498,7 +553,9 @@ def _record_exception(
         self._save_results(self.results)
 
         if exceptions:
-            # TODO: needs a better message...
+            failed_packages = self._extract_failed_packages(exceptions)
+            if failed_packages:
```
AthreyVinay: Nitpick: Using the walrus operator makes this more succinct.

tmt/steps/prepare/__init__.py
```diff
@@ -221,6 +221,61 @@ def summary(self) -> None:
...
+
+        # Get test dependencies from discover step
+        dependencies_to_tests = self.plan.discover.dependencies_to_tests
+
+        # Show failed packages with their associated tests
+        self.info('')
+        self.info('failed packages', '', 'red', shift=1)
+
+        for failed_package in failed_packages:
+            matching_tests = []
```
AthreyVinay: You don’t need to initialize matching_tests and then overwrite it.
```rst
matching_tests = dependencies_to_tests.get(failed_package, [])
```

tmt/steps/prepare/__init__.py
```diff
@@ -221,6 +221,61 @@ def summary(self) -> None:
...
+            if failed_package in dependencies_to_tests:
+                matching_tests = dependencies_to_tests[failed_package]
+            sorted_matching_tests = ', '.join(sorted(matching_tests))
+            self.info(failed_package, f'required by {sorted_matching_tests}', color='red', shift=2)
```
AthreyVinay: If no `matching_tests` found, only `required by` gets reported. Maybe add an explicit check?

tmt/steps/prepare/__init__.py
```diff
@@ -221,6 +221,61 @@ def summary(self) -> None:
         preparations = fmf.utils.listed(self.preparations_applied, 'preparation')
         self.info('summary', f'{preparations} applied', 'green', shift=1)
 
+    def _extract_failed_packages(self, exceptions: list[Exception]) -> set[str]:
+        """
+        Extract package names from installation exceptions.
+
+        Returns a list of package names that failed to install.
```
AthreyVinay: Docstring mismatch with the return type.

tmt/steps/prepare/__init__.py
```diff
@@ -221,6 +221,61 @@ def summary(self) -> None:
...
+        """
+
+        failed_packages: set[str] = set()
+        import re
+
+        for exc in exceptions:
+            # Extract failed packages from RunError stderr/stdout
+            if isinstance(exc, tmt.utils.RunError):
+                error_text = ''
```
AthreyVinay: Just this maybe?
```rst
error_text = '\n'.join(
            text for text in (exc.stderr, exc.stdout) if text
        )
```

happz: Or not merging them at all, e.g. with something like the following: an inner function to focus on a single exception, produce a sequence of package names, and then merge these "streams" into a single stream by the magic of `itertools.chain`ining them together. This iterable can be turned into a `set`. Not tested, but this could be a starting point:

```python
def _extract_failed_packages(self, exceptions: list[Exception]) -> set[str]:
    def _extract_from_exception(exception: Exception) -> Iterator[str]:
        if not isinstance(exc, tmt.utils.RunError):
            return

        if exc.stderr:
            yield from guest.package_manager.extract_package_name_from_ugly_package_manager_output(exc.stderr)

        if exc.stdout:
            yield from guest.package_manager.extract_package_name_from_ugly_package_manager_output(exc.stdout)

    return set(
        itertools.chain.from_iterable(
            _extract_from_exception(exc) for exc in exceptions
        )
    )
```

bajertom:
I have addressed comments and suggestions in b46f5c7e187e80867f43bad77ee3d0db98e88c9f
If recommended packages can not be installed, tmt will not raise an exception and therefore I had to use CommandOutput instead and parse the packages from there.

For the "unattributed_packages" - I don't know if that can even happen that packages would not be required/recommended - but maybe the "essential requires" like `beakerlib` and `/usr/bin/flock` could fall into this category?

tmt/steps/prepare/__init__.py
bajertom: Used the iterator solution in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/steps/prepare/__init__.py
bajertom: Fixed in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/steps/prepare/__init__.py
bajertom: Check added in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/steps/prepare/__init__.py
bajertom: Thanks, improved in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/steps/prepare/__init__.py
bajertom: Thanks, improved in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/steps/prepare/__init__.py
bajertom: Moved there in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/steps/prepare/__init__.py
bajertom: I've delegated the package name extraction task to package managers in b46f5c7e187e80867f43bad77ee3d0db98e88c9f

tmt/package_managers/__init__.py
```diff
@@ -270,6 +270,16 @@ def check_presence(self, *installables: Installable) -> dict[Installable, bool]:
 
         raise NotImplementedError
 
+    def extract_package_name_from_package_manager_output(self, output: str) -> Iterator[str]:
```
happz, line 281: Please, decorate with `@abc.abstractmethod` to make clear to linters this is an abstract method.

tmt/package_managers/apt.py
```diff
@@ -79,6 +80,13 @@
 """  # noqa: E501
 
 
+# Compiled regex patterns for APT error messages
+_UNABLE_TO_LOCATE_PATTERN = re.compile(r'Unable to locate package\s+([^\s]+)', re.IGNORECASE)
+_E_UNABLE_TO_LOCATE_PATTERN = re.compile(
+    r'E:\s+Unable to locate package\s+([^\s]+)', re.IGNORECASE
```
happz: Would `(?:E:\s+)?` be possible? Then we would need just one pattern, with the `E:` prefix being optional, the rest seems to be the same.

tmt/package_managers/__init__.py
happz: Or not. After seeing the actual methods, how about moving the patterns into classes, making something like `_PACKAGE_NAME_IN_PM_OUTPUT_PATTERNS = [...]`, and the default implementation of this method can do what all the methods do: iterate over a list of patterns, match, `yield`. They are really repeating the same set of lines, right? The only thing that's different is the actual set of patterns.

tmt/steps/discover/__init__.py
```diff
@@ -280,6 +280,26 @@ def required_tests(self) -> list[TestOrigin]:
...
         return tests
 
+    @property
+    def dependencies_to_tests(self) -> tuple[dict[str, list[str]], dict[str, list[str]]]:
+        """
+        A tuple containing two dictionaries mapping dependencies to tests (required & recommended)
+        """
+
+        required_dependencies_to_tests: dict[str, list[str]] = {}
+        recommended_dependencies_to_tests: dict[str, list[str]] = {}
```
happz: You can use `collections.defaultdict`, `defaultdict(list)`, and skip the `setdefault()` calls below.

tmt/steps/prepare/install.py
```diff
@@ -950,6 +963,85 @@ class PrepareInstall(tmt.steps.prepare.PreparePlugin[PrepareInstallData]):
...
 
+    def _extract_failed_packages_from_outputs(
+        self, outputs: list[tmt.utils.CommandOutput], guest: 'Guest'
+    ) -> set[str]:
+        """
+        Extract package names from installation outputs.
+
+        Returns a set of package names that failed to install.
+        """
+        import itertools
```
happz: It's a standard library, let's import it on top of the module.

tmt/package_managers/__init__.py
bajertom: Moved to classes. Should the compiles themselves be also inside a class?

tmt/package_managers/apt.py
bajertom: I've used the non-capture group like you suggested.

bajertom:
I've addressed the comments in latest commits, now extending the tests remains.

tmt/package_managers/__init__.py
happz: Oh yes, why not?

```python
class ApkEngine(PackageManagerEngine):
    install_command = Command('add')


    _engine_class = ApkEngine

    _PACKAGE_NAME_IN_PM_OUTPUT_PATTERNS = [
        re.compile(r'unable to locate package\s+([^\s]+)', re.IGNORECASE),
        re.compile(r'ERROR:\s+([^\s:]+):\s+No such package', re.IGNORECASE)
    ]
```

skycastlelily:
There are some conflicts.
You may should extend the test,at least  verification that the correct package names are extracted from various error message formats.
And you might want to adding debug logging for regex pattern matching results in the `tmt/steps/prepare/install.py` file.
For the following reasons, 
 ### Benefits of This Logging
1. **Troubleshooting**: When package failures aren't detected, you can see exactly what the package manager output was and whether patterns matched
2. **Pattern Validation**: Helps verify that regex patterns work correctly across different package manager versions
3. **Performance Monitoring**: Shows if certain patterns are never matching (indicating they might be obsolete)
4. **User Support**: Provides detailed information when users report attribution issues
And,This debug logging would only appear when running with verbose logging enabled (`-v` or `--debug`), so it won't clutter normal output but provides valuable insight when needed.

Here is an example change:
```python
def extract_package_name_from_package_manager_output(self, output: str) -> Iterator[str]:
    """Extract package names from package manager output using regex patterns."""
 
    
    self.logger.debug(f"Analyzing package manager output for failed packages:\n{output}")
    
    matched_packages = []
    for i, pattern in enumerate(self._PACKAGE_NAME_IN_PM_OUTPUT_PATTERNS):
        pattern_matches = []
        for match in pattern.finditer(output):
            package_name = match.group(1)
            pattern_matches.append(package_name)
            matched_packages.append(package_name)
            yield package_name
        
        if pattern_matches:
            self.logger.debug(f"Pattern {i} matched packages: {pattern_matches}")
        else:
            self.logger.debug(f"Pattern {i} found no matches")
    
    if not matched_packages:
        self.logger.debug("No failed packages extracted from output")
    else:
        self.logger.debug(f"Total extracted packages: {matched_packages}")
```

skycastlelily (approved):
Other than the nitpick and extending test suggestion, LGTM

tmt/package_managers/__init__.py
```diff
@@ -240,6 +241,10 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
     _engine_class: type[PackageManagerEngineT]
     engine: PackageManagerEngineT
 
+    #: Patterns for extracting failed package names from error output.
+    #: Subclasses override this with their own specific patterns.
+    _PACKAGE_NAME_IN_PM_OUTPUT_PATTERNS: list[re.Pattern[str]] = []
```
```suggestion
    _PACKAGE_NAME_IN_PM_OUTPUT_PATTERNS: ClassVar[list[re.Pattern[str]]] = []
```
But also any other naming that could be used to make it more clear that the pattern is about packages that were not installed?

tmt/steps/discover/__init__.py
```diff
@@ -280,6 +281,26 @@ def required_tests(self) -> list[TestOrigin]:
...
+        recommended_dependencies_to_tests: dict[str, list[str]] = defaultdict(list)
+
+        for test_origin in self.tests(enabled=True):
+            test = test_origin.test
+            test_name = test.name
+            # Collect dependencies separately for required and recommended
+            for dependency in test.require:
+                required_dependencies_to_tests[str(dependency)].append(test_name)
+            for dependency in test.recommend:
+                recommended_dependencies_to_tests[(str(dependency))].append(test_name)
```
LecrisUT: I think this would include the library dependency types also

tmt/steps/prepare/install.py
```diff
@@ -1067,6 +1159,24 @@ def go(
...
+        try:
+            installer.install()
+        except Exception as exc:
+            # Extract and show failed packages if this is a RunError
+            # Convert exception to CommandOutput and use the unified extraction method
+            if isinstance(exc, tmt.utils.RunError):
+                failed_packages = self._extract_failed_packages_from_outputs([exc.output], guest)
+                if failed_packages:
+                    self._show_failed_packages_with_tests(failed_packages)
+            raise
```
LecrisUT, line 1109: We do want plugins to output more relevant results instead of just raising, could do it in this PR IMO.

tmt/package_managers/__init__.py
LecrisUT: Seems resolved.

tmt/steps/prepare/install.py
happz: That would be quite a distraction, for turning `prepare/install` into a plugin that returns results I would recommend a dedicated PR. It took several iterations with `prepare/shell` and `prepare/ansible`, `prepare/install` is even weirder, with its redressing of package managers as "installers" :/

LecrisUT: But in this case the return is already defined as a `PluginOutcome` and it can be done more incrementally, in this case just returning instead of raising. Yes we could do it more cleanly to have the installers control the outcome or such.

tmt/package_managers/__init__.py
bajertom: Renamed to `_FAILED_PACKAGE_INSTALLATION_PATTERNS` and added ClassVar

tmt/steps/discover/__init__.py
bajertom: It does work with uninstallable library:
```
        package: 5 packages requested
            bash
            coreutils
            grep
            library(yum/false-library)
            sed
        
            Required packages failed to install, aborting:
                library(yum/false-library): required by: /tests/FIRST
        fail: Command 'rpm -q --whatprovides coreutils 'library(yum/false-library)' sed grep bash || dnf5 install -y  coreutils 'library(yum/false-library)' sed grep bash' returned 1.
 ```
however only if the library is requested in this form: `library(yum/false-library)`.
Due to recent changes - beaker-tasks repository being dropped from Testing Farm Image Mode - this format is now needed:
```
require:
  - type: library
    url: https://github.com/beakerlib/yum
    name: /false-library
```
However this ends up with `After beakerlib processing, tests may have only simple requirements` error, which is kind of confusing for the user. It originates from Prepare class where it checks whether dependency is 'simple' and throws an exception if it isn't, so it does not have a chance to pass the info to anything else. Not sure how to proceed here..

Comment (happz) [reply tmt/steps/discover/__init__.py]: > It does work with uninstallable library:
> 
> ```
>         package: 5 packages requested
>             bash
>             coreutils
>             grep
>             library(yum/false-library)
>             sed
>         
>             Required packages failed to install, aborting:
>                 library(yum/false-library): required by: /tests/FIRST
>         fail: Command 'rpm -q --whatprovides coreutils 'library(yum/false-library)' sed grep bash || dnf5 install -y  coreutils 'library(yum/false-library)' sed grep bash' returned 1.
> ```
> 
> however only if the library is requested in this form: `library(yum/false-library)`. Due to recent changes - beaker-tasks repository being dropped from Testing Farm Image Mode - this format is now needed:
> 
> ```
> require:
>   - type: library
>     url: https://github.com/beakerlib/yum
>     name: /false-library
> ```
> 
> However this ends up with `After beakerlib processing, tests may have only simple requirements` error, which is kind of confusing for the user. It originates from Prepare class where it checks whether dependency is 'simple' and throws an exception if it isn't, so it does not have a chance to pass the info to anything else. Not sure how to proceed here..

This error message appears because the library was either not recognized, or failed to install or something similar, and remained on the list. First, the libraries are resolved, cloned, installed, etc., and removed from `require` and `recommend` lists; then, this check verifies only the package-like items are left. So something is not working correctly in the processing of libraries on the list of required packages.

@LecrisUT could it be connected to one of your PRs?

Comment (LecrisUT) [reply tmt/steps/discover/__init__.py]: > Due to recent changes - beaker-tasks repository being dropped from Testing Farm Image Mode - this format is now needed:

Would be nice if we can officially drop it. I have not done the impact-check. On the Fedora side there is none, so probably `beaker-tasks` is the only provider for those?

> However this ends up with `After beakerlib processing, tests may have only simple requirements` error, which is kind of confusing for the user. It originates from Prepare class where it checks whether dependency is 'simple' and throws an exception if it isn't, so it does not have a chance to pass the info to anything else. Not sure how to proceed here..

The error handling for that should be done in the discover step directly, not on the `PrepareInstall`. I need to check what is going on there, because I thought that that would already be the case, can you help me with a minimum reproducer?

As for proceeding with this, can you filter out all `Library` types, and for now, check explicitly for the `BeakerlibLibrary` + `.format == "rpm"`, as those are the only sources of `library(...)`. All the other `Library` types we drop and try to revisit after the `Library` refactors.

> So something is not working correctly in the processing of libraries on the list of required packages.
> 
> @LecrisUT could it be connected to one of your PRs?

Yes and no. I have not touched that logic yet, so cannot say for certain what is going on, but something is wrong for sure. The main logic that is handling `Library` dependencies is in
https://github.com/teemtee/tmt/blob/746105f16ad0d015dd71de552b9ac1431658bb07/tmt/libraries/__init__.py#L140-L149
https://github.com/teemtee/tmt/blob/746105f16ad0d015dd71de552b9ac1431658bb07/tmt/steps/discover/__init__.py#L450-L456
My understanding so far is that it is trying to go through the list and pick out all the `Library` types and drop all library types unless they were defined as `library(yum/false-library)` in which case forward it as a normal dependency for `PrepareInstall` to handle [^1]. Thinking about it more, normal `Library` types that are not `BeakerlibLibrary` + `.format == "rpm"` should really not be present there, so maybe there is a regression, let me check more closely.

[^1]: The logic in there is super convoluted and unnecessary, will try to refactor it

Comment (bajertom) [reply tmt/steps/discover/__init__.py]: Reproducer: any test with non-existing library likes this
```
 - type: library
   url: https://github.com/beakerlib/yum
   name: /false-library
```
or just
```
- url: https://github.com/beakerlib/yum
  name: /false-library
```
in `require` section of its main.fmf. Running it results in https://github.com/teemtee/tmt/blob/main/tmt/base.py#L712-L735 throwing
```
fail: Invalid requirement: DependencyFmfId(fmf_root=None, git_root=None, default_branch=None, url='https://github.com/beakerlib/yum', ref=None, path=None, name='/false-library', destination=None, nick=None, type='library')
```
plus the exception
```
plan failed.

Cause number 1:

    After beakerlib processing, tests may have only simple requirements
```
coming from here https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/__init__.py#L275-L317

Comment (happz) [reply tmt/steps/discover/__init__.py]: Yeah, I'd expect these library-like requirements be consumed by the layer responsible for beakerlib libraries and other complex requirements.

Comment (bajertom) [reply tmt/steps/discover/__init__.py]: I've made code changes to reflect the libraries problem. Now it raises at a Discover level without passing anything to Prepare.

Having uninstallable library results in:
```
        fail: Failed to process beakerlib libraries (/bad-library) for test '/tests/SECOND', '/tests/THIRD'.
        fail: Failed to process beakerlib libraries (/false-library) for test '/tests/THIRD'.
        fail: Failed to process beakerlib libraries (/wrong-library) for test '/tests/FIRST', '/tests/SECOND', '/tests/THIRD'.
    report
        how: display
        summary: no results found
    cleanup
        warn: Nothing to cleanup, no guests provisioned.

plan failed.

Cause number 1:

    Failed to process beakerlib libraries.
```

File: tmt/package_managers/apk.py [hunk 15]
```diff
@@ -155,6 +155,12 @@ class Apk(PackageManager[ApkEngine]):
 
     _engine_class = ApkEngine
 
+    # Compiled regex patterns for APK error messages
+    _UNABLE_TO_LOCATE_PATTERN = re.compile(r'unable to locate package\s+([^\s]+)', re.IGNORECASE)
+    _NO_SUCH_PACKAGE_PATTERN = re.compile(r'ERROR:\s+([^\s:]+):\s+No such package', re.IGNORECASE)
+
+    _FAILED_PACKAGE_INSTALLATION_PATTERNS = [_UNABLE_TO_LOCATE_PATTERN, _NO_SUCH_PACKAGE_PATTERN]
```
happz: More polishing :) If `_UNABLE_TO_LOCATE_PATTERN` is not used, it does not need to have a name, and it can be simply an item of the `_FAILED_PACKAGE_INSTALLATION_PATTERNS` list:

```python
_FAILED_PACKAGE_INSTALLATION_PATTERNS = [
    re.compile(r'unable to locate package\s+([^\s]+)', re.IGNORECASE),
    re.compile(r'ERROR:\s+([^\s:]+):\s+No such package', re.IGNORECASE)
]
```

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +471,37 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+            lines: list[str] = []
+
+            for library_name in sorted(library_failures.keys()):
+                tests = ', '.join(f"'{name}'" for name in sorted(library_failures[library_name]))
+                lines.append(
+                    f"Failed to process beakerlib libraries ({library_name}) for test {tests}."
+                )
+
+            for line in lines:
+                self._logger.fail(line)
```
happz: You can emit these lines right from the loop above, no need to collect them in `lines`.

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +471,37 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+                ]
+
+                library_names = {
+                    (dependency.name or '/')
+                    for dependency in unresolved_dependencies
+                    if isinstance(dependency, tmt.base.DependencyFmfId)
+                }
+
+                for library_name in library_names:
+                    library_failures[library_name].add(test.name)
```
happz: I think it would be nice to also report the other unresolved dependencies - we learn about libraries only, but since we are in the business of reporting unresolved libraries, we can take it all the way. There's at least `DependencyFile`. Maybe it would be possible to collect unresolved dependencies in general, e.g. separate them into buckets by their class names, then by their "name", and collect "their" tests there? A final report would do the libraries in some nice way, then the `DependencyFile`, and then if there's any other class name left, some generic "don't know how to present it nicer yet, but this we failed to resolve" output.


Maybe something along the following:

```python
# Collect all unresolved dependencies, per class, each dependency with its own list of tests.
unresolved_dependencies = collections.defaultdict(lambda: collections.defaultdict(list))

...

for dependency in (*test.require, *test.recommend):
    if isinstance(dependency, DependencySimple):
        continue

    unresolved_dependencies[dependency.__class__.__name__][dependency].append(test)

...

# This is the end, right?

# Report unresolved libraries
for dependency, tests in unresolved_dependencies.pop(tmt.base.DependencyFmfId.__name__, {}).items():
    test_names = ', '.join(f"'{name}'" for name in sorted(test.name for test in tests)
    self._logger.fail(
        f"Failed to process beakerlib library {dependency.name or "/"}), required by tests {test_names}."
    )

if unresolved_dependencies:
    # Rough reporting for dependencies we don't know how to report in a better way
    ...
```

tmt/package_managers/apk.py
bajertom: Addressed in 5dfec9c84c9218c985281212205cdd148cf6cb2a

tmt/steps/discover/__init__.py
bajertom: Addressed in 5dfec9c84c9218c985281212205cdd148cf6cb2a

tmt/steps/discover/__init__.py
bajertom: Addressed in 5dfec9c84c9218c985281212205cdd148cf6cb2a
Now the outcome of unresolved library or file dependencies looks like this:
```
....
        fail: Failed to process beakerlib libraries (/bad-library) for test '/tests/SECOND', '/tests/THIRD'.
        fail: Failed to process beakerlib libraries (/common-functions) for test '/tests/FIRST'.
        fail: Failed to process beakerlib libraries (/false-library) for test '/tests/THIRD'.
        fail: Failed to process beakerlib libraries (/wrong-library) for test '/tests/FIRST', '/tests/SECOND', '/tests/THIRD'.
        fail: Failed to process file dependencies (/completely/wrong/path) for test '/tests/SECOND'.
        fail: Failed to process file dependencies (/path/that/does/not/exist) for test '/tests/THIRD'.
        fail: Failed to process file dependencies (/path/that/does/not/exist, /another/wrong/path) for test '/tests/FIRST'.
    report
        how: display
        summary: no results found
    cleanup
        warn: Nothing to cleanup, no guests provisioned.

plan failed.

Cause number 1:

    Failed to process some dependencies.
```
When there are invalid packages, it does not get to that stage at all and raises already here.

bajertom: I just found out that it displays valid library as invalid one if there is also an invalid file dependency present in the test metadata, so this needs more work.

tmt/steps/discover/__init__.py
```diff
@@ -456,17 +457,76 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+                # If beakerlib processing fails, use the original dependencies
+                # for reporting purposes and continue with other tests.
+                except tmt.utils.MetadataError:
+                    dependencies = (*test.require, *test.recommend)
+
+                for dependency in dependencies:
+                    if isinstance(dependency, tmt.base.DependencySimple):
+                        continue
+
+                    class_name = dependency.__class__.__name__
```
LecrisUT: Can you make an issue to eventually improve this logic from the library package part?

tmt/libraries/__init__.py
```diff
@@ -212,7 +212,7 @@ def already_fetched(lib: Dependency) -> bool:
                 processed_recommend.update(set(recommends))
                 gathered_libraries.extend(libraries)
         # Regular package require/recommend
-        except LibraryError:
+        except (LibraryError, tmt.utils.MetadataError):
```
LecrisUT: Why do we need this change?

tmt/steps/discover/__init__.py
```diff
@@ -575,6 +635,26 @@ def discover_tests(
...
         phase.apply_policies()
 
+    @property
+    def dependencies_to_tests(self) -> tuple[dict[str, list[str]], dict[str, list[str]]]:
+        """
+        A tuple containing two dictionaries mapping dependencies to tests (required & recommended)
+        """
+
+        required_dependencies_to_tests: dict[str, list[str]] = defaultdict(list)
+        recommended_dependencies_to_tests: dict[str, list[str]] = defaultdict(list)
```
LecrisUT, line 685: Not keen on calculating these each time the property is executed, but ok to deffer this for later. Ideally this will be populated during the `plugin.go()` and `tmt.libraries.dependencies`, but these need a bigger refactor to do so.

tmt/steps/discover/__init__.py
bajertom: Fixed in 7c9ff651c9f105d76a093978437b3dfff65dfe33

LecrisUT (commented):
Overall LGTM, other than one unclear change

tmt/libraries/__init__.py
bajertom: When there are both **valid** and **invalid** libraries AND an **invalid** file dependency in test metadata at the same time, it results in false negative - the **valid** library being reported as **invalid**. See the output in https://github.com/teemtee/tmt/pull/4430#discussion_r2764451378.
`/common-functions` should not be there, but it is because:

when invalid file dependency is being processed in [tmt/libraries/__init__.py.dependencies()](https://github.com/teemtee/tmt/blob/1e67cc3be9b1fba4d6a8e87f8a1db480662f15f0/tmt/libraries/__init__.py#L153) it throws `tmt.utils.MetadataError` and exits the loop immediately. It propagates to https://github.com/teemtee/tmt/pull/4430/changes#diff-5f59acdec9a1fdc81c556e6ca1c7c872a25d72f1be0259a8f727b9dce4a45a44L450 where it is caught by the except `tmt.utils.MetadataError` clause and then the unprocessed dependencies are being used. Because the [tmt/libraries/__init__.py.dependencies()](https://github.com/teemtee/tmt/blob/1e67cc3be9b1fba4d6a8e87f8a1db480662f15f0/tmt/libraries/__init__.py#L153)  never returns the processed dependencies, unprocessed dependencies are used and the **valid** library is still seen as `DependencyFmfId` instead of `DependencySimple` and therefore later reported as invalid.

With the `tmt.utils.MetadataError` included, it continues the loop even if it encounters **invalid** file dependency, then after it ends looping it returns the processed dependencies to DiscoverPlugin.install_libraries(), where it is filtered out - the **valid** library would be seen as `DependencySimple`, whereas both **invalid** library(still `DependencyFmfId`) and **invalid** file dependency would get reported as invalid.

It's kind of a corner case where the user has to mess up twice, but it could happen I guess?

Anyway looking at it now after writing this explanation, this comment
```
# If beakerlib processing fails, use the original dependencies
# for reporting purposes and continue with other tests.
```
is misleading, because the except `tmt.utils.MetadataError` right below it is never reached. It's caught every time in [tmt/libraries/init.py.dependencies()](https://github.com/teemtee/tmt/blob/1e67cc3be9b1fba4d6a8e87f8a1db480662f15f0/tmt/libraries/__init__.py#L153) and then the function returns. The whole try/except block could be removed and `dependencies = (*test.require, *test.recommend)` used just once.

bajertom: Removed the try/except block in 59d69aefa267f855ef5d250ee9b3a989dfe95f3e, but the `tmt.utils.MetadataError` check logic still applies.

LecrisUT: > When there are both valid and invalid libraries AND an invalid file dependency in test metadata at the same time, it results in false negative - the valid library being reported as invalid.

This is about a valid library depending on an invalid library? Well one way to view it is that the parent library is also invalid because it cannot be resolved. I would keep the old behavior if that is the case (at least for now until we detangle it a bit more)

If this is not about that, then seems like a bug (not surprised if it falls in this category)

tmt/steps/discover/__init__.py
LecrisUT: The `try ... except` was removed here, but the comment still applies for the general code of dealing with these dependencies.

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +473,55 @@ def install_libraries(self, source: Path, target: Path) -> None:
                     target_location=target,
                 )
 
+                dependencies = (*test.require, *test.recommend)
+
+                for dependency in dependencies:
+                    if isinstance(dependency, tmt.base.DependencySimple):
+                        continue
+
+                    class_name = dependency.__class__.__name__
```
happz: `dependency_class_name` - there's a lot of classes involved, knowing it's the dependency one helps.

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +473,55 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+
+                    if isinstance(dependency, tmt.base.DependencyFmfId):
+                        dep_key = dependency.name or '/'
```
happz: I would squeeze this value into `[...]` below. The value is not reused, the name exists only to appear once, and `dep_name` is a risky name too: there are various "*dependenc*"-like names, suddenly a very shortened "dep" appears.

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +473,55 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+                        unresolved_dependencies[class_name][dep_key].add(test.name)
+
+                    elif isinstance(dependency, tmt.base.DependencyFile):
+                        for pattern in dependency.pattern or []:
+                            unresolved_dependencies[class_name][pattern].add(test.name)
+
+                    else:
+                        dep_key = str(dependency)
```
happz: Same as above, `[str(dependency)]` would work too.

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +473,55 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+        # without rerunning tmt repeatedly.
+        if unresolved_dependencies:
+
+            def _report_unresolved_dependencies(
+                dependencies: dict[str, dict[str, set[str]]],
+                class_name: str,
+                label: Optional[str] = None,
+            ) -> None:
+                display_label = label or f"{class_name} dependency"
+                for dep_key, tests in sorted(dependencies.pop(class_name, {}).items()):
```
happz: Maybe `dependency_name`? Or `dependency_key`, but not `dep`. Throughout the code they are called `dependencies` and `dependency_whatever`, `dep` does not fit this pattern.

tmt/steps/discover/__init__.py
```diff
@@ -468,6 +473,55 @@ def install_libraries(self, source: Path, target: Path) -> None:
...
+                    test_names = ', '.join(f"'{name}'" for name in sorted(tests))
+                    self._logger.fail(
+                        f"Failed to process {display_label} ({dep_key}) for test {test_names}."
+                    )
+
+            # Known types first
+            for class_name, label in (
```
happz: `dependency_class_name`

tmt/steps/discover/__init__.py
happz: Would `cached_property` work? Depends when this is property is used, whether the list of tests (and their requirements) is already finalized.

tmt/libraries/__init__.py
bajertom: No, it was not a valid library depending on an invalid one, they were two independent libraries, one valid, one invalid, both required by one test. Result of the introduced code was that both were rendered invalid in the report. So it was a bug, but not in the original code, but in my PR. Later commit fixed that.

Scenario of a valid library depending on an invalid one is a scenario I actually haven't thought of - good catch.

tmt/steps/discover/__init__.py
LecrisUT: My issues is that when the timing when these dependencies are calculated is ill-defined, especially since the `test.(require|recommend)` change after `tmt.libraries.dependencies` is called in `install_libraries`. I don't think we have a good way to solve this in this PR as it would require more library refactoring.

tmt/libraries/__init__.py
bajertom: As of right now if a valid library requires an invalid one, only the invalid one is being reported for that specific test case.

tmt/steps/discover/__init__.py
bajertom: Addressed in 4b7d11468506999051b89ad48580921777ed98af.

tmt/steps/discover/__init__.py
bajertom: Addressed in 4b7d11468506999051b89ad48580921777ed98af.

tmt/steps/discover/__init__.py
bajertom: Addressed in 4b7d11468506999051b89ad48580921777ed98af.

tmt/steps/discover/__init__.py
bajertom: Addressed in 4b7d11468506999051b89ad48580921777ed98af.

tmt/steps/discover/__init__.py
bajertom: Addressed in 4b7d11468506999051b89ad48580921777ed98af.

tmt/steps/discover/__init__.py
happz: So this thread can be resolved?

LecrisUT: Yes, just to be added to the follow-up pile. (Might even get to this one since if we continue with the library makeover)

bajertom:
### Bug fix
 I have fixed a bug that showed *all* the recommended packages, both valid and invalid, by removing 

```
re.compile(r'no package provides\s+([^\s\n]+)', re.IGNORECASE)
```
because that line is misleading and packages are getting installed after that, which gave false negatives. I've added 
```
re.compile(r'Could not find a package for:\s*([^\s]+)', re.IGNORECASE)
```
which was missing to make it work correctly for yum.

### Debuginfo packages
I have adjusted tests for required and recommended packages. In the recommend sections I've found out that debuginfo packages are handled differently:

1. debuginfo repository is installed, enabled
2. "package-debuginfo" is debuginfo-installed by it's basename "package"
```
rpm -q --whatprovides /usr/bin/debuginfo-install || yum install -y  /usr/bin/debuginfo-install && rpm -q --whatprovides /usr/bin/debuginfo-install && debuginfo-install -y --skip-broken
```
How to handle this? With the current state of PR tmt is gonna say that
```
"package name" failed to install, required by "test name".
```
Should it say "package-debuginfo"?

### Verbosity increase?
I am really not sure whether increasing the verbosity in [tests/prepare/recommend/test.sh](https://github.com/teemtee/tmt/pull/4430/changes#diff-106499552af1d6f685e07283bcf6215fe1020aeccbe188b2cf19149e92ca05b4) is the correct way to go. Without it the lines 
```
rlAssertGrep "forest: recommended by: /test/debuginfo" $rlRun_LOG
```
in the `debuginfo` scenario are not grepped, but only for CentOS7 (due to yum I guess).

tmt/libraries/__init__.py
LecrisUT: Had to look deeper into this. The issue here is quite more complex, it can either come from:
- `FileLibrary` failing, but in that case it is reporting the wrong library
- some other *unexpected* error. I have looked through the `raise.*MetadataError`, but I did not find anything that I would expect to be called here which would trigger this

For now can you revert and if anything let's dive deeper with an example of how this fails.

LecrisUT:
> ### Bug fix
> 
> I have fixed a bug that showed _all_ the recommended packages, both valid and invalid, by removing
> 
> ```
> re.compile(r'no package provides\s+([^\s\n]+)', re.IGNORECASE)
> ```
> 
> because that line is misleading and packages are getting installed after that, which gave false negatives. I've added

I assume this is when it is trying to do `rpm -q`. That thing needs a revision to split it in two, one part to see which packages to install and the other to install them. This change could actually help do that.

tmt/libraries/__init__.py
bajertom: Without the `tmt.utils.MetadataError` getting raised in file.py and caught in the libraries/__init__.py, it unfortunately does not work and gives out the `Patterns {patterns} don't match any files in '{self.source_location}'` error message from file.py.

I've changed the `tmt.utils.MetadataError` in file.py to `LibraryError` and removed the error message, as it will now inform the user via the `_report_unresolved_dependencies()` from discover/__init__.py.

What do you think? Is this acceptable? Changed in 2e774cec0f8d95e968969f1b0828866848063ec9

LecrisUT: > Without the `tmt.utils.MetadataError` getting raised in file.py and caught in the libraries/**init**.py, it unfortunately does not work and gives out the `Patterns {patterns} don't match any files in '{self.source_location}'` error message from file.py.

Uph, and that would have created installation issues and unpredictable installations all around even before that :face_exhaling:.

> I've changed the `tmt.utils.MetadataError` in file.py to `LibraryError` and removed the error message, as it will now inform the user via the `_report_unresolved_dependencies()` from discover/**init**.py.

I don't think that would help because it would just fall through into
```
        except LibraryError:
            if dependency in original_require:
                processed_require.add(dependency)
            if dependency in original_recommend:
                processed_recommend.add(dependency)
```
Ah, but we are catching any errors at this stage and not even going into `PrepareInstall`. Yeah, seems like it would be a good approach for now

LecrisUT (approved):

tmt/steps/prepare/install.py
```diff
@@ -1002,7 +1095,40 @@ def go(
...
+                )
+
+                failed_packages.update(
+                    str(package).removesuffix("-debuginfo")
+                    for package, present in presence.items()
+                    if not present
+                )
+
+            if failed_packages:
+                self._show_failed_packages_with_tests(failed_packages)
```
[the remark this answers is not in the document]
bajertom: Debuginfo packages are stripped of their "debuginfo" suffix when handled by the installer. If they are **required** and fail to install, they will be logged with their base name.

If they are **recommended** with --skip-broken, the same thing applies except for an edge case, where in CentOS7 yum tries to debuginfo-install debuginfo packages in random order[^1].

If the queue starts with a valid debuginfo package followed by an invalid one, eg `grep` and then `forest`, in the output we can see the "Could not find a package for: <package name>", that is being search for in `_extract_failed_packages_from_output`:
```
cmd: rpm -q --whatprovides /usr/bin/debuginfo-install || yum install -y  /usr/bin/debuginfo-install && rpm -q --whatprovides /usr/bin/debuginfo-install && debuginfo-install -y --skip-broken grep forest
            stdout: yum-utils-1.1.31-54.el7_8.noarch
            stdout: yum-utils-1.1.31-54.el7_8.noarch
            stdout: Loaded plugins: fastestmirror, ovl
            stdout: enabling base-debuginfo
            stdout: Determining fastest mirrors
            stdout: Could not find debuginfo pkg for dependency package glibc-2.17-326.el7_9.3.x86_6
 ==========>stdout: Could not find a package for: forest <==========
            stdout: --> Running transaction check
            stdout: ---> Package grep-debuginfo.x86_64 0:2.20-3.el7 will be installed
            stdout: ---> Package pcre-debuginfo.x86_64 0:8.32-17.el7 will be installed
            stdout: ---> Package yum-plugin-auto-update-debug-info.noarch 0:1.1.31-54.el7_8 will be installed
            stdout: --> Finished Dependency Resolution
            stdout: 
            stdout: Dependencies Resolved

```

However if the queue starts with in invalid package followed by valid ones, eg `forest` and then `grep`, the output looks like this:
```
cmd: rpm -q --whatprovides /usr/bin/debuginfo-install || yum install -y  /usr/bin/debuginfo-install && rpm -q --whatprovides /usr/bin/debuginfo-install && debuginfo-install -y --skip-broken forest grep
            stdout: yum-utils-1.1.31-54.el7_8.noarch
            stdout: yum-utils-1.1.31-54.el7_8.noarch
            stdout: Loaded plugins: fastestmirror, ovl
            stdout: enabling base-debuginfo
            stdout: Determining fastest mirrors
            stdout: Could not find debuginfo pkg for dependency package glibc-2.17-326.el7_9.3.x86_64
            stdout: --> Running transaction check
            stdout: ---> Package grep-debuginfo.x86_64 0:2.20-3.el7 will be installed
            stdout: ---> Package pcre-debuginfo.x86_64 0:8.32-17.el7 will be installed
            stdout: ---> Package yum-plugin-auto-update-debug-info.noarch 0:1.1.31-54.el7_8 will be installed
            stdout: --> Finished Dependency Resolution
            stdout: 
            stdout: Dependencies Resolved
```
So an extra check for presence has to be done and the `debuginfo` suffix has to be stripped manually to align with the behavior of installers treating it by its basename. This way the `_show_failed_packages_with_tests` always shows the basename of the package to the user.

[^1]: that is because https://github.com/teemtee/tmt/blob/225ca30a9e096512449c3341e6a0f335c7317204/tmt/utils/__init__.py#L3300 does not preserve order of elements, so the packages end up in the debuginfo-install in a random order

bajertom:
In cc830f9dab31053610c6060163896b2a4a8e1d27 I've changed raising LibraryError instead of SpecificationError to allow the discover step to catch these failures and report which tests required the  unresolvable library, rather than aborting immediately with no test context.

In 199fc61883794740c0265c7a633eff58c53fb3e6 I've added output catching logic for bootc installer as well.

In the latest I have added back error message about patterns not being found for LibraryError.

tmt/steps/prepare/install.py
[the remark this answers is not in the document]
happz: TBH, a shortened version of this comment would be nice if presented as a comment in the code. Package managers are full of traps like this, and seeing the line of thinking behind the algorithm can be invaluable 6 months from now.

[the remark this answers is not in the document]
bajertom: It already is:
 ```
 # Output from yum is non-deterministic. It depends on order in which
 # the invalid debuginfo package is getting installed. If first, its error messages
 # are omitted. If after a valid package, the error messages are included.
 # So we're checking for presence on the system. 
 ```
 https://github.com/teemtee/tmt/pull/4430/changes#diff-2d4f47a0625c721a2b882d2bf06558193efeef8c220f169d4b5e59169363f62cR1117-R1120 
 
 Should I extend it a bit?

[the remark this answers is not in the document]
happz: Ah, I missed that. Never mind then.

happz (approved):

happz:
@bajertom rebase, please, there seem to be some conflicts.

happz:
@bajertom I'm afraid one more rebase is needed.

happz:
@bajertom oh, and the release note is missing...

docs/releases/pending/4430.fmf
```diff
@@ -0,0 +1,5 @@
+description: |
+    When package installation fails during the ``prepare`` :ref:`/plugins/prepare/install`
+    step, tmt now reports which specific packages failed and which tests required
```
happz:
```suggestion
    When package installation fails during the :ref:`/plugins/prepare/install`
    phase, tmt now reports which specific packages failed and which tests required
```

PR #4435, opened 2025-12-17
Split the doc-tests into multiple jobs

LecrisUT:
Sometimes these jobs are failing and blocks the other checks in the queue. But these jobs are not really gating so we could just run them in parallel. The GH actions for us is cheap by comparison anyway.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/.github/workflows/doc-tests.yml
+++ b/.github/workflows/doc-tests.yml
@@ -17,14 +17,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        builder: [ linkcheck, html ]
+        builder: [ linkcheck, html, lint ]
         include:
           # Run default html builder with warnings as error
           - builder: html
             args: -W
+          - builder: lint
+            sphinx_builder: html
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
       - uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-tests.yml
```diff
@@ -11,43 +11,48 @@ permissions:
...
+          # Run default html builder with warnings as error
+          - builder: html
+            args: -W
+          - builder: lint
+            sphinx_builder: html
 
     steps:
       - uses: actions/checkout@v6
-        with:
-          persist-credentials: false
```
psss, line 21: This seems to be reverting #3325, was that a bad idea?

LecrisUT: Had to look what it is about. I would say it is inconsequential. We are using the default `GITHUB_TOKEN` which is short lived, and even if it persists, we do not have private repos that would be exposed with it. If they wanted to abuse it, they would use `gh` cli directly and that gets the `GITHUB_TOKEN` regardless of the `actions/checkout`

Reverted anyway

LecrisUT:
> Plus the `linkcheck` job failed with:

Just needed to rebase. It was linkcheck failure over the christmas holidays. It's just Anubis-like protection that was fixed later on.

happz (approved):

PR #4437, opened 2025-12-17
Mock provision: make detection of finished processes more deterministic

mkoncek:
Based on: #4173 
I found out that various environments (`fedora-rawhide-x86_64`) print different number of empty strings on their stdout as the commands spawned in the chroots finish. This causes some environments to hang.
This PR makes the detection more deterministic.

Pull Request Checklist

* [ ] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [ ] include a release note

### DIFF
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -447,6 +447,10 @@ class MockShell:
                 # guaranteed.
                 if len(events) == 1 and events[0][0] == returncode_fd:
                     content = os.read(returncode_fd, 16)
+                    if not content:
+                        # EPOLLHUP with no data: the writer (echo $?>) has not
+                        # yet opened the FIFO. Keep waiting until it does.
+                        continue
                     try:
                         returncode = int(content.decode('ascii').strip())
                     except ValueError:
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -186,8 +186,6 @@ class MockShell:
         Invoke commands in the mock shell without checking for errors.
         This is done by writing the commands to the shell, ending each with
         a newline.
-        Each command invocation causes the shell to print a newline when the
-        command has finished.
         """
 
         assert self.epoll is not None
@@ -195,24 +193,29 @@ class MockShell:
         assert self.mock_shell.stdin is not None
         assert self.mock_shell.stdout is not None
         assert self.mock_shell.stderr is not None
+        finished_keyword = 'TMT_FINISHED_EXEC'
 
-        self.mock_shell.stdin.write(''.join(command + '\n' for command in commands))
-        # Issue a command writing a binary zero on the standard output after all
+        self.mock_shell.stdin.write('\n'.join(commands))
+        # Issue a command writing a terminator on the standard output after all
         # the previous commands are finished.
-        self.mock_shell.stdin.write('echo -e \\\\x00\n')
+        self.mock_shell.stdin.write(f'\necho {finished_keyword}\n')
         self.mock_shell.stdin.flush()
 
         # Wait until we read the binary zero from stdout.
-        loop = True
-        while loop and self.mock_shell.poll() is None:
+        while True:
+            mock_shell_result = self.mock_shell.poll()
+            if mock_shell_result is not None:
+                raise tmt.utils.ProvisionError(f'Mock shell abruptly exited: {mock_shell_result}.')
             events = self.epoll.poll()
             for fileno, _ in events:
                 if (
                     fileno == self.mock_shell_stdout_fd
-                    and self.mock_shell.stdout.read() == '\x00\n'
+                    and self.mock_shell.stdout.read() == f'{finished_keyword}\n'
                 ):
-                    loop = False
                     break
+            else:
+                continue
+            break
         for line in self.mock_shell.stderr.readlines():
             self.parent.debug('mock', line.rstrip(), color='blue', level=2)
 
@@ -440,8 +443,7 @@ class MockShell:
                     break
                 for fileno, _ in events:
                     if fileno == self.mock_shell_stdout_fd:
-                        # Various environments may print variable number of
-                        # times here.
+                        # Mock prints newlines on stdout.
                         self.mock_shell.stdout.read()
                     elif fileno == self.mock_shell_stderr_fd:
                         # Whatever we sent on mock shell's input it prints on
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -424,6 +424,15 @@ class MockShell:
             yield  # type: ignore[misc]
 
             while self.mock_shell.poll() is None:
+                # The `epoll.poll` call returns a list of pairs of all the
+                # events, that are currently ready. The pair consists of the
+                # file descriptor and the event type (we only registered
+                # select.EPOLLIN - ready for reading).
+                #
+                # `epoll.poll` is level-based: each time it is invoked, it
+                # returns a list of descriptors which are ready and it will
+                # contain those descriptors until we actually read from the
+                # descriptor.
                 events = self.epoll.poll(timeout=timeout)
 
                 if len(events) == 0:
@@ -433,8 +442,9 @@ class MockShell:
 
                 # The command is finished when the returncode is written to its
                 # file.
-                # We want to break loop after we handled all the other epoll
-                # events because the event ordering is not guaranteed.
+                # We want to break loop after we handled all the epoll events
+                # other than `returncode` because the event ordering is not
+                # guaranteed.
                 if len(events) == 1 and events[0][0] == returncode_fd:
                     content = os.read(returncode_fd, 16)
                     returncode = int(content.decode('utf-8').strip())
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -201,16 +201,15 @@ class MockShell:
         self.mock_shell.stdin.write(f'\necho {finished_keyword}\n')
         self.mock_shell.stdin.flush()
 
-        # Wait until we read the binary zero from stdout.
+        # Wait until we read the `finished_keyword`.
         while True:
             mock_shell_result = self.mock_shell.poll()
             if mock_shell_result is not None:
                 raise tmt.utils.ProvisionError(f'Mock shell abruptly exited: {mock_shell_result}.')
             events = self.epoll.poll()
             for fileno, _ in events:
-                if (
-                    fileno == self.mock_shell_stdout_fd
-                    and self.mock_shell.stdout.read() == f'{finished_keyword}\n'
+                if fileno == self.mock_shell_stdout_fd and self.mock_shell.stdout.read().endswith(
+                    f'{finished_keyword}\n'
                 ):
                     break
             else:
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -447,8 +447,13 @@ class MockShell:
                 # guaranteed.
                 if len(events) == 1 and events[0][0] == returncode_fd:
                     content = os.read(returncode_fd, 16)
-                    returncode = int(content.decode('utf-8').strip())
-                    returncode_io.try_unregister()
+                    try:
+                        returncode = int(content.decode('ascii').strip())
+                    except ValueError:
+                        # Missing `returncode` is handled outside of the loop.
+                        break
+                    finally:
+                        returncode_io.try_unregister()
                     break
                 for fileno, _ in events:
                     if fileno == self.mock_shell_stdout_fd:
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -197,16 +197,21 @@ class MockShell:
         assert self.mock_shell.stderr is not None
 
         self.mock_shell.stdin.write(''.join(command + '\n' for command in commands))
+        # Issue a command writing a binary zero on the standard output after all
+        # the previous commands are finished.
+        self.mock_shell.stdin.write('echo -e \\\\x00\n')
         self.mock_shell.stdin.flush()
 
-        # Wait until the previous commands finished.
-        loop = len(commands)
-        while loop != 0 and self.mock_shell.poll() is None:
+        # Wait until we read the binary zero from stdout.
+        loop = True
+        while loop and self.mock_shell.poll() is None:
             events = self.epoll.poll()
             for fileno, _ in events:
-                if fileno == self.mock_shell_stdout_fd:
-                    loop -= 1
-                    self.mock_shell.stdout.read()
+                if (
+                    fileno == self.mock_shell_stdout_fd
+                    and self.mock_shell.stdout.read() == '\x00\n'
+                ):
+                    loop = False
                     break
         for line in self.mock_shell.stderr.readlines():
             self.parent.debug('mock', line.rstrip(), color='blue', level=2)
@@ -424,16 +429,23 @@ class MockShell:
                     # kill the process spawned inside the mock shell
                     pass
 
-                # The command is finished when mock shell prints a newline on its
-                # stdout. We want to break loop after we handled all the other
-                # epoll events because the event ordering is not guaranteed.
-                if len(events) == 1 and events[0][0] == self.mock_shell_stdout_fd:
-                    self.mock_shell.stdout.read()
+                # The command is finished when the returncode is written to its
+                # file.
+                # We want to break loop after we handled all the other epoll
+                # events because the event ordering is not guaranteed.
+                if len(events) == 1 and events[0][0] == returncode_fd:
+                    content = os.read(returncode_fd, 16)
+                    returncode = int(content.decode('utf-8').strip())
+                    returncode_io.try_unregister()
                     break
                 for fileno, _ in events:
-                    # Whatever we sent on mock shell's input it prints on the stderr
-                    # so just discard it.
-                    if fileno == self.mock_shell_stderr_fd:
+                    if fileno == self.mock_shell_stdout_fd:
+                        # Various environments may print variable number of
+                        # times here.
+                        self.mock_shell.stdout.read()
+                    elif fileno == self.mock_shell_stderr_fd:
+                        # Whatever we sent on mock shell's input it prints on
+                        # the stderr so just discard it.
                         self.mock_shell.stderr.read()
                     elif fileno == stdout_fd:
                         content = os.read(stdout_fd, 128)
@@ -445,12 +457,6 @@ class MockShell:
                         stream_err += content
                         if not content:
                             stderr_io.try_unregister()
-                    elif fileno == returncode_fd:
-                        content = os.read(returncode_fd, 16)
-                        if not content:
-                            returncode_io.try_unregister()
-                        else:
-                            returncode = int(content.decode('utf-8').strip())
 
             stdout = stream_out.string
             stderr = stream_err.string
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -446,6 +446,7 @@ class MockShell:
                 # other than `returncode` because the event ordering is not
                 # guaranteed.
                 if len(events) == 1 and events[0][0] == returncode_fd:
+                    # Read the return code number string into a large enough buffer.
                     content = os.read(returncode_fd, 16)
                     if not content:
                         # EPOLLHUP with no data: the writer (echo $?>) has not
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -709,6 +709,8 @@ class GuestMock(tmt.Guest):
         source = source or self.plan_workdir
         destination = destination or source
 
+        if options.delete:
+            self.mock_shell.execute(Command('rm', '-rf', str(destination)), logger=self._logger)
         if source.is_dir():
             self.mock_shell.execute(Command('mkdir', '-p', str(destination)), logger=self._logger)
             p = self.mock_shell._spawn_command(
--- a/tmt/steps/provision/mock.py
+++ b/tmt/steps/provision/mock.py
@@ -446,8 +446,13 @@ class MockShell:
                 # other than `returncode` because the event ordering is not
                 # guaranteed.
                 if len(events) == 1 and events[0][0] == returncode_fd:
-                    # Read the return code number string into a large enough buffer.
-                    content = os.read(returncode_fd, 16)
+                    try:
+                        # Read the return code number string into a large enough buffer.
+                        content = os.read(returncode_fd, 16)
+                    except BlockingIOError:
+                        # EAGAIN: the writer opened the FIFO but has not written data yet.
+                        # Keep waiting.
+                        continue
                     if not content:
                         # EPOLLHUP with no data: the writer (echo $?>) has not
                         # yet opened the FIFO. Keep waiting until it does.
--- a/tests/provision/mock/main.fmf
+++ b/tests/provision/mock/main.fmf
@@ -1,3 +1,5 @@
+require+:
+  - tmt+provision-mock
 tag+:
   - provision-only
   - provision-mock
--- a/tmt/package_managers/mock.py
+++ b/tmt/package_managers/mock.py
@@ -102,7 +102,7 @@ class _MockPackageManager(PackageManager[MockEngine]):
     """
 
     probe_command = Command('/usr/bin/false')
-    probe_priority = 140
+    probe_priority = 130
     _engine_class = MockEngine
 
     def check_presence(self, *installables: Installable) -> dict[Installable, bool]:
--- a/tmt/package_managers/mock.py
+++ b/tmt/package_managers/mock.py
@@ -102,7 +102,7 @@ class _MockPackageManager(PackageManager[MockEngine]):
     """
 
     probe_command = Command('/usr/bin/false')
-    probe_priority = 130
+    probe_priority = 140
     _engine_class = MockEngine
 
     def check_presence(self, *installables: Installable) -> dict[Installable, bool]:


mkoncek:
@happz Rebased on the current `HEAD`.

tmt/steps/provision/mock.py
```diff
@@ -181,16 +180,19 @@ def _simple_execute(self, *commands: str) -> None:
         assert self.mock_shell.stderr is not None
 
         self.mock_shell.stdin.write(''.join(command + '\n' for command in commands))
+        self.mock_shell.stdin.write('echo -e \\\\x00\n')
```
LecrisUT: Well the intent of these changes is unintuitive, can you elaborate on what is going on here?

mkoncek: Rebased with more elaborative comments.

mkoncek:
The primary cause of this bug was the assumption that each finished command prints a single newline on stdout after it is finished. This turned out not to be true in some environments. So we print a binary zero after the initial commands are finished and then loop until we read that binary zero back.

Later the command completion is detected by returncode being written into its pre-determined file, instead of waiting on a newline on stdout again.

tmt/steps/provision/mock.py
```diff
@@ -181,16 +180,21 @@ def _simple_execute(self, *commands: str) -> None:
         assert self.mock_shell.stderr is not None
 
         self.mock_shell.stdin.write(''.join(command + '\n' for command in commands))
+        # Issue a command writing a binary zero on the standard output after all
+        # the previous commands are finished.
+        self.mock_shell.stdin.write('echo -e \\\\x00\n')
```
LecrisUT: So iiuc the binary zero is arbitrary. Can we instead make it `echo TMT_FINISHED_EXEC` or such?

tmt/steps/provision/mock.py
```diff
@@ -181,16 +180,21 @@ def _simple_execute(self, *commands: str) -> None:
...
         self.mock_shell.stdin.flush()
 
-        # Wait until the previous commands finished.
-        loop = len(commands)
-        while loop != 0 and self.mock_shell.poll() is None:
+        # Wait until we read the binary zero from stdout.
+        loop = True
+        while loop and self.mock_shell.poll() is None:
```
```suggestion
        while self.mock_shell.poll() is None:
```
Do we still need `loop` in here? If we do maybe some better comments is in order. You could probably use `for ... else` to make the logic clearer, maybe
```python
while True:
    for fileno, _ in events:
        if check_finished():
            break
    else:
        # Did not find the finish command, do another poll
        continue
    # check_finished was successful, we can finish now
    break
```
(but if I read correctly you do not need the `while True` and `self.mock_shell.poll() is None` would take care of finishing the loop?)

tmt/steps/provision/mock.py
```diff
@@ -428,12 +440,6 @@ def _spawn_command(
                         stream_err += content
                         if not content:
                             stderr_io.try_unregister()
-                    elif fileno == returncode_fd:
-                        content = os.read(returncode_fd, 16)
-                        if not content:
-                            returncode_io.try_unregister()
-                        else:
-                            returncode = int(content.decode('utf-8').strip())
```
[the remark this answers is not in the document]
mkoncek: The "event loop" is the whole `while` loop one level earlier, not just the `for` loop.

tmt/steps/provision/mock.py
mkoncek: Yes, possible.

tmt/steps/provision/mock.py
mkoncek: `self.mock_shell.poll() is None` return is unexpected, it would mean the mock shell process exited abruptly.

mkoncek: I believe there is no other way than a double loop. It takes some time for the epoll events to arrive and each `epoll.poll()` returns a list.

mkoncek: `for... else` is possible, I was not sure about whether it is actually more readable.

mkoncek:
Also:
```python
        self.mock_shell.stdin.write(''.join(command + '\n' for command in commands))
        # Issue a command writing a binary zero on the standard output after all
        # the previous commands are finished.
        self.mock_shell.stdin.write('echo -e \\\\x00\n')
```
could be rewritten as:
```python
        self.mock_shell.stdin.write('\n'.join(commands))
        # Issue a command writing a binary zero on the standard output after all
        # the previous commands are finished.
        self.mock_shell.stdin.write('\necho -e \\\\x00\n')
```
And then we could use a different string to mark the end of commands.

Afaik in that initialization we only execute things like `rm`, `mkdir`, `mkfifo` and such so there are no arbitrary strings written to stdout.

tmt/steps/provision/mock.py
LecrisUT: > `self.mock_shell.poll() is None` return is unexpected, it would mean the mock shell process exited abruptly.

Ah, so the only purpose is to make sure `poll` is called early inside the loop. Would be more readable if it is just added there, and it should be checked for `None` with a `raise` as an unexpected event.

> `for... else` is possible, I was not sure about whether it is actually more readable.

Well, it's more readable than the variable `loop`, for me at least 😅. Ultimately if you can make a relevant comment to track the logic that it is trying to do, then either is fine.

tmt/steps/provision/mock.py
```diff
@@ -408,16 +412,23 @@ def _spawn_command(
...
+                    returncode = int(content.decode('utf-8').strip())
+                    returncode_io.try_unregister()
                     break
                 for fileno, _ in events:
-                    # Whatever we sent on mock shell's input it prints on the stderr
-                    # so just discard it.
-                    if fileno == self.mock_shell_stderr_fd:
+                    if fileno == self.mock_shell_stdout_fd:
+                        # Various environments may print variable number of
+                        # times here.
```
LecrisUT: What is the various environments referring here?

mkoncek: As mentioned in the description of the PR, this was first discovered about a month ago on `fedora-rawhide-x86_64`, other older chroots behave the old way.

mkoncek:
@LecrisUT I went over your concerns and tried to apply fixes.

LecrisUT (approved):
Looks good enough for me. Could you rebase please?

Unrelated, but can you also open a PR to revert and give more context on what needs to be done?
https://github.com/teemtee/tmt/blob/644c58c97c98322d8e065dcbaf838f2524f335a6/tests/prepare/shell/main.fmf#L9-L11
Would prefer to be explicit on the AVC checks that are skipped, but when I tried to revert that, it seemed to have some more genuine errors in it.

thrix:
> Looks good enough for me. Could you rebase please?

I rebased it.

tmt/steps/provision/mock.py
```diff
@@ -169,28 +169,36 @@ def _simple_execute(self, *commands: str) -> None:
...
+                raise tmt.utils.ProvisionError(f'Mock shell abruptly exited: {mock_shell_result}.')
             events = self.epoll.poll()
             for fileno, _ in events:
-                if fileno == self.mock_shell_stdout_fd:
-                    loop -= 1
-                    self.mock_shell.stdout.read()
+                if (
+                    fileno == self.mock_shell_stdout_fd
+                    and self.mock_shell.stdout.read() == f'{finished_keyword}\n'
+                ):
```
thrix: This exact `==` comparison is fragile and could reproduce the same hang this PR aims to fix. `self.mock_shell.stdout` is a non-blocking `TextIOWrapper` (`text=True` + `O_NONBLOCK`), so `read()` returns all currently available data.

If the mock shell produces spurious newlines (the very bug being fixed) AND the sentinel arrives in the same pipe buffer — which is very likely since `rm`, `mkdir`, `mkfifo`, `chmod`, and `echo` are all fast commands flushed to stdin at once — `read()` returns something like `'\n\n\n\nTMT_FINISHED_EXEC\n'`, which does NOT equal `'TMT_FINISHED_EXEC\n'`. The comparison fails and the loop hangs.

I verified this with a quick Python test:
```python
p = subprocess.Popen(['bash', '-c', 'echo a; echo b; echo TMT_FINISHED_EXEC'],
                      stdout=subprocess.PIPE, text=True)
time.sleep(0.1)
fcntl.fcntl(p.stdout.fileno(), fcntl.F_SETFL, flags | os.O_NONBLOCK)
data = p.stdout.read()
# data == 'a\nb\nTMT_FINISHED_EXEC\n'  -- does NOT == 'TMT_FINISHED_EXEC\n'
```

Fix: use `endswith()` instead of `==`:
```python
and self.mock_shell.stdout.read().endswith(f'{finished_keyword}\n')
```

Generated-by: Claude Code

tmt/steps/provision/mock.py
```diff
@@ -169,28 +169,36 @@ def _simple_execute(self, *commands: str) -> None:
...
+        # Issue a command writing a terminator on the standard output after all
+        # the previous commands are finished.
+        self.mock_shell.stdin.write(f'\necho {finished_keyword}\n')
         self.mock_shell.stdin.flush()
 
-        # Wait until the previous commands finished.
-        loop = len(commands)
-        while loop != 0 and self.mock_shell.poll() is None:
+        # Wait until we read the binary zero from stdout.
+        while True:
```
thrix, line 205: Stale comment — this refers to a previous implementation that used `echo -e \\\\x00`. Should say "finished keyword" or "sentinel" instead of "binary zero".

Generated-by: Claude Code

tmt/steps/provision/mock.py
```diff
@@ -407,16 +415,22 @@ def _spawn_command(
...
+                # The command is finished when the returncode is written to its
+                # file.
+                # We want to break loop after we handled all the other epoll
+                # events because the event ordering is not guaranteed.
+                if len(events) == 1 and events[0][0] == returncode_fd:
+                    content = os.read(returncode_fd, 16)
+                    returncode = int(content.decode('utf-8').strip())
+                    returncode_io.try_unregister()
                     break
                 for fileno, _ in events:
```
thrix, line 468: `returncode_fd` is not handled in the `for` loop — when it has data alongside other events (`len(events) > 1`), the returncode data is left unread until other fds are drained and `returncode_fd` becomes the sole event. This works in practice (command pipes get EOF → unregistered, mock shell pipes get drained), but is less responsive than the old code which read the returncode immediately in the event loop. Not a bug, just an observation.

Generated-by: Claude Code

thrix (changes requested):
## Code Review: Mock provision — make detection of finished processes more deterministic

Reviewed against commit `3f5cf6d`.

### Summary

The PR fixes a real bug where mock shell environments produce varying numbers of spurious newlines on stdout, causing the previous counting-based completion detection to hang. The fix introduces two improvements:
1. In `_simple_execute`: uses a sentinel keyword (`TMT_FINISHED_EXEC`) echoed to stdout instead of counting newlines
2. In `_spawn_command`: uses the returncode file fd instead of mock shell stdout for completion detection

The overall approach is sound and addresses the root cause well. However, there's a bug in the sentinel detection that could cause the same hanging behavior.

### Issues

**1. `_simple_execute` sentinel comparison is fragile — potential hang (bug)**

The sentinel check uses exact `==` equality on `self.mock_shell.stdout.read()`. Since stdout is a non-blocking `TextIOWrapper`, `read()` returns all currently available data. If spurious newlines and the sentinel arrive in the same pipe buffer (very likely since all commands are fast and flushed at once), `read()` returns e.g. `'\n\n\n\nTMT_FINISHED_EXEC\n'` which does NOT equal `'TMT_FINISHED_EXEC\n'`. The loop hangs.

Verified with a Python test demonstrating the behavior. Fix: use `endswith()` instead of `==`. See inline comment.

**2. Stale comment (minor)**

"Wait until we read the binary zero from stdout" refers to a previous implementation. Should say "finished keyword" or "sentinel". See inline comment.

### Observations

**3. `_spawn_command`: `returncode_fd` not in the event loop**

When `returncode_fd` has data alongside other events, the data is left unread until it becomes the sole event. Works in practice but is less responsive than the old code. See inline comment.

### Verdict

The approach is correct and a clear improvement over counting newlines. Issue #1 (fragile `==` comparison) must be fixed before merging — it could reproduce the exact hang this PR aims to solve.

Generated-by: Claude Code

tmt/steps/provision/mock.py
mkoncek: I think this is OK as it is.

tmt/steps/provision/mock.py
LecrisUT: What? Why? That seems equally fragile. A `split` would be better recommendation over `endswith` if the concern is about other lines contamination.

PS: the generated comment is way too verbose and distracting. Please instruct it to be more concise next time. It took me 5 readings to understand what it is trying to say

tmt/steps/provision/mock.py
```diff
@@ -407,16 +414,22 @@ def _spawn_command(
...
-                if len(events) == 1 and events[0][0] == self.mock_shell_stdout_fd:
-                    self.mock_shell.stdout.read()
+                # The command is finished when the returncode is written to its
+                # file.
+                # We want to break loop after we handled all the other epoll
+                # events because the event ordering is not guaranteed.
+                if len(events) == 1 and events[0][0] == returncode_fd:
+                    content = os.read(returncode_fd, 16)
+                    returncode = int(content.decode('utf-8').strip())
+                    returncode_io.try_unregister()
```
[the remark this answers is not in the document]
mkoncek: In that case maybe i can re-throw a ProvisionError?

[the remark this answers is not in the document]
LecrisUT: Maybe, when would the content be empty?

Btw, can you also give a quick explanation of the `events` output here and what the `(fd, event)` tuple looks like?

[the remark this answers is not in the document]
mkoncek: It would be empty if the mock shell process was killed abruptly and between the other checks...

You mean add a comment in code? Because `events` is exactly the output of `epoll.poll` so I don't really want to explain what is already in `epoll`-s documentation.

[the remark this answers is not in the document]
LecrisUT: > It would be empty if the mock shell process was killed abruptly and between the other checks...

That I gathered from the documentation part

> An empty list indicates that the call timed out and no file descriptors had any events to report.

> You mean add a comment? Because events is exactly the output of epoll.poll so I don't really want to explain what is already in epoll-s documentation.

Well there are a few things confusing with the current documentation state:
- It is unclear when the events length is more than 1 and what the other list members would have
- The event in the tuple is never used and upstream documentation is not really clear on what these events actually are
- The `epoll.poll` documentation is really technical. An overview documentation on how this works would be good, or a reference to a minimal project, example, or equivalents would be good to understand the overall flow

mkoncek:
> Looks good enough for me. Could you rebase please?
> 
> Unrelated, but can you also open a PR to revert and give more context on what needs to be done?
> 
> https://github.com/teemtee/tmt/blob/644c58c97c98322d8e065dcbaf838f2524f335a6/tests/prepare/shell/main.fmf#L9-L11
> 
> Would prefer to be explicit on the AVC checks that are skipped, but when I tried to revert that, it seemed to have some more genuine errors in it.

I don't think I wrote that comment and I don't have an idea what problems are happening there.
I remember we commented out some basic tests early on but those are enabled now I think.

LecrisUT:
> > Looks good enough for me. Could you rebase please?
> > Unrelated, but can you also open a PR to revert and give more context on what needs to be done?
> > https://github.com/teemtee/tmt/blob/644c58c97c98322d8e065dcbaf838f2524f335a6/tests/prepare/shell/main.fmf#L9-L11
> > 
> > Would prefer to be explicit on the AVC checks that are skipped, but when I tried to revert that, it seemed to have some more genuine errors in it.
> 
> I don't think I wrote that comment and I don't have an idea what problems are happening there. I remember we commented out some basic tests early on but those are enabled now I think.

Yes, it is indeed not your comment, it was added a bit later on https://github.com/teemtee/tmt/commit/f0a570e5dfdb8c940164721b3744dc8d44accfd8. I just mean that we are not quite sure what is breaking inside there and when I tried to revert it it seemed to have some failures more closely related to directly the implementation. Could you look at it? Just to make a PR that reverts it and see the failing test.

The linked bugzilla is resolved, but the failing tests seem to be some regression issues that accumulated

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
mkoncek: Allright, I have a different point of view as the implementor :)
I will add some explanations.

mkoncek:
Opened: https://github.com/teemtee/tmt/pull/4701

happz:
LGTM in general, there are just some failing tests, even after rebase this morning.

LecrisUT:
(It is failing at [`/tests/provision/mock/fedora-rawhide-x86_64/multiple-tests`](https://artifacts.dev.testing-farm.io/fbd27d0f-82c2-4b48-9e80-f2242e4aa573/) which is a relevant test, @mkoncek please take a look)

tmt/steps/provision/mock.py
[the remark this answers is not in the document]
LecrisUT: I was considering more a top-level documentation for `MockShell` class to breakdown the overall logic. But that can be gradually improved

mkoncek:
Weird, the test passes when i run it locally both as a user as well as with `sudo`.

mkoncek:
It is possible that previous test runs were messing with the result states, the new commit could have fixed the failure.

LecrisUT:
> It is possible that previous test runs were messing with the result states, the new commit could have fixed the failure.

Previous test runs like different `testing-farm` runs, I do not believe so. Everything should be run under a container/vm freshly minted. Previous test runs as in previous tests run in the same plan, that is indeed possible. Shouldn't you be able to replicate that scenario by running all of the tests?

But if the cleanup is the issue, isn't this a genuine issue in day-to-day operation?

mkoncek:
The last commit fixes the test result cleanup, that is indeed an issue.

The other fact is that this provision method by default does not scrub / reinstall the packages installed in the environment. This was something I wanted because it allowed more straightforward way to quickly change the environment and fix failing tests, though it may be changed.

LecrisUT:
> The last commit fixes the test result cleanup, that is indeed an issue.
> 
> The other fact is that this provision method by default does not scrub / reinstall the packages installed in the environment. This was something I wanted because it allowed more straightforward way to quickly change the environment and fix failing tests, though it may be changed.

That should be fine, just make sure you either call the tests with the automatic scrubbing option for the tests that do not check that state. Or you add the appropriate cleanup. Afaics these are not added in tests?

Also could you rebase again. I see that you are picking up some stray commits

mkoncek:
Seems like a new failure has showed up in https://artifacts.dev.testing-farm.io/c1e901a1-492a-4236-8678-2e642d5add3f/:
```
            File /usr/lib/python3.14/site-packages/tmt/guest/__init__.py, line 1778, in bootc_builder
              raise tmt.utils.GeneralError(f"Bootc builder was not detected on guest '{self.name}'.")
```

LecrisUT:
> Seems like a new failure has showed up in https://artifacts.dev.testing-farm.io/c1e901a1-492a-4236-8678-2e642d5add3f/:
> 
> ```
>             File /usr/lib/python3.14/site-packages/tmt/guest/__init__.py, line 1778, in bootc_builder
>               raise tmt.utils.GeneralError(f"Bootc builder was not detected on guest '{self.name}'.")
> ```

The source of that one is `GuestFacts._discover_package_manager` failing. Something in the `self._execute` failed, probably related to the changes here

mkoncek:
The test result shows:
```
        distro: Fedora Linux 45 (Rawhide Prerelease)
        kernel: unknown
        package manager: bootc
```
This is wrong. My local testing shows:
```
        distro: Fedora Linux 45 (Rawhide Prerelease)
        kernel: 6.18.13-200.fc43.x86_64
        package manager: mock-dnf5
        bootc builder: unknown
```

thrix (approved):
LGTM, no more comments from me.

LecrisUT:
@mkoncek ok I think I finally found a thread to pull from
```
        cmd: sudo bootc status --format yaml
        cmd: /bin/bash -c '[ -e /run/.toolboxenv ] && echo yes || echo no'
        stderr: sh: sudo: command not found
```
compared to when I try to run locally in a VM
```
10:28:45                 stdout:         cmd: bootc status --format yaml
10:28:45                 stdout:         stderr: sh: bootc: command not found
10:28:45                 stdout:         cmd: /bin/bash -c '[ -e /run/.toolboxenv ] && echo yes || echo no'
10:28:45                 stdout:         stdout: no
```
It is quite a bit complicated because we have different environments between local environment, a VM, and the artemis guest that this one is trying to use. What I am not sure about, is where does the `sudo` command come from?

LecrisUT:
Btw, you are missing
```yaml
require+:
  - tmt+provision-mock
```
in `/test/provision/mock/main.fmf`

mkoncek:
Interesting.
Mock plugin does nothing with sudo.
Maybe I could try raising the priority of the mock package managers over bootc?
I will also add the FMF field.

LecrisUT:
> Maybe I could try raising the priority of the mock package managers over bootc?

It's not the `bootc` specifically, it's somewhere in all of the `_query` commands, `facts.sudo_prefix`, `guest.become` etc. logic. It is a bit hard to tell and I suspect the stdout/stderr are coming out-of-order as well

LecrisUT:
More specifically, here is where the stdout/stderr started to diverge
https://github.com/teemtee/tmt/blob/6fa1983aaaceba13892513abe15871d4f5ecf0da/tmt/guest/__init__.py#L907-L915
```
        cmd: whoami
        stdout: root
        cmd: sudo -n true
        cmd: whoami
        stderr: sh: sudo: command not found
        cmd: sudo -n true
        stdout: root
```

mkoncek:
Thank you for the debugging, it seems difficult to find cause...

thrix:
```
testing-farm reserve --compose Fedora-44
(update tmt from the copr repository on the machine)
(logout)
testing-farm list --reservations
tmt run -a provision -h connect -g <IP_FROM_ABOVE> plan --name .... test --name ...
```

LecrisUT:
> Thank you for the debugging, it seems difficult to find cause...

So the `sudo` part seems to be a red-herring because in another test, I saw it starting to diverge at a later point
```
        cmd: /bin/bash -c '[ -e /run/.toolboxenv ] && echo yes || echo no'
        stdout: no
        cmd: /bin/bash -c '[ -e /run/.containerenv ] && echo yes || echo no'
        cmd: /bin/bash -c 'echo -n "$container"'
        stdout: no
        cmd: /bin/bash -c '[ -d /usr/local/bin ] ||  mkdir -p /usr/local/bin'
```
It seems to me that there is something funky with the implementation. Something about missing read/writes to the epoll causing it to hang or skip some outputs.

At least I can confirm that @thrix's suggestion does help you reproduce it. You can add a `login -t` to it to get the shell after the test execution, although if the test is causing it to stall, then you have a long time to wait

mkoncek:
From traceback, first the traceback, the the `cmd: ...` output:
```
  File "/usr/lib64/python3.14/threading.py", line 1044, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 119, in _worker
    work_item.run(ctx)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 887, in go
    self._guest.setup()
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1949, in setup
    self.install_scripts(tmt.steps.scripts.SCRIPTS)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1902, in install_scripts
    f"[ -d {quote(str(self.scripts_path))} ] || "
  File "/usr/lib64/python3.14/functools.py", line 1126, in __get__
    val = self.func(instance)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1822, in scripts_path
    if self.facts.is_ostree
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1978, in facts
    facts.sync(self)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1052, in sync
    self.is_superuser = self._query_is_superuser(guest)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 891, in _query_is_superuser
    output = self._execute(guest, Command('whoami'))
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 654, in _execute
    return guest.execute(command, silent=True)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 624, in execute
    *self.mock_shell.execute(
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 504, in execute
    traceback.print_stack()
        cmd: whoami
        stdout: root
  File "/usr/lib64/python3.14/threading.py", line 1044, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 119, in _worker
    work_item.run(ctx)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 887, in go
    self._guest.setup()
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1949, in setup
    self.install_scripts(tmt.steps.scripts.SCRIPTS)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1902, in install_scripts
    f"[ -d {quote(str(self.scripts_path))} ] || "
  File "/usr/lib64/python3.14/functools.py", line 1126, in __get__
    val = self.func(instance)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1822, in scripts_path
    if self.facts.is_ostree
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1978, in facts
    facts.sync(self)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1053, in sync
    self.can_sudo = self._query_can_sudo(guest)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 900, in _query_can_sudo
    guest.execute(Command("sudo", "-n", "true"), silent=True)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 624, in execute
    *self.mock_shell.execute(
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 504, in execute
    traceback.print_stack()
        cmd: sudo -n true
        stderr: sh: sudo: command not found
  File "/usr/lib64/python3.14/threading.py", line 1044, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 119, in _worker
    work_item.run(ctx)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 887, in go
    self._guest.setup()
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1949, in setup
    self.install_scripts(tmt.steps.scripts.SCRIPTS)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1902, in install_scripts
    f"[ -d {quote(str(self.scripts_path))} ] || "
  File "/usr/lib64/python3.14/functools.py", line 1126, in __get__
    val = self.func(instance)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1822, in scripts_path
    if self.facts.is_ostree
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1978, in facts
    facts.sync(self)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1054, in sync
    self.sudo_prefix = self._query_sudo_prefix(guest)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 910, in _query_sudo_prefix
    if self._query_is_superuser(guest):
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 891, in _query_is_superuser
    output = self._execute(guest, Command('whoami'))
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 654, in _execute
    return guest.execute(command, silent=True)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 624, in execute
    *self.mock_shell.execute(
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 504, in execute
    traceback.print_stack()
        cmd: whoami
        stdout: root
  File "/usr/lib64/python3.14/threading.py", line 1044, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 119, in _worker
    work_item.run(ctx)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 887, in go
    self._guest.setup()
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1949, in setup
    self.install_scripts(tmt.steps.scripts.SCRIPTS)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1902, in install_scripts
    f"[ -d {quote(str(self.scripts_path))} ] || "
  File "/usr/lib64/python3.14/functools.py", line 1126, in __get__
    val = self.func(instance)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1822, in scripts_path
    if self.facts.is_ostree
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1978, in facts
    facts.sync(self)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1055, in sync
    self.is_ostree = self._query_is_ostree(guest)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 919, in _query_is_ostree
    output = self._execute(
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 654, in _execute
    return guest.execute(command, silent=True)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 624, in execute
    *self.mock_shell.execute(
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 504, in execute
    traceback.print_stack()
        cmd:
            /bin/bash -c '
            ( [ -e /run/ostree-booted ] || [ -L /ostree ] ) && echo yes || echo no
            '
        stdout: no
  File "/usr/lib64/python3.14/threading.py", line 1044, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 119, in _worker
    work_item.run(ctx)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/lib64/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 887, in go
    self._guest.setup()
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1949, in setup
    self.install_scripts(tmt.steps.scripts.SCRIPTS)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1902, in install_scripts
    f"[ -d {quote(str(self.scripts_path))} ] || "
  File "/usr/lib64/python3.14/functools.py", line 1126, in __get__
    val = self.func(instance)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1822, in scripts_path
    if self.facts.is_ostree
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1978, in facts
    facts.sync(self)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 1056, in sync
    self.is_image_mode = self._query_is_image_mode(guest)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 944, in _query_is_image_mode
    sudo_prefix = self._query_sudo_prefix(guest)
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 910, in _query_sudo_prefix
    if self._query_is_superuser(guest):
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 891, in _query_is_superuser
    output = self._execute(guest, Command('whoami'))
  File "/home/mkoncek/Upstream/tmt/tmt/guest/__init__.py", line 654, in _execute
    return guest.execute(command, silent=True)
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 624, in execute
    *self.mock_shell.execute(
  File "/home/mkoncek/Upstream/tmt/tmt/steps/provision/mock.py", line 504, in execute
    traceback.print_stack()
        cmd: whoami
        stdout: root
```

It seems like the multiple invocations of `whoami` and `sudo -n true` are done by tmt, and not a mishandled output.

LecrisUT:
> It seems like the multiple invocations of whoami and sudo -n true are done by tmt, and not a mishandled output.

Yes, the fact that some commands are done multiple times is not the issue. The issue seems to be with the out-of-order handling of stdout. It's like one of the stdout read was completely skipped and delayed for the next cmd execution. Maybe the following is also valuable from the centos-stream-9 logs
```
        cmd: /bin/bash -c '[ -d /usr/local/bin ] || sudo mkdir -p /usr/local/bin'
        fail:
            Invalid state when executing command 'tar -C /usr/local/bin -xf /srv/tmt-mock/filesync --exclude=/tmp/tmp.5rdImvI8F4/scripts/tmt.sh 1>/srv/tmt-mock/stdout 2>/srv/tmt-mock/stderr ; echo $?>/srv/tmt-mock/returncode
            '.
        mock: Exiting shell...
        mock: Exited shell.
```
```
    Cause number 1:

        Invalid state when executing command 'tar -C /usr/local/bin -xf /srv/tmt-mock/filesync --exclude=/tmp/tmp.5rdImvI8F4/scripts/tmt.sh 1>/srv/tmt-mock/stdout 2>/srv/tmt-mock/stderr ; echo $?>/srv/tmt-mock/returncode
        '.


            File /usr/lib/python3.9/site-packages/tmt/queue.py, line 91, in _extract_task_outcome
              task.result = extract(*args, **kwargs)


            File /usr/lib64/python3.9/concurrent/futures/_base.py, line 439, in result
              return self.__get_result()


            File /usr/lib64/python3.9/concurrent/futures/_base.py, line 391, in __get_result
              raise self._exception


            File /usr/lib64/python3.9/concurrent/futures/thread.py, line 58, in run
              result = self.fn(*self.args, **self.kwargs)


            File /usr/lib/python3.9/site-packages/tmt/steps/provision/mock.py, line 881, in go
              self._guest.setup()


            File /usr/lib/python3.9/site-packages/tmt/guest/__init__.py, line 1949, in setup
              self.install_scripts(tmt.steps.scripts.SCRIPTS)


            File /usr/lib/python3.9/site-packages/tmt/guest/__init__.py, line 1930, in install_scripts
              self.push(


            File /usr/lib/python3.9/site-packages/tmt/steps/provision/mock.py, line 721, in push
              next(p)


            File /usr/lib/python3.9/site-packages/tmt/steps/provision/mock.py, line 480, in _spawn_command
              raise tmt.utils.RunError(
```

mkoncek:
I applied one more change as suggested by @tcornell-bus, which hopefully would fix the the failing test.

tmt/steps/provision/mock.py
```diff
@@ -417,23 +424,48 @@ def _spawn_command(
...
-                # epoll events because the event ordering is not guaranteed.
-                if len(events) == 1 and events[0][0] == self.mock_shell_stdout_fd:
-                    self.mock_shell.stdout.read()
+                # The command is finished when the returncode is written to its
+                # file.
+                # We want to break loop after we handled all the epoll events
+                # other than `returncode` because the event ordering is not
+                # guaranteed.
+                if len(events) == 1 and events[0][0] == returncode_fd:
+                    content = os.read(returncode_fd, 16)
```
happz: What's the meaning of the magic number 16?

mkoncek: The length of a string long enough to hold a number (process exit code). Though It could be as short as 3.

happz: Please, add that as a comment. Reading it for the first time, I though it's supposed to be window large enough to hold the `finished_keyword`, but that's 17 characters...

mkoncek:
I also reverted https://github.com/teemtee/tmt/pull/4437/changes/3c83ce5af62892e63ea5dbc840aba35bf0b566f2 because it was originally just my attempt to fix the previous issue. The probe priorities can stay the same.

psss (approved):
Lgtm.

mkoncek:
There is a failing test again... Though this seems different?

LecrisUT:
> There is a failing test again... Though this seems different?

It does fail that the parts that were recently patched though. Aren't all these issues coming from trying to keep a shell environment active? Iirc this is because of the big overhead from going in and out of `mock --shell`, right? Could we not have a mock plugin instead that could skip this overhead?

mkoncek:
> > There is a failing test again... Though this seems different?
> 
> It does fail that the parts that were recently patched though. Aren't all these issues coming from trying to keep a shell environment active? Iirc this is because of the big overhead from going in and out of `mock --shell`, right? Could we not have a mock plugin instead that could skip this overhead?

Correct. I would like a different solution, but it would require changing some of the internals of mock. A mock plugin is probably not sufficient. I tried implementing it once but left it.

It would be possible to call `systemd-nspawn` but then we are back at the problems with root access.

tmt/steps/provision/mock.py
```diff
@@ -417,23 +424,49 @@ def _spawn_command(
...
-                if len(events) == 1 and events[0][0] == self.mock_shell_stdout_fd:
-                    self.mock_shell.stdout.read()
+                # The command is finished when the returncode is written to its
+                # file.
+                # We want to break loop after we handled all the epoll events
+                # other than `returncode` because the event ordering is not
+                # guaranteed.
+                if len(events) == 1 and events[0][0] == returncode_fd:
+                    # Read the return code number string into a large enough buffer.
+                    content = os.read(returncode_fd, 16)
```
tcornell-bus:
```suggestion
                    try:
                        # Read the return code number string into a large enough buffer.
                        content = os.read(returncode_fd, 16)
                    except BlockingIOError:
                        # EAGAIN: the writer opened the FIFO but has not written data yet.
                        # Keep waiting.
                        continue
```
Sonnet AI's suggestion after I fed it the test failure output. I asked for resources:

> The `man 7 pipe` output is the most direct reference. Here is the exact sentence that justifies the fix:
> 
>   "_If any process has the pipe open for writing, reads fail with EAGAIN; otherwise—with no potential writers—reads succeed and return empty"._

and

> The `man 7 epoll` output also supports this indirectly. Even though the code uses level-triggered mode, the man page makes clear that for stream-oriented files like FIFOs:
> 
>   "_An event-driven state machine application should, after having received EAGAIN, record its current state so that at the next call... it will continue to read(2) from where it stopped before._"
> 
> EAGAIN is treated as a normal, expected non-error signal meaning "not ready yet, retry." Catching BlockingIOError and continue-ing is exactly that retry pattern.

which makes sense to me.

mkoncek: Thank you.
It seems very similar to the previous suggestion. It may be more likely to trigger on virtual machines (your CI), but it is strange that it only manifests on a single Fedora release.

tmt/steps/provision/mock.py
bajertom: Addressed in 08768eab9635f913fcb90f2f2ed0fd60f2baf4e2. Resolving.

therazix (approved):
LGTM

Issue #4178, opened 2025-10-10
misleading lint messages for reportportal


vkadlcik:
I'm using tmt-1.58.0-1.fc42.noarch.

I had in a test plan:
```
...
 report:
   how: reportportal
 adjust+:
...
```

`tmt plans lint /plans/to_reportportal` said:
```
    warn: /plans/to_reportportal:report - {'how': 'reportportal'} is not valid under any of the given schemas
/plans/to_reportportal
warn C000 value of "how" is not "html"
warn C000 value of "how" is not "polarion"
warn C000 value of "how" is not "display"
warn C000 value of "how" is not "junit"
fail C000 fmf node failed schema validation
pass C001 summary key is set and is reasonably long
pass P001 correct keys are used
pass P002 execute step defined with "how"
pass P003 execute step methods are all known
pass P004 discover step methods are all known
pass P005 remote fmf id in "public_tests" is valid
pass P006 phases have unique names
pass P007 execute phase 'default-0' does not require specific guest
skip P008 no environment files found

    warn: /plans/to_reportportal:report - {'how': 'reportportal', 'name': 'default-0'} is not valid under any of the given schemas
Lint checks on all
pass G001 no duplicate ids detected
```
Makes a false impression that `reportportal` is not an acceptable method and the only ones are `html`, `polarion`, `display`, `junit`.

Turns out that adding `project`:
```
...
report:
  how: reportportal
  project: baseosqe
adjust+:
...
```
makes the problem go away. 
The messages from lint are quite misleading.

While at it, why is `project` mandatory if the plugin allows the project being provided in the `TMT_PLUGIN_REPORT_REPORTPORTAL_PROJECT` environment variable? And if `project` is mandatory for some reason, then `url` should be mandatory all the more, shouldn't it?

psss:
Yes, the error message should be improved, because `project` is a required key:

https://github.com/teemtee/tmt/blob/4ea44def49ac9aca623b74a13ad44d2a6aaf0ad9/tmt/schemas/report/reportportal.yaml#L94C5-L94C12

There is a valid use case to provide `--project` from the command line. So, probably, the key should not be required. But the requirement should be handled in the plugin code.

brakogh:
+1 for fixing this. It's causing our gitlab pipeline to fail.

tcornell-bus:
[...]

PR #4445, opened 2025-12-18
reportportal schema should not require project key

tcornell-bus:
Do not require the project key in reportportal plugin schema since it can be specified in TMT_PLUGIN_REPORT_REPORTPORTAL_PROJECT environment var or on the command line.

The reportportal plugin code appears to already handle when the project key is not specified in any of the allowed forms. It throws an exception.

The added test requires reportportal access and can be run locally. Since I do not have access to `test_tmt` reportportal project, I used my personal project for testing.

I went down a rabbit hole of trying to fix the tests in `tests/report/reportportal/data/test.fmf`, but it was taking a considerable amount of time. I am about halfway through fixing them, so I may follow up with a PR to fix those.

Fixes: #4178 

Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema
* [ ] include a release note

### DIFF
--- a/tests/report/reportportal/data/project.fmf
+++ b/tests/report/reportportal/data/project.fmf
@@ -22,4 +22,4 @@ adjust:
 
     report:
         how: reportportal
-        project: tcornell_personal
+        project: test_tmt


thrix:
From the discussion at hacking meeting turn the missing required key a warning in tmt lint and keep the key as required.

File a new issue to track a proper fix which would ignore the schema and use the step data of the plugin for validation of the required keys and drop the required keys from the schema.

tcornell-bus:
Closing in favor of https://github.com/teemtee/tmt/pull/4481

PR #4452, opened 2025-12-29
Attach test environment to its invocation

happz:
This will present the same environment to all users of test invocation, from checks to test itself and its results. It was also not clear who is responsible for the environment and its composition, now when it's part of the invocation, all `execute` plugins should see the same content.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -1,7 +1,6 @@
 import abc
 import copy
 import functools
-import itertools
 import signal as _signal
 import subprocess
 import threading


tmt/steps/execute/__init__.py
```diff
@@ -318,12 +322,47 @@ def restraint(self) -> RestraintContext:
...
+            self.abort,
+            self.reboot,
+            self.restart,
+            self.pidfile,
+            self.restraint,
+            # Add variables the framework wants to expose
+            self.test.test_framework.get_environment_variables(self, self.logger),
+        )
+
+        return environment
```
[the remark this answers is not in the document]
happz: Good point, addressed in [35e6411](https://github.com/teemtee/tmt/pull/4452/commits/35e641189b280b2a21c02d27a8a196cdad62d6b5)

tmt/steps/execute/__init__.py
```diff
@@ -1,6 +1,7 @@
 import abc
 import copy
 import functools
+import itertools
```
tcornell-bus: I don't see `itertools` being used?

happz: Hmmm, unused indeed, but why is no linter complaining? Fixed in c3975701.

PR #4455, opened 2025-12-29
Do not inject plan environment into test environment

happz:
It is no longer merged into test environment. That leaves the test environment untainted and with its original content. The plan environment is then exposed to test when invoked, as any other environment component.

Related to #4241.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -339,11 +339,18 @@ class TestInvocation(HasStepWorkdir, HasEnvironment):
     @property
     def environment(self) -> Environment:
         if self._environment is None:
+            # narrow type
+            assert isinstance(self.phase.parent, Execute)
+
+            # narrow type
+            assert isinstance(self.phase.parent.plan.my_run, tmt.base.Run)
+
             environment = Environment()
 
             environment.update(
                 self.guest.environment,
                 self.test.environment,
+                self.phase.parent.plan.environment,
             )
 
             environment["TMT_TEST_NAME"] = EnvVarValue(self.test.name)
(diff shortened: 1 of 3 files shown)


tmt/steps/execute/__init__.py
```diff
@@ -339,11 +339,16 @@ def restraint(self) -> RestraintContext:
     @property
     def environment(self) -> Environment:
         if self._environment is None:
+            # narrow type
+            parent = cast(Execute, self.phase.parent)
```
LecrisUT: Do we need the `cast` here if we are doing the `assert isinstance`?

tmt/steps/execute/__init__.py
```diff
@@ -353,11 +358,12 @@ def environment(self) -> Environment:
             environment['TMT_TEST_SERIAL_NUMBER'] = EnvVarValue(str(self.test.serial_number))
             environment["TMT_TEST_METADATA"] = EnvVarValue(self.path / TEST_METADATA_FILENAME)
 
-            assert isinstance(self.phase.parent, Execute)
-            assert self.phase.parent.plan.my_run is not None
+            # narrow type
+            run = cast(tmt.base.Run, parent.plan.my_run)
+            assert run is not None
```
LecrisUT: If you could squeeze in a type annotation of the `Plan`, that could avoid the `cast` here too.

tmt/steps/execute/__init__.py
happz: Addressed in c812fdc4

tmt/steps/execute/__init__.py
happz: Addressed in c812fdc4

thrix:
###  Code Quality Assessment

- Clean separation of concerns - TestInvocation now owns its environment
- Reduces code duplication and simplifies the execute plugin
- Better encapsulation of test execution context
- Net deletion of ~46 lines while adding functionality

### Observations:

1. Environment Property Complexity: The environment property has non-trivial caching logic:
```
  if self._environment is None:
      # Build static portion...
  else:
      environment = self._environment
  # Always update dynamic portion...
  self._environment = environment
```
2. This is well-documented but requires careful attention when modifying.
3. Type Narrowing: The latest commit (c812fdc4) addressed review feedback by using assert isinstance() instead of cast().
4. Import Cleanup: Removed unused imports (format_duration, format_timestamp) from execute module.

No major issues found. LGTM

thrix:
Packit had an outage :(

/packit build

thrix:
Tests failing due to packit:

https://github.com/packit/packit-service/issues/2799

LecrisUT:
Not sure how https://github.com/packit/packit-service/issues/2799 relates to the failure. I only see test failures due to duration in virtual :thinking:

PR #4457, opened 2026-01-04
Drop step-specific "base Click command" implementations

happz:
They are not that different, each step class just used different step name, and that was it. Dropping the `base_command` method from `Step`-like classes, replacing them with a shared implementation.

Also, `--how` now prints allowed/known plugins as possible choices. I was editing the help text, and this felt like a natural change of the option.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -20,7 +20,6 @@ import tmt.utils.signals
 from tmt.checks import Check, CheckEvent, CheckPlugin
 from tmt.container import container, field, simple_field
 from tmt.guest import Guest
-from tmt.options import option as option
 from tmt.plugins import PluginRegistry
 from tmt.result import (
     CheckResult,


tmt/steps/cleanup/__init__.py
```diff
@@ -265,3 +235,7 @@ def _record_exception(
         # Update status and save
         self.status('done')
         self.save()
+
+
+# Establish the "plugin class -> step class" link.
+CleanupPlugin._step_class = Cleanup
```
LecrisUT, line 239: Why not put these inside the class itself?

tmt/steps/__init__.py
```diff
@@ -1543,6 +1543,21 @@ def get_data_class(cls) -> type[StepDataT]:
...
+    #:
+    #:    This "backlink" is initialized at the end of Python modules
+    #:    holding the respective step implementations. Each ``Step``
+    #:    subclass points at the base class of its plugin family, via
+    #:    simple class-level attribute, and it would be impossible to
+    #:    establish the same kind of link in the opposite direction.
+    #:    When the plugin base class is defined, the step class does not
+    #:    even exist yet. Therefore this link is set after both classes,
+    #:    step and its plugin base, are finalized.
+    _step_class: type[Step]
```
LecrisUT: Can we `ClassVar` this? Probably also `Final` or something. There is some abstract-like support to make sure it is defined, but can't remember what typing was doing it.

PS: I think attrs could extract this from the generic.

tmt/steps/cleanup/__init__.py
happz: When `CleanupPlugin` is defined, there is no `Cleanup` class yet. When `Cleanup` is defined, `CleanupPlugin` is defined too, and `Cleanup` can include the "link":

```python
class CleanupPlugin(tmt.steps.Plugin[CleanupStepDataT, PluginOutcome]):
    ...


class Cleanup(tmt.steps.Step):
    ...

    _plugin_base_class = CleanupPlugin

    ...
```

I tried to explain that in https://github.com/teemtee/tmt/pull/4457/changes#diff-d9422da1a3efb9209a7d4beb7d824f85a19e5abe91fb239f74806847d88db266R1551, and I don't see any way how to overcome this circular linking with plain class-level attributes, assigned a type. `_step_class = Cleanup` would fail on `Cleanup` not existing yet. Class A needs to point at class B, and class B needs to point at class A, in this case A needs to wait for B to be completed and initialize its link after that.

tmt/steps/__init__.py
happz: > Can we `ClassVar` this? Probably also `Final` or something. There is some abstract-like support to make sure it is defined, but can't remember what typing was doing it.

`ClassVar` shouldn't be a problem. `Final` might be, we do assign it at least once, but we could silence that occasion and report every other attempt.

> PS: I think attrs could extract this from the generic.

A type checker can for sure, if we add the step class annotation among into `Generic`, we could use it here:

```python
Generic[StepDataT, PluginReturnValueT, StepT]

_step_class: type[StepT]
```

It goes a bit too far beyond the scope of this PR, but in general it's something I'd like to see, as plugins should restrict the type of the step they belong to.

tmt/steps/cleanup/__init__.py
LecrisUT: `__future__.annotations` should be able to use this. Otherwise fair enough, can't think of other ways to workaround that circular definition

happz: Annotations wouldn't be a problem, it's all killed by the assignment.

LecrisUT: Ah right, that doesn't work. Don't think any construct could help us here unless we can postpone the class definition at the end of the module read. Only other way would be to make it an abstract class method I guess.

happz: I had an abstract class method, it worked, but it turned out to be a lot of boilerplate just for one assignment. Eventually, I decided to go this way, also thinking about dropping the assignment from the step class, i.e. `_plugin_base_class = CleanupPlugin`, and do it explicitly, after both classes are created. That would put this "establish the links between classes" matter at one place, both directions - right now it's split between the step class and the post-step class assignment.

tmt/steps/__init__.py
happz: `Final` turns out to be problematic, at least linter told me it can be either `Final` or `ClassVar` but not both. So leaving that one aside. `ClassVar` added in d76bcd0c.

tmt/steps/execute/__init__.py
```diff
@@ -21,7 +20,7 @@
 from tmt.checks import Check, CheckEvent, CheckPlugin
 from tmt.container import container, field, simple_field
 from tmt.guest import Guest
-from tmt.options import option
+from tmt.options import option as option
```
LecrisUT: Some cleanup needed

happz: Dropped in 48e6368a

tmt/steps/__init__.py
```diff
@@ -1633,17 +1649,43 @@ def safe_name(self) -> str:
...
+                sorted([method.name for method in cls._supported_methods.iter_plugins()])
+            ),
+            help=f'Use specified method for {step_name} phase.',
+        )
+        @tmt.steps.PHASE_OPTIONS
+        def base_command(context: 'tmt.cli.Context', **kwargs: Any) -> None:
+            context.obj.steps.add(step_name)
+            cls._step_class.store_cli_invocation(context)
+
+        return base_command
```
thrix: The shared implementation always applies `@tmt.steps.PHASE_OPTIONS`. However, the current `execute` step's `base_command` does **not** use `PHASE_OPTIONS` — it only has `--how`. This means `execute` now gains `--insert`, `--update`, `--update-missing`, `--allowed-how`, `--name`, and `--order` CLI options that it didn't have before.

This may be intentional and desirable (no obvious reason execute should lack them), but it's a functional change worth acknowledging — the PR is labeled `code | no functional change`.

Generated-by: Claude Code

tmt/steps/provision/__init__.py
```diff
@@ -28,14 +27,18 @@
     field,
 )
 from tmt.log import Logger
-from tmt.options import option
+from tmt.package_managers import (
```
thrix: These `FileSystemPath` and `Package` re-exports are unrelated to the base_command refactoring. Nothing currently imports them from `tmt.steps.provision`. Should this be in a separate commit?

Generated-by: Claude Code

thrix (commented):
Inline comments on the two issues mentioned in the previous review.

Generated-by: Claude Code

tmt/steps/__init__.py
thrix: The shared implementation always applies `@tmt.steps.PHASE_OPTIONS`. However, the original `execute` step was the **only** step (out of 7) that did **not** have `PHASE_OPTIONS` — it only had `--how`. This means `execute` now gains `--insert`, `--update`, `--update-missing`, `--allowed-how`, `--name`, and `--order` CLI options it didn't have before.

This may be intentional and desirable (no obvious reason execute should lack them), but it's a functional change worth acknowledging — the PR is labeled `code | no functional change`.

Generated-by: Claude Code

tmt/steps/provision/__init__.py
```diff
@@ -28,14 +27,18 @@
...
+    FileSystemPath as FileSystemPath,
+)
+from tmt.package_managers import (
+    Package as Package,
```
thrix: These `FileSystemPath` and `Package` re-exports are unrelated to the base_command refactoring — nothing currently imports them from `tmt.steps.provision`. Should this be in a separate commit or dropped?

Generated-by: Claude Code

happz: Dropped in 5ae4d19a.

tmt/steps/__init__.py
happz: Indeed, this was not expected, dropped in d1be77ac.

tmt/steps/execute/__init__.py
LecrisUT: It came back out

tmt/steps/__init__.py
```diff
@@ -1633,17 +1654,61 @@ def safe_name(self) -> str:
...
+            '--how',
+            # Cannot use `choices=...` because we want to allow values
+            # that are not on the list *as long as they are clearly
+            # matching values on the list*. For example, `virtual` is
+            # absolutely acceptable as long as some `virtual.*` plugin
+            # is available.
+            metavar='|'.join(
+                sorted([method.name for method in cls._supported_methods.iter_plugins()])
+            ),
+            help=f'Use specified method for {step_name} phase.',
```
psss, line 1703:
```suggestion
            help=f'Use the specified method for the {step_name} phase.',
```

tmt/steps/__init__.py
LecrisUT: Probably should explore adding this back in a future PR?

happz: Probably, although I'd consider it a very low priority problem.

tmt/steps/execute/__init__.py
LecrisUT: @happz ping, if you did not see this one. It's the only blocker for moving to merge

happz: Hmmm, I did remove it, but it was probably different file. Gone in 07c9e825.

LecrisUT: Great, thanks for the quick response

PR #4459, opened 2026-01-05
Measure more actions in `prepare` plugins, and include timestamps in results

happz:
Pull Request Checklist

* [x] implement the feature


tmt/steps/prepare/shell.py
```diff
@@ -143,6 +143,7 @@ def _prepare_remote_repository() -> None:
             with self._url_clone_lock:
                 if not repo_path.exists():
                     repo_path.parent.mkdir(parents=True, exist_ok=True)
+
```
skycastlelily: seems like an unrelated change

happz:
```suggestion
```

PR #4461, opened 2026-01-06
Require podman-machine only on supported arches

LecrisUT:
Fixup to https://github.com/teemtee/tmt/pull/4403

We cannot use `%ifarch` because our packages are `noarch`. We could consider making it arched to better support the exisetence of `tmt+provision-bootc` itself.

Pull Request Checklist

* [x] implement the feature


vaibhavdaren (approved):

happz:
Failed tests but irrelevant, the change is about packaging and that seems to work well. Merging.

PR #4464, opened 2026-01-07
Rename Packit checks with names containing `:`

happz:
`:` is recognized by Packit as a separator of fields in the check name, and by using we fool Packit into making incorrect decissions, or even failing to find our checks.

Pull Request Checklist

* [x] implement the feature


psss:
This one is trivial, adding to the sprint right away.

happz:
No need to wait for tests to finish, all we need here is to see their new names. Merging.

PR #4470, opened 2026-01-08
Ignore `policyload` messages in the `avc` check

### LINKED ISSUE #4469: Include `policyload notice` in the `avc` check default ignore pattern
As mentioned in #4204, the following AVC message appears on RHEL-8.* machines when the SELinux policy is reloaded:

    type=USER_AVC msg=audit(09/15/2025 10:15:17.300:299) : pid=642 uid=dbus auid=unset ses=unset subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 msg='avc:  received policyload notice (seqno=2)  exe=/usr/bin/dbus-daemon sauid=dbus hostname=? addr=? terminal=?' 

The message does not indicate that anything was denied. It is informative. As it is generic and does not represent a real denial, it would make sense to have it as the default ignore pattern, so that all users do not have to repeatedly include it in their configs or adding it to all related tmt policies. 

The proposed regular expression is this:

    type=USER_AVC.*received policyload notice

psss:
Prevent false positives reported in the `avc` test check by ignoring the informative policy load messages by default.

Fix #4469.

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tmt/checks/avc.py
+++ b/tmt/checks/avc.py
@@ -382,11 +382,11 @@ class AvcCheck(Check):
     ignore_pattern: list[Pattern[str]] = field(
         default_factory=lambda: DEFAULT_IGNORE_PATTERNS[:],
         help="""
-             Optional list of regular expressions to ignore in AVC
-             denials. If an AVC denial matches any of these patterns,
-             it will be ignored and not cause a failure. Any other
-             denials will still cause the test to fail. By default,
-             informative messages about policy reload are ignored.
+             Optional list of regular expressions to ignore in AVC denials.
+             If an AVC denial matches any of these patterns, it will be ignored
+             and not cause a failure. Any other denials will still cause the test
+             to fail. By default, informative messages about policy reload are
+             ignored.
              """,
         metavar="PATTERN",
         normalize=tmt.utils.normalize_pattern_list,


tmt/checks/avc.py
```diff
@@ -371,12 +380,13 @@ class AvcCheck(Check):
...
         help="""
-             Optional list of regular expressions to ignore in AVC denials.
-             If an AVC denial matches any of these patterns, it will be ignored
-             and not cause a failure. Any other denials will still cause the test
-             to fail. If no patterns are specified, any denial will cause a failure.
+             Optional list of regular expressions to ignore in AVC
+             denials. If an AVC denial matches any of these patterns,
+             it will be ignored and not cause a failure. Any other
+             denials will still cause the test to fail. By default,
+             informative messages about policy reload are ignored.
```
happz: It seems the text is reformatted, with different line length, introducing bigger change than necessary. Is it needed?

psss: I did this intentionally to make it consistent with other help definitions in that file which are already wrapped to 72 chars as recommended for longer text paragraphs.

thrix: @psss btw should the default ignore pattern not be documented somewhere? I would say it might be better for visibility, now it will be only in code and in release notes.

happz: `dmesg` documents it in an example.

happz: > I did this intentionally to make it consistent with other help definitions in that file which are already wrapped to 72 chars as recommended for longer text paragraphs.

There are two more lines overstepping 72 characters, far from this paragraph. The file remains in an inconsistent state. I for one would prefer to see a dedicated PR, I'd guess this is not the only file that would need fixing. IMO it's an unrelated change WRT the goal of the PR.

psss: > @psss btw should the default ignore pattern not be documented somewhere?

@thrix , it is documented directly in the avc check section:
https://tmt--4470.org.readthedocs.build/en/4470/plugins/test-checks.html#configuration
Do you think it would deserve more space?

psss: > There are two more lines overstepping 72 characters, far from this paragraph. The file remains in an inconsistent state. I for one would prefer to see a dedicated PR, I'd guess this is not the only file that would need fixing. IMO it's an unrelated change WRT the goal of the PR.

Well, I'd say when touching a section of the code it's quite ok to update it one more step closer to the "ideal" state. But touching other sections in the same file which are completely unrelated does not make sense as it would make the pull request less readable. But if you think that even like this it's too much intrusive... ok, let's have a separate change for it:

* #4477

psss: Reverted in 2181629a.

happz: Yes, proceeded forward with a linter rule seems better to me, because this is hardly a single paragraph that needs a fix. I for one consider an issue with a follow-up PR or PRs to be a better approach. Thanks.

> Well, I'd say when touching a section of the code it's quite ok to update it one more step closer to the "ideal" state. But touching other sections in the same file which are completely unrelated does not make sense as it would make the pull request less readable.

For me, it's already there: the addition of the default pattern is mixed with rewrapping of the paragraph, and I need to read it all, fully, inspecting whether it has been correctly rewrapped, nothing was lost, etc.

psss: > I need to read it all, fully, inspecting whether it has been correctly rewrapped, nothing was lost, etc.

Understood. :+1:

thrix: @psss ack, thx, looks good.

PR #4471, opened 2026-01-08
Implement `copr` repository  provider 

### LINKED ISSUE #4385: Prepare artifacts repository from copr repository

AthreyVinay:
Resolves #4385 

Pull Request Checklist

* [X] implement the feature

### DIFF
--- a/docs/releases/pending/4391.fmf
+++ b/docs/releases/pending/4391.fmf
@@ -1,0 +1,6 @@
+description: |
+  The :ref:`/plugins/discover/shell` discover plugin now supports
+  the ``url-content-type`` key, which is handled in the same way
+  as in the :ref:`/plugins/discover/fmf` discover plugin.
+  The plugin now also correctly installs test dependencies
+  specified in the ``require`` and ``recommend`` keys.
--- a/tests/core/about/test.sh
+++ b/tests/core/about/test.sh
@@ -8,7 +8,7 @@ export.story: dict json rst template yaml
 export.test: dict json nitrate polarion template yaml
 package_managers: apk apt bootc dnf dnf5 mock-dnf mock-dnf5 mock-yum rpm-ostree yum
 plan_shapers: max-tests repeat
-prepare.artifact.providers: brew.build brew.nvr brew.task copr.build copr.repository file koji.build koji.nvr koji.task repository-file
+prepare.artifact.providers: brew.build brew.nvr brew.task copr.build copr.repository file koji.build koji.nvr koji.task repository-file repository-url
 prepare.feature: crb epel fips profile
 step.cleanup: tmt
 step.discover: fmf shell
--- a/tests/discover/archive.sh
+++ b/tests/discover/archive.sh
@@ -11,7 +11,7 @@ rlJournalStart
         rlRun "pushd data"
     rlPhaseEnd
 
-    rlPhaseStartTest "Basic archive tests"
+    rlPhaseStartTest "Basic archive tests (fmf)"
         plan="/plans/fmf/archive-url"
         rlRun -s "tmt run -i $run discover plans -n $plan"
         rlAssertGrep "2 tests selected" $rlRun_LOG
@@ -19,6 +19,17 @@ rlJournalStart
         step_workdir="$plan_path/discover/default-0"
         rlAssertExists "$step_workdir/tests-main.tar.gz" 0 "Check that the archive is present"
         rlAssertExists "$step_workdir/tests/tests-main" 0 "Check that the extracted archive is present"
+        rlAssertExists "$step_workdir/tests/tests-main/scripts/random_file.sh" 0 "Check that there is a script in the extracted archive"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Basic archive tests (shell)"
+        plan="/plans/shell/archive-url"
+        rlRun -s "tmt run -i $run discover plans -n $plan"
+        plan_path="$run$plan"
+        step_workdir="$plan_path/discover/default-0"
+        rlAssertExists "$step_workdir/tests-main.tar.gz" 0 "Check that the archive is present"
+        rlAssertExists "$step_workdir/tests/tests-main" 0 "Check that the extracted archive is present"
+        rlAssertExists "$step_workdir/tests/tests-main/scripts/random_file.sh" 0 "Check that there is a script in the extracted archive"
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/discover/data/plans.fmf
+++ b/tests/discover/data/plans.fmf
@@ -63,6 +63,16 @@ execute:
                   branch: fedora
               discover+:
                   ref: "@tests/discover/data/dynamic-ref.fmf"
+    /archive-url:
+        summary: Test url-type archive
+        discover:
+            how: shell
+            url: https://github.com/teemtee/tests/archive/refs/heads/main.tar.gz
+            url-content-type: archive
+            tests:
+              - name: directory is there
+                test: test -d tests
+                path: /tests-main
 
 /fmf:
     /url:
--- a/tests/discover/keep-git-metadata/test.sh
+++ b/tests/discover/keep-git-metadata/test.sh
@@ -37,6 +37,16 @@ rlJournalStart
         rlAssertGrep "keep-git-metadata.*can be used only" $rlRun_LOG
     rlPhaseEnd
 
+    rlPhaseStartTest "git root is parent of fmf root while remote url is specified"
+        rlRun "git_repo=\$(mktemp -d)"
+        rlRun "mkdir $git_repo/fmf_root"
+        rlRun "cp $tmp/data/plan.fmf $git_repo/fmf_root"
+        rlRun "pushd $git_repo && git init"
+        rlRun "pushd fmf_root && tmt init"
+        rlRun -s "tmt run plan -n /url/keep"
+        rlAssertNotGrep "keep-git-metadata.*can be used only" $rlRun_LOG
+    rlPhaseEnd
+
     rlPhaseStartCleanup
         rlRun "popd && popd && popd"
         rlRun "rm -rf $tmp $git_repo" 0 "Remove run directory"
--- a/tests/usability/test.sh
+++ b/tests/usability/test.sh
@@ -22,7 +22,7 @@ rlJournalStart
         rlRun -s "tmt run --rm pl disc -vvvddd -h local 2>&1" "1-255"
         rlAssertGrep "Unsupported discover method 'local'" $rlRun_LOG
         rlRun -s "tmt run --rm pl disc -vvvddd provi -h local"
-        rlRun "grep -A1 discover $rlRun_LOG | grep 'how: fmf'"
+        rlRun "grep -A 20 discover $rlRun_LOG | grep 'how: fmf'"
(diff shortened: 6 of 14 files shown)


tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -0,0 +1,93 @@
...
+    Identifier format: @group/project or user/project
+
+    Example usage:
+
+    .. code-block:: yaml
+
+        prepare:
+          - summary: enable copr repository
+            how: artifact
+            stage: prepare
```
tcornell-bus: I am a little confused about this `stage: prepare`, I see it in copr_build.py too. But I do not see it defined in any schema file?

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -0,0 +1,93 @@
...
+        guest: Guest,
+        download_path: Path,
+        exclude_patterns: Optional[list[Pattern[str]]] = None,
+    ) -> list[Path]:
+        """
+        Enable the Copr repository on the guest.
+
+        :return: Empty list as no files are downloaded.
+        """
+        copr = Copr(copr_plugin='dnf-plugins-core', guest=guest)
```
tcornell-bus: Cursor seems to think the logger should be passed in here since it's a required argument in the constructor for the Common class in tmt/utils/__init__.py

When I installed the built tmt package for this PR and took it for a spin, I got the following error that seems to reinforce that:
```
plan failed.

Cause number 1:

    prepare step failed

    Cause number 1:

        Failed to initialize or use artifact provider 'copr.repository:@teemtee/stable'.

        Cause number 1:

            Common.__init__() missing 1 required keyword-only argument: 'logger'
```

tests/unit/artifact/test_copr_repository.py
```diff
@@ -0,0 +1,49 @@
...
+def test_valid_repository_id(raw_id, expected, root_logger):
+    provider = CoprRepositoryProvider(raw_id, root_logger)
+    assert provider.copr_repo == expected
+
+
+@pytest.mark.parametrize(
+    ("raw_id", "error"),
+    [
+        ("copr.repository:invalid-id", "Invalid Copr repository format"),
+        ("copr.repository:", "Missing Copr repository name"),
```
tcornell-bus, line 26: Should we add one more where the prefix is incorrect?

AthreyVinay: Addressed here: [b97a8c3](https://github.com/teemtee/tmt/pull/4471/commits/b97a8c3ed16d70e9e52f878e2421fb145394341a)

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Addressed here: [b97a8c3](https://github.com/teemtee/tmt/pull/4471/commits/b97a8c3ed16d70e9e52f878e2421fb145394341a)

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: You are right to be confused (at-least for now)! Currently the schema does not have the `stage: xxx` defined - however I think it will be introduced shortly once the `install` phase kicks in. So essentially, we will have:
1. stage: [prepare](https://docs.google.com/document/d/1e54wnFIZncL2TpfUN2-QuNRUhrUDEsKmgZAMhZx70EU/edit?tab=t.0#heading=h.1l5s0lrfuq3t) 
2. stage: [install](https://docs.google.com/document/d/1e54wnFIZncL2TpfUN2-QuNRUhrUDEsKmgZAMhZx70EU/edit?tab=t.0#heading=h.qss3evvgdsto)

Again this is based on my understanding - @psss WDYT?

tmt/steps/prepare/artifact/providers/copr_repository.py
tcornell-bus: Thanks for the quick update! The new complaint seems to be:
```
'Copr' object has no attribute 'copr_plugin'
```

AthreyVinay: Thanks Therese for testing this out again. 

I have pushed the fix now :)

tmt/steps/prepare/artifact/providers/copr_repository.py
LecrisUT: Let's keep it out for now. I don't think the `stage` part was particularly discussed, and not sure how it would differentiate in the current state. We can always add it back when we get to it.

tests/unit/artifact/conftest.py
```diff
@@ -53,3 +53,9 @@ def get_provider(provider_id: str):
         return provider_class(provider_id, logger=root_logger)
 
     return get_provider
+
+
+@pytest.fixture
+def mock_copr_class():
+    with patch('tmt.steps.prepare.artifact.providers.copr_repository.Copr') as mock_copr:
```
LecrisUT: Don't like how this one is pointing to a non-canonical name.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -0,0 +1,93 @@
...
+    def _extract_provider_id(cls, raw_provider_id: str) -> ArtifactProviderId:
+        prefix = 'copr.repository:'
+        if not raw_provider_id.startswith(prefix):
+            raise ValueError(f"Invalid Copr repository provider format: '{raw_provider_id}'.")
+
+        value = raw_provider_id[len(prefix) :]
+        if not value:
+            raise ValueError("Missing Copr repository name.")
+
+        if not re.match(r'^(@)?[^/]+/[^/]+$', value):
```
LecrisUT: Pre-compile the regex.

tmt/steps/prepare/install.py
```diff
@@ -229,15 +229,16 @@ def rpm_check(self, package: str, mode: str = '-q') -> None:
...
     guest: Guest
 
     # Keep this method around, to correctly support Python's method resolution order.
-    def __init__(self, *args: Any, guest: Guest, **kwargs: Any) -> None:
-        super().__init__(*args, guest=guest, **kwargs)
+    def __init__(self, *, guest: Guest, copr_plugin: Optional[str] = None, **kwargs: Any) -> None:
+        super().__init__(guest=guest, **kwargs)
 
         self.guest = guest
+        self.copr_plugin = copr_plugin or self.copr_plugin
```
LecrisUT: Why the changes here?

LecrisUT: It would be better to generalize the `PrepareInstall.go` logic where the `installer` is defined and use an `installer` directly with empty `dependencies` etc.

AthreyVinay: > Why the changes here?

See: https://github.com/teemtee/tmt/pull/4471#discussion_r2682810920.
Basically, Copr needed `copr_pplugin`  be set as an attribute, not passed to `__init__`
I couldve gone down 2 routes: 
1. As I have done - i.e. handle in the constructor
2. On the caller, set it after creating the instance. 

I felt the former was a better/cleaner approach.

> It would be better to generalize the PrepareInstall.go logic where the installer is defined and use an installer directly with empty dependencies etc.

I like the idea. I’ll refactor the installer selection logic in a follow-up PR.

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: fixed: [address more comments...](https://github.com/teemtee/tmt/pull/4471/commits/cdda787a39ba1e1b09bdeb7f8312d534ef31e0f7)

tests/unit/artifact/conftest.py
AthreyVinay: fixed: [address more comments...](https://github.com/teemtee/tmt/pull/4471/commits/cdda787a39ba1e1b09bdeb7f8312d534ef31e0f7)

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: fixed: [address more comments...](https://github.com/teemtee/tmt/pull/4471/commits/cdda787a39ba1e1b09bdeb7f8312d534ef31e0f7)

tmt/steps/prepare/install.py
LecrisUT: > > It would be better to generalize the PrepareInstall.go logic where the installer is defined and use an installer directly with empty dependencies etc.
> 
> I like the idea. I’ll refactor the installer selection logic in a follow-up PR.

Should note that that is how it can support epel7 and other formats. But main concern is remembering to revert the changes here. If you can prepare a draft PR dependent on it then I'm 👍.

AthreyVinay: I'll take the 2nd approach - i.e. set it as an attribute after the instantiation for now (just because Ill do a proper PR later, dont want to delay this!). This will revert any changes on this file - so should be safe to go ahead.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -0,0 +1,96 @@
...
+        download_path: Path,
+        exclude_patterns: Optional[list[Pattern[str]]] = None,
+    ) -> list[Path]:
+        """
+        Enable the Copr repository on the guest.
+
+        :return: Empty list as no files are downloaded.
+        """
+        copr = install.Copr(logger=self.logger, guest=guest)
+        copr.copr_plugin = "dnf-plugins-core"
```
thrix, line 94: @AthreyVinay can we add somehwere in issues to solve later this problem?

```
  1. Hardcoded copr_plugin (copr_repository.py:93-94):
  copr = install.Copr(logger=self.logger, guest=guest)
  copr.copr_plugin = "dnf-plugins-core"
  1. This hardcodes dnf-plugins-core, but different package managers need different plugins:
    - dnf: dnf-plugins-core
    - dnf5: dnf5-command(copr)
    - yum: yum-plugin-copr
```

thrix (approved):
LGTM

tmt/steps/prepare/artifact/providers/copr_repository.py
LecrisUT: Discussed previously https://github.com/teemtee/tmt/pull/4471#discussion_r2688219669. Will generalize installer in `PrepareInstall.go`

happz (approved):

tmt/steps/prepare/install.py
AthreyVinay: Leaving this in for visibility:
copr = install.Copr(logger=self.logger, guest=guest)
  copr.copr_plugin = "dnf-plugins-core"
  1. This hardcodes dnf-plugins-core, but different package managers need different plugins:
    - dnf: dnf-plugins-core
    - dnf5: dnf5-command(copr)
    - yum: yum-plugin-copr

AthreyVinay: Also updated the TODO issue with this requirement.

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: @tcornell-bus If you can verify if this fixes the issue? Thanks :)

thrix:
virtual provisioner tests timeout is a known issue

PR #4472, opened 2026-01-09
Apply plugin envvars to phases not on command line

### LINKED ISSUE #4270: testing-farm inject TMT_PLUGIN_REPORT_REPORTPORTAL_TOKEN failed
in my tmt main.fmf file I define a faker_token as below:

```
report:
  - name: tmt local report
    how: html
  - name: reportportal
    how: reportportal
    project: mhou_personal
    url: https://reportportal-rhelnet.apps.dno.ocp-hub.prod.psi.redhat.com
    log-size-limit: 32M
    token: faker_token
    suite-per-plan: true
    launch: debug_restraint test on $@{distro}
```

then I try assign my token when submit testing-farm through --tmt-environment "TMT_PLUGIN_REPORT_REPORTPORTAL_TOKEN=xxx"
But the job report failed to reportportal.
https://artifacts.osci.redhat.com/testing-farm/69493ec8-5a7f-44c8-b4e1-ea288d14fe84/

```
    report
        how: html
        name: tmt local report
        output: /var/ARTIFACTS/work-planabwvay5x/rt-kernel_misc_test/plan/report/tmt-local-report/index.html
        how: reportportal
        name: reportportal
        launch: debug_restraint test on RHEL-10.0-ctc
    report
        how: html
        name: tmt local report
        output: /var/ARTIFACTS/work-planabwvay5x/rt-kernel_misc_test/plan/report/tmt-local-report/index.html
        how: reportportal
        name: reportportal
        launch: debug_restraint test on RHEL-10.0-ctc
    cleanup
        queued cleanup task #1: default-0 on default-0 (standalone)
        
        cleanup task #1: default-0 on default-0 (standalone)
        how: tmt
        summary: 1 task completed

plan failed

Cause number 1:

    Received non-ok status code 401 from ReportPortal: {"error":"invalid_token","error_description":"Invalid access token"}

---^---^---^---^---^---
```

happz:
Plugin envvars take effect even when plugin is not mentioned on CLI

This was a major drawback when using envvars to populate or change phases' configuration: plugin-specific envvars had no effect unless the plugin was invoked via command line. This was caused by Click consuming them, and without mentioning the plugin on command line, Click was not invoked, therefore nothing consumed said envvars.

The patch adds one more pass into the application of CLI invocations where we collect all these envvars, and then we use Click to process them. The produced values are then added into raw step data.

We do not use custom parsers or type casting, we identify the plugin in question from the envvar name, and use its Click command and its parameters to perform the necessary steps for us. This should be equivalent to what Click does when crunching the command line.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note


tmt/steps/__init__.py
```diff
@@ -1096,18 +1131,96 @@ def _patch_raw_datum(
...
+
+                    if plugin_param is None:
+                        raise GeneralError(
+                            f"Failed to find the '{param_name.replace('_', '-')}' key"
+                            f" of the '{self.step_name}/{how}' plugin."
+                        )
+
+                    raw_datum[key_to_option(param_name)] = plugin_param.type_cast_value(  # type: ignore[literal-required]
+                        plugin_context, str(envvar_value)
+                    )
```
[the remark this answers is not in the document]
happz: 6f4570da

tmt/steps/__init__.py
```diff
@@ -1040,6 +1046,35 @@ def _iter_options() -> Iterator[tuple[str, Any]]:
         # and we will get back to them once we're done with those we can apply immediately.
         postponed_invocations: list[tmt.cli.CliInvocation] = []
 
+        # A container of all plugin-specific environment variables. Here we collect all
+        # plugin-specific environment variables, and separate them into buckets per plugin. Note
+        # that we do not care about *phases* - there is no way to apply environment variable to
+        # one specific phase, they are very strong and affect all phases spawned from the same
+        # plugin, i.e. with the same `how`.
+        environment_invocations: dict[str, dict[str, tuple[EnvVarName, EnvVarValue]]] = (
```
LecrisUT: Uph, quite the convoluted type. Can you make it a dedicated type? We can hook the plugin into it and do the check for allowed step/plugin/option more localized

happz: I can make a dedicated type, but would it help? It’s used just here, I can add describing comments and reformat it right here, to make it easiser to follow.

Although I don’t understand which plugin and how the check would be more localized, if you vould elaborate, it’d help.

LecrisUT: > I can make a dedicated type, but would it help?

Yes, at the bare minimum it keeps the related functions/logic in close proximity. E.g. the logic now under the `enumerate(environment_invocations.items())` can be wrapped in a single function

> Although I don’t understand which plugin

It can be extended the other way around, with the StepPlugins announcing to this class what step/plugin/options are available instead of doing the search the other way around. In that way all of the filtering logic in the enumeration can be moved to the section just below this (`Environment.from_environ().items()`)

tests/cli/plugin-envvars/main.fmf
```diff
@@ -0,0 +1,4 @@
+summary: Verify some basic output of `tmt about`
```
AthreyVinay: Ctrl C + Ctrl V?

tmt/steps/__init__.py
```diff
@@ -1040,6 +1046,35 @@ def _iter_options() -> Iterator[tuple[str, Any]]:
...
+
+            # ... or for different step. Neither of these groups is interesting here and now.
+            step_name = match.group(1).lower()
+
+            if step_name != self.step_name:
+                continue
+
+            # Since the environment variable names are upper-cased only, plugin and field names
+            # require a bit of character manipulation.
+            plugin_name: str = match.group(2).lower().replace('_', '-')
```
therazix: We could use the `key_to_option` function, like on the line below.
```suggestion
            plugin_name: str = key_to_option(match.group(2).lower())
```

tmt/steps/__init__.py
```diff
@@ -1110,18 +1145,96 @@ def _patch_raw_datum(
...
+                    plugin_param = plugin_params.get(param_name)
+
+                    if plugin_param is None:
+                        raise GeneralError(
+                            f"Failed to find the '{param_name.replace('_', '-')}' key"
+                            f" of the '{self.step_name}/{how}' plugin."
+                        )
+
+                    raw_datum[key_to_option(param_name)] = plugin_param.type_cast_value(  # type: ignore[literal-required]
+                        plugin_context, str(envvar_value)
```
tcornell-bus: There is at least one envvar that is `multiple=True` and the `str()` here seems to break it. reportportal has `upload-log-pattern`, which can be a list. Adding this envvar works fine when the report plugin is mentioned in the CLI when running with tmt-1.72.0
```
tcornell ~/github/teemtee/tmt/tests/cli/plugin-envvars/data (apply-plugin-envvars-to-phases-not-on-command-line) $ TMT_PLUGIN_REPORT_REPORTPORTAL_UPLOAD_LOG_PATTERN="cool.log my-special-log" tmt -vvv --feeling-safe --log-topic=cli-invocations run -a report -h reportportal --project my_personal_project
/var/tmp/tmt/run-176
[...redacted for brevity, it finishes successfully...]

tcornell ~/github/teemtee/tmt/tests/cli/plugin-envvars/data (apply-plugin-envvars-to-phases-not-on-command-line) $ grep -E "ReportPlugin.delegate" /var/tmp/tmt/run-176/log.txt
14:31:36     ReportPlugin.delegate(step=report, data=None, raw_data={'name': 'default-0', 'how': 'reportportal', <...redacted...>, 'upload-log-pattern': ('cool.log', 'my-special-log')})
```
When I run this same command in dev mode I get an exception:
```
Cause number 1:

    Value must be an iterable.
```

I added two other comments in this review that should reproduce the issue.

gpt flagged this and I verified.

tests/cli/plugin-envvars/test.sh
```diff
@@ -0,0 +1,47 @@
...
+        rlAssertGrep "warn: Found environment variables for plugin 'report/display', but the plugin is not used by the plan '/'. The following environment variables will have no effect:" $rlRun_LOG
+        rlAssertGrep "warn: TMT_PLUGIN_REPORT_DISPLAY_DISPLAY_GUEST" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartTest "Verify unknown options are reported"
+        rlRun -s "TMT_PLUGIN_REPORT_HTML_HIDE_GUEST=never $run_tmt" 2
+
+        rlAssertGrep "Failed to find the 'hide-guest' key of the 'report/html' plugin." $rlRun_LOG
+    rlPhaseEnd
+
```
tcornell-bus, line 40:
```suggestion

    rlPhaseStartTest "Verify multi-value plugin option"
        rlRun "run_tmt_reportportal=\"tmt --context report_portal=1 -vvv --feeling-safe --log-topic=cli-invocations run --id $rundir --scratch --dry\""

        rlRun -s "TMT_PLUGIN_REPORT_REPORTPORTAL_UPLOAD_LOG_PATTERN=\"cool.log my-special-log\" $run_tmt_reportportal"
        rlAssertGrep "ReportPlugin.delegate\(step=report, data=None, raw_data=\{'how': 'reportportal',.*'upload-log-pattern': \('cool.log', 'my-special-log'\).*\)" $rundir/log.txt -E
    rlPhaseEnd

```

tests/cli/plugin-envvars/data/main.fmf
```diff
@@ -0,0 +1,10 @@
+provision:
+  how: local
+
+execute:
+  script: /bin/true
+
+report:
+  - how: html
+    absolute-paths: false
+    display-guest: auto
```
tcornell-bus, line 10:
```suggestion
    display-guest: auto

adjust:
  - when: report_portal is defined
    report:
      - how: reportportal
        project: fake-project
```

tmt/steps/__init__.py
happz: `envvar_value` is `EnvVarValue` instance, which is just a fancy `str`; `str()` call here is a no-op in runtime, `str()` just tells mypy & co. everything's fine. There will be a bug for sure, but probably not because of `str()` call.

tmt/steps/__init__.py
happz: I gave it some thought and 1. I'm not sure I fully understand what you're proposing, a quick dirty draft would help, and 2. I kind of like the approach I took here.

> > I can make a dedicated type, but would it help?
> 
> Yes, at the bare minimum it keeps the related functions/logic in close proximity. E.g. the logic now under the `enumerate(environment_invocations.items())` can be wrapped in a single function
> 
> > Although I don’t understand which plugin
> 
> It can be extended the other way around, with the StepPlugins announcing to this class what step/plugin/options are available instead of doing the search the other way around. In that way all of the filtering logic in the enumeration can be moved to the section just below this (`Environment.from_environ().items()`)

The thing is, I already have the environment variables, the interesting ones, and it's trivial to find which plugin they belong to. Starting from plugins, I'd have to iterate over phases and ask them whether they claim ownership of the given interesting envvar when I already can tell, from the envvar name, whether they would or wouldn't. Thus, taking an interesting envvar and then filtering phases to find matching `how` seems cleaner to me. Plus, there may be e.g. misspelled envvars, correct `how` but wrong key name, and the plugin would most likely not claim the ownership as it wouldn't match any of its options. Starting from envvars, I can know which plugin is responsible, and I emit nicer message. Maybe I just don't understand the picture behind your comment.

LecrisUT: > 1. I'm not sure I fully understand what you're proposing, a quick dirty draft would help

```python
@container
class PluginEnvironment:
    PATTERN: ClassVar[re.pattern] = re.compile(...)
    step: StepName
    plugin: str
    option: str
    value: str

    @classmethod
    def parse_var(cls, name: str, value: str) -> Optional[Self]:
        if not (match := cls.PATTERN.match(name)):
            return None
        step, plugin, option = match.groups()
        if step not in STEPS:
            logger.warn(...)
            return None
        ...
```

Some cached variables and better optimization is needed especially if we want to warn only once. But the main thing there is that it splits the monolith of `_apply_cli_invocations` into more contained chunks. Even if you take the whole functions you have currently and wrap them in more manageable functions that is still a win for me.

---

About the ways to expand this, once we have better wrappers for `@container` and `field`, we can avoid the `self._plugin_base_class._supported_methods.get_plugin(how)` and `plugin_class.class_.command()` and have these be called when `@container` and `field` are executed. Then in something like `parse_var` above you would only add the checks from navigating the `Step` variables
```python
        if plugin not in step.plugins:
            logger.warn(...)
            return None
        if option not in step.plugins[plugin].options:
            logger.warn(...)
            return None
```

tests/cli/plugin-envvars/main.fmf
happz: Fixed in afc3c26d.

tmt/steps/__init__.py
happz: No longer relevant.

tmt/steps/__init__.py
happz: I think I can't use the reproducer, because `reportpoertal` ignores `--dry`: https://github.com/teemtee/tmt/issues/4851

But I've found the issue and fixed it in afc3c26d when I run the test command it correctly handles the list (and then tries to talk to a real RP instance...)

tmt/steps/__init__.py
happz: Well, I tried something in afc3c26d.

tmt/steps/__init__.py
tcornell-bus: Thanks for filing the issue :)

tests/cli/plugin-envvars/test.sh
```diff
@@ -0,0 +1,53 @@
...
+        rlAssertGrep "warn: Found environment variable for plugin 'report/display', but the plugin is not used by the plan '/'. The following environment variable will have no effect: TMT_PLUGIN_REPORT_DISPLAY_DISPLAY_GUEST" $rlRun_LOG
+    rlPhaseEnd
+
+    rlPhaseStartTest "Verify unknown options are reported"
+        rlRun -s "TMT_PLUGIN_REPORT_HTML_HIDE_GUEST=never $run_tmt" 2
+
+        rlAssertGrep "Failed to find the 'hide-guest' key of the 'report/html' plugin." $rlRun_LOG
+    rlPhaseEnd
+
+# Disabled: report/reportportal ignores `--dry`, we can't have it connecting anywhere.
```
tcornell-bus, line 41: minor request to add the filed issue to this comment

tcornell-bus (approved):
LGTM

tests/cli/plugin-envvars/test.sh
happz: Added in 27f8e2f9

tmt/steps/__init__.py
LecrisUT: Yep, looks great now. Just a few technical comments that I will follow up in there.

tmt/steps/__init__.py
```diff
@@ -503,6 +509,141 @@ class WhereableStepData(SerializableContainer):
...
+        return self.key.replace('-', '_')
+
+    # We do not want to invent our custom parser for turning an environment
+    # variable into plugin step data value. Instead, we need to reach the
+    # Click command of the plugin, because that command already knows how
+    # to handle its input. This requires a bit of traversal. Get the right
+    # plugin class, then its command, and then its parameters.
+
+    @functools.cached_property
+    def plugin_class(self) -> Optional['Method']:
```
LecrisUT, line 551: The `cached_property` here is not that useful when it is bound to the instance isn't it? A quick fix would be to make it cached on the `@staticmethod` (note that `@classmethod` + `@property` is deprecated). Could add it as an additional layer to this property to keep the interface.

AthreyVinay (approved):
lgtm.

tmt/steps/__init__.py
```diff
@@ -503,6 +509,141 @@ class WhereableStepData(SerializableContainer):
...
+
+            step_name, plugin_name, option_name = tuple(
+                component.lower() for component in match.groups()
+            )
+
+            if not is_step_name(step_name):
+                continue
+
+            if step_name != step.name:
+                continue
```
LecrisUT, line 634:
```suggestion
            if step_name != step.name:
                continue
            assert is_step_name(step_name)  # narrow type
```
Good either way, but technically more efficient.

tmt/steps/__init__.py
```diff
@@ -503,6 +509,141 @@ class WhereableStepData(SerializableContainer):
...
+
+    @functools.cached_property
+    def plugin_command(self) -> click.Command:
+        """
+        Plugin command-line command that owns the option-to-value conversion.
+        """
+
+        assert self.plugin_class is not None
+
+        return self.plugin_class.class_.command()
```
LecrisUT: This will blow up in the debugger because it's a property. Can make it safely return `None` or move the plugin warning in `from_environment`?

LecrisUT (commented):
LGTM other than the `cached_property` comment. Just 2 other minor improvement suggestions

tmt/steps/__init__.py
happz: I don't think that's possible, the return value is derived from `self.step` + `self.how`, both are instance attributes. It returns a plugin that would be responsible for this particular environment variable.

tmt/steps/__init__.py
happz: I don't see the issue. `self.plugin_class` is a property, `self.plugin_class.class_` is an attribute of `Method`, `self.plugin_class.class_.command` is a classmethod of `BasePlugin`. What am I missing?

tmt/steps/__init__.py
happz: Hmmm, I like mine more, because first I make sure I have the right type, `StepName`, then I compare two values of the same type.

tmt/steps/__init__.py
LecrisUT: The `assert self.plugin_class is not None` is valid when `self.plugin_class` is `None` and it goes through the warning path. This would only blow up when you are in the debugger and break just before that warning, so not a blocking issue for me on this.

tmt/steps/__init__.py
LecrisUT: I mean
```python
    @functools.cache
    @staticmethod
    def _get_plugin_class(step: 'Step', how: str) -> Optional['Method']:
        return step._plugin_base_class._supported_methods.get_plugin(how)

    @property
    def plugin_class(self) -> Optional['Method']: 
        """
        Plugin class according to :py:attr:`how`, or ``None``.
        """

        return self._get_plugin_class(self.step, self.how)
```

Alternatively, we could also revisit this when we get to the `attrs`/`pydantic` customization, or look into caching the `get_plugin` or similar paths, since it technically should not be that expensive in practice.

happz: I don't see the benefit, why is it better? `plugin_class` feeds `self.step` and `self.how` to a static method, why not using them right away? Who are other users of `_get_plugin_class`?

tmt/steps/__init__.py
happz: > The `assert self.plugin_class is not None` is valid when `self.plugin_class` is `None`

How? I don't follow, sorry, how could `self.plugin_class is not None` be valied when "`self.plugin_class` is `None`"?

tmt/steps/__init__.py
LecrisUT: It's about `@functools.cache` and making it work. `@functools.cached_property` would not have any effect because `self` is different as the `name`,`value`, and `key` always differ. The snippet below illustrates the logic difference
```python
>>> import functools
... import dataclasses
...
... @dataclasses.dataclass
... class A:
...     unique: int
...     _shared: str
...     
...     @functools.cached_property
...     def shared(self) -> str:
...         print("Running cached_property")
...         return self._shared
...
... a_1 = A(1, "A")
... a_2 = A(2, "A")
... print(a_1.shared)
... print(a_2.shared)
Running cached_property
A
Running cached_property
A
>>> class A2(A):
...     @staticmethod
...     @functools.cache
...     def _get_shared(shared: str) -> str:
...         print("Running cached staticmethod")
...         return shared
...     
...     @property
...     def shared(self) -> str:
...         print("Running property wrapper")
...         return self._get_shared(self._shared)
...
... a2_1 = A2(1, "A2")
... a2_2 = A2(2, "A2")
... print(a2_1.shared)
... print(a2_2.shared)
Running property wrapper
Running cached staticmethod
A2
Running property wrapper
A2
```

tmt/steps/__init__.py
LecrisUT: > > The `assert self.plugin_class is not None` is valid when `self.plugin_class` is `None`
> 
> How? I don't follow, sorry, how could `self.plugin_class is not None` be valied when "`self.plugin_class` is `None`"?

Hmm, not that it is "valid", but that `assert self.plugin_class is not None` will trigger (`raise AssertionError`)

tmt/steps/__init__.py
happz: I'm sorry but I really, really don't understand what you are trying to point out.

`functools.cached_property` on instance method is perfectly correct, it's literally the example from its documentation, https://docs.python.org/3/library/functools.html#functools.cached_property. Yes, `self`, `name`, `value`, `how` would be different, but they are supposed to be different, they are all instance properties, and `plugin_class` is not shared in any way between instances. It is supposed to give me a plugin class **for this given plugin-specific environment variable**. `PluginEnvVar(name='TMT_PLUGIN_REPORT_HTML_...').plugin_class` must - and is - different than `PluginEnvVar(name='TMT_PLUGIN_REPORT_JUNI_...').plugin_class`. There is no `_shared` these two **instances** would share. Plugin class depends on `self.step` and `self.how` alone. I don't see where static or class methods or properties would be needed, as this is all about distinct instances.

> It's about `@functools.cache` and making it work. `@functools.cached_property` would not have any effect because `self` is different as the `name`,`value`, and `key` always differ.

Yes, it would have an effect: once invoked, its return value is stored in `self.__dict__`, replaces the method with actual value. All future access to this instance property would use the precomputed value. It's really attached to `self` alone, and there is no information shared by different `PluginEnvVar` instances.

psss (approved):
Nice! Looks good. Thanks for fixing this.

tmt/steps/__init__.py
LecrisUT: Ok, let's be concrete, using the `cached_property`, for each `PluginEnvVar` this saves ~2 function calls in `plugin_command`, between `TMT_PLUGIN_REPORT_*` there is no caching so you would miss out on `n-1` calls where `n` is the number of options for the same step-plugin.

The output of `plugin_class` is identical between any `TMT_PLUGIN_REPORT_*` variables, and more importantly, so are `plugin_command` and `plugin_command_params`.

Does it make sense to add the caching via a dedicated `staticmethod`? No not really, caching should happen at `get_plugins` instead. The purpose of the `staticmethod` and the snippet is just to illustrate the `n-1` calls that are missed when `step + how` are identical (equivalent to `_shared` in the snippet).

Do you save much from even caching the `get_plugins`, I don't think so, it's just a small wrapper over `dict.get`.

What about `plugin_command` or `plugin_command_params`, yes, those are actually quite expensive, and we are missing `n-1` cachable calls in the current form.

happz: Ahaaaa, no I understand what you are after.

> Ok, let's be concrete

Yes, please.

> using the `cached_property`, for each `PluginEnvVar` this saves ~2 function calls in `plugin_command`, between `TMT_PLUGIN_REPORT_*` there is no caching so you would miss out on `n-1` calls where `n` is the number of options for the same step-plugin.
> 
> The output of `plugin_class` is identical between any `TMT_PLUGIN_REPORT_*` variables, and more importantly, so are `plugin_command` and `plugin_command_params`.

So you were proposing to add caching so `PluginEnvVar` instances targetting the same plugin would benefit from already having the plugin class, plugin command, plugin command params. Now this I do understand, and I think I can change code to add this "layer" - although I would mind leaving it for later.

tmt/steps/__init__.py
happz: I see. It's supposed to be the `# narrow type` kind of assert. Added a comment in a34a70d0 to make it clear. Eventually, this particula place would benefit from encoding invariants as types, e.g. a subclass without `Optional` for envvar we already know that have a backing plugin.

tmt/steps/__init__.py
tcornell-bus: @happz could you make a quick maintenance issue for this and link here?

LecrisUT (approved):

tmt/steps/__init__.py
happz: https://github.com/teemtee/tmt/issues/4853

tmt/steps/__init__.py
```diff
@@ -139,6 +140,11 @@
 
 CODE_BLOCK_REGEXP = re.compile(r"^\s*\.\. code-block::.*$\n", re.MULTILINE)
 
+#: A regular expression to cut the special, plugin-specific environment
+#: variable names into their components: step, plugin, and option names.
+PLUGIN_ENVIRONMENT_VARIABLE_NAME_PATTERN = re.compile(
+    r'TMT_PLUGIN_(?P<step>[A-Z]+)_(?P<plugin>[A-Z]+)_(?P<option>[A-Z_]+)'
```
psss, line 146: Hm, one thought: What if the plugin has a multi-word name? For example the `verify-installation`? Variable name in such case would be `TMT_PLUGIN_PREPARE_VERIFY_INSTALLATION_SOMETHING` and this would be incorrectly matched above.

happz: Yes, that is a genuine issue. Do you want to get it fixed in this PR?

psss: I don't think it would be a blocker, let's track that in a follow-up issue.

happz: https://github.com/teemtee/tmt/issues/4857

happz:
Something canceled the last remaining job, again.

happz:
And again.

happz:
And again.

LecrisUT:
The job is running under https://dashboard.packit.dev/jobs/testing-farm/1535847 and it should at the very least push out the old result when it finishes. So let's wait a bit for it

happz:
The last job is green.

PR #4473, opened 2026-01-11
Fixing the console log path by simplifying guest log updates

### LINKED ISSUE #4466: Invalid console path provided during the `testcloud` provision
When verbose mode is used with `provision --how virtual` local path to the console log is printed:

    provision
        how: virtual
        order: 50
        ansible: GuestAnsible(group=None, vars={})
        qcow: Fedora-Cloud-Base-Generic-43-1.6.x86_64.qcow2
        effective hardware: {}
        memory: 2048 MB
        disk: 40 GB
        name: tmt-ole-tkmalsjw
        key: /var/tmp/tmt/console/default/plan/provision/default-0/id_ecdsa
        progress: booting...
        console: /var/tmp/tmt/console/default/plan/provision/default-0/logs/console.txt
        ...

After #4104 the path is not working. It seems that `console.txt.link` is used as the filename instead.

happz:
* `virtual` console log was incorrect, after changes made in #4104.
* Console log is different than other other guest logs, as it appears, out of thin air, as a file on the disk. tmt does not need to download it, it just needs to make it visible to user.
* the possibility of using a custom directory for log storage was making the API needlessly complicated - we do not expect to have custom directories, we are barely starting fetching the logs.

All in all, I tried to fix the issue with an incorrect console log path being presented to users, while doing so I had to modify the guest log API, to make `GuestLog` instances aware of their filesystem storage. Dropping the custom logdir option made things much simpler, and it is not going to be missed now.

Fixes #4466.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -2443,7 +2443,7 @@ class Guest(
 
         return []
 
-    @property
+    @functools.cached_property
     def logdir(self) -> Path:
         """
         Path to store logs
@@ -2474,24 +2474,20 @@ class Guest(
 
         self.guest_logs.append(log)
 
-    def setup_logs(self, *, dirpath: Optional[Path] = None, logger: tmt.log.Logger) -> None:
+    def setup_logs(self, *, logger: tmt.log.Logger) -> None:
         """
         Notify all registered logs their collection will begin.
 
-        :param dirpath: a directory to save into. If not set,
-            :py:attr:`logdir`, or current working directory will be used.
         :param logger: logger to use for logging.
         """
 
         for log in self.guest_logs:
             log.setup(logger=logger)
 
-    def teardown_logs(self, *, dirpath: Optional[Path] = None, logger: tmt.log.Logger) -> None:
+    def teardown_logs(self, *, logger: tmt.log.Logger) -> None:
         """
         Notify all registered logs their collection will no longer continue.
 
-        :param dirpath: a directory to save into. If not set,
-            :py:attr:`logdir`, or current working directory will be used.
         :param logger: logger to use for logging.
         """
 


psss:
Nice! Thanks for fixing this. I suggest to include this in the next sprint.

tmt/steps/provision/__init__.py
```diff
@@ -1502,11 +1510,12 @@ def staging_file(self, final_filepath: Path, logger: tmt.log.Logger) -> Iterator
...
 
     @abc.abstractmethod
-    def update(self, filepath: Path, logger: tmt.log.Logger) -> None:
+    def update(self, *, final: bool = False, logger: tmt.log.Logger) -> None:
         """
         Fetch the up-to-date content of the log, and save it into a file.
 
-        :param filepath: a file to save into.
+        :param final: if set, this is the last attempt to update the log
+            before the guest cleanup.
```
LecrisUT: Why not have a different interface for it to `finalize()` the log or use the `teardown`?

tmt/steps/provision/__init__.py
```diff
@@ -2437,17 +2446,14 @@ def essential_requires(cls) -> list['tmt.base.Dependency']:
         return []
 
     @property
```
```suggestion
    @cached_property
```
since the interface is clearer that it is always a fixed Path now.

tmt/steps/provision/__init__.py
```diff
@@ -2470,49 +2476,47 @@ def collect_log(self, log: GuestLog, hint: Optional[str] = None) -> None:
...
-            log.setup(logger)
+            log.setup(logger=logger)
 
-    def teardown_logs(self, logger: tmt.log.Logger) -> None:
+    def teardown_logs(self, *, dirpath: Optional[Path] = None, logger: tmt.log.Logger) -> None:
         """
         Notify all registered logs their collection will no longer continue.
+
+        :param dirpath: a directory to save into. If not set,
+            :py:attr:`logdir`, or current working directory will be used.
```
LecrisUT: Does not seem to be used

tmt/steps/provision/__init__.py
happz: Definitely not bundling an update with the `teardown()`. It's still an update, it's just the last one, and `finalize()` seems like what `teardown()` could be called. `final=True` seemed like the reasonable choice here.

LecrisUT: `final` seems fragile to me. Can you expand on why the copying of the logs is a no-go for the `teardown()`?

happz: Sure. I want to keep `setup` and `teardown` focused on actions that do not involve the actual update of the log. Both would be just one-time events in the lifetime of a guest log, while update would be a periodically invoked method. It is possible to perform an update in `teardown`, but then the distinction between the sequence of "setup your workspace" -> "make updates, one, many, whatever" -> "clean up after" would be much less clear.

`final: bool` won it for me over "final update" method, as I saw no benefit in having a dedicated method. Not a big issue though, I could live with an extra method with clear purpose, if there's an advantage in having it. Better than `teardown` doing two things instead of one.

On the other hand, this debate may be pointless as I think about the `final` more - it exists so `virtual` could replace its symlink with an actual content, which suddenly seems more suitable for `teardown`. It's not an update per se, it's more like closing a book. So I'd probably move this to `teardown`, and then there'd be no need for `final` at this moment.

LecrisUT: > On the other hand, this debate may be pointless as I think about the `final` more - it exists so `virtual` could replace its symlink with an actual content, which suddenly seems more suitable for `teardown`. It's not an update per se, it's more like closing a book. So I'd probably move this to `teardown`, and then there'd be no need for `final` at this moment.

Indeed that. I was considering that instead of `final` we would just run that snippet each time, but then I figured out what it was actually trying to do (convert the symlink _in place_). Then it felt like a finalization action which got me confused with teardown scope.

skycastlelily (approved):
LGTM, I'm not able to reproduce the issue anymore^^

tmt/steps/provision/__init__.py
happz: Addressed in d2e957a5

tmt/steps/provision/__init__.py
happz: Dropped in 6cac0481

tmt/steps/provision/__init__.py
happz: Fixed in 6cac0481

PR #4449, opened 2025-12-24
Add example tests for artifact providers

vaibhavdaren:
Add simple, example-based integration tests for artifact providers:
- koji.build: Fetches packages from Koji using build IDs
- repository-url: Configures external package repositories
- multi: Demonstrates using multiple providers together

Once the design is approved, additional providers (koji.task, koji.nvr, file, copr.build, brew.*) can be added following the same pattern.

Related to #4420

PR #4474, opened 2026-01-11
Add repository-url artifact provider

vaibhavdaren:
Implements a simple repository-url provider that accepts baseurl as input. The provider creates a .repo file with the given baseurl and makes packages available through the package manager.

Usage:
```
  prepare:
    how: artifact
    provide:
      - repository-url:https://download.docker.com/linux/fedora/$releasever/$basearch/stable
```
Includes test that verifies docker-ce-cli installation from the Docker CE repository baseurl.

Pull Request Checklist

* [ ] implement the feature

---

**Next PR** : The provides arugment should be a ProvideSpec Object.  It should take multiple fields, so that we can get multiple inputs for the provider like id , name, priority etc.

### DIFF
--- a/tmt/steps/prepare/artifact/providers/repository_url.py
+++ b/tmt/steps/prepare/artifact/providers/repository_url.py
@@ -68,6 +68,7 @@ class RepositoryUrlProvider(ArtifactProvider[RpmArtifactInfo]):
         self, artifact: RpmArtifactInfo, guest: Guest, destination: tmt.utils.Path
     ) -> None:
         """This provider only discovers repos; it does not download individual RPMs."""
+        # FIXME: Change this to UnsupportedOperationError once its available
         raise AssertionError("RepositoryUrlProvider does not support downloading individual RPMs.")
 
     def fetch_contents(


tmt/steps/prepare/artifact/providers/repository_baseurl.py
```diff
@@ -0,0 +1,126 @@
...
+from tmt.steps.provision import Guest
+
+# FIXME: use priority from https://github.com/teemtee/tmt/pull/4454
+# Default priority for the repository (lower value = higher priority)
+DEFAULT_PRIORITY = 50
+
+
+# ignore[type-arg]: TypeVar in provider registry annotations is
+# puzzling for type checkers. And not a good idea in general, probably.
+@provides_artifact_provider('repository-baseurl')  # type: ignore[arg-type]
```
LecrisUT: `repository-url` as the name for this please.

vaibhavdaren: Isolating the changes to smaller PRs :  https://github.com/teemtee/tmt/pull/4476. Renaming the old provider to correct name.

tests/prepare/artifact/providers/repository-url/main.fmf
```diff
@@ -0,0 +1,11 @@
...
+description: |
+    Test repository-url artifact provider which configures external
+    package repositories from baseurl.
+tag+:
+  - provision-only
+  - provision-container
+adjust:
+  - when: distro != fedora
+    enabled: false
+    because: Test uses Fedora-specific repository
```
LecrisUT: Comments in https://github.com/teemtee/tmt/pull/4449 re-apply here. It seems you will have to make one of these PRs depend on the other.

tmt/steps/prepare/artifact/providers/repository_url.py
```diff
@@ -0,0 +1,108 @@
...
+from tmt.steps.prepare.artifact.providers import (
+    ArtifactProvider,
+    ArtifactProviderId,
+    Repository,
+    provides_artifact_provider,
+)
+from tmt.steps.prepare.artifact.providers.repository import _REPO_NAME_GENERATOR
+from tmt.steps.provision import Guest
+
+# TODO: Implement repository priority support using https://github.com/teemtee/tmt/pull/4454
```
LecrisUT: The link is a PR. Do you need to add anything on top of that PR or just it will automagically solve itself after merge.

tmt/steps/prepare/artifact/providers/repository_url.py
```diff
@@ -0,0 +1,108 @@
...
+
+    The provider generates a .repo file with the given baseurl, which will be installed to the
+    guest's ``/etc/yum.repos.d/`` directory, and lists RPMs available in the repository
+    without downloading them, acting as a discovery-only provider. Artifacts are all available
+    RPM packages listed in the repository.
+
+    :param raw_provider_id: The full provider identifier, starting with 'repository-url:'.
+    :param logger: Logger instance for outputting messages.
+    :raises ValueError: If the provider identifier format is invalid or the baseurl is missing.
+    """
```
LecrisUT, line 40: Do note that this docstring would need to be refactored when the documentation of `PrepareArtifact` is made dynamic. So I am not actually reviewing this.

tmt/steps/prepare/artifact/providers/repository_url.py
```diff
@@ -0,0 +1,108 @@
...
+
+    repository: Repository
+
+    def __init__(self, raw_provider_id: str, logger: tmt.log.Logger):
+        super().__init__(raw_provider_id, logger)
```
LecrisUT, line 48: Can be dropped

tmt/steps/prepare/artifact/providers/repository_url.py
vaibhavdaren: Once that PR is merged , I can use the `DEFAULT_REPOSITORY_PRIORITY` constant. All that would be left would be changes to the `repository-url` provider related to priority.

tmt/steps/prepare/artifact/providers/repository_url.py
```diff
@@ -0,0 +1,108 @@
...
+        # Repository provider does not enumerate individual artifacts.
+        # The repository is installed and packages are available through the package manager.
+        # There is no need to download individual artifact files.
+        return []
+
+    def _download_artifact(
+        self, artifact: RpmArtifactInfo, guest: Guest, destination: tmt.utils.Path
+    ) -> None:
+        """This provider only discovers repos; it does not download individual RPMs."""
+        raise AssertionError("RepositoryUrlProvider does not support downloading individual RPMs.")
```
AthreyVinay, line 72: @LecrisUT Questioning the use of `AssertionError` as opposed to  `NotImplementedError` - looks odd for a runtime behavior. This is copy-pasted for most of the providers - I am guilty of it too.

LecrisUT: To me `AssertionError` seems more appropriate because we do not expect this function to ever be called. `NotImplementedError` reads more as eventually we would address it. It is not codified one way or another, but it would be good if we can have a more elevated error reporting for `AssertionError`, `assert`, etc. since if those fail we really need to be made aware of.

tests/prepare/artifact/providers/repository-url/main.fmf
vaibhavdaren: @LecrisUT  : I think, I should remove the tests from this PR and in a subsequent PR add tests for the repository-url provider.  Once #4449  is done. It will be ez to add tests for any provider.

LecrisUT: Fine with that

tmt/steps/prepare/artifact/providers/repository_url.py
AthreyVinay: I agree & disagree with the above statement:

Agreement: `NotImplementedError` definitely "indicates while the class is being developed to indicate that the real implementation still needs to be added. " Had to read through the old docs again!
So, yeah, `NotImplementedError` out of picture.

Disagreement: `AssertionError` is raised when an assert fails, so in practice it’s for internal sanity checks.  Personally, feel this is not appropriate.

A middle ground, creating a custom exception `UnsupportedOperationError`. I have pushed a commit against my own PR: [a75c22a](https://github.com/teemtee/tmt/pull/4471/commits/a75c22ab671da677a807081c145b4a660cf1b156). WDYT?

thrix: Custom exception sounds better to me

thrix (approved):
LGTM, I would expect some tests though!

vaibhavdaren:
> LGTM, I would expect some tests though!

I totally agree. I have removed the tests added in https://github.com/teemtee/tmt/pull/4474/changes/a1f747e12f4bb900cabe07be431b3f058f8279ce. I intend to re-add them once PR #4449 is merged (In a subsequnt PR) so that we can utilize the underlying common functions introduced there.

tmt/steps/prepare/artifact/providers/repository_url.py
vaibhavdaren: @AthreyVinay: Added a FIXME note to be  Addressed in subsequent PR. Commit : https://github.com/teemtee/tmt/pull/4474/changes/26b93a781f2d28cb822640a95de85216c615d436. I think we can discuss this but this does not seem like a blocker.

AthreyVinay: No discussion needed ;) The PR is good enough for me.

LecrisUT: There is no consensus on the usage of `AssertionError` and `NotImplementedError`, so it's a project-by-project basis on what meaning these can have.

We ourselves do not do special handling of these error types, but maybe we should and having a document in contribution section of how and we handle these would be helpful.

As our users are primarily interfacing us via the CLI and fmf, I will borrow some design patterns from CMake and how to split the error types:
- Warnings including user definition errors, deperecations, etc.
- Assertion-like errors: Things that if the code entered we know it is a bug on our side
- User facing errors: Coming from missing field in fmf, network issues etc
- Author/Maintainer errors: Errors relevant primarily to the maintainer of the fmf file. This is overlapping with deprecation warnings and user facing errors, but more focused on who should be pinged about these issues

thrix:
Tests still running due to https://github.com/packit/packit-service/issues/2799

Issue #3618, opened 2025-03-21
tmt lint: should fail if trying to extend a key when parent doesn exit


bgoncalv:
When trying to use `+` to add key to an item not previously defined, tmt lint doesn't fail. What is confusing for the user perspective as the key/value is not set to the plan.

example: 

```
discover:
    how: fmf

execute:
    how: tmt

/bare:
    adjust:
      - when: package-name ~= kernel-64k
        provision+:
            hardware+:
                virtualization:
                    is-virtualized: false
```

provision is not set, therefore ` hardware+:` is not set, but there is no error indicating this is invalid.

If user set, just ` hardware:` it works:
```
            hardware:
                virtualization:
                    is-virtualized: false
```

psss:
The `lint` actually reported the problem:

    warn C000 fmf node failed schema validation

Could we change this to fail? Possible problems with short plugin names `virtual` vs. `virtual.testcloud`?

Decision: Convert the check from `warn` to `fail` & announce the change in the release notes.

psss:
Not sure if @bajertom will be around for the next release, dropping to backlog for now.

tcornell-bus:
Hi @psss and @happz based on today's discussion in slack, I came back to this issue to propose reverting this fix for the time being while we fix the issue more holistically, such as the linter hints @happz suggested. But one thing is confusing me about this issue: tmt lint does not warn about the actual issue (which is the `hardware+` key) during linting. 

Looking at line [1075 in tmt/base.py](https://github.com/teemtee/tmt/blob/1af4d82c3acc53ff7c87ed86145d289acaaf48c5/tmt/base.py#L1075) it should be warned about, but the regex is missing the `+` char. When I add it to the regex (making it `([a-zA-Z0-9', \-+]+)`) I get more helpful warnings like:
```
warn C000 key "hardware+" not recognized by schema /schemas/provision/virtual
```
Is there a reason why the `+` char is not included in the regex? In fact, should the regex just be `(.+)`?

happz:
> Hi [@psss](https://github.com/psss) and [@happz](https://github.com/happz) based on today's discussion in slack, I came back to this issue to propose reverting this fix for the time being while we fix the issue more holistically, such as the linter hints [@happz](https://github.com/happz) suggested. But one thing is confusing me about this issue: tmt lint does not warn about the actual issue (which is the `hardware+` key) during linting.
> 
[...]

PR #4481, opened 2026-01-13
Revert the tmt lint schema failure

tcornell-bus:
The original intention was to implement what is said in this [comment](https://github.com/teemtee/tmt/pull/4445#issuecomment-3744315786). The issue I ran into is that when I have `tmt lint` emit a schema (pass or warn) if the only error is the `detect_missing_required_properties`, then it still fails because almost always this check `detect_enum_violations` also fails when a required key is missing. `detect_enum_violations` is triggered when `how` is used and the required key is missing because it cycles through all of the values and tells you when the first 9/10 did not match. 

Cursor analysis of why this occurs (better explanation than mine):
> Explaining why enum violations only appear when the required property is missing:
> The schema uses `oneOf` for the report step, which requires exactly one schema to match. Here's what happens:
> When the `project` key (required property) is present:
> The `reportportal` schema matches (`how: reportportal` and `project` are present)
> Since `oneOf` requires exactly one match and reportportal matches, validation succeeds
> No enum violations occur because the validator doesn't try other schemas
> When the `project` key (required property) is missing:
> The `reportportal` schema fails (missing required `project`)
> Since `oneOf` requires exactly one match and `reportportal` failed, the validator tries other schemas (display, html, junit, polarion)
> Each alternative fails because `how: reportportal` doesn't match their enum values (they expect how: display, how: html, etc.)
> These enum violations are reported as the validator tries each alternative

So, it's not simple to turn _only_ the missing required key into a warning in `tmt lint` since even if that error is not counted, another error still occurs, still resulting in an overall `tmt lint` failure.

After some discussion with @LecrisUT, @happz, and @thrix, reverting the FAIL to a WARN would unblock the CI failures for #4178, while we work on a better overall solution where #3618 can be reimplemented, and #4178 is still resolved. The ticket for the better solution is #4483 

This PR reverts the tmt lint schema failure implemented to fix issue #3618.

Fixes: #4178

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/tests/lint/all/test.sh
+++ b/tests/lint/all/test.sh
@@ -62,7 +62,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Check --fix for tests"
-        rlRun -s "tmt lint --fix fix" 1 "Fix the test"
+        rlRun -s "tmt lint --fix fix" 0 "Fix the test"
         rlAssertGrep 'relevancy converted into adjust' $rlRun_LOG
         rlRun -s "tmt lint fix" 0 "Should pass after fix"
     rlPhaseEnd
--- a/tests/lint/test/test.sh
+++ b/tests/lint/test/test.sh
@@ -28,7 +28,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Old yaml"
-        rlRun -s "tmt test lint old-yaml" 1
+        rlRun -s "tmt test lint old-yaml" 0
         rlAssertGrep "warn: /old-yaml:enabled - 'yes' is not of type 'boolean'" $rlRun_LOG
         rlAssertGrep 'warn C000 fmf node failed schema validation' $rlRun_LOG
     rlPhaseEnd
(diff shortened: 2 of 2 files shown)


tcornell-bus:
I do want to state an alternative to reverting the FAIL to a WARN (this PR). Instead, we could include enum violations (`detect_enum_violations`) in the "warn only" category, like missing required (`detect_missing_required_properties`). Which would fix the issue I ran into, but I was not prepared to unilaterally decide that enum violations are now only a warning as well.

The discussion in slack led to this PR, but I do not think I explained this alternative solution there.

LecrisUT:
> I do want to state an alternative to reverting the FAIL to a WARN (this PR). Instead, we could include enum violations (`detect_enum_violations`) in the "warn only" category, like missing required (`detect_missing_required_properties`). Which would fix the issue I ran into, but I was not prepared to unilaterally decide that enum violations are now only a warning as well.

About `detect_enum_violations`, if it can be confined to only the `how`, then maybe yes, otherwise those are valid errors.

happz:
@tcornell-bus please, check the tests.

psss:
> Fixes: #4178

So, #4178 should be added to the sprint as well, right?

happz:
> > Fixes: #4178
> 
> So, #4178 should be added to the sprint as well, right?

Yes. Added to the sprint.

[osh-diff-scan:fedora-rawhide-x86_64](https://github.com/teemtee/tmt/pull/4481/checks?check_run_id=60663123374) is hardly a critical check, will make sure it passes eventually. Merging.

PR #4486, opened 2026-01-14
To/from "state" helpers for tmt state files

happz:
tmt saves its state info - step data, test data, guest data, etc. - in YAML format, plent yof `*.yaml` files in a workdir. It's fine, it's readable by both machines and humans, it works.

It turns out YAML is not necessarily the best format out there, as YAML can be costly to generate when dealing with large structures and large collections.

Two things happen in this PR:

* To allow some experimentation, `TMT_STATE_FORMAT` environment variable tells tmt which format to use when storing its data. As of now, only `yaml` is supported.
* "to/from JSON" helpers are refactored to align with recent changes to "to/from YAML" helpers, to provide similar function names, annotations and so on, and they are made available for selection via `TMT_STATE_FORMAT`, making JSON the second supported format for tmt state files.

Moar formats, moar fun!

A set of simple helpers, `to_state` and `from_state` is defined, based on the aforementioned envvar, and patch changes all places where tmt stores state on disk to use them instead of `{to,from}_yaml`. `Common.{read,write}_state()` will help with that, instead of `Common.{read,write}()` methods.

Now we can play with different format, and even if we learn the format is not the culprit of ineffective storage of large collections of results, we will have nice and tidy internal API to work with.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [ ] include a release note

### DIFF
--- a/tmt/base/core.py
+++ b/tmt/base/core.py
@@ -80,6 +80,7 @@ from tmt.utils import (
     HasRunWorkdir,
     Path,
     ShellScript,
+    StateFormat,
     WorkdirArgumentType,
     normalize_shell_script,
     to_yaml,
@@ -2934,6 +2935,74 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
 
         return self.workdir
 
+    @functools.cached_property
+    def state_format_marker_filepath(self) -> Path:
+        return self.run_workdir / 'state-format'
+
+    @functools.cached_property
+    def state_format(self) -> StateFormat:
+        try:
+            format_name = self.state_format_marker_filepath.read_text().strip()
+
+        except FileNotFoundError:
+            state_format = tmt.utils.get_state_format()
+
+            self.debug(
+                "No state format marker file found,"
+                f" using the default state format '{state_format.name}'."
+            )
+
+            return state_format
+
+        except Exception as exc:
+            raise GeneralError('Failed to read state format marker.') from exc
+
+        state_format = tmt.utils.get_state_format(format=format_name)
+
+        self.debug(
+            f"State format marker file found, using the '{state_format.name}' state format."
+        )
+
+        return state_format
+
+    def read_state(self, filepath: Path) -> Any:
+        """
+        Read a stored state from the given file.
+
+        .. important::
+
+            No deserialization is performed, it is the responsibility of the
+            caller to turn loaded structure, consisting of built-in-like
+            types, into objects of desired classes, e.g. by the power of
+            :py:meth:`tmt.container.SerializableContainer.deserialize`.
+
+        :param filepath: file to read the state from.
+        :returns: stored state as Python data structure.
+        """
+
+        return self.state_format.from_state(
+            self.read_file(Path(f'{filepath}{self.state_format.suffix}'))
+        )
+
+    def write_state(self, filepath: Path, data: Any) -> None:
+        """
+        Write a state into the given file.
+
+        .. important::
+
+            No serialization is performed, it is the responsibility of the
+            caller to turn internal objects into built-in-like Python types,
+            e.g. by the power of
+            :py:meth:`tmt.container.SerializableContainer.serialize`.
+
+        :param filepath: file to write the state into.
+        :param data: state as Python data structure.
+        """
+
+        return self.write_file(
+            Path(f'{filepath}{self.state_format.suffix}'), self.state_format.to_state(data)
+        )
+
     def load_workdir(self, *, with_logfiles: bool = True) -> None:
         """
         Prepare the run workdir and associated.
@@ -3078,8 +3147,13 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         """
         Save list of selected plans and enabled steps
         """
+
+        self.state_format_marker_filepath.unlink(missing_ok=True)
+        self.state_format_marker_filepath.write_text(self.state_format.name)
+
         assert self.tree is not None  # narrow type
         assert self._cli_context_object is not None  # narrow type
+        assert self.workdir is not None  # narrow type
         data = RunData(
             root=str(self.tree.root) if self.tree.root else None,
             plans=[plan.name for plan in self._plans] if self._plans is not None else None,
@@ -3087,7 +3161,7 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
             environment=self.environment,
             remove=self.remove,
         )
-        self.write_state('run', data.to_serialized())
+        self.write_state(self.workdir / 'run', data.to_serialized())
 
     def load_from_workdir(self) -> None:
         """
@@ -3105,8 +3179,11 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         #  which in turn is only called by status and clean, both cases where we do not want
         #  to attach the logfile loggers to.
         self.load_workdir(with_logfiles=False)
+
+        assert self.workdir is not None  # narrow type
+
         try:
-            self.data = RunData.from_serialized(self.read_state('run'))
+            self.data = RunData.from_serialized(self.read_state(self.workdir / 'run'))
 
         except tmt.utils.FileError:
             self.debug('Run data not found.')
@@ -3136,8 +3213,10 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         """
         from tmt.base.plan import Plan
 
+        assert self.workdir is not None  # narrow type
+
         try:
-            self.data = RunData.from_serialized(self.read_state('run'))
+            self.data = RunData.from_serialized(self.read_state(self.workdir / 'run'))
 
         except tmt.utils.FileError:
             self.debug('Run data not found.')
--- a/tmt/base/core.py
+++ b/tmt/base/core.py
@@ -82,7 +82,6 @@ from tmt.utils import (
     ShellScript,
     WorkdirArgumentType,
     normalize_shell_script,
-    to_state,
     to_yaml,
     verdict,
 )
--- a/tmt/base/core.py
+++ b/tmt/base/core.py
@@ -82,6 +82,7 @@ from tmt.utils import (
     ShellScript,
     WorkdirArgumentType,
     normalize_shell_script,
+    to_state,
     to_yaml,
     verdict,
 )
@@ -3087,7 +3088,7 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
             environment=self.environment,
             remove=self.remove,
         )
-        self.write(Path('run.yaml'), to_yaml(data.to_serialized()))
+        self.write_state('run', data.to_serialized())
 
     def load_from_workdir(self) -> None:
         """
@@ -3106,9 +3107,8 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         #  to attach the logfile loggers to.
         self.load_workdir(with_logfiles=False)
         try:
-            self.data = RunData.from_serialized(
-                tmt.utils.yaml_to_dict(self.read(Path('run.yaml')))
-            )
+            self.data = RunData.from_serialized(self.read_state('run'))
+
         except tmt.utils.FileError:
             self.debug('Run data not found.')
             return
@@ -3138,9 +3138,8 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         from tmt.base.plan import Plan
 
         try:
-            self.data = RunData.from_serialized(
-                tmt.utils.yaml_to_dict(self.read(Path('run.yaml')))
-            )
+            self.data = RunData.from_serialized(self.read_state('run'))
+
         except tmt.utils.FileError:
             self.debug('Run data not found.')
             return
--- a/tmt/steps/cleanup/__init__.py
+++ b/tmt/steps/cleanup/__init__.py
@@ -77,7 +77,7 @@ class Cleanup(tmt.steps.StepWithQueue[CleanupStepData, PluginOutcome]):
         A set of members of the step workdir that should not be removed.
         """
 
-        return {*super()._preserved_workdir_members, 'results.yaml'}
+        return {*super()._preserved_workdir_members, f'results{tmt.utils.STATE_FILENAME_SUFFIX}'}
 
     def wake(self) -> None:
         """
--- a/tmt/steps/cleanup/__init__.py
+++ b/tmt/steps/cleanup/__init__.py
@@ -77,7 +77,14 @@ class Cleanup(tmt.steps.StepWithQueue[CleanupStepData, PluginOutcome]):
         A set of members of the step workdir that should not be removed.
         """
 
-        return {*super()._preserved_workdir_members, f'results{tmt.utils.STATE_FORMAT.suffix}'}
+        members = {
+            *super()._preserved_workdir_members,
+        }
+
+        if self.plan.my_run:
+            members = {*members, f'results{self.plan.my_run.state_format.suffix}'}
+
+        return members
 
     def wake(self) -> None:
         """
--- a/tmt/steps/cleanup/__init__.py
+++ b/tmt/steps/cleanup/__init__.py
@@ -77,7 +77,7 @@ class Cleanup(tmt.steps.StepWithQueue[CleanupStepData, PluginOutcome]):
         A set of members of the step workdir that should not be removed.
         """
 
-        return {*super()._preserved_workdir_members, f'results{tmt.utils.STATE_FILENAME_SUFFIX}'}
+        return {*super()._preserved_workdir_members, f'results{tmt.utils.STATE_FORMAT.suffix}'}
 
     def wake(self) -> None:
         """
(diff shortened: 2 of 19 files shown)


tmt/steps/cleanup/__init__.py
```diff
@@ -78,7 +78,7 @@ def _preserved_workdir_members(self) -> set[str]:
         A set of members of the step workdir that should not be removed.
         """
 
-        return {*super()._preserved_workdir_members, 'results.yaml'}
+        return {*super()._preserved_workdir_members, f'results{tmt.utils.STATE_FILENAME_SUFFIX}'}
```
LecrisUT: What would happen if you continue a run and forgot to set `TMT_STATE_FORMAT` consistently. I think it's better for `Run` to own this info.

tmt/utils/__init__.py
```diff
@@ -2396,47 +2396,111 @@ def run(
...
 
         except OSError as error:
-            raise FileError(f"Failed to write into '{path}' file.") from error
+            raise FileError(f"Failed to write into '{filepath}'.") from error
+
+    # TODO: too many of these in the code, let's get rid of them later.
+    read = read_file
+    write = write_file
+
+    def read_state(self, name: str) -> Any:
```
LecrisUT: Can we move these to `Run` for the same reason mentioned in the other comment?

tmt/steps/cleanup/__init__.py
happz: Second run would fail because of being unable to find the expected `run.$suffix` file.

tmt/utils/__init__.py
happz: I'd prefer leaving that for later, as all these methods are now attached to `Common` and used by many objects via `self`. I'm already dissecting `Common` object into smaller pieces, this could be one of such PRs.

tmt/steps/cleanup/__init__.py
LecrisUT: Don't remember if I had another comment that mysteriously vanished, but the idea I wanted to highlight here is that this variable can be owned by `Run` and read the state of `run.(yaml|json)` for the current state that it should (re-)use moving forward. On a new run, it can use the similar logic to take it from the env variable

tmt/utils/__init__.py
LecrisUT: But all of the `read_state` are new occurrences, so those should be fine to be rehomed right? I realized that Github highlighting makes it ambiguous, I mean `read_state` to be moved, not `read/write`

tmt/steps/cleanup/__init__.py
happz: Yeah, but it's a lot of wiring. I'd like to separate these methods from `Common` completely, then `Run` can do this magic and just pass this "read/write state" functionality to its children. It's also very related to workdirs I'm trying to cut out, as state is pretty exclusively stored in workdirs, so that's also an option, to have this functionality attached to a workdir (and configured by `Run`). So, TL;DR, yeah, I'm on board, it's just all so entangled :(

tmt/utils/__init__.py
happz: Hm, `read_state`/`write_state` could be easier, I can give it a try.

See also https://github.com/teemtee/tmt/pull/4486#discussion_r3071588366

tmt/steps/cleanup/__init__.py
LecrisUT: But unless Github search is lying, all of the current usages are navigable to the `run` object. The only exception is `read/write_state`, so I guess it's fine if that is the prerequisite for this.

tmt/utils/__init__.py
happz: > Hm, `read_state`/`write_state` could be easier, I can give it a try.

Hmmm, but then again, there's `self.read()` used, that would need to traverse its parents to find the `Run` instance to call its `read_state()`, but since `read()` is defined in `Common` and `read_state()` would be in `Run`, this would be base calling child implementation. `Common` has no notion of what hierarchy these tests and plans and runs build upon `Common` base :/

I would have to ut out `read/write` out of `Common` for it to become aware of this relationship, which makes it https://github.com/teemtee/tmt/pull/4486#discussion_r3071588366: it's all entangled, 100% needs to be separated, it will happen, but it seems to go beyond the scope of this simpler PR.

tmt/steps/cleanup/__init__.py
happz: Ok, if you'd insist rewiring `read/write` needs to happen first, then this PR would need to go out of the sprint, as it would be blocked on that, and workdir refactoring.  I can't see the move of `read_state/write_state` under `Run` as easy as it would make `Common.read` depend on `Run.read_state`, i.e. base would rely on child class, which I think would not be right. Therefore moving these new methods to `Run` is IMO not doable today in a simple way.

tmt/utils/__init__.py
LecrisUT: > Hmmm, but then again, there's `self.read()` used, that would need to traverse its parents to find the `Run` instance to call its `read_state()`, but since `read()` is defined in `Common` and `read_state()` would be in `Run`, this would be base calling child implementation.

You lost me, I don't see how `self.read` is calling `read_state`, just the other way around. Yes, the workdir is ambiguous, but my thought is that you could have a property in `Run` that would get only this suffix from either existing run workdir or env variable. I think it is rather self-contained, but if I am missing something, than I am fine with deferring it.

happz: Hmmm, maybe I need to read it few more times. Let's see after the lunch, I'll start from the scratch.

happz: > ... but my thought is that you could have a property in `Run` that would get only this suffix from either existing run workdir or env variable.

Slightly more complicated: invoke `TMT_STATE_FORMAT=json tmt run ...`, then invoke `tmt run --last ...` - the format from envvar would be the default, `yaml`, the one saved in `run.json` would be `json`, but to load that value would require loading `run.json` into `RunData`, but since the active format is now `yaml`, this would fail... So either some kind of discovery, e.g. trying out formats one by one till the first one succeeds, or saving the format in a dedicated file to be read first.

happz: An overhaul in fcd37385: per-run state format is now saved in a dedicated file in rundir; it's loaded and used by each run, initialized by the default format. Plus plenty more docstrings, new type, and so on. Seems a lot cleaner to me.

tmt/steps/finish/__init__.py
```diff
@@ -67,7 +67,7 @@ def _preserved_workdir_members(self) -> set[str]:
         A set of members of the step workdir that should not be removed.
         """
 
-        return {*super()._preserved_workdir_members, 'results.yaml'}
+        return {*super()._preserved_workdir_members, f'results{tmt.utils.STATE_FORMAT.suffix}'}
```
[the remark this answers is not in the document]
happz: Yeah, addressed in bb3203f3.

tmt/utils/__init__.py
```diff
@@ -3715,21 +3759,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+
+    except Exception as exc:
+        raise GeneralError(f"tmt state file format '{format}' is not supported.") from exc
+
+
+#: The default state format. It is initialized via ``TMT_STATE_FORMAT``
+#: environment variable.
+STATE_FORMAT: Final[StateFormat] = get_state_format(
+    format=os.environ.get('TMT_STATE_FORMAT', 'yaml').lower()
+)
```
LecrisUT, line 3889: Please move this above the function and rename it to something like `STATE_FORMAT_FROM_ENV`

LecrisUT (approved):

tmt/utils/__init__.py
happz: * since this calls the `get_state_format()` function, it cannot be moved above said function.
* I prefer `STATE_FORMAT`, it's **the** state format to use. It could be `DEFAULT_STATE_FORMAT` as well, but I don't want it to announce how it's created, i.e. `_FROM_ENV` seems needles because 1. there's no other `STATE_FORMAT` to distinguish from on this level, and 2. consumer of this variable should not worry about which `(DEFAULT_)STATE_FORMAT` to use.

LecrisUT: The reason is that with all the `StateFormat`, `_STATE_FORMATS`, `STATE_FORMAT` that look to similar it is hard to read, initially I thought it was a typo. Any other option to not have so many similar names would be good, maybe consolidating these as `ClassVar` to `StateFormat`

happz: Renamed `_STATE_FORMATS` to `_SUPPORTED_STATE_FORMATS`, `STATE_FORMAT` to `DEFAULT_STATE_FORMAT` in b767c151. Can't do much about `StateFormat` though. I'm not fond of class vars, so leaving these as module-level items until we decide if/whether it's time to move module-level items into classes in general.

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+        one.
+    :returns: stored state as Python data structure.
+    """
+
+    state_format = get_state_format(format=format_name)
+
+    return state_format.from_state(Path(f'{filepath}{state_format.suffix}').read_text())
+
+
+def write_state(filepath: Path, data: Any, format: Optional[str] = None) -> None:
```
tcornell-bus: minor: Should `format` be `format_name` to match the other functions?

tmt/steps/finish/__init__.py
```diff
@@ -67,7 +67,10 @@ def _preserved_workdir_members(self) -> set[str]:
         A set of members of the step workdir that should not be removed.
         """
 
-        return {*super()._preserved_workdir_members, 'results.yaml'}
+        return {
+            *super()._preserved_workdir_members,
+            f'results{tmt.utils.DEFAULT_STATE_FORMAT.suffix}',
```
tcornell-bus: The other steps use `self.plan.my_run.state_format.suffix` but this one is using `DEFAULT_STATE_FORMAT` directly, which seems like a bug. Cursor pointed this out and I agree.

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+}
+
+
+def get_state_format(format: Optional[str] = None) -> StateFormat:
+    """
+    Get a state format by the given name.
+
+    :param format: if set, the format of matching name is looked up.
+    :returns: state format by the name of ``format``, or the default
+        format, :py:data:`STATE_FORMAT`.
```
tcornell-bus:
```suggestion
        format, :py:data:`DEFAULT_STATE_FORMAT`.
```
Minor docsting issue

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+    Turn Python data structure into a stored state.
+
+    .. important::
+
+        No serialization is performed, it is the responsibility of the
+        caller to turn internal objects into built-in-like Python types,
+        e.g. by the power of
+        :py:meth:`tmt.container.SerializableContainer.serialize`.
+
+    :param data: Python data structure.
```
tcornell-bus, line 3925:
```suggestion
    :param state: Python data structure.
```
Minor: the actual parameter is named `state`

tmt/base/core.py
```diff
@@ -2932,6 +2933,74 @@ def run_workdir(self) -> Path:
...
+    def write_state(self, filepath: Path, data: Any) -> None:
+        """
+        Write a state into the given file.
+
+        .. important::
+
+            No serialization is performed, it is the responsibility of the
+            caller to turn internal objects into built-in-like Python types,
+            e.g. by the power of
+            :py:meth:`tmt.container.SerializableContainer.serialize`.
```
tcornell-bus, line 2996:
```suggestion
            :py:meth:`tmt.container.SerializableContainer.to_serialized`.
```
should this be `to_serialized`?

tmt/base/core.py
```diff
@@ -2932,6 +2933,74 @@ def run_workdir(self) -> Path:
...
+    def read_state(self, filepath: Path) -> Any:
+        """
+        Read a stored state from the given file.
+
+        .. important::
+
+            No deserialization is performed, it is the responsibility of the
+            caller to turn loaded structure, consisting of built-in-like
+            types, into objects of desired classes, e.g. by the power of
+            :py:meth:`tmt.container.SerializableContainer.deserialize`.
```
tcornell-bus, line 2977:
```suggestion
            :py:meth:`tmt.container.SerializableContainer.from_serialized`.
```

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+def write_state(filepath: Path, data: Any, format: Optional[str] = None) -> None:
+    """
+    Write a state into the given file.
+
+    .. important::
+
+        No serialization is performed, it is the responsibility of the
+        caller to turn internal objects into built-in-like Python types,
+        e.g. by the power of
+        :py:meth:`tmt.container.SerializableContainer.serialize`.
```
tcornell-bus:
```suggestion
        :py:meth:`tmt.container.SerializableContainer.to_serialized`.
```

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+def to_state(state: Any, format_name: Optional[str] = None) -> str:
+    """
+    Turn Python data structure into a stored state.
+
+    .. important::
+
+        No serialization is performed, it is the responsibility of the
+        caller to turn internal objects into built-in-like Python types,
+        e.g. by the power of
+        :py:meth:`tmt.container.SerializableContainer.serialize`.
```
tcornell-bus:
```suggestion
        :py:meth:`tmt.container.SerializableContainer.to_serialized`.
```

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+def read_state(filepath: Path, format_name: Optional[str] = None) -> Any:
+    """
+    Read a stored state from the given file.
+
+    .. important::
+
+        No deserialization is performed, it is the responsibility of the
+        caller to turn loaded structure, consisting of built-in-like
+        types, into objects of desired classes, e.g. by the power of
+        :py:meth:`tmt.container.SerializableContainer.deserialize`.
```
tcornell-bus:
```suggestion
        :py:meth:`tmt.container.SerializableContainer.from_serialized`.
```

tmt/utils/__init__.py
```diff
@@ -3723,21 +3767,212 @@ def yaml_to_list(data: str, *, yaml_type: Optional[YamlTypType] = 'safe') -> lis
...
+def from_state(state: str, format_name: Optional[str] = None) -> Any:
+    """
+    Turn stored state into Python data structure.
+
+    .. important::
+
+        No deserialization is performed, it is the responsibility of the
+        caller to turn loaded structure, consisting of built-in-like
+        types, into objects of desired classes, e.g. by the power of
+        :py:meth:`tmt.container.SerializableContainer.deserialize`.
```
tcornell-bus:
```suggestion
        :py:meth:`tmt.container.SerializableContainer.from_serialized`.
```

tcornell-bus (commented):
Mostly docsting update requests since some things in the PR were changed and a possible minor bug.

tmt/utils/__init__.py
happz: Correct, addressed in 6e57372d.

tmt/steps/finish/__init__.py
happz: Fixed in 6e57372d.

tmt/utils/__init__.py
happz: Fixed in 6e57372d.

tmt/utils/__init__.py
happz: Fixed in 6e57372d.

tmt/base/core.py
happz: Fixed in 6e57372d.

tmt/base/core.py
happz: Fixed in 6e57372d.

tmt/utils/__init__.py
happz: Fixed in 6e57372d.

tmt/utils/__init__.py
happz: Fixed in 6e57372d.

tmt/utils/__init__.py
happz: Fixed in 6e57372d.

tmt/utils/__init__.py
happz: Fixed in 6e57372d.

tcornell-bus (approved):

PR #4497, opened 2026-01-15
Adding a basic code profiler

happz:
Easy to use as an decorator, for ad hoc measurements.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/utils/_profiler.py
+++ b/tmt/utils/_profiler.py
@@ -1,0 +1,143 @@
+import cProfile
+import datetime
+import functools
+import io
+import pstats
+import time
+from collections.abc import Iterator
+from typing import Callable, Optional, TypeVar
+
+# Using bootstrap logger on purpose: a profiler can be invoked while
+# functions, classes and methods are still being defined and imported,
+# and the proper logger may not be available yet.
+from tmt._bootstrap import _BOOTSTRAP_LOGGER
+from tmt._compat.typing import ParamSpec
+from tmt.container import container
+
+T = TypeVar('T')
+P = ParamSpec('P')
+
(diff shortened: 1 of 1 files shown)


LecrisUT:
What are the plans with this, vs running `python -m cProfile -m tmt` directly?

happz:
> What are the plans with this, vs running `python -m cProfile -m tmt` directly?

That will certainly work, but it's terribly noisy. When interested where time is spent during `virtual` provisioning, decorating only the interesting methods for the sake of investigation seems easier to ingest.

LecrisUT:
I'm not sure how noisy it would be, it should be just a sub-stack in the flame graph, but probably it would make the code too slow overall and harder to actually create the profile graph.

tmt/utils/_profiler.py
```diff
@@ -0,0 +1,143 @@
...
+
+                    return fn(*args, **kwargs)
+
+                finally:
+                    profiler.stop()
+
+                    _BOOTSTRAP_LOGGER.info(
+                        f"# Profiler report on '{fn.__name__}'",
+                        '\n'.join(profiler.format(sort_stats=sort_stats, limit=limit)),
+                    )
```
LecrisUT, line 139: Could we output to a file instead of the logger? At the very least PyCharm would support opening it with `Tools -> Open CProfile snapshot`

LecrisUT (approved):

tmt/utils/_profiler.py
happz: It should be possible. Do you have a specific format in mind? Which file?

Issue #4440, opened 2025-12-17
Library conflicts with different url


HouMinXi:
if I define a library which come from different git repository, tmt run discover will raise error.
```
houminxi@houminxi:~/code/kernel$ pwd
/home/houminxi/code/kernel  #<------tmt-root path
houminxi@houminxi:~/code/kernel$ tmt run discover --how fmf --path . --name /networking/openvswitch/topo
/var/tmp/tmt/run-027

/default/plan
    discover
        how: fmf
        directory: .
        cloned-commit-hash: 1f4bf067
        cloned-commit-hash: a55bfd6

plan failed.

The exception was caused by 1 earlier exceptions

Cause number 1:

    Library 'kernel/common' with url 'https://gitlab.cee.redhat.com/mhou/kernel' conflicts with already fetched library from 'https://gitlab.cee.redhat.com/kernel-qe/kernel'.
```
I have my private repository(https://gitlab.cee.redhat.com/mhou/kernel) which frok from https://gitlab.cee.redhat.com/kernel-qe/kernel
I define /networking/openvswitch/topo/main.fmf as below:
```
summary: ovs topo testing
description: |
    FIXME: more description about ovs topo
contact: Rick Alongi <ralongi@redhat.com>
component:
  - openvswitch
test: bash ./runtest.sh
framework: beakerlib
require:
  - type: library
    url: https://gitlab.cee.redhat.com/mhou/kernel
    name: /networking/common
    ref: mhou_tmp_branch
  - type: library
    url: https://gitlab.cee.redhat.com/mhou/kernel
    name: /networking/openvswitch/common
    ref: mhou_tmp_branch
duration: 24h
extra-summary: /kernel/networking/openvswitch/topo
```
extra-task: /kernel/networking/openvswitch/topo

PR #4499, opened 2026-01-15
Refactor Library classes

LecrisUT:
To summarize the changes here:
- Make Library and sub-classes into `@container`, moving all of the `__init__` and `library_factory` logic into a constructor `cls.from_identifier`
- Split the `BeakerLib` class into distinct `BeakerLibUrl` and `BeakerLibPath` with more fine-grained type-checking
- Split the `__init__` and `fetch` logic in `BeakerLib` depending on each type that it can be

A precursor to #4440.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/libraries/beakerlib.py
+++ b/tmt/libraries/beakerlib.py
@@ -29,6 +29,7 @@ LIBRARY_REGEXP = re.compile(r'^library\(([^/]+)(/[^)]+)\)$')
 DEFAULT_REPOSITORY_TEMPLATE = 'https://github.com/beakerlib/{repository}'
 DEFAULT_DESTINATION = 'libs'
 
+# TODO: This can probably be dropped? Why do we strip the .git only for some?
 # List of git forges for which the .git suffix should be stripped
 STRIP_SUFFIX_FORGES = [
     'https://github.com',
@@ -347,8 +348,8 @@ class BeakerLibFromUrl(BeakerLib):
         # Strip the '.git' suffix from url for known forges
         url = identifier.url
         for forge in STRIP_SUFFIX_FORGES:
-            if url.startswith(forge) and url.endswith('.git'):
-                url = url.rstrip('.git')
+            if url.startswith(forge):
+                url = url.removesuffix('.git')
         repo = identifier.nick
         if not repo:
             repo_search = re.search(r'/([^/]+?)(/|\.git)?$', url)


tmt/libraries/beakerlib.py
```diff
@@ -465,3 +298,248 @@ def fetch(self) -> None:
...
+    # TODO: url is actually a required field, but we cannot use kw_only yet.
+    #: Full git repository url
+    url: str = ""
+    #: Git revision (branch, tag or commit)
+    ref: Optional[str] = None
+    path: Optional[Path] = None
+
+    def __post_init__(self) -> None:
+        # Set default source directory, used for files required by a library
+        self.source_directory: Path = self.path or self.fmf_node_path
```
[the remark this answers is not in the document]
LecrisUT: False, the parent does not have `__post_init__`

happz:
This PR did not exist when the current sprint was planned this afternoon, and it does not refer to an issue planned for this sprint. What issue is this related to?

LecrisUT:
> This PR did not exist when the current sprint was planned this afternoon, and it does not refer to an issue planned for this sprint. What issue is this related to?

It's precursor to #4440.

tmt/libraries/__init__.py
```diff
@@ -33,31 +28,89 @@ class LibraryError(Exception):
...
     """
 
-    def __init__(
-        self,
+    #: Optional 'parent' object inheriting from tmt.utils.Common can be
+    #: provided in order to share the cache of already fetched libraries.
+    parent: tmt.utils.Common
+    _logger: tmt.log.Logger
+    identifier: Dependency
+    format: str
```
psss, line 48: Could you please add a short docstring for `format`? It's not obvious at the first sight.

tmt/libraries/beakerlib.py
```diff
@@ -465,3 +298,244 @@ def fetch(self) -> None:
             cast(Optional['_RawDependency'], library_node.get('recommend', [])),
             self.parent._logger,
         )
+
+
+@container
+class BeakerLibUrl(BeakerLib):
```
psss:
```suggestion
class BeakerLibFromUrl(BeakerLib):
```

At the first sight the name was a bit confusing to me (sounds like the instance is a url), maybe like this it would be a bit more obvious?

tmt/libraries/beakerlib.py
```diff
@@ -465,3 +298,244 @@ def fetch(self) -> None:
...
+        if self.path:
+            tmt.utils.filesystem.copy_tree(
+                clone_dir / self.path.unrooted() / '.fmf',
+                directory / self.path.unrooted() / '.fmf',
+                self._logger,
+            )
+
+
+@container
+class BeakerLibPath(BeakerLib):
```
psss:
```suggestion
class BeakerLibFromPath(BeakerLib):
```

Similarly here.

tmt/libraries/__init__.py
```diff
@@ -33,31 +28,89 @@ class LibraryError(Exception):
...
     General library class
 
     Used as parent for specific libraries like beakerlib and file
     """
 
-    def __init__(
-        self,
+    #: Optional 'parent' object inheriting from tmt.utils.Common can be
+    #: provided in order to share the cache of already fetched libraries.
+    parent: tmt.utils.Common
```
happz, line 42: Is it optional when it's not `Optional`?

tmt/libraries/__init__.py
```diff
@@ -33,31 +28,89 @@ class LibraryError(Exception):
...
+    _logger: tmt.log.Logger
+    identifier: Dependency
+    format: str
+    #: Library prefix (git repository name or nick if provided)
+    repo: Path
+    #: Library suffix (folder containing the library code)
+    name: str
```
happz, line 55: Since you are moving this around, please, add some examples to docstrings. The attribute is called `repo`, it's `Path`, but the docstring says "Library prefix" and "git repository name or nick if provided" - TBH, I have absolutely no idea what the attribute is used for or whether the combination of these statements is even remotely correct.

"folder containing the library code" - it's a folder, but it's `Path`... And now I'm lost.

tmt/libraries/__init__.py
LecrisUT: Copied from the old docstring. No it is not optional and in fact it is guaranteed to be `DiscoverPlugin`. Tried to refactor it to narrow the type, but that patch is much more complicated, so I am deferring that.

tmt/libraries/__init__.py
LecrisUT: Definitely docstring improvement is needed. Left it out for now to make the review simpler as these is only moving code and docstring around. Will make another commit on top of it for the docstring improvements alone.

happz: Yeah, I'm not blaming you, the library code has been an opaque blob for me for years. I was hoping the refactoring could spring up a refresh of docstrings and naming, after which I could maybe catch up with what's happening in that black box :)

tmt/libraries/__init__.py
LecrisUT: Ok, please check the docstring improvements in bef50a3

tmt/libraries/__init__.py
LecrisUT: Clarified it a bit and added a todo note.

tmt/libraries/__init__.py
LecrisUT: > Yeah, I'm not blaming you

Didn't take it as such :)

> I was hoping the refactoring could spring up a refresh of docstrings and naming

Here we are: bef50a3. Hopefully more to come when the logic inside is detangled more, for not just focused on the basics of it.

tmt/libraries/__init__.py
psss: Thanks!

psss (approved):
Thanks for the changes, now looks good to me.

tmt/libraries/beakerlib.py
```diff
@@ -465,3 +299,246 @@ def fetch(self) -> None:
...
+    path: Optional[Path] = None
+
+    @property
+    def hostname(self) -> str:
+        """
+        Get hostname from url or default to local
+        """
+
+        if self.url:
+            matched = re.match(r'(?:git|http|https|ssh)(?:@|://)(.*?)[/:]', self.url)
```
happz, line 328: Since we're making things better, https://docs.python.org/3.9/library/urllib.parse.html would be the right way to parse an URL.

LecrisUT: It will be dropped completely in https://github.com/teemtee/tmt/pull/4503 :)

tmt/libraries/beakerlib.py
```diff
@@ -465,3 +299,246 @@ def fetch(self) -> None:
...
+        assert parent is not None  # narrow type
+        assert isinstance(identifier, DependencyFmfId)  # narrow type
+        assert identifier.url  # narrow type
+        parent.debug(f"Detected library '{identifier.to_minimal_spec()}'.", level=3)
+
+        # Strip the '.git' suffix from url for known forges
+        url = identifier.url
+        for forge in STRIP_SUFFIX_FORGES:
+            if url.startswith(forge) and url.endswith('.git'):
+                url = url.rstrip('.git')
```
therazix: Shouldn't we use `removesuffix` instead? `rstrip` could keep removing letters beyond the `.git` suffix

LecrisUT: Yes, and it is also duplicated with the logic in utils.git, need to change it to use something from there

happz: Will there be a commit for this PR, or a follow-up PR?

LecrisUT: If there is already something in `utils.git` to replace it with, I'll squeeze it in here, otherwise I think better to punt?

happz: Check and decide, please.

LecrisUT: There is no utility for this, but I think the need for this goes away with #4503. Added a note for now and will mull it over. For now, will just use @therazix suggestion

happz: Ok, resolving then.

PR #4500, opened 2026-01-16
Fix Testing Farm API docs link

thrix:
Since the 2026-01.1 release, the API link without a path redirects to the documentation, which is now available under the /docs path. Remove the /redoc suffix as it's no longer needed.


LecrisUT:
Fixes linkcheck so let's get this in for the sprint

happz:
Documentation-only, merging.

PR #4503, opened 2026-01-16
Allow beakerlib libraries with clashing repo and name

### LINKED ISSUE #4440: Library conflicts with different url

LecrisUT:
The original code should already have allowed for libraries with different `name` part but same repo, but as shown in practice that did not happen. This is because the string construction in of `str(Beakerlib)` only took the last part of the `name`, disallowing `/foo/common`, `/bar/common`. But beakerlib does not seem to actually have that restriction.

Summary of changes:
- Reworked how the library's repo are cloned allowing for a big degree of collision
- Allow for whole libraries to collide and take the first definition
- Adjust the order of library fetching to account for the relaxed condition above
- Moved the `non-existing-url` handling to the `git_clone` utility
- Adjusted the display of the beakerlib library to display the full `(repo)/(name)` as it would be used in `rlImport`

Pull Request Checklist

* [x] implement the feature
   (Better repository for testing needed: https://github.com/teemtee/tests/issues/23)
* [x] include a release note

Depends on #4499
Closes #4440

### DIFF
--- a/tmt/libraries/__init__.py
+++ b/tmt/libraries/__init__.py
@@ -196,7 +196,9 @@ def resolve_dependencies(
             #  explicitly outside of the try..except.
             if not isinstance(dependency, DependencySimple):
                 logger.warning(f"Library '{dependency}' failed unexpectedly")
-                continue
+                # TODO: we should not add these to the *_to_install, but currently
+                #  DiscoverPlugin.install_libraries handles the logging of unresolved
+                #  libraries in general.
             if dependency in original_require:
                 require_to_install.add(dependency)
             if dependency in original_recommend:


psss:
Seems there are some [failing](https://artifacts.dev.testing-farm.io/787e85cd-6bc5-46b2-bb07-e6673745aa47/) tests:

```
Library 'openssl/certgen (https://github.com/beakerlib/openssl)' conflicts with previously fetched 'openssl/certgen (https://github.com/beakerlib/openssl#master)'.
:: [ 13:55:37 ] :: [   FAIL   ] :: Command 'tmt run -arvvvddd plan --name 'rpm|fmf|nick|duplicate'' (Expected 0, got 2)
```

LecrisUT:
> Seems there are some [failing](https://artifacts.dev.testing-farm.io/787e85cd-6bc5-46b2-bb07-e6673745aa47/) tests:
> 
> ```
> Library 'openssl/certgen (https://github.com/beakerlib/openssl)' conflicts with previously fetched 'openssl/certgen (https://github.com/beakerlib/openssl#master)'.
> :: [ 13:55:37 ] :: [   FAIL   ] :: Command 'tmt run -arvvvddd plan --name 'rpm|fmf|nick|duplicate'' (Expected 0, got 2)
> ```

Let's postpone this one. The test failures are related to the feature that I was planning to address later one, but it seems I cannot fix the tests without that feature.

> * ~Allow for whole libraries to collide and take the first definition~ Postponed
> 
> * [ ]  ~Adjust the order of library fetching to account for the relaxed condition above~ Postponed

psss:
> Let's postpone this one. The test failures are related to the feature that I was planning to address later one, but it seems I cannot fix the tests without that feature.

Ack, makes sense.

tmt/libraries/beakerlib.py
```diff
@@ -191,94 +166,52 @@ def _do_fetch(self, directory: Path) -> None:
...
+        # TODO: Use phase_workdir instead since this is for sure DiscoverPlugin
+        assert self.parent.workdir
+        local_repo_path = self.parent.workdir / self.dest / self.repo
+        local_library_path = local_repo_path / self.fmf_node_path
+        if cached_library := self._library_cache.get(local_library_path):
+            # TODO: Allow conflicting libraries from dependencies
+            # TODO: consider nullifying the require/recommend lists to avoid circular loops
+            raise tmt.utils.GeneralError(
+                f"Library '{self._show_ref()}' conflicts with previously fetched "
+                f"'{cached_library._show_ref()}'."
```
skycastlelily: it seems that the following comment from ai reviewer is valid:
- **Critical** [Bug]: The `fetch` method raises a `GeneralError` unconditionally if the library is found in `_library_cache`. This breaks idempotency: if two tests require the same library, the second one will fail instead of reusing the existing one.
  - Suggestion: Restore the logic that checks if the cached library matches the requested one (same URL/ref) and reuses it if they match. Only raise an error if they map to the same path but have different origins.

```
**Code example:**
`if cached_library := self._library_cache.get(local_library_path):
    # Check if it is the same library (reuse)
    if isinstance(self, BeakerLibFromUrl) and isinstance(cached_library, BeakerLibFromUrl):
         if self.url == cached_library.url and self.ref == cached_library.ref:
             self.parent.debug(f"Library '{self}' already fetched.", level=3)
             self.tree = cached_library.tree
             return

    # Raise error only if it's a genuine conflict
    raise tmt.utils.GeneralError(...)`
```

LecrisUT: As the TODO notes indicate, yes it must be handled. Initially I hoped I could mimic the older behavior, but it didn't seem possible [^1]. So indeed will have to adress it in this PR. Description will reflect the state accordingly, so check that part.

[^1]: https://github.com/teemtee/tmt/pull/4503#issuecomment-3804557105

skycastlelily: JTBC, and have ai reviewer leave a good impression,you only mentioned  " # TODO: Allow conflicting libraries from dependencies" ,but not no-conflicting  senario, ie , the older " self.parent.debug(f"Library '{self}' already fetched.", level=3)". I guess that's why and what ai reviewer is suggesting :smile:

docs/releases/pending/4503.fmf
```diff
@@ -0,0 +1,10 @@
+/git_clone:
```
happz, line 1: The filename probably can be renamed now.

tmt/libraries/__init__.py
```diff
@@ -150,46 +139,50 @@ def fetch(self) -> None:
...
-    # Cut circular dependencies to avoid infinite recursion
-    def already_fetched(lib: Dependency) -> bool:
-        if not imported_lib_ids:
-            return True
-        return lib not in imported_lib_ids
-
-    to_fetch = original_require + original_recommend
-    for dependency in filter(already_fetched, to_fetch):
-        # Library require/recommend
+    # TODO: These should actually be `set[DependencySimple]`
```
happz, line 180: It'd be nice to share why they aren't...

tmt/libraries/beakerlib.py
```diff
@@ -236,94 +211,55 @@ def _do_fetch(self, directory: Path) -> None:
         """
         The actual beakerlib fetch logic.
         """
-        # TODO: This can be dropped when addressing #4440
+
+    # TODO: Move these inside the identifier
+    @abc.abstractmethod
+    def _show_ref(self) -> str:
```
happz: Doesn't really show anything, IIUIC. Maybe `ref_formatted` or something similar? Plus, it could be a property`, seems to take no input.

tmt/libraries/__init__.py
LecrisUT: This is partly due to the covariant/invariant type-definitions. We could probably detangle it, but would be more complicated patch

happz: Yeah, I guessed it might be related. Would you mind mentioning the reason in the comment? I doubt I would be able to guess it in six months.

tmt/utils/git.py
```diff
@@ -789,6 +793,9 @@ def clone_the_repo(
     if can_change:
         url = clonable_git_url(url)
 
+    if url in NON_EXISTING_GIT_URL:
+        raise tmt.utils.GitUrlError(f"Already know that '{url}' does not exist.")
```
vaibhavdaren, line 797: a better message would be helpful :)  
```suggestion
        raise tmt.utils.GitUrlError(f"Skipping git clone: The repository URL '{url}' is known to be unreachable.")
```

LecrisUT: Copied from the older implementation. I do not like the design overall, this is an internal message, not one that needs to be shouted to the user. If it failed to git clone, it's a failure, and if it hit this one, we did not properly stop at the previous step.

The "skipping git clone" is not guaranteed, we only know that that would happen if we know the context of how this is called. This is a raise exception, so it would be creating an error depending on where it is called and how it is handled.

Something more important would be to navigate backwards and check whether the message is actually used, if it should just be a debug message like before, or something.

tcornell-bus:
Hi @LecrisUT would you mind taking a look at the tests? Failures look related to the PR

LecrisUT:
> Hi @LecrisUT would you mind taking a look at the tests? Failures look related to the PR

Yep, thanks for the ping, I forgot to delete the `-a` which was the main point of running only `discover` :facepalm:

therazix:
Test `/tests/libraries/unresolved` seems to be failing. @LecrisUT could you take a look?

LecrisUT:
> Test `/tests/libraries/unresolved` seems to be failing. @LecrisUT could you take a look?

Seems like fallout from #4430. Moving it back to implement, I'll try yo havr it done tomorrow.

LecrisUT:
@therazix it should be good now.

@bajertom I just noticed
```
/plan/certificate/missing/metadata
    discover
        directory: /home/cle/PycharmProjects/tmt/tests/discover/libraries/data
        how: fmf
        tests: missing/metadata
        cloned-commit-hash: 810f6a9
        fail: Failed to process beakerlib libraries (/) for test '/certificate/missing/metadata'.
```
We need better logging still :sweat_smile:

LecrisUT:
Ok, so the blocking PR for wow is: https://gitlab.cee.redhat.com/baseos-qe/integration_scripts/-/merge_requests/365

LecrisUT:
It's unblocked now

PR #4505, opened 2026-01-18
 Accept default repository priority for repositories created by prepare/artifact

vaibhavdaren:
Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/schemas/prepare/artifact.yaml
+++ b/tmt/schemas/prepare/artifact.yaml
@@ -18,7 +18,7 @@ properties:
   provide:
     $ref: "/schemas/common#/definitions/one_or_more_strings"
 
-  default_repository_priority:
+  default-repository-priority:
     type: integer
 
 required:
--- a/tmt/steps/prepare/artifact/__init__.py
+++ b/tmt/steps/prepare/artifact/__init__.py
@@ -35,8 +35,7 @@ class PrepareArtifactData(PrepareStepData):
         metavar='PRIORITY',
         help="""
             Default priority for created artifact repositories. Lower values mean
-            higher priority in package managers. Default: 50. Also uses environment
-            variable ``TMT_PLUGIN_PREPARE_ARTIFACT_DEFAULT_REPOSITORY_PRIORITY``.
+            higher priority in package managers.
             """,
     )
 


tmt/steps/prepare/artifact/__init__.py
```diff
@@ -29,6 +29,16 @@ class PrepareArtifactData(PrepareStepData):
...
 
+    priority: int = field(
+        default=50,
+        option='--priority',
+        metavar='PRIORITY',
+        help="""
+            Priority for artifact repositories. Lower values mean higher priority
+            in package managers. Default: 50.
+            """,
+    )
```
LecrisUT: I wonder if we should be making this a global property to be set. The definition and usage here is specific to rpms, and if we want to generalize it, then we may hit issues with conflicting definitions, e.g. deb packages use the inverse logic, higher values mean higher priority

The alternative is that we allow the `provide` to be a list of dict and the user can pass it for each provide item. But I don't think that would work on the cli interface though, without some black magic similar to how we handle `--how` flag. On the other hand, I don't think we need to support the detailed definition in the cli, the user could just edit the fmf file if needed, `tmt try` or something else.

tests/unit/artifact/test_copr_repository.py
```diff
@@ -13,7 +13,7 @@
     ],
 )
 def test_valid_repository_id(raw_id, expected, root_logger):
-    provider = CoprRepositoryProvider(raw_id, root_logger)
+    provider = CoprRepositoryProvider(raw_id, priority=50, logger=root_logger)
```
LecrisUT: Please move these to use `artifact_provider` fixture instead.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -29,6 +29,16 @@ class PrepareArtifactData(PrepareStepData):
         normalize=tmt.utils.normalize_string_list,
     )
 
+    priority: int = field(
```
vaibhavdaren: From hacking session , We discussed to change this to : 
change to `default_repository_priority`.

vaibhavdaren: Addressed in 417794048c4d67a32862d272beb1cf86e2d3e954.

tests/unit/artifact/test_copr_repository.py
vaibhavdaren: Addressed in d96954e59ca26783592abc787600d3432c5cf36b
Can send a seprate PR for this if required and we want to keep this less chunky

tests/unit/artifact/test_copr_repository.py
```diff
@@ -28,23 +28,23 @@ def test_valid_repository_id(raw_id, expected, root_logger):
 )
 def test_invalid_repository_id(raw_id, error, root_logger):
     with pytest.raises(ValueError, match=error):
-        CoprRepositoryProvider(raw_id, root_logger)
+        CoprRepositoryProvider(raw_id, repository_priority=50, logger=root_logger)
```
LecrisUT: Consistency issue. We do not care about the `invalid:` case because that would not be acessible

vaibhavdaren: Thought so as well. addressed in 091dc5ce4a5b0deba740e86e29cd935d53fcdda5.

tmt/schemas/prepare/artifact.yaml
```diff
@@ -18,5 +18,8 @@ properties:
   provide:
     $ref: "/schemas/common#/definitions/one_or_more_strings"
 
+  default_repository_priority:
```
happz: Specification keys are always using `-` to separate words.

```suggestion
  default-repository-priority:
```

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -29,6 +29,17 @@ class PrepareArtifactData(PrepareStepData):
...
     )
 
+    default_repository_priority: int = field(
+        default=50,
+        option='--default-repository-priority',
+        metavar='PRIORITY',
+        help="""
+            Default priority for created artifact repositories. Lower values mean
+            higher priority in package managers. Default: 50. Also uses environment
+            variable ``TMT_PLUGIN_PREPARE_ARTIFACT_DEFAULT_REPOSITORY_PRIORITY``.
```
happz: Please, don't mention the default value, it will be displayed by tmt and rendered into HTML docs automatically. Also, I advise against mentioning the environment variable, because pretty much all CLI options can be specified via environment variables, and this one is not special enough to deserve its own to be mentioned explicitly. See https://tmt--4505.org.readthedocs.build/en/4505/plugins/prepare.html#id1, it's rendered into HTML, and once I'm done with https://github.com/teemtee/tmt/pull/4472, we can enable display of envvars in CLI help as well.

happz:
PR name would deserve an update, to better reflect the change to something like "Accept default repository priority for repositories created by prepare/artifact"

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Yes, help text should be simpler. Addressed in ec2fbbb77e8d9ff02139ac54bc4ffd402f1fef26. Thanks for pointing this out.

PR #4509, opened 2026-01-19
Document that there is no context auto-detection

### LINKED ISSUE #4444: Add big note/warning block into the "Context" documentation that the context is *not* autodiscovered by tmt
We see users from time to time asking why their tests with `distro == fedora` do not run on their machine that's clearly a Fedora, as it was spawned from a Fedora container/VM image. The "Context" page does not make it clear that the context is not discovered and populated by tmt, and that it's fully owned by the user. A couple of examples would help as well ("`tmt run provision --image fedora:43` will run all enabled tests, not just those for Fedora, because **you** did not specify `-c distro=fedora-43`").

psss:
Explicitly document that there is no support for detecting context from the provisioned environment or other command line options. Provide a couple more examples to demonstrate the usage and make it completely obvious that both provision image and distro context dimension need to be provided by the user.

Fix #4444.

Pull Request Checklist

* [x] write the documentation
* [x] include a release note

### DIFF
--- a/spec/context/main.fmf
+++ b/spec/context/main.fmf
@@ -27,12 +27,14 @@ description: |
     .. _fmf context: https://fmf.readthedocs.io/en/stable/context.html
 
     .. note::
-
-        There is no support for detecting context from other
-        command line options or from the provisioned environment.
         Always make sure the needed context dimensions are
         explicitly provided either on the command line or in the
         plan.
+
+        There is no support for detecting context from other
+        command line options or from the provisioned environment.
+        Similarly, provision options, such as distro image, are
+        not detected from the context values.
 
         Because of the chicken-and-egg problem (plan definitions,
         including the ``provision`` step, can be modified using
@@ -64,6 +66,12 @@ example:
     tmt --context @ci.yaml run
 
   - |
+    # In most use cases you want to define both distro image for
+    # the provision step and distro context dimension because
+    # context is not detected from image and vice versa.
+    tmt --context distro=fedora-44 run --all provision --how virtual --image fedora-44
+
+  - |
     # When context is not defined, adjust rules checking for the
     # specific distro version will not be applied
     tmt run --all provision --how virtual --image fedora-44
@@ -72,11 +80,6 @@ example:
     # Distro images are not detected from the context value,
     # default image will be used when not explicitly specified
     tmt --context distro=fedora-44 run --all provision --how container
-
-  - |
-    # Define both provision image and the distro context dimension
-    # if you need test or plan adjust rules to be applied
-    tmt --context distro=fedora-44 run --all provision --how virtual --image fedora-44
 
 link:
   - implemented-by: /tmt/base.py


spec/context/main.fmf
```diff
@@ -11,26 +11,67 @@ description: |
...
-    The context definition provided directly on the command line
-    overrides defined dimensions. All context dimension values are
-    handled in case-insensitive way.
+    All context dimensions are optional and all context dimension
+    values are handled in case-insensitive way. It is possible to
+    define your own dimensions to describe the context, just make
+    sure the :ref:`/spec/context/dimension` name does not conflict
+    with reserved names.
 
     __ https://fmf.readthedocs.io/en/latest/context.html
```
LecrisUT: To avoid future surprises, can we convert this link to a named ref link? Technically could also use `intersphinx` to reference it.

spec/context/main.fmf
```diff
@@ -11,26 +11,67 @@ description: |
...
 
-example: |
+    .. note::
```
LecrisUT, line 29: How about an additional note mentioning that we also do not use the contexts, and that all usages must be defined by the user in the `fmf` file, at least for now.

spec/context/main.fmf
psss: Sure: 4a181654

spec/context/main.fmf
psss: I'm not completely sure what do you mean by `we also do not use the contexts`.

LecrisUT: E.g. for `ProvisionPodman`, we do not take the `--image` from the context `distro`, instead it is a defaulted from hard-coded `fedora`.

happz: The lines below made this clear to me:

> There is no support for detecting context from other command line options or from the provisioned environment. Always make sure the needed context dimensions are explicitly provided either on the command line or in the plan.

LecrisUT: > The lines below made this clear to me:
> 
> > There is no support for detecting context from other command line options or from the provisioned environment. Always make sure the needed context dimensions are explicitly provided either on the command line or in the plan.

To me this covers the detection, not the utilization. I.e. sure `distro` is not populated from the environment provisioned, but also (what is missing) defining `distro` alone does not mean that it will be used to create the appropiate image in `provision` when you `tmt -c distro=centos run -a provision --how=container`

psss: I see, so it's about the other way round, about detecting images from context. What about including an example like this 28e05ce3 to make this clear?

LecrisUT: > What about including an example like this [28e05ce](https://github.com/teemtee/tmt/commit/28e05ce3fe11db92b2abe1e1f96ee4b29933121d) to make this clear?

I think that may have the opposite effect because we are still documenting the usage of a `tmt -c ... run` without the context of the fmf file. Maybe the context could be a `distro=rhel-10` and the comments could be more explicit about "the following _does not_ guarantee a rhel container is provisioned".

But also, this seems a bit hidden in the example, would prefer at least a sentence or a note in the main text to accompany it.

psss: > Maybe the context could be a `distro=rhel-10` and the comments could be more explicit about "the following _does not_ guarantee a rhel container is provisioned".

I'm lost here. What do you mean by `does not guarantee`? That sounds super confusing to me. Context does not affect distro image choice at all so how it could even suggest to have some influence on it?

Examples are all demonstrating usage on the command line, I don't think it would be good to combine them here. 

> But also, this seems a bit hidden in the example, would prefer at least a sentence or a note in the main text to accompany it.

Sure, can add it to the description as well. Once we've clarified the above.

LecrisUT: > > Maybe the context could be a `distro=rhel-10` and the comments could be more explicit about "the following _does not_ guarantee a rhel container is provisioned".
> 
> I'm lost here. What do you mean by `does not guarantee`? That sounds super confusing to me. Context does not affect distro image choice at all so how it could even suggest to have some influence on it?

*Words are hard* :grin:. I mean it as something like "you need to make sure you provide an equivalent `--image` flag, or have an `adjust`". Any wording to indicate that the context alone has no special effect (probably easier to formulate in the note)

psss: Ok, one more attempt in 649e9f41. Added an explicit mention to the note and started the combinations examples with:

    # In most use cases you want to define both distro image for
    # the provision step and distro context dimension because
    # context is not detected from image and vice versa.

LecrisUT: Looks good :)

psss:

PR #4517, opened 2026-01-22
Remove outdated `rhel-8` workarounds and comments

### LINKED ISSUE #3705: Remove unnecessary "RHEL-8 workarounds"
For example "Because of old jsonschema package on RHEL-8, we cannot use foo" instances.
Searching for "rhel-8", "epel-8", "el8" will probably uncover more items that are no longer necessary, or can be improved/simplified with not having to support el8.  

Note that el8 guests are supported.

mcasquer:
As tmt running in RHEL 8 is no longer supported, remove the schema workarounds referencing RHEL-8

Fixes #3705 

* [x] modify the json schema

### DIFF
--- a/tmt/schemas/common.yaml
+++ b/tmt/schemas/common.yaml
@@ -121,12 +121,24 @@ definitions:
     additionalProperties: false
 
     properties:
-      url: true
-      ref: true
-      path: true
-      name: true
-      nick: true
-      type: true
+      url:
+        type: string
+
+      ref:
+        type: string
+
+      path:
+        type: string
+
+      name:
+        type: string
+        pattern: "^/"
+
+      nick:
+        type: string
+
+      type:
+        $ref: "/schemas/common#/definitions/require_type"
 
     required:
       - type
@@ -243,20 +255,36 @@ definitions:
     additionalProperties: false
 
     properties:
-      url: true
-      ref: true
-      path: true
-      name: true
+      url:
+        type: string
+
+      ref:
+        type: string
+
+      path:
+        type: string
+
+      name:
+        type: string
+        pattern: "^/"
 
   # helper used by beakerlib_library and fmf_id
   fmf_id_base:
     type: object
     minProperties: 1
     properties:
-      url: true
-      ref: true
-      path: true
-      name: true
+      url:
+        type: string
+
+      ref:
+        type: string
+
+      path:
+        type: string
+
+      name:
+        type: string
+        pattern: "^/"
 
   require_file:
     type: object
--- a/tmt/schemas/plan.yaml
+++ b/tmt/schemas/plan.yaml
@@ -136,11 +136,19 @@ properties:
         additionalProperties: false
 
         properties:
-          # https://github.com/teemtee/tmt/issues/1258
-          url: true
-          ref: true
-          path: true
-          name: true
+          url:
+            # https://github.com/teemtee/tmt/issues/1258
+            type: string
+
+          ref:
+            type: string
+
+          path:
+            type: string
+
+          name:
+            type: string
+            format: regex
 
           importing:
             type: string
--- a/tests/plan/export/test.sh
+++ b/tests/plan/export/test.sh
@@ -91,14 +91,7 @@ rlJournalStart
 
     rlPhaseStartTest "Invalid format"
         rlRun -s "tmt plan export --how weird" 2
-
-        if rlIsRHELLike "=8"; then
-            # RHEL-8 and Centos stream 8 usually offer an older Click package that has slightly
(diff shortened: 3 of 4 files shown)


tmt/schemas/common.yaml
```diff
@@ -120,28 +120,13 @@ definitions:
 
     additionalProperties: false
 
-    # NOTE: Because of old jsonschema package on RHEL-8, we cannot use
-    # `url: true` and inherit `url` attributes from parent schema. It
-    # seems like we have to repeat properties schemas here as well.
```
LecrisUT, line 125: I had my suspicion about this comment, and I tried it with the following example:
```yaml
/tests:
  /valid:
    test: something
    require:
      - url: https://github.com/beakerlib/beakerlib
        type: library
        name: /example
  /invalid:
    test: something
    require:
      - url: 12345
        type: library
        name: { }
```
With the changes here, the invalid one always passes the schema, while previously it failed correctly. Best option here is to just delete these comments and revert the rest.

tmt/schemas/plan.yaml
```diff
@@ -135,23 +135,11 @@ properties:
 
         additionalProperties: false
 
-        # NOTE: Because of old jsonschema package on RHEL-8, we cannot use
-        # `url: true` and inherit `url` attributes from parent schema. It
-        # seems like we have to repeat properties schemas here as well.
         properties:
-          url:
-            # https://github.com/teemtee/tmt/issues/1258
```
LecrisUT, line 143: Make sure this comment is preserved, it is still very much relevant

LecrisUT (changes requested):

tmt/schemas/common.yaml
mcasquer: So the options such as `url: true` are not valid, we still need to redefine them, right?

tmt/schemas/plan.yaml
mcasquer: Done :)

tmt/schemas/common.yaml
LecrisUT: Yes, the previous way of defining them was correct. Maybe just needing more `$ref` indirections.

mcasquer:
What about the tmp_path* fixtures that are not available at RHEL-8, I see we have a workaround at https://github.com/teemtee/tmt/blob/746105f16ad0d015dd71de552b9ac1431658bb07/tests/unit/conftest.py#L45

I'm not sure of the magnitude of what this implies, or if it's good as it is now :)
cc @LecrisUT @happz

LecrisUT:
> What about the tmp_path* fixtures that are not available at RHEL-8, I see we have a workaround at
> 
> https://github.com/teemtee/tmt/blob/746105f16ad0d015dd71de552b9ac1431658bb07/tests/unit/conftest.py#L45
> 
> I'm not sure of the magnitude of what this implies, or if it's good as it is now :) cc @LecrisUT @happz

Based on
https://github.com/teemtee/tmt/blob/746105f16ad0d015dd71de552b9ac1431658bb07/tests/unit/conftest.py#L39-L44
We still need to keep it, but the `try ... except` in there can be simplified. What I don't know if it was invesitgated, what happens if we also provide a `pytest.fixture` with the same name `tmp_path` or have any way to keep the tmp_path name but edit the output type.

But the minimum blast radius would be to just simplify the `try ... except` and that should not have any side-effects, so start from there.

happz:
> What about the tmp_path* fixtures that are not available at RHEL-8, I see we have a workaround at
> 
> https://github.com/teemtee/tmt/blob/746105f16ad0d015dd71de552b9ac1431658bb07/tests/unit/conftest.py#L45
> 
> I'm not sure of the magnitude of what this implies, or if it's good as it is now :) cc @LecrisUT @happz

If you keep `tmppath` and `tmppath_factory` working as they are today, and drop the backward compatibility, the it should be harmful and remove one old piece of code.

tmt/schemas/common.yaml
mcasquer: Kept the original logic, comments have been removed

mcasquer:
References to tmt on RHEL 8 have been removed, as discussed internally this topic opened the door to look into the dependencies too, e.g. the following pre-commit entry must be updated/removed:
https://github.com/teemtee/tmt/blob/0e222cad206b1853dfa3b4b3fd2d5a385acef91c/.pre-commit-config.yaml#L182
This will covered in a different issue + PR

LecrisUT (approved):

psss:
Failures are irrelevant

* missing `fedora-44` compose
* Cannot download repodata/repomd.xml: All mirrors were tried

Merging.

PR #4521, opened 2026-01-23
Add `pending` and `skip` state to the `junit` report

### LINKED ISSUE #4533: Pending tests are reported as passed in Polarion report
Any test ending in pending state (not executed) is marked as passed in Polarion.
This is caused by missing skipped tag in the report for junit on which Polarion report is based on.

KwisatzHaderach:
The template and `junit` report plugin was missing a new option for result state `pending`. Adding that to the plugin and template along with `skipped` status so the Polarion report no longer marks non-executed cases as passed.

Fixes #4533 

Pull Request Checklist

* [x] implement the feature
* [x] include a release note

### DIFF
--- a/docs/releases/pending/4521.fmf
+++ b/docs/releases/pending/4521.fmf
@@ -1,3 +1,4 @@
 description: |
-    Adding missing skipped/error tags to :ref:`/plugins/report/junit`
-    report template and include pending state in these reports as error.
+    The :ref:`/plugins/report/junit` report template has been updated
+    to correctly map test result ``skip`` to the ``skipped`` tag and
+    ``pending`` result to the ``error`` tag.


psss:
Thanks much for fixing this! Seems like an important blocker, adding as a `must` have into the sprint.

tmt/steps/report/junit/templates/_base.xml.j2
```diff
@@ -12,7 +12,7 @@
                         <error type="error" message="{{ result.result.value | e }}">{{ result.failure_logs | failures | join("\n") | e }}</error>
                     {% elif result.result.value == 'fail' %}
                         <failure type="failure" message="{{ result.result.value | e }}">{{ result.failure_logs | failures | join("\n") | e }}</failure>
-                    {% elif result.result.value == 'info' %}
+                    {% elif result.result.value == 'skip' or result.result.value == 'info' or result.result.value == 'pending' %}
```
happz: Is "skipped" the right junit outcome for "pending" result? For "skip", sure, but "pending" means it should have run, it would have, but something happened and the test did not run at all. It's not a happy outcome, would "skipped" seen as similarly unhappy outcome in junit and/or polarion?

Maybe we could modify just the `polarion.xml.j2` outcome, since IIUIC this is trying to tailor the output to achieve something in polarion reporting.

KwisatzHaderach: there are 3 options: failure, error and skipped
from these I think skipped is the best one, in Polarion this ends up as non-executed while error ends up as blocked, which is also somewhat reasonable, but I don't want put it on the same level as errored out tests (tests that very likely broke the tested machine)

psss: Hm, this is actually a very good point. Let's imagine the following scenario:

* tmt plan discovers 100 tests
* executes the first 10 which happily pass
* then something weird happens and test execution is interrupted
* remaining 90 tests are marked as `pending`
* will appear as `skipped` in the `polarion` report

So... this, all in all, would appear in Polarion as "everything went well"? Couldn't something like this hide an infrastructure errors?

It's true that `tmt` would in such case return non-zero exit code, but not sure how that is exactly handled/visible in Polarion. Would users be able to clearly see that there was an error?

KwisatzHaderach: technically yes, in the current (with this MR) form it would have 1 error and 89 pending/skipped tests, I don't think there is a way for the machine to break outside of any test execution during the execute step, or am I wrong here? can we end up with 10 pass and 90 pending?

again, I don't have a problem to put pending on the same level as error, it will just look weird and that is the only other option, no other states are recognised by Polarion (unless you want fail for these)

psss: > I don't think there is a way for the machine to break outside of any test execution during the execute step, or am I wrong here?

Yes, I believe that can happen: Imagine the guest becomes unresponsive/unreachable after executing the first set of passing tests. Then the rest should be marked as "something went wrong, something unexpected and user should investigate".

> again, I don't have a problem to put pending on the same level as error, it will just look weird and that is the only other option, no other states are recognised by Polarion (unless you want fail for these)

Then, I guess, yeah, `error` sounds to be the closest mapping to our `pending` status. Added to tomorrow's hacking session agenda to confirm the approach.

psss: In `junit` the `skip` status is a "happy" status, representing something that should not run (e.g. test is not relevant for this environment, we do not need to run the test). Our `pending` means that a specific test was discovered, scheduled and should have been executed, it is missing. It's a terrible outcome, should not go unnoticed.

Decision from the hacking session:

* the `pending` test result should be mapped to `error`
* add a note explaining what happened to give user an idea about what went wrong

docs/releases/pending/4521.fmf
```diff
@@ -0,0 +1,3 @@
+description: |
+    Adding missing skipped tags to junit report template
+    and include pending state in these reports as skipped.
```
psss: Release note needs update as well.

tests/report/junit-pending/data/.fmf/version
```diff
@@ -0,0 +1 @@
+1
```
psss, line 1: Let's rather keep the tree structure and store all `junit` tests under a single directory to have `/tests/report/junit/basic` and `/tests/report/junit/pending`.

tests/report/junit-pending/main.fmf
```diff
@@ -0,0 +1,5 @@
+summary: Verify that pending test is marked properly in junit report
+duration: 5m
```
psss: Sometimes the `virtual` provision can be slower, let's raise the limit a bit?

tmt/steps/report/junit/templates/_base.xml.j2
```diff
@@ -8,11 +8,11 @@
                 {% set test_duration = result.duration | duration_to_seconds | float %}
 
                 <testcase name="{{ result.name | e }}" {% if test_duration %}time="{{ test_duration }}"{% endif %}>
-                    {% if result.result.value == 'error' or result.result.value == 'warn' %}
+                    {% if result.result.value == 'error' or result.result.value == 'warn' or result.result.value == 'pending' %}
                         <error type="error" message="{{ result.result.value | e }}">{{ result.failure_logs | failures | join("\n") | e }}</error>
                     {% elif result.result.value == 'fail' %}
                         <failure type="failure" message="{{ result.result.value | e }}">{{ result.failure_logs | failures | join("\n") | e }}</failure>
-                    {% elif result.result.value == 'info' %}
+                    {% elif result.result.value == 'skip' or result.result.value == 'info' %}
```
psss, line 15: Might be useful to have these mappings explicitly documented. What about adding just a short section to the [junit](https://tmt.readthedocs.io/en/stable/plugins/report.html#junit) plugin docs which would list all result statuses supported by `tmt` and their corresponding `junit` value?

docs/releases/pending/4521.fmf
```diff
@@ -0,0 +1,3 @@
+description: |
+    Adding missing skipped/error tags to junit report template
```
psss:
```suggestion
    Adding missing skipped/error tags to :ref:`/plugins/report/junit` report template
```

And the new docs could be linked from the release note.

tmt/steps/report/junit.py
```diff
@@ -523,6 +527,13 @@ class ReportJUnit(tmt.steps.report.ReportPlugin[ReportJUnitData]):
         report:
             how: junit
             file: test.xml
+
+    Possible test results and their mapping to junit tags:
+
+    * Passed - no specific tag, test is passed
```
happz: I would recommend sticking to outcome labels used by specification, https://tmt.readthedocs.io/en/stable/spec/results.html, plus the help can be nicer to read with the help of some ReST magic. How about the following?

```ReST
Possible test results and their mapping to junit tags:

* ``pass`` - no specific tag, test has succeeded
* ``fail`` - ``<failure />`` tag
* ``skip`` or ``info`` - ``<skipped />`` tag
* ``error``, ``warn``, or ``pending`` - ``<error />`` tag
```

happz:
@KwisatzHaderach isn't the patch also adding `skip`? If so, the title and description would deserve an update.

psss (approved):
Except for status names formatting and mentioning the `skip` status as well looks good. Thanks!

docs/releases/pending/4521.fmf
```diff
@@ -0,0 +1,3 @@
+description: |
+    Adding missing skipped/error tags to :ref:`/plugins/report/junit`
```
psss: This should mention adding `pending` and `skip` as `error` was not missing, right?

KwisatzHaderach: since this talks about tags, then I think it's actually correct, pending is now adding error tags, so yeah, error tag was missing too

docs/releases/pending/4521.fmf
```diff
@@ -0,0 +1,3 @@
...
+    report template and include pending state in these reports as error.
```
psss:
```suggestion
    The :ref:`/plugins/report/junit` report template has been updated
    to correctly map test result ``skip`` to the ``skipped`` tag and
    ``pending`` result to the ``error`` tag.
```

I see, understood. What about something like this to make both mapping changes clear?

psss:
Failing job is an irrelevant repo failure:

```
content: Error: Failed to download metadata for repo
'epel-debuginfo': Yum repo downloading error: Downloading
error(s):
repodata/2afd50c1af19fd84b87020f32df13d038c093ee1666801f7dbf7b9a2347a3ac9-primary.xml.xz
- Cannot download, all mirrors were already tried without success;
repodata/fc38e6dbf9ce4a3fe82b1ef9fe388a0ff20ae242a5a402eeef9c00e6b651ea3f-filelists.xml.xz
- Cannot download, all mirrors were already tried without success
```

Merging.

PR #4523, opened 2026-01-27
Step `__init__` shall accept list of raw step data only, no other types are allowed

happz:
This particular normalization happened inside the method, which opened its `_raw_data` and `data` operations to way more types than necessary. `Step.__init__` now accepts `list[_RawStepData]` and nothing else. Plus, the parameter name is now `raw_data` rather than `data` as `data` labels data classes and their instances rather than raw, unnormalized input.

Pull Request Checklist

* [x] implement the feature


therazix (approved):
LGTM

LecrisUT (approved):
Title is not quite correct (it used to accept before, but now it *must* be a list)

PR #4525, opened 2026-01-27
Support prepare shell scripts in image mode

### LINKED ISSUE #4495: Add bootc support for prepare shell step
## Summary

Add support for bootc guests in the `prepare/shell` step, allowing shell scripts to be executed as part of the bootc image build process rather than on the live guest.

## Background

When using bootc guests, most preparation commands should be collected and executed during the container image build (via Containerfile), followed by a reboot. Only test commands should run on the live guest.

Related draft PR: #3947

## Proposed Implementation

Based on discussions in #3947, the implementation should:

1. **Hide bootc behavior behind the Guest class API** - Plugins should not need to be aware they are talking to a bootc guest. The different behavior should be transparent to plugin users.

2. **Add `make_changes` parameter to `Guest.execute()`** - Commands that modify the system (default `make_changes=True`) would be collected for the Containerfile, while probe/check commands (`make_changes=False`) would execute immediately on the live guest.

3. **Determine when to build the collected Containerfile** - Options include:
   - Lazy approach: Build when `test_session=True` is seen (test is about to run)
   - Add explicit `on_start`/`on_finish` events to phase/step workflow

4. **Support `Guest.push()` for Containerfile COPY operations** - Needed for cases like `prepare/shell` with URL + script, where content needs to be available during the build:
   ```yaml
   prepare:
       how: shell
       url: https://github.com/teemtee/tmt.git
       ref: main
       script: cd $TMT_PREPARE_SHELL_URL_REPOSITORY && make docs
   ```

## Open Questions

- How to handle the `finish` step (no test after it, so collected commands would remain in the buffer)?
- Which specific commands need to run live vs. be collected (probe commands, package presence checks, bootc status commands)?

## References

- Draft PR: #3947
- Related to bootc package manager implementation pattern (separating command construction and execution)

thrix:
## Summary

This PR enables `prepare/shell` scripts to work with guests in image mode by collecting commands and applying them via container image rebuild.

Closes #4495

### Image Mode Support for Prepare/Shell

- **Add `is_image_mode` guest fact**: Detects if the guest was booted from a bootc image by checking `bootc status` output
- **Add `CommandCollector` interface**: Abstract base class for guests that support deferred command execution
- **Add `immediately` parameter to `execute()`**: Controls whether commands run immediately or are collected for batch execution:
  - `immediately=True` (default): Execute command right away
  - `immediately=False`: Command may be collected for later batch execution on image mode guests
- **Add `on_step_complete()` hook**: Notifies guests when a step completes, triggering the container build and reboot for collected commands
- **Add bind mount for tmt workdir**: Mounts the run workdir (`-v` flag with `:Z` SELinux label) during `podman build` so prepare scripts can access plan data
- **Add `_save_deferred_run_outcome()` helper**: Records INFO results for commands collected for deferred execution

### Bootc Package Manager Improvements

- **Add dry-run guards**: Skip container build operations in dry-run mode

### How It Works

1. When `facts.is_image_mode` is `True`, commands with `immediately=False` are collected into Containerfile RUN directives
2. At step completion, `on_step_complete()` triggers the container build via `flush_collected()`
3. The new image is built with `podman build`, applied via `bootc switch`, and the guest is rebooted
4. Commands with `immediately=True` (default) are always executed immediately

## Checklist

* [x] implement the feature
* [x] include a release note

Assisted-by: Claude Code

### DIFF
--- a/tmt/steps/provision/__init__.py
+++ b/tmt/steps/provision/__init__.py
@@ -1584,18 +1584,6 @@ class CommandCollector(abc.ABC):
 
         raise NotImplementedError
 
-    def on_step_complete(self, step: 'tmt.steps.Step') -> None:
-        """
-        Called when a step completes execution.
-
-        Flush collected commands if there are some.
-
-        :param step: the step that has completed.
-        """
-        if self.has_collected_commands:
-            self.flush_collected()
-            return
-
 
 class Guest(
     tmt.utils.HasRunWorkdir,
@@ -2870,6 +2858,17 @@ class GuestSsh(Guest, CommandCollector):
 
         self.info("building container image from collected commands", "green")
         self.package_manager.build_container()
+
+    def on_step_complete(self, step: 'tmt.steps.Step') -> None:
+        """
+        Called when a step completes execution.
+
+        Flush collected commands if there are some.
+
+        :param step: the step that has completed.
+        """
+        if self.has_collected_commands:
+            self.flush_collected()
 
     @functools.cached_property
     def _ssh_guest(self) -> str:


tmt/package_managers/bootc.py
```diff
@@ -228,10 +228,10 @@ def build_container(self) -> None:
                         f'  ( podman pull {base_image} || podman pull containers-storage:{base_image} )'  # noqa: E501
                         f'  || bootc image copy-to-storage --target {base_image}'
                         ')"'
-                    )
+                    ),
```
thrix:
```suggestion
                    )
```

tmt/package_managers/bootc.py
```diff
@@ -228,10 +228,10 @@ def build_container(self) -> None:
...
                 )
                 self.guest.execute(
-                    ShellScript(f'cat <<EOF > {containerfile_path!s} \n{containerfile} \nEOF')
+                    ShellScript(f'cat <<EOF > {containerfile_path!s} \n{containerfile} \nEOF'),
```
thrix:
```suggestion
                    ShellScript(f'cat <<EOF > {containerfile_path!s} \n{containerfile} \nEOF')
```

tmt/package_managers/bootc.py
```diff
@@ -243,7 +243,7 @@ def build_container(self) -> None:
                 self.guest.execute(
                     ShellScript(
                         f'{self.guest.facts.sudo_prefix} podman build -t {image_tag} -f {containerfile_path} {self.guest.step_workdir}'  # noqa: E501
-                    )
+                    ),
```
thrix:
```suggestion
                    )
```

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2825,134 @@ def __init__(
...
+
+    def collect_command(
+        self,
+        command: Union[tmt.utils.Command, tmt.utils.ShellScript],
+        *,
+        cwd: Optional[Path] = None,
+        env: Optional[tmt.utils.Environment] = None,
+    ) -> None:
+        """
+        Collect a command for later Containerfile execution.
```
thrix:
```suggestion
        Collect a command for later container image building.
```

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2825,134 @@ def __init__(
...
+            script_parts.append(command.to_script())
+        else:
+            script_parts.append(command)
+
+        # Combine all parts into a single script
+        full_script = ShellScript.from_scripts(script_parts)
+
+        # Add to the package manager's engine
+        bootc_pm.engine.open_containerfile_directives()
+        bootc_pm.engine.containerfile_directives.append(f"RUN {full_script}")
```
thrix: We should bind mount the tmt workdir here, so the building has access to those files

E.g.

```
RUN --mount=type=bind,source=.,target=/src,options=z 
```

thrix: This issue can be seen in the tests:

https://artifacts.osci.redhat.com/testing-farm/739bc9d1-1375-4fe9-9fa5-5b0900748729/

```
            out: STEP 1/2: FROM quay.io/testing-farm/centos-bootc:stream10
            out: STEP 2/2: RUN export TMT_TOPOLOGY_YAML=/var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/tree/tmt-test-topology-default-0-default-0.yaml; export TMT_TOPOLOGY_BASH=/var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/tree/tmt-test-topology-default-0-default-0.sh; export TMT_VERSION=1.66.0.dev23+gf6d0e6b62; export TMT_TREE=/var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/tree; export TMT_PLAN_DATA=/var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/data; export TMT_PLAN_ENVIRONMENT_FILE=/var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/data/variables.env; export TMT_PLAN_SOURCE_SCRIPT=/var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/data/plan-source-script.sh; cd /var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/tree; set -eo pipefail; dnf -y install tree 
            err: /bin/sh: line 1: cd: /var/tmp/tmt/tmp.wG0Afdo6E4/plans/centos-stream-10/prepare-shell/tree: No such file or directory
```

tmt/steps/provision/__init__.py
```diff
@@ -1699,6 +1774,22 @@ def bootc_builder(
             guest=self, logger=self._logger
         )
 
+    @property
+    def is_bootc_guest(self) -> bool:
```
thrix: Should this not go to facts?

tmt/steps/provision/__init__.py
```diff
@@ -2214,6 +2311,12 @@ def execute(
         :param cwd: if set, execute command in this directory on the guest.
         :param env: if set, set these environment variables before running the command.
         :param friendly_command: nice, human-friendly representation of the command.
+        :param test_session: if True, this is the actual test being run.
```
thrix: this looks like an unrelated change, maybe open another MR documenting other missing params, seems there are more

tmt/steps/provision/__init__.py
thrix: Resolved. `podman build` support volume mounts \o/. Mounting inside the whole workdir.

tmt/steps/provision/__init__.py
thrix: could be moved, but would not help much.

tmt/steps/provision/__init__.py
```diff
@@ -3066,6 +3328,7 @@ def execute(
         env: Optional[tmt.utils.Environment] = None,
         friendly_command: Optional[str] = None,
         test_session: bool = False,
+        make_changes: bool = False,
```
LecrisUT: I am not convinced by the need of `make_changes` currently, only the `provision` part seems to not make changes since those only do probes. The need to make changes or not could instead be controlled as a guest property instead based on which step it is in (only make changes in prepare/finish?).

It may be that during provision you want to make changes, but in that case you already know what provisioned guest it is and if needs a special handling for bootc or if it is done live.

The default is also unclear to me why it should be one way or the other. It would be better to make it explicitly required parameter if such a design is used.

tmt/steps/provision/__init__.py
```diff
@@ -3249,7 +3532,11 @@ def push(
 
         try:
             if options.create_destination:
-                self.execute(Command("mkdir", "-p", destination.parent), silent=True)
+                # Use make_changes=False because this must execute immediately,
+                # not be collected for bootc guests (rsync that follows needs the directory)
+                self.execute(
+                    Command("mkdir", "-p", destination.parent), silent=True, make_changes=False
```
LecrisUT: Here is an example where the interface makes it confusing. This command very much makes changes, but the interface makes it appear like it does not.

At the same time does this code path even get traversed in the bootc case? Isn't it overwritten?

tmt/steps/provision/__init__.py
```diff
@@ -1699,6 +1774,22 @@ def bootc_builder(
...
+    def is_bootc_guest(self) -> bool:
+        """
+        Check if this guest is running a bootc-based system.
+
+        Returns True when the guest's detected package manager is 'bootc',
+        indicating that commands should be collected into a Containerfile
+        rather than executed immediately.
+        """
+        # Access __dict__ directly to avoid triggering the facts property getter
+        # which would call sync() if facts aren't ready, causing infinite recursion
```
LecrisUT: Well don't make it a property then, just call `.facts` directly when needed.

tmt/steps/provision/__init__.py
```diff
@@ -2657,10 +2783,15 @@ class GuestSshData(GuestData):
     )
 
 
-class GuestSsh(Guest):
+class GuestSsh(Guest, GuestCommandCollector):
```
LecrisUT: Do we need to cram all of this into the generic `GuestSsh`. Bootc guest is very much a special guest type with nuanced implementations so let's make it its own class.

E.g. even if we are using `provision/connect` and we are ssh-ing to a guest that is bootc, we cannot recreate the image in there so it would conflict with the one that we create via podman.

tmt/steps/provision/__init__.py
thrix: > I am not convinced by the need of make_changes currently, only the provision part seems to not make changes since those only do probes. The need to make changes or not could instead be controlled as a guest property instead based on which step it is in (only make changes in prepare/finish?).

yeah, it is a first try that works, mostly based on the ideas in https://github.com/teemtee/tmt/pull/3947. Anyway, I am fine changing it the way you describe.

tmt/steps/provision/__init__.py
thrix: ack, will check.

tmt/steps/provision/__init__.py
happz: Sounds legit, I'd expect the mixin to be added into the bootc guest, as that's the one that needs to support collection of commands, while Artemis or beaker guests, both `GuestSsh` subclasses, do not.

tmt/steps/provision/__init__.py
happz: Indeed, I would turn this into a fact. We already have two facts, package manager and bootc builder, i.e. a package manager to use for building commands on behalf of bootc package manager. We can have a fact just like we have `is_container` or `is_ostree`.

tmt/package_managers/bootc.py
```diff
@@ -101,6 +101,10 @@ def _get_current_bootc_image(self) -> str:
             raise tmt.utils.PrepareError("Failed to extract bootc image info.") from error
 
     def _get_base_containerfile_directives(self) -> list[str]:
+        # In dry run mode, use a placeholder image
+        if self.guest.is_dry_run:
+            return ['FROM dry-run-bootc-image:latest']
```
happz: Why a dummy image?

tmt/steps/provision/__init__.py
```diff
@@ -2214,6 +2311,16 @@ def execute(
...
+        :param test_session: if True, this is the actual test being run.
+        :param make_changes: if True, the command modifies system state.
+            For bootc guests implementing :py:class:`GuestCommandCollector`,
+            commands with ``make_changes=True`` are collected into a Containerfile
+            rather than executed immediately. Commands with ``make_changes=False``
+            (default, probes/checks) are always executed immediately.
+            For bootc guests implementing :py:class:`GuestCommandCollector`,
+            commands with ``make_changes=True`` are collected into a Containerfile
+            rather than executed immediately. Commands with ``make_changes=False``
+            (probes, checks) are always executed immediately.
```
happz: * `make_changes` can be interpreted as an order, "make changes!", that's how I read it, which is puzzling, especially `make_changes=False`. It would be nice to find a better name. `makes_changes` wouldn't be an order anymore, but a statement about the command, for example. Let's see whether we can find a better label for this category of commands.
* bootc sneaking into so many comments in classes tha tshouldn't be aware of bootc at all. In `Guest` class, I shouldn't really care how `bootc` plugin handles `make_changes=True`. The important thing is, with `make_changes=True`, some plugins may chose to postpone these commands, and perform them later, e.g. in a batch, and that `make_changes=False` means the command is executed here and now. Mentions of `GuestCommandCollector` do not belong here.

tmt/steps/provision/__init__.py
happz: Exactly, `make_changes=False` is really puzzling.

tmt/steps/provision/__init__.py
```diff
@@ -3082,10 +3345,30 @@ def execute(
         :param cwd: execute command in this directory on the guest.
         :param env: if set, set these environment variables before running the command.
         :param friendly_command: nice, human-friendly representation of the command.
+        :param test_session: if True, this is the actual test being run.
+        :param make_changes: whether the command modifies system state.
+            This parameter is used by bootc guests to determine whether to
+            collect commands for the Containerfile. Use make_changes=True for
+            commands that modify the system (e.g., package installs, file writes),
+            and make_changes=False for probe/diagnostic commands (e.g., whoami,
+            cat /proc/stat). For regular guests, this parameter is ignored.
```
happz: Bootc sneaking into `GuestSsh`.

tmt/steps/provision/__init__.py
```diff
@@ -3082,10 +3345,30 @@ def execute(
...
         """
 
         sourced_files = sourced_files or []
 
+        # For bootc guests: flush any pending commands before running a test
+        if test_session and self._has_bootc_collected_commands():
+            self._flush_bootc_commands()
```
happz: Right, bootc guests - good, do your work, but stay in your class, please. Same below.

tmt/steps/provision/__init__.py
happz: > I am not convinced by the need of `make_changes` currently, only the `provision` part seems to not make changes since those only do probes. The need to make changes or not could instead be controlled as a guest property instead based on which step it is in (only make changes in prepare/finish?).

It also installs scripts, that's a change, may create directories, so I'm not that sure about exclusivity of not making changes in `provision`.

> It may be that during provision you want to make changes, but in that case you already know what provisioned guest it is and if needs a special handling for bootc or if it is done live.

But it all relates to commands, doesn't it? Let's say there's a guest property controlling whether commands are collected or not, but which commands? How would the generic machinery recognize that this command is supposed to be put on a pile with other commands?

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2829,134 @@ def __init__(
...
+        persists across reboots, so explicit COPY is not always needed.
+        """
+        # For now, files are pushed normally and persist in /var
+        # The bootc package manager handles this via push() calls
+        self.debug(f"File copy noted for bootc: {source} -> {destination}")
+
+    @property
+    def has_collected_commands(self) -> bool:
+        """Check if there are pending commands."""
+        return self._has_bootc_collected_commands()
```
happz: Yeah, this screams the implementation is in wrong place, `GuestSsh.has_collected_commands` calling `GuestSsh._has_bootc_collected_commands`.

happz:
Hmmm, or maybe I misunderstood the target audience. @thrix what is the use case, what plugins are the target group that need to support this? I'm starting to think I took "bootc" as "booc provision plugin", but that's not the case, is it?

tests/prepare/bootc/test.sh
```diff
@@ -0,0 +1,35 @@
...
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "pushd data"
+    rlPhaseEnd
+
+    # Test prepare/shell on bootc guest - should use Containerfile collection
+    rlPhaseStartTest "Prepare/shell on bootc guest - install tree package"
+        rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
```
psss: Should go to the `Setup` phase.

tests/prepare/bootc/test.sh
```diff
@@ -0,0 +1,35 @@
...
+        # Verify bootc switch was called
+        rlAssertGrep "switching to new image" $rlRun_LOG
+
+        # Verify reboot happened
+        rlAssertGrep "rebooting to apply new image" $rlRun_LOG
+
+        # Verify tree --version ran successfully in the test
+        rlAssertGrep "tree v" $rlRun_LOG
+
+        rlRun "rm -rf $run" 0 "Remove run directory"
```
psss: Should go to the `Cleanup` phase.

tmt/steps/provision/__init__.py
psss: Agreed, let's make the name more straightforward. Perhaps something like:

    guest.execute(..., enqueue=True)
    guest.execute(..., can_be_postponed=True)

or

    guest.execute(..., immediately=False)

for the `bootc` use case when commands should not be executed right away.

tmt/steps/provision/__init__.py
psss: Yeah, I'd also say the granularity will have to be per-command to allow some immediate actions as well.

tmt/steps/provision/__init__.py
psss: Yeah, makes sense.

psss (commented):
Overall the approach sounds reasonable, I agree that the parameter name and value is quite confusing and that the `bootc` stuff should not go directly into the `GuestSsh` class but should be handled separately.

happz:
> Overall the approach sounds reasonable, I agree that the parameter name and value is quite confusing and that the `bootc` stuff should not go directly into the `GuestSsh` class but should be handled separately.

Actually, we discussed this today with @thrix , and it turns out `GuestSsh` is the right place after all. The target here is not the "`bootc` provision plugin", but "a guest powered by bootc" - e.g. a Beaker or 1minutetip machine, governed and managed by bootc & co. That why there is the "is bootc" flag/fact; if true, on the remote end, behind the SSH connection, is a guest spawned from an image, and requires bootc workflow to install packages, run playbooks and so on. So it's not bootc leaking into `GuestSsh` as I originally thought, but new mode of guest behavior.

LecrisUT:
> Actually, we discussed this today with @thrix , and it turns out `GuestSsh` is the right place after all. The target here is not the "`bootc` provision plugin", but "a guest powered by bootc" - e.g. a Beaker or 1minutetip machine, governed and managed by bootc & co. That why there is the "is bootc" flag/fact; if true, on the remote end, behind the SSH connection, is a guest spawned from an image, and requires bootc workflow to install packages, run playbooks and so on. So it's not bootc leaking into `GuestSsh` as I originally thought, but new mode of guest behavior.

To clarify, this is about the flags and facts, not about the logic, right? If so, agreed. To me the handling logic of the `make_change` and such should be in a dedicated guest type.

happz:
> > Actually, we discussed this today with @thrix , and it turns out `GuestSsh` is the right place after all. The target here is not the "`bootc` provision plugin", but "a guest powered by bootc" - e.g. a Beaker or 1minutetip machine, governed and managed by bootc & co. That why there is the "is bootc" flag/fact; if true, on the remote end, behind the SSH connection, is a guest spawned from an image, and requires bootc workflow to install packages, run playbooks and so on. So it's not bootc leaking into `GuestSsh` as I originally thought, but new mode of guest behavior.
> 
> To clarify, this is about the flags and facts, not about the logic, right? If so, agreed.

We discussed it in the context of both flags and facts and the functionality, since the functionality, as mentioned above, is supposed to work across all SSH-backed guests.

> To me the handling logic of the `make_change` and such should be in a dedicated guest type.

It sounds like a trait to me, or a mixin. Being reachable over SSH seems as another one to me. Unfortunately, eventually these mixins would all be used together, as `artemis`, `beaker`, `connect`, `minute`, and `virtual` guests are the only SSH-backed guests, and they would include the SSH mixin and the bootc mixin since they must support both. And `container` and `mock` would include none of these two. So we'd still have two groups, one with both SSH and bootc support, and one group with neither... I guess it's dragged by the history and lack of time.

LecrisUT:
> Unfortunately, eventually these mixins would all be used together, as `artemis`, `beaker`, `connect`, `minute`, and `virtual` guests are the only SSH-backed guests, and they would include the SSH mixin and the bootc mixin since they must support both.

But they would still have distinct implementation for each one. Am considering something like `executor`/`execution_driver` etc. that abstracts the usage of the `guest.execute` more explicitly. This can then be specialized granularly for each ssh execution types to support bootc or not, and inside the bootc a similar split for how to create the bootc image and how to provide it to each `virtual`, `beaker`, etc. provision types.

tmt/steps/provision/__init__.py
LecrisUT: Would be fine with the granularity if the naming is consistent, and we agree on what the default should be (do changes immediately or not?) or always require passing the intent (breaking whatever plugin did not adjust). To me the default seems to depend on which step we are in, but also fine if we make the default consistent across the steps for simplicity.

tmt/steps/provision/__init__.py
LecrisUT: :+1: for something along the lines of `immediately`, `run_now`, etc.

thrix:
> > > Actually, we discussed this today with @thrix , and it turns out `GuestSsh` is the right place after all. The target here is not the "`bootc` provision plugin", but "a guest powered by bootc" - e.g. a Beaker or 1minutetip machine, governed and managed by bootc & co. That why there is the "is bootc" flag/fact; if true, on the remote end, behind the SSH connection, is a guest spawned from an image, and requires bootc workflow to install packages, run playbooks and so on. So it's not bootc leaking into `GuestSsh` as I originally thought, but new mode of guest behavior.
> > 
> > 
> > To clarify, this is about the flags and facts, not about the logic, right? If so, agreed.
> 
> We discussed it in the context of both flags and facts and the functionality, since the functionality, as mentioned above, is supposed to work across all SSH-backed guests.
> 
> > To me the handling logic of the `make_change` and such should be in a dedicated guest type.
> 
> It sounds like a trait to me, or a mixin. Being reachable over SSH seems as another one to me. Unfortunately, eventually these mixins would all be used together, as `artemis`, `beaker`, `connect`, `minute`, and `virtual` guests are the only SSH-backed guests, and they would include the SSH mixin and the bootc mixin since they must support both. And `container` and `mock` would include none of these two. So we'd still have two groups, one with both SSH and bootc support, and one group with neither... I guess it's dragged by the history and lack of time.

@happz should I work on also making the SSH functionality a mixin here? Feels like a separate patch rather?

thrix:
> > Unfortunately, eventually these mixins would all be used together, as `artemis`, `beaker`, `connect`, `minute`, and `virtual` guests are the only SSH-backed guests, and they would include the SSH mixin and the bootc mixin since they must support both.
> 
> But they would still have distinct implementation for each one. Am considering something like `executor`/`execution_driver` etc. that abstracts the usage of the `guest.execute` more explicitly. This can then be specialized granularly for each ssh execution types to support bootc or not, and inside the bootc a similar split for how to create the bootc image and how to provide it to each `virtual`, `beaker`, etc. provision types.

@LecrisUT I will open an alternative implementation using an `executor/execution_driver` abstraction pattern and we can compare which we like more. This one I will keep as a mixin.

tmt/steps/provision/__init__.py
thrix: it is a proper fact now

tmt/steps/provision/__init__.py
thrix: Resolving, this comment is obsolete from the latest conversation

happz:
> > > > Actually, we discussed this today with @thrix , and it turns out `GuestSsh` is the right place after all. The target here is not the "`bootc` provision plugin", but "a guest powered by bootc" - e.g. a Beaker or 1minutetip machine, governed and managed by bootc & co. That why there is the "is bootc" flag/fact; if true, on the remote end, behind the SSH connection, is a guest spawned from an image, and requires bootc workflow to install packages, run playbooks and so on. So it's not bootc leaking into `GuestSsh` as I originally thought, but new mode of guest behavior.
> > > 
> > > 
> > > To clarify, this is about the flags and facts, not about the logic, right? If so, agreed.
> > 
> > 
> > We discussed it in the context of both flags and facts and the functionality, since the functionality, as mentioned above, is supposed to work across all SSH-backed guests.
> > > To me the handling logic of the `make_change` and such should be in a dedicated guest type.
> > 
> > 
> > It sounds like a trait to me, or a mixin. Being reachable over SSH seems as another one to me. Unfortunately, eventually these mixins would all be used together, as `artemis`, `beaker`, `connect`, `minute`, and `virtual` guests are the only SSH-backed guests, and they would include the SSH mixin and the bootc mixin since they must support both. And `container` and `mock` would include none of these two. So we'd still have two groups, one with both SSH and bootc support, and one group with neither... I guess it's dragged by the history and lack of time.
> 
> @happz should I work on also making the SSH functionality a mixin here? Feels like a separate patch rather?

Definitely a separate patch. A rock in our shoe we do not necessarily need to remove today, but it's worth looking into one day.

tmt/steps/provision/__init__.py
thrix: changed to `immediately=False` as default (with the rename of course)

tmt/steps/provision/__init__.py
thrix: moved to `immediately`

tmt/steps/provision/__init__.py
```diff
@@ -935,6 +936,32 @@ def _query_is_ostree(self, guest: 'Guest') -> Optional[bool]:
...
+        # Check if bootc command exists and system is booted from a bootc image.
+        # Uses same logic as bootc package manager probe: check that neither
+        # booted nor image are null in the status output.
+        output = self._execute(
+            guest,
+            ShellScript(
+                """
+                type bootc >/dev/null 2>&1 || { echo no; exit 0; }
+                sudo bootc status --format yaml 2>/dev/null \
+                    | grep -qE 'booted: null|image: null' && echo no || echo yes
```
thrix: needs proper review if this is the correct way

tmt/steps/provision/__init__.py
```diff
@@ -1520,6 +1549,78 @@ def update(self, *, logger: tmt.log.Logger) -> None:
...
+    Used by bootc guests where prepare commands should be collected into a
+    Containerfile rather than executed immediately on the live guest.
+
+    When a guest implements this mixin, commands with ``immediately=False``
+    are collected rather than executed immediately. The collected commands
+    are then executed in batch when :py:meth:`flush_collected` is called
+    (typically at the end of a step).
+
+    Commands with ``immediately=True`` (default) or ``test_session=True``
+    are always executed immediately, even on guests that implement this mixin.
```
thrix: this should documentation should rather move I would say

tests/prepare/bootc/test.sh
thrix: thanks, moved

tmt/steps/provision/__init__.py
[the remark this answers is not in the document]
thrix: outdated

tmt/steps/provision/__init__.py
```diff
@@ -3082,10 +3345,30 @@ def execute(
...
 
         sourced_files = sourced_files or []
 
+        # For bootc guests: flush any pending commands before running a test
+        if test_session and self._has_bootc_collected_commands():
+            self._flush_bootc_commands()
+
+        # For bootc guests: collect system-modifying commands instead of executing
+        # Commands with make_changes=True (default) are collected for later
+        # execution via a Containerfile build, unless test_session=True.
```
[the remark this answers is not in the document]
thrix: outdated

tmt/steps/provision/__init__.py
```diff
@@ -2214,6 +2324,16 @@ def execute(
...
+        :param test_session: if True, this is the actual test being run.
+        :param make_changes: if True, the command modifies system state.
+            For bootc guests implementing :py:class:`GuestCommandCollector`,
+            commands with ``make_changes=True`` are collected into a Containerfile
+            rather than executed immediately. Commands with ``make_changes=False``
+            (default, probes/checks) are always executed immediately.
+            For bootc guests implementing :py:class:`GuestCommandCollector`,
+            commands with ``make_changes=True`` are collected into a Containerfile
+            rather than executed immediately. Commands with ``make_changes=False``
+            (probes, checks) are always executed immediately.
```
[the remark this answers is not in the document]
thrix: outdated

tmt/steps/provision/__init__.py
thrix: looks ok, tested manually.

tmt/steps/provision/__init__.py
thrix: it did

tmt/steps/provision/__init__.py
thrix: Should be cleaner now this has nothing to do with bootc provisioner.

tmt/steps/provision/__init__.py
thrix: Resolving, it is `has_collected_commands` now. And it was agreed the implementation is now in `GuestSsh`.

tests/prepare/bootc/test.sh
thrix: resolved

tmt/steps/provision/__init__.py
thrix: it is `immediately` now

tmt/steps/provision/__init__.py
thrix: resolving, I believe this is outdated, please re-review

tmt/steps/provision/__init__.py
```diff
@@ -3082,10 +3251,25 @@ def execute(
...
         sourced_files = sourced_files or []
 
+        # For guests in image mode collect non testing commands with
+        # immediately=False for later batch execution.
+        # Prevent infinite recursion accessing facts in execute.
+        # Facts gathering calls execute also.
+        facts = self.__dict__.get('facts')
+        if facts is not None and facts.in_sync and self.facts.is_image_mode and not immediately:
+            self.collect_command(command, cwd=cwd, env=env)
+            return tmt.utils.CommandOutput(stdout="", stderr="")
```
happz: Hmmm, an idea: we could use `@overload` to create special annotation when `execute` is called with `immediate=False`, and mark `execute` as returning `None`, for example. Because, if dummy `CommandOuput` is returned, it's not possible to distinguish it from an actual one, and some code may easily use it, not realizing it's meaningless, and make decisions based on its content.

```python
output = guest.execute("enable-some-strong-security-kernel-module", immediate=False)

if 'enabled' not in output.stdout:
    raise GeneralError"("Failed to enable the strong security kernel module.")
```

And that's wrong because the command was postponed and collected and neither plugin author nor reviewers noticed that it would not really run and produce testable output. Returning dummy output seems like the riskiest option available.

tests/prepare/bootc/data/plans.fmf
```diff
@@ -0,0 +1,14 @@
+discover:
+    how: fmf
+provision:
+    how: virtual
+    image: https://artifacts.dev.testing-farm.io/images/CentOS-Stream-10-image-mode-x86_64.qcow2
```
LecrisUT: Shouldn't this be defined from the makefile? At the very least this should be at the `/centos-stream-10` level

tmt/steps/provision/__init__.py
LecrisUT: Alternatively, check `immediately` first and then run the `self.facts`.

tmt/steps/provision/__init__.py
LecrisUT: Can we `@override` the type-checking with and without `immediately` to return `None` instead? If we have anywhere in the code that we explicitly set `immediately=False` and we were trying to process the `CommandOutput` then it should be an error.

tmt/steps/provision/__init__.py
happz: Which, BTW, is exactly what's happening in `prepare/shell` :(

```python
            output, error, timer = Stopwatch.measure(_invoke_script, command, script_environment)

            if error is not None:
                if isinstance(error, tmt.utils.RunError):
                    self._post_action_pull(
                        guest=guest,
                        path=self.phase_workdir,
                        pull_options=pull_options,
                        exceptions=outcome.exceptions,
                    )

                return self._save_failed_run_outcome(
                    log_filepath=script_log_filepath,
                    label=script_name,
                    timer=timer,
                    guest=guest,
                    command=command,
                    exception=error,
                    outcome=outcome,
                )

            if output is None:
                return self._save_error_outcome(
                    label=script_name,
                    timer=timer,
                    note='Command produced no output but raised no exception',
                    guest=guest,
                    outcome=outcome,
                )

            self._post_action_pull(
                guest=guest,
                path=self.phase_workdir,
                pull_options=pull_options,
                exceptions=outcome.exceptions,
            )

            self._save_success_outcome(
                log_filepath=script_log_filepath,
                label=script_name,
                timer=timer,
                guest=guest,
                output=output,
                outcome=outcome,
            )
```

`guest.execute()` pretends nothing happened, and `prepare/shell` reports success, saves logs, resorcd successful result, tells user it succeeded. It will pull phase workdir after each script needlessly.

Seeing it I think we need a sentinel, a special value meaning "accepted but postponed", and plugins need to learn to recognize. The transparency of the image mode is good, command collection, image rebuild, reboot, all that works, but we must not allow plugins pretend like the action succeeded - it did not, it did not even run yet.

```python
# Somewhere in tmt.utils:
CommandCollected = CommandOutput(None, None)


```python
            output, error, timer = Stopwatch.measure(_invoke_script, command, script_environment)

            if error is not None:
                if isinstance(error, tmt.utils.RunError):
                    self._post_action_pull(
                        guest=guest,
                        path=self.phase_workdir,
                        pull_options=pull_options,
                        exceptions=outcome.exceptions,
                    )

                return self._save_failed_run_outcome(
                    log_filepath=script_log_filepath,
                    label=script_name,
                    timer=timer,
                    guest=guest,
                    command=command,
                    exception=error,
                    outcome=outcome,
                )

            # This would be new branch - nothing really happened, so no success, but `INFO`
            # result with a note stating "you're in image mode, not really running this right now"
            # or something like that.
            if output is CommandCollected:
                self._save_collected_run_outcome(
                    log_filepath=script_log_filepath,
                    label=script_name,
                    timer=timer,
                    guest=guest,
                    outcome=outcome,
                )

                continue

            if output is None:
                return self._save_error_outcome(
                    label=script_name,
                    timer=timer,
                    note='Command produced no output but raised no exception',
                    guest=guest,
                    outcome=outcome,
                )

            self._post_action_pull(
                guest=guest,
                path=self.phase_workdir,
                pull_options=pull_options,
                exceptions=outcome.exceptions,
            )

            self._save_success_outcome(
                log_filepath=script_log_filepath,
                label=script_name,
                timer=timer,
                guest=guest,
                output=output,
                outcome=outcome,
            )
```

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2793,79 @@ def __init__(
...
+        Collect a command for image mode container build.
+
+        Adds a RUN directive to the bootc package manager's Containerfile.
+        Uses the same environment and cwd handling as regular execute().
+        """
+
+        if not isinstance(self.package_manager, tmt.package_managers.bootc.Bootc):
+            raise tmt.utils.GeneralError(
+                'The `collect_command` function called for incorrect package manager.'
+            )
```
LecrisUT: This conflicts with the definitions of the other methods (here raising an error, but the other being allowed)

tmt/steps/provision/__init__.py
```diff
@@ -935,6 +938,32 @@ def _query_is_ostree(self, guest: 'Guest') -> Optional[bool]:
...
+        """
+        # Check if bootc command exists and system is booted from a bootc image.
+        # Uses same logic as bootc package manager probe: check that neither
+        # booted nor image are null in the status output.
+        output = self._execute(
+            guest,
+            ShellScript(
+                """
+                type bootc >/dev/null 2>&1 || { echo no; exit 0; }
+                sudo bootc status --format yaml 2>/dev/null \
```
LecrisUT: Don't like the hard-coding of `sudo` in here. Is it really necessary?

tmt/steps/provision/__init__.py
```diff
@@ -935,6 +938,32 @@ def _query_is_ostree(self, guest: 'Guest') -> Optional[bool]:
...
+        a booted image in its status. See https://containers.github.io/bootc/.
+        """
+        # Check if bootc command exists and system is booted from a bootc image.
+        # Uses same logic as bootc package manager probe: check that neither
+        # booted nor image are null in the status output.
+        output = self._execute(
+            guest,
+            ShellScript(
+                """
+                type bootc >/dev/null 2>&1 || { echo no; exit 0; }
```
LecrisUT: This is trying to check for the presence of `bootc` command right? Would be nice if we can avoid it or unify it. At a quick glance, why don't we just use `_query` and let it fail if `bootc` failed for whatever reason. We implicitly do that for the other cases anyway.

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2793,79 @@ def __init__(
...
+        Delegates to the bootc package manager's build_container() method.
+        """
+        if not self.facts.is_image_mode or not isinstance(
+            self.package_manager, tmt.package_managers.bootc.Bootc
+        ):
+            return
+
+        if not self.package_manager.engine.containerfile_directives:
+            self.debug("No collected commands to flush.")
+            return
```
LecrisUT: Why not call `has_collected_commands` here, and it also seems like it will never be called because it is gated in `on_step_complete`.

tmt/steps/provision/__init__.py
```diff
@@ -2257,6 +2336,17 @@ def stop(self) -> None:
...
+    def on_step_complete(self, step: 'tmt.steps.Step') -> None:
+        """
+        Called when a step completes execution on this guest.
+
+        Flush collected commands for Image Mode if there are some.
+
+        :param step: the step that has completed.
+        """
+        if isinstance(self, CommandCollector) and self.has_collected_commands:
+            self.flush_collected()
```
LecrisUT: The definition seems generic, but the help and the implementation are specific to image mode. Can you make it a no-op and move the definition to `CommandCollector`/`GuestSsh`?

tmt/steps/prepare/__init__.py
```diff
@@ -504,6 +504,11 @@ def _record_exception(
                 causes=exceptions,
             )
 
+        # Notify guests that the prepare step has completed.
+        # This allows guests in image mode to apply collected commands to the running system.
```
LecrisUT: image mode specific comment.

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2793,79 @@ def __init__(
...
+        if isinstance(command, tmt.utils.Command):
+            script_parts.append(command.to_script())
+        else:
+            script_parts.append(command)
+
+        # Combine all parts into a single script
+        full_script = ShellScript.from_scripts(script_parts)
+
+        # Add to the package manager's engine
+        self.package_manager.engine.open_containerfile_directives()
```
LecrisUT, line 2833: Unrelated to this PR, but this should be a context manager.

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2793,79 @@ def __init__(
...
+
+        # Add working directory change (properly quoted like in execute())
+        if cwd:
+            script_parts.append(ShellScript(f'cd {quote(str(cwd))}'))
+
+        # Add the actual command
+        if isinstance(command, tmt.utils.Command):
+            script_parts.append(command.to_script())
+        else:
+            script_parts.append(command)
```
LecrisUT: Why the list of shellscripts? Prefer if this would use the built-in `+=` etc. of the `ShellScript`

tmt/steps/provision/__init__.py
LecrisUT: :+1: to the type-checking.

I am not sure about specializing `CommandCollected` though. Would prefer if we specialize on the fact of the guest, something like `guest.facts.will_deffer_scripts`. Also would bypass the whole `Stopwatch.measure` in that case.

happz: > 👍 to the type-checking.
> 
> I am not sure about specializing `CommandCollected` though. Would prefer if we specialize on the fact of the guest, something like `guest.facts.will_deffer_scripts`.

We would still need to return something from `guest.execute`, something that could not be mistaken for an actual command output. It can be `None` though, since there really is no command output to present, `None` would communicate the message of "acceptable but not executed" as well. But that would require some amount of `CommandOutput` becoming `Optional[CommandOutput]`.

> Also would bypass the whole `Stopwatch.measure` in that case.

I wouldn't, it makes "run the action" simpler to handle. We would have two ways to run something, both calling `guest.execute` in some way... Doesn't look appealing.

LecrisUT: > > Also would bypass the whole `Stopwatch.measure` in that case.
> 
> I wouldn't, it makes "run the action" simpler to handle. We would have two ways to run something, both calling `guest.execute` in some way... Doesn't look appealing.

Generally agree, but it seems that a special handling must be added either way. Unless we fix the handling of `if output is None:` to mean execution was deferred? Leaning towards this one unless I am missing something in how the output of `Stopwatch.measure` is defined differently?

happz: Yes, the `if output is None:` would suddenly have an actual meaning. Right now, it's unexpected - no exception, but no command output - and I was unable to convince mypy that there are just two options, output + no exception or no output + exception, therefore we have this strange branch, to appease linters. `Stopwatch.measure()` will not return `None` and `None`, but I failed to convince type checkers. Using `None` as a sentinel would give it a meaning.

thrix: sure.

tmt/package_managers/bootc.py
thrix: Added explanation, and used a empty list rather.

tests/prepare/bootc/data/plans.fmf
thrix: Moved to the /centos-stream-10 level

tmt/steps/provision/__init__.py
thrix: Should be done @LecrisUT

tmt/steps/provision/__init__.py
thrix: bootc always must run as root:

```
bash-5.3$ bootc status
error: Status: Preparing for write: Querying root privilege: This command must be executed as the root user
bash-5.3$ sudo bootc status
● Booted image: ghcr.io/thrix/workstation:latest
        Digest: sha256:57d6f32adc196b2a228be8b3a9e46ce699b20472e6f035cf03b0cfcbd6876156 (amd64)
       Version: 43.20260112.0 (2026-01-13T07:46:01Z)

  Rollback image: ghcr.io/thrix/workstation:latest
          Digest: sha256:bd7577b49763be3621b1989f015c447d6ac43d754cfdad14ea0a30897c9ad2d4 (amd64)
         Version: 43.20260110.0 (2026-01-11T07:43:56Z)
```

But I can try to detect if we are root and add it only in that case?

tmt/steps/provision/__init__.py
thrix: I can try to make it simpler.

tmt/steps/prepare/__init__.py
thrix: removed.

tmt/steps/provision/__init__.py
thrix: moved

tmt/steps/provision/__init__.py
thrix: @happz hopefully addressed, please check.

tmt/steps/provision/__init__.py
thrix: it was on purpose, but seems like an overkill, made it return None

tmt/steps/prepare/shell.py
```diff
@@ -268,13 +269,22 @@ def _invoke_script(
...
-                    label=script_name,
-                    timer=timer,
-                    note='Command produced no output but raised no exception',
-                    guest=guest,
-                    outcome=outcome,
+                # Command was collected for deferred batch execution (image mode).
+                # Record an INFO result and continue to the next script.
+                from tmt.result import PhaseResult, ResultGuestData
+
+                outcome.results.append(
```
happz: Please, add a helper for this action into `self` - we will have more such situations. See `_save_failed_run_outcome` and `_save_error_outcome`, the new helper could be something like `_save_deferred_run_outcome`:

```python
    def _save_deferred_run_outcome(
        self,
        *,
        label: str,
        timer: Stopwatch,
        guest: 'Guest',
        command: Optional[Union[Command, ShellScript]] = None,
        outcome: PluginOutcome,
    ) -> PluginOutcome:
        from tmt.result import PhaseResult, ResultGuestData

        outcome.results.append(
            PhaseResult(
                name=label,
                result=tmt.steps.ResultOutcome.INFO,
                note=['Command collected for deferred execution'],
                guest=ResultGuestData.from_guest(guest=guest),
                start_time=timer.start_time_formatted,
                end_time=timer.end_time_formatted,
                duration=timer.duration_formatted,
            )
        )

        return outcome
```

tmt/steps/provision/__init__.py
LecrisUT: Hmm, yeah, probably not worth it. Getting `sudo_prefix` is quite complicated within the fact querying. Would probably need some refactoring around it.

tmt/steps/provision/__init__.py
```diff
@@ -3056,16 +3231,17 @@ def setup(self) -> None:
                     mkdir -p {workdir_root};
                     setfacl -d -m o:rX {workdir_root}
                     """
-                )
+                ),
             )
 
-    def execute(
+    def execute(  # type: ignore[override]
```
happz: I know it sucks, but it would be probably better to introduce `Optional` in the `Guest` class. I'm not sure how big the fallout would be...

tmt/steps/provision/__init__.py
thrix: it was dropped, guarded anyway in on_step_complete.

Moved the debug message there.

tmt/steps/provision/__init__.py
thrix: resolving, follow up comments deal with this.

tmt/steps/provision/__init__.py
thrix: trying

tmt/steps/provision/__init__.py
thrix: ok, ty

tmt/steps/prepare/shell.py
thrix: @happz this one is addressed, PTAL pls

tmt/steps/provision/__init__.py
thrix: @LecrisUT this one was changed to mimic the collecting of remote_commands.

tmt/steps/provision/__init__.py
thrix: Resolving, it was addressed.

tmt/steps/provision/__init__.py
thrix: @LecrisUT not able to comply with this one, `self._query` seems to be used to extract some fact from commands, but that is not something I need here.

tmt/steps/provision/__init__.py
thrix:
```
tmt/package_managers/__init__.py:278: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/__init__.py:285: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/__init__.py:292: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/__init__.py:295: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/__init__.py:305: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/__init__.py:316: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/__init__.py:327: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/steps/provision/__init__.py:547: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/provision/__init__.py:563: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/provision/__init__.py:2692: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/provision/__init__.py:2693: error: Item "None" of "Optional[CommandOutput]" has no attribute "stderr"  [union-attr]
tmt/steps/provision/__init__.py:2695: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/execute/__init__.py:520: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/bootc.py:107: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/bootc.py:205: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/prepare/install.py:242: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/prepare/install.py:243: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/prepare/distgit.py:240: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/steps/prepare/distgit.py:283: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/rpm_ostree.py:145: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/mock.py:102: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/mock.py:145: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/package_managers/dnf.py:227: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/apt.py:252: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/apt.py:252: error: Item "None" of "Optional[CommandOutput]" has no attribute "stderr"  [union-attr]
tmt/package_managers/apk.py:165: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/package_managers/apk.py:165: error: Item "None" of "Optional[CommandOutput]" has no attribute "stderr"  [union-attr]
tmt/checks/dmesg.py:90: error: Incompatible return value type (got "Optional[CommandOutput]", expected "CommandOutput")  [return-value]
tmt/checks/coredump.py:240: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/checks/coredump.py:245: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/checks/coredump.py:282: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/checks/coredump.py:306: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/checks/coredump.py:329: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/checks/coredump.py:403: error: Item "None" of "Optional[CommandOutput]" has no attribute "stdout"  [union-attr]
tmt/checks/avc.py:151: error: Incompatible return value type (got "tuple[Optional[CommandOutput], None]", expected "Union[tuple[CommandOutput, None], tuple[None, RunError]]")  [return-value]
Found 35 errors in 14 files (checked 145 source files)
```

tmt/steps/provision/__init__.py
thrix: @LecrisUT fixed, PTAL

tmt/steps/provision/__init__.py
LecrisUT: Still relevant, you can simplify the if now with:
```python
        # For guests in image mode collect non testing commands with
        # immediately=False for later batch execution.
        if not immediately and self.facts.is_image_mode:
            self.collect_command(command, sourced_files=sourced_files, cwd=cwd, env=env)
            return None
```

LecrisUT (approved):
Just a minor comment (https://github.com/teemtee/tmt/pull/4525#discussion_r2788681767). I guess the `overload` issue is addressed. It is quite a bit messy, hopefully it can be improved if we move the execute logic to its own thing

tmt/steps/prepare/shell.py
thrix: Resolving, should be addressed

tmt/steps/provision/__init__.py
thrix: Resolved by happz

tmt/package_managers/bootc.py
```diff
@@ -251,10 +261,10 @@ def build_container(self) -> None:
                         f'  ( podman pull {base_image} || podman pull containers-storage:{base_image} )'  # noqa: E501
                         f'  || bootc image copy-to-storage --target {base_image}'
                         ')"'
-                    )
+                    ),
                 )
                 self.guest.execute(
-                    ShellScript(f'cat <<EOF > {containerfile_path!s} \n{containerfile} \nEOF')
+                    ShellScript(f'cat <<EOF > {containerfile_path!s} \n{containerfile} \nEOF'),
```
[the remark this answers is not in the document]
thrix: the content of containerfile is something we create!

tmt/steps/provision/__init__.py
thrix: indeed, I do not understand why, but it is cool. 

Dropped recursion check ...

thrix:
> Just a minor comment ([#4525 (comment)](https://github.com/teemtee/tmt/pull/4525#discussion_r2788681767)). I guess the `overload` issue is addressed. It is quite a bit messy, hopefully it can be improved if we move the execute logic to its own thing.
> 
> Edit: Also please adjust the description, e.g. "Rename make_changes to immediately" is invalid

@LecrisUT updated

tmt/steps/provision/__init__.py
thrix: @LecrisUT filed under #4568

tmt/steps/provision/__init__.py
```diff
@@ -2257,6 +2339,23 @@ def stop(self) -> None:
 
         raise NotImplementedError
 
+    def on_step_complete(self, step: 'tmt.steps.Step') -> None:
```
happz: Hmm, this suggests there is a conflict: `on_step_complete` is part of the `CommandCollector` API, but it is defined here too, as if `Guest` supports `CommandCollector` API, partially, but does nothing unless being `CommandCollector` instance. If `on_step_complete` is supposed to be part of `Guest` API, then it should not be part of `CommandCollector` API which is not part of `Guest` API.

tmt/steps/provision/__init__.py
```diff
@@ -2580,7 +2679,7 @@ def mkdtemp(
         """
 
         output = self.execute(
-            self._construct_mkdtemp_command(prefix=prefix, template=template, parent=parent)
+            self._construct_mkdtemp_command(prefix=prefix, template=template, parent=parent),
```
happz: Unrelated change, added comma.

tmt/steps/provision/__init__.py
```diff
@@ -3056,16 +3231,18 @@ def setup(self) -> None:
                     mkdir -p {workdir_root};
                     setfacl -d -m o:rX {workdir_root}
                     """
-                )
+                ),
```
happz: Unrelated change.

tmt/steps/provision/__init__.py
thrix: @LecrisUT ^

thrix: I moved it on his request :dancers:

tmt/package_managers/bootc.py
```diff
@@ -263,9 +273,11 @@ def build_container(self) -> None:
 
                 assert self.guest.parent is not None
 
+                # Mount run_workdir so scripts have access to tmt files during build.
+                # Use :Z for SELinux private label.
                 self.guest.execute(
                     ShellScript(
-                        f'{self.guest.facts.sudo_prefix} podman build -t {image_tag} -f {containerfile_path} {self.guest.step_workdir}'  # noqa: E501
+                        f'{self.guest.facts.sudo_prefix} podman build -v {self.guest.run_workdir}:{self.guest.run_workdir}:Z -t {image_tag} -f {containerfile_path} {self.guest.run_workdir}'  # noqa: E501
```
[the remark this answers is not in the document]
thrix: @gemini-code-assist nah, thanks, but this should be ok. those paths are governed by us

tmt/steps/provision/__init__.py
LecrisUT: The previous version was doing an `isinstance` and then calling it. So would much prefer to have a generic no-op then that.

The conflict is probably with marking the other one as `abc` which ie no longer needed and it is instead part of normal guest api

tmt/steps/provision/__init__.py
```diff
@@ -2698,6 +2802,77 @@ def __init__(
...
+        if isinstance(command, tmt.utils.Command):
+            collected_commands += command.to_script()
+        else:
+            collected_commands += command
+
+        collected_command = collected_commands.to_element()
+
+        # Add to the package manager's engine
+        self.package_manager.engine.open_containerfile_directives()
+        self.package_manager.engine.containerfile_directives.append(f"RUN {collected_command}")
```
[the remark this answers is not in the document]
thrix: @gemini-code-assist how would you fix it?

tmt/steps/provision/__init__.py
thrix: thanks, removed

tmt/steps/provision/__init__.py
thrix: thanks, removed

tmt/steps/provision/__init__.py
thrix: @happz should be resovled as agreed in Slack:

https://redhat-internal.slack.com/archives/C08J5QPSSKH/p1770763591145219?thread_ts=1770758355.224859&cid=C08J5QPSSKH

tmt/steps/provision/__init__.py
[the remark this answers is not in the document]
thrix: as discussed, user already has root on the machine anyway.

Issue #4055, opened 2025-09-12
Prepare artifacts list with version details


psss:
Store a list of all packages included in the artifacts repository including the precise versions.

AthreyVinay:
@psss The logs already include downloaded artifact metadata. Is the expectation to extract and persist this information in a structured artifact list(exists here: https://github.com/teemtee/tmt/blob/a6de51c23e25ce21ac4dd47d36a6efcbe2893ee2/tmt/steps/prepare/artifact/providers/__init__.py#L111) or is relying on log output sufficient? Thaks.

vaibhavdaren:
**Notes from discussion** in review meeting.

Dump list of all artifacts that are provided by a specific provider in yaml format and persist in `{workdir}` to be then later used to compare whats specified vs installed .

AthreyVinay:
After some work done to refactor `ArtifactInfo` #4526 , I think the problem statement  of this issue should be:

`artifacts.yaml` tracks only what was actually downloaded, not what was merely "available". 

This means those providers that acutally download (koji, brew, copr-build, file) artifacts would call `get_installable_artifacts()` to persist whats downloaded & made available for installation rather that providers that enable(repository-url, copr-repository...) repositories.

psss:
> `artifacts.yaml` tracks only what was actually downloaded, not what was merely "available".

Hm, seems we'll need some more discussion here. See also #4546.

AthreyVinay:
Yep - good to have the discussion, however a lot of text described in #4546 - easy to get confused as to what the objective is.

AthreyVinay:
@psss Please see the comment on the submitted pr to resolve this: https://github.com/teemtee/tmt/pull/4558#issuecomment-3859970753

PR #4526, opened 2026-01-27
Refactor the `ArtifactInfo` class

AthreyVinay:
An effort towards resolving: #4055 (part-1)

following changes made:
1. All subclasses of `ArtifactInfo` now removed
2. Updated attributs of `ArtifactInfo` to what was discussed during the artifact plugin design sync (`id`, `version`, `name`, `location`, `provider`). The `__str__` of this container represents: `<provider-id>-<nvra>`
3. Introduced `Version` object and its subclass (`RpmVersion` with its factory methods) to represent `nvra` & `nevra` construction
4. Refactored all classes (src and tests) where the above was referenced
5. ~Refactored `artifacts` propeorty to `get_installable_artifacts` method. Reason being only persist information about artifacts that would be downloaded i.e. list from providers that "download" and not "enable".~
6. Removed `ArtifactInfoT` generic since see point 1.

couple more to finish before moving to review:
~1. move `ArtifactInfo` to someplace else visible to package_managers too.~ The providers **shall** be responsible to create `ArtifactInfo` objects.

### DIFF
--- a/tmt/steps/prepare/artifact/providers/__init__.py
+++ b/tmt/steps/prepare/artifact/providers/__init__.py
@@ -129,7 +129,7 @@ class ArtifactInfo:
 
         TODO: Transient for now, modify based on the decision made here: https://github.com/teemtee/tmt/issues/4546
         """
-        return str(self.version)
+        return self.version.nvra
 
     @property
     def name(self) -> str:
@@ -140,7 +140,7 @@ class ArtifactInfo:
         """
         This is the filename of the artifact.
         """
-        return f"{self.provider.id}-{self.version}.rpm"
+        return f"{self.id}.rpm"
 
     def __str__(self) -> str:
         return f"{self.version} ({self.provider.id})"
--- a/tests/unit/artifact/test_base.py
+++ b/tests/unit/artifact/test_base.py
@@ -39,7 +39,7 @@ def test_download_artifacts(tmp_path, root_logger):
     provider = MockProvider("mock:123", repository_priority=50, logger=root_logger)
 
     paths = provider.fetch_contents(guest, tmp_path, [])
-    file_path = tmp_path / "123-mock-1.0-1.x86_64.rpm"
+    file_path = tmp_path / "mock-1.0-1.x86_64.rpm"
     assert file_path in paths
     assert file_path.exists()
     assert file_path.read_text() == "ok"


tmt/steps/prepare/artifact/__init__.py
```diff
@@ -195,7 +176,7 @@ def go(
         self,
         *,
         guest: Guest,
-        environment: Optional[Environment] = None,
+        environment: typing.Optional[Environment] = None,
```
happz: I'd prefer dropping `import typing` in favor of `from typing import Optional`. In our code we tend towards importing these names rather than using `typing.foo`.

LecrisUT: I know there is at least a ruff to do the opposite (`ICN003`), but it doesn't seem we have the opposite ([issue to track](https://github.com/astral-sh/ruff/issues/21025))

AthreyVinay: Addressed here: [99b5726](https://github.com/teemtee/tmt/pull/4526/commits/99b57260e7989bdbad0f455393a4542590dfd85b)

tmt/steps/prepare/artifact/providers/repository_url.py
```diff
@@ -58,15 +56,14 @@ def _extract_provider_id(cls, raw_provider_id: str) -> ArtifactProviderId:
             raise ValueError("Missing repository baseurl.")
         return value
 
-    @property
-    def artifacts(self) -> Sequence[RpmArtifactInfo]:
+    def get_installable_artifacts(self) -> Sequence[ArtifactInfo]:
```
vaibhavdaren: I think all the artifacts are installable. Lets name it something more appropriate. 
We can decide during hacking session up for suggestions.

tmt/package_managers/__init__.py
```diff
@@ -304,22 +294,6 @@ def install_repository(self, repository: "Repository") -> CommandOutput:
         """
         return self.guest.execute(self.engine.install_repository(repository))
 
-    def list_packages(self, repository: "Repository") -> list[str]:
```
vaibhavdaren, line 307: I don't see the point in deleting this? How will I list the artifacts from a repository if I do not have this functionality in the package manager?

tests/unit/artifact/test_repository.py
```diff
@@ -229,107 +228,6 @@ def test_url_no_path(root_logger):
             Repository.from_url(url="https://example.com/", logger=root_logger)
 
 
-@pytest.mark.parametrize(
```
vaibhavdaren, line 232: Please instead of deleting these tests , please change these tests to UnitTesting the `VersionFromFilename` class from your code : https://github.com/teemtee/tmt/pull/4526/changes#diff-e997bb498ff41c9b641bfe2a553423df7ca65c98b88be174212bd1b4ed4c0352R94.

Since this code is already reviewed it already contains the various test cases.

tmt/steps/prepare/artifact/providers/repository_url.py
AthreyVinay: > I think all the artifacts are installable.

Not all artifacts are installable. Repository providers return [] because they just enable repos without downloading packages. 

> Lets name it something more appropriate

The name "installable" specifically means "downloaded and ready for installation," which distinguishes them from merely "available" artifacts.

tmt/package_managers/__init__.py
AthreyVinay: Why would you want to list artifacts from a repository? Enumerating 100's of packages that might not be downloaded is unnecessary. Also as discussed in the sync today, `artifacts.yaml` would only be populated for packages that are downloaded. 

We can surely add it back when it is actually used somewhere :)

tests/unit/artifact/test_repository.py
AthreyVinay: Please see `test_koji.py`

tmt/steps/prepare/artifact/providers/koji.py
```diff
@@ -61,26 +62,71 @@ def import_koji(logger: tmt.log.Logger) -> None:
         raise tmt.utils.GeneralError("Could not import koji package.") from error
 
 
-@container
-class ScratchRpmArtifactInfo(RpmArtifactInfo):
+@container(frozen=True)
+class VersionFromRpmMeta(Version):
```
happz: Strongly recommending keeping just `RpmVersion` class, with `from_rpm_meta()` or `from_filename()` classmethods serving as "constructors". That is the actual difference, in how the "rpm version" object is created, not in what fields it has or which methods it has, therefore there is no need to keep different classes because of their origin.

tmt/steps/prepare/artifact/providers/koji.py
```diff
@@ -215,30 +261,27 @@ def contribute_to_shared_repo(
...
+        version_info = VersionFromRpmMeta.create(rpm_meta)
         path = (
             f"packages/{package_name}/"
-            f"{version}/{release}/"
-            f"{arch}/"
-            f"{name}-{version}-{release}.{arch}.rpm"
+            f"{version_info.version}/"
+            f"{version_info.release}/"
+            f"{version_info.arch}/"
+            f"{version_info}.rpm"
```
happz: I'd recommend using `.nvra` instead of implicit string conversion, it will be much clearer which field is expected here.

tmt/steps/prepare/artifact/providers/file.py
```diff
@@ -77,12 +63,17 @@ def _extract_provider_id(cls, raw_provider_id: str) -> ArtifactProviderId:
...
         return ArtifactProviderId(raw_provider_id)
 
-    @cached_property
-    def artifacts(self) -> Sequence[PackageAsFileArtifactInfo]:
-        artifacts: list[PackageAsFileArtifactInfo] = []
+    def make_rpm_artifact(self, path: str) -> ArtifactInfo:
+        filename = tmt.utils.Path(path).name
+        version_info = VersionFromFilename.create(filename)
+
+        return ArtifactInfo(version=version_info, location=path, provider=self)
```
happz: The following would be absolutely acceptable:

```python
return ArtifactInfo(
    version=VersionFromFilename.create(tmt.utils.Path(path).name),
    location=path,
    provider=self
)
```

Neither `filename` nor `version_info` are used multiple times, or cache some large, complicated value. There is barely any benefit in having standalone names for these values.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -33,30 +33,55 @@ class UnsupportedOperationError(RuntimeError):
...
     @property
-    @abstractmethod
-    def location(self) -> str:
-        raise NotImplementedError
+    def name(self) -> str:
+        return self.version.name
 
     def __str__(self) -> str:
-        return self.id
+        return f"{self.provider.id}-{self.version}.rpm"
```
LecrisUT: Is this used in code or only in log. If the latter, how about
```suggestion
        return f"{self.version} ({self.provider.id})"
```

AthreyVinay: ~Just in the log - I dont have a preference. This was decided during the sync - but happy to go ahead with what you suggested.~
Scratch that - its used in the code(to fetch contents as this provides the path) as the the name needs to end with `.rpm` .

tmt/steps/prepare/artifact/providers/file.py
AthreyVinay: Addressed here: [9154f4d](https://github.com/teemtee/tmt/pull/4526/commits/9154f4d1be68511625c22efb58c40260a2bc9552)

tmt/steps/prepare/artifact/providers/koji.py
AthreyVinay: Addressed here: [9154f4d](https://github.com/teemtee/tmt/pull/4526/commits/9154f4d1be68511625c22efb58c40260a2bc9552)

tmt/steps/prepare/artifact/providers/koji.py
AthreyVinay: Addressed here: [9154f4d](https://github.com/teemtee/tmt/pull/4526/commits/9154f4d1be68511625c22efb58c40260a2bc9552)

tmt/package_managers/__init__.py
vaibhavdaren: Regarding the sync, I agree that `artifacts.yaml` should track downloaded packages, but that doesn't negate the need to query the remote repository for its state. . @LecrisUT @psss @happz  what do you think?

The use case isn't just about populating `artifacts.yaml` locally. We need a way to audit or verify the contents of the upstream repository programmatically. If we remove this, we are effectively blinding the package manager to anything.

tmt/steps/prepare/artifact/providers/koji.py
```diff
@@ -61,26 +62,70 @@ def import_koji(logger: tmt.log.Logger) -> None:
         raise tmt.utils.GeneralError("Could not import koji package.") from error
 
 
-@container
-class ScratchRpmArtifactInfo(RpmArtifactInfo):
+@container(frozen=True)
+class RpmVersionClass(Version):
```
happz: Why did you include `Class` in the name of the, hm, class? `RpmVersion` would work as well.

AthreyVinay: Yep - overlooked the name. Shall be fixed.

tmt/package_managers/__init__.py
happz: Hmmm, I would keep those methods around. We don't need to remove them today, in this PR, we can drop them when it's clear we don't need them.

I guess we'll see what components we will need for the workflow proposed during the last sync: run something like `rpm -q --provided $a_required_token_from_test_requires` and make sure the output comes from the shared repo we constructed - if there is the package in this repo, of course. This might be where we'd need the list of artifacts from repositories "installed" by providers, even those like Copr or repository-file, to find out whether the package is offered by our repositories rather than system ones. But right now whether these methods should stay or be removed seems unrelated to moving and renaming of `ArtifactInfo` PR.

AthreyVinay: > Regarding the sync, I agree that artifacts.yaml should track downloaded packages, but that doesn't negate the need to query the remote repository for its state. . @LecrisUT @psss @happz what do you think?
The use case isn't just about populating artifacts.yaml locally. We need a way to audit or verify the contents of the upstream repository programmatically. If we remove this, we are effectively blinding the package manager to anything.


I still dont feel the need to leave this code - is there a concrete use-case? 
1. Are you trying to verify a repository hasn't changed between test runs?
2. Are you auditing what packages are available before enabling a repo?
3. Something else?

Repository providers are  loose - you enable the repo and trust DNF to pull the right packages. If you need strict control, use downloaded artifacts.

Again, open for suggestions :)

AthreyVinay: > But right now whether these methods should stay or be removed seems unrelated to moving and renaming of ArtifactInfo PR.

Not entirely - as this was linked to another issue that was created to move `ArtifactInfo` to package manger classes. However, there could be a housekeeping pr that should remove these (currently) unused methods in artifact plugin.

tmt/steps/prepare/artifact/providers/koji.py
AthreyVinay: Fixed: [refactor: update class name as suggested](https://github.com/teemtee/tmt/pull/4526/commits/652fa412e67d4eb997972aa8bb080a6898cd6ded)

tmt/package_managers/__init__.py
happz: I would leave the decision after the verification is drafted. Because I suspect we will need to know content of these repositories:

1. run `rpm -q --provides ...` to find out which package provides the token test required.
2. check which repository sourced the package by asking dnf.

The thing is, system repositories are still perfectly eligible, not all requirements will be sourced by `artifact` repos - if my test requires `make`, and I'm testing Firefox, I am most likely fine with using the system one. It must be one of the `artifact` repos only when `make` is available in one of the `artifact` repos - and this is where we would need to list content of `artifact` repos, so we get a list of available packages. If we find out the package in question is on the list but dnf does say it was installed from the system repo, we would worry - if it's not on the list, we are fine with whatever dnf says.

So, yeah, I think I do have a use case for "list packages from this repository" methods - not just our shared repository, but I think we will need it for other repositories as well. That's why I would leave the decision whether to drop these for different PR, this one should focus on moving the `ArtifactInfo` class and `Version` and so, and make sure all its uses are correct and updated. If we realize there is no use case, we can drop them - in another PR. Let's focus on one thing and one thing only in each PR.

AthreyVinay: Fair point!

For this PR: I'll keep `list_packages() ` as-is and focus on the `ArtifactInfo/Version` refactoring.
For later: When we draft the verification implementation, we can decide if `list_packages()` is needed or if there's a better approach.

Tahnks.

AthreyVinay: Rebased & removal reverted for this pr.

tmt/steps/prepare/artifact/providers/__init__.py
psss: We dropped `ArtifactInfo` subclasses so this should be dropped as well?

tmt/steps/prepare/artifact/providers/copr_build.py
```diff
@@ -203,36 +204,23 @@ def _fetch_results_json(self) -> list[dict[str, str]]:
...
 
-        artifact._raw_artifact["url"] = urljoin(base_url + "/", filename)
-        return artifact
+        return ArtifactInfo(
+            version=version_info, location=urljoin(base_url + "/", filename), provider=self
+        )
 
-    @cached_property
-    def artifacts(self) -> Sequence[RpmArtifactInfo]:
+    def get_installable_artifacts(self) -> Sequence[ArtifactInfo]:
```
happz: This seems similar to the `list_packages` issue, change not really related to the moving of `ArtifactInfo`. What is the reason `artifacts` property is replaced with `get_installable_artifacts` method (which, then again, could be `installable_artifacts` property...)? `def artifacts(self) -> Sequence[RpmArtifactInfo]:` -> `def artifacts(self) -> Sequence[ArtifactInfo]:` would be enough for the purpose of this patch, i.e. slightly fancy moving the class.

AthreyVinay: Ok - so, the reason this was added in this patch was I thought I would make all the changes related to `ArtifactInfo` here before I created a MR to list the artifacts #4558, assuming the next patch only had changes related to persisting to yaml. Now, in the hidsight, maybe this should've made its way in that?

The thought process to rename `artifacts` to `get_installable_artifacts()` was that `artifacts` was too generic and this only gets populated for the downloadble providers as the repo ones jsut return an `[]`, as stated here: https://github.com/teemtee/tmt/pull/4526#discussion_r2783548471
It was not just a property but a `cached_property` and I thought maybe the consumers of this might not be ready for this to be populated..... and again... since it was renamed to `_get....()`, I made this a method.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Nice one, thanks.
[address comments: Remove ArtifactInfo generic type parameter](https://github.com/teemtee/tmt/pull/4526/commits/6cdc6ac27299134844c378746cf514fc24948626)

tmt/steps/prepare/artifact/providers/copr_build.py
psss: > `def artifacts(self) -> Sequence[RpmArtifactInfo]:` -> `def artifacts(self) -> Sequence[ArtifactInfo]:` would be enough for the purpose of this patch, i.e. slightly fancy moving the class.

Agreed, let's limit changes not essentially related to minimum, this will speed up the review and merge rate.

> The thought process to rename `artifacts` to `get_installable_artifacts()` was that `artifacts` was too generic and this only gets populated for the downloadble providers as the repo ones jsut return an `[]`, as stated here: [#4526 (comment)](https://github.com/teemtee/tmt/pull/4526#discussion_r2783548471)

Yeah, there might be some adjustments needed. But let's not dive into discussion about those here. Let's do small changes step by step.

AthreyVinay: Fixed:
[reverted renaming of artifact property](https://github.com/teemtee/tmt/pull/4526/commits/8613f866b4cac870c2fe7b184d04535cb7e69360)

tmt/steps/prepare/artifact/providers/repository_url.py
AthreyVinay: Fixed: 
[reverted renaming of artifact property](https://github.com/teemtee/tmt/pull/4526/commits/8613f866b4cac870c2fe7b184d04535cb7e69360)

tests/unit/artifact/test_copr_repository.py
```diff
@@ -2,6 +2,8 @@
 
 import pytest
 
+from tmt.steps.prepare.artifact.providers.copr_repository import CoprRepositoryProvider
+
```
psss: Seems that the class is not used. Interesting that it was not caught by `ruff`. Ah, we have it disabled:

* #4566

tmt/steps/prepare/artifact/providers/__init__.py
psss: Thanks!

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -33,41 +33,62 @@ class UnsupportedOperationError(RuntimeError):
...
-class ArtifactInfo(ABC):
+class ArtifactInfo:
     """
     Information about a single artifact, e.g. a package.
     """
 
-    _raw_artifact: Any
+    version: Version
+    location: str
+    provider: "ArtifactProvider"
```
vaibhavdaren, line 123: Will each ArtifactInfo instance now contain a full ArtifactProvider object? When thousands of artifacts are loaded (e.g., large Koji builds or multiple repo files), this could lead to significant memory bloat . Would it be better to store only a provider_id string and look up the provider when needed?

happz: It's not a full object in its entirety, just a pointer.

tmt/steps/prepare/artifact/providers/koji.py
```diff
@@ -61,26 +62,70 @@ def import_koji(logger: tmt.log.Logger) -> None:
         raise tmt.utils.GeneralError("Could not import koji package.") from error
 
 
-@container
-class ScratchRpmArtifactInfo(RpmArtifactInfo):
+@container(frozen=True)
+class RpmVersion(Version):
```
vaibhavdaren: Now that FileArtifactProvider depends on RpmVersion, we have a 'provider-to-provider' dependency. Can we move RpmVersion to a more central location?

tests/unit/artifact/test_repository.py
vaibhavdaren: The regex-based tests that were removed actually served as documentation for the 'happy path' and edge cases of RPM naming conventions in tmt. Even though the regex is gone, the requirements for the parser haven't changed.

Can we port those @pytest.mark.parametrize cases over to a unit test for RpmVersion? This will help us confirm that the new class handles complex strings (like tmt+export-polarion...) as correctly as the old function did?

tmt/steps/prepare/artifact/providers/__init__.py
LecrisUT: That does not sound right. The name is inconsistent with the actual filename downloaded

tests/unit/artifact/test_repository.py
AthreyVinay: Sure, 
for `RpmVersion.from_rpm_meta` : [e73c2ea](https://github.com/teemtee/tmt/pull/4526/commits/e73c2ea62e919cfe4eb9b7b17e04b22d2e156f24)
for `RpmVersion.from_filename` : [f3734d2](https://github.com/teemtee/tmt/pull/4526/commits/f3734d2fdfc7f583f516bea2b662673113bd4709)

tmt/steps/prepare/artifact/providers/koji.py
AthreyVinay: Done: [refactor: move RpmVersion class to provider base](https://github.com/teemtee/tmt/pull/4526/commits/e6a2e0125d13d6f6da330f450b3f6a2651d57240)

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Sorry, I do not understand the comment.
Eg: Expected filename: `123-mock-1.0-1.x86_64.rpm` vs what you are suggesting: `mock-1.0-1.x86_64 (123)`

tests/unit/artifact/test_copr_repository.py
AthreyVinay: ~Shall I wait for it to be fixed in the above pr?~

This import is needed to register the provider in the plugin registry. Just realised after trying to remove it.

tmt/steps/prepare/artifact/providers/__init__.py
LecrisUT: `123-mock-1.0-1.x86_64.rpm` is not an actual physical filename and it might be confusing, so this part is suspicious to me

>  its used in the code(to fetch contents as this provides the path)

AthreyVinay: So, in `ArtifactProvider.fetch_contents()`:
```python
local_path = download_path / str(artifact)
```
This expects `str(artifact)` to return a string with `.rpm` to create the path.

tests/unit/artifact/test_repository.py
vaibhavdaren: I think there is a bug in code
```
Unit Test: 
>>> RpmVersion.from_filename("docker-ce-1:20.10.7-3.el8.x86_64")
RpmVersion(name='docker-ce', version='1:20.10.7', release='3.el8', arch='x86_64', epoch=0)

It should be 
RpmVersion(name='docker-ce', version='1:20.10.7', release='3.el8', arch='x86_64', epoch=1)

Comment (AthreyVinay) [reply tests/unit/artifact/test_repository.py:232]: Hm, this is interesting. I thought filenames dont include `epoch` field (at-least for koji; see: https://koji.fedoraproject.org/koji/buildinfo?buildID=2909847). Can you please point me to the source of this rpm so I can validate the metadata? Tks.

Comment (AthreyVinay) [reply tests/unit/artifact/test_repository.py:232]: From the examples that I see [here](https://download.docker.com/linux/centos/8/x86_64/stable/Packages/), I do not find `epoch`  in the filenames. Maybe its just a `NEVRA` string.

Comment (vaibhavdaren) [reply tests/unit/artifact/test_repository.py:232]: > filenames dont include epoch. 

The test itself was wrong. Confirmed from https://rpm.org/docs/4.19.x/manual/spec.html.

Comment (AthreyVinay) [reply tests/unit/artifact/test_repository.py:232]: Sure - np!

File: tmt/steps/prepare/artifact/providers/__init__.py (hunk 8 above)
Comment (LecrisUT) [reply tmt/steps/prepare/artifact/providers/__init__.py]: Not sure about doing it like that, much rather be very explicit when we are requesting the `filename` in there.

About using a renamed rpm filename, that may backfire on us if somewhere in the chain like `createrepo` does not support the filename format or misinterprets it. Did a quick test locally and at least it does not die, but did not test with more complicated characters.

File: tests/unit/artifact/test_file.py [hunk 14]
```diff
@@ -13,14 +13,16 @@
...
 )
 def test_file_artifact_provider_patterns(tmp_path, pattern, expected_names, artifact_provider):
-    # Create test files
-    for name in expected_names:
-        (tmp_path / name).touch()
+    if "*" in pattern:
+        for i in range(3):
+            (tmp_path / f"package-{i}-1.0-1.fc43.x86_64.rpm").touch()
+    else:
+        (tmp_path / pattern).touch()
```
LecrisUT: Can we keep using the `expected_names` here?

tests/unit/artifact/test_file.py
```diff
@@ -35,26 +37,73 @@ def test_file_artifact_provider_deduplicates_globs(tmp_path, caplog, artifact_pr
...
+def test_version_from_filename_parsing(filename, expected):
+    version = RpmVersion.from_filename(filename)
+
+    assert version.name == expected["name"]
+    assert version.epoch == expected["epoch"]
+    assert version.version == expected["version"]
+    assert version.release == expected["release"]
+    assert version.arch == expected["arch"]
+    assert version.nvra == expected["nvra"]
+    assert str(version) == expected["nvra"]
```
LecrisUT: Can you use the `to_dict` (part of `tmt.container`, might have the name wrong) and check the equality of the whole dict? The current form would not catch any new fields introduced.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: As discussed in the review, makes sense to explicitly call out the `filename` of an artifact instead of reusing the `__str__`. 
Please see: [3db5aa2](https://github.com/teemtee/tmt/pull/4526/commits/3db5aa2ca096cbf4dec58d3ddabbf7122235c2ed)

tests/unit/artifact/test_copr_repository.py
LecrisUT: That is quite funky. There is a load plugin method, but I did not look into how that works. Do you have the bandwitdh to check it?

tests/unit/artifact/test_file.py
AthreyVinay: Okay - so just noticed `to_dict()` belongs to `DataContainer` which makes `Version` inherit from it - maybe just leave it as simple `container`? I will use `vars(version)` instead.

AthreyVinay: Addressed: [cb25be9](https://github.com/teemtee/tmt/pull/4526/commits/cb25be9d18116a4eb988ecb741d68c0eb4005d95)

tests/unit/artifact/test_file.py
AthreyVinay: Addressed: [cb25be9](https://github.com/teemtee/tmt/pull/4526/commits/cb25be9d18116a4eb988ecb741d68c0eb4005d95)

tests/unit/artifact/test_copr_repository.py
AthreyVinay: Maybe to unblock this - we can look at adding this to the housekeeping sprint?

psss:
It's a bit late for the release and this change does not directly affect users.
Let's not block on it and merge as one of the first things in the next sprint.

tests/unit/artifact/test_copr_repository.py
AthreyVinay: ok... so did a bit of digging.
there's a _`explore_packages()` hidden in `explore()` that loads plugins on start... however this does not get called automagically for unit tests, we need to explicitly call it to load the providers. Added a fixture for this. See: [0119fe4](https://github.com/teemtee/tmt/pull/4526/commits/0119fe4bb9e65eeed30121c7f0d775feeb8f326e)
Maybe we could try lazy-loading in the plugin registry (so it auto-discovers on first access)? Obviously as a separate change though.

LecrisUT: Fine for now, there are some pytest constructs that can avoid adding it manually to each test (define a fixture to be used under a conftest scope), but we can get to that later.

AthreyVinay: 'm  a bit confused now - isnt that what I have pushed? a fixture `_load_plugins()` and making `artifact_provider()` depend on it. Am i missing something?

therazix:
The failed rawhide tests are unrelated to the changes in this PR. Merging.

Issue #4546, opened 2026-02-03
Enhance artifact install plugin to verify RPM packages using hashes for same-NVR scenarios


vaibhavdaren:
### Summary

We need a reliable way to verify that the requested RPM packages were installed correctly in the artifact install plugin, especially when multiple scratch builds share the same NVR.

### Scenario

- A package like `curl-1.2.3` exists in system or custom repos.
- User provides a scratch build with the same NVR via Koji task ID.
- Artifact provider downloads the build, creates a high-priority DNF repo, and installs it.
- Need to confirm the rpm from a particular scratch build was installed, not the system one.

### Discussion Points

- Use RPM hashes like `sha256header` (RHEL 8+) or `sha1header` (for compatibility with RHEL <8).
- Example: `rpm -qp --qf "%{sha256header}\n" package.rpm`.
- Hashes identify packages uniquely, but identical hashes mean identical content (acceptable for tmt).
- Repo data doesn't include these headers; DNF repoquery can't fetch them without downloading.
- For repos, SHA-256 digest of RPM file is in `primary.xml` (under `/metadata/package/checksum`), accessible via librepo/libdnf API.
- For older distros (RHEL 6/7), use `sha1header`; consider marking same-NVR cases as unsupported or error out.
- Goal: Support local downloads (Koji/Brew/Copr/local/remote URLs) and remote DNF repos.

### References

- [Slack thread](https://redhat-internal.slack.com/archives/C04RFANDEN4/p1769768409515449)

LecrisUT:
We should discuss more on the merit and how it plays with alternative of checking from something like `dnf info` and the source of the package.

vaibhavdaren:
**Note**: What we have already in place (example from [Integration Tests for Koji Build](https://github.com/teemtee/tmt/tree/main/tests/prepare/artifact/providers/koji-build)):

1. We fetch the package from Koji using the build ID:
   https://github.com/teemtee/tmt/blob/1e67cc3be9b1fba4d6a8e87f8a1db480662f15f0/tests/prepare/artifact/providers/koji-build/test.sh#L20-L22

2. We make the downloaded package available on the guest via `tmt artifact shared repository`.

3. The package is then installed using the `require` instruction in the test metadata:
   https://github.com/teemtee/tmt/blob/1e67cc3be9b1fba4d6a8e87f8a1db480662f15f0/tests/prepare/artifact/providers/koji-build/data/main.fmf#L15-L20

[...]

PR #4510, opened 2026-01-19
Add integration tests for the `file` provider

vaibhavdaren:
Pull Request Checklist
* [x] implement the feature
* [x] Fix for single RPM file 

Related to #4420.

PR #4549, opened 2026-02-04
Add integration test for NVR edge cases 

vaibhavdaren:
Add test to verify file artifact provider correctly handles NVR edge cases. 
This is continued from discussion on the PR for file provider:   https://github.com/teemtee/tmt/pull/4510#issuecomment-3848015990

Related: #4546

#### Why it falls back on Rawhide

The `tmt-artifact-shared` repository is created with **priority 50**. However, because packages are currently requested by **name** (e.g., `jq`) rather than by the exact **NVR** (Name-Version-Release), `dnf` has too much flexibility.

When the downloaded RPM (e.g., `jq-1.8.1-2.fc44.x86_64`) has dependencies that are incompatible with the Rawhide host, `dnf` determines it cannot install that specific file from the artifact repo. Because the request was simply for `jq`, it silently resolves the package from `updates` or `fedora` instead to satisfy the request.

#### Proposed Path Forward

To ensure the integrity of the test environment, we should implement the following:

1. **Strict NVR Enforcement:** Modify the installation logic to request the exact NVR. This forces `dnf` to either install the specific artifact or fail if dependencies aren't met.
2. **Post-Installation Verification:** After a package is installed, verify the source repository. If the package was pulled from a system repo instead of `tmt-artifact-shared` (or a repo defined by the provider), the process should raise an error with detailed logs.

#### Why it works with dependencies included

When we download using `--resolve --alldeps`, the artifact repository contains the full dependency chain. This allows `dnf` to satisfy the entire transaction locally from `tmt-artifact-shared` without needing to query default system repositories. You can see this passing in the test case:
[/tests/prepare/artifact/nvr-edge-cases/package-with-dependency](/tests/prepare/artifact/nvr-edge-cases/package-with-dependency)


tests/prepare/artifact/nvr-edge-cases/package-with-dependency/data/test.sh
```diff
@@ -0,0 +1,14 @@
...
+rpm -q jq
+
+# Verify jq came from the tmt-artifact-shared repository
+dnf info --installed jq | grep -Eq "From repo(sitory)?\s*:\s*tmt-artifact-shared"
+
+# Verify nano is installed (from directory with multiple RPMs)
+rpm -q nano
+
+# Verify nano came from the tmt-artifact-shared repository
+dnf info --installed nano | grep -Eq "From repo(sitory)?\s*:\s*tmt-artifact-shared"
```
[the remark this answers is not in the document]
vaibhavdaren: The test is supposed to fail in fedora-rawhide. Waiting for the failed run. So, I can get the required logs.  I will make the suggested changes post that.

vaibhavdaren:
**Note**🥇: Isolated the failing test ^_^ in [testing-farm:fedora-rawhide-x86_64:full](https://github.com/teemtee/tmt/runs/62868640050)    
https://artifacts.dev.testing-farm.io/f4e74ca8-7d1a-43f6-961c-e62e586f157b

LecrisUT:
Ok, did some local testing to see how this behaves, indeed the priority takes precedence over all this. @vaibhavdaren @AthreyVinay  Do you have a document to put the test cases and behavior?

vaibhavdaren:
> Ok, did some local testing to see how this behaves, indeed the priority takes precedence over all this. @vaibhavdaren @AthreyVinay Do you have a document to put the test cases and behavior?

We dont have a file, I think we can use https://github.com/teemtee/tmt/issues/4546 to document our findings.

vaibhavdaren:
When dependency are present in the shared artifact dir the tests pass on rawhide: https://artifacts.dev.testing-farm.io/361bd67e-9794-4863-8d91-f2152506bc83

LecrisUT:
> When dependency are present in the shared artifact dir the tests pass on rawhide: https://artifacts.dev.testing-farm.io/361bd67e-9794-4863-8d91-f2152506bc83

Can you elaborate? Basically the fact that `jq` built F43 failed is expected, but the fact that it fallbacked to system is unexpected and we **must** prevent that in the final design, i.e. the installation **must** fail with that exact packages requested. If you make it install `jq-1.8.1-2.fc44.x86_64` instead of `jq` it will fail as we expect it to. I have put some more details of observations in the google docs, but I will move it to #4546, let's move the discussion there.

vaibhavdaren:
### Analysis: Unexpected Fallback to System Repositories

It sounds like we’ve pinpointed the exact point of failure. `dnf` is doing what it’s designed to do : find *a* working solution when what we actually want is for it to **fail fast** if our specific artifact is incompatible.
We are on the same page: while the failure of the `jq` build on F43 is expected, the silent fallback to system repositories is a behavior we must prevent in the final design.

#### Why it falls back on Rawhide

The `tmt-artifact-shared` repository is created with **priority 50**. However, because packages are currently requested by **name** (e.g., `jq`) rather than by the exact **NVR** (Name-Version-Release), `dnf` has too much flexibility.

When the downloaded RPM (e.g., `jq-1.8.1-2.fc44.x86_64`) has dependencies that are incompatible with the Rawhide host, `dnf` determines it cannot install that specific file from the artifact repo. Because the request was simply for `jq`, it silently resolves the package from `updates` or `fedora` instead to satisfy the request.

#### Proposed Path Forward

To ensure the integrity of the test environment, we should implement the following:

1. **Strict NVR Enforcement:** Modify the installation logic to request the exact NVR. This forces `dnf` to either install the specific artifact or fail if dependencies aren't met.
2. **Post-Installation Verification:** After a package is installed, verify the source repository. If the package was pulled from a system repo instead of `tmt-artifact-shared` (or a repo defined by the provider), the process should raise an error with detailed logs.

#### Why it works with dependencies included

When we download using `--resolve --alldeps`, the artifact repository contains the full dependency chain. This allows `dnf` to satisfy the entire transaction locally from `tmt-artifact-shared` without needing to query default system repositories. You can see this passing in the test case:
`/tests/prepare/artifact/nvr-edge-cases/package-with-dependency`

LecrisUT:
> #### Why it falls back on Rawhide
> 
> The `tmt-artifact-shared` repository is created with **priority 50**. However, because packages are currently requested by **name** (e.g., `jq`) rather than by the exact **NVR** (Name-Version-Release), `dnf` has too much flexibility.

Note that dnf has a flag `--from-repo`, from my testing it worked as intended, but not very well with name-only, mostly with NVR. Don't remember if I tried it with the failing package as well, so try that.

> 1. **Strict NVR Enforcement:** Modify the installation logic to request the exact NVR. This forces `dnf` to either install the specific artifact or fail if dependencies aren't met.

Edge case to consider here is what if 2 repos provide the same NVR. My guess is that `--from-repo` would take care of this issue, but it's best if it is confirmed.

> Why it works with dependencies included

This should not be relevant, and should be avoided. The case here is a perfect example of why we should not be doing it as it breaks what is expected to have failed.

AthreyVinay:
>  My guess is that `--from-repo` would take care of this issue, but it's best if it is confirmed.

Would be interested to know the findings too.

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,73 @@
...
+REPO_LOW="$REPO_BASE/repo-low"
+BUILD_DIR="/tmp/rpm-build-nvr-test"
+
+dnf install -y rpm-build createrepo_c
+
+mkdir -p "$REPO_HIGH" "$REPO_LOW"
+mkdir -p "$BUILD_DIR"/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
+
+# Spec for v1.0 (goes into the high-priority repo)
+cat > "$BUILD_DIR/SPECS/v1.spec" << 'SPECEOF'
```
LecrisUT: Please make thesr concrete files

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,73 @@
...
+PACKAGE="dummy-nvr-test"
+REPO_BASE="/tmp/nvr-test"
+REPO_HIGH="$REPO_BASE/repo-high"
+REPO_LOW="$REPO_BASE/repo-low"
+BUILD_DIR="/tmp/rpm-build-nvr-test"
+
+dnf install -y rpm-build createrepo_c
+
+mkdir -p "$REPO_HIGH" "$REPO_LOW"
+mkdir -p "$BUILD_DIR"/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
```
LecrisUT: This folder structure is not necessary, the specs can be in arbitrary location

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,73 @@
...
+
+set -ex
+
+PACKAGE="dummy-nvr-test"
+REPO_BASE="/tmp/nvr-test"
+REPO_HIGH="$REPO_BASE/repo-high"
+REPO_LOW="$REPO_BASE/repo-low"
+BUILD_DIR="/tmp/rpm-build-nvr-test"
+
+dnf install -y rpm-build createrepo_c
```
LecrisUT: Building with mock is preferred. And splitting the srpm and binary rpm build would be useful since the srpm could be cached and reused across tested distros (and even across runs when that would be possible)

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,73 @@
...
+%files
+
+%changelog
+SPECEOF
+
+rpmbuild --define "_topdir $BUILD_DIR" -bb "$BUILD_DIR/SPECS/v1.spec"
+rpmbuild --define "_topdir $BUILD_DIR" -bb "$BUILD_DIR/SPECS/v2.spec"
+
+find "$BUILD_DIR/RPMS" -name "${PACKAGE}-1.0*.rpm" -exec cp {} "$REPO_HIGH/" \;
+find "$BUILD_DIR/RPMS" -name "${PACKAGE}-2.0*.rpm" -exec cp {} "$REPO_LOW/" \;
```
LecrisUT: The `find` is so unnecessary. Just put that glob directly in the `cp`

tests/prepare/artifact/nvr-debug/data/plan.fmf
```diff
@@ -0,0 +1,35 @@
...
+    discover:
+        how: fmf
+        filter: 'name: /tc12'
+    provision:
+        how: container
+    prepare:
+        - name: Build dummy RPMs and create local repos
+          how: shell
+          order: 40
+          script: ./build-repos.sh
```
LecrisUT: Why is this run inside the plan?

tests/prepare/artifact/nvr-debug/data/tc11-high.repo
```diff
@@ -0,0 +1,6 @@
+[test-high-priority]
+name=Test High Priority Repo (dummy-nvr-test v1.0)
+baseurl=file:///tmp/nvr-test/repo-high
+enabled=1
+gpgcheck=0
+priority=1
```
LecrisUT: Let's avoid extreme numbers. `50` and `30` seem fine mid-range values

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,63 @@
...
+                --provide repository-file:file://$(pwd)/tc11-low.repo" \
+            0 "TC 1.1: high-priority repo (v1.0) beats low-priority repo (v2.0)"
+    rlPhaseEnd
+
+    rlPhaseStartTest "TC 1.2 - Same priority falls back to NVR comparison"
+        rlRun "tmt run -i $run --scratch -vvv --all \
+            plan -n /tc12 \
+            provision -h $PROVISION_HOW --image $TEST_IMAGE_PREFIX/$image_name \
+            prepare --insert --how artifact \
+                --provide repository-file:file://$(pwd)/tc12-a.repo \
```
LecrisUT: The usage of artifact provide seems arbitrary here if you do all of the building in the nested plan. If you move the build up then at least you can reuse the artifacts.

tests/prepare/artifact/nvr-debug/data/build-repos.sh
vaibhavdaren: Done! I’ve optimized this by caching the SRPMs. They are now generated on the host logic inside `build-srpms/` and saved to /data. This ensures we're using a single, consistent source artifact for all downstream builds across different distributions.

tests/prepare/artifact/nvr-debug/rpm-data/build-srpms.sh
```diff
@@ -0,0 +1,24 @@
...
+mkdir -p "$SRPM_DIR"
+
+mock --buildsrpm --spec "$SCRIPT_DIR/v1.spec" --sources "$SCRIPT_DIR/" \
+    --resultdir "$SRPM_DIR" --enable-network
+mock --buildsrpm --spec "$SCRIPT_DIR/v2.spec" --sources "$SCRIPT_DIR/" \
+    --resultdir "$SRPM_DIR" --enable-network
+
+# This is done to speed up the tests
+echo "SRPMs staged in $SRPM_DIR — commit them with:"
+echo "  git add data/srpms/*.src.rpm && git commit"
```
LecrisUT: Put these in `nvr-debug/test.sh` and do not git commit these. We can figure out a better caching of these later (I will even start the first parts of that design now)

tests/prepare/artifact/nvr-debug/rpm-data/build-srpms.sh
```diff
@@ -0,0 +1,24 @@
...
+# Run this whenever the spec files change, then commit data/srpms/.
+#
+# Usage:
+#   ./build_srpms.sh
+#   git add data/srpms/*.src.rpm
+#   git commit -m "Update committed SRPMs"
+
+set -ex
+
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
```
LecrisUT: Unnecessarily complicated
```suggestion
SCRIPT_DIR="$(dirname $0)"
```
That's fine enough.

tests/prepare/artifact/nvr-debug/rpm-data/build-srpms.sh
```diff
@@ -0,0 +1,24 @@
...
+SRPM_DIR="$SCRIPT_DIR/../data/srpms"
+
+mkdir -p "$SRPM_DIR"
+
+mock --buildsrpm --spec "$SCRIPT_DIR/v1.spec" --sources "$SCRIPT_DIR/" \
+    --resultdir "$SRPM_DIR" --enable-network
+mock --buildsrpm --spec "$SCRIPT_DIR/v2.spec" --sources "$SCRIPT_DIR/" \
+    --resultdir "$SRPM_DIR" --enable-network
```
LecrisUT:
```suggestion
for spec in $SCRIPT_DIR/*.spec; do
  mock --buildsrpm --spec "$SCRIPT_DIR/v1.spec" --sources "$SCRIPT_DIR/" \
    --resultdir "$SRPM_DIR"
done
```
Don't see why we wouldn't build all srpms

tests/prepare/artifact/nvr-debug/data/plan.fmf
```diff
@@ -0,0 +1,11 @@
+summary: NVR priority and version selection tests
+discover:
+    how: fmf
+provision:
+    how: container
+prepare:
+    - name: Build dummy RPMs and create local repos
+      how: shell
+      script: ./build-repos.sh
```
LecrisUT:
```suggestion
    - name: Build dummy RPMs and create local repos
      order: 20
      how: shell
      script: ./build-repos.sh
```
Still not a fan of building these inside the inner tests, but at least run this before the artifacts

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,33 @@
...
+# by tmt as part of the normal repo sync.
+# Both test cases (tc11, tc12) share these repos within the same container.
+
+set -ex
+
+PACKAGE="dummy-nvr-test"
+BUILD_DIR="$(mktemp -d)"
+REPO_BASE="/tmp/nvr-test"
+
+dnf install -y rpm-build createrepo_c
```
LecrisUT: Move this in an prepare install step so that we can run these on the other distros also

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,33 @@
...
+
+mkdir -p "$REPO_BASE/tc11-high" "$REPO_BASE/tc11-low" \
+         "$REPO_BASE/tc12-a"   "$REPO_BASE/tc12-b"
+
+# Build binary RPMs from the host-built SRPMs for this container's distro.
+rpmbuild --define "_topdir $BUILD_DIR" --rebuild ./srpms/${PACKAGE}-1.0*.src.rpm
+rpmbuild --define "_topdir $BUILD_DIR" --rebuild ./srpms/${PACKAGE}-2.0*.src.rpm
```
LecrisUT: for loop build

tests/prepare/artifact/nvr-debug/data/plan.fmf
vaibhavdaren: As dicussed on slack moved the logic to runner.

vaibhavdaren:
From the hacking session : Added the first two test cases from the https://github.com/teemtee/tmt/issues/4546#issuecomment-3908515743
https://artifacts.dev.testing-farm.io/94c23875-cc56-4c77-879b-c2ab3f76f0ae/#work-the-restubje3ckx_plans-provision-container-the-rest_8_default-0. 
The tests should definitely pass now.

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,62 @@
...
+        done
+        PACKAGE="dummy-nvr-test"
+        BUILD_DIR="$(mktemp -d)"
+        for srpm in "$SRPM_DIR"/${PACKAGE}-*.src.rpm; do
+            rlRun "rpmbuild --define '_topdir $BUILD_DIR' --rebuild '$srpm'" 0 "Build binary RPM from $srpm"
+        done
+        rlRun "mkdir -p '$SCRIPT_DIR/data/rpms'"
+        rlRun "cp '$BUILD_DIR'/RPMS/noarch/${PACKAGE}-*.rpm '$SCRIPT_DIR/data/rpms/'"
+        rlRun "rm -rf '$BUILD_DIR'"
+        rlRun "pushd data"
```
LecrisUT: Instead of doing the build inside the `data`, make a `tmpdir` and copy the data and `pushd` into there. See for example
https://github.com/teemtee/tmt/blob/ebf609c6c38dace4130a97889956e9f6ba67313b/tests/libraries/local/test.sh#L13-L17

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,62 @@
...
+# (repository-file) installs the relevant .repo files before each TC runs.
+
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+. ../../../images.sh || exit 1
+. ../lib/common.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
+        SCRIPT_DIR="$(dirname $0)"
```
LecrisUT: `dirname` is unnecessary, we are always assuming this script is run in the cwd. See the `pushd data`

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,62 @@
...
+            rlRun "rpmbuild --define '_topdir $BUILD_DIR' --rebuild '$srpm'" 0 "Build binary RPM from $srpm"
+        done
+        rlRun "mkdir -p '$SCRIPT_DIR/data/rpms'"
+        rlRun "cp '$BUILD_DIR'/RPMS/noarch/${PACKAGE}-*.rpm '$SCRIPT_DIR/data/rpms/'"
+        rlRun "rm -rf '$BUILD_DIR'"
+        rlRun "pushd data"
+        rlRun "run=$(mktemp -d)" 0 "Create run directory"
+        DATA_DIR="$(pwd)"
+        setup_distro_environment
+    rlPhaseEnd
```
LecrisUT: Please split this in smaller phases. I would say setting up the `mktemp` and `setup_distro_environment` can live here for consistency, but the rpm building should be in a separate `rlPhaseStartTest`

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,62 @@
...
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
+        SCRIPT_DIR="$(dirname $0)"
+        SRPM_DIR="$SCRIPT_DIR/data/srpms"
+        rlRun "mkdir -p '$SRPM_DIR'"
+        for spec in "$SCRIPT_DIR"/rpm-data/*.spec; do
+            rlRun "sudo mock --buildsrpm --spec '$spec' --sources '$SCRIPT_DIR/rpm-data/' \
+                --resultdir '$SRPM_DIR' --enable-network" 0 "Build SRPM from $spec"
```
LecrisUT:
```suggestion
            rlRun "mock --buildsrpm --spec '$spec' --sources '$SCRIPT_DIR/rpm-data/' \
                --resultdir '$SRPM_DIR'" 0 "Build SRPM from $spec"
```
- Heck no `sudo`
- No `enable-network` unless you justify it

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,62 @@
...
+        done
+        PACKAGE="dummy-nvr-test"
+        BUILD_DIR="$(mktemp -d)"
+        for srpm in "$SRPM_DIR"/${PACKAGE}-*.src.rpm; do
+            rlRun "rpmbuild --define '_topdir $BUILD_DIR' --rebuild '$srpm'" 0 "Build binary RPM from $srpm"
```
LecrisUT: Can also use `mock` here. Not really an issue for this simple test repos though, both `rpmbuild` and `mock` are fine, just be consistent with it. The separate srpm and binary rpm should still be the case because it will make it easier to reuse and cache stuff.

tests/prepare/artifact/nvr-debug/data/tc11.fmf
```diff
@@ -0,0 +1,15 @@
...
+    Repos used:
+      - test-high-priority repo (priority=30) provides dummy-nvr-test-1.0
+      - test-low-priority repo  (priority=50) provides dummy-nvr-test-2.0
+
+    Expected result: dummy-nvr-test-1.0 is installed because the lower
+    numeric priority value (30) overrides the higher package version.
+
+test: ./tc11.sh
+require:
+  - dummy-nvr-test
```
LecrisUT: To not split too much the files (especially when there is relevant documentation associated with it), can you embed the test inside the plan file using `discover/shell` (should be able to add the `require` inside that also.

tests/prepare/artifact/nvr-debug/test.sh
```diff
@@ -0,0 +1,62 @@
...
+        setup_distro_environment
+    rlPhaseEnd
+
+    rlPhaseStartTest "TC 1.1 - Repository priority overrides package version"
+        rlRun "tmt run -i $run --scratch -vvv --all \
+            plan --name /plan-tc11 \
+            provision -h $PROVISION_HOW --image $TEST_IMAGE_PREFIX/$image_name \
+            prepare --insert --how artifact \
+                --provide 'repository-file:file://$DATA_DIR/tc11-high.repo' \
+                --provide 'repository-file:file://$DATA_DIR/tc11-low.repo'" \
```
LecrisUT: Can't these be embedded in the static fmf file in these cases (the `parsed.path` should account for what a relative path means (can take it relative to fmf tree root for now) but otherwise seems like it would work). Then you can just do a loop of the the whole `rlPahseStartTest` like in
https://github.com/teemtee/tmt/blob/ebf609c6c38dace4130a97889956e9f6ba67313b/tests/core/enabled/test.sh#L10-L21

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -ex
+
+REPO_BASE="/tmp/nvr-test"
+
+for spec in "$@"; do
+    package="${spec%%:*}"
+    repo="${spec#*:}"
+    mkdir -p "$REPO_BASE/$repo"
+    cp "./rpms/${package}"*.rpm "$REPO_BASE/$repo/"
```
LecrisUT: This will be problematic in the future when we test same NVR but different contents because the names will clash at build time. Let's make the assumption that all repo content (building and separating in folders) is defined by the layer above in their own folders (although at that point could also do `createrepo` there).

The build steps could be like
```
# Loop over the state of the repos
for repo in repo-1.0 repo-2.0 repo-1.0-broken; do
  # Loop over all spec files in the repos
  pushd $repo
  for spec_file  in *.spec; do
    # for now all of these can build and save the *.rpm in the same cwd
    build_srpm()
    build_rpm()
    create_repo()  # in-place
  done
  popd 
done
```
And inside this script it would just put everything under a predictable file location for the `.repo` files to be able to consume, not even bothering with filtering what is actually used.

tests/prepare/artifact/nvr-debug/data/plan-tc11.fmf
```diff
@@ -0,0 +1,42 @@
...
+            rpm -q dummy-nvr-test | grep -E "^dummy-nvr-test-1\.0-"
+            dnf repoquery --installed dummy-nvr-test --qf "%{from_repo}" | grep -Fx "test-high-priority"
+provision:
+    how: container
+prepare:
+    - name: Install createrepo_c
+      order: 10
+      how: install
+      package:
+        - createrepo_c
```
LecrisUT: This can be moved in `main.fmf`

vaibhavdaren: moved the common part to `main.fmf` addressed in https://github.com/teemtee/tmt/pull/4549/changes/9acd9300cb0af28da4410dede3c2c394270715e8.

LecrisUT: You forgot
```yaml
execute:
    how: tmt
```

vaibhavdaren: tq. fixed in https://github.com/teemtee/tmt/pull/4549/commits/1da478bbf381b882d374c11c9944a88f69ea4a00.

tests/prepare/artifact/nvr-debug/data/main.fmf
```diff
@@ -0,0 +1,11 @@
...
+    how: container
+
+execute:
+    how: tmt
+
+prepare:
+    - name: Install createrepo
+      how: install
+      package:
+        - createrepo
```
tcornell-bus: This prepare step gets overwritten by the prepare step in plan-tc11 and plan-tc12. I think it can be deleted entirely since the tests themselves do not use createrepo and they are currently passing anyway without this step.

vaibhavdaren: Nice find. Addressed in https://github.com/teemtee/tmt/pull/4549/commits/4179c37559e23a6cc0bfaa88f20f18dbbf21e9c4.

tests/prepare/artifact/nvr-debug/data/build-repos.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Build local RPM repos from per-repo spec directories.
+# For each subdirectory containing .spec files, builds binary RPMs
+# and runs createrepo_c in-place.
```
AthreyVinay: Possible documentation mis-match? `createrepo_c` vs calling `createrepo .`

thrix: yeah, the package is called `createrepo_c` but it provides `createrepo` executable. IN some older releases it was only `createrepo` package. I would just rename to `createrepo`.

tests/prepare/artifact/nvr-priority/main.fmf
```diff
@@ -0,0 +1,30 @@
...
+
+    TC 1.2 - Same Priority favors NVR:
+        When repository priorities are equal, DNF5 falls back to
+        standard RPM version comparison and installs the highest NVR.
+
+link:
+  - verifies: https://github.com/teemtee/tmt/issues/4546
+
+require:
+  - tmt
```
psss:
```suggestion
require+:
```

It's a bit better to extend the required packages. In this way we don't lose/override any useful requirements defined in the parent.

tmt/steps/prepare/artifact/providers/repository.py
```diff
@@ -81,8 +83,18 @@ def fetch_contents(
         # It returns an Empty list, as no individual artifact files are downloaded.
 
         self.logger.info(f"Initializing repository provider with URL: {self.id}")
-        # TODO: This should not be using Repository.from_url
-        self.repository = Repository.from_url(url=self.id, logger=self.logger)
+
+        parsed = urlparse(self.id)
+        if parsed.scheme == 'file':
+            # Read .repo file from the local (controller) filesystem.
+            self.logger.info(f"Reading repository file from local path: {parsed.path}")
```
thrix, line 90: @vaibhavdaren for debugging purposes, could we also print here the absolute path as a debug output pls?

tests/prepare/artifact/nvr-priority/data/build-repos.sh
```diff
@@ -0,0 +1,19 @@
...
+
+for repo_dir in */; do
+    pushd "$repo_dir"
+    build_dir=$(mktemp -d)
+    for spec in *.spec; do
+        rpmbuild --define "_topdir $build_dir" -bb "$spec"
+    done
+    find "$build_dir/RPMS" -name "*.rpm" -exec cp {} . \;
+    rm -rf "$build_dir"
+    createrepo .
```
AthreyVinay, line 17: Since you are iterating through all subdirectories which are are not repo directories (eg. `.fmf/`), there is a possibility that this line here will create `repodata/` directories in non repo directories too. Maybe an explicit check to only go over directories that contain `.spec` files?

tests/prepare/artifact/nvr-priority/main.fmf
thrix, line 1: Why we call the directory `nvr-debug` could we use some other, more descriptive name please?
Consider renaming to something more descriptive like nvr-priority or repository-priority, matching what the 
tests actually verify (DNF repository priority and NVR resolution behavior).

thrix (changes requested):

tests/prepare/artifact/nvr-priority/main.fmf
vaibhavdaren: I'll change it to `nvr-priority` seems nice. Addressed in f802a19b830c671b96fe584ce6d5b233010f07d4.

tests/prepare/artifact/nvr-priority/data/build-repos.sh
vaibhavdaren: Thanks for the review! The concern makes sense at first glance, but `*/` in bash                                                                does not match hidden directories (those starting with `.`) unless `shopt -s dotglob` is explicitly set. Since this script only sets `nullglob`,  `.fmf/` will never be matched by the loop.                                                                                                           
                                                                                                                                                       
  In the actual directory layout, `*/` will only ever match the intended repo directories: `tc11-high/`, `tc11-low/`, `tc12-a/`, `tc12-b/` — all of which contain `.spec` files by design, so no stray `repodata/` directories will be created.

tmt/steps/prepare/artifact/providers/repository.py
vaibhavdaren: @thrix : Addressed in f802a19b830c671b96fe584ce6d5b233010f07d4.

tests/prepare/artifact/nvr-priority/data/plan-tc12.fmf
```diff
@@ -0,0 +1,31 @@
...
+              - test-same-priority-b repo (priority=50) provides dummy-nvr-test-2.0
+
+            Expected result: dummy-nvr-test-2.0 is installed because both repos
+            have equal priority and DNF5 falls back to RPM version comparison.
+        require:
+          - dummy-nvr-test
+        test: |
+            set -ex
+            rpm -q dummy-nvr-test | grep -E "^dummy-nvr-test-2\.0-"
+            dnf repoquery --installed dummy-nvr-test --qf "%{from_repo}" | grep -Fx "test-same-priority-b"
```
psss: Bundling the full `tmt` test using the `discover.shell` approach sounds a little bit awkward. What about rather defining a genuine `tmt` test? Both plan and test related to the `tc12` test case could be still defined in the same `tc12.fmf` file:

```yaml
/plan:
    discover:
        how: fmf
    ...

/test:
    summary: TC 1.2 - Same priority falls back to NVR comparison
    description: |
        Verify that when repositories share the same priority, DNF5 selects
        the package with the highest NVR (Name-Version-Release).

        Repos used:
          - test-same-priority-a repo (priority=50) provides dummy-nvr-test-1.0
          - test-same-priority-b repo (priority=50) provides dummy-nvr-test-2.0

        Expected result: dummy-nvr-test-2.0 is installed because both repos
        have equal priority and DNF5 falls back to RPM version comparison.
    require:
      - dummy-nvr-test
    test: |
        set -ex
        rpm -q dummy-nvr-test | grep -E "^dummy-nvr-test-2\.0-"
        dnf repoquery --installed dummy-nvr-test --qf "%{from_repo}" | grep -Fx "test-same-priority-b"
```

psss (commented):
Looks good, just two rather minor suggestions.

tests/prepare/artifact/nvr-priority/data/plan-tc12.fmf
vaibhavdaren: Nice suggestion looks cleaner this way.  Addressed in e07519aced64d3a8b8f4c19274657b20b56caf58.

PR #4555, opened 2026-02-05
Add config and persona hint for Gemini Code Assist

psss:
Let's try if we can tune up Gemini's behaviour to our needs.
https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -2,5 +2,15 @@
 
 ## Review Persona & Tone
 - **Be Concise:** Provide feedback in the fewest words possible.
-- **No Fluff:** Do not include pleasantries, praise (e.g., "Great job"), or conversational filler.
-- **Direct Imperatives:** Use imperative mood (e.g., "Rename variable..." instead of "I suggest renaming...").
+- **No Fluff:** Do not include pleasantries, praise (e.g., "Great job"),
+  or conversational filler.
+- **Direct Imperatives:** Use imperative mood (e.g., "Rename
+  variable..." instead of "I suggest renaming...").
+
+## Avoid Repetitive Feedback
+- **Single Occurrence Only:** If a pattern (e.g., a specific linting
+  error or architectural suggestion) occurs multiple times across
+  different files, comment on it **only once**.
+- **Summary Level:** For project-wide issues found in particular PR,
+  mention them in the PR summary rather than leaving multiple identical
+  inline comments.


.gemini/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+memory_config:
+    disabled: false
+
+code_review:
+    disable: false
+    comment_severity_threshold: MEDIUM
+    max_review_comments: -1
+    pull_request_opened:
+        help: false
+        summary: true
```
LecrisUT: Is this the "Summary of Changes" thing that is posted the first time a PR is opened? Shall we disable it, since it seems it only regurgitates the content of the description, and best guesses of the changes. If anything we should be making sure that the contents of the description is up to date and describes the PR well enough rather than the comment it generates there.

psss: Makes sense, let's turn it off: 08b4710d

psss: Hmmm... thinking about it a bit more... actually, gemini is quite good at providing summaries and sometimes it describes the changes better than our commit descriptions. It can also serve as an inspiration for improving the pull request description. Now I'm not completely sure we should disable it...

LecrisUT: Afaict ([example](https://github.com/teemtee/tmt/pull/4566#issuecomment-3879342284)), it does not update the summary with each commit made, and if that is the case it would be a source of distraction when it comes to reviewing what is the overall PR state.

LecrisUT:
Anpther issue spotted. This one is so repetitive: https://github.com/teemtee/tmt/pull/4571#discussion_r2790617494. It is the same change repated, we know to update all ocurances

vaibhavdaren:
Nice catch @LecrisUT. 
> Anpther issue spotted. This one is so repetitive: [#4571 (comment)](https://github.com/teemtee/tmt/pull/4571#discussion_r2790617494). It is the same change repated, we know to update all ocurances

There is no `deduplicate-comments` key in the config,  the way to solve this would be through `.gemini/styleguide.md`

```md
## Avoid Repetitive Feedback
**Single Occurrence Only:** If a pattern (e.g., a specific linting error or architectural suggestion) occurs multiple times across different files, comment on it **only once**.
**Summary Level:** For project-wide issues found in this PR, mention them in the PR summary rather than leaving multiple identical inline comments.
```

mcasquer:
There's nothing we can add to the `ignore_patterns` ? What about `/packaging` or `/examples` ?

happz (approved):

.gemini/config.yaml
psss: Ok, let's keep it turned off for now.

psss:
> ```
> ## Avoid Repetitive Feedback
> **Single Occurrence Only:** If a pattern (e.g., a specific linting error or architectural suggestion) occurs multiple times across different files, comment on it **only once**.
> **Summary Level:** For project-wide issues found in this PR, mention them in the PR summary rather than leaving multiple identical inline comments.
> ```

Ok, let's try it. Added in 1283578a.

.gemini/config.yaml
```diff
@@ -0,0 +1,16 @@
+have_fun: false
+
+memory_config:
+    disabled: false
+
+code_review:
+    disable: false
```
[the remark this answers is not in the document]
psss: This one was copy-pasted from the docs example:

* https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github

Schema confirms they use both `disable` and `disabled`.

psss:
> There's nothing we can add to the `ignore_patterns` ? What about `/packaging` or `/examples` ?

I'd rather start without explicit exclusions and add them later if clear need pops up. Both packaging and examples can contain mistakes so might we worth review.

happz:
Configuration-only, no need for full test coverage.

PR #4563, opened 2026-02-10
Pass plan environment variables to container via --env-file (#4482)

### LINKED ISSUE #4482: Support passing environment variables to `podman run` in container provisioner
## Problem

Some container images require environment variables to be set during `podman run` (via `-e` flag) for proper initialization. Currently, tmt does not provide a way to pass environment variables directly to the `podman run` command when using the container provisioner.

This is different from the test environment variables that are available during test execution - these variables need to be present when the container is started.

## Use Case

When running tests against containers that require runtime configuration via environment variables (e.g., containers that perform initialization based on environment variables at startup), there's no way to pass these variables to the container.

## Proposed Solution

Add a new `run-environment` option to the container provisioner that allows specifying environment variables to be passed to `podman run`:

```yaml
provision:
    how: container
    image: my-image:latest
    run-environment:
        MY_VAR: "value"
        ANOTHER_VAR: "${TMT_PLAN_DATA}"
```

The variables should support injection of tmt environment variables (like `TMT_*` variables), allowing dynamic configuration based on the tmt run context.

These would translate to `-e MY_VAR=value -e ANOTHER_VAR=/path/to/data` flags in the `podman run` command.

## Additional Context

This feature would be useful for:
- Containers that require license keys or configuration tokens at startup
- Containers with conditional initialization logic based on environment
- Integration testing scenarios where container behavior needs to be controlled at runtime

thrix:
The container provision plugin now writes plan environment variables to a file and passes it to `podman run` via `--env-file`. This allows container images that require environment variables for initialization to receive them at container startup time.

**Important:** Environment variables containing newline characters are skipped with a warning, as podman's env-file format does not support multiline values (one line = one variable).

Fixes #4482

Assisted-by: Claude Code

## Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [x] include a release note


tmt/steps/provision/podman.py
```diff
@@ -212,6 +212,57 @@ def _setup_network(self) -> list[str]:
...
+
+        if not environment:
+            return []
+
+        # Filter out variables with newlines - podman's env-file format
+        # does not support multiline values (one line = one variable)
+        filtered_env: dict[str, str] = {}
+        for key, value in environment.items():
+            str_value = str(value)
+            if '\n' in str_value:
```
happz: * `filtered_env` should stick to being `tmt.utils.Environment`, as it is a collection of environment variables.
* `value` should be a subclass of `str`, it should be possible to use it everywhere `str` is used, therefore `str()` should not be needed.

I'd propose this then:

```diff
diff --git a/tmt/steps/provision/podman.py b/tmt/steps/provision/podman.py
index 44bcad07..9989d250 100644
--- a/tmt/steps/provision/podman.py
+++ b/tmt/steps/provision/podman.py
@@ -239,17 +239,16 @@ class GuestContainer(tmt.Guest):

         # Filter out variables with newlines - podman's env-file format
         # does not support multiline values (one line = one variable)
-        filtered_env: dict[str, str] = {}
+        filtered_env = tmt.utils.Environment()
         for key, value in environment.items():
-            str_value = str(value)
-            if '\n' in str_value:
+            if '\n' in value:
                 self.warn(
                     f"Environment variable '{key}' contains a newline character. "
                     "Podman's env-file format does not support multiline values, "
                     "skipping this variable."
                 )
             else:
-                filtered_env[key] = str_value
+                filtered_env[key] = value

         if not filtered_env:
             return []
```

thrix: thx

thrix: @happz addressed

tmt/steps/provision/podman.py
```diff
@@ -212,6 +212,56 @@ def _setup_network(self) -> list[str]:
...
+        environment = self.parent.plan.environment
+
+        if not environment:
+            return []
+
+        # Filter out variables with newlines - podman's env-file format
+        # does not support multiline values (one line = one variable)
+        filtered_env = tmt.utils.Environment()
+        for key, value in environment.items():
+            if '\n' in value:
```
AthreyVinay, line 244: A very naive question: what happens or is it even possible that there is a `=` in the key?

thrix: @AthreyVinay It is very stupid, it just takes all the things after the first `=`:

```
❯ cat envfile 
FILE=a=b=c

❯ podman run --env-file=envfile alpine printenv FILE
a=b=c
```

tests/provision/container/environment/data/main.fmf
```diff
@@ -0,0 +1,15 @@
...
+    discover:
+        how: fmf
+    provision:
+        how: container
+        image: fedora
+    execute:
+        how: tmt
+
+/test:
+    test: cat /etc/os-release
```
LecrisUT: This does not seem to be used. What about using `env` instead?

PR #4564, opened 2026-02-10
Detect and error on duplicate artifact NVRAs

AthreyVinay:
from sync...
**Specific example:** 
User provides two scratch builds of curl
Two koji tasks with the same nvr (content is different, e.g. two ways / patches how to fix the bug)
Raise an error (as we cannot decide/check what should be installed)

**NOTE**: This pr also renamed `raw_provider_id` -> `raw_id`.

- [x]  extend tests

### DIFF
--- a/tmt/steps/prepare/artifact/__init__.py
+++ b/tmt/steps/prepare/artifact/__init__.py
@@ -1,4 +1,4 @@
-from typing import Any, ClassVar, Optional
+from typing import ClassVar, Optional
 
 import tmt.base.core
 import tmt.steps
@@ -198,7 +198,6 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
         # Initialize all providers and have them contribute to the shared repo
         providers: list[ArtifactProvider] = []
         seen_nvras: dict[str, str] = {}
-        providers_data: list[dict[str, Any]] = []
 
         # --- Pass 1: Initialize all providers and validate for duplicate NVRAs ---
         for raw_id in self.data.provide:
@@ -213,13 +212,6 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
                 )
 
                 self._detect_duplicate_nvras(provider, seen_nvras)
-
-                providers_data.append(
-                    {
-                        'id': provider.raw_id,
-                        'artifacts': provider.artifact_metadata,
-                    }
-                )
 
                 providers.append(provider)
 
@@ -274,7 +266,7 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
             logger.debug(f"Installed repository '{repo.name}'.")
 
         # Persist artifact metadata to YAML
-        self._save_artifacts_metadata(providers_data)
+        self._save_artifacts_metadata(providers)
 
         # Report configuration summary
         logger.info(
@@ -306,16 +298,26 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
 
             seen_nvras[nvra] = raw_id
 
-    def _save_artifacts_metadata(self, providers_data: list[dict[str, Any]]) -> None:
+    def _save_artifacts_metadata(self, providers: list[ArtifactProvider]) -> None:
         """
         Persist the metadata of artifacts to a YAML file.
 
         Groups artifacts by provider.
         """
 
+        metadata = {
+            'providers': [
+                {
+                    'id': provider.raw_id,
+                    'artifacts': provider.artifact_metadata,
+                }
+                for provider in providers
+            ]
+        }
+
         metadata_file = self.plan_workdir / self.ARTIFACTS_METADATA_FILENAME
 
         try:
-            metadata_file.write_text(tmt.utils.to_yaml({'providers': providers_data}, start=True))
+            metadata_file.write_text(tmt.utils.to_yaml(metadata, start=True))
         except OSError as error:
             raise tmt.utils.FileError(f"Failed to write into '{metadata_file}' file.") from error
--- a/tests/unit/artifact/test_base.py
+++ b/tests/unit/artifact/test_base.py
@@ -57,15 +57,7 @@ def test_persist_artifact_metadata(tmp_path, mock_provider):
     prepare.ARTIFACTS_METADATA_FILENAME = 'artifacts.yaml'
 
     PrepareArtifact._detect_duplicate_nvras(prepare, mock_provider, {})
-
-    providers_data = [
-        {
-            "id": mock_provider.raw_id,
-            "artifacts": mock_provider.artifact_metadata,
-        }
-    ]
-
-    PrepareArtifact._save_artifacts_metadata(prepare, providers_data)
+    PrepareArtifact._save_artifacts_metadata(prepare, [mock_provider])
 
     # Verify YAML
     yaml_file = tmp_path / "artifacts.yaml"


tcornell-bus:
Hi @AthreyVinay looks like there are some conflicts. Could you take a look please?

AthreyVinay:
> Could you take a look please?

Thanks @tcornell-bus , done.

AthreyVinay:
A question to the reviewers:
Should duplicate detection happen before any downloads (and leave a clean state if error-ed i.e. do not leave partial artifactson disk), or whether the current behavior (interleaved with downloads) is acceptable?

Thanks.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -267,19 +278,32 @@ def essential_requires(self) -> list[tmt.base.core.Dependency]:
             tmt.base.core.DependencySimple('/usr/bin/createrepo'),
         ]
 
-    def _save_artifacts_metadata(self, providers: list[ArtifactProvider]) -> None:
+    def _collect_artifacts_metadata(
```
tcornell-bus: The name of this function is a bit misleading. I couldn't quickly tell if this was where the duplicates were checked for. Maybe the `_detect_nvra_duplicates` part should be its own function?

tcornell-bus:
> Should duplicate detection happen before any downloads (and leave a clean state if error-ed i.e. do not leave partial > artifactson disk), or whether the current behavior (interleaved with downloads) is acceptable?

Checking before any downloading makes the most sense to me. I think the user would prefer it to fail quickly, rather than halfway down the download list. It requires two loops, but things would be cleaner and clearer, in my opinion.

AthreyVinay:
> I think the user would prefer it to fail quickly, rather than halfway down the download list.

Thank you @tcornell-bus, makes sense to me too.
Addressed here: [Move duplicate NVRA detection into a separate pass that runs...](https://github.com/teemtee/tmt/pull/4564/commits/5feea223e352b84c2bd5853924fb240c4bbab692)

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Addressed here: [d00f947](https://github.com/teemtee/tmt/pull/4564/commits/d00f94779769707e2a87c3c681678a368b5f0ae9)

tcornell-bus:
Should this also handle when there are two different repository-files that provide the same package NVRA? Or is this only to detect when `koji:<task-id-1>` and `koji:<task-id-2>` point to the same NVRA?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -209,8 +213,29 @@ def go(
...
+                self._detect_duplicate_nvras(provider, seen_nvras)
+
+                providers_data.append(
+                    {
+                        'id': provider.raw_provider_id,
+                        'artifacts': provider.artifact_metadata,
+                    }
+                )
+
+                providers.append((provider, provider_id_sanitized))
```
LecrisUT: Would rather make the provider own its `sanitized_id` than use a tuple

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -267,19 +292,29 @@ def essential_requires(self) -> list[tmt.base.core.Dependency]:
...
-    def _save_artifacts_metadata(self, providers: list[ArtifactProvider]) -> None:
+    def _detect_duplicate_nvras(
+        self, provider: ArtifactProvider, seen_nvras: dict[str, str]
+    ) -> None:
+        """
+        Check for duplicate NVRAs across providers.
+        """
+        raw_provider_id = provider.raw_provider_id
+
+        for artifact_dict in provider.artifact_metadata:
```
happz: Please, no type-like suffixes. The name should reflect what the value is, not what its type is. `artifact_info` would work nicely.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -267,19 +292,29 @@ def essential_requires(self) -> list[tmt.base.core.Dependency]:
             tmt.base.core.DependencySimple('/usr/bin/createrepo'),
         ]
 
-    def _save_artifacts_metadata(self, providers: list[ArtifactProvider]) -> None:
+    def _detect_duplicate_nvras(
+        self, provider: ArtifactProvider, seen_nvras: dict[str, str]
+    ) -> None:
```
LecrisUT, line 287: Depending on how the installation goes, we might want to extend this check for names (and `Provides`?). Fine with keeping it only on NVRs right now until we figure out what conflicts types we actually want to error on.

happz:
> Should this also handle when there are two different repository-files that provide the same package NVRA? Or is this only to detect when `koji:<task-id-1>` and `koji:<task-id-2>` point to the same NVRA?

As of now, I'd focus on non-repository providers like koji and brew. Providers contributing their payload to the shared repository. Repositories brought by repository-like providers may be disabled except in one particular test, or mapped into VMs, plenty of ways for a potential NVR conflict to never materialize; on the other hand, package-like providers contributing same NVR to the same repository is broken already.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -267,19 +292,29 @@ def essential_requires(self) -> list[tmt.base.core.Dependency]:
...
 
         Groups artifacts by provider.
         """
-        providers_data = [
-            {
-                'id': provider.raw_provider_id,
-                'artifacts': provider.artifact_metadata,
-            }
-            for provider in providers
-        ]
```
LecrisUT, line 282: Do we need this change?

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Nice, addressed here: [Rename raw_provider_id to raw_id and add sanitized_id attribute](https://github.com/teemtee/tmt/pull/4564/commits/1aa09da455963d36046f21419cdafbbd15945440)

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Still relevant after the latest commit?

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Apologies, went in as part of this: [1aa09da](https://github.com/teemtee/tmt/pull/4564/commits/1aa09da455963d36046f21419cdafbbd15945440)

but should be addreesed, ps: https://github.com/teemtee/tmt/blob/1aa09da455963d36046f21419cdafbbd15945440/tmt/steps/prepare/artifact/__init__.py#L301

AthreyVinay:
> As of now, I'd focus on non-repository providers like koji and brew. Providers contributing their payload to the shared repository

Maybe an explicit check before we detect duplcates? .... something along the lines of:
``` python
if provider.contribute_to_shared_repo();
    self._detect_duplicate_nvras()
```

tmt/steps/prepare/artifact/__init__.py
LecrisUT: I don't see it being relevant. You do not use `providers_data` anywhere outside of this function.

LecrisUT:
> > As of now, I'd focus on non-repository providers like koji and brew. Providers contributing their payload to the shared repository
> 
> Maybe an explicit check before we detect duplcates? .... something along the lines of:
> 
> ```python
> if provider.contribute_to_shared_repo();
>     self._detect_duplicate_nvras()
> ```

Fine either way. If there are no `.artifacts` provided then we do not have anything to use the duplicate nvr check for. Although the plain name of `.artifacts` for that makes things increasingly confusing, since we are converging to using that property to mean `installable_artifacts` more explicitly (we do not query or populate that for repo file, and I don't think we should start to do either)

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Maybe there's a confusion - that's probably the old code that you are still seeing.

Please see the refactor:
https://github.com/teemtee/tmt/blob/1aa09da455963d36046f21419cdafbbd15945440/tmt/steps/prepare/artifact/__init__.py#L217-L222

and its usage here:
https://github.com/teemtee/tmt/blob/1aa09da455963d36046f21419cdafbbd15945440/tmt/steps/prepare/artifact/__init__.py#L277

LecrisUT: > Maybe there's a confusion - that's probably the old code that you are still seeing.
> 
> Please see the refactor:
> 
> https://github.com/teemtee/tmt/blob/1aa09da455963d36046f21419cdafbbd15945440/tmt/steps/prepare/artifact/__init__.py#L217-L222
> 
> and its usage here:
> 
> https://github.com/teemtee/tmt/blob/1aa09da455963d36046f21419cdafbbd15945440/tmt/steps/prepare/artifact/__init__.py#L277

Yes, but isn't that (`_save_artifacts_metadata`) *the only* place where it is used, making this change unnecessary?

AthreyVinay:
> Although the plain name of .artifacts for that makes things increasingly confusing, since we are converging to using that property to mean installable_artifacts more explicitly

Agree - there was some discussion around it in a pr here: https://github.com/teemtee/tmt/pull/4526#discussion_r2788596594

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Aha - I think I now got what you mean / expect. Maybe just pass `providers` to `_save_artifacts_metadata` and derive metadata at save time. Makes sense.

AthreyVinay: Addressed here: [40e7121](https://github.com/teemtee/tmt/pull/4564/commits/40e7121565d306f5eabf226d5cb6fb438e438fc9)

Also, now that I remember the reason it was done that way was to not iterate over the providers for another pass.

PR #4566, opened 2026-02-10
Enable the `ruff` check for unused imports

psss:
Would be nice to prevent importing irrelevant stuff.


tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -18,7 +18,9 @@
 from tmt.container import container, simple_field
 from tmt.plugins import PluginRegistry
 from tmt.steps.provision import Guest
-from tmt.utils import GeneralError, Path, ShellScript, retry
+from tmt.utils import GeneralError, Path
+from tmt.utils import ShellScript as ShellScript
+from tmt.utils import retry as retry
```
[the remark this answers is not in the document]
happz: +1 from me, it is absolutely obscure :/ I'd prefer we'd drop unused imports like this, as reimports seem very unlikely. `tmt.steps.prepare.artifact.providers.*` can and should import `tmt.utils.retry`. And so on.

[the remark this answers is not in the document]
psss: Weird, why would `ruff` want to do something like this??? Interesting...

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -18,9 +18,7 @@
 from tmt.container import container, simple_field
 from tmt.plugins import PluginRegistry
 from tmt.steps.provision import Guest
-from tmt.utils import GeneralError, Path
-from tmt.utils import ShellScript as ShellScript
-from tmt.utils import retry as retry
+from tmt.utils import GeneralError, Path, ShellScript, retry  # noqa: F401
```
LecrisUT: Huh, what was going on here?

tmt/templates/__init__.py
```diff
@@ -1,7 +1,7 @@
 import functools
 from typing import Any, Optional, Union
 
-import tmt
+import tmt as tmt
```
LecrisUT: Wanted to look into why it was recommending this. My only assumption is that it is

> If an import statement is used to re-export a symbol as part of a module's public interface, consider using a "redundant" import alias, which instructs Ruff (and other tools) to respect the re-export, and avoid marking it as unused, as in:
>
> ```python
> from module import member as member
> ```

But that doesn't seem right in these cases.

Incidentally it does suggest a workaround of

> Alternatively, you can use __all__ to declare a symbol as part of the module's interface, as in:
> 
> ```python
> # __init__.py
> import some_module
>
> __all__ = ["some_module"]
> ```

LecrisUT: Although would like it if we used
```python
from tmt import utils
```
as ways to make the namespaces a bit shorter when we otherwise use `tmt.utils.GeneralError` -> `utils.GeneralError`

happz: I'd prefer dropping those unused imports, even if they serve as "reimports", which does not seem to be very common in our code. Shortened version - `utils.GeneralError` - does not look appealing to me, it seems like it can be very puzzling among the rest `tmt.foo` imports.

`GeneralError` itself would be acceptable without the namespace entirely, like other common and frequently used names - `Command`, `Plan`, etc.

LecrisUT: > I'd prefer dropping those unused imports, even if they serve as "reimports", which does not seem to be very common in our code.

Yes, and the key issue is to find why did ruff consider this as a re-import. If we are genuinely re-importing (which I doubt because it would mean we are calling `import tmt.utils.tmt`, but there are genuine leftover cases like `from tmt.utils import Path`) then I think patching the usage would make it work.

> Shortened version - `utils.GeneralError` - does not look appealing to me, it seems like it can be very puzzling among the rest `tmt.foo` imports.
> 
> `GeneralError` itself would be acceptable without the namespace entirely, like other common and frequently used names - `Command`, `Plan`, etc.

We will easily get too swamped with all of the `from ... import`. I would rather call `from tmt.utils import error` and `error.GeneralError` wherever we have clearly defined namespaced packages where we can do that. `base.Command`, `base.Plan` do not feel like a bad compromise.

tmt/steps/prepare/artifact/providers/__init__.py
psss: Just unused imports. I've removed them completely.

tmt/templates/__init__.py
psss: Yeah, I'd agree here with completely removing the unused imports. Using the short version of `GeneralError` and friends sounds also a bit better to me then introducing the prefixes.

tmt/steps/provision/__init__.py
```diff
@@ -1,94 +1,45 @@
...
+from tmt.utils import Path
+from tmt.utils.hints import get_hint as get_hint
+from tmt.utils.wait import Deadline as Deadline
+from tmt.utils.wait import Waiting as Waiting
 
 if TYPE_CHECKING:
     import tmt.base
     import tmt.cli
-    from tmt._compat.typing import TypeAlias
+    from tmt._compat.typing import TypeAlias as TypeAlias
```
LecrisUT: Some more re-exports here.

tmt/lint.py
```diff
@@ -82,7 +82,7 @@ def lint_manual_valid_markdown(self) -> LinterReturn:
 from tmt.container import container
 
 if TYPE_CHECKING:
-    import tmt.base
+    pass
```
LecrisUT: Can drop

tmt/steps/scripts.py
```diff
@@ -3,20 +3,16 @@
...
-import tmt.utils.wait
 from tmt.container import container
 from tmt.utils import (
     Path,
 )
 from tmt.utils.templates import render_template_file
 
 if TYPE_CHECKING:
-    import tmt.steps.provision
+    pass
```
LecrisUT: Can drop

tmt/steps/provision/__init__.py
psss: Should be covered by b9827871.

tmt/lint.py
psss: Should be covered by b9827871.

tmt/steps/scripts.py
psss: Should be covered by b9827871.

PR #4572, opened 2026-02-11
Require `setuptools` until `beaker-client` drops `pkg_resources`

psss:
In the newest `setutools` the `pkg_resources` has been deprecated:

    UserWarning: pkg_resources is deprecated as an API. See
    https://setuptools.pypa.io/en/latest/pkg_resources.html. The
    pkg_resources package is slated for removal as early as 2025-11-30.
    Refrain from using this package or pin to Setuptools<81.

Let's temporarily require specific version of `setuptools` until `beaker-client` fixes the problem:

 * https://github.com/beaker-project/beaker/issues/232

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,9 +48,6 @@ dependencies = [
     "pygments>=2.7.4",
     "requests>=2.25.1",
     "ruamel.yaml>=0.16.6",
-    # Needed because beaker-client uses pkg_resources which are deprecated
-    # https://github.com/beaker-project/beaker/issues/232
-    "setuptools<81",
     "urllib3>=1.26.5, <3.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -48,6 +48,9 @@ dependencies = [
     "pygments>=2.7.4",
     "requests>=2.25.1",
     "ruamel.yaml>=0.16.6",
+    # Needed because beaker-client uses pkg_resources which are deprecated
+    # https://github.com/beaker-project/beaker/issues/232
+    "setuptools<81",
```
LecrisUT: Move to `provision-beaker`?

psss: Ah, right: 124641eb

psss:
All green! Finally! \o/

Issue #4570, opened 2026-02-10
Consider `executor`/`execution_driver` abstraction for guest execute


thrix:
## Proposal

Consider implementing an `executor`/`execution_driver` abstraction that abstracts the usage of `guest.execute` more explicitly.

This can then be specialized granularly for each SSH execution types to support bootc or not, and inside the bootc a similar split for how to create the bootc image and how to provide it to each `virtual`, `beaker`, etc. provision types.

The idea is that while mixins would all be used together (as `artemis`, `beaker`, `connect`, `minute`, and `virtual` guests are SSH-backed and would include both SSH and bootc mixins), they would still have distinct implementations for each one.

## Original discussion

https://github.com/teemtee/tmt/pull/4525#issuecomment-3847809856

LecrisUT:
My vision on this is to move all of the logic we have in the `guest.execute` and similar under its own class that can change for each guest based on the facts gathered (similar to how the `package_manager` changes based on those). We can then also have it own adjacent stuff like `StopWatch`. A quick sketch

```python
class Guest:
    executor: Executor

    def _post_init():
        # This is used in the fact gathering
        self.executor = self.initial_executor(self)

    def _post_gather_facts():
        # From here on this one is used and would be aware of gathering commands etc.
        self.executor = self.executor_from_facts(self, self.facts)

    def execute(...):
        return self.executor.execute(...)
```

psss:
Just starting the discussion on this...

PR #4574, opened 2026-02-11
Draft: Add executor abstraction for provision step

thrix:
## Summary

Relates to #4570

This implements an executor/execution_driver abstraction that separates command execution logic from guest connectivity, enabling cleaner bootc/image-mode support across different provision types.

### Key changes:
- Add `ExecutionDriver` and `DeferrableExecutor` base classes in `tmt/steps/provision/executor/`
- Implement `SSHExecutor`, `LocalExecutor`, `ContainerExecutor`, `MockExecutor`
- Add `BootcContainerfileExecutor` and `DeferrableSSHExecutor` for bootc support
- Integrate executors with Guest classes via `_create_executor()` method
- Remove `CommandCollector` mixin (replaced by executor pattern)

### Benefits:
- **Separation of concerns**: Transport (SSH/local/container) separate from execution mode
- **Composability**: Easy to add new execution strategies
- **Extensibility**: Bootc support can be added to other guest types
- **Backward compatibility**: External `Guest.execute()` API unchanged
- **Testability**: Executors can be unit tested independently

### Architecture

```
ExecutionDriver (ABC)                    # Base abstraction
    |
    +-- SSHExecutor                      # Execute via SSH
    +-- LocalExecutor                    # Execute locally
    +-- ContainerExecutor                # Execute via podman exec
    +-- MockExecutor                     # Execute via mock shell

DeferrableExecutor(ExecutionDriver)      # Commands can be batched
    |
    +-- BootcContainerfileExecutor       # Collect into Containerfile, build, reboot

DeferrableSSHExecutor(DeferrableExecutor)  # Composes SSH + Bootc deferral
```

## Test plan

- [x] Run `tests/prepare/bootc/test.sh` - all 6 assertions pass
- [ ] Run full test suite
- [ ] Type checking with mypy

🤖 Generated with [Claude Code](https://claude.ai/code)

Assisted-by: Claude Code


tmt/steps/provision/__init__.py
```diff
@@ -1673,10 +1634,33 @@ def __init__(
         """
 
         self.guest_logs = []
+        self._executor: Optional['ExecutionDriver'] = None
```
LecrisUT, line 1637: I have a different view here. Would prefer for this executor to be swapped in and out, specifically, use a default executor initially -> gather the facts -> conditionally replace the executor with the one to be used beyond the `provision` step.

tmt/steps/provision/__init__.py
LecrisUT, line 1060: Unrelated changes, muddies the picture.

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+class ExecutionDriver(abc.ABC):
+    """
+    Base abstraction for command execution on a guest.
+
+    Executors encapsulate HOW commands are executed, separated from
+    the Guest which represents WHAT system is being targeted.
+
+    This follows the Strategy pattern - different executors can be
+    swapped to change execution behavior without modifying the Guest.
+    """
```
LecrisUT, line 37: The base classes should be outside of the `__init__` to avoid circular dependency issues

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+        self._logger.info(message, *args, **kwargs)
+
+    @abc.abstractmethod
+    def execute(
+        self,
+        command: Union[Command, ShellScript],
+        *,
+        cwd: Optional[Path] = None,
+        env: Optional[Environment] = None,
+        friendly_command: Optional[str] = None,
```
LecrisUT, line 69: Stuff like `friendly_command` need not be inside the `execute`, it should be handled outside of it. This driver should be minimal in its intent, just execute a command or shellscript.

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        pass
+
+    def cleanup(self) -> None:
+        """
+        Cleanup executor resources.
+
+        Optional hook for executors that need cleanup after use.
+        """
+        pass
```
LecrisUT, line 123: If it is not yet used, do not define. Adding them is trivial when needed

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+        logger: tmt.log.Logger,
+    ) -> None:
+        """
+        Initialize the executor.
+
+        :param guest: the guest this executor operates on.
+        :param logger: logger to use for logging.
+        """
+        self.guest = guest
+        self._logger = logger
```
LecrisUT, line 52: We had a discussion on the issues with logger. One issue that was pointed is that there are sections of the code where we modify the logger which propagates everywhere due to the assignment. We should stop assigning it and always do a `descend` to avoid these issues.

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+
+    @abc.abstractmethod
+    def defer(
+        self,
+        command: Union[Command, ShellScript],
+        *,
+        cwd: Optional[Path] = None,
+        env: Optional[Environment] = None,
+        sourced_files: Optional[list[Path]] = None,
+    ) -> None:
```
LecrisUT, line 146: Having special interfaces for the driver goes against the point of having the generic driver. Any interface here must be connected to the ones defined in the main `ExecutionDriver` and nowhere in the code should there be an `isinstance(..., DeferrableExecutor)` or equivalent.

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+
+        :param step: the step that has completed.
+        """
+        import tmt.steps
+
+        if self.has_pending_commands:
+            self.info("Flushing collected commands")
+            self.flush()
+        else:
+            self.debug("No deferred commands to flush.")
```
LecrisUT, line 195: This is unnecessary, just have `flush` be the command without logging and make it a generic to the base one, which is specialized in this. The abstraction should be on `execute_deferred_commands`

tmt/steps/provision/executor/__init__.py
```diff
@@ -0,0 +1,195 @@
...
+        test_session: bool = False,
+        tty: bool = False,
+        silent: bool = False,
+        log: Optional[tmt.log.LoggingFunction] = None,
+        interactive: bool = False,
+        on_process_start: Optional[OnProcessStartCallback] = None,
+        on_process_end: Optional[OnProcessEndCallback] = None,
+        sourced_files: Optional[list[Path]] = None,
+        **kwargs: Any,
+    ) -> CommandOutput:
```
LecrisUT, line 79: `Optional[CommandOutput]`.

And the main thing that we can do with this driver is to move some of the specialization that we have encountered, like with the `StopWatch` in `prepare/shell`.

Issue #4529, opened 2026-01-28
Move `Guest` classes out of `tmt.steps.provision`


psss:
Introduce a new module `tmt.guest` and store all `Guest` classes there.

psss:
This was proposed for the housekeeping sprint 10.

psss:
One more pull request will be needed for downstream changes.

PR #4583, opened 2026-02-16
Move Guest base classes into tmt.guest module

happz:
Basically what's not about the `provision` step itself is moved out of `tmt.steps.provision`, and since this part's about guests, `tmt.guest` is born. Guests are objects multiple steps work with, this makes them less depending on `provision` and more open to nicer things in the future, like nicer imports and type annotations.

Implements #4529.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -620,11 +620,11 @@ class GuestFacts(SerializableContainer):
     #: commands executed on this guest.
     capabilities: dict[GuestCapability, bool] = field(
         default_factory=cast(Callable[[], dict[GuestCapability, bool]], dict),
-        serialize=lambda capabilities: {
-            capability.value: enabled for capability, enabled in capabilities.items()
-        }
-        if capabilities
-        else {},
+        serialize=lambda capabilities: (
+            {capability.value: enabled for capability, enabled in capabilities.items()}
+            if capabilities
+            else {}
+        ),
         unserialize=lambda raw_value: {
             GuestCapability(raw_capability): enabled
             for raw_capability, enabled in raw_value.items()
@@ -1067,7 +1067,7 @@ class GuestFacts(SerializableContainer):
         def _value(field: str, label: str) -> tuple[str, str, str]:
             v = getattr(self, field)
 
-            return field, label, v if v else 'unknown'
+            return field, label, v or 'unknown'
 
         def _flag(field: str, label: str) -> tuple[str, str, str]:
             v = getattr(self, field)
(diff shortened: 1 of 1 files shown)


tcornell-bus:
Cursor found a Doc reference that is due for an update:
> File: docs/code/classes.rst line 287
> Current: # various different classes (derived from tmt.steps.provision.Guest),
> Should be: # various different classes (derived from tmt.guest.Guest),

tmt/steps/provision/artemis.py
```diff
@@ -4,6 +4,7 @@
 import requests
 
 import tmt
+import tmt.guest
 import tmt.hardware
 import tmt.log
```
AthreyVinay, line 9: Some stale / unused imports. Probablyl be fixed by #4566

happz: Correct, there will be some, and I plan to let them rot till #4566 lands and resolves them.

tcornell-bus:
Hi @happz are the failing tests unrelated to this PR?

happz:
> Hi @happz are the failing tests unrelated to this PR?

Very, very likely related to my changes.

LecrisUT:
LGTM overall, but what are the plans for further splitting of this. There seems to be quite a lot of utilities defined in there, and at the very least would be better to have a more dedicated `guest.py` and `guest_ssh.py` submodules. Could you make it a package in this PR to make that transition a bit easier?

happz:
> LGTM overall, but what are the plans for further splitting of this.

Mine, or general? Mine, at this moment, are minimal: my goal was to cut out `Guest` base out of `tmt.steps.provision`, to make the latter smaller & to untangle some import conundrums I observed. I can imagine e.g. (re)boot-related code being placed in its own submodule, same for facts, then maybe traits extracted from `Guest` and `GuestSsh`.

> There seems to be quite a lot of utilities defined in there, and at the very least would be better to have a more dedicated `guest.py` and `guest_ssh.py` submodules.

Yeah, and of that I am not sure, that `guest_ssh.py`. I can imagine `tmt/guest/traits/{bootc,ssh}.py`, for example. That's another reason why I constrained myself to just moving things from place A to place B, as I believe we don't have a clear agreement on how to refactor base guest classes. So, yes, something in the future, sure, but I would prefer not to partition the code anymore at this point.

> Could you make it a package in this PR to make that transition a bit easier?

Sure, turned into a subpackage in 89f91f2f. For reasons mentioned above, I'd prefer to not cut it into smaller pieces till we plan how to cut the base guest classes.

psss:
Testing priority sync.

Issue #4544, opened 2026-02-03
Let's try using python instead of beakerlib for testing `tmt`


psss:
Convert a single test, see how it’s going, then agree on the next steps
@happz has some initial experiments in this stash. We could start from there.

PR #4584, opened 2026-02-16
Experiment with tmt tests in Python & Pytest

happz:
Kicking off by rewriting a simpler beakerlib test into pytest lingo.

Brings also smaller changes in unrelated tests to use the same CLI runner where possible, to prevent different behavior.

Implements #4544.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/integration/test_nitrate.py
+++ b/tests/integration/test_nitrate.py
@@ -1,7 +1,6 @@
 import os
 import shutil
 import tempfile
-from typing import TYPE_CHECKING
 
 import nitrate
 import pytest
@@ -12,9 +11,6 @@ from ruamel.yaml import YAML
 import tmt.log
 from tests import CliRunner
 from tmt.utils import ConvertError, Path
-
-if TYPE_CHECKING:
-    pass
 
 # Prepare path to examples
 TEST_DIR = Path(__file__).parent
--- a/tests/integration/test_polarion.py
+++ b/tests/integration/test_polarion.py
@@ -1,5 +1,4 @@
 import os
-from typing import TYPE_CHECKING
 
 import pytest
 from fmf import Tree
@@ -8,9 +7,6 @@ from tests import CliRunner
 from tmt.identifier import ID_KEY
 
 from .test_nitrate import TEST_DIR, Base
-
-if TYPE_CHECKING:
-    pass
 
 
 @pytest.mark.skip(reason="Only works locally for now")


tests/unit/test_cli.py
```diff
@@ -93,42 +95,37 @@ def test_create():
         'story create -t full test',
     ]
     for command in commands:
-        result = runner.invoke(tmt.cli._root.main, command.split())
+        result = run_tmt(command.split())
```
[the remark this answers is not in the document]
happz: Addressed in 1d6f78d5

tests/__init__.py
```diff
@@ -26,25 +29,56 @@ def reset_common() -> None:
...
-    def invoke(
+    def __init__(
+        self,
+        charset: str = "utf-8",
+        env: Optional[Mapping[str, Optional[str]]] = None,
+        echo_stdin: bool = False,
+        catch_exceptions: bool = True,
+        mix_stderr: bool = False,
+    ) -> None:
+        super().__init__(charset=charset, env=env, echo_stdin=echo_stdin)
```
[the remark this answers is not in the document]
happz: Addressed in 05ddcea9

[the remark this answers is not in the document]
happz: Addressed in 05ddcea9

plans/main.fmf
```diff
@@ -10,11 +10,15 @@ prepare+:
...
       - python3-pip
+      - python3-pytest
 
 # Use the internal executor
 execute:
     how: tmt
 
+environment:
+  PYTEST: python3 -m pytest -vvv -ra --showlocals
+
```
LecrisUT, line 35: Do we want to be extra verbose? It would make navigating failure quite cumbersome until a proper framework can be provided. Although, it seems the verbosity only triggers at test failure?

tests/core/about/main.fmf
```diff
@@ -2,3 +2,6 @@ summary: Verify some basic output of `tmt about`
 description: |
     Make sure `tmt about` reports some basic info about tmt: plugins
     for now, more info in the future.
+
+framework: shell
+test: $PYTEST
```
LecrisUT: Could add more explicitly `./` to show what is being tested

tests/__init__.py
```diff
@@ -26,29 +29,58 @@ def reset_common() -> None:
...
+    def __call__(
+        self,
+        *args: str,
+        command: Optional[click.BaseCommand] = None,
+        input: Optional[Union[str, bytes, IO[Any]]] = None,
+        env: Optional[Mapping[str, Optional[str]]] = None,
+        color: bool = False,
+        **kwargs: Any,
+    ) -> click.testing.Result:
+        pass
```
LecrisUT, line 50: Is this needed? I thought that `mypy` can take the type annotation from `conftest.py` and the fixture (probably requiring quotations in the tests since we do not do `future.annotations`).

tests/__init__.py
```diff
@@ -26,29 +29,58 @@ def reset_common() -> None:
...
 class CliRunner(click.testing.CliRunner):
-    def invoke(
+    def __init__(
+        self,
+        charset: str = "utf-8",
+        env: Optional[Mapping[str, Optional[str]]] = None,
+        echo_stdin: bool = False,
+        catch_exceptions: bool = True,
+    ) -> None:
+        super().__init__(charset=charset, env=env, echo_stdin=echo_stdin)
```
LecrisUT: Is the missing `catch_exceptions` here intentional? I don't see we pass it explicitly at the constructor anywhere also.

plans/main.fmf
happz: `-vvv` is already the "standard" verbosity of pytest invocations in tmt test suite. I don't mind changing it - it's probably no longer needed, pytest itself ignores the third `-v` but some plugins may be more verbose - but not in this PR.

tests/__init__.py
happz: I believe it is needed, because without it, I don't know what would replace it in the following signatures:

```python
@pytest.fixture(name='run_tmt')
def fixture_run_tmt() -> RunTmt:

def test_invalid_yaml_syntax(run_tmt: 'RunTmt'):
```

mypy can infer a lot of things, but this I think still requires an annotation. mypy can then tell me whether `CliRunner().invoke` matches my annotation or not.

LecrisUT: I mean it could be in `conftest.py` and unimported

tests/core/about/main.fmf
happz: Added in a4720f53.

tests/__init__.py
happz: It's probably better to use `kwargs` here. I was told `catch_exceptions` is better placed in `invoke()` where it's correctly propagated to Click.

Changed in a4720f53 and f6c87cc3.

tests/__init__.py
happz: I see. No, I did not observe this behavior, here or in other projects, in signatures of tests like `def test_invalid_yaml_syntax(run_tmt: 'RunTmt')` I still needed an annotation. No sign mypy would let me write just `test_invalid_yaml_syntax(run_tmt)` and infer the type from the corresponding fixture. First time hearing about this feature.

LecrisUT: Ah, in the case I had in mind it did `from conftest` [^1]. It does not make the test folder a package, and it doesn't have deeper file nesting. Guess it might not be a thing.

[^1]: https://github.com/scikit-build/scikit-build-core/blob/f4429b08316a485835c57e0beff256a7b505bc6c/tests/test_editable_unit.py#L19

tests/__init__.py
LecrisUT: But why ignore those in the constructor, and even more so when we do not even use the `catch_exceptions=False` in the constructors?

happz: What do you suggest?

LecrisUT: Just to use the default contructor.

happz: I see. I simplified in 0696d5bf, but I'd still like to use custom `__init__`, to enforce some basic behavior. Later we can add more parameters if needed, they are indeed unused at the moment, but enforcing encoding and echo from the beginning seems like a good idea to me.

tests/integration/test_nitrate.py
```diff
@@ -9,13 +9,12 @@
...
 from ruamel.yaml import YAML
 
-import tmt.cli._root
 import tmt.log
 from tests import CliRunner
 from tmt.utils import ConvertError, Path
 
 if TYPE_CHECKING:
-    from tests import RunTmt
+    pass
```
LecrisUT: Can remove this instead. And same in `test_polarion`

happz: Addressed in 918fbcf9

LecrisUT:
Oh wow, first time I see a race condition in RTD: https://app.readthedocs.org/projects/tmt/builds/31553371/
Should be fine though, previous build did not complain.

therazix:
Failed documentation build seems unrelated. Merging.

PR #4585, opened 2026-02-16
Add generic tmt context to AGENTS.md

happz:
And AGENTS.md is born from Claude instructions. This way we should cover other agents as well as Claude.

Pull Request Checklist

* [x] implement the feature


.claude/CLAUDE.md
```diff
@@ -2,6 +2,180 @@
...
+```bash
+# Run pre-commit checks
+pre-commit run --all-files
+```
+
+### Code Quality
+```bash
+# Run pre-commit checks
+pre-commit run --all-files
+```
```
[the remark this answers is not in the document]
happz: Addressed in 12fcda18

therazix (approved):

AthreyVinay (approved):

AGENTS.md
```diff
@@ -0,0 +1,395 @@
+# tmt AI Instructions
+
+This file contains instructions for AI assistants (Claude Code, Gemini CLI, GitHub Copilot, etc.) when working with the tmt codebase. Different sections cover different tasks.
+
+## Overview
+
+TMT (Test Management Tool) is a comprehensive Python-based testing framework that provides
+a user-friendly way to work with tests. It implements the Metadata Specification using the Flexible
+Metadata Format (FMF) for storing test execution data directly within git repositories.
```
vaibhavdaren: explicitly stating "FMF (which utilizes YAML syntax)" will prevent the AI from hallucinating a proprietary syntax when asked to write a new FMF file.

AGENTS.md
```diff
@@ -0,0 +1,395 @@
...
+
+## Development Commands
+
```
vaibhavdaren: "When tasked with executing builds or tests in the terminal, use the following commands:" This gives the AI a clear condition for when to care about this block.

AGENTS.md
```diff
@@ -0,0 +1,395 @@
...
+1. Create plugin module in appropriate `tmt/steps/<step>/` directory
+2. Inherit from base plugin class of the step
+3. Implement required methods (`go()`, `show()`, `wake()`)
+4. Add plugin registration and CLI options
+
+### Test Development
+
+- Unit tests should mock external dependencies
+- Integration tests can use the container framework
+- Follow existing test patterns in `tests/` directory
```
vaibhavdaren: Too ambiguous for an AI. You are forcing the agent to guess which file represents your gold standard.
Give it a specific anchor. Change this to: "Follow existing test patterns (e.g., refer to tests/prepare/install)"

AGENTS.md
```diff
@@ -0,0 +1,395 @@
...
+
+### Code Standards
+
+- Python 3.9+ required
+- Type hints enforced via mypy and pyright with strict settings
+- Comprehensive linting via ruff (configured in `pyproject.toml`)
+- Code formatting via ruff (configured in `pyproject.toml`)
+- Comprehensive docstring requirements
+- Security-focused linting with bandit checks provided by ruff
```
vaibhavdaren: make these instructions active. 
```
### Code Standards

- **Target Python 3.9+:** Use modern syntax only.
- **Use strict type hints:** Fully type all signatures, arguments, and return types to satisfy `mypy` and `pyright`.
- **Adhere to `ruff`:** Format and lint code strictly according to `pyproject.toml`.
- **Write comprehensive docstrings:** Document modules, classes, and functions.
- **Write secure code:** Pass all `bandit` security checks enforced by `ruff`.
```

AGENTS.md
```diff
@@ -0,0 +1,395 @@
...
+
+
+## Release Notes Generation
```
vaibhavdaren: if you load this entire AGENTS.md file into your AI's context every time you start a chat, you are forcing the AI to read and "remember" how to format release notes even when you just want it to help you debug a unit test in `tmt/steps/provision`. So this should be moved to a seprate file.

happz: Which file then? If these are instructions for an agent which shall be in the `AGENTS.md` file, where should we put instructions that are supposed to be read by an agent, but only in some situations?

AGENTS.md
happz: Which test shall I recommend? @psss @thrix @LecrisUT any tips?

AGENTS.md
happz: Updated in cd346633.

(Although this will *not* prevent agent from hallucinating invalid syntax. The way they fixed the problem of invalid JSON blobs was a layer of JSON (and YAML...) syntax validator that keeps returning the produced snippets back to the LLM until it passes as syntactically correct...)

AGENTS.md
happz: Updated in cd346633.

AGENTS.md
happz: Updated in cd346633.

AGENTS.md
LecrisUT: `/tests/prepare/install` probably because it has the context of running on multiple images.

AGENTS.md
vaibhavdaren: I'd suggest moving the actual release note instructions into a dedicated file, something like `docs/agents/release-notes-prompt.md` or somewhere else.

What you keep in `AGENTS.md` is just a pointer to that new file. You could add a simple rule in AGENTS.md that says something like:

```md
If the task involves generating release notes,
load and follow the instructions in docs/agents/release-notes-prompt.md.
```

This way, `AGENTS.md` stays lean and acts as a router. The heavy, specific formatting instructions are only loaded into the AI's context when release note generation is explicitly requested. If a developer is just asking the AI to debug a unit test, those extra instructions are ignored, keeping the AI focused entirely on the code.


@thrix is there anything better we could do here?

LecrisUT: > something like `docs/agents/release-notes-prompt.md` or somewhere else.

Can we please not contaminate other directories with machine language

psss: What do you suggest then as an alternative? Keeping everything related to agents in a single huge file? That does not seem to be very maintainable to me. Having a dedicated folder for agent skills where each area would have its only file sounds reasonable to me. If we agree on the specific location there will not be any "contamination" with other docs.

LecrisUT: > What do you suggest then as an alternative? Keeping everything related to agents in a single huge file? That does not seem to be very maintainable to me.

The thing is not human readable/maintainable from the beginning IMO.

>  Having a dedicated folder for agent skills where each area would have its only file sounds reasonable to me.

Yes, a dedicated folder would be preferred.

AGENTS.md
```diff
@@ -0,0 +1,399 @@
...
+   ```
+
+6. **Move pending entries**: Check if any release notes exist in `docs/releases/pending/` for PRs in this milestone:
+   - Move matching fmf files to the release directory
+   - Example: `git mv docs/releases/pending/1234.fmf docs/releases/1.65.0/1234.fmf`
+   - This handles notes that were added during PR development
+
+7. **Review**: Ensure all significant changes for the milestone are covered and the formatting matches the existing style in the `docs/releases/` directory.
+
+## Using This Instruction with Different AI Tools
```
psss: This section and the following ones (MCP Setup and Custom Instructions) are human-oriented documentation, not an interesting reading for agents, let's move them to the contribute page.

AGENTS.md
```diff
@@ -0,0 +1,399 @@
+# tmt AI Instructions
+
+This file contains instructions for AI assistants (Claude Code, Gemini CLI, GitHub Copilot, etc.)
+when working with the tmt codebase. Different sections cover different tasks.
+
+## Overview
+
+TMT (Test Management Tool) is a comprehensive Python-based testing framework that provides
```
psss:
```suggestion
tmt (Test Management Tool) is a comprehensive Python-based testing framework that provides
```

`tmt` and `fmf` should be always written in lower-case, except for envvar, class and similar exceptions...

AGENTS.md
happz: Added in 0b73a5f7.

AGENTS.md
happz: Yes, let's keep assistant instructions separated from user-focused documentation. Moved in 0b73a5f7.

AGENTS.md
happz: Moved in 0b73a5f7.

AGENTS.md
happz: Fixed in 0b73a5f7.

vaibhavdaren (approved):
Thankyou for the changes. LGTM.

psss (approved):

AGENTS.md
psss: Thanks.

psss:

PR #4586, opened 2026-02-16
proposal: Make `PluginRegistry` auto-load plugins on first access

AthreyVinay:
Currently, plugins are only loaded when `tmt.plugins.explore()` is explicitly called, which happens automatically in `tmt run` but not for unit tests. See discussion: https://github.com/teemtee/tmt/pull/4526#discussion_r2789144462

This tries to make `PluginRegistry` automatically trigger plugin discovery on first access. Also, `ALREADY_EXPLORED` makes the `explore()` call idempotent.

### DIFF
--- a/tests/unit/artifact/conftest.py
+++ b/tests/unit/artifact/conftest.py
@@ -2,17 +2,11 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 
-import tmt.plugins
 from tmt.steps.prepare import install
 from tmt.steps.prepare.artifact import get_artifact_provider
 from tmt.steps.prepare.artifact.providers import koji as koji_module
 from tmt.steps.prepare.artifact.providers.brew import BrewArtifactProvider
 from tmt.steps.prepare.artifact.providers.koji import KojiArtifactProvider
-
-
-@pytest.fixture(scope='session')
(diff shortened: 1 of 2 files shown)


happz:
-1 from me: explicit is better than implicit, I not a fan of implicit magic, import-time stuff one cannot avoid, and I despise side-effects. The mere fact of accessing a plugin registry will cause the whole plugin discovery to trigger, and it cannot be disabled. Guarding access to the plugin registry, and touching it at the right time, is much harder than simply deciding when the plugin discovery would be explicitly invoked.

Test fixture seemed like a good solution to me. Pytest supports also fixtures without any actual value, see https://docs.pytest.org/en/stable/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures:

```python
@pytest.fixture
def discovered_plugins():
  ... explore() ...


@pytest.mark.usefixtures('discovered_plugins')
def test_foo():
  ...
```

The great advantage of this is the clear and explicit marking of tests that require real-life plugins to be discovered first. With implicit magic happening in the background it's much harder to tell whether a given test needs something more than its arguments.

AthreyVinay:
Sure - this was more of a proposal and got a deserved answer. Thanks.

PR #4589, opened 2026-02-19
Add a unit test for check result combinations

### LINKED ISSUE #4578: Improve test coverage for results handling and interpretation
We really need that unit test to collect all combinations and corner cases and test what we expect to happen.

See #3540.

therazix:
Resolves #4578 

Assisted-by: Cursor

Pull Request Checklist

### DIFF
--- a/tests/unit/test_results.py
+++ b/tests/unit/test_results.py
@@ -994,10 +994,10 @@ def test_result_interpret_with_checks(case: CheckPhasesCase) -> None:
         ),
     ],
 )
-def test_check_phases_duplicate_phase(case: CheckPhasesDuplicateCase) -> None:
+def test_check_phases_duplicate(case: CheckPhasesDuplicateCase) -> None:
     """
-    Test the interpretation of check results with duplicate
-    phases (phases with the same name and event).
+    Test the interpretation of duplicate check results
+    (check results with the same name and event).
     """
     result = Result(
         name="test-case",


tests/unit/test_results.py
```diff
@@ -226,30 +226,58 @@ def test_result_interpret_check_phases() -> None:
...
+    assert interpreted_before.check[1].result == ResultOutcome.PASS  # check1 AFTER_TEST
+    assert interpreted_before.check[2].result == ResultOutcome.PASS  # check2 BEFORE_TEST (INFO)
+
+    assert interpreted_after.check[0].result == ResultOutcome.PASS  # check1 BEFORE_TEST
+    assert interpreted_after.check[1].result == ResultOutcome.FAIL  # check1 AFTER_TEST
+    assert interpreted_after.check[2].result == ResultOutcome.PASS  # check2 BEFORE_TEST (INFO)
+
+    assert interpreted_both.check[0].result == ResultOutcome.FAIL  # check1 BEFORE_TEST
+    assert interpreted_both.check[1].result == ResultOutcome.FAIL  # check1 AFTER_TEST
+    assert interpreted_both.check[2].result == ResultOutcome.PASS  # check2 BEFORE_TEST (INFO)
```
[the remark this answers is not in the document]
LecrisUT: Yes to the parametriztion. No to including the `name` as it is arbitrary.

tests/unit/test_results.py
```diff
@@ -270,6 +298,577 @@ def test_result_interpret_edge_cases() -> None:
...
+    ),
+    [
+        # Test interpret RESPECT:
+        pytest.param(
+            ResultOutcome.PASS,
+            ResultInterpret.RESPECT,
+            ResultOutcome.PASS,
+            CheckResultInterpret.RESPECT,
+            ResultOutcome.PASS,
+            [],
```
LecrisUT: Can you kwargs these for quick readability? The param names could also use some work because `result_*`, `expected_*` kinda clash. Maybe `overall_outcome`, `check_outcome`.

tests/unit/test_results.py
```diff
@@ -270,6 +298,577 @@ def test_result_interpret_edge_cases() -> None:
...
+)
+def test_check_phases_duplicate_phase(
+    result_outcome: ResultOutcome,
+    interpret: ResultInterpret,
+    check_result_outcome1: ResultOutcome,
+    check_result_outcome2: ResultOutcome,
+    check_interpret: CheckResultInterpret,
+    expected_outcome: ResultOutcome,
+    expected_note_contains: list[str],
+) -> None:
```
LecrisUT: Docstrings on the tests to explain these would be useful, particularly to highlight how this differs from the previous case, and what is the intention with the duplicates (or link to the story)

tests/unit/test_results.py
LecrisUT: Isn't this becoming a special case of the other ones or vice-versa?

tests/unit/test_results.py
happz: IIUIC, kwargs probably not possible with `pytest.param`, we'd need to bundle these into a named tuple or a dataclass, but we could use default values there, e.g.

```python
@dataclasses.dataclass
class Foo:
    result_outcome: ResultOutcome = ResultOutcome.PASS
    interpret: ResultInterpret = ResultInterpret.RESPECT
    check_result_outcome: ResultOutcome = ResultOutcome.PASS
    check_interpret: CheckResultInterpret = CheckResultInterpret.RESPECT
    expected_outcome: ResultOutcome = ResultOutcome.PASS
    expected_note_contains: list[str] = dataclasses.field(default_factory=[])

...

pytest.param(
    Foo(
        check_result_outcome=ResultOutcome.FAIL,
        expected_outcome=ResultOutcome.FAIL,
        expected_note_contains=["check 'check1' failed", "original test result: pass"],
    ),
    id="pass-respect-fail-respect",
),
```

But that's also verbose...

tests/unit/test_results.py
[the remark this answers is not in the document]
therazix: Changed in fa0e2d627c21c3ae18e1465949c6be72e6a5e83e.

tests/unit/test_results.py
therazix: Changed to dataclasses in fa0e2d627c21c3ae18e1465949c6be72e6a5e83e.

tests/unit/test_results.py
therazix: The intention of the `test_check_phases_duplicate_phase` test is to verify the correct behavior when the same check is defined multiple times (same name and event). Added docstring in fa0e2d627c21c3ae18e1465949c6be72e6a5e83e.

tests/unit/test_results.py
therazix: Yes, I added some more test cases, and now there is no need for that test. Removed in fa0e2d627c21c3ae18e1465949c6be72e6a5e83e.

tests/unit/test_results.py
```diff
@@ -270,6 +211,802 @@ def test_result_interpret_edge_cases() -> None:
...
+                    "original test result: pass",
+                ],
+            ),
+            id="pass-respect-fail-warn-respect",
+        ),
+    ],
+)
+def test_check_phases_duplicate_phase(case: CheckPhasesDuplicateCase) -> None:
+    """
+    Test the interpretation of check results with duplicate phases.
```
LecrisUT: Duplicate phases being phases with exact same name? I guess that is a special case of `upgrade`? Maybe could be more explicit here.

tests/unit/test_results.py
```diff
@@ -270,6 +211,802 @@ def test_result_interpret_edge_cases() -> None:
...
+            CheckResult(name="check1", result=case.check_outcome2, event=CheckEvent.AFTER_TEST),
+        ],
+    )
+
+    interpret_checks = {
+        "check1": case.check_interpret,
+        "check2": CheckResultInterpret.RESPECT,
+    }
+
+    interpreted = result.interpret_result(case.result_interpret, interpret_checks)
```
AthreyVinay: Nit: There’s a bit of repetition around interpret_checks and the note assertion block across these tests.
Might be worth extracting a small helper (e.g. make_interpret_checks / assert_note_contains) to reduce duplication. Totally optional though — current version is clear

mcasquer:
Nothing to add apart from my mate's comments, execution looks good
```
test_results.py ........................................................................                                                               [100%]

===================================================================== 72 passed in 0.80s =====================================================================
```

tests/unit/test_results.py
therazix: It's not meant to be a special case of `upgrade`. I added it to cover a scenario where multiple checks with the same name and event are present based on [this comment](https://github.com/teemtee/tmt/pull/3540#issuecomment-2669036365):
> I presume this will break when multiple checks of same name are present.

Changed the docstring in f712cbb42c73a1d915654e95485cb9729e2df37b.

tests/unit/test_results.py
therazix: Changed in f712cbb42c73a1d915654e95485cb9729e2df37b.

tests/unit/test_results.py
LecrisUT: > It's not meant to be a special case of `upgrade`

Misphrased that part. I meant more like do we have an example of where that is the case? Duplicate phases with same name sounds quite suspicious given that we have a `itertools.count` that should disambiguate all of them.

tests/unit/test_results.py
AthreyVinay: Thanks for the improvement.... just another minor nit. However, "optional again". we can avoid the `else`:
``` python
assert interpreted.result == case.overall_outcome
if not case.note_contains:
    assert not interpreted.note
    return

assert interpreted.note
for expected_note in case.note_contains:
    assert expected_note in interpreted.note
```
Again, just OPTIONAL!

therazix: Changed in 7144d03751d39d0810e98a3571f88a73783dffbe.

tests/unit/test_results.py
therazix: As far as I know, nothing prevents you to define a test like this:
```
test: echo "Test"
framework: shell
check:
  - dmesg
  - dmesg
```
Which will have 4 check results with the same `name` and `event`. (2 for before-test, and 2 for after-test)

tests/unit/test_results.py
```diff
@@ -16,6 +18,43 @@
...
+    result_outcome: ResultOutcome
+    result_interpret: ResultInterpret
+    check_outcome1: ResultOutcome
+    check_outcome2: ResultOutcome
+    check_interpret: CheckResultInterpret
+    overall_outcome: ResultOutcome
+    note_contains: list[str]
+
+
+def assert_result(case: Union[CheckPhasesCase, CheckPhasesDuplicateCase], result: Result) -> None:
```
skycastlelily, line 42: You might want to  change Union[CheckPhasesCase, CheckPhasesDuplicateCase]  to CheckPhasesCase | CheckPhasesDuplicateCase,so "from typing import Union" could be removed.

therazix:
> LGTM, just nitpicks:
> 
> you might want to extend the test a bit:
- The first scenario is already covered by `pass-respect-fail-info` test.
- I added the test for the second scenario in 91c660e12027faacd46e7bf86e11e67481084efe.
- I don't think it's possible to have a check with an empty name, so I didn't add a test for the third scenario.

tests/unit/test_results.py
therazix: In the rest of the codebase, we use `Union` more often, so I would keep it as it is.

LecrisUT: We cannot change it without using future.annotations, and that part would need to be its own change because there are some fixes needed for it to work

tests/unit/test_results.py
LecrisUT: But that's duplicate checks, not duplicate phases?

therazix: Yes, but they also have the same phases. And by phase I mean 'before-test' or 'after-test'.

tests/unit/test_results.py
skycastlelily: Just out of curiosity, dose tmt support Python < 3.10? :)

happz: From `pyproject.toml`:

```toml
requires-python = ">=3.9"
```

tests/unit/test_results.py
LecrisUT: > And by phase I mean 'before-test' or 'after-test'.

So not plan's step phases? Well that's some confusing terminology :upside_down_face:

therazix: I was using the term "phase" based on [this comment](https://github.com/teemtee/tmt/pull/3540#issuecomment-2669036365):
> Also there are before/after phases and users don't specify expected outcome per phase.

But you are right, it could be confusing. I renamed the function and modified the docstring in ccdf15356ceef85f0ede488d200cfebd9ff16183. Is that okay?

PR #4591, opened 2026-02-22
Prefer `Stopwatch.measure` over context manager

happz:
It allows explicit control over return value and exceptions raised by the code. `with` + `try` allows that two, but it's two extra levels of indentation - or we do not bother capturing exceptions, letting them bubble up, but that is a bad policy when we are slowly moving towards capturing results and reporting errors consistently.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/checks/avc.py
+++ b/tmt/checks/avc.py
@@ -3,7 +3,7 @@ import re
 import textwrap
 import time
 from re import Pattern
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional
 
 import jinja2
 
@@ -19,6 +19,7 @@ from tmt.utils import (
     ShellScript,
     Stopwatch,
     render_command_report,
+    safe_call,
 )
 from tmt.utils.hints import hints_as_notes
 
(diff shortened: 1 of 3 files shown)


therazix (approved):

LecrisUT (approved):
LGTM, just one suggestion to mark the `__enter__` as `@deprecated` with explicit opt-out for `measure()`

happz:
> LGTM, just one suggestion to mark the `__enter__` as `@deprecated` with explicit opt-out for `measure()`

I wouldn't call it deprecated, `measure()` is just easier to use given what we often measure, and it returns useful values with less work (return value, exception, timer, without any `try`/`except`).

LecrisUT:
> > LGTM, just one suggestion to mark the `__enter__` as `@deprecated` with explicit opt-out for `measure()`
> 
> I wouldn't call it deprecated, `measure()` is just easier to use given what we often measure, and it returns useful values with less work (return value, exception, timer, without any `try`/`except`).

Sure, fine enough. The suggestion was in order to pick up any stray usages of that pattern, mostly for ourselves. Non-blocker suggestion, so fine either way.

PR #4598, opened 2026-02-24
Fix installation of local rpm in Image Mode

thrix:
The #4525 added volume mounting the workdir into the container during
building while changing the build context directory. Unfortunately
this broke the local rpm installation, which relied on the RPMs
to be placed in the step workdir.

The fix is simple, remove the copying of the rpm files as they
are already provided by the volume mount.

Add tests covering the use case.

Related to [TFT-4368](https://issues.redhat.com/browse/TFT-4368)

Pull Request Checklist

* [x] implement the feature
* [ ] include a release note

### DIFF
--- a/tests/prepare/artifact/lib/common.sh
+++ b/tests/prepare/artifact/lib/common.sh
@@ -101,6 +101,7 @@ get_koji_nvr() {
     local profile="$3"
     unset KOJI_NVR  # Clear any previous value
 
+    local profile_option=""
     [ -n "$profile" ] && profile_option="--profile $profile"
 
     # Get the latest tagged build for the package


tests/prepare/bootc/data/plans.fmf
```diff
@@ -1,14 +1,23 @@
 discover:
     how: fmf
+
+provision:
```
LecrisUT, line 4:
```suggestion
/centos-stream-10:
    provision:
```

tmt/steps/prepare/install.py
```diff
@@ -649,12 +649,6 @@ def install_local(self) -> None:
             PackagePath(self.package_directory / filename.name) for filename in self.local_packages
         ]
 
-        self._engine.containerfile_directives.append(f'RUN mkdir -p {self.package_directory}')
-
-        files = " ".join(str(file.relative_to(self.guest.step_workdir)) for file in filelist)
-
-        self._engine.containerfile_directives.append(f'COPY {files} {self.package_directory}')
-
```
LecrisUT, line 663: It would be useful to keep track what paths are being mounted. Can we assume that everything under `run_workdir` will be mounted and any changes in there will persist (changes would propagate to the tmt runner's copy)?

tests/prepare/bootc/test.sh
```diff
@@ -27,6 +27,28 @@ rlJournalStart
         rlAssertGrep "tree v" $rlRun_LOG
     rlPhaseEnd
 
+    # Test prepare/install on bootc guest - should install rpm via Containerfile
+    rlPhaseStartTest "Prepare/install on bootc guest - install tree package from rpm"
+        # Download tree RPM
+        rlRun "curl -LO https://mirror.stream.centos.org/10-stream/BaseOS/x86_64/os/Packages/tree-2.1.0-8.el10.x86_64.rpm"
```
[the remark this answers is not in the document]
thrix: @gemini-code-assist how would you solve it? :)

tests/prepare/bootc/data/plans.fmf
thrix: @LecrisUT thanks, will fix in an hour

tests/prepare/bootc/test.sh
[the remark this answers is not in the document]
LecrisUT: Something like how we do for `koji build` to get the latest package would be preferred
https://github.com/teemtee/tmt/blob/84ebfb2795dcc85eada317de287966fd10a3c2d2/tests/prepare/artifact/lib/common.sh#L47-L59
Or what @vaibhavdaren is working on with the creating our own rpms. I wouldn't worry too much about this right now until we have proper way to handle these more generally.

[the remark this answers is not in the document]
thrix: @LecrisUT agreed, checking

tmt/steps/prepare/install.py
thrix: @LecrisUT Keep track how? document it or?

Yeah the idea with this volume mounting was to persist data, in case somebody would create some data in the `prepare/shell` step (where this was introduced), which is run in image mode done in the build container, the data created there would persist in the workdir as it would if the script would be run on the host machine.

tests/prepare/bootc/test.sh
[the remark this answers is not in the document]
thrix: @LecrisUT went with `koji -p stream` with `centpkg` installed to get the concrete latest tree rpm name.

tests/prepare/bootc/data/plans.fmf
thrix: Moved under `/centos-stream-10`

tmt/steps/prepare/install.py
LecrisUT: Just a comment pointing to the source code where the volume mounting is defined. Over here is a relevant spot because it is creating raw container commands umlike the other places where it goes through the abstraction layer.

tests/prepare/bootc/test.sh
```diff
@@ -27,8 +27,33 @@ rlJournalStart
         rlAssertGrep "tree v" $rlRun_LOG
     rlPhaseEnd
 
+    # Test prepare/install on bootc guest - should install rpm via Containerfile
+    rlPhaseStartTest "Prepare/install on bootc guest - install tree package from rpm"
+        # Download tree RPM
+        rlRun -s "koji -p stream list-tagged --arch x86_64 --rpms --latest --quiet c10s-candidate tree-pkg | head -1"
```
LecrisUT: Why not just koji download-build directly?

Would prefer to align it and reuse parts of the `artifacts/lib` stuff for simpler migration in the future. Not a blocker for now though.

tests/prepare/bootc/data/plans.fmf
```diff
@@ -1,14 +1,26 @@
...
+            how: shell
+            script:
+              - dnf -y install tree
+
+    /prepare-install:
+        summary: "Test prepare/install on bootc guest - install tree package from local rpm"
+        prepare:
+            how: install
+            package:
+              - tree-2.1.0-8.el10.x86_64.rpm
```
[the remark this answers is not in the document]
thrix: outdated, using a variable to inejct the correct rpm now

tests/prepare/bootc/test.sh
thrix: @LecrisUT how you imagine it would be done with `download-build`? I do not know what is the latest version of tree? I am not able to find how it could download the latest build of `tree-pkg` package ...

```
❯ koji download-build --help
Usage: koji download-build [options] <n-v-r|build_id>

Downloads files from the specified build entry
Note: scratch builds do not have build entries. Use download-task for those
(Specify the --help global option for a list of other help options)

Options:
  -h, --help            show this help message and exit
  -a ARCH, --arch=ARCH  Only download packages for this arch (may be used
                        multiple times)
  --type=TYPE           Download archives of the given type, rather than rpms
                        (maven, win, image, remote-sources)
  --latestfrom=LATESTFROM
                        Download the latest build from this tag
  --debuginfo           Also download -debuginfo rpms
  --task-id             Interperet id as a task id
  --rpm                 Download the given rpm
  --key=KEY             Download rpms signed with the given key
  --fallback-unsigned   When used with --key: download unsigned if signed
                        packages not found
  --topurl=URL          URL under which Koji files are accessible
  --noprogress          Do not display progress meter
  -q, --quiet           Suppress output
```

LecrisUT: Using both of these commands
https://github.com/teemtee/tmt/blob/96977e19f2f7e8c189554f8b7e8c59aacb28b8cb/tests/prepare/artifact/providers/file/test.sh#L30-L33

tmt/steps/prepare/install.py
thrix: @LecrisUT can you check?

tests/prepare/bootc/test.sh
thrix: Used improved `get_koji_nvr` with support for `profiles` and `koji download-build`

thrix:

LecrisUT (approved):

tests/prepare/bootc/test.sh
```diff
@@ -27,8 +28,33 @@ rlJournalStart
         rlAssertGrep "tree v" $rlRun_LOG
     rlPhaseEnd
 
+    # Test prepare/install on bootc guest - should install rpm via Containerfile
+    rlPhaseStartTest "Prepare/install on bootc guest - install tree package from rpm"
+        # Download tree RPM
+        get_koji_nvr "tree-pkg" "c10s-candidate" "stream"
+        rlRun "koji --profile stream download-build --arch=x86_64 $KOJI_NVR" 0 "Download tree RPM"
+        rlRun "TREE_RPM=$(ls tree-*.rpm)"
```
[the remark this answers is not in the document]
thrix: naaah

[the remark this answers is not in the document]
thrix: I would ignore this for now. We know this package has only a single rpm. I would not add any safeguards here.

PR #4599, opened 2026-02-24
Fix the xref formatting of `:tmt:story:` role to normal text

### LINKED ISSUE #4579: Adjust the `:tmt:story:` link formatting
Looking at the rendered text, I find it a little bit confusing that now the story links are shown in a very similar style as the preformated text like keys and options. It's a bit hard to quickly identify which is clickable and what is not. Could we use the previous style so that links are clearly distinguishable?

Also for longer texts like `Results specification documentation` below, I'd say, the large rectangle around the whole title does not suit very well.

## Before

Image

## After

Image

_Originally posted by @psss in https://github.com/teemtee/tmt/pull/4288#discussion_r2800006641_

LecrisUT:
This fixes a regression after #4288 where the format was changed to a code-format.

---

Pull Request Checklist

* [x] implement the feature

Fixes #4579

### DIFF
--- a/docs/ext/tmt_domain/domain.py
+++ b/docs/ext/tmt_domain/domain.py
@@ -14,8 +14,7 @@ from sphinx.util.nodes import make_refnode
 import tmt
 import tmt.log
 
-from .base import TmtXRefRole
-from .story import AutoStoryDirective, StoryDirective, StoryIndex
+from .story import AutoStoryDirective, StoryDirective, StoryIndex, StoryRole
 
 if typing.TYPE_CHECKING:
     from sphinx.addnodes import pending_xref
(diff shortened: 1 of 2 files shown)


happz:
> I don't understand all the sphinx magic,

Same here. I can't shake the feeling this is a step back... We can now use `:tmt:story:` but we pay for it with a black box full of magic only one person understands :(

LecrisUT:
> > I don't understand all the sphinx magic,
> 
> Same here. I can't shake the feeling this is a step back... We can now use `:tmt:story:` but we pay for it with a black box full of magic only one person understands :(

I do keep it as minimal as possible and stick to the reference in upstream. The debugger helps a lot with the navigation. But if there are parts that are still to magic, I can try to document them further, and check with upstream on how to improve some of it.

bajertom:
I tried the change and :tmt:story: links now appear as normal links instead of code-formatted monospace. However, as others, I don't really have much clue about the docs engine behind it so.. LW(orking)TM.

LecrisUT:
Documentation only change, relevant tests are passing, merging

PR #4616, opened 2026-02-26
Move `TMT_SOURCE_DIR` envvar setting out of the `discover` step

happz:
It is added to test environment, spoiling it and making it harder for us and users to think about envvar precedence order.

Variable is now set unconditionaly for all tests, by the execute plugin. That makes it present for all tests, but used by subset only.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -177,7 +177,9 @@ class TestInvocation(HasStepWorkdir, HasEnvironment):
             if discover_phase.name == self.phase.discover_phase:
                 return discover_phase
 
-        raise tmt.utils.GeneralError('Failed to find discover phase for the test.')
+        raise tmt.utils.GeneralError(
+            f"Failed to find discover phase '{self.phase.discover_phase}' for the test."
+        )
 
     @property
     def step_workdir(self) -> Path:
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -163,6 +163,23 @@ class TestInvocation(HasStepWorkdir, HasEnvironment):
     exceptions: list[Exception] = simple_field(default_factory=list)
 
     @property
+    def discovery_phase(self) -> DiscoverPlugin[Any]:
+        """
+        Discovery phase the test comes from.
+        """
+
+        if isinstance(self.phase.discover, DiscoverPlugin):
+            return self.phase.discover
+
+        for discover_phase in self.phase.discover.phases(
+            classes=DiscoverPlugin,  # type: ignore[type-abstract]
+        ):
+            if discover_phase.name == self.phase.discover_phase:
+                return discover_phase
+
+        raise tmt.utils.GeneralError('Failed to find discover phase for the test.')
+
+    @property
     def step_workdir(self) -> Path:
         return self.phase.step_workdir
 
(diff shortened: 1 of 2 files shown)


tmt/steps/execute/__init__.py
```diff
@@ -162,6 +162,25 @@ class TestInvocation(HasStepWorkdir, HasEnvironment):
...
+        """
+
+        if isinstance(self.phase.discover, DiscoverPlugin):
+            return self.phase.discover
+
+        for discover_phase in self.phase.discover.phases(
+            classes=DiscoverPlugin,  # type: ignore[type-abstract]
+        ):
+            if discover_phase.name == self.phase.discover_phase:
+                return discover_phase
```
LecrisUT, line 178: This feels like a fragile navigation to be patched later `self.phase.discover_phase` should already be the the phase not the phase name.

LecrisUT (approved):

tmt/steps/execute/__init__.py
happz: The navigation itself should work fairly reliably. Valid point about the naming and the actual value of `discover_phase`, and I have a WiP patch to do something about it - unfortunately the `upgrad` plugin is involved, and that one does some ugly things to hook into the testing process, so it's not a clear win :/

LecrisUT: Wouldn't it work by changing `discover_phase` -> `discover_phase.name`?

happz: `discover` plays a prominent role in the `upgrade` plugin, I ended up with more lines than just that :/ In any case, it'd be a different patch, as it does touch parts unrelated to `TMT_SOURCE_DIR` move.

tmt/steps/execute/__init__.py
```diff
@@ -162,6 +162,25 @@ class TestInvocation(HasStepWorkdir, HasEnvironment):
...
     exceptions: list[Exception] = simple_field(default_factory=list)
 
+    @property
+    def discover_phase(self) -> DiscoverPlugin[Any]:
+        """
+        Discover phase the test comes from.
+        """
+
+        if isinstance(self.phase.discover, DiscoverPlugin):
+            return self.phase.discover
```
psss, line 172: This part took me some time to understand what's actually happening. Suggest to make that part of the code at least a bit less confusing in #4790.

psss (approved):
Looks good, just got confused a bit by the `Execute.discover` property, which is not problem of this pull request.

tmt/steps/execute/__init__.py
happz: Yes, the ugliness was mentioned (and postponed for future PR) in https://github.com/teemtee/tmt/pull/4616#discussion_r2863462354

happz: Be sure to run full test suite, or at least the upgrade tests - all the juggling with discover step vs discover phase is there to accommodate upgradeability testing.

therazix (approved):

tmt/steps/execute/__init__.py
```diff
@@ -162,6 +162,23 @@ class TestInvocation(HasStepWorkdir, HasEnvironment):
...
+
+        for discover_phase in self.phase.discover.phases(
+            classes=DiscoverPlugin,  # type: ignore[type-abstract]
+        ):
+            if discover_phase.name == self.phase.discover_phase:
+                return discover_phase
+
+        raise tmt.utils.GeneralError('Failed to find discover phase for the test.')
```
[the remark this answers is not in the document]
happz: Has been addressed already.

PR #4620, opened 2026-02-27
Drop manually specifying copr repo in packit

LecrisUT:
Not quite sure why this was used, but the fact that it is only set for rawhide and epel9 is fishy.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -63,20 +63,6 @@ _:
   - &test-base
     job: tests
     trigger: pull_request
-    tf_extra_params:
-      environments:
-        - &copr-teemtee-stable
-          artifacts:
-            # Testing Farm fails to install the repository when a multiline
-            # string is used. The script looks broken, as if one piece of
-            # the process incorrectly loads it, producing broken repo file
-            # URL.
-            - type: repository-file
-              # yamllint disable-line rule:line-length
-              id: https://copr.fedorainfracloud.org/coprs/g/teemtee/stable/repo/fedora-rawhide/group_teemtee-stable-fedora-rawhide.repo
-            - type: repository-file
-              # yamllint disable-line rule:line-length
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -63,20 +63,6 @@ _:
...
-            # Testing Farm fails to install the repository when a multiline
-            # string is used. The script looks broken, as if one piece of
-            # the process incorrectly loads it, producing broken repo file
-            # URL.
-            - type: repository-file
-              # yamllint disable-line rule:line-length
-              id: https://copr.fedorainfracloud.org/coprs/g/teemtee/stable/repo/fedora-rawhide/group_teemtee-stable-fedora-rawhide.repo
-            - type: repository-file
-              # yamllint disable-line rule:line-length
-              id: https://copr.fedorainfracloud.org/coprs/g/teemtee/stable/repo/epel-9/group_teemtee-stable-epel-9.repo
```
psss: These repositories were intentionally enabled in #3452 to allow testing with fresh `fmf` bits as soon as they are released. As discussed today, it would be good to verify that this actually still works, because of concerns about the exact python version required. We could possibly do that if https://github.com/teemtee/fmf/pull/293 is finished and we do a new `fmf` release.

LecrisUT: [This packit issue](https://github.com/packit/packit/issues/2719) could be useful here. Basically we could just add the appropriate dependency at the copr level and let copr do its magic there.

psss:
Blocked on the hacking session discussion, moving to the sprint backlog.

bajertom:
It's blocked on https://github.com/packit/packit/issues/2719

psss:
Agreed to keep it in the backlog until the packit issue is implemented or rawhide python version breaks the setup. Dropping from the sprint.

PR #4621, opened 2026-02-27
Allow adjusting individual tests

### LINKED ISSUE #1793: Support `environment` in the `test` key of `fmf` discover
### Agreed Solution

Here's solution we've reached after the discussion: Add support for a dictionary with `name` and `environment` keys under the `test` key:

```yaml
discover:
    how: fmf
    test:
      - name: /tests/area/feature
        environment:
            VARIABLE: ONE
      - name: /tests/area/feature
        environment:
            VARIABLE: TWO
```

It would be kind of consistent with the `shell` discover plugin where `name` is used for defining the test name.

### Related Issues

* https://issues.redhat.com/browse/RHELTEST-407
* https://issues.redhat.com/browse/TFT-4210

### Original Suggestion

It would be handy if it would be possible to define test parameters for a test as a part of a test name.
Example how it could look like:
```
discover:
  test:
    - /my/test/foo?PARAM1=bar&PARAM2=baz
```
Clearly such parameters would have to be urlencoded or something similar. Additionally, somehow it would have to be defined/clear that such a test name+parameters won't be understood as a test name pattern and matched. This could be a bit tricky since test name patters could be regular expressions containing also character `?`. ATM I do not have a solution for that.

My use case is that I want to parametrize tests in my plan and cannot use `environment` definitions because that particular test can be used in the plan multiple times, each time with different parameters. I know I can create virtual (?) tests like `/my/test/foo/setup1` where environment definitions can be defined but this won't work in a situation where a) parameters may differ and it won't be practical to create extra virtual test for each option and b) I may not be the owner of that actual test repository.

A real-life example could be a tmt alternative to `/distribution/install/brew-build` task where users would specify build NVR or buildID to be used during an installation.

LecrisUT:
This extends the discover/fmf to allow manipulating particular tests, e.g.
```yaml
discover:
  how: fmf
  test:
    - /test/without_adjust
    - name: /test/with_adjust
      environment+:
        DO_TEST: case_1
```

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [ ] ~~mention the version~~ Cannot, see #4627
* [x] include a release note

Fixes #1793

### DIFF
--- a/tmt/steps/discover/fmf.py
+++ b/tmt/steps/discover/fmf.py
@@ -546,7 +546,8 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
 
     Note that unlike the more generic ``adjust-tests``, only one adjust
     rule is allowed. Everything except for the ``name`` key is treated as
-    the adjust rule.
+    the adjust rule, and are applied on top and at the end of any previous
+    rules defined with ``adjust-tests``.
     """
 
     _data_class = DiscoverFmfStepData


tmt/steps/discover/fmf.py
```diff
@@ -696,21 +769,43 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
...
+                ).tests(
+                    filters=filters,
+                    names=[test_adjust.name],
+                    conditions=["manual is False"],
+                    unique=False,
+                    links=link_needles,
+                    includes=includes,
+                    excludes=excludes,
+                )
+            )
```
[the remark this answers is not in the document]
LecrisUT: Aware, but not going to do it here

tmt/steps/discover/fmf.py
```diff
@@ -505,8 +505,6 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
     specified via ``test``, so those tests will also be selected even if
     not modified.
 
-    .. _/discover/fmf/adjust-tests:
-
```
LecrisUT: We cannot even use global targets in these documents because these are rendered twice, once in `plugins/*` and once in `autodoc`

docs/releases/pending/4261.fmf
```diff
@@ -0,0 +1,4 @@
+description: |
+  The :ref:`discover fmf</plugins/discover/fmf>` plugin can now adjust
+  individual tests using the ``test`` key. See the plugin documentation for
+  more detail.
```
tcornell-bus: Could we add a link to the adjust-tests section of the plugin? The user may not know to look at the section in particular.

psss:
Seems that `/tests/core/environment` is [failing](https://artifacts.dev.testing-farm.io/5001a0fe-77b8-429c-89a3-1fa21138ccb8/) plus a bunch of others with:

```
summary: 0 tests selected
warning: No tests found, finishing plan.
```

docs/releases/pending/4261.fmf
psss:
```suggestion
  more details.
```

LecrisUT:
> Seems that `/tests/core/environment` is [failing](https://artifacts.dev.testing-farm.io/5001a0fe-77b8-429c-89a3-1fa21138ccb8/) plus a bunch of others with:
> 
> ```
> summary: 0 tests selected
> warning: No tests found, finishing plan.
> ```

Oh, thanks for the poke. Probably all are failing because of the cli handling/normalization. I'll see what's going on there.

psss:
A question: Just noticed that the `test` key is stored in `step.yaml` with the `__class__` details:

For this simple example:

```yaml
test:
  - one
  - two
  - name: one
    environment:
        greeting: bye
```

I get the following:

```yaml
test:
  - name: one
    adjust-rule:
    __class__:
        module: tmt.steps.discover.fmf
        name: TestsWithAdjusts
  - name: two
    adjust-rule:
    __class__:
        module: tmt.steps.discover.fmf
        name: TestsWithAdjusts
  - name: one
    adjust-rule:
        environment:
            greeting: bye
    __class__:
        module: tmt.steps.discover.fmf
        name: TestsWithAdjusts
```

Do we want/need that? Wouldn't a normalized `dict` be enough?

```yaml
test:
  - name: one
  - name: two
  - name: one
    environment:
        greeting: bye
```

When storing `tests.yaml` for example the `link` key does not contain these. I guess those two should be handled in a similar way, wdyt?

tmt/steps/discover/fmf.py
```diff
@@ -14,10 +14,80 @@
...
-from tmt.container import container, field
+from tmt.container import SerializableContainer, SpecBasedContainer, container, field
 from tmt.steps.prepare.distgit import insert_to_prepare_step
-from tmt.utils import Command, Path
+from tmt.utils import Command, NormalizationError, Path
+
+
+class _RawTestsWithAdjusts(TypedDict, total=True):
+    name: str
+    adjust_rule: Optional[_RawAdjustRule]
```
psss: I'm not sure about the `adjust_rule` here. Couldn't we rather store adjust keys together with `name`? This would simplify the `step.yaml` format and possibly also the implementation? What about just providing an `adjust_rule` property which would just filter out the `name` key?

tmt/steps/discover/fmf.py
```diff
@@ -696,21 +790,44 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
...
-        return tree.tests(
-            filters=filters,
-            names=names,
-            conditions=["manual is False"],
-            unique=False,
-            links=link_needles,
-            includes=includes,
-            excludes=excludes,
-        )
+        if tests_without_adjust:
```
psss: This condition branch should also be enabled when no `test` filter is provided, e.g. discover all tests.

tmt/steps/discover/fmf.py
```diff
@@ -14,10 +14,80 @@
...
+            return cls(name=spec)
+        return cls(**spec)
+
+    def to_spec(self) -> Union[str, _RawTestsWithAdjusts]:
+        if self.adjust_rule:
+            return _RawTestsWithAdjusts(
+                name=self.name,
+                adjust_rule=self.adjust_rule,
+            )
+        return self.name
```
LecrisUT, line 52: > A question: Just noticed that the test key is stored in step.yaml with the __class__ details:

@psss Re ☝️, yes that was my intent with the part here, but it seems that did not happen. Seems like I am missing something and probably not accounting for the `recipe.yaml` part either. @therazix if you have some hints about the `recipe.yaml` part?

docs/releases/pending/4261.fmf
LecrisUT: Unfortunately we cannot because it is defined as a section in the docstring. Tried with a global target, but that failed on me because that docstring is rendered twice and fails at the global target.

tmt/steps/discover/fmf.py
psss: I meant `tests.yaml` not `recipe.yaml`. But yeah, recipes might be related here as well.

tmt/steps/discover/fmf.py
LecrisUT: True, forgot about that. We are still missing something because it should not have saved as a `TestsWithAdjusts` at all if `adjust_rule` was `None`.

tmt/steps/discover/fmf.py
happz: `__class__` comes from `serialize()`, and it's important for `unserialize()`. It has no place in *spec* though, which should be as if coming from an fmf file.

tmt/steps/discover/fmf.py
LecrisUT: Resolved in 3722295

tmt/steps/discover/fmf.py
LecrisUT: Indeed that's why the testing-farm failed. Fixed in e2f9896

tmt/steps/discover/fmf.py
LecrisUT: Resolved by calling to/from spec. And `recipe.yaml` seems to just work because it takes the resolved tests data :)

tests/discover/data/plans.fmf
```diff
@@ -233,3 +233,14 @@ execute:
...
+    /adjust-individual-tests:
+        discover:
+            how: fmf
+            test:
+            # With adjust
+            - name: /tests/discover1
+              duration+: "*2"
+            # Without any adjust
+            - /tests/discover2
+            - name: /tests/discover3
```
LecrisUT, line 246: I was considering adding the running duplicate tests in here as well, but it does not quite fit well. Anyone know where the duplicated tests is covered?

psss: I believe the duplicates are covered here:

https://github.com/teemtee/tmt/blob/f2510a3552c533a492891979cbecfe54854486d9/tests/test/select/test.sh#L147-L204

LecrisUT:
> But several CI tests seem to be failing

Hmm I see a `recipe.yaml` failure (legit, will see what's going on), `upgrade` (less clear on this one), and `beakerlib.xml` (just needs rebase). Thanks for the comment.

tmt/steps/discover/fmf.py
```diff
@@ -14,10 +14,82 @@
...
+
+
+def normalize_tests_with_adjusts(
+    key_address: str,
+    value: Any,
+    logger: tmt.log.Logger,
+) -> list[TestsWithAdjusts]:
+    def normalize_raw_item(raw_item: Any, index: Optional[int] = None) -> TestsWithAdjusts:
+        if isinstance(raw_item, TestsWithAdjusts):
+            return raw_item
```
LecrisUT, line 62: @therazix Need your opinion on this thing. It seems I am encountaring this because when doing `tmt run --recipe` it goes first through:
- `unserialize` from `RecipeManager.load` -> tests become `TestsWithAdjust`
- `normalization` from `Step.data` (`_normalize_data`) -> tries to normalize the data as if it came from cli/fmf but it was actually already normalized

Something feels wrong

happz: > * `normalization` from `Step.data` (`_normalize_data`) -> tries to normalize the data as if it came from cli/fmf but it was actually already normalized

This shouldn't hurt, normalizing already normalized data should be idempotent operation. If it gets different outcome every time it's applied, something's wrong.

happz: Right, it took me a while... This should be correct handling, normalization callback can easily receive already normalized data, in their internal representation. In such a case, nothing better to do, return the already normalized value.

tmt/steps/discover/fmf.py
```diff
@@ -696,21 +789,53 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
...
+                # Recalculate the adjusted tree only if necessary
+                adjusted_tree = tree
+                if test.adjust_rule:
+                    adjusted_tree = tmt.Tree(
+                        logger=self._logger,
+                        path=tree_path,
+                        fmf_context=self.step.plan.fmf_context,
+                        additional_rules=[*self.data.adjust_tests, test.adjust_rule],
+                    )
+                tests.extend(
```
happz: Would `tests += adjusted_tree.tests(...)` work? It could save us one indentation level below.

tmt/steps/discover/fmf.py
LecrisUT: Yeah, but it is suspicious that this is the first time this occurs, and I wonder if we would not start to have similar issues when expanding the recipe loader beyond the discover step.

tests/test/select/test.sh
```diff
@@ -161,30 +161,53 @@ rlJournalStart
...
+        rlAssertGrep 'tests: /tier/two, /tier/one, /tier/two and /tier/two' $output
+        rlAssertGrep 'summary: 4 tests selected' $output
         rlRun "grep -A 1 summary $output | tail -1 | grep '/tests/tier/two'"
         rlRun "grep -A 2 summary $output | tail -1 | grep '/tests/tier/one'"
         rlRun "grep -A 3 summary $output | tail -1 | grep '/tests/tier/two'"
+        rlRun "grep -A 4 summary $output | tail -1 | grep '/tests/tier/two'"
+        # Check the order from tests.yaml as well to distinguish the adjusted test
+        tests_yaml="$run$plan/discover/tests.yaml"
+        rlAssertExists "$tests_yaml"
+        # First one has the adjust
```
psss: Did you mean `Third one`? The first one has `result: respect`.

docs/releases/pending/4621.fmf
```diff
@@ -0,0 +1,4 @@
+description: |
```
psss, line 1: Typo in filename? Should be `4621.fmf`?

tmt/steps/discover/fmf.py
```diff
@@ -452,6 +531,21 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
...
+       discover:
+           how: fmf
+           test:
+             - name: ^/test/one$
+               check+:
+                 - how: avc
+
+    Note that unlike the more generic ``adjust-tests``, only one adjust
+    rule is allowed. Everything except for the ``name`` key is treated as
+    the adjust rule.
```
psss: Perhaps it would be good to explicitly document which one is performed first / which one has preference?

tmt/steps/discover/fmf.py
```diff
@@ -696,21 +789,49 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
...
-            names=names,
-            conditions=["manual is False"],
-            unique=False,
-            links=link_needles,
-            includes=includes,
-            excludes=excludes,
-        )
+        if not self.data.test or not any(test.adjust_rule for test in self.data.test):
+            # - not test: we do not have any names filter (i.e. we take all tests)
+            # - not any(.adjust_rule): we do not any tests with custom adjust
```
```suggestion
            # - not any(.adjust_rule): we do not have any tests with custom adjust
```

docs/releases/pending/4621.fmf
```diff
@@ -0,0 +1,4 @@
...
+  The :ref:`discover fmf</plugins/discover/fmf>` plugin can now adjust
+  individual tests using the ``test`` key. See the plugin documentation for
+  more details.
```
psss, line 4: Hm, Claude suggested a bit more detailed release note:

    The :ref:`discover fmf</plugins/discover/fmf>` plugin now supports adjusting
    individual tests inline. Tests can be specified as objects with a ``name`` key
    and adjustment fields, for example ``duration+: "*2"`` to double the test duration.
    This allows fine-grained control over test properties without requiring global
    ``adjust-tests`` rules.

Doesn't sound bad, what do you think?

tmt/steps/discover/fmf.py
```diff
@@ -696,21 +789,49 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
...
+            for test in self.data.test:
+                # Recalculate the adjusted tree only if necessary
+                adjusted_tree = tree
+                if test.adjust_rule:
+                    adjusted_tree = tmt.Tree(
+                        logger=self._logger,
+                        path=tree_path,
+                        fmf_context=self.step.plan.fmf_context,
+                        additional_rules=[*self.data.adjust_tests, test.adjust_rule],
+                    )
```
psss, line 825: Here `else` clause with `adjusted_tree = tree` could make it a tiny bit more readable. Just a nitpick.

tmt/steps/discover/fmf.py
LecrisUT: In what sense? Like the order of the rule in here vs `adjust-tests`?

tmt/steps/discover/fmf.py
```diff
@@ -14,10 +14,82 @@
...
+):
+    name: str
+
+    adjust_rule: Optional[_RawAdjustRule] = None
+
+    @classmethod
+    def from_spec(cls, spec: Union[str, _RawTestsWithAdjusts]) -> Self:
+        if isinstance(spec, str):
+            return cls(name=spec)
+        name = spec.pop("name")
```
psss: Modifying provided `dict` parameters probably not a good idea in general. Not sure if we can expect any clash here (e.g. the spec dictionary used elsewhere). But still... might be good to do this cleanly?

docs/releases/pending/4621.fmf
LecrisUT: I prefer not to give too much details in the release notes because they can quickly be outdated and it duplicates the documentation that needs to be tracked. Would rather manually try to construct the ref to the document and anchor if it comes to it.

tmt/steps/discover/fmf.py
LecrisUT: Fair, would really want us to catch when that is the case since it sounds like there might be something that would need a re-review.

Edit: staged the fix, will add it with the other review changes
Edit: and applied

tmt/steps/discover/fmf.py
```diff
@@ -359,6 +435,9 @@ class DiscoverFmf(tmt.steps.discover.DiscoverPlugin[DiscoverFmfStepData]):
     detailed information about how exactly the regular expressions are
     handled.
 
+    The ``tests`` key accepts a dict with at least a key ``name``. See
```
thrix: Isn't the name of the key `test` not `tests` ?

```suggestion
    The ``test`` key accepts a dict with at least a key ``name``. See
```

thrix (approved):
No more comments from me outside of what is already mentioned.

tmt/steps/discover/fmf.py
LecrisUT: Elaborated in 7ce66c9 in case that is what you meant

psss: Yes, exactly. Thanks.

docs/releases/pending/4621.fmf
psss: Ok.

tmt/steps/discover/fmf.py
therazix: @LecrisUT Sorry, I didn't get the ping notification for some reason. I experienced something similar while working on recipes with `Links`. Normalization happened twice as well. I will have to look into it.

bajertom: Is it a blocker for this PR? @LecrisUT

LecrisUT: > Is it a blocker for this PR? @LecrisUT

No. Ultimately that would have to be fixed more globally in the recipe PR, but this one would be fine to go earlier than that. It is good that we have a confirmation that it is mostly the new recipe process that brought this change and we did not miss something somewhere else.

PR #4632, opened 2026-03-02
Allow and skip empty elements in Command construction

LecrisUT:
Technically a bug because if the first element is `""` it will become `" actual_command"` which can have some tricky consequences

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -1096,7 +1096,7 @@ CommonDerivedType = TypeVar('CommonDerivedType', bound='Common')
 #: A single element of command-line.
 _CommandElement = str
 #: A single element of raw command line in its ``list`` form.
-RawCommandElement = Union[str, Path]
+RawCommandElement = Union[str, Path, None]
 #: A raw command line form, a list of elements.
 RawCommand = list[RawCommandElement]
 
@@ -1197,7 +1197,7 @@ class Command:
     """
 
     def __init__(self, *elements: RawCommandElement) -> None:
-        self._command = [str(element) for element in elements]
+        self._command = [str(element) for element in elements if element]
 
     def __str__(self) -> str:
         return self.to_element()


tmt/utils/__init__.py
```diff
@@ -1096,7 +1096,7 @@ def get_output(self) -> Optional[str]:
 #: A single element of command-line.
 _CommandElement = str
 #: A single element of raw command line in its ``list`` form.
-RawCommandElement = Union[str, Path]
+RawCommandElement = Union[str, Path, None]
```
happz, line 1099: I can see the issue with `Command('ls', '')`, but how would `None` get it? Is the relaxation of this type needed? What will be easier if we could let `None` travel through our code unchecked until dropped in `Command`? That seems risky.

LecrisUT: Well without this whenever we _want_ to use this feature we would have to use an empty `""` explicitly. `None` seems more explicit in that case as an element that will not be expanded.

happz: Hmm, isn't `None` or `""` appearing as command element the actual issue, i.e. something feeding `Command` with things that cannot function as command elements? Dropping it would make things work, but that really seems like silencing errors. If you ran into something like that, we should probably fix the callsite, and maybe raise a `ValueError` to catch future violations. Empty strings are sneaky bastards, we should root them out as soon as we spot them, if they travel through our calls, we are probably missing checks somewhere.

LecrisUT: Well this happens in constructs such as
https://github.com/teemtee/tmt/blob/5ba7562c4dd9acd61b0aefa99bfea93e6dce9d12/tmt/package_managers/dnf.py#L41-L42
which I think would be perfectly valid

LecrisUT: I mean prefectly valid to avoid the prior `if` check and use the whole thing

https://github.com/teemtee/tmt/blob/5ba7562c4dd9acd61b0aefa99bfea93e6dce9d12/tmt/package_managers/bootc.py#L95-L98

Becoming just

```python
command = Command(self.guest.facts.sudo_prefix, 'bootc')
```

happz: Interesting. So dropping the test, and just constructing the command, relying on `Command` filtering out all invalid items?

I get the use case, but it really seems risky to me, because once this "negative" kind of items is allowed, sooner or later an unwanted and unexpected one would sneak in, without us noticing :/

If I had my way, I would root out empty strings completely, including the `sudo_prefix`, because the emptiness is often misused to track "undefined", "unset", or "unknown" kind of info. Maybe we could turn `sudo_prefix` into something friendlier to other kids... Not sure what though, I'd like to think about this first.

happz: > I mean prefectly valid to avoid the prior `if` check and use the whole thing
> 
> https://github.com/teemtee/tmt/blob/5ba7562c4dd9acd61b0aefa99bfea93e6dce9d12/tmt/package_managers/bootc.py#L95-L98
> 
> Becoming just
> 
> ```python
> command = Command(self.guest.facts.sudo_prefix, 'bootc')
> ```

Got it.

It might seem I'm needlessly paranoid... But I really dislike empty strings being silently dropped; with the history of people (mis)using empty strings, it's practically guaranteed the silence will harm us one day. So I'd like to find something we could verify. Maybe a type for strings that can be empty, and it does not bear any boolean-like meaning, who knows.

thrix (approved):
Makes sense. An empty first element in `Command` would produce `" actual_command"` with a leading space, which is a subtle bug. The `if element` filter correctly skips both `None` and empty strings, neither of which are valid command elements.

Issue #3894, opened 2025-07-21
tmt upgrade execution order issue when there are 2 or more configs in discovery


guoguojenna:
Define 2 configs in discovery of a plan, the actual execution order is: 
```
old-config-1 -> upgrade -> new-config-1; then old-config-2 -> upgrade -> new-config-2
```
this must fail because it runs upgrade 2 times.

The expected order is:
```
old-config-1, old-config-2  -> upgrade -> new-config-1,  new-config-2
```

Any ideas?

Example:
```
summary: iscsi 9to10 upgrade testing
discover:
  - name: sts-prepare
    how: fmf 
    filter:
      -tag:sts-prepare
  - name: iscsi-local
    how: fmf 
    filter:
      - tag:iscsi
      - tag:local
provision:
    image: RHEL-9.7%
execute:
    how: upgrade
    url: https://gitlab.cee.redhat.com/oamg/rhel-major-upgrade
    upgrade-path: 9to10.cut
report:
    how: html

```

guoguojenna:
[log.txt](https://github.com/user-attachments/files/21339216/log.txt)

psss:
Thanks for the issue and the reproducer! I was able to reproduce the problem with this minimal `main.fmf` config:

```yaml
/test:
    test: echo hi

/plan:
    discover:
      - how: fmf
      - how: fmf
    provision:
        how: container
        image: fedora:41
    execute:
        how: upgrade
        url: https://github.com/teemtee/upgrade
        upgrade-path: /paths/fedora41to42
        test: /tasks/prepare
```

Adding to the backlog with the `should` priority.

guazhangRH:
Yes, Hit same issue.

guoguojenna:
> Thanks for the issue and the reproducer! I was able to reproduce the problem with this minimal `main.fmf` config:
> 
> /test:
>     test: echo hi
> 
> /plan:
>     discover:
>       - how: fmf
>       - how: fmf
>     provision:
>         how: container
>         image: fedora:41
>     execute:
>         how: upgrade
>         url: https://github.com/teemtee/upgrade
>         upgrade-path: /paths/fedora41to42
>         test: /tasks/prepare
> 
> Adding to the backlog with the `should` priority.

Sweet, thanks!

tcornell-bus:
+2 for getting this fixed. Two more users ran into it.

psss:
Raising priority to `high` as more and more users are running into this.
Tentatively proposing for the next sprint.

PR #4633, opened 2026-03-02
Fix support for multiple discover phases in upgrade execute plugin

tcornell-bus:
Extract task iterator logic for upgrade plugin into a dedicated tasks method following the same pattern as internal (tmt) execute.

Allow upgrade execute plugin to handle multiple discover phases by iterating through each phase when running tests before and after upgrade.

Fix test name restoration to occur per-phase to avoid iteration order issues.

Assisted-by: Cursor

Fixes: #3894

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -1,4 +1,5 @@
 import abc
+import copy
 import functools
 import itertools
 import signal as _signal
@@ -978,14 +979,15 @@ class ExecutePlugin(tmt.steps.Plugin[ExecuteStepDataT, None]):
             level=3,
         )
 
+    @property
     @abc.abstractmethod
     def tasks(
         self,
-    ) -> Iterator[tuple['ExecutePlugin[ExecuteStepData]', list['Guest']]]:
+    ) -> Iterator[tuple[Optional[str], list['Guest']]]:
         """
         Iterate over tasks to be enqueued for execution.
 
-        :yields: tuple of two items, a plugin phase copy
+        :yields: tuple of two items, a discover phase name
             and the list of guests it should run on.
         """
         raise NotImplementedError
@@ -1218,7 +1220,12 @@ class Execute(tmt.steps.Step):
                 queue.enqueue_action(phase=phase)
 
             elif isinstance(phase, ExecutePlugin):
-                for phase_copy, guests in phase.tasks():
+                for discover_phase_name, guests in phase.tasks:
+                    # For each discover phase, we need a copy of the execute
+                    # plugin, so we could point it to that discover phase rather than
+                    # let it "see" all tests, or test in different discover phase.
+                    phase_copy = cast(ExecutePlugin[ExecuteStepData], copy.copy(phase))
+                    phase_copy.discover_phase = discover_phase_name
                     queue.enqueue_plugin(phase=phase_copy, guests=guests)
 
         failed_tasks: list[Union[ActionTask, PluginTask[ExecuteStepData, None]]] = []
--- a/tmt/steps/execute/upgrade.py
+++ b/tmt/steps/execute/upgrade.py
@@ -1,4 +1,3 @@
-import copy
 from collections.abc import Iterator
 from typing import Any, Optional, Union, cast
 
@@ -12,7 +11,7 @@ import tmt.utils
 from tmt.container import container, field, key_to_option
 from tmt.steps.discover import Discover, DiscoverPlugin, DiscoverStepData, normalize_ref
 from tmt.steps.discover.fmf import DiscoverFmf, DiscoverFmfStepData
-from tmt.steps.execute import ExecutePlugin, ExecuteStepData
+from tmt.steps.execute import ExecutePlugin
 from tmt.steps.execute.internal import ExecuteInternal, ExecuteInternalData
 from tmt.steps.prepare import PreparePlugin
 from tmt.steps.prepare.install import PrepareInstallData
@@ -230,20 +229,31 @@ class ExecuteUpgrade(ExecuteInternal):
     def discover(self, plugin: Optional[DiscoverPlugin[DiscoverStepData]]) -> None:
         self._discover = plugin
 
+    @property
     def tasks(
         self,
-    ) -> Iterator[tuple['ExecutePlugin[ExecuteStepData]', list['tmt.guest.Guest']]]:
-        # upgrade plugin is expected to (potentially) execute multiple
+    ) -> Iterator[tuple[Optional[str], list['tmt.guest.Guest']]]:
+        # upgrade plugin is expected to execute multiple
         # discover phases on old, perform the upgrade, then execute
         # those same discover phases again on new. All of this should occur
-        # in a single task.
-        phase_copy = cast(ExecutePlugin[ExecuteStepData], copy.copy(self))
-        # We want to run all discover phases on old and new, so set discover_phase to None.
-        # This comes with the responsibility to later only run discover phases that are enabled
-        # for that guest, based on discover.enabled_by_when and discover.enabled_on_guest(guest).
-        phase_copy.discover_phase = None
+        # in a single task, so that the upgrade happens only once
+        # (due to how the upgrade plugin is currently structured).
 
-        yield (phase_copy, self.step.plan.provision.ready_guests)
+        # TODO: The logic for this plugin could be simplified if it were refactored to make
+        # the before upgrade self._run_test_phase() an execute task,
+        # the actual upgrade step an execute task,
+        # then the after upgrade self._run_test_phase() an execute task, all specified here.
+
+        # TODO: It would be nice if the discover_phase variable was a list instead of string,
+        # so that we could specify multiple discover phases to run, instead of all
+        # (by setting it to None) or a single specified one.
+        # Then we would not need to do the discover_phase shuffle in self._run_test_phase().
+
+        # Due to the above, we have to set discover_phase to None for now, which tells
+        # self._run_tests() to run all discover phases. This comes with the responsibility
+        # to later only run discover phases that are enabled for that guest, based on
+        # discover.enabled_by_when and discover.enabled_on_guest(guest) in self._run_test_phase().
+        yield (None, self.step.plan.provision.ready_guests)
 
     def go(
         self,
--- a/tests/execute/upgrade/data/plan.fmf
+++ b/tests/execute/upgrade/data/plan.fmf
@@ -14,13 +14,17 @@ execute:
 /no-path:
     summary: Basic upgrade test with no upgrade path
 /path:
-    summary: Basic upgrade test with multiple discover phases
+    summary: Basic upgrade test with upgrade path
+    execute+:
+        upgrade-path: $@{upgrade-path}
+/add-phase:
+    summary: Upgrade test with upgrade path and extra discover phase
+    execute+:
+        upgrade-path: $@{upgrade-path}
     discover+:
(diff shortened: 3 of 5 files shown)


tests/execute/upgrade/main.fmf
```diff
@@ -34,3 +34,12 @@ tier: 4
     tag+:
       - provision-only
       - provision-virtual
+
+/multi-phase-discover:
```
tcornell-bus: Should I add the second discover phase to the `/full` test to save time?

tmt/steps/execute/__init__.py
```diff
@@ -978,6 +977,19 @@ def timeout_hint(self, invocation: TestInvocation) -> None:
             level=3,
         )
 
+    @abc.abstractmethod
+    def gather_tasks(
+        self,
+    ) -> list[tuple['ExecutePlugin[ExecuteStepData]', list['Guest']]]:
```
happz: This could be `Iterator[...]`, and we could get rid of building the actual list:

```python
for discover in self.step.plan.discover.phases(classes=(DiscoverPlugin,)):
    if not discover.enabled_by_when:
        continue

    phase_copy = cast(ExecutePlugin[ExecuteStepData], copy.copy(self))
    phase_copy.discover_phase = discover.name

    yield (
        phase_copy,
        [
            guest
            for guest in self.step.plan.provision.ready_guests
            if discover.enabled_on_guest(guest)
        ],
    )

# And
yield (phase_copy, self.step.plan.provision.ready_guests)
```

The list is needed only to be iterated over, it's never passed anywhere else, never logged, never saved, never printed out. We don't need it.

tmt/steps/execute/upgrade.py
```diff
@@ -227,6 +228,30 @@ def discover(self) -> Union[Discover, DiscoverFmf]:
...
+    def gather_tasks(
+        self,
+    ) -> list[tuple['ExecutePlugin[ExecuteStepData]', list['tmt.guest.Guest']]]:
+        """
+        Return tasks to be enqueued for execution.
+
+        upgrade plugin is expected to (potentially) execute multiple
+        discover phases on old, perform the upgrade, then execute
+        those same discover phases again on new. All of this should occur
+        in a single task.
```
happz: I'd recommend moving this detail about implementation into a comment in the body of this method (same in `internal.py`), and dropping the docstring completely. IDEs and Spinx will use docstring from the parent method.

tests/execute/upgrade/main.fmf
tcornell-bus: addressed in cf99977544706501448f4bb8e7779119146086c0

tmt/steps/execute/upgrade.py
```diff
@@ -469,21 +494,36 @@ def _run_test_phase(self, guest: tmt.guest.Guest, prefix: str, logger: tmt.log.L
...
+                    extra_environment=Environment({STATUS_VARIABLE: EnvVarValue(prefix)}),
+                    logger=logger,
+                )
 
-        self._remove_old_results(prefix)
+                # Restore test names immediately after this phase
+                for test, original_name in test_name_backups:
+                    test.name = original_name
+
+        self.discover_phase = original_discover_phase
```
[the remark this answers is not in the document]
tcornell-bus: addressed in cf99977544706501448f4bb8e7779119146086c0

tmt/steps/execute/upgrade.py
tcornell-bus: addressed in cf99977544706501448f4bb8e7779119146086c0

tmt/steps/execute/__init__.py
tcornell-bus: addressed in cf99977544706501448f4bb8e7779119146086c0

tcornell-bus:
packit /test

tmt/steps/execute/__init__.py
```diff
@@ -1205,29 +1216,9 @@ def go(self, force: bool = False) -> None:
...
-                        phase=phase_copy,
-                        guests=[
-                            guest
-                            for guest in self.plan.provision.ready_guests
-                            if discover.enabled_on_guest(guest)
-                        ],
-                    )
+            elif isinstance(phase, ExecutePlugin):
+                for phase_copy, guests in phase.tasks():
+                    queue.enqueue_plugin(phase=phase_copy, guests=guests)
```
LecrisUT: Suggest to move the copying of the phases inside this part. It is not intuitive to always provide a copy in the `phase.tasks`

tmt/steps/execute/__init__.py
```diff
@@ -978,6 +977,18 @@ def timeout_hint(self, invocation: TestInvocation) -> None:
             level=3,
         )
 
+    @abc.abstractmethod
+    def tasks(
+        self,
+    ) -> Iterator[tuple['ExecutePlugin[ExecuteStepData]', list['Guest']]]:
```
LecrisUT: If we do not do the explicit copy, this can be made a property

tmt/steps/execute/upgrade.py
```diff
@@ -228,6 +230,21 @@ def discover(self) -> Union[Discover, DiscoverFmf]:
...
+    ) -> Iterator[tuple['ExecutePlugin[ExecuteStepData]', list['tmt.guest.Guest']]]:
+        # upgrade plugin is expected to (potentially) execute multiple
+        # discover phases on old, perform the upgrade, then execute
+        # those same discover phases again on new. All of this should occur
+        # in a single task.
+        phase_copy = cast(ExecutePlugin[ExecuteStepData], copy.copy(self))
+        # We want to run all discover phases on old and new, so set discover_phase to None.
+        # This comes with the responsibility to later only run discover phases that are enabled
+        # for that guest, based on discover.enabled_by_when and discover.enabled_on_guest(guest).
+        phase_copy.discover_phase = None
```
LecrisUT: This part lost me. Can you elaborate what is going on here?

tmt/steps/execute/upgrade.py
```diff
@@ -228,6 +230,21 @@ def discover(self) -> Union[Discover, DiscoverFmf]:
     def discover(self, plugin: Optional[DiscoverPlugin[DiscoverStepData]]) -> None:
         self._discover = plugin
 
+    def tasks(
+        self,
+    ) -> Iterator[tuple['ExecutePlugin[ExecuteStepData]', list['tmt.guest.Guest']]]:
```
LecrisUT: I am lost on why this is not a similar iteration over `step.plan.discover.phases`

LecrisUT: Are there plans to expand this? E.g. could the logic of `_run_test_phase(BEFORE_UPGRADE_PREFIX`) -> `_perform_upgrade` -> `_run_test_phase(AFTER_UPGRADE_PREFIX)` be included directly here?

tmt/steps/execute/__init__.py
tcornell-bus: Fair point, if I move the copy out, it would need to return the intended `discover_phase` string as well in the tuple. I don't see a problem with that though

tmt/steps/execute/upgrade.py
tcornell-bus: For the internal.py tmt execute plugin, it does multiple execute tasks; each task is to run a different discover phase. To do that, it copies the whole plugin and uses the `discover_phase` variable to make that copy of the plugin only work on that discover phase.
Example: 5 discovery phases would result in 5 tmt execute tasks, where there are 5 copies of the plugin and each copy points to a different discover_phase.

For the upgrade plugin, it should run each discover phase once before the upgrade, then again after the upgrade. We only want the upgrade part to run _once_. Which is why the task format for tmt execute does not work here.
Example: 5 discovery phases should result in 1 upgrade execute task, where there is 1 copy of the plugin and  it points to None in discover_phase variable because:
```
    #: If set, plugin should run tests only from this discover phase.
    discover_phase: Optional[str] = None
```
Then `self._run_tests()` -> `self.prepare_tests()` -> `self.discover.tests(phase_name=self.discover_phase, enabled=True)`  where the phase_name param is:
```
        :param phase_name: if set, return only tests discovered by the
            phase of this name. Otherwise, tests discovered by all
            phases are returned.
```
The two options are to specify a single discovery phase to run or all the discover phases by setting `discover_phase` to `None`. Further down, I still have it cycle through the discover phases one at a time because we need to filter out disabled phases, but until that point, the best option is to assume `all` discover phases.

I think the `discover_phase` variable idea could be refactored, but it is not a simple refactor.

tmt/steps/execute/upgrade.py
tcornell-bus: I think that could be done, but the upgrade plugin would need a bigger refactor that I was trying to avoid here. I think it would be a good refactor for a housekeeping sprint?

tmt/steps/execute/upgrade.py
happz: > I think the `discover_phase` variable idea could be refactored, but it is not a simple refactor.

It can, it should, but it really isn't easy. I hope to get to it eventually, already stumbled over it in another PR.

tmt/steps/execute/__init__.py
LecrisUT: > it would need to return the intended `discover_phase` string as well in the tuple

Subject to future refactoring of `discover_phase`. Can you add a comment when you do that to indicate what you would have preferred instead so we can do the refactoring of that more easily?

tmt/steps/execute/upgrade.py
LecrisUT: Sure, just figuring out the plans for this would be helpful when coming back to this later on.

tmt/steps/execute/upgrade.py
LecrisUT: > For the upgrade plugin, it should run each discover phase once before the upgrade, then again after the upgrade. We only want the upgrade part to run _once_. Which is why the task format for tmt execute does not work here.

Ah ok. So with the idea of making the upgrade as one of the tasks here, it would actually simplify this quite a bit right?

Ok things are a bit clearer. Just adding some TODO notes on how we can better simplify this logic would be helpful.

tcornell-bus: Added a TODO in 0502fee86b4030ca5dda1a1efad4267251339fbd  and made an issue to refactor discover_phase https://github.com/teemtee/tmt/issues/4640

tmt/steps/execute/upgrade.py
tcornell-bus: I think this was addressed in this comment [thread](https://github.com/teemtee/tmt/pull/4633/files#r2878587997).

tcornell-bus: Added a TODO in 0502fee86b4030ca5dda1a1efad4267251339fbd and filed this issue https://github.com/teemtee/tmt/issues/4641 so it doesn't get forgotten.

tmt/steps/execute/__init__.py
tcornell-bus: Addressed in 0502fee86b4030ca5dda1a1efad4267251339fbd

tmt/steps/execute/__init__.py
tcornell-bus: Addressed in 0502fee86b4030ca5dda1a1efad4267251339fbd

LecrisUT (approved):
Great, thanks a lot for the context and trackers

PR #4630, opened 2026-03-02
Skip known bootc failure for now

LecrisUT:
Thanks to #4248 we can actually be more granular about this now :wink:.

Tracked in https://bugzilla.redhat.com/show_bug.cgi?id=2443825

PR #4634, opened 2026-03-02
Revert #4630

LecrisUT:
Pull Request Checklist

* [ ] Follow up on the AVC issues
  * [x] `type=AVC.*comm=bootupctl`: Probably out of our control, requested help in https://bugzilla.redhat.com/show_bug.cgi?id=2443825
  * [ ] `type=SELINUX_ERR.*oldcontext=system_u:system_r:install_t.*newcontext=system_u:system_r:mount_t`: smells like an issue of our own creation

### DIFF
--- a/tests/provision/bootc/main.fmf
+++ b/tests/provision/bootc/main.fmf
@@ -5,16 +5,3 @@ tag+:
 require:
   - tmt+provision-virtual
 duration: 2h
-
-adjust+:
-  - when: distro > fedora-42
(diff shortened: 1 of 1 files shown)


LecrisUT:
https://github.com/fedora-selinux/selinux-policy/pull/3162 was merged, but the problem still persists. Also the other issue (`type=SELINUX_ERR` failure) has not been explored yet

Issue #4625, opened 2026-03-02
Support Ansible-based feature plugins in image mode


thrix:
Follow up for #4624 to use the same procedure.

thrix:
Exchanged with #4624 in the sprint, we will not fix the genreic ansible support

PR #4636, opened 2026-03-03
Support Ansible prepare step in image mode

### LINKED ISSUE #4624: Implement ansible prepare step support in image mode
Using the following steps:

1. starting the base container using podman
2. connecting to it using container.podman.podman Ansible connection plugin
3. applying the playbook
4. saving the container image
5. rebooting to the new configuration

thrix:
## Summary
- Add image mode (bootc) support for the `ansible` prepare plugin
- When running on a bootc guest, starts a container, runs `ansible-playbook` with `containers.podman.podman` connection plugin, commits the result, switches via `bootc switch`, and reboots
- Includes integration test following the `tests/prepare/bootc/` pattern

## Details

On image mode guests, `/usr` is read-only and system changes must be baked into the container image. The existing shell and install prepare plugins handle this via Containerfile directives, but the ansible plugin had no image mode support.

The implementation:
1. Flushes any pending containerfile directives from prior shell/install phases
2. Ensures `ansible-playbook` is available on the guest (installs `ansible-core` if needed)
3. Gets the current bootc image and ensures it's in podman storage
4. Starts a container from the image
5. Runs each playbook with `ansible-playbook -c containers.podman.podman`
6. Commits the container to a new image
7. Switches via `bootc switch --transport containers-storage` and reboots

## Test plan
- [ ] Run existing prepare tests for regressions
- [ ] Run new `tests/prepare/bootc-ansible/` test with an image mode guest
- [ ] Verify log output shows: container start, ansible-playbook with podman connection, commit, switch, reboot

Closes #4624


tmt/steps/prepare/ansible.py
```diff
@@ -173,6 +182,273 @@ def _preserved_workdir_members(self) -> set[str]:
...
+        ansible-core via the bootc package manager, which triggers a
+        container image build, bootc switch, and reboot.
+        """
+
+        try:
+            guest.execute(Command('which', 'ansible-playbook'), silent=True)
+            logger.debug('ansible-playbook is already available on the guest.')
+        except tmt.utils.RunError:
+            logger.info('ansible', 'installing ansible-core on the guest', 'green')
+            guest.package_manager.install(Package('ansible-core'))
```
LecrisUT, line 203: Why is ansible required on the guest?

tmt/steps/prepare/ansible.py
```diff
@@ -173,6 +182,273 @@ def _preserved_workdir_members(self) -> set[str]:
...
+            if not current_image.startswith(LOCALHOST_BOOTC_IMAGE_PREFIX):
+                logger.debug(f'Ensuring image {current_image} is in podman storage.')
+                guest.execute(
+                    ShellScript(
+                        f'{guest.facts.sudo_prefix} /bin/bash -c "('
+                        f'  ( podman pull {current_image}'
+                        f'    || podman pull containers-storage:{current_image} )'
+                        f'  || bootc image copy-to-storage --target {current_image}'
+                        ')"'
+                    )
```
LecrisUT, line 339: Why the indirect bash

tmt/steps/prepare/ansible.py
```diff
@@ -173,6 +182,273 @@ def _preserved_workdir_members(self) -> set[str]:
...
+                        raw_playbook, pb_index, guest, logger
+                    )
+
+                    # Build the ansible-playbook command to run on the guest
+                    ansible_cmd = (
+                        f'{guest.facts.sudo_prefix} ansible-playbook'
+                        f' -c containers.podman.podman'
+                        f" -i '{container_name},'"
+                        f' {guest_playbook}'
+                    )
```
LecrisUT, line 382: :-1: to this approach. It uses both a `containers` connection and it also runs on the guest!?

LecrisUT (commented):
I do not like this approach. Please elaborate why you went with this and what other alternatives have you considered.

tmt/steps/prepare/ansible.py
thrix: @LecrisUT this should happen on the guest to:

1. mitigate having an external container registry in place to share the images (or find some other crazy way how to get the base image from the guest and push it to the guest). The base container image is available only on the guest

2. Be similar to a regular container rebuild, less suprises, easier to explain how it works

LecrisUT: > mitigate having an external container registry

The local container registry of the bootc runtime should be usable right?

> Be similar to a regular container rebuild, less suprises

I disagree with this creating less surprises. E.g. if the package being tested is anything inside the ansible dependency stack, and this fails at the prepare step, how would one debug this? Having a minimal dependency footprint would be the best for the "less surprises"

---

Before anything, I would like a better breakdown of any other alternative approaches. From what I understand of bootc it is supposed to be possible to do local builds from the current state, don't know if `bootc image copy-to-storage`, but there seem to be many ways of interacting with bootc and would like to have a more thorough breakdown of it

thrix: > The local container registry of the bootc runtime should be usable right?

it has nothing to do with bootc runtime, it is the guest local registry on the machine

> I disagree with this creating less surprises. E.g. if the package being tested is anything inside the ansible dependency stack, and this fails at the prepare step, how would one debug this? Having a minimal dependency footprint would be the best for the "less surprises"

Not sure I follow, the failed preparation should be visible from the tmt logs?

> Before anything, I would like a better breakdown of any other alternative approaches. From what I understand of bootc it is supposed to be possible to do local builds from the current state,

Local to what? guest or host?

> don't know if bootc image copy-to-storage, but there seem to be many ways of interacting with bootc and would like to have a more thorough breakdown of it

I am not aware of any alternative approaches that would not require what I wrote

thrix:
It was decided we will not support Ansible for now, only fix the playbooks for feature plugins to be image-mode friendly.

thrix:
As discussed, closing this for now, we will go with #4625 by adjusting the playbooks to not use tasks which will not work in image mode (e.g. dnf)

PR #4638, opened 2026-03-03
Add `virtualization.confidential` hardware requirement

### LINKED ISSUE #4637: Add `virtualization.confidential` hardware requirement
## Summary

Add a new boolean hardware requirement `virtualization.confidential` to specify that a guest should be provisioned as a confidential VM with a hardware-based Trusted Execution Environment (TEE).

## Motivation

Confidential VMs protect data while in use through hardware-level encryption and isolation. The underlying technologies include:

- **AMD SEV-SNP** (Secure Encrypted Virtualization - Secure Nested Paging)
- **Intel TDX** (Trust Domain Extensions)
- **ARM CCA** (Confidential Compute Architecture)

All major cloud providers support this: Azure Confidential VMs, GCP Confidential VMs, and AWS Nitro. This is inherently a virtualization feature — it protects the VM from the hypervisor — and fits naturally alongside the existing `virtualization.is-virtualized` and `virtualization.is-supported` flags.

## Proposed usage

```yaml
virtualization:
    confidential: true
```

## Implementation scope

- Add `confidential: true|false` to the `virtualization` hardware spec
- Add parsing support in `tmt/hardware.py`
- Add JSON schema support
- Update tests

No provisioner (Beaker/mrack) transformer is included initially, consistent with how `virtualization.is-supported` is handled today.

## Links

https://issues.redhat.com/browse/TFT-4391

thrix:
## Summary

Add a new boolean hardware requirement `virtualization.confidential` for specifying that a guest should be provisioned as a confidential VM with a hardware-based Trusted Execution Environment (TEE).

Confidential VMs use technologies like AMD SEV-SNP, Intel TDX, or ARM CCA to protect data while in use by encrypting VM memory from the hypervisor. All major cloud providers (Azure, GCP, AWS) support this.

### Changes

- **Spec**: Added `confidential: true|false` to `spec/hardware/virtualization.fmf`
- **Parser**: Added `'confidential'` to the flag constraints in `_parse_virtualization()`
- **Schema**: Added `confidential` boolean property to the virtualization block in `hardware.yaml`
- **Tests**: Updated `FULL_HARDWARE_REQUIREMENTS` and expected output in `test_hardware.py`

No provisioner transformer is included initially, consistent with how `virtualization.is-supported` is handled.

Closes #4637 [TFT-4391](https://issues.redhat.com/browse/TFT-4391)

Assisted-by Claude Code

## Pull Request Checklist

* [x] implement the feature
* [x] modify the json schema
* [x] mention the version
* [x] include a release note

### DIFF
--- a/tests/unit/provision/mrack/test_hw.py
+++ b/tests/unit/provision/mrack/test_hw.py
@@ -232,6 +232,7 @@ def test_maximal_constraint(root_logger: Logger) -> None:
                 'and': [
                     {'system': {'hypervisor': {'_op': '==', '_value': ''}}},
                     {},
+                    {},
                     {'system': {'hypervisor': {'_op': 'like', '_value': 'xen'}}},
                 ]
             },


spec/hardware/virtualization.fmf
```diff
@@ -1,6 +1,9 @@
 summary: |
     Select or provision a guest with requested virtualization properties.
 
+    .. versionadded:: 1.42
+       ``virtualization.confidential``
```
LecrisUT: In this case is the support automatically added for both `beaker` and `artemis` and it is just waiting on the support from that side? Is it the same case for the other 2? In one way or another these notes need adjustment

docs/releases/pending/4637.fmf
```diff
@@ -0,0 +1,6 @@
+description: |
+    Introduced the :ref:`/spec/hardware/virtualization` hardware
+    requirement ``virtualization.confidential`` to specify that a guest
+    should be provisioned as a confidential VM with a hardware-based
+    Trusted Execution Environment (TEE) such as AMD SEV-SNP, Intel TDX,
+    or ARM CCA.
```
LecrisUT: I would say, cut it off at "confidential VM" and add an external link. Could probably add that in the spec fmf file as well.

docs/releases/pending/4637.fmf
```diff
@@ -0,0 +1,6 @@
+description: |
+    Introduced the :ref:`/spec/hardware/virtualization` hardware
```
tcornell-bus: This wasn't rendered right, I think it needs to be:
```suggestion
    Introduced the :tmt:story:`/spec/hardware/virtualization` hardware
```

thrix: ty!

spec/hardware/virtualization.fmf
thrix: No, `confidential` support is not automatically added to provisioners - it needs explicit transformer support on their side. Same applies to `is-supported` which also has no provisioner implementation.

The `implemented-by` notes (updated in this PR for both artemis and mrack) explicitly state they only support `is-virtualized` and `hypervisor`, making it clear that `confidential` and `is-supported` are not implemented.

The `versionadded` note follows the existing pattern used across other hardware specs (e.g., `device.fmf`, `gpu.fmf`, `zcrypt.fmf`) - just the version and the key name, no provisioner details. Provisioner support gets documented via `versionchanged` when it lands.

Is this sufficient or would you like a specific adjustment?

thrix: @LecrisUT PTAL pls

LecrisUT: > The `implemented-by` notes (updated in this PR for both artemis and mrack) explicitly state they only support `is-virtualized` and `hypervisor`, making it clear that `confidential` and `is-supported` are not implemented.

That is super confusing. That does not fit in with the [hardware table](https://tmt.readthedocs.io/en/stable/plugins/provision.html#hardware-requirement-support) which shows that `artemis` supports the whole `virtualization`, yet I do not see any explicit reference to `is-virtualized` or `hypervisor` or anything in that module.

> The `versionadded` note follows the existing pattern used across other hardware specs (e.g., `device.fmf`, `gpu.fmf`, `zcrypt.fmf`) - just the version and the key name, no provisioner details. Provisioner support gets documented via `versionchanged` when it lands.

Context matters here. In the current form it reads to me that `virtualization.confidential` is added and all providers would support it.

thrix:
@tcornell-bus kindly asking for re-review

spec/hardware/virtualization.fmf
thrix: Fair point. I've updated the `versionadded` note to explicitly say "(not yet supported by any provisioner plugin)" which should address the immediate ambiguity.

Regarding the hardware table confusion - the table is auto-generated from the spec fmf files (`docs/ext/generate_hardware_matrix.py`), one row per `/spec/hardware/*.fmf` file. To properly show per-sub-requirement support, we'd need to split `virtualization.fmf` into a directory with separate child nodes (`hypervisor`, `is-virtualized`, `is-supported`, `confidential`), each with their own `implemented-by` links. The table would then show exactly which provisioner supports which sub-requirement without relying on footnotes.

That's a bigger structural change though. Would you prefer:
1. Just the `versionadded` clarification in this PR, with the table split as a follow-up issue?
2. The full spec split done in this PR?

LecrisUT: > Fair point. I've updated the `versionadded` note to explicitly say "(not yet supported by any provisioner plugin)" which should address the immediate ambiguity.

Also fine with not having it mentioned as `.. versionchanged` at all since it is a no-op.

Table should be something to be revisited later when there is a clearer vision on how to improve the look. I would just inject single table rows inside the hardware stories.

happz:
@thrix  please, set the "Size" of this PR.

spec/hardware/virtualization.fmf
thrix: Thanks. I pushed the `versionadded` clarification and also added a footnote to the hardware support table via the generator (`docs/ext/generate_hardware_matrix.py`). I'd prefer to keep the `versionadded` so users know when the spec key was introduced, just with the explicit "not yet supported" qualifier.

Table improvements tracked in #4666.

thrix:
> @thrix please, set the "Size" of this PR.

@happz done, thx

spec/hardware/virtualization.fmf
```diff
@@ -1,6 +1,10 @@
 summary: |
     Select or provision a guest with requested virtualization properties.
 
+    .. versionadded:: 1.42
+       ``virtualization.confidential`` (not yet supported by any
+       provisioner plugin)
```
psss, line 6: Hm, `virtualization.confidential` added in `1.42`?

psss (commented):
Looks ok, just the `versionadded` confusion.

spec/hardware/virtualization.fmf
```diff
@@ -1,6 +1,10 @@
 summary: |
     Select or provision a guest with requested virtualization properties.
 
+    .. versionadded:: 1.42
```
thrix:
```suggestion
    .. versionadded:: 1.70
```

LecrisUT:
@thrix the PR is failing at
> FAILED provision/mrack/test_hw.py::test_maximal_constraint - AssertionError: ...

Moving it back to implement for now

LecrisUT:
rawhide test failure is a known issue fixed with a rebase, but the UI did not allow me to rebase

PR #4650, opened 2026-03-06
Add a glossary section

LecrisUT:
This serves as a place where we can dump all definitions of the terms that we use across the project. During the documentation, this can then be consumed with the `:term:` role giving a reference link to the definition of the term. If we combine this with [sphinx-tippy](https://sphinx-tippy.readthedocs.io/en/latest/index.html) we get a nice pop-up with the contents.

Currently the glossary is not meant to be alphabetically ordered to be able to group together similarly related terms. We should also be able to use multiple `.. glossary` and group them via sections if a meaningful separation is available

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -1,0 +1,16 @@
+========
+Glossary
+========
+
+Here you can find some common terms used throughout the tmt project.
+
+Machines
+--------
+
+.. glossary::
+
+   guest
+      The machine where the tests are being run.
+
+   runner
+      The machine that is running the ``tmt`` command.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -35,5 +35,5 @@ Table of Contents
     Questions <questions>
     Contribute <contribute>
     Code <code/index>
-    Releases <releases/index>
     glossary
+    Releases <releases/index>
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -36,3 +36,4 @@ Table of Contents
     Contribute <contribute>
     Code <code/index>
     Releases <releases/index>
+    glossary


docs/glossary.rst
```diff
@@ -0,0 +1,13 @@
+========
+Glossary
+========
+
+Here you can find some common terms used throughout the tmt project.
+
+.. glossary::
+
+   runner
```
happz, line 15: Can the entries be sorted in a more visible way, e.g. lexicographically?

LecrisUT: > Currently the glossary is not meant to be alphabetically ordered to be able to group together similarly related terms. We should also be able to use multiple .. glossary and group them via sections if a meaningful separation is available

Partially, I would still prefer to have similar concepts to be close to each other first. The fully alphabetical list should be automatically generated from the index, let me find the link for that

happz: > > Currently the glossary is not meant to be alphabetically ordered to be able to group together similarly related terms. We should also be able to use multiple .. glossary and group them via sections if a meaningful separation is available
> 
> Partially, I would still prefer to have similar concepts to be close to each other first. The fully alphabetical list should be automatically generated from the index, let me find the link for that

Can the entries of these "concept groups" be sorted in a more visible way, e.g. lexicographically?

LecrisUT: > Can the entries of these "concept groups" be sorted in a more visible way, e.g. lexicographically?

Yep, that's fine, just need to find appropriate wording for the groups

Hmm, we do not seem to have a global glossary term index https://github.com/sphinx-doc/sphinx/issues/14331 :/

thrix (approved):
LGTM, two minor things to address before merging:

1. Missing trailing period on the "guest" definition (line 13) — "runner" has one, "guest" doesn't.
2. Lexicographic ordering within concept groups as discussed with @happz.

Nothing else found.

Generated-by: Claude Code

docs/glossary.rst
LecrisUT: Name is hard. I went with naming this group as `Machines` for now

happz (approved):

docs/index.rst
```diff
@@ -36,3 +36,4 @@ Table of Contents
     Contribute <contribute>
     Code <code/index>
     Releases <releases/index>
+    glossary
```
psss, line 38: What about moving `Glossary` one item up? It's kind of related to the `Code` section and it is nice to have `Releases` at the very end, as they are kinda special.

psss (commented):
Looks good, just one suggestion.

docs/index.rst
LecrisUT: Can we postpone this for https://github.com/teemtee/tmt/pull/4651?

psss: That one is not included in the current sprint so it will take couple weeks to complete. Here it is just one line switch so I'd vote to address is directly.

happz: Isn't glossay mainly for tmt users, rather than developers? I'd say it's not as related to `Code` as it is to `Guide` or `Specification`, providing short explanations of terms used heavily in these sections to people not that familiar with the terminology.

LecrisUT: Oh, I thought it was in the current sprint. Well, the current position is rather arbitrary in here, so sure I will move it. Let's focus on #4651 on where to actually put it though. If it does have enough approvals I will just add it ad-hoc (maybe with a distinct label to un-count the size)

LecrisUT: All ok with b4833be?

psss: Yeah, no need to dive deeper here. Let's finish the areas in #4651.

psss (approved):

LecrisUT:
Documentation only change and everything looks fine, merging

PR #4651, opened 2026-03-06
Separate the TOC sections

LecrisUT:
Hopefully this helps with navigability a bit

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -24,16 +24,33 @@ Table of Contents
 ==================
 
 .. toctree::
-    :maxdepth: 2
+    :maxdepth: 1
+    :caption: About
 
     Overview <overview>
+    Releases <releases/index>
+    Repository <https://github.com/teemtee/tmt>
+
+.. toctree::
+    :maxdepth: 1
+    :caption: User
+
     Guide <guide>
+    Examples <examples>
+    Questions <questions>
+
+.. toctree::
+    :maxdepth: 1
+    :caption: Reference
+
     Specification <spec>
     Plugins <plugins/index>
-    Examples <examples>
-    Stories <stories>
-    Questions <questions>
-    Contribute <contribute>
-    Code <code/index>
     glossary
-    Releases <releases/index>
+
+.. toctree::
+    :maxdepth: 1
+    :caption: Contributor
+
+    Contribute <contribute>
+    Stories <stories>
+    Code <code/index>
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -24,7 +24,7 @@ Table of Contents
 ==================
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
     :caption: About
 
     Overview <overview>
@@ -32,7 +32,7 @@ Table of Contents
     Repository <https://github.com/teemtee/tmt>
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
     :caption: User
 
     Guide <guide>
@@ -40,7 +40,7 @@ Table of Contents
     Questions <questions>
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
     :caption: Reference
 
     Specification <spec>
@@ -48,7 +48,7 @@ Table of Contents
     glossary
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
     :caption: Contributor
 
     Contribute <contribute>


vaibhavdaren (approved):

happz:
Why not, although I'd wait for few more approvals, the actual grouping might be a point of discussion.

psss:
Copy pasting the structure brainstorm from #4704:

- How it works under the hood
  - The basic split/flow of `user tree` + (user config) -> `run` -> `step` -> `phase` -> `test`
  - The various trees and directories
  - The split between guest
  - Plugins
  - (Recipe eventually)
  - Glossary
  - Advance topics:
    - Phases that create other phases
    - Profiles/Policies
    - Phases that jump across steps, e.g. `dist-git` discover and `upgrade`
- What feature is available right now/How to use `tmt`
  - Specification/Features what you can do: multi-host, restart, hardware, etc.
  - External plugins
  - Examples and guides
  - Projects that use tmt and their feedback
- Code details
  - Python API
  - Config reference/index

LecrisUT:
> Copy pasting the structure brainstorm from #4704:
> 
>     * How it works under the hood
>       
>       * The basic split/flow of `user tree` + (user config) -> `run` -> `step` -> `phase` -> `test`
>       * The various trees and directories
>       * The split between guest
>       * Plugins
>       * (Recipe eventually)
>       * Glossary
>       * Advance topics:
>         
>         * Phases that create other phases
>         * Profiles/Policies
>         * Phases that jump across steps, e.g. `dist-git` discover and `upgrade`
> 
>     * What feature is available right now/How to use `tmt`
>       
>       * Specification/Features what you can do: multi-host, restart, hardware, etc.
>       * External plugins
>       * Examples and guides
>       * Projects that use tmt and their feedback
> 
>     * Code details
>       
>       * Python API
>       * Config reference/index

I think this would still be useful to consider and refine, and I think we can refine it with the context of this PR. For this PR I want this to focus on just ordering the documents that we have

docs/index.rst
```diff
@@ -24,16 +24,33 @@ Table of Contents
...
     Specification <spec>
     Plugins <plugins/index>
-    Examples <examples>
     Stories <stories>
+
+.. toctree::
+    :maxdepth: 1
+    :caption: Guide
+
+    Guide <guide>
```
AthreyVinay, line 38: Nitpick (with my limited knowledge): `Guide` entry under `Guide` section looks redundant... no?

LecrisUT: Naming suggestions welcome :slightly_smiling_face: (and main purpose for review process? :slightly_smiling_face:)

AthreyVinay: What about `User Guide`? or something like `Getting Started`? Or what if we remove the entry and show this on the parent?

LecrisUT: > Or what if we remove the entry and show this on the parent?

Hmm, I don't think it's possible. Sphinx has `Self`, but I think that does the opposite (create an entry pointing to the current document, compared to removing the entry and moving it to the `caption`?). It might also be a bit confusing since it would not be obvious that that caption is clickable compared to the others (or compared to typical sphinx docs)

> What about `User Guide`? or something like `Getting Started`?

I think `Getting Started` fits well for the current contents. Let's keep the discussion open for other's visibility

AthreyVinay: > Hmm, I don't think it's possible. 

Hence my limited knowledge :) Happy to keep this open!

AthreyVinay (approved):

AthreyVinay:
Lgtm apart from the test failures.

LecrisUT:
> Lgtm apart from the test failures.

Oh, thanks did not notice

docs/getting-started.rst
```diff
@@ -1,7 +1,7 @@
 .. _guide:
 
 ======================
-    Guide
+Getting started
```
psss: I'm not sure about this rename: The scope of the Guide is not just [The First Steps](https://tmt.readthedocs.io/en/stable/guide.html#the-first-steps), which is actually the first section but:

    This guide will show you the way through the dense forest of
    available tmt features, commands and options...

    ... we will slowly dive deeper and deeper so that you don’t
    miss any essential functionality

I think `Guide` fits better to cover this broader scope.

docs/index.rst
```diff
@@ -24,16 +24,33 @@ Table of Contents
...
+
+    Repository <https://github.com/teemtee/tmt>
+    Releases <releases/index>
+    glossary
     Contribute <contribute>
+
+.. toctree::
+    :maxdepth: 1
+    :caption: API docs
+
```
psss, line 53: Grouping the relevant sections together nicely improves the readability, good idea! Looking at the actual grouping I have some suggestions:

* `Documentation` sounds very generic and covers basically whole site, it's all documentation after all
* `About` could also be a good starting / kick-off section
* `Stories` are sometimes confusing users as they also cover ideas not implemented yet, they are kind of list of requirements, probably fit the developer are a bit more?
* `Repository` not sure if the repository link really needs / deserves to take up space in the main menu

What about creating groups by the target audience? Plus the "intro" section? It could look like this:

    ABOUT

    Overview
    Specification
    Glossary
    Releases

    USER

    Guide
    Plugins
    Examples
    Questions

    CONTRIBUTOR

    Contribute
    Stories
    Code

The sections would cover these areas:

* **About** ... What is this project about? How it works? What's new?
* **User** ... How can I use it? Who is using this? Give me some inspiration!
* **Contributor** ... How can I contribute to the project?

One more thought: What about having a `Reference` section group for all reference like materials?

    ABOUT

    Overview
    Releases

    USER

    Guide
    Examples
    Questions

    REFERENCE

    Specification
    Plugins
    Glossary

    CONTRIBUTOR

    Contribute
    Stories
    Code

What do you think?

psss (changes requested):
Very nice idea to group the sections! I'm proposing a couple changes to the actual grouping.

docs/index.rst
LecrisUT: I like the split by the audience, and the second variant looks closest.

> One more thought: What about having a Reference section group for all reference like materials?

:+1: and I would put `Code` in there as well for now. Yes we have some interesting guide-like section about the `steps` and `phases`, but that should be rehomed and split between the user and contributor. Otherwise I like the split for the audience

Looking at https://github.com/teemtee/tmt/pull/4651#issuecomment-4089008573 I think I can fit most of the stuff in these categories as well.

> `Stories` are sometimes confusing users as they also cover ideas not implemented yet, they are kind of list of requirements, probably fit the developer are a bit more?

Can make it splittable by implement state. But the `Stories` name I think is rather confusing, like `user/stories` and `contributor/stories` bring to mind more like testimonies/reviews.

> `Repository` not sure if the repository link really needs / deserves to take up space in the main menu

I would prefer if it does, and any other relevant quick links. I often land on documentation sites like https://www.attrs.org/en/stable/ and the `Repository` there is quite useful for navigation. In those places though the `About` is more near the bottom. I think we could split the `About`, do you have in mind another name that could be used to group those? Or we can make `Overview` be a lone one without a section since that is the only thing from there that should be highlighted

Btw, the `index.rst` is not anywhere in the toctrees, is there a place where we can add it, or make `overview` be the landing page?

psss: > I would put `Code` in there as well for now.

The `Code` area contains developer-only material, wouldn't it fit more `Contribute`?

> Can make it splittable by implement state. But the `Stories` name I think is rather confusing, like `user/stories` and `contributor/stories` bring to mind more like testimonies/reviews.

No better name comes to my mind, perhaps `Ideas`? But these are really `tmt stories` so... dunno.

> I would prefer if it does, and any other relevant quick links. I often land on documentation sites like https://www.attrs.org/en/stable/ and the `Repository` there is quite useful for navigation.

Hm, what about adding it to `About` then?

```
ABOUT

Overview
Releases
Repository
```

> Btw, the `index.rst` is not anywhere in the toctrees, is there a place where we can add it, or make `overview` be the landing page?

I wouldn't put `overview` as the landing page, it's a bit too long. What would be the benefit of putting `index.rst` into toctrees?

LecrisUT: > > I would put `Code` in there as well for now.
> 
> The `Code` area contains developer-only material, wouldn't it fit more `Contribute`?

I am thinking about the `autodoc` which is a code reference rather than contribute. There is one section about the split of steps, and phases which would probably be better in `About`. I would like to move at least `autodoc` out of the contribute, but I think that needs a different set of changes, so I will postpone it.

> > Btw, the `index.rst` is not anywhere in the toctrees, is there a place where we can add it, or make `overview` be the landing page?
> 
> I wouldn't put `overview` as the landing page, it's a bit too long. What would be the benefit of putting `index.rst` into toctrees?

Nothing really, just completion and if it fits anywhere

> Hm, what about adding it to `About` then?

For lack of better idea, let's just keep `About` like that and at the top, and see if we get new ideas later on?

docs/getting-started.rst
LecrisUT: Reverted

docs/index.rst
LecrisUT: @vaibhavdaren @AthreyVinay the reviews were reset due to significant change. Please see :point_up: discussion

psss: Thanks!

docs/index.rst
```diff
@@ -24,16 +24,33 @@ Table of Contents
...
-    :maxdepth: 2
+    :maxdepth: 1
+    :caption: About
 
     Overview <overview>
+    Releases <releases/index>
+    Repository <https://github.com/teemtee/tmt>
+
+.. toctree::
+    :maxdepth: 1
```
psss:
```suggestion
    :maxdepth: 2
```

Can we raise the `maxdepth` here and for the following sections? I find it quite useful that on the landing page it's possible to quickly `ctrl-f` for keywords one is interested in. The first section should stay with `1` as the list of releases is too long and not interesting.

psss (approved):
Looks good, just a small suggestion.

docs/index.rst
LecrisUT: Sure

happz (approved):

happz:
Documentation-only, merging.

PR #4652, opened 2026-03-06
Revert "Do not check for `journal.xml` presence, temporarily (#4648)"

happz:
This reverts commit f2510a3552c533a492891979cbecfe54854486d9.


thrix (approved):
Straightforward revert of the temporary workaround. Re-enables journal.xml presence checks.

PR #4653, opened 2026-03-06
Consolidate installation logic into package managers

### LINKED ISSUE #4605: Move installer logic to package managers
@happz summary from a slack thread:
The `InstallBase` class and its subclasses in `tmt/steps/prepare/install.py` should be removed. Their functionality should be split between the `PrepareInstall` plugin itself and the respective package manager implementations.

AthreyVinay:
resolves #4605

### DIFF
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -35,7 +35,6 @@ import tmt
 import tmt.hardware
 import tmt.log
 import tmt.package_managers
-import tmt.package_managers.bootc
 import tmt.steps
 import tmt.steps.scripts
 import tmt.utils
@@ -2814,6 +2813,8 @@ class GuestSsh(Guest, CommandCollector):
         Uses the same environment and cwd handling as regular execute().
         """
 
+        import tmt.package_managers.bootc
+
         sourced_files = sourced_files or []
 
         if not self.facts.is_image_mode or not isinstance(
@@ -2851,6 +2852,8 @@ class GuestSsh(Guest, CommandCollector):
     def has_collected_commands(self) -> bool:
         """Check if there are collected commands to be applied."""
 
+        import tmt.package_managers.bootc
+
         if not self.facts.is_image_mode or not isinstance(
             self.package_manager, tmt.package_managers.bootc.Bootc
         ):
@@ -2864,6 +2867,9 @@ class GuestSsh(Guest, CommandCollector):
 
         Delegates to the bootc package manager's build_container() method.
         """
+
+        import tmt.package_managers.bootc
+
         if not self.facts.is_image_mode or not isinstance(
             self.package_manager, tmt.package_managers.bootc.Bootc
         ):


tmt/steps/prepare/install.py
```diff
@@ -973,6 +168,158 @@ class PrepareInstall(tmt.steps.prepare.PreparePlugin[PrepareInstallData]):
...
+                )
+
+                failed_packages.update(
+                    str(package).removesuffix("-debuginfo")
+                    for package, present in presence.items()
+                    if not present
+                )
+
+            if failed_packages:
+                self._show_failed_packages_with_tests(failed_packages)
```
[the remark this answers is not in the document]
AthreyVinay: I think this always existed. The original `InstallBootc` only worked because it explicitly cast to Bootc and called `build_container() ` directly, which was tightly coupled. For this pr, I can make `finalize_installation` return `Optional[CommandOutput]` and capture it in `_install()` & bootc `skip_missing` detection can be addressed separately.

tmt/package_managers/bootc.py
```diff
@@ -335,7 +335,51 @@ def install_debuginfo(
         self,
         *installables: Installable,
         options: Optional[Options] = None,
-    ) -> CommandOutput:
+    ) -> None:
```
skycastlelily: shouldn't all the the return type of install_debuginfo functions be the same?

tmt/steps/prepare/install.py
```diff
@@ -1068,70 +416,45 @@ def go(
         if self.is_dry_run:
             return outcome
 
+        self.guest = guest
```
happz: No, `self.guest` cannot exist. `prepare/install` may and is invoked on multiple guests, at the same time - that's why the guest to operate on is a parameter of this method. `self.guest` assignment will result in phases being applied to a mess of different guests.

tmt/package_managers/__init__.py
```diff
@@ -343,3 +343,48 @@ def create_repository(self, directory: Path) -> CommandOutput:
...
+        self,
+        *installables: Installable,
+        options: Optional[Options] = None,
+    ) -> Optional[CommandOutput]:
+        """
+        Install packages stored on a remote URL
+        """
+        return self.install(*installables, options=options)
+
+    def enable_copr(self, repositories: list[str]) -> None:
```
happz: How about `*repositories: str`?

tmt/steps/prepare/install.py
AthreyVinay: Yeah - makes sense too.

Fixed: [3ccb1bc](https://github.com/teemtee/tmt/pull/4653/commits/3ccb1bce578dc7c550a9f61ca127a0f5c0afb0dc)

tmt/package_managers/__init__.py
```diff
@@ -343,3 +343,48 @@ def create_repository(self, directory: Path) -> CommandOutput:
...
+        Enable requested copr repositories
+        """
+        if repositories:
+            raise PrepareError(
+                f"Package manager '{self.NAME}' does not support enabling COPR repositories."
+            )
+
+    def finalize_installation(self) -> Optional[CommandOutput]:
+        """
+        Perform any post-installation steps. Currently, only container image builds are supported.
```
happz: In a base class, I don't care what particular implementation exist. The method is here to perform something after installation, and that's all I need to know when looking at its documentation.

tmt/package_managers/__init__.py
AthreyVinay: Fixed: [111d845](https://github.com/teemtee/tmt/pull/4653/commits/111d8458f2b01fc1cc56607ae92fb43aea6c9108)

tmt/package_managers/__init__.py
AthreyVinay: Fixed: [3217158](https://github.com/teemtee/tmt/pull/4653/commits/32171588383e129861f4ce6828532c66a99167b3)

tmt/package_managers/__init__.py
```diff
@@ -306,7 +306,7 @@ def install_debuginfo(
         self,
         *installables: Installable,
         options: Optional[Options] = None,
-    ) -> CommandOutput:
+    ) -> Optional[CommandOutput]:
```
LecrisUT: Do we really need this change. Would prefer to always provide a command and if it was requested on something that it was not supported, it should raise an error.

Similar for the cases below.

tmt/package_managers/apk.py
```diff
@@ -191,5 +191,34 @@ def install_debuginfo(
...
+        options = options or Options()
+        return self.install(
+            *installables,
+            options=Options(
+                excluded_packages=options.excluded_packages,
+                skip_missing=options.skip_missing,
+                allow_untrusted=True,
+                check_first=False,
+            ),
+        )
```
LecrisUT: This handling of options seems superfluous if you are going to override. Could instead unfreeze the options.

tmt/steps/prepare/install.py
```diff
@@ -1071,67 +418,40 @@ def go(
...
-            dependencies=self.data.package,
-            directories=self.data.directory,
-            exclude=self.data.exclude,
+        options = Options(
+            excluded_packages=[Package(p) for p in self.data.exclude],
+            skip_missing=self.data.missing == 'skip',
         )
 
+        if not self.data.package and not self.data.directory:
+            self.debug("No packages for installation found.", level=3)
```
LecrisUT: Seems weird that this is printed but things are continuing to execute. Can you move this after the enable_copr so that you can return right away after printing this?

tmt/package_managers/__init__.py
LecrisUT: Would `*repositories` be preferable over `Iterable[str]` though :thinking:

tmt/steps/prepare/install.py
```diff
@@ -1071,67 +418,40 @@ def go(
...
+            # to whatever the code may raise, and therefore preserve the information attached
+            # to the original exception.
+            self.warn('Installation failed, trying again after metadata refresh.')
 
-            if failed_packages:
-                self._show_failed_packages_with_tests(failed_packages)
+            try:
+                # Refresh cache in case of recent but not updated change to repodata
+                guest.package_manager.refresh_metadata()
+                self._install(guest, options)
```
LecrisUT, line 448: This could also be moved to be handled by the PackageManager. Maybe in a subsequent task

tmt/steps/prepare/install.py
```diff
@@ -1071,67 +418,40 @@ def go(
...
+            except Exception as exc2:
+                if isinstance(exc2, tmt.utils.RunError):
+                    failed_packages = self._extract_failed_packages_from_outputs(
+                        [exc2.output], guest
+                    )
+                    if failed_packages:
+                        self._show_failed_packages_with_tests(failed_packages)
+                raise exc2 from exc1
```
LecrisUT, line 456: Similarly, this should also be owned by the PackageManager

tmt/steps/prepare/install.py
```diff
@@ -973,6 +168,158 @@ class PrepareInstall(tmt.steps.prepare.PreparePlugin[PrepareInstallData]):
...
+        self.local_packages: list[PackagePath] = []
+        self.remote_packages: list[PackageUrl] = []
+        self.debuginfo_packages: list[Package] = []
+
+        # Detect local, debuginfo and repository packages
+        for dependency in dependencies:
+            if re.match(r"^http(s)?://", dependency):
+                self.remote_packages.append(PackageUrl(dependency))
+            elif dependency.endswith(".rpm"):
+                self.local_packages.append(PackagePath(dependency))
```
LecrisUT, line 191: This highlights that this logic needs to move inside the package manager. Fine if `_prepare_install_local` becomes a method of the package manager.

I am fine with this being deferred.

tmt/steps/prepare/install.py
```diff
@@ -973,6 +168,158 @@ class PrepareInstall(tmt.steps.prepare.PreparePlugin[PrepareInstallData]):
...
+            local_packages = [
+                PackagePath(self.package_directory / p.name) for p in self.local_packages
+            ]
+            if output := guest.package_manager.install_local(
+                *self._list_installables('local package', *local_packages),
+                options=options,
+            ):
+                install_outputs.append(output)
+            summary = fmf.utils.listed([str(p) for p in self.local_packages], 'local package')
+            self.info('total', f"{summary} installed", 'green')
```
LecrisUT, line 273: The only thing that would be an obstacle from moving the whole `_install` to be owned by the PackageManager would be these reports. Probably this can be repatriated if this whole logic is owned by an `install_all` or equivalent function of the PackageManager.

tmt/package_managers/bootc.py
```diff
@@ -335,7 +335,51 @@ def install_debuginfo(
...
         self.engine.install_debuginfo(*installables, options=options)
 
-        return self.build_container() or CommandOutput(stdout=None, stderr=None)
+    def install_from_repository(
+        self,
+        *installables: Installable,
+        options: Optional[Options] = None,
+    ) -> None:
```
thrix: Several `Bootc` methods (`install_from_repository`, `install_from_url`, `install_local`, `install_debuginfo`) declare `-> None` but the base class `PackageManager` declares these as `-> Optional[CommandOutput]`.

While returning `None` is valid for `Optional[CommandOutput]`, the explicit `-> None` annotation is inconsistent and would cause type checker issues for callers expecting `Optional[CommandOutput]`.

This also has a functional impact: in `PrepareInstall._install()`, the walrus operators like:
```python
if output := guest.package_manager.install_from_repository(...):
    install_outputs.append(output)
```
will never append to `install_outputs` for bootc, so `skip_missing` failed package detection via output parsing won't work for bootc guests. (This was a pre-existing limitation, but the new code makes it more explicit.)

tmt/package_managers/rpm_ostree.py
```diff
@@ -193,4 +194,75 @@ def install_debuginfo(
...
+
+        Dnf5(guest=self.guest, logger=self._logger).enable_copr(*repositories)
+
+    def sort_packages(
+        self,
+        *installables: Installable,
+        options: Options,
+    ) -> None:
+        """Sort packages into required and recommended based on presence and skip_missing."""
+        self.required: list[Installable] = []
```
thrix: `sort_packages()` sets `self.required` and `self.recommended` as instance attributes that are then consumed by `install_from_repository()`. These are not declared on the class, only created at runtime, making the pattern fragile.

Consider returning a tuple instead:
```python
def sort_packages(self, ...) -> tuple[list[Installable], list[Installable]]:
    required, recommended = [], []
    ...
    return required, recommended
```
And in `install_from_repository`:
```python
required, recommended = self.sort_packages(*installables, options=options)
```

tmt/steps/prepare/install.py
thrix, line 191: The old `InstallMock.install_local()` had special path handling that prepended `self.guest.root_path` to transform host paths into mock-buildroot-accessible paths:

```python
# Old code in InstallMock:
PackagePath(self.guest.root_path / self.package_directory.relative_to("/") / filename.name)
```

The mock package managers in `tmt/package_managers/mock.py` don't override `install_local()`, so they fall through to the base `PackageManager.install_local()` which just calls `self.install()` without any path transformation. Worth verifying that local package installation still works correctly for mock-provisioned guests.

tmt/package_managers/rpm_ostree.py
happz: In general, let's try to avoid `self` in combination with a mutable value, like a list or dictionary. And in the context of code that can and will be running on multiple guests at the same time it's much safer to stick to locals, parameters, and return values.

LecrisUT: It is also an existing issue, let's not focus on this in this PR

happz: > It is also an existing issue, let's not focus on this in this PR

I don't think it's a completely existing issue. `self.required` now is not the same thing as `self.required_packages` before this PR. The latter `self` was `InstallRpmOstree` instance, freshly created in each invocation of `PrepareInstall.go()`, and thrown away after `go()` method finished; that made it fairly contained under `go()` invocation. `self.required` now is an attribute of `RpmOstree` instance, the only package manager instance attached to the guest, and it is not thrown away once `PrepareInstall.go()` finishes. Every user of `guest.package_manager` could see remnants of previous uses, and if we forget to clear the list before using it once again...

Besides that, there is no reason to introduce it and ask for trouble, `install_from_repository()` does not need to get these values through `self`, neither `required` nor `recommend` need to be attributes of `self`. They **were** attributes, and it wasn't the best idea then, now it would be just worse for no benefit at all. Rewritten to use a return value, the shared mutable state we can forget to protect or reset is gone, no chance to make a mistake:

```python
        required, recommended = self.sort_packages(*installables, options=options)

        for package in recommended:
            self.info('package', str(package), 'green')
            try:
                self.install(package)
            except RunError as error:
                self.debug(f"Package installation failed: {error}")
                self.warn(f"Unable to install recommended package '{package}'.")

        if required:
            return self.install(*required)

        return None
```

tmt/package_managers/__init__.py
AthreyVinay: My 2 cents.... `*repositories` makes intent explicit.

tmt/package_managers/bootc.py
AthreyVinay: This should be fixed here: [2922616](https://github.com/teemtee/tmt/pull/4653/commits/2922616d3d0e78ce57af9e6dc66a7988ca048825)

Bootc's methods now return -> `CommandOutput` matching the base class exactly. Also, `skip_missing` output gap: With -> `CommandOutput` and install_outputs.append(...) (no walrus), Bootc's CommandOutput(stdout=None, stderr=None) placeholder is always appended.

tmt/package_managers/__init__.py
AthreyVinay: This should be fixed here: [2922616](https://github.com/teemtee/tmt/pull/4653/commits/2922616d3d0e78ce57af9e6dc66a7988ca048825)

tmt/package_managers/bootc.py
thrix: ty

tmt/package_managers/rpm_ostree.py
thrix: I agree with @happz. This is NOT the same as the pre-existing pattern - the old `self.required_packages` lived on a fresh `InstallRpmOstree` created per `go()` call and discarded after, while `self.required` now lives on the long-lived `RpmOstree` PM instance. The fix happz proposed is minimal and removes the risk entirely.

tmt/package_managers/bootc.py
thrix: Verified fix in `2922616`. Looks good.

tmt/package_managers/rpm_ostree.py
LecrisUT: It is not much of a blocker for me because this function is only called and used inside `install_from_repository`, so at least for now there is no potential of misuse. I don't even know why this separation of `required` and `recommend` is needed in the first place since this is executed once since the handling of required vs recommend would be handled one layer higher at the PrepareInstall.

It is a simple fix of course, so fine with me if it is a blocker for others.

tmt/package_managers/apk.py
AthreyVinay: Thanks, fixed: [e95f51f](https://github.com/teemtee/tmt/pull/4653/commits/e95f51fd84ef6801d56189f5e4e7cf471e6613ba)

tmt/package_managers/rpm_ostree.py
AthreyVinay: refactored: [d2d3283](https://github.com/teemtee/tmt/pull/4653/commits/d2d32839486772c45feb7cc2ad462f2c08e89812)

tmt/steps/prepare/install.py
AthreyVinay: Fixed: [e87004a](https://github.com/teemtee/tmt/pull/4653/commits/e87004a7a184d2a2f1aabe9b5ab0b0aafac3f7dc)

tmt/steps/prepare/install.py
AthreyVinay: `install_local()` is overridden in `_MockPackageManager` with the same path transformation -`PackagePath(self.guest.root_path / p.relative_to('/'))` here:

https://github.com/teemtee/tmt/blob/e87004a7a184d2a2f1aabe9b5ab0b0aafac3f7dc/tmt/package_managers/mock.py#L195

tmt/steps/prepare/install.py
AthreyVinay: Agreed, noted in the description.

tmt/steps/prepare/install.py
AthreyVinay: Noted.

tmt/steps/prepare/install.py
AthreyVinay: Noted.

tmt/steps/prepare/install.py
AthreyVinay: Noted.

tmt/steps/prepare/install.py
thrix: Confirmed - `_MockPackageManager.install_local()` at line 181 of `mock.py` has the `root_path` transformation. My concern was unfounded. Thanks for pointing this out.

tmt/package_managers/rpm_ostree.py
thrix: Looks good, thanks for the fix.

happz:
@AthreyVinay  please, set the "Size" of this PR.

AthreyVinay:
> @AthreyVinay please, set the "Size" of this PR.

Done - however @psss  is there  a guide similar to story points?

tmt/package_managers/__init__.py
```diff
@@ -124,7 +124,7 @@ def escape_installables(*installables: Installable) -> Iterator[str]:
 
 
 # TODO: find a better name, "options" is soooo overloaded...
-@container(frozen=True)
+@container
 class Options:
```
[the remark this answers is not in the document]
AthreyVinay: Intended.

bajertom:
Failed on Rawhide - avc check only, actual test result is Pass. Log is incomplete, but the previously failing test has passed, no other fails can be seen in the log. Merging.

PR #4656, opened 2026-03-06
Try to get a more reasonable stacklevel

LecrisUT:
Try to provide a more reasonable `stacklevel` to the inner `logging.Logger._log`, skipping any functions that are wrappers around the loggers. The references in the logger's record should point to the most relevant function that requested the log command, or any such indirect commands like the `guest.run`.

---

Found that this would be useful in #4642 for getting a more accurate source of the caller and I split it out from there to simplify the review.

The stack can be further improved as it goes, as shown in the second commit of this PR, but that can be gradually addressed.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/checks/avc.py
+++ b/tmt/checks/avc.py
@@ -144,7 +144,6 @@ def _run_script(
         topic: Optional[tmt.log.Topic] = None,
         stacklevel: int = 1,
     ) -> None:
-        stacklevel += 1
         logger.verbose(
             key=key,
             value=value,
@@ -152,7 +151,7 @@ def _run_script(
             shift=shift,
             level=level,
             topic=topic,
-            stacklevel=stacklevel,
+            stacklevel=stacklevel + 1,
         )
 
     try:
--- a/tmt/utils/__init__.py
+++ b/tmt/utils/__init__.py
@@ -2194,14 +2194,13 @@ class Common(_CommonBase, metaclass=_CommonMeta):
         Show a message unless in quiet mode
         """
 
-        stacklevel += 1
         self._logger.info(
             key,
             value=value,
             color=color,
             shift=shift,
             topic=topic,
-            stacklevel=stacklevel,
+            stacklevel=stacklevel + 1,
         )
 
     def verbose(
@@ -2220,7 +2219,6 @@ class Common(_CommonBase, metaclass=_CommonMeta):
         In quiet mode verbose messages are not displayed.
         """
 
-        stacklevel += 1
         self._logger.verbose(
             key,
             value=value,
@@ -2228,7 +2226,7 @@ class Common(_CommonBase, metaclass=_CommonMeta):
             shift=shift,
             level=level,
             topic=topic,
-            stacklevel=stacklevel,
+            stacklevel=stacklevel + 1,
         )
 
     def debug(
@@ -2247,7 +2245,6 @@ class Common(_CommonBase, metaclass=_CommonMeta):
         In quiet mode debug messages are not displayed.
         """
 
-        stacklevel += 1
         self._logger.debug(
             key,
             value=value,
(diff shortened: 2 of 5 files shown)


tmt/utils/__init__.py
```diff
@@ -2188,12 +2188,21 @@ def info(
...
         shift: int = 0,
         topic: Optional[tmt.log.Topic] = None,
+        stacklevel: int = 1,
     ) -> None:
         """
         Show a message unless in quiet mode
         """
 
-        self._logger.info(key, value=value, color=color, shift=shift, topic=topic)
+        stacklevel += 1
```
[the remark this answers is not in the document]
LecrisUT: Technically possible, but that would make the stack level calculation even more complicated.

happz:
What is a “stacklevel”, and what is it good for? If there’s a vhange to output, would it be possible to inckude an example or two to demonstrate the change, e.g. before/after?

LecrisUT:
> What is a “stacklevel”, and what is it good for? If there’s a vhange to output, would it be possible to inckude an example or two to demonstrate the change, e.g. before/after?

It's related to the stack trace and so far I don't think we use it anywhere yet (will do in #4642). It basically controls how many levels up the stack trace do you navigate (from inside `logging.Logger._log` call) to save the origin of the caller in the `record`. Best place to see this imo is in #4642.

skycastlelily:
> > What is a “stacklevel”, and what is it good for? If there’s a vhange to output, would it be possible to inckude an example or two to demonstrate the change, e.g. before/after?
> 
> It's related to the stack trace and so far I don't think we use it anywhere yet (will do in #4642). It basically controls >how many levels up the stack trace do you navigate (from inside `logging.Logger._log` call) to save the origin of the >caller in the `record`. Best place to see this imo is in #4642.

Hmm, imo it's better to put   _log  in this mr , #4642 is mainly about warning file , and Warnings classes，plus the mr will be merged before #4642.Not strong opinion though:)

tmt/checks/avc.py
```diff
@@ -142,8 +142,18 @@ def _output_logger(
...
-        logger.verbose(key=key, value=value, color=color, shift=shift, level=level, topic=topic)
+        stacklevel += 1
+        logger.verbose(
+            key=key,
+            value=value,
+            color=color,
+            shift=shift,
+            level=level,
+            topic=topic,
+            stacklevel=stacklevel,
```
happz: Instead of `stacklevel += 1`, would `stacklevel=stacklevel + 1` work? One less added line, and the increased `stacklevel` isn't seem to be used anywhere else in this scope. It could reduce the size of the whole patch.

LecrisUT: Yes, it's mostly for unification with `self._logger._log`. Fine with either format if we are consistent with the approach.

happz: I wouldn't limit it to just `_logger.log`, there are also calls to `verbose` and other logger methods that use this approach of `stacklevel += 1`.

LecrisUT: Yes, I singled out that one, because that is the source of it and where the relevant comments about the need for `stacklevel` is

LecrisUT: Done in  432b60c

LecrisUT:
> ```
> The PR fixes source location reporting in log records.                                                                                                                                       
>                                                                                                                                                                                                
>   Python's logging module records the filename, line number, and function name of the code that emitted each log message. Without stacklevel, every log record in tmt would show tmt/log.py    
>   (the internal Logger._log method) as the source, because that's where self._logger._log() is actually called.                                                                                
>                                                                                                                                                                                                
>   With this patch, log records correctly point to the actual caller - e.g., the plugin or step code that called logger.verbose() or self.info(). This matters when:
> 
>   - Using log formatters that include %(filename)s:%(lineno)d (standard Python logging pattern)
>   - Debugging with file-based log handlers that show source locations
>   - Any tooling that inspects LogRecord.pathname / LogRecord.lineno to trace where a message originated
> 
>   Before this PR, all log records would point to the same internal location regardless of which code emitted them. After it, each record correctly identifies its origin in the business logic.
> 
>   The PR description notes this was split out from #4642, which specifically needs accurate caller information.
> ```
> 
> @LecrisUT I would make the MR description a bit more descriptive, for example this AI summary is a lot better for me then the current description.

I do not like the verbosity there at all. If there is a concise version I could consider, but I do not want the commit message to be so full of tangents

Edit: tried to manually improve the message.

PR #4658, opened 2026-03-06
Fix catastrophic regex backtracking in shell framework

### LINKED ISSUE #4657: Catastrophic regex backtracking in shell framework _extract_failures on long lines
## Summary

The `_extract_failures()` function in `tmt/frameworks/shell.py` uses a regex pattern that causes catastrophic backtracking when test output contains very long lines (e.g. 1M+ characters). This results in tmt processes appearing stuck at 75% CPU for hours, blocking test execution.

## Root cause

The regex at line 20:

```python
re.findall(r'.*\b(?:error|fail)\b.*', log, re.IGNORECASE | re.MULTILINE)
```

The leading and trailing `.*` combined with word boundary checks causes the Python regex engine to backtrack extensively on long lines that don't contain "error" or "fail". With `re.MULTILINE`, `.*` matches within each line, but on a 1M-character line the engine explores an enormous number of positions before giving up.

**Measured impact:**
- 10,000 characters: **5.4 seconds**
- 50,000+ characters: **does not complete within 10 seconds**
- 1,000,000 characters: **would take hours** (observed 130+ minutes CPU time in production)

The worker thread holds the GIL while executing `re.findall()`, blocking the main thread waiting in `concurrent.futures.as_completed()`.

## How we hit this

Testing Farm jobs from the [Hummingbird](https://gitlab.com/redhat/hummingbird) project were getting stuck. The test output (`Verify-reproducibility-1/output.txt`) contained base64-encoded [in-toto](https://in-toto.io/) attestation payloads dumped via `bash -x` tracing — single lines of 1,058,495 and 467,425 characters.

~30 jobs were affected simultaneously.

## Proposed fix

Replace the single `re.findall()` with a line-by-line `re.search()` without the leading/trailing `.*`:

```python
return [
    line for line in log.splitlines()
    if re.search(r'\b(?:error|fail)\b', line, re.IGNORECASE)
    ]
```

This is a simple linear scan — O(n) per line with no backtracking. Tested on the same 1.6 MB file: completes in **1.1 seconds** vs hours with the old regex.

## Blog post

Detailed write-up of the debugging session: https://vadkerti.net/posts/debugging-stuck-tmt-processes-with-claude-code/

thrix:
## Summary

Fix `_extract_failures()` in `tmt/frameworks/shell.py` which causes tmt processes to hang for hours when test output contains very long lines.

- Replace `re.findall(r'.*\b(?:error|fail)\b.*', ...)` with a line-by-line `re.search()` approach that avoids catastrophic regex backtracking
- Add unit tests for `_extract_failures()` including a regression test with 1M-character lines

Fixes #4657

## Context

Testing Farm jobs from the [Hummingbird](https://gitlab.com/redhat/hummingbird) project were getting stuck in "running" state for hours. The test output files contained base64-encoded in-toto attestation payloads — single lines of 1M+ characters — causing the regex engine to backtrack indefinitely.

Detailed write-up: https://vadkerti.net/posts/debugging-stuck-tmt-processes-with-claude-code/

## Test plan

- [x] Unit tests pass (`hatch -e dev run -- python -m pytest tests/unit/test_shell.py -v`)
- [ ] Existing shell framework integration tests still pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,7 +142,6 @@ dependencies = [
     "pytest",
     "python-coveralls",
     "pytest-container",
-    "pytest-timeout",
     "pytest-xdist",
     "requre",
     "pre-commit",
--- a/tests/unit/test_shell.py
+++ b/tests/unit/test_shell.py
@@ -118,19 +118,8 @@ def test_extract_failures_file_error(invocation: MagicMock) -> None:
             ),
             id='long-line-word-boundary-no-false-positive',
         ),
-        pytest.param(
-            FailureCase(
-                log_content=(
-                    f'start\n{_LONG_SEGMENT} a b c d e f g h i j k l m n o p '
-                    f'{_LONG_SEGMENT}\nend\n'
-                ),
-                expected=[],
-            ),
-            id='long-line-many-word-boundaries-no-match',
-        ),
     ],
 )
-@pytest.mark.timeout(10)
 def test_extract_failures_long_lines(invocation: MagicMock, case: FailureCase) -> None:
     """
     Verify correct handling of very long lines (~50k characters).
@@ -141,20 +130,18 @@ def test_extract_failures_long_lines(invocation: MagicMock, case: FailureCase) -
 
     Benchmarks with 50k-character lines:
 
-    ====================================  ===========  ===========
-    Case                                  Old regex    New method
-    ====================================  ===========  ===========
-    long line, no match + error line      11.335 s     < 0.001 s
-    error embedded in long line           0.001 s      < 0.001 s
-    word boundary, no false positive      49.290 s     < 0.001 s
-    many word boundaries, no match        133.628 s    < 0.003 s
-    ====================================  ===========  ===========
+    ==================================  ===========  ===========
+    Case                                Old regex    New method
+    ==================================  ===========  ===========
+    long line, no match + error line    11.335 s     < 0.001 s
+    error embedded in long line         0.001 s      < 0.001 s
+    word boundary, no false positive    49.290 s     < 0.001 s
+    ==================================  ===========  ===========
 
     The new ``splitlines()`` + per-line ``re.search()`` runs in < 0.001 s
     for all cases. A 0.1 s threshold sits well between the two methods:
     any result above 0.1 s indicates a regression toward the old behavior.
-    The ``@pytest.mark.timeout(10)`` serves as an additional safety net
-    to prevent CI from hanging if a regression is catastrophic.
+    The explicit time assertion serves as a safeguard against regressions.
     """
     invocation.phase.step.plan.execute.read.return_value = case.log_content
 


thrix:
@gemini-code-assist please review again

tests/unit/test_shell.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def _make_invocation(log_content: str) -> MagicMock:
+    """Create a mock TestInvocation that returns given log content on read."""
+    mock = MagicMock()
+    mock.phase.step.plan.execute.read.return_value = log_content
+    return mock
+
+
+class TestExtractFailures:
```
happz: Please, no classes, they do not play well with pytest fixtures.

thrix: @happz should be resolved.

tmt/frameworks/shell.py
```diff
@@ -17,7 +17,9 @@ def _extract_failures(invocation: 'TestInvocation', log_path: Path) -> list[str]
     except tmt.utils.FileError:
         return []
 
-    return re.findall(r'.*\b(?:error|fail)\b.*', log, re.IGNORECASE | re.MULTILINE)
+    return [
+        line for line in log.splitlines() if re.search(r'\b(?:error|fail)\b', line, re.IGNORECASE)
+    ]
```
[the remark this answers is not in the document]
thrix: should be resolved.

tests/unit/test_shell.py
```diff
@@ -0,0 +1,78 @@
...
+        result = _extract_failures(invocation, Path("dummy.log"))
+        assert result == ["error one", "fail two", "error three"]
+
+    def test_word_boundary(self) -> None:
+        """Words like 'errorless' or 'failover' should not match."""
+        invocation = _make_invocation("errorless operation\nfailover complete\n")
+        assert _extract_failures(invocation, Path("dummy.log")) == []
+
+    def test_file_error(self) -> None:
+        import tmt.utils
```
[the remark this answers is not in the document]
thrix: Resolved

thrix:
@gemini-code-assist please review again

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+import tmt.utils
+from tmt.frameworks.shell import _extract_failures
+from tmt.utils import Path
+
+
+def _make_invocation(log_content: str) -> MagicMock:
+    """Create a mock TestInvocation that returns given log content on read."""
+    mock = MagicMock()
+    mock.phase.step.plan.execute.read.return_value = log_content
+    return mock
```
LecrisUT: We use fixture, please make this a proper fixture

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def test_word_boundary() -> None:
+    """Words like 'errorless' or 'failover' should not match."""
+    invocation = _make_invocation("errorless operation\nfailover complete\n")
+    assert _extract_failures(invocation, Path("dummy.log")) == []
+
+
+def test_file_error() -> None:
+    invocation = MagicMock()
```
LecrisUT: Inconsistent documentaiton of  tests

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+def test_error_match() -> None:
+    invocation = _make_invocation("line1\nsome error occurred\nline3\n")
+    result = _extract_failures(invocation, Path("dummy.log"))
+    assert result == ["some error occurred"]
+
+
+def test_fail_match() -> None:
+    invocation = _make_invocation("test passed\ntest fail here\ndone\n")
+    result = _extract_failures(invocation, Path("dummy.log"))
+    assert result == ["test fail here"]
```
LecrisUT: So much repetition. Please make these parametrized

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+    result = _extract_failures(invocation, Path("dummy.log"))
+    elapsed = time.monotonic() - start
+
+    # Must complete in under 5 seconds (the old regex took 5+ seconds
+    # on just 10k characters, and would never complete on 1M characters)
+    assert elapsed < 5.0, (
+        f"_extract_failures took {elapsed:.1f}s on a log with a 1M-char line; "
+        f"likely catastrophic regex backtracking"
+    )
+    assert result == ["some error here"]
```
LecrisUT: Checking test times in tests is very iffy to me. But in absence of benchmark infrastructure 🤷 . I am also fine with just ignoring this test since it fixed a rather edge-case situation

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    # Build a log with a 1M-character line (similar to base64 attestation data)
+    long_line = "A" * 1_000_000
+    log_content = f"start\n{long_line}\nsome error here\nend\n"
+
+    invocation = _make_invocation(log_content)
+
+    start = time.monotonic()
+    result = _extract_failures(invocation, Path("dummy.log"))
+    elapsed = time.monotonic() - start
```
LecrisUT: The choice of `monotic` here seems arbitrary?

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Must complete in under 5 seconds (the old regex took 5+ seconds
+    # on just 10k characters, and would never complete on 1M characters)
+    assert elapsed < 5.0, (
```
LecrisUT: 5 seconds is a very arbitrary cut-off. If the old one would have finished these in 5.001 seconds, and this change brings it down to 4.999, this test does not really tell anything. Similarly if we make changes to the code and it gradually increases the test time to the point that it takes 4.999 tests still pass.

If there is log-scale changes then confirm the intended change are at least an order of magnitude difference and account more realistically buffers of what time scale it is expected

tests/unit/test_shell.py
```diff
@@ -0,0 +1,83 @@
...
+    on very long lines.
+
+    The original implementation used re.findall(r'.*\\b(?:error|fail)\\b.*', ...)
+    which caused O(n^2) or worse backtracking on long lines without matches,
+    hanging tmt processes for hours on 1M+ character lines (e.g. base64-encoded
+    in-toto attestation payloads in container build logs).
+    """
+    # Build a log with a 1M-character line (similar to base64 attestation data)
+    long_line = "A" * 1_000_000
+    log_content = f"start\n{long_line}\nsome error here\nend\n"
```
LecrisUT: What if the long line does not have a new-line that separates from the `some error` part, what if it is comprised of multiple words which would match the `\b`?

thrix: checking

tests/unit/test_shell.py
thrix: Converted _make_invocation to a `@pytest.fixture`

tests/unit/test_shell.py
thrix: Added consistent docstrings on all test functions

tests/unit/test_shell.py
thrix: Combined 6 individual test functions into a single parametrized test_extract_failures with descriptive IDs

tests/unit/test_shell.py
thrix: Added comment explaining why time.monotonic() is the correct choice (unaffected by system clock adjustments)

tests/unit/test_shell.py
thrix: Changed to 30 seconds — generous ceiling that still catches catastrophic backtracking (old regex never completed on 1M chars in 10s of minutes)

tests/unit/test_shell.py
thrix: Kept the test but integrated timing assertion into the parametrized long-line tests rather than a standalone test, with clear documentation of the regression it prevents

tests/unit/test_shell.py
thrix: Added 3 parametrized long-line cases: error on separate line, error embedded within the long line (no newline separator), and word-boundary false positive withing long line

thrix:

tests/unit/test_shell.py
thrix: resolving as outdated, please open a new comment if needed.

tests/unit/test_shell.py
thrix: resolving as outdated, please open a new comment if needed.

tests/unit/test_shell.py
LecrisUT: > Added comment explaining why time.monotonic() is the correct choice (unaffected by system clock adjustments)

Who/How do you expect these things to be run. Sounds like ridiculous overkill, but fine, at least it would not hurt other than being inconsistent with the other places that we use `time`

tests/unit/test_shell.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def test_extract_failures_file_error() -> None:
+    """Verify _extract_failures returns empty list when the log file cannot be read."""
+    invocation = MagicMock()
+    invocation.phase.step.plan.execute.read.side_effect = tmt.utils.FileError('not found')
+    assert _extract_failures(invocation, Path('dummy.log')) == []
+
+
+_LONG_LINE_CASES: list[tuple[str, str, list[str]]] = [
```
LecrisUT: Please no list of tuple of list of ... Make it structured in a dataclass

tests/unit/test_shell.py
```diff
@@ -0,0 +1,120 @@
...
+        'start\n{long}errorless{long}\nend\n',
+        [],
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    ('log_template', 'expected_template'),
+    [(log, expected) for _, log, expected in _LONG_LINE_CASES],
+    ids=[name for name, _, _ in _LONG_LINE_CASES],
```
LecrisUT: Use `pytest.param` to put the `id` next to the value. Maybe also store it in the dataclass

tests/unit/test_shell.py
```diff
@@ -0,0 +1,120 @@
...
+@pytest.fixture
+def make_invocation():
+    """Factory fixture creating a mock TestInvocation returning given log content."""
+
+    def _factory(log_content: str) -> MagicMock:
+        mock = MagicMock()
+        mock.phase.step.plan.execute.read.return_value = log_content
+        return mock
+
+    return _factory
```
LecrisUT: Why the whole factory? You do not need it, let it provide just the mock object and mock the relevant part in the tests. That would make it more consistent with all other instances

tests/unit/test_shell.py
```diff
@@ -0,0 +1,120 @@
...
+    start = time.monotonic()
+    result = _extract_failures(invocation, Path('dummy.log'))
+    elapsed = time.monotonic() - start
+
+    assert result == expected
+
+    # The old regex would never complete on lines this long — it took 5+
+    # seconds on just 10k characters. The splitlines approach finishes in
+    # well under 1 second. Use 30 seconds as a generous ceiling to avoid
+    # flakiness on slow CI while still catching catastrophic backtracking.
```
LecrisUT: I don't like this. The difference is well within a single order of magnitude. It seems that the test is not catching anything significant.

tests/unit/test_shell.py
```diff
@@ -0,0 +1,120 @@
...
+    (
+        'error embedded in long line without newline separator',
+        'start\n{long} error in the middle {long}\nend\n',
+        ['{long} error in the middle {long}'],
+    ),
+    (
+        'long line with word boundary - no false positive',
+        'start\n{long}errorless{long}\nend\n',
+        [],
+    ),
```
LecrisUT: This just shows that what we are doing in `_extract_failures` is likely wrong, e.g. the user could have something
```
def test_accuracy()
    value = get_value()
    assert abs(value - refrence) < 1E-10
    print("All is good, the value is within the acceptable error bar") 
```
Why are we catching any word error and concluding that if it is present it must be relevant. It is a very poorman's guess and the question is if anyone is actually relying on these.

Would consider adding a deprecation note and see who actually complains and/or cares about them. A 0 second evaluation is faster still

tests/unit/test_shell.py
```diff
@@ -0,0 +1,120 @@
...
+
+    The original implementation used ``re.findall(r'.*\\b(?:error|fail)\\b.*', ...)``
+    which caused catastrophic backtracking on long lines (O(n^2) or worse),
+    hanging tmt processes for hours on 1M+ character lines (e.g. base64-encoded
+    in-toto attestation payloads in container build logs).
+
+    The current implementation uses ``str.splitlines()`` and per-line
+    ``re.search()`` which processes each line in linear time.
+    """
+    long_segment = 'A' * 1_000_000
```
LecrisUT: Move this construction outside of the test, and make it more lorem-ipsum-like with more word separators and variations that it could randomly hit.

tests/unit/test_shell.py
LecrisUT: > what if it is comprised of multiple words which would match the \b?

Is not addressed, but also the time difference between them is unclear

tests/unit/test_shell.py
LecrisUT:
```
    # The old regex would never complete on lines this long — it took 5+
    # seconds on just 10k characters.
```
Unless we know the scaling, how long it takes on 10k characters is unclear if it's relevant for 1M.

tests/unit/test_shell.py
thrix: @LecrisUT sure, but that is not for this issue, I will file a new one

tests/unit/test_shell.py
LecrisUT: Well, here are the measurements that I got using the 50000 length:
- Old method
  - Case 1: 11.335 s
  - Case 2: 0.001 s
  - Case 3: 49.290 s
- New method: All 0.001 s

What I would propose is documenting this difference explicitly. Choose whatever size you want and run them fully and document explicitly how much they (can) differ. E.g. in this case all of them are expected to be instantaneous in the new case so cutting in the middle, having a threshold of `0.1 s` guarantees that on either side will for sure fail and for sure pass, mentioning how much the time difference is expected:
- at the time of writing the test with the old method expecting `XX s`
- at the time of writing the test with the current method expected below measurable `0.001 ms`

---

Running these does highlight one weird quirk between Case 1 and Case 2. I would not have expected that the latter to be that much faster in the old method given the documentation in the test

tcornell-bus: I ran the test locally without the code fix (to make sure it fails) and it hung until I did Ctrl+C. Gemini suggested I add `@pytest.mark.timeout(5)` which force fails the test after 5 (or however many) sec and it can move onto the next test. We would need `pytest-timeout` installed. It worked when I added it, and it should hopefully prevent our CI from getting stuck if anyone breaks the regex again.

thrix: @tcornell-bus thanks for pytest-timeout, that should make this safer added.

tests/unit/test_shell.py
thrix: @LecrisUT resolving this one, we are discussing this in other thread.

tests/unit/test_shell.py
thrix: @LecrisUT can you recheck?

tests/unit/test_shell.py
thrix: Added `FailureCase` `tmt.container`

tests/unit/test_shell.py
thrix: did not knwo about `pytest.param`, it is there

tests/unit/test_shell.py
thrix: yeah, you are right, inly a Mock object returned now

tests/unit/test_shell.py
thrix: @LecrisUT does it look better now?

tests/unit/test_shell.py
thrix: Created under https://github.com/teemtee/tmt/issues/4739

tests/unit/test_shell.py
thrix: @LecrisUT it is gone outside

tests/unit/test_shell.py
```diff
@@ -0,0 +1,169 @@
...
+        pytest.param(
+            FailureCase(
+                log_content=(
+                    f'start\n{_LONG_SEGMENT} a b c d e f g h i j k l m n o p '
+                    f'{_LONG_SEGMENT}\nend\n'
+                ),
+                expected=[],
+            ),
+            id='long-line-many-word-boundaries-no-match',
+        ),
```
LecrisUT: This case is not needed with the form of `_LONG_SEGMENT` now

pyproject.toml
```diff
@@ -142,6 +142,7 @@ dependencies = [
     "pytest",
     "python-coveralls",
     "pytest-container",
+    "pytest-timeout",
```
LecrisUT: Afaict, [python-pytest-timeout](https://src.fedoraproject.org/rpms/python-pytest-timeout) is not available on epel9, so let's skip it. Plus this is not added in the correct place anyway, and the explicit time is still being checked.

tcornell-bus: If a regression occurs without the `@pytest.mark.timeout(10)` check the test could hang for minutes while its waiting for it to complete.

LecrisUT: Yeah, but realistically it shouldn't occur. At most it would only be 1 minute and it very much depends on the hardware, phase of the moon and who knows what else. Personally I am fine with just dropping that test completely, and documenting the expectation in the commit description

thrix: @happz @psss seems we are back an forth on the usefullness of this test. Is the broader consensus among the elders to drop this? I and @tcornell-bus are more in the keep it camp, while @LecrisUT obviously finds it uneeded. My QE heart tells me a regression test is needed, but because we plan to drop this part hopefully in #4739, no strong feelings

tcornell-bus: Oh ok, I didn't realize this check may get dropped entirely. I vote we remove the `pytest-timeout` check, so we don't add a dependency that may never get removed again. But it would be nice to keep the test. Who knows how long until the check gets dropped?

tests/unit/test_shell.py
thrix: @LecrisUT removed

pyproject.toml
thrix: removed `pytest-timeout`

LecrisUT:
linkcheck failure is know, merging

Issue #4646, opened 2026-03-04
Link the verify install plugin to the artifact plugin


vaibhavdaren:
## Problem Statement
As a developer working on the `verify-installation` plugin (PR #4660), I need two follow-up improvements to make the plugin easier to test in isolation and more useful in real-world workflows:

1. **Test structure is unnecessarily split.** The current tests use separate `data-centos`/`data-fedora` directories.

2. **The plugin has no integration with `prepare/artifact`.** When a user specifies packages via `prepare/artifact` for a provider whose artifacts are declared in `artifacts.yaml` (e.g. Copr, Brew, koji), the `verify-installation` plugin currently has no way to automatically know which repository those packages are expected to come from. The user must manually duplicate this information. Instead, the plugin should be able to derive the expected repo from the artifact provider configuration already present in the plan.

## Out of Scope
- Adding new providers to `artifacts.yaml` (that is a separate concern).
- Support for package managers other than dnf4/dnf5 (already tracked separately).

vaibhavdaren:
PR which will close 1 : https://github.com/teemtee/tmt/pull/4723
PR which will close 2: https://github.com/teemtee/tmt/pull/4726

PR #4660, opened 2026-03-09
Add a new feature plugin to verify package installation

### LINKED ISSUE #4644: Create a Verify Install plugin which checks if packages are installed from the correct repo
```yaml
    prepare:
      - how: verify-installation
        order: 76
        verify:
          - package: make
            expected_repo: tmt-artifact-shared
          - package: make-devel
            expected_repo: tmt-artifact-shared
          - package: diffutils
            expected_repo: fedora
    execute:
````

TBD

### PR DESCRIPTION
**Pull request by vaibhavdaren**
Introduce `prepare/verify-installation` plugin to verify that a package installation come from a specific pre-declared repository

Closes #4644

---

Pull Request Checklist:

* [x] implement the feature
* [x] write the documentation
* [x] adjust plugin docstring
* [x] modify the json schema
* [x] include a release note

### DIFF
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -18,10 +18,10 @@ class SpecialPackageOrigin(str, enum.Enum):
     special package states returned by :py:meth:`PackageManager.get_package_origin`.
     """
 
-    # Package is not installed on the guest.
+    #: Package is not installed on the guest.
     NOT_INSTALLED = '<not-installed>'
-    # Package is installed but its source repository cannot be determined
-    # (e.g. pre-installed in a container image).
+    #: Package is installed but its source repository cannot be determined
+    #: (e.g. pre-installed in a container image).
     UNKNOWN = '<unknown>'
 
 
@@ -251,7 +251,9 @@ class PackageManagerEngine(tmt.utils.Common):
         Empty lines are allowed and will be ignored by the caller.  If
         the origin field is omitted the package is treated as having an
         unknown source repository (equivalent to
-        :py:attr:`SpecialPackageOrigin.UNKNOWN`).
+        :py:attr:`SpecialPackageOrigin.UNKNOWN`).  Packages whose name
+        does not appear in the output at all are treated as not installed
+        (equivalent to :py:attr:`SpecialPackageOrigin.NOT_INSTALLED`).
 
         :param packages: Package names to query.
         :returns: A shell script to list source repositories for the given packages.
@@ -398,7 +400,7 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
             parts = line.split(maxsplit=1)
             package = parts[0]
             # Omitted origin field → unknown source repository.
-            result[package] = parts[1].strip() if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
+            result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
         return result
 
     def create_repository(self, directory: Path) -> CommandOutput:

### CODE REVIEW

File: tmt/steps/prepare/verify_installation.py [hunk 1]
```diff
@@ -0,0 +1,212 @@
...
+
+
+def _parse_verify_string(key_address: str, value: str) -> VerifyMapping:
+    """Parse 'package,repo' format into a VerifyMapping."""
+    parts = value.split(',')
+    if len(parts) != 2:
+        raise tmt.utils.NormalizationError(
+            key_address, value, "a 'package_name,repository_name' string"
+        )
+    return VerifyMapping(package=Package(parts[0].strip()), expected_repo=parts[1].strip())
```
LecrisUT: Would rather not have this provided as comma-separated values, prefer to have the more explicit yaml dict. Can make this not have a cli counterpart until someone actually requests it.

tmt/package_managers/__init__.py
```diff
@@ -222,6 +222,16 @@ def list_packages(self, repository: "Repository") -> ShellScript:
...
 
+    def get_installed_repo(self, package: str) -> ShellScript:
+        """
+        Return a shell script that outputs the repository a package was installed from.
+
+        :param package: The package name to query.
+        :returns: A shell script whose stdout contains the repository name.
+        :raises NotImplementedError: If the package manager does not support this query.
+        """
+        raise NotImplementedError
```
LecrisUT, line 262: Docstring does not fit with the previous

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,22 @@ def list_packages(self, repository: "Repository") -> list[str]:
 
         return stdout.strip().splitlines()
 
+    def get_installed_repo(self, package: str) -> Optional[str]:
```
LecrisUT: Why the `Optional` here? When is it an expected output

tmt/package_managers/dnf.py
```diff
@@ -200,6 +201,17 @@ def list_packages(self, repository: Repository) -> ShellScript:
...
+    def get_installed_repo(self, package: str) -> ShellScript:
+        """
+        Return a shell script that outputs the repository a package was installed from.
+
+        Uses ``repoquery --installed --queryformat %{from_repo}`` to query the RPM database.
+        """
+        return ShellScript(
+            f'{self.command.to_script()} repoquery --installed'
+            f' --queryformat "%{{from_repo}}" {shlex.quote(package)}'
+        )
```
LecrisUT: This can accept a list of packages. Try to refactor this to query all packages at once whenever possible.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,212 @@
...
+        # only build a ShellScript locally, so this raises NotImplementedError
+        # immediately if the package manager doesn't support repo queries.
+        try:
+            guest.package_manager.engine.get_installed_repo('')
+        except NotImplementedError:
+            self.warn(
+                f"Package source verification not supported for "
+                f"'{guest.facts.package_manager}' package manager, skipping."
+            )
+            return outcome
```
LecrisUT: Rather fail completely if it gets to this point

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,212 @@
...
+        logger: Logger,
+    ) -> tmt.steps.PluginOutcome:
+        """Perform package source verification."""
+        outcome = super().go(guest=guest, environment=environment, logger=logger)
+
+        if self.is_dry_run:
+            return outcome
+
+        if not self.data.verify:
+            self.verbose('verify', 'no packages to verify', level=2)
```
LecrisUT: Why the use of `level` here?

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,212 @@
...
+
+        failures = self._verify_packages(guest, self.data.verify)
+
+        if failures:
+            self._report_failures(failures)
+            raise tmt.utils.PrepareError(
+                f"Package source verification failed for {len(failures)} package(s)."
+            )
+
+        self.info('verify', 'all packages verified successfully', 'green')
```
LecrisUT: Don't think these should have the `verify` key attached to it.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,212 @@
...
+            'green',
+        )
+
+        failures = self._verify_packages(guest, self.data.verify)
+
+        if failures:
+            self._report_failures(failures)
+            raise tmt.utils.PrepareError(
+                f"Package source verification failed for {len(failures)} package(s)."
+            )
```
LecrisUT: Here is a good opportunity to design this properly. Please add any failures directly in the `outcome`

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,212 @@
...
+            actual_repo = self._get_package_repository(guest, verify_mapping.package)
+
+            if actual_repo != verify_mapping.expected_repo:
+                failures.append(
+                    VerificationFailure(
+                        package=verify_mapping.package,
+                        expected_repo=verify_mapping.expected_repo,
+                        actual_repo=actual_repo,
+                    )
+                )
```
LecrisUT: Instead of gathering and then processing the log, you can add specific `PhaseResult` with a failure and the cause as you process each pacakge

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,176 @@
...
+            if actual_repo == verify_mapping.expected_repo:
+                continue
+
+            if actual_repo is None:
+                note = (
+                    f"Package '{package}': expected repo '{verify_mapping.expected_repo}'"
+                    f", but the package is not installed or its source repository"
+                    f" could not be determined."
+                )
+            else:
```
thrix, line 128: Verification failures are recorded as `PhaseResult(result=FAIL)` in `outcome.results`, but the prepare step only raises `PrepareError` when `outcome.exceptions` has entries (see `tmt/steps/prepare/__init__.py` line 499). FAIL results are recorded but **do not stop test execution**.

The docstring above says "Verification failure will cause the prepare step to fail and prevent test execution" - but the code doesn't match this promise.

To actually fail the prepare step, either raise `PrepareError` after collecting all failures:

```python
if any(r.result == ResultOutcome.FAIL for r in outcome.results):
    failed = [r.name for r in outcome.results if r.result == ResultOutcome.FAIL]
    raise tmt.utils.PrepareError(
        f"Package verification failed for: {', '.join(failed)}"
    )
```

Or append to `outcome.exceptions` so the prepare step framework handles it.

tmt/schemas/prepare/verify-installation.yaml
```diff
@@ -0,0 +1,55 @@
...
+      type: object
+      additionalProperties: false
+      required:
+        - package
+        - expected_repo
+      properties:
+        package:
+          type: string
+        expected_repo:
+          type: string
```
thrix: tmt uses kebab-case (hyphens) for user-facing YAML keys: `dist-git-source`, `is-virtualized`, `modified-ref`, etc. `expected_repo` with underscore is inconsistent.

Should be `expected-repo` in the schema (and correspondingly in the normalization function and test data). The container's `NormalizeKeysMixin` handles hyphen-to-underscore conversion for Python attribute names automatically.

tmt/package_managers/dnf.py
```diff
@@ -200,6 +201,13 @@ def list_packages(self, repository: Repository) -> ShellScript:
             """
         )
 
+    def get_installed_repos(self, packages: list[str]) -> ShellScript:
+        package_args = ' '.join(shlex.quote(p) for p in packages)
+        return ShellScript(
+            f'{self.command.to_script()} repoquery --installed'
```
thrix: `Dnf5Engine` inherits this without override. Worth verifying that `dnf5 repoquery --installed --queryformat "%{name} %{from_repo}\n"` works correctly on dnf5 - the `--queryformat` syntax and `%{from_repo}` tag may behave differently between dnf4 and dnf5.

Also, the `yum` engine (via `YumEngine`) would inherit this too. On older systems where `yum` is the real yum (not a dnf symlink), `yum repoquery` may not support this syntax.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+    mappings: list[VerifyMapping] = []
+    for raw_item in cast(list[object], value):
+        if (
+            not isinstance(raw_item, dict)
+            or 'package' not in raw_item
+            or 'expected_repo' not in raw_item
+        ):
+            raise tmt.utils.NormalizationError(
+                key_address, raw_item, "a dict with 'package' and 'expected_repo' keys"
```
thrix: This normalization function manually checks for `'expected_repo'` (underscore) in the raw YAML dict. Since tmt convention is kebab-case in YAML, the fmf/container normalization layer normally converts hyphens to underscores for Python attributes. But this manual check bypasses that layer, forcing users to write underscores in YAML.

If the key is renamed to `expected-repo` (per tmt convention), this check should look for `'expected-repo'` in the raw dict, or better yet, let the container normalization handle it automatically (similar to how `DiscoverFmfStepData` fields are normalized).

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,176 @@
...
+        serialize=lambda mappings: [
+            {'package': m.package, 'expected_repo': m.expected_repo} for m in mappings
+        ],
+        unserialize=lambda data: [
+            VerifyMapping(
+                package=Package(item['package']),
+                expected_repo=item['expected_repo'],
+            )
+            for item in cast(list[dict[str, str]], data)
+        ],
```
LecrisUT: These can be handled indirectly with `to_spec`/`from_spec` of `DataContainer`

tmt/package_managers/__init__.py
```diff
@@ -222,6 +222,17 @@ def list_packages(self, repository: "Repository") -> ShellScript:
         """
         raise NotImplementedError
 
+    def get_installed_repos(self, packages: list[str]) -> ShellScript:
```
LecrisUT: Use `Iterable` whenever possible. Alternatively `*packages` is also fine. `list` can be dangerous because it allows mutating the object of the caller.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,188 @@
...
+                PhaseResult(
+                    name=package,
+                    result=ResultOutcome.FAIL,
+                    note=[note],
+                    guest=ResultGuestData.from_guest(guest=guest),
+                )
+            )
+
+        if not has_failures:
+            self.info('All packages verified successfully.', color='green')
```
LecrisUT: Fine to just skip this. The user would have a failure message if anything is unexpected.

tmt/steps/prepare/verify_installation.py
LecrisUT: That seems like a bug on the `Prepare` step. When do we not want to stop the run when a step/phase has failed?

tests/prepare/verify-installation/data/main.fmf
```diff
@@ -0,0 +1,59 @@
+/plan/success:
+    discover:
+        how: fmf
+        test: /test/success
+    provision:
+        how: container
+    prepare:
+      - how: artifact
+        provide:
+          - koji.build:2761810
```
LecrisUT: Why is this hard-coded?

tests/prepare/verify-installation/data/test-success.sh
```diff
@@ -0,0 +1,22 @@
...
+# Verify it came from the tmt-artifact-shared repository
+dnf info --installed make-devel | grep -E "From repo(sitory)?\s*:\s*tmt-artifact-shared"
+
+# Verify diffutils is installed
+rpm -q diffutils
+
+# Verify it came from the fedora repository
+dnf info --installed diffutils | grep -E "From repo(sitory)?\s*:\s*fedora"
+
+echo "SUCCESS: All packages verified to be from expected repositories"
```
LecrisUT: Why do we do these internal checks again when they are covered by other tests?

tests/prepare/verify-installation/test.sh
```diff
@@ -0,0 +1,43 @@
...
+
+        setup_distro_environment
+    rlPhaseEnd
+
+    rlPhaseStartTest "Test successful verification"
+        rlLog "Verify that package from artifact repository passes verification"
+
+        rlRun -s "tmt run -i \$run/success --scratch -vvv --all \
+            plan --name /plan/success \
+            provision -h \$PROVISION_HOW --image \$TEST_IMAGE_PREFIX/\$image_name 2>&1" 0 "Run verification test with correct repo"
```
LecrisUT:
```suggestion
            provision -h \$PROVISION_HOW --image \$TEST_IMAGE_PREFIX/\$image_name" 0 "Run verification test with correct repo"
```
This is explicitly what `-s` in [`rlRun`](https://beakerlib.readthedocs.io/en/latest/manual.html#rlrun) does.

tests/prepare/verify-installation/test.sh
```diff
@@ -0,0 +1,43 @@
...
+
+    rlPhaseStartTest "Test verification failure"
+        rlLog "Verify that plugin processes all packages and then fails with wrong repo"
+
+        rlRun -s "tmt run -i \$run/failure --scratch -vvv --all \
+            plan --name /plan/failure \
+            provision -h \$PROVISION_HOW --image \$TEST_IMAGE_PREFIX/\$image_name 2>&1" 1 "Verification should fail with wrong repo"
+
+        rlAssertGrep "3 packages" $rlRun_LOG
+        rlAssertGrep "fail make-devel" $rlRun_LOG
```
LecrisUT: Include `rlAssertNotGrep` for the checks that are expected to no appear

tmt/package_managers/dnf.py
```diff
@@ -200,6 +201,13 @@ def list_packages(self, repository: Repository) -> ShellScript:
...
+            f' --queryformat "%{{name}} %{{from_repo}}\\n" {package_args}'
+        )
```
LecrisUT: Make this a  `Command` and you will avoid the whole f-string formatting workarounds

tmt/package_managers/__init__.py
```diff
@@ -338,6 +349,23 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        :param packages: Package names to query.
+        :returns: A mapping of package names to source repository names.
+            Packages not installed or with undetermined source are omitted.
+        """
+        script = self.engine.get_installed_repos(packages)
+        output = self.guest.execute(script)
+        result: dict[str, str] = {}
+        for line in (output.stdout or '').strip().splitlines():
+            parts = line.split()
+            if len(parts) == 2:
```
LecrisUT: Assert `len(parts)` or do something for the invalid case.

tmt/package_managers/__init__.py
```diff
@@ -338,6 +349,23 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        """
+        Get the repositories packages were installed from.
+
+        :param packages: Package names to query.
+        :returns: A mapping of package names to source repository names.
+            Packages not installed or with undetermined source are omitted.
+        """
+        script = self.engine.get_installed_repos(packages)
+        output = self.guest.execute(script)
+        result: dict[str, str] = {}
```
LecrisUT: Prepopulate the result with the packages as `None` or uninstalled. That would make it clearer in the navigation what the actual result is.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,188 @@
...
+        if not self.data.verify:
+            self.verbose('No packages to verify.')
+            return outcome
+
+        self.info(
+            fmf.utils.listed([m.package for m in self.data.verify], 'package'),
+            color='green',
+        )
+
+        packages = [str(m.package) for m in self.data.verify]
```
LecrisUT: Why do you make the `.package` be a `Package` if you do not use it anywhere?

tmt/steps/prepare/verify_installation.py
thrix: Good question. Looking at the prepare step code (`__init__.py` lines 486-495), when a plugin returns `outcome.result` with FAIL results but no exceptions, the results are saved AND `self.preparations_applied += 1` is incremented (the preparation is counted as successful). Only `outcome.exceptions` entries trigger `PrepareError`.

So this seems like a design choice rather than a bug - the framework treats FAIL results as informational, and only exceptions as fatal. For this specific plugin though, FAIL should be fatal. Raising `PrepareError` after collecting all failures would be the right approach here.

tmt/package_managers/dnf.py
thrix: Verified experimentally on Fedora with dnf5 - this is more severe than initially stated:

```
$ dnf5 repoquery --installed --queryformat '%{name} %{from_repo}\n' bash
bash 9f8cc11c5319459e848dd72cc876329c

$ dnf5 info --installed bash | grep repo
From repository : 9f8cc11c5319459e848dd72cc876329c
```

`%{from_repo}` returns a UUID hash on dnf5, not a human-readable repo name like `fedora`. So `expected_repo: fedora` would **always fail** on dnf5 systems. This may be container-specific (repo metadata cleaned in base image), but since the tests use `provision: how: container`, it's directly relevant.

tmt/steps/prepare/verify_installation.py
LecrisUT: @happz @psss thoughts? This seems counter to the effort of using the `PhaseOutcome` and preparing nicer failure messages for the user.

happz: Seems fishy, let me check the code & get back to you.

happz: Right, this is by design, and it seems like a gap:

* when the plugin crashes and raises an exception, the exception is turned into `ERROR` result. The queue stops there, `queue.run()` does not continue - this is what you don't see.
* when the plugin returns, even if it returns failed results, or error results, queue does not detect this, it would continue with the next task.

So, this calls for a fix in `prepare` and `finish` steps, to better handle plugins that failed. Turn `continue`s into `break`s, to explicitly see no other phase runs. And a check which results were produced by the plugin, and stop on `FAIL` and `ERROR` (and `WARN`?). I can make the patch, it's my mess after all.

LecrisUT: > And a check which results were produced by the plugin, and stop on `FAIL` and `ERROR` (and `WARN`?)

Sounds good, just `FAIL` and `ERROR`. Would be nice if these also correctly follow the `tmt run` exit-code.

happz: That's handled elsewhere, and they would turn into a `ProvisionError` so they should lead to the same outcome as plugins ending with an exception.

LecrisUT: Ooph. It is what it is. Wanted to avoid the different ways that oculus is displaying issues when it is an error vs fail, and when it has an outcome or not. But that's for a later improvement once we have more reliable results.yaml

happz: https://github.com/teemtee/tmt/pull/4667

vaibhavdaren: Fixed by appending to outcome.exceptions alongside the FAIL PhaseResult entries, which causes the prepare step to abort as expected. Added a `FIXME` note referencing #4667 — once that lands, the explicit exception should no longer be needed since the prepare step will stop the run on FAIL outcomes directly.

tmt/package_managers/dnf.py
vaibhavdaren: @thrix : packages installed during a container image build via kiwi get a 32-character hex UUID as from_repo because kiwi uses UUID-named repo section IDs and discards the name→UUID mapping after the build. 
Added Dnf5.get_installed_repos override that normalises any 32-char hex UUID to  @System, consistent with how other package managers signal "source cannot be determined". Packages installed freshly (not pre-baked) still return human-readable repo names correctly.

**Note**: The purpose of the plugin is to tests the packages in required block are installed from the designated artifact repos. 
What should be the behaviour of the new plugin if a package is pre-installed , it checks for `@system` for now.
Is there something better we could do for this case here?

Up for suggestions @LecrisUT @happz.

vaibhavdaren:
Addressed comments from code review in f128998fdc3808c42a7a4fd560efc60e6c4c5d1b.

tmt/package_managers/dnf.py
LecrisUT: We do not control pre-installed packages, and this would only show up in the actual-repo. There are plenty of special formats that the from_repo can have, e.g. it can show up as `<unknown>` due to a known bug in the system upgrade. I would say we just display it however it is presented. It is more on dnf side to make sure a sensible format is provided and we can improve the report based on further feedback and suggestions

tmt/package_managers/__init__.py
```diff
@@ -222,6 +222,16 @@ def list_packages(self, repository: "Repository") -> ShellScript:
         """
         raise NotImplementedError
 
+    def get_installed_repos(self, packages: Iterable[str]) -> Command:
```
happz: "get installed repos" - I'd expect it to return a list of installed repositories. How about `get_package_origin` or something similar to indicate the goal is to find out from where was a package installed?

tmt/package_managers/__init__.py
```diff
@@ -222,6 +222,16 @@ def list_packages(self, repository: "Repository") -> ShellScript:
...
+        """
+        Get a command to query the repository each package was installed from.
+
+        :param packages: Package names to query.
+        :returns: A command whose stdout lists ``name repo`` per line for each installed package.
```
happz: If you force the command to share this info on its stdout, and the actual format, is it necessary to return the whole command, instead of e.g. tuples of package and its origin?

tmt/package_managers/__init__.py
happz: Hmmm, different take: why command, why not `ShellScript`? The purpose of an engine is to construct commands, not to run them. Commands may be delayed, executed via a containerfile, etc.

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,26 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        """
+        all_packages = list(packages)
+        command = self.engine.get_installed_repos(all_packages)
+        output = self.guest.execute(command)
+        result: dict[str, Optional[str]] = dict.fromkeys(all_packages)
+        for line in (output.stdout or '').strip().splitlines():
+            parts = line.split()
+            if len(parts) == 2:
+                result[parts[0]] = parts[1]
+            elif parts:
```
happz: Shouldn't this be `else`? `parts` makes it needlessly constrained - unlikely to not be true, but why even mention it when we want to emit a warning in case `parts` does not have two items?

tmt/package_managers/dnf.py
```diff
@@ -283,6 +293,18 @@ class Dnf5Engine(DnfEngine):
...
+#: On dnf5, packages installed during a container image build via kiwi store a 32-character
+#: hex UUID as ``from_repo`` instead of a human-readable repo name.  The UUID was the repo
+#: section ID used by kiwi's temporary repo files, whose name→UUID mapping is discarded after
+#: the build.  We normalise these to ``DEFAULT-SYSTEM-REPO`` to indicate that the package
+#: was installed as part of the base system image and its source cannot be determined.
+_DNF5_UUID_RE = re.compile(r'^[0-9a-f]{32}$')
+
+#: Sentinel repo name used for packages whose installation source cannot be determined
+#: (e.g. packages baked into a kiwi container image).
+DEFAULT_SYSTEM_REPO = 'DEFAULT-SYSTEM-REPO'
```
happz: There's hardly a default system repo. Which one on, a RHEL box, `AppStream`, or `BaseOS`, is the default one? I'd recommend marking it as something said kiwi did, "mangled by kiwi" or something similar.

tmt/package_managers/dnf.py
```diff
@@ -292,10 +314,20 @@ class Dnf5(Dnf):
...
+            pkg: (DEFAULT_SYSTEM_REPO if repo is not None and _DNF5_UUID_RE.match(repo) else repo)
+            for pkg, repo in result.items()
+        }
+
 
 class YumEngine(DnfEngine):
     _base_command = Command('yum')
 
+    def get_installed_repos(self, packages: Iterable[str]) -> Command:
+        raise NotImplementedError
```
happz, line 370: Why? A comment would be nice, to explain a reader why this is not supported with yum.

tmt/schemas/prepare/verify-installation.yaml
```diff
@@ -0,0 +1,55 @@
...
+        - package
+        - expected-repo
+      properties:
+        package:
+          type: string
+        expected-repo:
+          type: string
+    description: |
+      List of package and expected repository mappings, each as a dict
+      with ``package`` and ``expected-repo`` keys.
```
happz:
```yaml
how: verify-installation
verify:
  - foo: AppStream
  - bar: BaseOS
  ...
```

Wouldn't this be simpler?

```yaml
how: verify-installation
verify:
  foo: AppStream
  bar: BaseOS
  ...
```

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,203 @@
...
+
+        try:
+            installed_repos = guest.package_manager.get_installed_repos(
+                m.package for m in self.data.verify
+            )
+        except NotImplementedError as err:
+            raise tmt.utils.PrepareError(
+                f"Package source verification not supported for "
+                f"'{guest.facts.package_manager}' package manager."
+            ) from err
```
happz: It's very nice to see a plugin return proper result as the one below, why not doing so here as well?

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,203 @@
...
+        except tmt.utils.RunError as err:
+            outcome.results.append(
+                PhaseResult(
+                    name=self.name,
+                    result=ResultOutcome.ERROR,
+                    note=[f"Failed to query package repositories: {err}"],
```
happz: Use `tmt.utils.render_exception_as_notes()` for the correct rendering.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,203 @@
...
+                    f" '{verify_mapping.expected_repo}', but the package is not installed"
+                    f" or its source repository could not be determined."
+                )
+            else:
+                note = (
+                    f"Package '{verify_mapping.package}': expected repo"
+                    f" '{verify_mapping.expected_repo}', actual '{actual_repo}'."
+                )
+
+            outcome.results.append(
```
happz, line 134: Since we emit a result per failed package, we should probably emit a pass per successfully verified package too.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,203 @@
...
+                PhaseResult(
+                    name=verify_mapping.package,
```
happz: `name` should include phase name, i.e. `self.name`. `f'{self.name} / {verify_mapping.package}'` would work very well.

tmt/package_managers/dnf.py
vaibhavdaren: Changed to command in the subsequent commit, looks much clean. Thanks for the suggestion.

tmt/steps/prepare/verify_installation.py
vaibhavdaren: Added the result for the pass case as well.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,193 @@
...
+            color='green',
+        )
+
+        try:
+            package_origins = guest.package_manager.get_package_origin(self.data.verify.keys())
+        except (NotImplementedError, tmt.utils.RunError) as err:
+            error: Exception = (
+                tmt.utils.PrepareError(
+                    f"Package source verification not supported for "
+                    f"'{guest.facts.package_manager}' package manager."
```
thrix, line 89: Agreeing with happz's earlier comment — the `NotImplementedError` handler raises `PrepareError` directly, while the `RunError` handler below properly emits a `PhaseResult(result=ERROR)` and appends to `outcome.exceptions`. The inconsistency means `NotImplementedError` is treated as an unhandled plugin crash by the prepare step framework (generic "Plugin raised an unhandled exception" result at `prepare/__init__.py:474`), instead of a clean error with a descriptive message.

For consistency, handle it the same way as `RunError`:
```python
except NotImplementedError:
    msg = (
        f"Package source verification not supported for "
        f"'{guest.facts.package_manager}' package manager."
    )
    outcome.results.append(
        PhaseResult(
            name=self.name,
            result=ResultOutcome.ERROR,
            note=[msg],
            guest=ResultGuestData.from_guest(guest=guest),
        )
    )
    outcome.exceptions.append(tmt.utils.PrepareError(msg))
    return outcome
```

Generated-by: Claude Code

thrix: Correction after verification: the user DOES see the descriptive error message in the log output via `_record_exception` → `logger.fail()`. The improvement is specifically about the `PhaseResult.note` — it would contain the descriptive message instead of the generic "Plugin raised an unhandled exception." This matters for CI systems that display results but not full logs.

Generated-by: Claude Code

vaibhavdaren: Ammened the commit : https://github.com/teemtee/tmt/pull/4660/commits/fea2349bf742c06ca406f9a446fae8091251c39c. Addressed comment by happz and thrix.

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,193 @@
...
+        default=tmt.steps.PHASE_ORDER_PREPARE_VERIFY_INSTALLATION,
+        help='Order in which the phase should be handled.',
+    )
+
+    verify: dict[str, str] = field(
+        default_factory=dict,
+        help="Mapping of package names to expected source repository names.",
+        normalize=_normalize_verify_mappings,
+        serialize=lambda d: d,
+        unserialize=lambda data: cast(dict[str, str], data),
```
LecrisUT: These are meaningless in this form

tmt/package_managers/__init__.py
LecrisUT: Still relevant. See `List packages available` vs `Return a shell script...`. The latter is too literal to what it does, while the former describes the purpose of the script. Similarly the `:returns:` in the previous does not explicitly say all that is in its stdout, output etc.

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,26 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+    def get_package_origin(self, packages: Iterable[str]) -> dict[str, Optional[str]]:
+        """
+        Get the repository each package was installed from.
+
+        :param packages: Package names to query.
+        :returns: A mapping of package names to source repository names.
+            Packages not installed or with undetermined source are mapped to ``None``.
+        """
+        all_packages = list(packages)
+        script = self.engine.get_package_origin(all_packages)
```
LecrisUT: Casting to `list` is unnecessary. You can create the dict first and reuse its keys

tmt/package_managers/dnf.py
```diff
@@ -386,10 +411,24 @@ class Dnf5(Dnf):
     probe_command = Command('dnf5', '--version')
     probe_priority = 60
 
+    def get_package_origin(self, packages: Iterable[str]) -> dict[str, Optional[str]]:
+        result = super().get_package_origin(packages)
+        return {
+            pkg: (KIWI_PREBAKED if repo is not None and _DNF5_UUID_RE.match(repo) else repo)
+            for pkg, repo in result.items()
+        }
+
```
LecrisUT: Too complicated for too little value. `KIWI_PREBAKED` is also incorrect because the container is not usually built with kiwi

tmt/schemas/prepare/verify-installation.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  summary:
+    $ref: "/schemas/core#/definitions/summary"
+
+  verify:
+    type: object
+    additionalProperties:
+      type: string
+    description: |
+      Mapping of package names to their expected source repository names.
```
LecrisUT: We don't do descriptions

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,193 @@
...
+    logger: Logger,
+) -> dict[str, str]:
+    """Normalize verify mappings from a dict mapping package names to expected repos."""
+    if not isinstance(value, dict):
+        raise tmt.utils.NormalizationError(
+            key_address, value, "a dict mapping package names to expected repo names"
+        )
+    # ignore[redundant-cast]: mypy infers `dict[Any, Any]` after the isinstance check
+    # while pyright settles for `dict[Unknown, Unknown]`; the cast helps pyright.
+    return {str(k): str(v) for k, v in cast(dict[Any, Any], value).items()}  # type: ignore[redundant-cast]
```
LecrisUT: `normalize_string_dict` already does the `dict[str, str]` normalization. It does extra stuff as well, but I don't mind rather reusing it

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,193 @@
...
+@container
+class PrepareVerifyInstallationData(PrepareStepData):
+    """Data class for verify-installation prepare plugin."""
+
+    @classmethod
+    def pre_normalization(cls, raw_data: tmt.steps._RawStepData, logger: Logger) -> None:
+        super().pre_normalization(raw_data, logger)
+        name = raw_data.get('name')
+        if name is not None and name.startswith(tmt.utils.DEFAULT_NAME):
+            raw_data['name'] = 'verify-installation'
```
LecrisUT: Why though?

tmt/steps/prepare/verify_installation.py
```diff
@@ -0,0 +1,193 @@
...
+            key_address, value, "a dict mapping package names to expected repo names"
+        )
+    # ignore[redundant-cast]: mypy infers `dict[Any, Any]` after the isinstance check
+    # while pyright settles for `dict[Unknown, Unknown]`; the cast helps pyright.
+    return {str(k): str(v) for k, v in cast(dict[Any, Any], value).items()}  # type: ignore[redundant-cast]
+
+
+@container
+class PrepareVerifyInstallationData(PrepareStepData):
+    """Data class for verify-installation prepare plugin."""
```
LecrisUT: We don't use docstrings on the other `StepData`. Would want to figure out how we can use it to know what can be inside it

tmt/package_managers/dnf.py
```diff
@@ -203,6 +204,18 @@ def list_packages(self, repository: Repository) -> ShellScript:
...
+        return (
+            self.command
+            + Command(
+                'repoquery',
+                '--installed',
+                '--queryformat',
+                r'%{name} %{from_repo}\n',
+                *[Package(p) for p in packages],
+            )
+        ).to_script()
```
LecrisUT, line 214: This behaves quite differently on dnf4 apparently
```console
$ podman run -ti centos:stream10
# dnf install -y make
# dnf repoquery --installed --queryformat "%{name} %{from_repo}\n" make dnf
dnf 

make baseos

```

vaibhavdaren:
```bash
[root@eab1c340eb4b /]# cat /etc/os-release 
NAME="CentOS Stream"
VERSION="10 (Coughlan)"
RELEASE_TYPE=stable
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="10"
PLATFORM_ID="platform:el10"
PRETTY_NAME="CentOS Stream 10 (Coughlan)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:centos:centos:10"
HOME_URL="https://centos.org/"
VENDOR_NAME="CentOS"
VENDOR_URL="https://centos.org/"
BUG_REPORT_URL="https://issues.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 10"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
[root@eab1c340eb4b /]# dnf repoquery --installed --queryformat "%{name} %{from_repo}\n" make tree jq dnf
dnf 

jq baseos

make baseos

tree baseos

```

Works the same?

LecrisUT: If you have not noticed, the new-line is implied in the query format and `%{from_repo}` expands to null for the pre-installed container repo

tmt/steps/prepare/verify_installation.py
vaibhavdaren: Oh @LecrisUT, I want a custom name for the plugin , i do not like the default thingy. Is there a recommened way to do this?

happz: Let's stick to the default value for now. Needless complication for MVP.

tmt/package_managers/dnf.py
vaibhavdaren: Thank you so much. I have a much clear picture now. 

| Case | dnf4 `repoquery` output | dnf5 `repoquery` output | tmt result |                                                                                                                                                                                                      
  |---|---|---|---|                                                                                                                                                                                                                                                            
  | Pre-installed in base image | `pkgname ` (empty `from_repo`, 1 token) | `pkgname 9f8cc11c5319459e848dd72cc876329c` (UUID, 2 tokens) | `PREBAKED` |                                                                                                                           
  | Installed by artifact | `pkgname tmt-artifact-shared` (2 tokens) | `pkgname tmt-artifact-shared` (2 tokens) | `"tmt-artifact-shared"` |                                                                                                                                      
  | Not installed | *(not listed in output)* | *(not listed in output)* | `None` → fail with *"package is not installed"* |   
  
  Will be  adding a test in subsequent commit for this.

tmt/package_managers/dnf.py
vaibhavdaren: Yes, changed to https://github.com/teemtee/tmt/pull/4660#discussion_r2927588432

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,26 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        output = self.guest.execute(script)
+        result: dict[str, Optional[str]] = dict.fromkeys(all_packages)
+        for line in (output.stdout or '').strip().splitlines():
+            parts = line.split(maxsplit=1)
+            if len(parts) == 2:
+                result[parts[0]] = parts[1]
+            else:
```
thrix: The `\n` in queryformat `r'%{name} %{from_repo}\n'` appears to be redundant — dnf adds its own newline per result. This causes double newlines in the output, producing empty lines.

Combined with dnf4's behavior of expanding `%{from_repo}` to empty for pre-installed packages (verified by @LecrisUT on CentOS Stream 10), the parsing loop generates spurious warnings. For a 4-package query on dnf4 where one package is pre-installed, I count **4 spurious warns**: 1 for the single-token `from_repo` line, 3 for empty lines between results.

The result dict is actually correct (pre-installed packages stay as `None` from `dict.fromkeys`), but the warnings are noisy and misleading since this is expected behavior.

Fix options:
1. Skip empty lines and don't warn for single-token lines (they're expected on dnf4):
```python
for line in (output.stdout or '').strip().splitlines():
    if not line.strip():
        continue
    parts = line.split(maxsplit=1)
    if len(parts) == 2:
        result[parts[0]] = parts[1]
```
2. Remove the trailing `\n` from the queryformat if dnf guarantees its own newlines.

Generated-by: Claude Code

tests/prepare/verify-installation/data-fedora/main.fmf
```diff
@@ -0,0 +1,53 @@
+/plan/success:
+    discover:
+        how: fmf
+        test: /test/success
+    provision:
+        how: container
+    prepare:
+      - how: artifact
+        provide:
+          - koji.build:KOJI_BUILD_ID
```
LecrisUT, line 12: Use inheritance to simplify all of this

tests/prepare/verify-installation/data/main.fmf
```diff
@@ -0,0 +1,53 @@
...
+        how: tmt
+
+/test/failure:
+    test: /bin/false
+    duration: 5m
+    summary: Verify package source verification fails with wrong repo
+    require:
+      - make
+      - make-devel
+      - diffutils
```
LecrisUT: Reuse `/test/success` instead. We do not want the failure to come from the test failing. It must fail before that guaranteed.

tests/prepare/verify-installation/data/main.fmf
```diff
@@ -0,0 +1,53 @@
...
+            make: tmt-artifact-shared
+            make-devel: tmt-artifact-shared
+            diffutils: fedora
+    execute:
+        how: tmt
+
+/test/success:
+    test: /bin/true
+    duration: 5m
+    summary: Verify successful package source verification
```
LecrisUT: Drop fields that do not contribute

tmt/package_managers/__init__.py
LecrisUT: > he `\n` in queryformat `r'%{name} %{from_repo}\n'` appears to be redundant — dnf adds its own newline per result. This causes double newlines in the output, producing empty lines.

It does not. That is only from `dnf < 5` afaict. Haven't checked dnf3 and yum of course

> The result dict is actually correct (pre-installed packages stay as `None` from `dict.fromkeys`), but the warnings are noisy and misleading since this is expected behavior.

No, this mixes both pre-installed on dnf4 with not installed in general which would not be reported during `repoquery`

tmt/package_managers/dnf.py
LecrisUT: Test would be difficult. A mock unit test for this does not tell us anything because we are hard-coding what we *believe* would be the output. The only reliable test to do here is to expand the testing to cover centos as well.

tmt/package_managers/dnf.py
LecrisUT: `KIWI_PREBAKED` -> `PREBAKED` still does not help. It is still too complicated for little value. If the from-repo is some weird UUID, fine we just report that. It should be dnf's responsibility to report it as it desires. If they do not do that because there is valuable information in that hash, why do we hide it on our side then?

tests/prepare/verify-installation/test.sh
```diff
@@ -0,0 +1,52 @@
...
+
+        rlRun -s "tmt run -i \$run/failure --scratch -vvv --all \
+            plan --name /plan/failure \
+            provision -h \$PROVISION_HOW --image \$TEST_IMAGE_PREFIX/\$image_name" 2 "Verification should fail with wrong repo"
+
+        rlAssertGrep "3 packages" $rlRun_LOG
+        rlAssertGrep "fail verify-installation / make-devel" $rlRun_LOG
+        rlAssertGrep "expected repo 'SOME_NON_EXISTENT_REPO'" $rlRun_LOG
+        rlAssertGrep "actual 'tmt-artifact-shared'" $rlRun_LOG
+        rlAssertNotGrep "All packages verified successfully." $rlRun_LOG
```
LecrisUT: With the report of the successful ones as well, do assergrep for those as well

tmt/package_managers/__init__.py
LecrisUT: Apparently it is true in weird circumstances: https://github.com/teemtee/tmt/pull/4660#discussion_r2926060708. A better handling is needed

tmt/schemas/prepare/verify-installation.yaml
thrix: Outdated — schema has been redesigned to use a plain dict, this naming concern no longer applies.

tmt/package_managers/dnf.py
thrix: Outdated — dnf5/yum compatibility has been addressed in subsequent commits with proper handling for pre-installed packages.

tmt/steps/prepare/verify_installation.py
thrix: Outdated — the normalization code has been redesigned.

tmt/steps/prepare/verify_installation.py
thrix: Addressed by author in commit fea2349 — NotImplementedError is now handled consistently with RunError, emitting a proper PhaseResult with ERROR outcome.

tmt/schemas/prepare/verify-installation.yaml
LecrisUT: @thrix please hit the resolve button, not a comment that floods the notifications

tmt/package_managers/dnf.py
vaibhavdaren: @LecrisUT Thanks for the review! :) 

If a package is pre-installed how should the new plugin address that? 
**Note** that : For `dnf5` it shows a UUID but for `dnf4` it returns an empty string.

If the usecase for the plugin is that we want to support packages that were installed by the artifact plugin only.
Then this will work, we can just ignore the pre installed stuff for now.

LecrisUT: > If a package is pre-installed how should the new plugin address that?
> **Note** that : For `dnf5` it shows a UUID but for `dnf4` it returns an empty string.

Good question. For `dnf5` it seems fine with whatever is provided, but `dnf4` is more questionable. Something to indicate `unknown` I would say. `dnf5` uses `<unknown>` but not sure under what circumstances. Maybe something like `[unknown]` to slightly distinguish from the other.

tests/prepare/verify-installation/dnf4/test.sh
```diff
@@ -0,0 +1,51 @@
...
+            plan --name /plan/success \
+            provision -h \$PROVISION_HOW --image \$TEST_IMAGE_PREFIX/centos/stream10/upstream:latest" \
+            0 "Run verification test with correct repos"
+
+        rlAssertGrep "pass .* / make" $rlRun_LOG
+        rlAssertGrep "pass .* / diffutils" $rlRun_LOG
+        rlAssertGrep "pass .* / docker-ce-cli" $rlRun_LOG
+        rlAssertGrep "All packages verified successfully." $rlRun_LOG
+        rlAssertGrep "1 test passed" $rlRun_LOG
+    rlPhaseEnd
```
LecrisUT: Please not like this with so much duplication. If needed, please use anything in [epel10](https://bodhi.fedoraproject.org/updates/?search=&releases=EPEL-10.1&releases=EPEL-10.2&releases=EPEL-10.3&page=1) and include epel feature. Can try `centpkg` for example. It doesn't use or test `brew` but that would be fine for now.

vaibhavdaren: Addressed in a1ea9d1683c32999c10a2cfac750b74590738dab

tmt/package_managers/dnf.py
vaibhavdaren: Addressed in e6afc7538469ee33c64fa673bdaa241723fb0ad9 and a1ea9d1683c32999c10a2cfac750b74590738dab.

tests/prepare/verify-installation/dnf4/test.sh
LecrisUT: It is still unnecessarily split between dnf4 and dnf5. Unify them and try to make it similar to `/tests/prepare/install`. Some `adjust` using the `distro` and such would be better

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,34 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        for line in (output.stdout or '').strip().splitlines():
+            # dnf4 produces a blank line after each entry due to the explicit \n
+            # in --queryformat combined with its own record separator.
+            # dnf5 does not have this issue and outputs one line per package.
+            if not line.strip():
+                continue
+            parts = line.split(maxsplit=1)
+            if len(parts) == 2:
+                result[parts[0]] = parts[1].strip()
+            elif len(parts) == 1 and parts[0] in result:
```
LecrisUT: `parts[0] in result` is unnecessary in various ways. Also add comments on what each of these cases represent

tmt/schemas/prepare/verify-installation.yaml
thrix: @LecrisUT yes, will do that more going forward, still learning.

tmt/package_managers/__init__.py
thrix: Outdated, resolving

thrix:
@vaibhavdaren btw tests are still failing hard? or will that be a focus at the end? Anyway, LGTM from my end now see below :)

thrix (approved):

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,35 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        Get the repository each package was installed from.
+
+        :param packages: Package names to query.
+        :returns: A mapping of package names to source repository names.
+            Packages not installed are mapped to ``None``.
+        """
+        result: dict[str, Optional[str]] = dict.fromkeys(packages)
+        script = self.engine.get_package_origin(result.keys())
+        output = self.guest.execute(script)
+        for line in (output.stdout or '').strip().splitlines():
```
happz, line 396: There are some loose ends I can't sort out:
 
* the dnf5/dnf4 notes below in the base class implementation suggest that dnf4 probably needs its own custom implementation, an implementation aware of dnf4 quirks.
* but then again, what's the contract between `engine.get_package_origin()` and its callers, what's the expected output? A line per package, in the `package origin` format - if the dnf4 implementation of this method sneaks in empty lines, it's violating the expectations. This may be problem for commands building images: engine is asked to create a command with expected output, the command would be then executed in container image build time, and the dnf4 command would produce unexpected output, possibly breaking further processing. Maybe we could extend the contract, to make it more liberal: a line per package, in the `package origin` format, empty lines are ignored. And we no longer need to worry about dnf4 in the base class, dnf4 engine would produce valid script.
* the "not installed via a repo" might be more common, other package managers may support installation of packages from files, therefore this might be worth an extension to the `get_package_origin()` contract: a line per package, in the `package origin` format, empty lines are ignored, `origin` is optional and if not specified, it means "unknown".

After this, we would get a simpler base implementation, fewer notes about dnf4 quirks in the base class, nicer protocol for `get_package_origin()`, and the need for a custom dnf4 implementation would be gone.

tmt/package_managers/__init__.py
```diff
@@ -338,6 +348,35 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+            # dnf5 does not have this issue and outputs one line per package.
+            if not line.strip():
+                continue
+            parts = line.split(maxsplit=1)
+            if len(parts) == 2:
+                # Expected format: "package-name repo-name"
+                result[parts[0]] = parts[1].strip()
+            elif len(parts) == 1:
+                # dnf4: empty %{from_repo} means package was not installed via a repo
+                result[parts[0]] = "[unknown]"
```
happz: Deserves to be a module-level constant.

tmt/package_managers/__init__.py
```diff
@@ -338,6 +361,29 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
 
+    def get_package_origin(self, packages: Iterable[str]) -> dict[str, Optional[str]]:
+        """
+        Get the repository each package was installed from.
+
+        :param packages: Package names to query.
+        :returns: A mapping of package names to source repository names.
+            Packages not installed are mapped to ``None``. Packages whose
+            source repository is unknown are mapped to
+            :py:data:`UNKNOWN_PACKAGE_ORIGIN`.
```
happz: This might be unfortunate, trying to squeeze multiple messages into a single string value - `None` as "not installed", `"<unknown>"` means "unknown origin", or an actual origin. Which is a string, just like the magic string that means "unknown".

I'd try to experiment a bit with an enum or something similar, so we can convey the special cases without mistaking them for package origin by accident. Something like this:

```python
class SpecialPackageOrigin(enum.Enum):
    NOT_INSTALLED = 'not-installed'
    UNKNOWN = 'unknown'

PackageOrigin: TypeAlias = Union[str, SpecialPackageOrigin]

# initialize all keys to `NOT_INSTALLED`, some would be overwritten by the actual repo name, some with `UNKNOWN`.


actual_origin = package_origins.get(package)

if actual_origin == expected_repo:
    ... pass, all green

failed_packages.append(package)

if actual_origin == SpecialPackageOrigin.NOT_INSTALLED:
    ...

elif actual_origin == SpecialPackageOrigin.UNKNOWN:
    ...

else:
    raise PrepareError("wow, unexpected enum member")
```

vaibhavdaren: Thanks for the suggestion, @happz . It makes sense to have a function that returns well-defined values such as `UNKNOWN` and `NOT_INSTALLED`.
The previous design was inconsistent, returning a mix of strings (e.g., `<unknown>`), `None`, and other values.
This has been addressed in c8a6f3ca3271a708270f3fd276e44c82216713fe.

tmt/package_managers/__init__.py
```diff
@@ -338,6 +373,32 @@ def list_packages(self, repository: "Repository") -> list[str]:
...
+        script = self.engine.get_package_origin(result.keys())
+        output = self.guest.execute(script)
+        for line in (output.stdout or '').strip().splitlines():
+            # Empty lines are allowed by the engine contract.
+            if not line.strip():
+                continue
+            parts = line.split(maxsplit=1)
+            package = parts[0]
+            # Omitted origin field → unknown source repository.
+            result[package] = parts[1].strip() if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
```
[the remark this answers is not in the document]
vaibhavdaren: split() without a separator never produces an all-whitespace token

tests/prepare/verify-installation/test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+            # Copy plan data and substitute the dynamic build ID
+            rlRun "cp -r data-fedora/. \$data_dir/" 0 "Copy test data"
+            rlRun "sed -i 's/KOJI_BUILD_ID/${KOJI_BUILD_ID}/g' \$data_dir/main.fmf" 0 "Substitute koji build ID"
+            rlRun "pushd \$data_dir"
+        else
+            build_container_image "centos/stream10/upstream:latest"
+            rlRun "image=\$TEST_IMAGE_PREFIX/centos/stream10/upstream:latest"
+            rlRun "pushd data-centos"
+        fi
```
LecrisUT, line 29: We would want to do the same for centos here, same `get_koji_build_id` and everything. At most the package name would differ, but even then you can use `centpkg` for both of them.

Just to get this PR out of the gate sooner, fine with deferring it, but please track this (and should address the growing pile soon)

tmt/package_managers/__init__.py
```diff
@@ -10,6 +11,20 @@
...
 from tmt.utils import Command, CommandOutput, GeneralError, Path, PrepareError, ShellScript
 
+
+class SpecialPackageOrigin(str, enum.Enum):
+    """
+    Sentinel values used in place of an actual repository name to convey
+    special package states returned by :py:meth:`PackageManager.get_package_origin`.
+    """
+
+    # Package is not installed on the guest.
```
```suggestion
    #: Package is not installed on the guest.
```
Iirc, these are docuemntable out-of-the-box

tmt/package_managers/__init__.py
```diff
@@ -222,6 +240,25 @@ def list_packages(self, repository: "Repository") -> ShellScript:
...
+        List source repositories for each installed package.
+
+        The script must emit one line per package in the format::
+
+            <name> <origin>
+
+        Empty lines are allowed and will be ignored by the caller.  If
+        the origin field is omitted the package is treated as having an
+        unknown source repository (equivalent to
+        :py:attr:`SpecialPackageOrigin.UNKNOWN`).
```
LecrisUT: If you will document it, document also that lines missing are treated as `NOT_INSTALLED`

tmt/package_managers/__init__.py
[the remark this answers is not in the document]
LecrisUT: By the same logic, `strip()` here is unnecessary

tests/prepare/verify-installation/test.sh
vaibhavdaren: Added in https://github.com/teemtee/tmt/issues/4646 as a requirement. I think we could have gone with simpler tests like https://github.com/teemtee/tmt/commit/f267ffb6bc07e221c5e31ded1e531fc86bbb8829 for this PR, just to keep it simpler. 
 
But all this pre-work would surely help us in the future.

LecrisUT: The reason for the simplification is that it brings us closer to unifying the tests in `/tests/prepare/artifact` as well and de-duplication once we can fixutrize them. Making sure the tests are as generic as possible will make those transitions all the more easier.

Issue #4531, opened 2026-01-29
Implement full loading of the recipe


therazix:
Implement full loading of run data, plans, and tests from the `recipe.yaml` file.

PR #4661, opened 2026-03-09
Add support for full recipe loading

therazix:
This PR implements a full recipe loading feature. All phases can now be loaded directly from the recipe. Serialization and deserialization were replaced with `to_spec/from_spec` methods, and the generated recipe will now contain only non-empty values to reduce its size.

Resolves: #4531 

Pull Request Checklist

* [x] implement the feature
* [ ] write the documentation
* [ ] adjust plugin docstring
* [ ] modify the json schema
* [ ] mention the version
* [x] include a release note

### DIFF
--- a/tmt/recipe.py
+++ b/tmt/recipe.py
@@ -276,15 +276,10 @@ class _RecipeStep(SpecBasedContainer[_RawRecipeStep, _RawRecipeStep], Serializab
 
     @classmethod
     def from_step(cls, step: 'Step') -> '_RecipeStep':
-        try:
-            step_data = step.data
-        except tmt.utils.SpecificationError:
-            # Do not save invalid phases
-            step_data = []
-
+        enabled = bool(step.enabled)
         return _RecipeStep(
-            enabled=bool(step.enabled),
-            phases=[phase.to_minimal_spec() for phase in step_data],
+            enabled=enabled,
+            phases=[phase.to_minimal_spec() for phase in step.data] if enabled else [],
         )
 
     # ignore[override]: does not match the signature on purpose, we need to pass logger
@@ -334,16 +329,11 @@ class _RecipeDiscoverStep(_RecipeStep):
 
     @classmethod
     def from_step(cls, step: 'Step') -> '_RecipeDiscoverStep':
-        try:
-            step_data = step.data
-        except tmt.utils.SpecificationError:
-            # Do not save invalid phases
-            step_data = []
-
         assert isinstance(step, Discover)
+        enabled = bool(step.enabled)
         return _RecipeDiscoverStep(
-            enabled=bool(step.enabled),
-            phases=[phase.to_minimal_spec() for phase in step_data],
+            enabled=enabled,
+            phases=[phase.to_minimal_spec() for phase in step.data] if enabled else [],
             tests=[_RecipeTest.from_test_origin(test_origin) for test_origin in step.tests()],
         )
 
@@ -374,16 +364,20 @@ class _RecipeExecuteStep(_RecipeStep):
 
     @classmethod
     def from_step(cls, step: 'Step') -> '_RecipeExecuteStep':
-        try:
-            step_data = step.data
-        except tmt.utils.SpecificationError:
-            # Do not save invalid phases
-            step_data = []
+        enabled = bool(step.enabled)
         return _RecipeExecuteStep(
-            enabled=bool(step.enabled),
-            phases=[phase.to_minimal_spec() for phase in step_data],
+            enabled=enabled,
+            phases=[phase.to_minimal_spec() for phase in step.data] if enabled else [],
             results_path=(step.step_workdir / 'results.yaml').relative_to(step.run_workdir),
         )
+
+    def to_fmf_spec(self) -> list[_RawStepData]:
+        spec = super().to_fmf_spec()
+        # Return the default execute phase if no phases were provided.
+        # This is needed so that the plan is discovered correctly.
+        if not spec:
+            return [_RawStepData(how='tmt')]
+        return spec
 
 
 @container
--- a/tests/recipe/data/import.yaml
+++ b/tests/recipe/data/import.yaml
@@ -69,59 +69,20 @@ plans:
             result: respect
     provision:
         enabled: false
-        phases:
-          - name: default-0
-            how: local
-            order: 50
-            become: false
+        phases: []
     prepare:
         enabled: false
-        phases:
-          - name: enable-latest-copr
-            how: install
-            order: 50
-            summary: Enable copr repo with the latest tmt bits
-            copr:
-              - '@teemtee/latest'
-            missing: fail
-          - name: disable-tag-repo
-            how: shell
-            order: 50
-            summary: Completely disable the tag repository
-            script:
-              - "dnf repolist | grep testing-farm-tag-repository || exit 0\nif command
-                -v dnf5; then\n    dnf config-manager setopt testing-farm-tag-repository.enabled=0\n\
-                else\n    dnf config-manager --set-disabled testing-farm-tag-repository\n\
-                fi\n"
+        phases: []
     execute:
         enabled: false
-        phases:
-          - name: default-0
-            how: tmt
-            order: 50
-            duration: 1h
-            ignore-duration: false
-            exit-first: false
-            interactive: false
-            restraint-compatible: false
-            no-progress-bar: false
+        phases: []
         results-path: plans/import/execute/results.yaml
     report:
         enabled: false
-        phases:
-          - name: default-0
-            how: display
-            order: 50
-            display-guest: auto
+        phases: []
     finish:
         enabled: false
-        phases:
-          - name: default-0
-            how: shell
-            order: 50
+        phases: []
     cleanup:
         enabled: false
-        phases:
-          - name: default-0
-            how: tmt
-            order: 50
+        phases: []
--- a/tests/recipe/load.sh
+++ b/tests/recipe/load.sh
@@ -20,6 +20,7 @@ rlJournalStart
         rlAssertGrep "pass /discover-shell/shell-test" $rlRun_LOG
         rlAssertGrep "summary: 3 tests executed" $rlRun_LOG
         rlAssertGrep "script: echo \"Finish phase\"" $rlRun_LOG
+        rlAssertGrep "total: 3 tests passed" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "Test recipe loading with remote repository"
@@ -27,10 +28,11 @@ rlJournalStart
         rlAssertGrep "summary: 1 test selected" $rlRun_LOG
         rlAssertGrep "pass /tests/one" $rlRun_LOG
         rlAssertGrep "summary: 1 test executed" $rlRun_LOG
+        rlAssertGrep "total: 1 test passed" $rlRun_LOG
     rlPhaseEnd
 
     rlPhaseStartTest "Test recipe loading of imported plan"
-        rlRun -s "tmt run -vvv --scratch --id $run --recipe import.yaml discover"
+        rlRun -s "tmt run -vvv --scratch --id $run --recipe import.yaml"
         rlAssertGrep "summary: 2 tests selected" $rlRun_LOG
     rlPhaseEnd
 


tmt/recipe.py
```diff
@@ -425,7 +442,7 @@ def tests(self, recipe: Recipe, plan_name: str) -> list[TestOrigin]:
             if plan.name == plan_name:
                 return [
                     TestOrigin(
-                        phase=test.discover_phase,
+                        phase=test.path.unrooted().parts[0],
```
LecrisUT: Uph, this is fragile. Is it really needed?

tmt/recipe.py
```diff
@@ -58,18 +50,16 @@ class _RecipeTest(SerializableContainer):
     restart_max_count: int
     restart_with_reboot: bool
     serial_number: int
-    discover_phase: str
     link: Optional['tmt.base.core.Links'] = field(
         serialize=lambda value: value.to_spec() if value else [],
-        unserialize=lambda value: _unserialize_links(value),
     )
```
thrix: The `unserialize` callback was removed but the type annotation still says `Optional['tmt.base.core.Links']`. When loading from serialized YAML via `from_serialized()`, the raw value (list of dicts) is stored instead of a `Links` object.

This doesn't cause runtime errors in the current save path (which creates from `Plan` objects where `link` IS a `Links` object), but the type mismatch could break if the loaded recipe is ever re-serialized, since the `serialize` callback calls `value.to_spec()` which would fail on raw data.

Same issue exists for `_RecipePlan.link` where `from_serialized` does `link=serialized.get('link')` (raw data).

tmt/schemas/report/display.yaml
```diff
@@ -22,6 +22,9 @@ properties:
   name:
     type: string
 
+  display-guest:
```
thrix: This `display-guest` addition appears unrelated to recipe loading. Should be in a separate commit or PR.

thrix (changes requested):
The full recipe loading direction is good. Three findings, one requiring a change:

1. **`Run.environment` silently ignores CLI `--environment` when recipe is loaded** - The recipe env unconditionally overrides CLI options. New `--environment` overrides on replay are silently lost. This should merge recipe env with CLI env, letting CLI take precedence.

2. **`link` field type mismatch after unserialization removal** - Raw data stored where `Links` object is expected. Not an active runtime bug but incorrect typing that could break on re-serialization paths.

3. **Unrelated schema change** - `display-guest` in `report/display.yaml` should be split out.

skycastlelily:
Besides other comments,  the code appears to assume the recipe file is well-formed, do you consider add some validation to the recipe file provided, say, a schema file?

happz:
@therazix please, set the "Size" of this PR.

tmt/base/core.py
```diff
@@ -3033,6 +3034,9 @@ def environment(self) -> Environment:
           options.
         """
 
+        if self.recipe is not None:
+            return self.recipe.run.environment.copy()
+
```
thrix: This early return completely bypasses `_environment_from_cli`, meaning CLI options like `--environment FOO=bar` are silently ignored when running from a recipe.

While the recipe faithfully captures the *original* run's environment (via `run.environment` in `save()`), any *new* CLI overrides on replay are lost. A user running:
```
tmt run --recipe recipe.yaml --environment FOO=bar
```
would expect `FOO=bar` to be set, but it won't be.

Consider merging instead, letting CLI overrides take precedence:
```python
if self.recipe is not None:
    return Environment({
        **self.recipe.run.environment,
        **self._environment_from_cli,
    })
```

tmt/recipe.py
thrix: Latent bug: the `unserialize` callback was removed, so when loading from YAML via `from_serialized()`, the raw value (list of dicts) is stored instead of a `Links` object.

This is fine when `link` is `[]` (falsy — the serialize callback short-circuits to `[]`), which is the case for all current test data. However, if a recipe test ever has a non-empty `link` (e.g., `[{"relates": "..."}]`), the serialize callback `lambda value: value.to_spec() if value else []` would call `.to_spec()` on a raw list, causing an `AttributeError`.

Either restore the `unserialize` callback or handle raw data in the serialize callback.

tmt/schemas/report/display.yaml
```diff
@@ -22,6 +22,9 @@ properties:
...
+    type: boolean
+
```
thrix: This `display-guest` addition appears unrelated to recipe loading. Should be in a separate commit or PR to keep changes atomic.

thrix (changes requested):
## Review: Add support for full recipe loading

Good progress on extending recipe loading from discover-only to all plan steps. The environment simplification and removal of `discover_phase` are clean.

### Issues

**Blocking:**
- **CLI environment silently ignored** (`tmt/base/core.py`): The early return bypasses `_environment_from_cli`, so `--environment FOO=bar` is silently dropped when using `--recipe`. See inline comment for suggested fix.

**Latent bug:**
- **`_RecipeTest.link` unserialization removed** (`tmt/recipe.py`): Works today because all test links are `[]` (falsy short-circuit), but would `AttributeError` on any recipe with non-empty test links. See inline comment.

**Hygiene:**
- **Unrelated schema change** (`tmt/schemas/report/display.yaml`): `display-guest` addition is not related to recipe loading — should be a separate commit/PR.
- **PR checklist**: All items are unchecked — docs, spec, schema, version, release note still needed.

### Verified non-issues

- **Path traversal in `discover_from_recipe`**: The `relative_to()` + `resolve()` + parent check is sufficient. The gemini-code-assist security warning is overblown.
- **Removed "non-existent plan" test**: Correct — with `tree.children.clear()`, the tree IS the recipe, so the old error case no longer applies.
- **Only saving `_environment_from_fmf` in `_RecipePlan`**: Reasonable — `_RecipeRun.environment` captures the full merged env, and intrinsics should be regenerated per run.

Generated-by: Claude Code

tmt/base/core.py
thrix: Generated-by: Claude Code

tmt/recipe.py
thrix: Generated-by: Claude Code

tmt/schemas/report/display.yaml
thrix: Generated-by: Claude Code

thrix:
@therazix would it be possible to provide a solid MR description for the changes, so it is easier to follow the changes, for example I am looking at this diff:

https://github.com/teemtee/tmt/pull/4661/changes#diff-f3bed0c5c8c7fcc2e8e296e3e85675c9c29c322c3a66f8def8f1d6d6d8dce260

And I would like to understand why this was changed, I would expect it would be mentioned that this is one of the improvements made to support full recipe loading ... (or something similar)

tmt/base/core.py
thrix: looks resolved to me

tmt/guest/__init__.py
```diff
@@ -1328,10 +1328,35 @@ def to_spec(self) -> tmt.steps._RawStepData:
...
 
         spec.pop('facts', None)  # type: ignore[typeddict-item]
+        spec.pop('-OPTIONLESS-FIELDS', None)  # type: ignore[typeddict-item]
         spec['ansible'] = self.ansible.to_spec() if self.ansible else {}  # type: ignore[typeddict-unknown-key]
+        spec['environment'] = self.environment.to_fmf_spec() if self.environment else {}  # type: ignore[typeddict-unknown-key]
+        spec['hardware'] = self.hardware.to_spec() if self.hardware else None  # type: ignore[typeddict-unknown-key]
 
         return spec
 
+    def to_minimal_spec(self) -> tmt.steps._RawStepData:
```
thrix, line 1339: I think even functions like these would deserve comments ...

tmt/steps/discover/__init__.py
```diff
@@ -588,6 +588,14 @@ def discover_from_recipe(self, logger: Optional[tmt.log.Logger] = None) -> None:
             if test_origin.phase == self.name
         ]
 
+        # Recreate test paths if they don't exist
+        for test in self._tests:
+            if test.path is None:
+                continue
+            path = self.test_dir / test.path.unrooted().relative_to(Path(self.safe_name) / 'tests')
+            if not path.exists():
+                path.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
thrix: Sounds like something to rather address?

tmt/recipe.py
```diff
@@ -89,9 +79,17 @@ class _RecipeTest(SerializableContainer):
...
 
+    # ignore[override]: does not match the signature on purpose, we need to pass logger
+    @classmethod
+    def from_serialized(cls, serialized: dict[str, Any], logger: Logger) -> '_RecipeTest':  # type: ignore[override]
+        raw_checks = serialized.pop('check', [])
+        serialized['check'] = []
+        recipe_test = super().from_serialized(serialized)
+        recipe_test.check = [Check.from_spec(check, logger) for check in raw_checks]
+        return recipe_test
```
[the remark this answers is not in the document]
thrix: worth documenting if expected

tmt/recipe.py
```diff
@@ -358,8 +356,27 @@ def from_plan(cls, plan: 'Plan') -> '_RecipePlan':
...
-        return {'discover': self.discover.to_spec()}
+        spec = self.to_dict()
+
+        for e in [
+            'environment_from_fmf',
+            'environment_from_importing',
+            'environment_from_cli',
+            'environment_from_intrinsics',
+        ]:
+            spec.pop(e)
```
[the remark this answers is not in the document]
thrix: sounds reasonable

tmt/steps/provision/connect.py
```diff
@@ -63,6 +63,45 @@ class ConnectGuestData(tmt.guest.GuestSshData):
...
+                **super().to_spec(),
+                'soft-reboot': str(self.soft_reboot)
+                if isinstance(self.soft_reboot, ShellScript)
+                else None,
+                'systemd-soft-reboot': str(self.systemd_soft_reboot)
+                if isinstance(self.systemd_soft_reboot, ShellScript)
+                else None,
+                'hard-reboot': str(self.systemd_soft_reboot)
+                if isinstance(self.systemd_soft_reboot, ShellScript)
+                else None,
```
thrix: `hard-reboot` referencing `systemds-soft-reboot`, does not sound right

tmt/steps/provision/connect.py
```diff
@@ -63,6 +63,45 @@ class ConnectGuestData(tmt.guest.GuestSshData):
...
+        field_map: dict[str, Callable[[Any], Any]] = {
+            'soft-reboot': lambda reboot: str(reboot) if isinstance(reboot, ShellScript) else None,
+            'systemd-soft-reboot': lambda reboot: (
+                str(reboot) if isinstance(reboot, ShellScript) else None
+            ),
+            'hard-reboot': lambda reboot: str(reboot) if isinstance(reboot, ShellScript) else None,
+        }
+
+        for key, transform in field_map.items():
+            value = getattr(self, key, None)
```
thrix: `field_map` items use hyphens, but this is accessing keys of self, which are with underscores?

tmt/recipe.py
thrix: resolving, looks outdated

tmt/schemas/report/display.yaml
thrix: resolving, outdated

tmt/schemas/report/display.yaml
thrix: resolving, outdated

tmt/steps/discover/__init__.py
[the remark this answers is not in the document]
therazix: This has been addressed.
https://github.com/teemtee/tmt/blob/d34c3f438029b4395cd499d6d86bec502d22b10f/tmt/steps/discover/__init__.py#L600-L604

tmt/recipe.py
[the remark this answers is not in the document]
therazix: This has been changed.

tmt/recipe.py
therazix: Test `path` should always be prefixed with the discover phase name when the recipe is generated, so I wanted to avoid duplication. But I agree, I reverted that change and kept the `discover_phase` attribute.

tmt/recipe.py
therazix: This has been addressed. `serialize/unserialize` has been replaced with `to_spec/from_spec`.

tmt/guest/__init__.py
therazix: Added in 62ddebcf5044872cc74f5d648ae79b89b622f4c6.

tmt/steps/provision/connect.py
therazix: Fixed in 62ddebcf5044872cc74f5d648ae79b89b622f4c6.

tmt/steps/provision/connect.py
therazix: Fixed in 62ddebcf5044872cc74f5d648ae79b89b622f4c6.

happz:
Two observations, and I'm open to discussion:

* quite a lot of code is related to addition of `to_spec` and `to_minimal_spec` - this would be nice to see as a standalone PR. There were some questions about how this works recently, refactoring this part of code would be worth on its own, with review not necessarily tied to recipes.
* the number of "crossroads", with "if recipe" and "else" branches, tends to grow. I was hoping recipe handling would follow the path of replacing tmt entities with those loaded from recipe, restoring their state. Every place where we need to look whether we're in the "recipe mode" deserves a hard scrutiny and the utmost prejudice: why this test exists, is it a sign of us not replacing a tmt object completely, is it a temporary solution before the next patch?

  For example, in `tmt.steps.Step.enabled`:

  ```python
  if self.plan.my_run.recipe:
      plan = self.plan.my_run.recipe.get_plan_by_name(self.plan.name)
      step = plan.get_step_by_name(self.step_name)
      return step.enabled
  ``

  What is `self` in this context? Isn't it an instance of `Step` subclass, restored from a recipe, in which case its `enabled` should be correctly set already by the recipe?

  I am not advocating against the existence of such checks, sometimes they might be unavoidable, I would just like to see every such check be explained, preferably with a hefty comment describing why it's needed, what would not work if the check wouldn't exist, and so on. Without a strict approach dscipline we can easily end up maintaining a growing list of tests and exceptions, eventually implement one behavior twice, once for "tmt as usual" and once for "tmt when loaded from recipe".

therazix:
> quite a lot of code is related to addition of `to_spec` and `to_minimal_spec` - this would be nice to see as a standalone PR.

I created a standalone PR #4727.

> the number of "crossroads", with "if recipe" and "else" branches, tends to grow.

I agree. I tried to get rid of them in 207f6c71429dcab89c986cc785d3e17338355c53.

> What is `self` in this context? Isn't it an instance of `Step` subclass, restored from a recipe, in which case its `enabled` should be correctly set already by the recipe?

If I understand it correctly, a step is only enabled when its name is in `_cli_context_object.steps`. There is no attribute we could set during recipe loading to enable or disable the step. I tried inserting the step names directly into `_cli_context_object.steps` in 207f6c71429dcab89c986cc785d3e17338355c53, as it's basically the same as what is invoked during recipe creation, but I'm not sure if it's a good solution. Or should I just add an attribute to the `Step` that the recipe would set during loading?

tmt/base/core.py
```diff
@@ -3032,6 +3032,17 @@ def _environment_from_cli(self) -> Environment:
...
+    @property
+    def _environment_from_recipe(self) -> Environment:
+        """
+        Environment variables from the recipe.
+        """
+
+        if self.recipe is None:
+            return Environment()
+
+        return self.recipe.run.environment.copy()
```
LecrisUT, line 3044: I wonder if the recipe should overwrite the other sources or inject itself somewhere in the list. Don't know what would be correct

tmt/base/core.py
```diff
@@ -3042,13 +3053,15 @@ def environment(self) -> Environment:
...
           options.
         """
 
         return Environment(
             {
                 **self._environment_from_workdir,
+                **self._environment_from_recipe,
                 **self._environment_from_cli,
             }
         )
```
LecrisUT, line 3073: The subclasses like `Plan` overwrite this, is that ok? I guess the recipe environment is only linked to the Run and not each components?

tmt/recipe.py
```diff
@@ -357,51 +448,123 @@ def from_plan(cls, plan: 'Plan') -> '_RecipePlan':
...
+
+        spec.pop('name')
+        spec['discover'] = self.discover.to_fmf_spec()
+        spec['provision'] = self.provision.to_fmf_spec()
+        spec['prepare'] = self.prepare.to_fmf_spec()
+        spec['execute'] = self.execute.to_fmf_spec()
+        spec['report'] = self.report.to_fmf_spec()
+        spec['finish'] = self.finish.to_fmf_spec()
+        spec['cleanup'] = self.cleanup.to_fmf_spec()
+        return {key: value for key, value in spec.items() if value not in (None, [], {})}
```
LecrisUT, line 505: Do we need this one? I do not expect the recipe to be in an `.fmf` file.

tmt/recipe.py
```diff
@@ -357,51 +448,123 @@ def from_plan(cls, plan: 'Plan') -> '_RecipePlan':
...
     def __init__(self, logger: Logger):
         super().__init__(logger=logger)
 
-    def load(self, path: Path) -> Recipe:
-        return Recipe.from_serialized(tmt.utils.yaml_to_dict(self.read(path)), self._logger)
+    def load(
+        self, run: 'Run', recipe_path: Optional[Path], fmf_tree: Optional[Tree]
+    ) -> Optional[Recipe]:
+        if recipe_path is None or fmf_tree is None:
+            return None
```
LecrisUT: Why the `Optional` in here? Would probably prefer for the check to be done on the higher level

tmt/recipe.py
```diff
@@ -434,8 +595,21 @@ def tests(self, recipe: Recipe, plan_name: str) -> list[TestOrigin]:
...
+        """
+        Load the names of enabled steps from the recipe and save them to the CLI context.
+        """
+        if self._cli_context_object is None:
+            return
+        for plan in recipe.plans:
+            for step_name in STEPS:
+                recipe_step = plan.get_step_by_name(step_name)
+                if recipe_step.enabled:
+                    self._cli_context_object.steps.add(step_name)
```
LecrisUT, line 623: This is a curious code. We are missing an interface if we are needed to edit the cli context like that. Can you open a follow-up issue on it?

tmt/recipe.py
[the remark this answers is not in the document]
thrix: @therazix can you look at this one and resolve?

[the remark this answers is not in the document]
therazix: This one is resolved, we no longer modify the `spec` or `serialized` dictionary.

tmt/base/core.py
therazix: As discussed in the review meeting, for now, we should only overwrite the environment. Changed in 7c44e057ff2568cebca4412bad4c8106e6353465.

tmt/base/core.py
therazix: Plan and other classes use the run environment and add their own things, which is expected. This particular case only replaces the run environment. Other environments, like those for `Plan` and `Test`, are loaded directly in the `_update_tree` function.

tmt/recipe.py
therazix: The `to_fmf_spec` methods handle the conversion from the recipe form to the structure that can replace the fmf tree in the `_update_tree` function. I could not think of a better name for them.

tmt/recipe.py
therazix: Changed in 7c44e057ff2568cebca4412bad4c8106e6353465.

thrix (approved):
No more comments from me, LGTM

tmt/recipe.py
therazix: As discussed in the review meeting, we should decouple step enablement from CLI context in the future. Created an issue #4741.

tmt/recipe.py
LecrisUT: :upside_down_face: Welp let's pile it on the review it when we get a pydantic/attrs backend

LecrisUT:
LGTM, ping me when the test issues are resolved

tmt/recipe.py
```diff
@@ -160,94 +270,124 @@ def to_test(self, logger: Logger) -> 'tmt.base.core.Test':
...
+    phases: list[_RawStepData]
 
-    # ignore[override]: does not match the signature on purpose, we need to pass logger
     @classmethod
-    def from_serialized(cls, serialized: dict[str, Any], logger: Logger) -> '_RecipeStep':  # type: ignore[override]
-        enabled = bool(serialized.get('enabled', False))
+    def from_step(cls, step: 'Step') -> '_RecipeStep':
+        try:
+            step_data = step.data
+        except tmt.utils.SpecificationError:
```
happz: This seems like silencing an error. Can this even happen, given that `step.data` should be valid phases? On the other hand, it's dynamically populated from raw data, that may fail, but then we do have invalid phase, e.g. win an unknown key, should the exception be propagated to user?

tmt/base/core.py
```diff
@@ -3044,14 +3044,19 @@ def environment(self) -> Environment:
...
-            }
-        )
+        if self.recipe is None:
+            return Environment(
+                {
+                    **self._environment_from_workdir,
+                    **self._environment_from_cli,
+                }
+            )
+        return self.recipe.run.environment.copy()
```
happz: I'd like to see `_environment_from_recipe` property, aligned with other `_environment_from_*` properties, `environment` would then create new `Environment` from either that, or `_environment_from_workdir` + `_environment_from_cli`.

tmt/recipe.py
therazix: The reason for this is that the test `/tests/steps/invalid` was failing. If a user specifies, for example, an invalid provision phase but selects only the discover step to be executed, recipe generation would fail due to that invalid provision phase, even though it was never executed.

tmt/base/core.py
therazix: Changed in 611ed5dac6218bd79bcdb30d9156a0ea60412b10.

tmt/recipe.py
happz: Shouldn't the recipe creation then also obey whether the step is enabled or not, and therefore avoiding the need to materialize step data of a step that was not requested to run?

LecrisUT: Smells like the https://github.com/teemtee/tmt/issues/4741 topic doesn't it?

happz: Related, it's about enabled steps, but the issue I think I see in this PR is the recipe generation does not consult enabled steps when creating a recipe, and must then silence errors that `tmt run` would not encounter.

LecrisUT: Wouldn't the override done in `_update_cli_context` overwrite (or I think just re-enable) the cli input, nullifying the step selection from the cli?

therazix: > Shouldn't the recipe creation then also obey whether the step is enabled or not, and therefore avoiding the need to materialize step data of a step that was not requested to run?

We can, but we need to keep at least the execute step, even though it's not enabled, because otherwise tmt would load a default plan instead. The execute step is required due to [this line](https://github.com/teemtee/tmt/blob/405e40a7ed38ea5a6c3a9987658fbf878cf546a1/tmt/base/core.py#L2556). And I would like to avoid another "if recipe" case.

happz: > Wouldn't the override done in `_update_cli_context` overwrite (or I think just re-enable) the cli input, nullifying the step selection from the cli?

I'm not sure, as I don't see the connection to the problem here. My worry is about recipe saving disabled steps: if a step was disabled by user for a good reason, e.g. when a plugin needed is not available, such a step would respond with an exception, and recipe would silently replaced it with emptiness. Which I think is not correct behavior.

Maybe the recipe should consume `step._raw_data` instead of `step.data`, that's a list of specs for individual phases, so pretty much ready to be saved, no materialization, no exceptions because of missing plugins. Seems like a fitting answer, I'm just not sure whether we wouldn't hit a failure elsewhere. I'd give this a try. Recipe can also skip disabled steps, or raise an exception. But the `step_data = []` with no message is IMO wrong.

therazix: I made it so that disabled steps are skipped during recipe generation. I also added a default execute phase that is injected during recipe loading if the recipe does not contain one. Otherwise, the plan from the recipe would not be discovered. 3a5b24bd9457f4e488209d532d85fa799d2a06b4

happz:
Eh, it's still far from satisfactory, but that's not @therazix's fault, it's just so complex field, with a tech debt of its own. Hopefully, this wouldn't be the last patch, and together with the transition of `dataclass`-based containers, a lot would clear up eventually.

PR #4662, opened 2026-03-09
Use dedicated temporary directory for all rsync pulls

happz:
This should prevent rsync process from creating temporary files in the destination directory tree. In case multiple rsync commands operate on the same directory structure, they may run into race condition:

* rsync A is pull into the tree, creates a temporary file to hold partially transfered file;
* rsync B is pushing the tree elsewhere, notices the file, and intends to push it;
* rsync A completes the transfer, removes the temporary file;
* rsync B sees the file is gone, and reports an error.

Isolating temporary files between rsync processes and from the manipulated tree should prevent this conflict.

The patch comes with new "tmpdir" attribute to provide run-specific temporary directories in a way compatible with `tempfile.TemporaryDirectory`.

Pull Request Checklist

* [x] implement the feature


tmt/guest/__init__.py
```diff
@@ -3538,17 +3540,20 @@ def pull(
...
-                    "rsync",
-                    *options.to_rsync(),
-                    "-e",
-                    self._ssh_command.to_element(),
-                    f"{self._ssh_guest}:{source}",
-                    destination,
-                ),
-                silent=True,
-            )
+            with tempfile.TemporaryDirectory(prefix='tmt.rsync-') as rsync_tempdir:
```
thrix: `tempfile.TemporaryDirectory()` defaults to `/tmp` (via `tempfile.gettempdir()`). On most Linux systems, `/tmp` and `/var/tmp` are on **different filesystems** (tmpfs vs disk). Verified on Fedora:

```
/tmp device:     41 (tmpfs)
/var/tmp device: 101 (disk)
```

tmt workdirs live under `/var/tmp/tmt/`. From the rsync man page for `--temp-dir`:

> When the scratch directory is on a different disk partition, rsync will not be able to rename each received temporary file over the top of the associated destination file, but instead must **copy it into place**. Rsync does this by copying the file over the top of the destination file, which means that **the destination file will contain truncated data during this copy**.

So the fix trades the original race (temp file vanishes -> rsync error) for a brief window where destination files could have truncated data during copy-in-place.

For small files this window is negligible, but worth being aware of. To avoid this entirely, the temp dir could be created on the same filesystem as the destination (e.g. `dir=str(destination.parent)` or a sibling of the workdir), though that requires care to keep it outside the tree being synced.

happz: Fair point, that is indeed possible. On the other hand, the reader of these files is tmt itself, and while the pull runs, nothing should consume files being copied. So this should be safe.

That said, we should probably prevent this if we can, because it might be unsafe outside of tmt, like in TF that invokes tmt and then keeps syncing its files to cold storage. This may easily happen in the middle of a large file being copied from `--temp-dir` into its destination, presenting incomplete picture to user via web UI.

happz: Addressed in bce2f718.

PR #4667, opened 2026-03-10
Prepare and finish step will now fail when phase fails

happz:
This was not automatic! Both steps failed and stopped running their phases when a phase _raised an exception_ and/or _returned an outcome with an exception attached_. These two conditions are no longer enough as we are expecting plugins that do not have any relevant exception, instead they gracefully deliver results reporting failed outcomes.

The patch teaches `prepare` and `finish` to recognize "negative" outcomes, `FAIL` and `ERROR`, and react as if a phase raised an exception, i.e. by reporting the error and stopping the run.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/tmt/queue.py
+++ b/tmt/queue.py
@@ -404,7 +404,11 @@ class Queue(list[TaskT]):
 
                 yield outcome
 
-            # TODO: make this optional
+            # TODO: with the `self._keep_running` this check is probably
+            # no longer needed, as the responsibility shifts more to the
+            # user of the queue, to decide when to stop. To drop it, we
+            # will need to review uses of `Queue`, which will happen as
+            # part of https://github.com/teemtee/tmt/issues/4668.
             if failed_tasks:
                 return
 


tmt/queue.py
```diff
@@ -376,7 +382,12 @@ def run(self) -> Iterator[TaskT]:
         instance of this class is yielded.
         """
 
-        for i, task in enumerate(self):
+        i = 0
+
+        while self:
+            task = self.pop(0)
```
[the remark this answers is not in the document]
happz: True, but very low priority. We do not expect the queue to host more than very low number of phases, few tens at max. In this range, the difference should be minimal. Filed #4668.

tmt/steps/finish/__init__.py
```diff
@@ -234,9 +243,16 @@ def _record_exception(
 
                         continue
```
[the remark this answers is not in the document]
happz: Addressed in 5124c374.

thrix (approved):
Clean, focused fix. Verified all changes against the actual branch code.

This directly addresses the design gap we identified during #4660 review - FAIL/ERROR results from plugins were silently recorded without stopping the prepare/finish steps. Now `_is_failed()` checks for both exceptions AND negative result outcomes, and `queue.stop()` + `break` halts execution immediately.

Key points verified:
- `_is_failed()` correctly checks `bool(exceptions) or any(result in (ERROR, FAIL))`
- `preparations_applied` only increments after the `_is_failed()` check passes
- The `continue` after exception processing was correctly replaced with fall-through to the failure check
- The queue's `_keep_running` flag acts as a safety net alongside the consumer's `break`
- Final guard changed from `if exceptions` to `if _is_failed()` to catch both failure modes

vaibhavdaren:
Apart from the PR. @thrix - Your code review setup is too cool now. Need a demo in the next hacking session ^_^ .

happz:
> LGTM^^, just a nitpick about duplicate-ish: do you consider create a shared _is_failed function,say: def _is_failed(exceptions: list, results: list) -> bool ,maybe in tmt/steps/**init**.py?

I hope there will be more opportunities to deduplicate code once other steps get on board of producing results.

tmt/steps/finish/__init__.py
```diff
@@ -192,6 +192,12 @@ def _record_exception(
 
                 exceptions.append(exc)
 
+            def _is_failed() -> bool:
+                return bool(exceptions) or any(
+                    result.result in (ResultOutcome.ERROR, ResultOutcome.FAIL)
+                    for result in results
+                )
```
LecrisUT, line 199: We are re-reading the whole results each time. Would be nice to instead have a helper to do the whole logic loop for one `outcome`. Not a blocker though, fixing the current status is of higher priority

tmt/queue.py
```diff
@@ -396,3 +407,17 @@ def run(self) -> Iterator[TaskT]:
             # TODO: make this optional
             if failed_tasks:
                 return
```
LecrisUT, line 413: This part is relevant to address. It seems the logic intended for `_is_failed` is already partially adressed here. But https://github.com/teemtee/tmt/issues/4668 is also relevant, so a quick comment on how the future refactoring of this in combination with that would look like would be useful

tmt/steps/finish/__init__.py
LecrisUT: Already raised and addressed https://github.com/teemtee/tmt/pull/4667#issuecomment-4037999198

tmt/queue.py
happz: Added comment in 4e282781.

LecrisUT (approved):

PR #4672, opened 2026-03-11
Implement the `sprint-overview` script

### LINKED ISSUE #4604: Duplicated issues, Jira Cloud integration, sprint overview
There's a couple of things needed to be done for the full syncing to Jira:

* [x] remove the duplicate issues created before the fix for pull request syncing has been deployed
* [x] make sure that when a pull request is merged the corresponding issue is closed
* [x] verify that the Jira Cloud integration works as expected
* [x] create a script listing all issues and pull requests in given sprint

psss:
Added a helper script showing all issues and pull requests included in the given sprint. Provides a human readable text overview and yaml export for further processing. This is the first half of the jira syncing automation. The second part will be completed once the Jira migration is done.

Fix #4604.

Assisted-by: Claude

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/scripts/sprint-overview
+++ b/scripts/sprint-overview
@@ -207,7 +207,10 @@ def display_items(sprint_name: str, items: list[Item]) -> None:
     open_count = len(categories.get("open-issue", [])) + len(categories.get("open-pull", []))
     closed_count = total_count - open_count
 
-    points = {key: sum(item.size or 0 for item in items) for key, items in categories.items()}
+    points = {
+        category: sum(item.size or 0 for item in category_items)
+        for category, category_items in categories.items()
+    }
 
     total_points = sum(points.values())
     open_points = points.get("open-issue", 0) + points.get("open-pull", 0)


scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+
+
+def display_items_yaml(
+    sprint_name: str, categories: dict[str, list[tuple[int, str, Optional[float], str]]]
+) -> None:
+    """Display categorized items in YAML format."""
+    # Calculate total story points by category
+    total_points: dict[str, int] = {}
+    for key, items in categories.items():
+        total_points[key] = int(sum(size or 0 for _, _, size, _ in items))
```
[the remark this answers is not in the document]
psss: Story points are actually `int`.

scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+        ("closed_prs", "Closed Pull Requests"),
+    ]
+
+    for key, label in sections:
+        if categories[key]:
+            print(f"{label} ({len(categories[key])}) - {total_points[key]:.0f} points:")
+            print_separator(char="-")
+            for number, title, size, _ in categories[key]:
+                size_str = f"{size:.0f}" if size is not None else "-"
+                print(f"  #{number:<6} [{size_str:>2}] {title}")
```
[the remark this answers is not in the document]
psss: Story points are actually `int`.

psss:
Example output:

```
> ./scripts/sprint-overview 'Sprint 11'
Fetching items from teemtee/project/1 for sprint 'Sprint 11'...
  Page 1: 40 items
Total: 40 items


================================================================================
Sprint: Sprint 11
Total items: 40
Total story points: 129
================================================================================

Open Issues (10) - 41 points:
--------------------------------------------------------------------------------
  #1270   [ 3] Clarify `tmt` mission and compare with others
  #3691   [ 1] Move to Public Ranch bare metal instances for plans needing VMs
  #4531   [ 5] Implement full loading of the recipe
  #4557   [ 5] Review flaky tests and make them reliable
  #4604   [ 5] Duplicated issues, Jira Cloud integration, sprint overview
  #4605   [ 8] Move installer logic to package managers
  #4618   [ 5] Save warning messages to be picked up by testing-farm
  #4637   [ 2] Add virtualization.confidential hardware requirement
  #4644   [ 5] Create a Verify Install plugin which checks if packages are installed from the correct repo
  #4645   [ 2] All providers should list the artifacts in artifacts.yaml

Open Pull Requests (8) - 24 points:
--------------------------------------------------------------------------------
  #3694   [ 1] Test virtual & bootc provision plugins in public
  #4280   [ 5] Add Polarion story export with test case linking and custom fields - features inside polarion
  #4638   [ 2] Add `virtualization.confidential` hardware requirement
  #4642   [ 3] Save all warnings encountered in a warnings file
  #4653   [ 3] Consolidate installation logic into package managers
  #4660   [ 5] Add a new feature plugin to verify package,repo combination
  #4661   [ 3] Add support for full recipe loading
  #4672   [ 2] Implement the `sprint-overview` script

Closed Issues (7) - 28 points:
--------------------------------------------------------------------------------
  #1793   [ 5] Support `environment` in the `test` key of `fmf` discover
  #3894   [ 5] Support multiple `discover` phases in `upgrade` execute
  #4440   [ 5] Library conflicts with different url
  #4519   [ 5] `tmt-reboot` does not work with `container` plugin
  #4578   [ 3] Improve test coverage for results handling and interpretation
  #4602   [ 3] Outline the consistent approach for handling paths
  #4608   [ 2] Kickstart template for `libvirt` cannot be found

Merged Pull Requests (15) - 36 points:
--------------------------------------------------------------------------------
  #4326   [ 3] Move plan stuff in its own submodule
  #4416   [ 3] Enforce test duration across reboots and restarts
  #4503   [ 3] Allow beakerlib libraries with clashing repo and name
  #4549   [ 5] Add integration test for NVR edge cases 
  #4564   [ 2] Detect and error on duplicate artifact NVRAs
  #4589   [ 3] Add a unit test for check result combinations
  #4615   [ 2] Rework Github release workflow
  #4619   [ 2] Deal with Fedora ELN no longer shipping legacy dnf tooling
  #4621   [ 3] Allow adjusting individual tests
  #4622   [ 1] Update Pre-commit Dependencies
  #4630   [ 1] Skip known bootc failure for now
  #4633   [ 3] Fix support for multiple discover phases in upgrade execute plugin
  #4643   [ 2] Force use of tty in `podman exec` for test session
  #4648   [ 1] Do not check for journal.xml presence, temporarily
  #4656   [ 2] Try to get a more reasonable stacklevel

================================================================================
Summary: 18 open (65 pts), 22 closed/merged (64 pts)
================================================================================
```

scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+        f"Summary: {open_count} open ({open_points:.0f} pts), "
+        f"{closed_count} closed/merged ({closed_points:.0f} pts)"
+    )
+    print_separator(after=True)
+
+
+def display_items_yaml(
+    sprint_name: str, categories: dict[str, list[tuple[int, str, Optional[float], str]]]
+) -> None:
+    """Display categorized items in YAML format."""
```
happz: Seems to me the script is diverting from the docstring format followed by the tmt code, namely one-liner docstrings, `Args:` instead of `:param ...`, `:returns:` not used when describing what the function returns, etc.

scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+            "closed": {"count": closed_count, "story_points": int(closed_points)},
+        },
+    }
+
+    print(yaml.dump(output, default_flow_style=False, sort_keys=False))
+
+
+def main() -> None:
+    """Main entry point."""
+    parser = argparse.ArgumentParser(
```
happz: Wouldn't Click be more in line with what tmt uses? I know it's "just a script", but it's introducing new way of defining CLI options...

scripts/sprint-overview
psss: Yeah, makes sense, adjusted in 92d53e68.

scripts/sprint-overview
```diff
@@ -0,0 +1,403 @@
...
+        action="store_true",
+        help="Output in YAML format for machine-readable processing",
+    )
+
+    args = parser.parse_args()
+
+    token = os.environ.get("GITHUB_PERSONAL_ACCESS_TOKEN")
+    if not token:
+        print("Error: GITHUB_PERSONAL_ACCESS_TOKEN environment variable not set", file=sys.stderr)
+        sys.exit(1)
```
LecrisUT: Unless we are doing write requests, we do not need a `GITHUB_PERSONAL_ACCESS_TOKEN`. From what I see, this is doing only reads on PRs, so I think it could be refactored to use the rest api as read-only instead

scripts/sprint-overview
```diff
@@ -0,0 +1,403 @@
+#!/usr/bin/python3
```
LecrisUT, line 1: Could you include the PEP723 header like

```suggestion
#!/usr/bin/python3
# /// script
# dependencies = [
#   "click",
#   "requests",
#   "ruamel.yaml",
# ]
# ///
```

scripts/sprint-overview
```diff
@@ -0,0 +1,403 @@
...
+"""
+List all issues and pull requests sprint with story points.
+
+Usage:
+    sprint-overview "Sprint 11"
+    sprint-overview "Sprint 11" --org myorg --project 2
+    sprint-overview "Sprint 11" --yaml
```
LecrisUT: Usage is already defined in the help message, let's not introduce a source of divergence

scripts/sprint-overview
```diff
@@ -0,0 +1,403 @@
...
+        "--org", default="teemtee", help="GitHub organization name (default: teemtee)"
+    )
+    parser.add_argument("--project", type=int, default=1, help="Project number (default: 1)")
+    parser.add_argument(
+        "--yaml",
+        action="store_true",
+        help="Output in YAML format for machine-readable processing",
+    )
+
+    args = parser.parse_args()
```
LecrisUT: Could we not just hard-code these? Maybe at most having either yaml format or the format to be used in the slack messages.

scripts/sprint-overview
```diff
@@ -0,0 +1,403 @@
...
+
+    # Calculate total story points by category
+    total_points: dict[str, float] = {}
+    for key, items in categories.items():
+        total_points[key] = sum(size or 0 for _, _, size, _ in items)
+
+    print_separator(before=True)
+    print(f"Sprint: {sprint_name}")
+    print(f"Total items: {total}")
+    print(f"Total story points: {sum(total_points.values()):.0f}")
```
LecrisUT: I don't like the format being defined by all of these print commands. Can we have at least an f-string of the overall format?

scripts/sprint-overview
```diff
@@ -0,0 +1,403 @@
...
+            }
+            pageInfo {
+              hasNextPage
+              endCursor
+            }
+          }
+        }
+      }
+    }
+    """
```
LecrisUT: Uph, isn't this super dark-magic for us? At least with the REST API we can point to the documentation of what is being requested: https://docs.github.com/en/rest/projects/items?apiVersion=2022-11-28#get-an-item-for-an-organization-owned-project

scripts/sprint-overview
psss: Right, updated in 7d37dae9.

scripts/sprint-overview
happz: Or a template. Like Jinja. Exactly the task for which templating engines were invented...

I don't understand why people insist on building complex, multiline, structured text output with raw `print()` calls... It's like there's some bonus for suffering. Yeah, let me compete in the table tennis tournament, but I insist on having both arms tied behind my back, holding paddle in my mouth!

scripts/sprint-overview
psss: Actually, without the token, we're not able to access the project data. I had to explicitly allow the `read:project` permission for that token.

scripts/sprint-overview
happz: And the whole concept of separating logic from the presentation layer. Interweaving both together makes maintenance harder.

scripts/sprint-overview
LecrisUT: Really? I can see them just fine from an incognito window. Does `read:project` restrict all reads or only reads to private repo/projects (the same like `read:repository` is unnecessary actually because of this).

If we do indeed need it, there should be a better way around this with github apps. Personal access tokens are a pain to manage, and would really like to avoid that hard dependency

scripts/sprint-overview
```diff
@@ -0,0 +1,407 @@
...
+            "open_issues": {
+                "count": len(categories["open_issues"]),
+                "story_points": total_points["open_issues"],
+                "items": [
+                    {
+                        "url": url,
+                        "title": title,
+                        "story_points": int(size) if size is not None else None,
+                    }
+                    for _, title, size, url in categories["open_issues"]
```
thrix: The `number` (issue/PR number) is discarded here with `_`, so YAML items only have `url`, `title`, `story_points`. The text output prominently shows `#number`, but the YAML output (intended for machine-readable processing per the `--yaml` help text) requires URL parsing to extract it. Since `number` is already in the tuple, consider including it:

```python
{
    "number": number,
    "url": url,
    "title": title,
    "story_points": int(size) if size is not None else None,
}
for number, title, size, url in categories["open_issues"]
```

Same applies to the other 4 category blocks below.

Generated-by: Claude Code

scripts/sprint-overview
```diff
@@ -0,0 +1,407 @@
...
+
+        if "errors" in data:
+            print(f"GraphQL errors: {data['errors']}", file=sys.stderr)
+            sys.exit(1)
+
+        items = data["data"]["organization"]["projectV2"]["items"]["nodes"]
+        print(f"  Page {page}: {len(items)} items", file=sys.stderr)
+
+        for item in items:
+            if item.get("content"):
```
thrix: This silently skips draft issues (which have no `content` in GitHub Projects). A brief comment explaining the intent would help — e.g. `# Skip draft issues which have no content type`.

Generated-by: Claude Code

LecrisUT: I don't like the review comment made by Claude here. There is an open discussion about the usage of graphql vs rest API, and "the thing" implies that it knows about this open discussion

> Agreement with existing feedback
>
> The following open review comments from other reviewers are valid and worth addressing:
>
> -  @LecrisUT: GraphQL vs REST API — Worth investigating if the REST API can access project items without a token, reducing the setup barrier.

But that means it's just creating noise on a potentially dead code. It should not have given a feedback under these circumstances.

More feedback to the "fine-tuning" of this generated review

scripts/sprint-overview
psss: Nice, didn't know about inline script metadata: 655bc47d

scripts/sprint-overview
psss: Ack: 1a186c91

scripts/sprint-overview
psss: Ok, why not: f6290381

scripts/sprint-overview
LecrisUT: Playing around with it, here are a few REST API queries that are useful:
- Main one: `curl -X GET "https://api.github.com/orgs/teemtee/projectsV2/1/items?per_page=100&q=sprint:@current&fields=126287851,126287852"`
  - The `q` part is the same that we use in the view (we can even reuse a specific pre-defined view if we do not want to hard code it)
  - `fields` is the main one to make navigation a bit easier
- Find the id to used for `fields`: `curl -X GET "https://api.github.com/orgs/teemtee/projectsV2/1/fields"`

psss: Right you are! The REST support for GitHub projects v2 was recently added:

* https://github.blog/changelog/2025-09-11-a-rest-api-for-github-projects-sub-issues-improvements-and-more/

I've simplified the script significantly. Please, re-review.

scripts/sprint-overview
psss: Yeah, that seems to be a limitation of GraphQL, that one need authentication. But REST is working, so no strong need for this. Just to prevent rate limiting.

scripts/sprint-overview
psss: > I don't understand why people insist on building complex, multiline, structured text output with raw `print()` calls... It's like there's some bonus for suffering. Yeah, let me compete in the table tennis tournament, but I insist on having both arms tied behind my back, holding paddle in my mouth!

This comment did not help. Actually, quite opposite. It took me some time to gather back motivation to continue working on this.

happz: > > I don't understand why people insist on building complex, multiline, structured text output with raw `print()` calls... It's like there's some bonus for suffering. Yeah, let me compete in the table tennis tournament, but I insist on having both arms tied behind my back, holding paddle in my mouth!
> 
> This comment did not help. Actually, quite opposite. It took me some time to gather back motivation to continue working on this.

I see, that was not the intention, sorry about the form if you find it inappropriate.

I do believe the question is valid though: why not use a template-based solution, and have advantage of presentation being separated from "business" logic, better readability, better control over whitespace? IIUIC this is not supposed to be an ad-hoc, used-once script, wouldn't the investment into its easier maintenance pay off?

scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+    all_items: list[tuple[str, str, int, str, Optional[float], str]] = []
+    cursor = None
+    page = 1
+
+    while True:
+        variables = {
+            "orgName": org,
+            "projectNumber": project,
+            "cursor": cursor,
+            "query": f'Sprint:"{sprint_name}"',
```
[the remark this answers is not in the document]
psss: Code removed.

scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+        if item_type == "Issue":
+            key = "open_issues" if state == "OPEN" else "closed_issues"
+        else:  # PullRequest
+            key = (
+                "open_prs"
+                if state == "OPEN"
+                else "merged_prs"
+                if state == "MERGED"
+                else "closed_prs"
+            )
```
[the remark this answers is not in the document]
psss: Done.

scripts/sprint-overview
```diff
@@ -0,0 +1,381 @@
...
+                    }
+                    for _, title, size, url in categories["closed_prs"]
+                ],
+            },
+        },
+        "summary": {
+            "open": {"count": open_count, "story_points": int(open_points)},
+            "closed": {"count": closed_count, "story_points": int(closed_points)},
+        },
+    }
```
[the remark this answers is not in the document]
psss: Simplified in d108e84f.

scripts/sprint-overview
psss: Here I'm not really sure. The main intention of the script is to just extract the list of issues and pull requests from the given sprint so that we can sync it with Jira. That task showed to be a bit more "interesting" than I originally thought so I've included the text overview for better debugging. So that part is basically just a byproduct, although quite a useful one I'd say. Anyway, for this use case (at least for now, of course we can extend and modify in the future) having a `header()` function and a `for item in items: print` loop sounds completely sufficient/appropriate to me.

scripts/sprint-overview
psss: That part of the code has been completely dropped.

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+        if field["name"] == "Size":
+            size_id = field["id"]
+        elif field["name"] == "Sprint":
+            sprint_id = field["id"]
+
+    if size_id is None or sprint_id is None:
+        print("Error: Could not find Size and Sprint fields in the project", file=sys.stderr)
+        sys.exit(1)
+
+    return size_id, sprint_id
```
LecrisUT: The IDs would not change, would prefer to hard-code these with a link to how to get them to avoid the additional request.

We can assert them when we use them later in `fetch_sprint_items`

scripts/sprint-overview
psss: Implementation changed substantially. I've verified that the new one correctly reports draft pull requests in the `Open Pull Requests` category.

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+import sys
+from collections import defaultdict
+from typing import Any, Optional
+
+import click
+import requests
+from ruamel.yaml import YAML
+
+GITHUB_API_URL = "https://api.github.com"
+GITHUB_API_VERSION = "2026-03-10"
```
LecrisUT, line 31: Please link to https://docs.github.com/en/rest/about-the-rest-api/breaking-changes to reference if we ever need to bump it

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+        url = (
+            f"{GITHUB_API_URL}/orgs/{GITHUB_ORGANIZATION}"
+            f"/projectsV2/{GITHUB_PROJECT_NUMBER}/{path}"
+        )
+    headers: dict[str, str] = {
+        "Accept": "application/vnd.github+json",
+        "X-GitHub-Api-Version": GITHUB_API_VERSION,
+    }
+    token = os.environ.get("GITHUB_PERSONAL_ACCESS_TOKEN")
+    if token:
```
LecrisUT:
```suggestion
    if token := os.environ.get("GITHUB_PERSONAL_ACCESS_TOKEN"):
```
nitpick

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+    :param sprint_name: name of the sprint to fetch items for.
+    :returns: a list of dictionaries with keys ``id``, ``type``, ``status``,
+        ``size``, ``url``, and ``title``.
+    """
+
+    print(
+        f"Fetching items from {GITHUB_ORGANIZATION}/project/{GITHUB_PROJECT_NUMBER}"
+        f" for sprint '{sprint_name}'...",
+        file=sys.stderr,
+    )
```
LecrisUT: If we are using `click`, we should go all they way to use `echo` where needed. My preference is either `echo` or `logging.logger`

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+
+    size_field_id, sprint_field_id = fetch_field_ids()
+
+    response = github_api_get(
+        "items",
+        params={
+            "per_page": str(GITHUB_ITEMS_PER_PAGE),
```
LecrisUT: Might as well do the `str` conversion at the top, but iiuc, the conversion does not do anything. Also do we need to do pagination? I do hope we do not have 100 items in the sprint, but the default was `30` and not sure if it is significant or related to maximum size of response

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+            "fields": f"{size_field_id},{sprint_field_id}",
+            "q": f'Sprint:"{sprint_name}"',
```
LecrisUT:
```suggestion
            "q": f"Sprint:{sprint_name}",
```
nitpick to be more consistent with the other quotations

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+
+            all_items.append(
+                {
+                    "id": content["number"],
+                    "type": item_type,
+                    "status": status,
+                    "size": size,
+                    "url": content["html_url"],
+                    "title": content["title"],
+                }
```
LecrisUT: Please make this a TypedDict or dataclass. Then you can define the `__str__` that would be used to format `display_items`. That would at least give us a central place to review the format.

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+    while True:
+        items = response.json()
+        print(f"Page {page}: {len(items)} items", file=sys.stderr)
+
+        for item in items:
+            content = item.get("content")
+            if not content:
+                continue
+
+            item_type = "pull" if item["content_type"] == "PullRequest" else "issue"
```
LecrisUT, line 165: Iirc there are also `card` types. Let's use the `content_type` as-is and work around it

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+            f"{GITHUB_API_URL}/orgs/{GITHUB_ORGANIZATION}"
+            f"/projectsV2/{GITHUB_PROJECT_NUMBER}/{path}"
+        )
+    headers: dict[str, str] = {
+        "Accept": "application/vnd.github+json",
+        "X-GitHub-Api-Version": GITHUB_API_VERSION,
+    }
+    token = os.environ.get("GITHUB_PERSONAL_ACCESS_TOKEN")
+    if token:
+        headers["Authorization"] = f"Bearer {token}"
```
LecrisUT: We can construct this header once at the beginning

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+from typing import Any, Optional
+
+import click
+import requests
+from ruamel.yaml import YAML
+
+GITHUB_API_URL = "https://api.github.com"
+GITHUB_API_VERSION = "2026-03-10"
+GITHUB_ORGANIZATION = "teemtee"
+GITHUB_PROJECT_NUMBER = 1
```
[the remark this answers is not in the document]
psss: Intentionally dropped.

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+            )
+
+        # Follow pagination via Link header
+        link_header = response.headers.get("Link", "")
+        match = re.search(r'<([^>]+)>;\s*rel="next"', link_header)
+        if not match:
+            break
+
+        response = github_api_get("", url=match.group(1))
+        page += 1
```
LecrisUT: This could be encoded directly in `github_api_get` and convert it to `yield`. Also pre-compile the regex with documentation to https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2026-03-10#using-link-headers

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+    """
+    try:
+        items = fetch_sprint_items(sprint_name)
+
+        if not items:
+            print(f"No items found in sprint '{sprint_name}'.")
+            return
+
+        if output_yaml:
+            YAML().dump(items, sys.stdout)
```
LecrisUT: Let's at least order the items by the `number` to make the diff more predictable

scripts/sprint-overview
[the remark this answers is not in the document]
psss: Actually `2026-03-10` is listed in [supported API versions](https://docs.github.com/en/rest/about-the-rest-api/api-versions?apiVersion=2026-03-10#supported-api-versions).

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+    closed_count = total - open_count
+    open_points = total_points.get("open-issue", 0) + total_points.get("open-pull", 0)
+    closed_points = sum(total_points.values()) - open_points
+
+    header(
+        f"Summary: {open_count} open ({open_points} pts), "
+        f"{closed_count} closed/merged ({closed_points} pts)",
+        char="=",
+        top=True,
+    )
```
LecrisUT: The formatting here is still to magical compared to a jinja
```jinja
{%- for key, label in sections.items() %}
{{ label }}
{{ "=" * len(label) }}
{%- for item in categories[key] %}
{{ item }}
{%- endfor %}
{%- endfor %}

Summary: {{ len(open) }} open ({{ sum(open)}} pts), {{ len(closed) }} closed ({{ sum(closed) }} pts)
```
It is hard to follow what is printed where and when with both `header` and `print` commands. The main negative to the jinja is the whitespace control, but we can converge on a proper format with trial and error.

Also, using a markdown instead of rst format makes formatting much easier, let's use that instead?

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+    '--yaml',
+    'output_yaml',
+    is_flag=True,
+    help='Output in YAML format for machine-readable processing.',
+)
+def main(sprint_name: str, output_yaml: bool) -> None:
+    """
+    List all issues and pull requests in a GitHub Project sprint with story points.
+
+    SPRINT_NAME: Name of the sprint (e.g., 'Sprint 11')
```
LecrisUT: Could move this to be an `option` and then the documentation is better placed

scripts/sprint-overview
```diff
@@ -0,0 +1,293 @@
...
+            categories = categorize_items(items)
+            display_items(sprint_name, categories)
+
+    except requests.exceptions.RequestException as error:
+        print(f"Error making API request: {error}", file=sys.stderr)
+        sys.exit(1)
+
+    except Exception as error:
+        print(f"Error: {error}", file=sys.stderr)
+        sys.exit(1)
```
LecrisUT: Don't think it's worth the special handling in here. It's not a publicly facing script and if it fails we would want to know exactly when and how it failed.

scripts/sprint-overview
LecrisUT: Will insist on a template or getting rid of the human-readable format.

The scope of when the human-readable format is unclear. Should it be used to kickstart the sprint status, to make a diff, or anything? If it is used anywhere non-programatically, then we might want to make tweaks on it based on new feedback coming in.

psss: Jinja template added in 1578b89a.

scripts/sprint-overview
psss: Field id hardcoded in 1578b89a.

scripts/sprint-overview
psss: Added.

scripts/sprint-overview
psss: Nitpicked.

scripts/sprint-overview
psss: Used `echo` instead of `print`.

scripts/sprint-overview
psss: `mypy` was complaining if this was not string, and `int` constant sounds more natural.

scripts/sprint-overview
psss: Nitpicked. But simple quotes are needed.

scripts/sprint-overview
psss: Converted to `dataclass`.

scripts/sprint-overview
psss: I don't think we're using cards. Let's keep that for a separate PR if needed.

scripts/sprint-overview
psss: Now not repeated in page handling. I'd say, looks better here, grouped with other headers.

scripts/sprint-overview
psss: Page handling moved.

scripts/sprint-overview
psss: All items sorted.

scripts/sprint-overview
psss: Jinja template used.

scripts/sprint-overview
psss: Converted to an option.

scripts/sprint-overview
psss: Exceptions grouped.

scripts/sprint-overview
```diff
@@ -0,0 +1,267 @@
...
+                    title=content["title"],
+                )
+            )
+
+    all_items.sort(key=lambda item: item.id)
+    click.echo(f"Total: {len(all_items)} items", err=True)
+    return all_items
+
+
+def display_items(sprint_name: str, items: list[Item]) -> None:
```
thrix, line 193: Would it be a big ask to use other name as `items` here, it is then below used again in dict/list comprehensions and can influence readability

scripts/sprint-overview
```diff
@@ -0,0 +1,267 @@
...
+            if item_type == "pull":
+                status = "merged" if content.get("merged_at") else content["state"]
+            else:
+                status = content["state"]
+
+            all_items.append(
+                Item(
+                    id=content["number"],
+                    type=item_type,
+                    repo=content["html_url"].split("/")[4],
```
thrix: Would it be possible to make this a bit more reboust, or at least leave a comment why we went with this britte solution. Also the GitHub Projects V2 REST API response likely includes a `repository` or `repository_url` field on content items which would be more robust.

scripts/sprint-overview
```diff
@@ -0,0 +1,267 @@
...
+        if not items:
+            click.echo(f"No items found in sprint '{sprint}'.")
+            return
+
+        if output_yaml:
+            YAML().dump([dataclasses.asdict(item) for item in items], sys.stdout)
+        else:
+            display_items(sprint, items)
+
+    except Exception as error:
```
thrix: are we fine catching broead exceptions like this? This can swallow the traceback making it harder to debug.

thrix (changes requested):
nitpicks mostly

scripts/sprint-overview
psss: It's aligned with the class name `Item`, used consistently across the whole code, covers both issues and pull requests, I don't see what is the possible confusion. Any suggestions for a better name?

scripts/sprint-overview
psss: Dropped completely in 092e4aba.

scripts/sprint-overview
psss: There's no common field for issues and pull requests. But I was able to find a repo name for each type separately: 092e4aba

scripts/sprint-overview
thrix: for example `display_items` ?

scripts/sprint-overview
thrix: ty!

scripts/sprint-overview
thrix: LGMT

scripts/sprint-overview
happz: -1 from me to `def display_items(..., display_items: list[Item])`, `items: list[Item]` is fairly clear, it's a list of items to display by the `display_items` function.

One thing that should be addressed though is this comprehension:

```python
points = {key: sum(item.size or 0 for item in items) for key, items in categories.items()}
```

`items` no longer labels the original argument, the list of items to display, it will label whatever the items had the last category in `categories`. This is risky, after this point if I use `items`, I'm no longer working with the input of this function. It is true that `items` is not accessed after this point, but that can change easily, causing confusion. Maybe `for category_name, category_items in categories.items()`?

psss: Good catch! Thanks, should be fixed in dfce2dd7640d3d501413058543578dedc4a49261.

LecrisUT:
@happz do you want to give it a quick look, or is it green light to merge this?

happz:
> @happz do you want to give it a quick look, or is it green light to merge this?

It would have been fine without my approval, but approved nevertheless, looks very good now.

PR #4396, opened 2025-12-02
Address two expected AVC denials on Fedora Rawhide

psss:
Three tests recently started to fail on Rawhide:

* `/tests/test/check/coredump` changed to `xfail`
* `/tests/prepare/shell` disabled for the `mock` provision plan
* `/tests/provision/mock/fedora-rawhide-x86_64/multiple-tests` changed to `xfail`

Pull Request Checklist

* [x] adjust the test coverage

PR #4674, opened 2026-03-11
Partially revert #4396

LecrisUT:
- https://bugzilla.redhat.com/show_bug.cgi?id=2415701 has been resolved in `selinux-policy-43.1-1.fc43`
- https://bugzilla.redhat.com/show_bug.cgi?id=2418343 is also tracked in that PR, but it used a too broad xfail

### DIFF
--- a/tests/test/check/main.fmf
+++ b/tests/test/check/main.fmf
@@ -49,7 +49,10 @@ duration: 20m
       - when: distro >= fedora-44
         check:
           - how: avc
-            result: xfail
+            ignore-pattern:
+              - "type=AVC.*comm=systemd-coredum.*scontext=system_u:system_r:systemd_coredump_t"
+              # From default pattern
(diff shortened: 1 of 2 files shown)


tests/test/check/main.fmf
```diff
@@ -49,7 +49,10 @@ duration: 20m
       - when: distro >= fedora-44
         check:
           - how: avc
-            result: xfail
+            ignore-pattern:
+              - "type=AVC.*comm=systemd-coredum.*scontext=system_u:system_r:systemd_coredump_t"
```
psss, line 53: Could we add short comment explaining why this `ignore-pattern` is actually needed?

LecrisUT: The previous tracker a few lines below

> `because: https://bugzilla.redhat.com/show_bug.cgi?id=2418343`

still applies

PR #4676, opened 2026-03-11
Expose typing.override

LecrisUT:
Separated from #4654. See there for more context of why it is needed


LecrisUT:
> The pull request **successfully** exposes `typing.override` through the `tmt._compat.typing` module.

For a "Large language model" this thing does not have a grasp of either language or coding. How would you unsuccessfully do that :roll_eyes:.

Thoughts on making it shut up if it doesn't have anything productive to contribute? Among this and graphite, I much prefer the latter because it's not like :nerd_face::point_up: every time

PR #4678, opened 2026-03-11
Convert /tests/policy/plan into pytest-based test

happz:
This one uses some jq queries, changes cwd, more complicated, but not terribly complex.

Did I mention jq? I added jq. I like jq. I don't like a mess of `in` and `get(..., {})` expressions to inspect a data structure.

Pull Request Checklist

* [x] implement the feature


tests/__init__.py
```diff
@@ -30,6 +39,52 @@ def reset_common() -> None:
...
+
+    cwd = Path.cwd()
+
+    os.chdir(path)
+
+    try:
+        yield path
+
+    finally:
+        os.chdir(cwd)
```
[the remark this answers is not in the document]
happz: No, there is no truce with multithreading. There is only one CWD, and once a thread changes it, it affects all other threads. There is no need to worry about restoring it safely between multiple threads, because by then tests are already broken.

tests/__init__.py
```diff
@@ -30,6 +39,52 @@ def reset_common() -> None:
...
+
+    return _with_cwd
+
+
+def jq_all(data: Any, query: str) -> Iterable[Any]:
+    """
+    Apply a jq filter on given data, and return the product.
+    """
+
+    return cast(Iterable[Any], _jq.compile(query).input(data).all())
```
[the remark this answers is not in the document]
happz: Not needed, there is no state we would like to preserve. Quite contrary, we want tracebacks to be reported.

tmt/_compat/importlib/metadata.py
```diff
@@ -1,10 +1,14 @@
...
 else:
-    from importlib_metadata import entry_points  # pyright: ignore[reportUnknownVariableType]
+    from importlib_metadata import (
+        entry_points,  # pyright: ignore[reportUnknownVariableType]
+        version,
+    )
 
 __all__ = [
     "entry_points",
+    "version",
```
LecrisUT, line 13: Why is this one needed to be exported? It is available from the start https://github.com/python/cpython/blob/v3.9.25/Lib/importlib/metadata.py

plans/main.fmf
```diff
@@ -12,6 +12,13 @@ prepare+:
       - python3-pip
       - python3-pytest
 
+  # Install *Python bindings for jq* - jq itself was installed above
+  # as a regular package. But Python tests use jq to inspect structures,
+  # and need jq Python bindings.
+  - how: shell
+    script: |
+      pip3 install jq
```
LecrisUT, line 20: Can we please avoid this? I rather help with figuring out what the blackmagic in the yq/jq was used and translate it to python instead.

tests/__init__.py
```diff
@@ -53,11 +113,15 @@ def __call__(
...
+        if _CLICK_VERSION >= (8, 2, 0):
+            super().__init__(charset='utf-8', echo_stdin=False)
+
+        else:
+            super().__init__(charset='utf-8', echo_stdin=False, mix_stderr=False)
 
     def invoke(  # type: ignore[override]
         self,
-        *args: str,
+        *args: Union[str, Path],
```
LecrisUT, line 124: I would refer not to extend this one, at the very least not inside this function

tests/policy/test_plan.py
```diff
@@ -0,0 +1,63 @@
...
+        ['feature', 17],
+        ['shell', None],
+    ], 'Verify that prepare step contains two phases'
+
+    assert jq_all(plans_exported, '.[] | .contact') == [["xyzzy"]], (
+        'Verify that contact key was populated'
+    )
+
+
+@with_cwd(DATA_DIR)
```
LecrisUT, line 53: Could we maybe avoid this and use `--root` instead? We could even put it in the `run_tmt` fixture along with some plan selection and stuff.

happz: I'd consider this a different problem. One that can be improved, but I tried to not invent too many new things.

plans/main.fmf
happz: There is no black magic, just a simple filter one can use to inspect a JSON/YAML structure. Native tools like indexing and `get()` can't keep up in simplicity and readability with `.[] | .prepare | .[] | [.how, .order]`.

Give me a tool as good as a jq filter, and I'd happily use it.

LecrisUT: It's more about readability `.[] | .prepare | .[] | [.how, .order]` vs
```python
def _filter(data: Any):
    for plan in data:
        for prepare in plan["prepare"]
            yield {
                "how": prepare["how"],
                "order": prepare["order"],
            }
```
And we don't even need the last part because that was there just to make it string-friendly comparable.

happz: Yeah, and the Python implementation of that simple filter is absolutely terrible. So much imperative work instead of declaring what I want to get. And the deeper a test would need to peek, with keys that may or may not be present, these custom helpers will get more and more complicated.

Sorry, you can't really convince me Python is better in this area, it's not its strongest suite. My idea is to use a tool that's, IMHO, better suited for the task, that's all.

This discussion cannot be resolved here, we clearly have different views on how the problem should be approached. Added to the hacking session agenda.

LecrisUT: After trying to use a python-only approach more, I've come around to this, and I'm :+1: to having something along the original lines

PR #4681, opened 2026-03-12
Refactor github workflow

LecrisUT:
Pull Request Checklist

* [x] implement the feature
  * [x] Move build wheel workflow to also be run on PRs
  * [x] Move all other components to have only 2 main workflows `ci` and `release`
  * [x] Simplify the naming structure
 
Depends-on: #4682

### DIFF
--- a/.github/workflows/step-shellcheck.yml
+++ b/.github/workflows/step-shellcheck.yml
@@ -7,7 +7,7 @@ on:
         description: Committish to test
       upload_sarif:
         type: boolean
-        description: Whether the sharif should be uploaded
+        description: Whether the sarif should be uploaded
 
 permissions: {}
 


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+      ref: ${{ inputs.ref }}
+    needs: [ pre-commit ]
+
+  pass:
+    name: ✅ Pass
+    if: always()
+    needs: [ pre-commit, build-wheel, shellcheck, doc-tests ]
+    runs-on: ubuntu-slim
+    steps:
+      - uses: re-actors/alls-green@release/v1
```
thrix, line 54: `ubuntu-slim` is a valid GitHub runner but currently in preview (per `actions/runner-images`). All existing workflows in this repo use `ubuntu-latest`. Preview runners may have availability or stability limitations — worth acknowledging.

Generated-by: Claude Code

.github/workflows/step-shellcheck.yml
```diff
@@ -1,12 +1,12 @@
...
-    branches: [ main ]
+  workflow_call:
+    inputs:
+      ref:
+        type: string
+        description: Committish to test
 
-permissions:
-  contents: read
+permissions: {}
```
thrix, line 13: With `permissions: {}` at workflow level and `security-events: write` at job level — for reusable workflows, the called workflow's GITHUB_TOKEN permissions are capped by the caller's. `ci.yml` has no explicit `permissions:`, so this depends on the teemtee org's default GITHUB_TOKEN settings. If the default doesn't include `security-events: write`, the ShellCheck SARIF upload would fail. Worth verifying after merge.

Generated-by: Claude Code

.github/workflows/ci.yml
mcasquer, line 54: I guess we have configured as 'permissive' the workflow actions in the repo, right? Otherwise something like: 
```yml
permissions:
  actions: read
```
would be needed, for the `pass` job to read the status of the other ones.

mcasquer (approved):
Raised one concern. Apart from that, LGTM

.github/workflows/ci.yml
LecrisUT: No, it does not need it. Its input comes directly from `${{ toJSON(needs) }}`. But good note, I should add empty permissions to this as well. I think I missed some other, will re-check

LecrisUT: For a github action that really does not depend on the OS at all [^1], this has no affect at all. (which is the whole point of `ubuntu-slim`)

[^1]: https://github.com/re-actors/alls-green/blob/a638d6464689bbb24c325bb3fe9404d63a913030/action.yml#L45-L60

.github/workflows/step-shellcheck.yml
LecrisUT: > ci.yml has no explicit permissions:, so this depends on the teemtee org's default GITHUB_TOKEN settings

What? No? What are you/it trying to say and is it remotely true?

> Worth verifying after merge.

No, this is run always, so if it has or lacks permissions now is reviewable right now.

LecrisUT: I reviewed the usage here. It might have been an error to upload them unconditionally for PRs as well. I have raised the issue with github https://github.com/github/codeql-action/issues/3578 to clarify what their process is. But for now I moved these to upload only on `main` commits

.github/workflows/ci.yml
thrix: @LecrisUT ack, I was not aware of `ubuntu-slim` so let it post it here for others more for awareness.

(not generated by AI)

.github/workflows/step-shellcheck.yml
thrix: @LecrisUT looking at that issue, so this nuance was indeed someting to be aware of? :)

(not generated by AI)

LecrisUT: > [@LecrisUT](https://github.com/LecrisUT) looking at that issue, so this nuance was indeed someting to be aware of? :)

Worth being aware, yes. Worth the gymnastic of first figuring it out what it is trying to say, questionable. A simple "Does this work with the `presmissions.security-events` in `jobs.lint`?" would have been sufficient

.github/workflows/step-shellcheck.yml
```diff
@@ -1,42 +1,62 @@
...
-  pull_request:
-    branches: [ main ]
+  workflow_call:
+    inputs:
+      ref:
+        type: string
+        description: Committish to test
+      upload_sarif:
+        type: boolean
+        description: Whether the sharif should be uploaded
```
psss:
```suggestion
        description: Whether the sarif should be uploaded
```

A typo?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+      security-events: write
+
+  doc-tests:
+    name: 📘 docs
+    uses: ./.github/workflows/step-doc-tests.yml
+    with:
+      ref: ${{ inputs.ref }}
+    needs: [ pre-commit ]
+
+  pass:
```
psss, line 48: What's the benefit of aggregating individual checks? If something fails, we would clearly see it nicely placed at the top of the check list, right?

psss (commented):
Looks nice, thanks! I'm just not sure about the all-green check benefit.

.github/workflows/ci.yml
LecrisUT: It gives a single point to manage the `Required` workflows, e.g. currently the `pre-commit` needs to be moved because the workflow name has changed due to the re-ordering. It is quite configurable though, e.g. you can specify which jobs are allowed to fail and the job will not turn red for those with a [job summary](https://github.com/teemtee/tmt/actions/runs/23139335647#summary-67211628790) breaking it down

psss: I see. Ok, let's try it! :+1:

LecrisUT:
Github actions only change, and it passes. Migrated the required to `CI / Pass`, merging

PR #4682, opened 2026-03-12
Fix issues with `check-wheel-contents`

LecrisUT:
Reported when the recent release workflow was run:
```
/tmp/baipp/dist/tmt-1.69.0-py3-none-any.whl: W002: Wheel contains duplicate files:
  tmt/steps/report/junit/templates/default.xml.j2
  tmt/steps/report/junit/templates/polarion.xml.j2
```

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -486,3 +486,8 @@ skip = "tests/execute/weird/data/weird.txt,tests/lint/plan/data/invalid_attr.fmf
 [tool.djlint]
 use_gitignore=true
 ignore="H005,H030,H031"
+
+[tool.check-wheel-contents]
+# W002: Wheel contains duplicate files
+#   Minor issue for us, it is fine if we have a few duplicate files
+ignore = ["W002"]


pyproject.toml
```diff
@@ -486,3 +486,8 @@ skip = "tests/execute/weird/data/weird.txt,tests/lint/plan/data/invalid_attr.fmf
 [tool.djlint]
 use_gitignore=true
 ignore="H005,H030,H031"
+
+[tool.check-wheel-contents]
+# W002: Wheel contains duplicate files
+#   Minor issue for us, it is fine if we have a few duplicate files (default.xml.j2 and polarion.xml.j2 for example)
+ignore = ["W002"]
```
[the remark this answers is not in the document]
LecrisUT: Go away

pyproject.toml
```diff
@@ -486,3 +486,8 @@ skip = "tests/execute/weird/data/weird.txt,tests/lint/plan/data/invalid_attr.fmf
 [tool.djlint]
 use_gitignore=true
 ignore="H005,H030,H031"
+
+[tool.check-wheel-contents]
+# W002: Wheel contains duplicate files
+#   Minor issue for us, it is fine if we have a few duplicate files (default.xml.j2 and polarion.xml.j2 for example)
```
happz: FTR, they are not duplicates, they just have, at this point of time, by accident, the same content. It'd be nice to mention this to prevent anyone trying to replace one with a symlink or something.

LecrisUT: I just dropped the specific mention of those. Even if they did replace it with symlinks, when the wheel is created those symlinks do not survive so it would not have "fixed" that issue

LecrisUT:
Does not affect tmt execution. Merging without waiting for CI, relevant check is in #4681 anyway

Issue #1602, opened 2022-10-17
Fix use of nitrate and pylero in export.py to avoid global state


happz:
* avoid mutable global members like `DEFAULT_PRODUCT` and `PolarionException`
* improve type annotations, avoid `Any` by using aliases/`NewType`
* preserve lazy import

PR #4695, opened 2026-03-13
`copr.repository` provider creates a `Repository` object

### LINKED ISSUE #4692: `copr.repository` provider should enumerate packages into artifacts.yaml
`CoprRepositoryProvider.fetch_contents()` calls `guest.package_manager.enable_copr()`, which uses the `dnf copr` plugin to install the repo. This never creates a `Repository` object (with known `repo_ids`), so `list_packages()` cannot be called on it. FIX IT!

AthreyVinay:
Resolves #4692

### DIFF
--- a/tmt/steps/prepare/artifact/providers/copr_repository.py
+++ b/tmt/steps/prepare/artifact/providers/copr_repository.py
@@ -128,7 +128,7 @@ class CoprRepositoryProvider(ArtifactProvider):
             )
         except tmt.utils.RunError as error:
             raise tmt.utils.PrepareError(
-                f"Failed to read '{repo_filename}' from the guest. "
+                f"Failed to read '{repo_filename}' from the guest."
             ) from error
 
         self.repository = Repository.from_content(output.stdout or '', self.copr_repo, self.logger)


tmt/package_managers/dnf.py
```diff
@@ -286,17 +309,7 @@ def _enable_copr_epel6(self, copr: str) -> None:
...
-        if not matched:
-            raise PrepareError(f"Invalid copr repository '{copr}'.")
-        group, name, project = matched.groups()
-        group = 'group_' if group else ''
-        # Prepare the repo file url
-        parts = [COPR_URL] + (['g'] if group else [])
-        parts += [name, project, 'repo', 'epel-6']
-        parts += [f"{group}{name}-{project}-epel-6.repo"]
-        url = '/'.join(parts)
+        url = build_copr_repo_url(copr, 'epel-6')
```
LecrisUT: Do we still support this? Copr repos in https://copr.fedorainfracloud.org/coprs do not have epel-6 chroot anymore. @psss @happz @thrix? Could probably still be accessed on internal copr, but at that point the hard-coded url would still get in the way.

tmt/steps/prepare/artifact/providers/_copr.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Initialize copr client session.
+        """
+        import_copr(self.logger)
+
+        try:
+            config = {"copr_url": "https://copr.fedorainfracloud.org"}
+            return Client(config)
+        except Exception as error:
+            raise tmt.utils.GeneralError("Failed to initialize Copr client session.") from error
```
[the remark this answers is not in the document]
AthreyVinay: The apis can raise more than just `CoprException`. Wold like to handle network errors too gracefully.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -76,18 +88,34 @@ def _download_artifact(self, artifact: ArtifactInfo, guest: Guest, destination:
             "CoprRepositoryProvider does not support downloading individual RPMs."
         )
 
+    def get_repositories(self) -> list[Repository]:
+        self.logger.info(f"Providing repository '{self.repository.name}' for installation")
+        return [self.repository]
```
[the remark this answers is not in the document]
AthreyVinay: in `go()`, fetch is called before collecting repositories.
```python
provider.fetch_contents(guest, download_path)                                                                                                                                                           
# ...
repositories.extend(provider.get_repositories())
```

tmt/package_managers/dnf.py
AthreyVinay: Nice one - havent seen epel-6 there.

thrix: I do not think copr for epel-6 is a reasonable test target, I would ignore it myself. I do not think we will get anything to be tested there. I would call it unsupported.

tmt/steps/prepare/artifact/providers/_copr.py
```diff
@@ -0,0 +1,120 @@
...
+    python -c 'import copr'
+
+* Users who installed tmt from PyPI should install the ``copr`` package
+  via ``pip install copr``.
+""",
+)
+
+
+COPR_URL = 'https://copr.fedorainfracloud.org/coprs'
+COPR_REPO_PATTERN = re.compile(r'^(@)?([^/]+)/([^/]+)$')
```
LecrisUT: Please use named groups to make it more explicit on what each group is meant to represent

tmt/steps/prepare/artifact/providers/_copr.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def parse_copr_repo(copr_repo: str) -> tuple[bool, str, str]:
```
LecrisUT: Don't quite like tuples, can we avaoid it please.

tmt/steps/prepare/artifact/providers/copr_build.py
```diff
@@ -97,28 +61,22 @@ def build_info(self) -> Optional["Munch"]:
         """
         return self._session.build_proxy.get(self.build_id)
 
+    @property
+    def _copr_owner(self) -> str:
+        assert self.build_info is not None
```
LecrisUT: Could move the assert into the `build_info` instead?

tmt/steps/prepare/artifact/providers/_copr.py
```diff
@@ -0,0 +1,120 @@
...
+    if not matched:
+        raise tmt.utils.PrepareError(f"Invalid copr repository '{copr_repo}'.")
+    is_group, name, project = matched.groups()
+    return bool(is_group), name, project
+
+
+def build_copr_repo_url(copr_repo: str, chroot: str) -> str:
+    """
+    Construct the URL for a COPR ``.repo`` file.
+    """
```
LecrisUT: This is just a bandaid until copr can provide https://github.com/fedora-copr/copr/issues/4119 right? Worth making a comment on that

tmt/steps/prepare/artifact/providers/_copr.py
```diff
@@ -0,0 +1,120 @@
...
+    global copr, Client
+    try:
+        import copr
+        from copr.v3 import Client
+    except ImportError as error:
+        from tmt.utils.hints import print_hints
+
+        print_hints('artifact-provider/copr', logger=logger)
+
+        raise tmt.utils.GeneralError("Could not import copr package.") from error
```
LecrisUT: Another instance of https://github.com/teemtee/tmt/pull/3350. We should have a more systematic place to put these wrappers (better still if we can avoid these wrappers altogether). Cannot find the tracker other than https://github.com/teemtee/tmt/issues/1602. Not an issue to be resolved here though.

tmt/steps/prepare/artifact/providers/_copr.py
```diff
@@ -0,0 +1,120 @@
...
+        Fetch and return the COPR project metadata.
+        """
+        try:
+            return self._session.project_proxy.get(
+                ownername=self._copr_owner, projectname=self._copr_project
+            )
+        except Exception as error:
+            raise tmt.utils.GeneralError(
+                f"Failed to fetch COPR project info for '{self._copr_owner}/{self._copr_project}'."
+            ) from error
```
LecrisUT: Could also make this a function that is called in the `__init__`? it would avoid the need for intermediate `_copr_owner` that is not used elsewhere.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -76,18 +91,34 @@ def _download_artifact(self, artifact: ArtifactInfo, guest: Guest, destination:
...
+
+        if chroot not in chroot_repos:
+            raise GeneralError(
+                f"COPR repository '{self.copr_repo}' has no chroot '{chroot}'. "
+                f"Available chroots: {', '.join(sorted(chroot_repos))}"
+            )
+
+        url = build_copr_repo_url(self.copr_repo, chroot)
+        self.logger.debug(f"Fetching COPR repository '{self.copr_repo}' from '{url}'.")
+        self.repository = Repository.from_url(url, self.logger)
```
LecrisUT: Don't like how the `self.repository` is initialized outside of the `__init__` logic, it could create issues more down the line.

LecrisUT: Hmm, can you check how the `copr enable` works and if we are not missing anything with this change? At the top of my head for concerns is the `additional_repos` in the project. BTW that is not handled in `copr_build` yet?

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -76,18 +91,34 @@ def _download_artifact(self, artifact: ArtifactInfo, guest: Guest, destination:
...
-            enabling Copr repositories.
+        Resolve the Copr repository ``.repo`` file for this guest.
         """
-        guest.package_manager.enable_copr(self.copr_repo)
+        chroot_repos = self.project_info.chroot_repos
+        os_release = guest.facts.os_release_content
+        chroot = (
+            f"{os_release.get('ID', '')}-{os_release.get('VERSION_ID', '')}"
+            f"-{guest.facts.arch or ''}"
+        )
```
LecrisUT: Uph, this seems quite a fragile way of defining it. Can you cross-reference with copr/dnf to see how do they handle it?

tmt/steps/prepare/artifact/providers/_copr.py
happz: I think it would be nice to use verbose patterns more often.

```python
COPR_REPO_PATTERN = re.compile(r"""
    ^
    (@)?     # something optional...
    ([^/]+)  # team name?
    /        # '/' separates team and project names
    ([^/]+)  # project name?
    $
    """)
```

LecrisUT: I have my gripes with verbose pattern because it gets too verbose for something that regex101 explains much better. I do not often care if a `.+` or `[^/]+` is used, and verbosely explaining the difference between those two are a distraction especially when the naming is well-defined on the other end.

Why I do prefer the named groups though is because it explains the _why_ this specific group is captured, which for me is the only thing that I am interested in the regex pattern. Anything else I go to regex101 and try out some patterns and try to break/improve it.

happz: Oh yes, absolutely use named groups. It's not in conflict with verbose patterns, we can and should use both.

tmt/package_managers/dnf.py
LecrisUT: Split off this change in https://github.com/teemtee/tmt/pull/4706

tmt/steps/prepare/artifact/providers/_copr.py
bajertom: What is wrong with tuples?

LecrisUT: Mostly readability issue. If the tuple definition is nearby, it is not that bad, but moving back and forward between files to know what which `str` is meant to be. There are NamedTuple and type-alias, but at that point, a dataclass is much neater

AthreyVinay: NamedTuple yes...
dataclass aka container a bit of an overkill for what we are trying to achieve here. WDYT?

happz: Dataclasses are actyally faster than named tuples :)

AthreyVinay: Agree - but we are not actually mutating it (need to set `Frozen=True`).

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Excellent point - not handled in either places. Would've been an install gotcha. Worth a separate change maybe?

LecrisUT: Sounds fine for a separate change

AthreyVinay: On a closer inspection, I dont think this patch deviate from the original dnf code... see:
https://github.com/rpm-software-management/dnf-plugins-core/blob/6e357ac84b20e43635969fdb65dec8ff66f4a408/plugins/copr.py#L555

We only embed whats in the `.repo` file & neither of us act on it.

AthreyVinay: `chroot` is only deternined from within the guest  & hence cannot initialize it in `__init__`.

LecrisUT: Hmm, confirmed with a test copr repo locally. That is odd, I'll poke copr folks for more details

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Fair enough - they do something like this:
https://github.com/rpm-software-management/dnf-plugins-core/blob/6e357ac84b20e43635969fdb65dec8ff66f4a408/plugins/copr.py#L458

tmt/steps/prepare/artifact/providers/copr_repository.py
LecrisUT: 😱 That stings. At least initialize it as `None` and `assert` it later on

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Intersting issue: https://github.com/rpm-software-management/dnf-plugins-core/issues/524

tmt/steps/prepare/artifact/providers/copr_repository.py
LecrisUT: Actually, scratch that, I had a PBKAC [^1]. Try it out with https://copr.fedorainfracloud.org/coprs/lecris/Test/ and you will get 2 repos defined in it

[^1]: https://matrix.to/#/!rALCnapvdNTxUTuENl:matrix.org/$YDNLRKBsvReLcmR94fOFWPOhVN1aEXk2UfuNpeQYimc?via=fedoraproject.org&via=matrix.org&via=fedora.im

tmt/steps/prepare/artifact/providers/copr_repository.py
LecrisUT: Using the same `distro` sounds reasonable and it normalizes the name more nicely
```python
>>> distro.id()
fedora
>>> distro.version()
43
```
Distinguishing between centos, centos+epel, rhel, rhel+epel would be a huge pain in the underside and apparently they do not do it well either. If we can assume that only one variant of the other is present, or that the explicit chroot to be tested is also passed, then it would be a bit easier.

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Thanks for this - since the runtime dep repos are embedded in the .repo file, this should get enabled for free.
I think we are good. WDYT?

LecrisUT: In the current form it would not extract those and embed in the `.repo` file though. It should produce:
```
[copr:copr.fedorainfracloud.org:lecris:Test]
name=Copr repo for Test owned by lecris
baseurl=https://download.copr.fedorainfracloud.org/results/lecris/Test/fedora-rawhide-x86_64/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/lecris/Test/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1

[coprdep:copr.fedorainfracloud.org:lecris:nightly]
name=Copr copr.fedorainfracloud.org/lecris/Test runtime dependency #1 - lecris/nightly
baseurl=https://download.copr.fedorainfracloud.org/results/lecris/nightly/fedora-rawhide-x86_64/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/lecris/nightly/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1
```

tmt/steps/prepare/artifact/providers/_copr.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

tmt/steps/prepare/artifact/providers/_copr.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

tmt/steps/prepare/artifact/providers/copr_build.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

tmt/steps/prepare/artifact/providers/_copr.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

tmt/steps/prepare/artifact/providers/_copr.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: fixed in [8014c88](https://github.com/teemtee/tmt/pull/4695/commits/8014c887ba14283620eb8739681e8004291bc3e9)

LecrisUT:
After a 1-on-1 with @AthreyVinay we found that the copr tests were not being run, `Repository.from_url` is misused here, and the creation from `dnf copr enable` is doing a lot of magic behind the scenes. Moving this to `implement` for now to focus on those issues

AthreyVinay:
Ok - so I have reverted to using  `dnf copr enable` + `guest.pull` instead of building the repo URL manually. @LecrisUT would like to know your thoughts on this! Thanks.

LecrisUT:
All looks nice and tidy in this form. Only comment would be that `self.repository` is not currently used, but maybe that is for a later change? ~~2~~3 things on the agenda following this at least:
- The `Repository.from_*` do not accurately account for multiple repos in the content
- A similar change is needed for `copr.build`
- This would probably fail on image mode

AthreyVinay:
> Only comment would be that self.repository is not currently used, but maybe that is for a later change?

That is accurate - for #4703

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -83,11 +110,22 @@ def fetch_contents(
...
-        :raises tmt.utils.PrepareError: If the package manager does not support
-            enabling Copr repositories.
+        Enable the Copr repository on the guest and retrieve the resulting
+        ``.repo`` file content.
         """
         guest.package_manager.enable_copr(self.copr_repo)
+
+        repo = parse_copr_repo(self.copr_repo)
+        owner = f'group_{repo.name}' if repo.is_group else repo.name
+        repo_filename = f"_copr:copr.fedorainfracloud.org:{owner}:{repo.project}.repo"
```
tcornell-bus, line 123: I am confused about the hard-coded `copr.fedorainfracloud.org`, copr can have other urls. Shouldn't this be pulled from the repo file?

AthreyVinay: Absolutely - if someone uses a private COPR hub, bound tobreak. Had brought this up with @LecrisUT and asked if it was a better option to read it off the client config. He felt its okay for now as we would need to open it from the `PrepareArtifacts` side to support that anyway.

LecrisUT: Yes, we first have to figure out how to pass the copr hub at the `PrepareArtifacts` level and actually pass and consume it in each of the providers (and probably other parts)

tcornell-bus: Maybe adding a quick TODO comment explaining what you said here would help?

happz:
@AthreyVinay relevant tests seem to be failing, please take a look. I'd like to review this first before diving into https://github.com/teemtee/tmt/pull/4703

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Added here: [fb24ed5](https://github.com/teemtee/tmt/pull/4695/commits/fb24ed546b8e85a0be0134201d8d571288b06d43)

AthreyVinay:
> @AthreyVinay relevant tests seem to be failing, please take a look.

Yes, this test `copr-repository` was never run earlier and hence this is the first time it was [enabled](https://github.com/teemtee/tmt/pull/4720) and is failing. Working on a probable fix.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -83,11 +110,23 @@ def fetch_contents(
...
+        # pull from guest to build Repository object with populated repo_ids
+        guest.pull(
+            source=Path(f"/etc/yum.repos.d/{repo_filename}"),
+            destination=download_path,
+        )
+
+        self.repository = Repository.from_file_path(
+            download_path / repo_filename, self.logger, name=self.copr_repo
+        )
         return []
```
thrix, line 135: Missing get_repositories() override. Without it, the Repository stored in `self.repository` is never returned to the framework. Add an override similar to `RepositoryFileProvider` and `RepositoryUrlProvider`.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -9,19 +9,44 @@
...
     ArtifactProvider,
     ArtifactProviderId,
+    Repository,
     UnsupportedOperationError,
     provides_artifact_provider,
 )
 from tmt.utils import Path
 
+COPR_REPO_PATTERN = re.compile(r'^(?P<group>@)?(?P<name>[^/]+)/(?P<project>[^/]+)$')
 COPR_REPOSITORY_PATTERN = re.compile(r'^(?:@[^/]+/[^/]+|[^@/]+/[^/]+)$')
```
thrix: Consider consolidating with `COPR_REPOSITORY_PATTERN` — they validate the same format with minor differences. It is very confusing that they are basically the same thing in terms of the name `REPO` vs  `REPOSITORY` as the difference in the name?

tests/unit/artifact/test_copr_repository.py
```diff
@@ -28,13 +28,23 @@ def test_invalid_repository_id(raw_id, error, artifact_provider):
...
-    result = provider.fetch_contents(mock_guest, tmppath / "artifacts")
+    with patch(
+        'tmt.steps.prepare.artifact.providers.copr_repository.Repository.from_file_path',
+        return_value=mock_repo,
+    ) as mock_from_file:
+        result = provider.fetch_contents(mock_guest, tmppath / "artifacts")
 
     mock_package_manager.enable_copr.assert_called_once_with('@teemtee/stable')
+    mock_guest.pull.assert_called_once()
+    mock_from_file.assert_called_once()
```
thrix: Assert on the expected `source` and `destination` arguments to verify the repo filename construction logic.

tests/unit/artifact/conftest.py
```diff
@@ -50,6 +51,13 @@ def mock_initialize_session(self, api_url=None, top_url=None):
         yield mock_koji
 
 
+@pytest.fixture
+def mock_copr():
```
thrix: Name is misleading maybe? This patches `CoprBuildArtifactProvider`, not `CoprRepositoryProvider`?Consider `mock_copr_build_session` or similar.

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: The parent provides the default implementation.

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -9,19 +9,44 @@
...
 
 
+@container(frozen=True)
+class CoprRepo:
+    is_group: bool
+    name: str
+    project: str
+
+
+def parse_copr_repo(copr_repo: str) -> CoprRepo:
```
happz: This looks like a construction function, how about making it a class method?

```python
class CoprRepo:
    ...

    @classmethod
    def from_id(cls, id: str) -> Self:
        ....
```

AthreyVinay: Addressed here: [7782a2c](https://github.com/teemtee/tmt/pull/4695/commits/7782a2cbf7ffd01c98cf50934ac401e4c7eb3691)

tests/unit/artifact/conftest.py
AthreyVinay: Addressed here: [7782a2c](https://github.com/teemtee/tmt/pull/4695/commits/7782a2cbf7ffd01c98cf50934ac401e4c7eb3691)

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Addressed here: [7782a2c](https://github.com/teemtee/tmt/pull/4695/commits/7782a2cbf7ffd01c98cf50934ac401e4c7eb3691)

tests/unit/artifact/test_copr_repository.py
AthreyVinay: Addressed here: [7782a2c](https://github.com/teemtee/tmt/pull/4695/commits/7782a2cbf7ffd01c98cf50934ac401e4c7eb3691)

AthreyVinay:
OK - So I think i see whats failing, this here: `rsync: no` is not letting `guest.pull` to copy the `.repo` file from the guest. `Repository.from_file_path()` then fails with a No such file or directory since it tries to read that file! I plan to avoid this by reading it with  `execute()` instead. :crossed_fingers:

LecrisUT:
> OK - So I think i see whats failing, this here: `rsync: no` is not letting `guest.pull` to copy the `.repo` file from the guest. `Repository.from_file_path()` then fails with a No such file or directory since it tries to read that file! I plan to avoid this by reading it with `execute()` instead. 🤞

:scream: well, that's not good. Should fix that at some point :tm:, but yeah, fine enough with me like that. I expect there are other image-mode flavored failures anyway

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -83,11 +112,22 @@ def fetch_contents(
...
         """
         guest.package_manager.enable_copr(self.copr_repo)
+
+        repo = CoprRepo.from_id(self.copr_repo)
+        owner = f'group_{repo.name}' if repo.is_group else repo.name
+        # TODO: Replace hardcoded hub URL by passing it from PrepareArtifacts to providers.
+        repo_filename = f"_copr:copr.fedorainfracloud.org:{owner}:{repo.project}.repo"
+
+        try:
+            output = guest.execute(tmt.utils.ShellScript(f"cat /etc/yum.repos.d/{repo_filename}"))
```
LecrisUT: nitpick: can construct this as `Command`. When possible to do so, please do, it can avoid one round conversion to-from command-shellscript

AthreyVinay: Addressed here: [6c6c6ec](https://github.com/teemtee/tmt/pull/4695/commits/6c6c6ece1462354e66a56f6b1dc282e020b37bce)

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -83,11 +112,24 @@ def fetch_contents(
...
+        repo_filename = f"_copr:copr.fedorainfracloud.org:{owner}:{repo.project}.repo"
+
+        try:
+            output = guest.execute(
+                tmt.utils.Command("cat", Path(f"/etc/yum.repos.d/{repo_filename}"))
+            )
+        except tmt.utils.RunError as error:
+            raise tmt.utils.PrepareError(
+                f"Failed to read '{repo_filename}' from the guest. "
+            ) from error
```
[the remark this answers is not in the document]
happz: +1 to trailing space removal, -1 to including `error` in the message, `from error` is used correctly.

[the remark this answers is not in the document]
AthreyVinay: Fied.

happz (approved):

LecrisUT:
lincheck will be fixed shortly and testing-farm failure is a fluke, merging

Issue #4645, opened 2026-03-04
All providers should list the artifacts in `artifacts.yaml`


vaibhavdaren:
Currently, discover-only providers:

- `repository-file`
- `repository-url`
- `copr.repository`

return an empty `artifacts` list. As a result, the `artifacts.yaml` file generated by `PrepareArtifact` does not include RPM metadata from repository-based providers.

Additionally, the `copr.repository` provider has issues in the `get_repository()` function as it does not return the repository name. The flow is a little different from the other providers.

## Impact

- `artifacts.yaml` does not contain repository RPM data from discover only providers.
- Behavior differs from build providers (`koji.*`, `brew.*`, `copr.build`)
- We do not have have tracking of the name of the repository added by the  copr provider.

## Goal

1. Ensure all repository-based providers enumerate available RPM packages and populate `artifacts.yaml`.
2. Fix and stabilize the `get_repository()` implementation in `copr.repository`.

### In Scope

- `repository-file`
- `repository-url`
- `copr.repository`
- `copr.repository.get_repository()` fix

### Out of Scope

- Changes to installation workflow
- Breaking changes to provider APIs

AthreyVinay:
@happz  @LecrisUT 
For build providers like `koji.*`, each entry has `version`, `nvra`, and `location` (per-RPM). Repository providers don't have individual RPMs to enumerate, so what should their entries look like? Should there be a `artifact_metadata` like `repository_metadata` property on the `ArtifactProvider` class and have `repo_name`, `base_url` & `repo_ids` as keys?

LecrisUT:
> * Behavior differs from build providers (`koji.*`, `brew.*`, `copr.build`)

Yes, because these have the rpms that we downloaded and that we want to make (more) sure that we install the correct one

> * `artifacts.yaml` does not contain repository RPM data from discover only providers.

Confusing wording, `discover` has a more specific meaning to us than what you mean in this issue. Yes, that info is not in there because we do not control the details of the repository like the priority in those cases and ultimately they are a distinct case.

> Additionally, the `copr.repository` provider has issues in the `get_repository()` function as it does not return the repository name. The flow is a little different from the other providers.

The name should be the `Repository.name`.

> All providers should list the artifacts in artifacts.yaml

[...]

PR #4703, opened 2026-03-16
List packages from repository providers in `artifacts.yaml`

AthreyVinay:
resolves: #4645

### DIFF
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -165,7 +165,7 @@ class Repository:
     @property
     def filename(self) -> str:
         """The name of the .repo file (e.g., 'my-repo.repo')."""
-        return f"{tmt.utils.sanitize_name(self.name)}.repo"
+        return f"{tmt.utils.sanitize_name(self.name, allow_slash=False)}.repo"
 
 
 class SpecialPackageOrigin(str, enum.Enum):


tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -300,6 +322,39 @@ def get_repositories(self) -> list['Repository']:
...
+                            version=RpmVersion.from_nevra(nevra),
+                            provider=self,
+                            location=repository.name,
+                        )
+                    )
+                except ValueError:
+                    self.logger.warning(f"Could not parse NEVRA '{nevra}'.")
+            self.logger.debug(
+                f"Enumerated {len(self._artifacts)} packages from repository '{repository.name}'."
+            )
```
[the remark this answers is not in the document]
AthreyVinay: fixed: [e0ec733](https://github.com/teemtee/tmt/pull/4703/commits/e0ec733c533f600200249ceb7388727681027bf1)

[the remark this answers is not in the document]
AthreyVinay: fixed: [e0ec733](https://github.com/teemtee/tmt/pull/4703/commits/e0ec733c533f600200249ceb7388727681027bf1)

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -86,6 +85,32 @@ def from_rpm_meta(cls, rpm_meta: dict[str, Any]) -> Self:
...
+
+            version_info = RpmVersion.from_nevra("curl-0:8.11.1-7.fc42.x86_64")
+            version_info = RpmVersion.from_nevra("curl-8.11.1-7.fc42.x86_64")
+        """
+        nvr_epoch, sep, arch = nevra.rpartition('.')
+        if not sep:
+            raise ValueError(f"Cannot parse arch from NEVRA '{nevra}'.")
+        parts = nvr_epoch.rsplit('-', 2)
+        if len(parts) != 3:
+            raise ValueError(f"Cannot parse NVR from NEVRA '{nevra}'.")
```
LecrisUT: Why not just `rsplit(maxsplit)` and expand the tuple. There is no expectation of this failing unless a non-nvr is provided.

But also this can be done in a single regex [`(?P<name>.*)-(?:(?P<epoch>\d+)\:)?(?P<version>.*)-(?P<release>.*)\.(?P<arch>.*)`](https://regex101.com/r/3nWbXB/1). I know we have this regex 1-2 places already even.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -191,16 +217,12 @@ def _extract_provider_id(cls, raw_id: str) -> ArtifactProviderId:
 
         raise NotImplementedError
 
-    @cached_property
+    @property
     @abstractmethod
     def artifacts(self) -> Sequence[ArtifactInfo]:
```
LecrisUT: Well, can make it non-abstract if you always provide a `_artifacts`

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -177,6 +202,7 @@ def __init__(self, raw_id: str, repository_priority: int, logger: tmt.log.Logger
         self.sanitized_id = tmt.utils.sanitize_name(raw_id, allow_slash=False)
 
         self.id = self._extract_provider_id(raw_id)
+        self._artifacts: list[ArtifactInfo] = []
```
LecrisUT: Add it to the class and document its intent

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -300,6 +322,39 @@ def get_repositories(self) -> list['Repository']:
         """
         return []
 
+    def enumerate_artifacts(
+        self, guest: Guest
+    ) -> None:  # TODO: refactor this once the NEVRA parsing is centralized.
+        """
+        Enumerate artifacts available from this provider after its repositories
+        have been installed on the guest.
+        """
```
LecrisUT, line 241: This needs a special handling when the artifacts are from `tmt-shared-repo`, otherwise all providers that write to it would overwrite each other.

tmt/steps/prepare/artifact/__init__.py
LecrisUT, line 214: The position and usage of this is now questionable again

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Kept the explicit check to give a useful error message when dnf repoquery emits unexpected non-NEVRA lines. Will change it to a regex match check.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Apologies - do not clearly understand the concern here. Can you please elaborte?

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: Yep - this was raised in the design decision. Open to ideas :)

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Fixed: [cf5cddd](https://github.com/teemtee/tmt/pull/4703/commits/cf5cddd5bd35113074bf68abb0907ff92750c3b9)

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Fixed: [cf5cddd](https://github.com/teemtee/tmt/pull/4703/commits/cf5cddd5bd35113074bf68abb0907ff92750c3b9)

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Fixed: [cf5cddd](https://github.com/teemtee/tmt/pull/4703/commits/cf5cddd5bd35113074bf68abb0907ff92750c3b9)

happz:
@AthreyVinay can you rebase, please?

AthreyVinay:
@LecrisUT I think it would make sense for #4734 to be merged first, then rebase #4695 first and then #4703 on top of that.  Then finally achieve whats intended (`tmt.package_managers.list_packages() `t o return a `list[RpmVersion]` instead of `list[str]`) #4736

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -86,6 +90,32 @@ def from_rpm_meta(cls, rpm_meta: dict[str, Any]) -> Self:
...
+            version_info = RpmVersion.from_nevra("curl-0:8.11.1-7.fc42.x86_64")
+            version_info = RpmVersion.from_nevra("curl-8.11.1-7.fc42.x86_64")
+        """
+        match = NEVRA_PATTERN.match(nevra)
+        if not match:
+            raise ValueError(f"Cannot parse NEVRA '{nevra}'.")
+        epoch_str = match.group('epoch')
+        return cls(
+            name=match.group('name'),
+            epoch=int(epoch_str) if epoch_str is not None else 0,
```
```suggestion
            epoch=int(match.group('epoch') or 0),
```

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -169,6 +199,10 @@ class ArtifactProvider(ABC):
     #: Lower values have higher priority in package managers.
     repository_priority: int
 
+    # Artifacts enumerated from this provider's repositories after they have
+    # been installed on the guest.
+    _artifacts: list[ArtifactInfo]
```
```suggestion
    #: Artifacts enumerated from this provider's repositories after they have
    #: been installed on the guest.
    _artifacts: list[ArtifactInfo]
```

LecrisUT:
> @LecrisUT I think it would make sense for #4734 to be merged first, then rebase #4695 first and then #4703 on top of that. Then finally achieve whats intended (`tmt.package_managers.list_packages() `t o return a `list[RpmVersion]` instead of `list[str]`).

Fine with me, if you can get the quick approvals for it. You have 1/2 it from me at least

AthreyVinay:
> > @LecrisUT I think it would make sense for #4734 to be merged first, then rebase #4695 first and then #4703 on top of that. Then finally achieve whats intended (`tmt.package_managers.list_packages() `t o return a `list[RpmVersion]` instead of `list[str]`).
> 
> Fine with me, if you can get the quick approvals for it. You have 1/2 it from me at least

Thanks Cris - looks like the tests are failing. Will try to look at them first.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Fixed: [9a4f30f](https://github.com/teemtee/tmt/pull/4703/commits/9a4f30f757952770917d81f3f9f8658e5eef9acb)

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Fixed: [9a4f30f](https://github.com/teemtee/tmt/pull/4703/commits/9a4f30f757952770917d81f3f9f8658e5eef9acb)

tmt/steps/prepare/artifact/providers/copr_repository.py
thrix: The regex `(?P<group>@)?(?P<name>[^/]+)` misparses `@/stable` as `name='@'` instead of rejecting it. Since _extract_provider_id validates first with the stricter COPR_REPOSITORY_PATTERN, this    
isn't reachable today — but it will be if `parse_copr_repo()` is ever called independently. Consolidate into one regex, or fix `COPR_REPO_PATTERN` to use `[^@/]+` for the non-group name case. 

Generated-by: Claude Code

tmt/steps/prepare/artifact/providers/copr_repository.py
```diff
@@ -117,7 +114,6 @@ def fetch_contents(
 
         repo = parse_copr_repo(self.copr_repo)
         owner = f'group_{repo.name}' if repo.is_group else repo.name
-        # TODO: Replace hardcoded hub URL by passing it from PrepareArtifacts to providers.
         repo_filename = f"_copr:copr.fedorainfracloud.org:{owner}:{repo.project}.repo"
```
thrix, line 120: If the file doesn't exist on the guest (naming convention change, different Copr instance), `guest.pull` will fail with an opaque error. Either wrap this in a `try/except` with a clear message,  or at minimum add a comment documenting the assumption.

Generated-by: Claude Code

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -191,20 +226,15 @@ def _extract_provider_id(cls, raw_id: str) -> ArtifactProviderId:
 
         raise NotImplementedError
 
-    @cached_property
-    @abstractmethod
+    @property
```
thrix, line 128: New providers can now forget to implement both `artifacts` and `enumerate_artifacts()` with no error. Consider adding a comment to the base class explaining the expected pattern (override artifacts for build-type providers, or rely on `enumerate_artifacts()` for repository-type providers).

Generated-by: Claude Code

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: This is updated in the base branch.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: This is valid - we can proceed in 2 ways:
1. As suggested, add a comment & forget about it.
2. Add a clear distinction(abstract base class) between download-provider & enable-provider and make these abstract methods there and all other provider classes inherit from it. This is my preference however includes  bit of a refactor. WDYT?

tmt/steps/prepare/artifact/providers/copr_repository.py
AthreyVinay: Fixed: [55abb72](https://github.com/teemtee/tmt/pull/4703/commits/55abb7217d52f8d0171874668efad9fb38b9caeb)

tmt/steps/prepare/artifact/providers/__init__.py
LecrisUT: Huh, I replied to this yesterday, what happened with it :scream:. Basically my request for this is to document in `_artifacts` explicitly which interfaces are expected to populate the fields in it. I'll go in more detail shortly

LecrisUT: So basically consider `copr_build`:
- `enumerate_artifacts` should populate the entries from all the `get_repositories` i.e. the artifacts inside the `dnf copr enable`
- `contribute_to_shared_repo` would also create another set of artifacts that need to appear in `_artifacts` that are downloaded and appear in `tmt-shared-repo`
- `enumerate_artifacts` *must not* be used for anything under `tmt-shared-repo` because at that point we have lost the link between the `dnf repoquery` used and which artifact provider created it. Should be fine in the current form because it uses `get_repositories` which do not include that one, but the documentation should be more explicit that is going on

Ultimately it would be good to document more explicitly what interfaces affect `_artifacts`, when and how. Currently there are 2 additive processes expected `enumerate_artifacts` and `fetch_content` which populate it. The former makes sense, but I would move the latter to `contribute_to_shared_repo`. And document that this the only expected interface with it + the interface to enumerate the artifacts, which is currently `_save_artifacts_metadata`.

tmt/steps/prepare/artifact/__init__.py
LecrisUT: Will postpone this when we get back to reviewing the tests for `copr_build` because that one is the case where these duplicate nvrs are expected.

tmt/steps/prepare/artifact/providers/__init__.py
LecrisUT: More detailed comment on the documentation in https://github.com/teemtee/tmt/pull/4703#discussion_r2982041970.

For the documentation of the artifacts, I would actually go another way and use `yield from`  and make this an `Iterable` instead to explicitly mark that this property is now expected to be a read-only view of `_artifacts`. But in order to that the logic that currently overrides this method needs to be moved, so I would deffer this for its own PR

tmt/steps/prepare/artifact/providers/copr_repository.py
LecrisUT: I do not believe the error message for that would be opaque since it is a very simple command, and if it is, well we need to improve the error format for it.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Would something like this enhance?
```python
@cached_property
@abstractmethod
def artifacts(self) -> Sequence[ArtifactInfo]:
    """
    Collect all artifacts available from this provider.
    
    Subclasses should follow one of two patterns:

    * **Download-only providers** (e.g. koji, copr_build): override this
      property (typically as a ``@cached_property``) to return the
      artifact list directly from the build metadata.
    * **Repository-type providers** (e.g. copr_repository): implement
      :meth:`get_repositories` so that :meth:`enumerate_artifacts` can
      populate ``_artifacts`` by querying the guest package manager.

    :returns: a list of provided artifacts.
    """

    raise NotImplementedError
```

LecrisUT: The other comment is relevant

> For the documentation of the artifacts, I would actually go another way and use `yield from` and make this an `Iterable` instead to explicitly mark that this property is now expected to be a read-only view of `_artifacts`. But in order to that the logic that currently overrides this method needs to be moved, so I would deffer this for its own PR

would rather have it in `_artifacts` and in `artifacts` have a TODO note to reshuffle stuff, and have that the more "user" facing (what it is meant to represent)

AthreyVinay: Addressed here: [40369e8](https://github.com/teemtee/tmt/pull/4703/commits/40369e8014c99519765fbb03ce56f38e050780d3)

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -153,11 +183,18 @@ class ArtifactProvider(ABC):
...
+    * **Download providers** (e.g. ``koji.build``): override :py:attr:`artifacts`
+      as a ``@cached_property``, implement :py:meth:`_download_artifact` and
+      :py:meth:`contribute_to_shared_repo`. Do not use :py:attr:`_artifacts`.
+
+    * **Repository providers** (e.g. ``copr.repository``): implement
+      :py:meth:`get_repositories`. After installation, :py:meth:`enumerate_artifacts`
+      queries the package manager and populates :py:attr:`_artifacts`.
+
+    .. TODO: Refactor into two distinct abstract base classes to make the
+       required interface explicit at the type level.
```
LecrisUT: This cannot be done because for `copr_build` you have a combination of both of them.

What should be refactored is the logic above there where "Download providers" override the `artifacts`. That is not correct because it looses the artifacts later on from `enumerate_artifacts`, and `copr_build` does both of them 🙃.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -169,6 +206,12 @@ class ArtifactProvider(ABC):
     #: Lower values have higher priority in package managers.
     repository_priority: int
 
+    #: Artifacts enumerated from this provider's repositories via
+    #: :py:meth:`enumerate_artifacts`. Populated only for repository providers
+    #: after their repositories are installed on the guest.
+    #: Not used by download providers (which override :py:attr:`artifacts` directly).
+    _artifacts: list[ArtifactInfo]
```
LecrisUT: Consequently, the ultimate purpose of this one will be to have the artifacts from both `enumerate_artifacts` and `contribute_to_shared_repo`. So you should document that it (will) contains (and accumulates) all known artifacts owned by this provider.

AthreyVinay: Addressed here: [49a1657](https://github.com/teemtee/tmt/pull/4703/commits/49a1657e0c246d57097340eec20aa8595b72403d)

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Addressed here: [49a1657](https://github.com/teemtee/tmt/pull/4703/commits/49a1657e0c246d57097340eec20aa8595b72403d)

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -300,6 +334,42 @@ def get_repositories(self) -> list['Repository']:
...
+        and populate :py:attr:`_artifacts`. Call this after repositories are installed.
+
+        For repository providers only. Does not include artifacts contributed to
+        the shared repository — those are handled by :py:meth:`contribute_to_shared_repo`.
+        """
+        for repository in self.get_repositories():
+            try:
+                nevras = guest.package_manager.list_packages(repository)
+            except tmt.utils.RunError:
+                self.logger.warning(
```
happz: While being a prototype, it would be nice to see more details about the error. Please, consider using `tmt.utils.show_exception_as_warning()` instead of `logger.warning()`.

AthreyVinay: Done: [1acaef1](https://github.com/teemtee/tmt/pull/4703/commits/1acaef1b73f40881cce58be863c2c3bd4801907b)

tmt/steps/prepare/artifact/providers/__init__.py
thrix: @AthreyVinay is this resolved?

LecrisUT: > @AthreyVinay is this resolved?

Aren't you supposed to decide that as the original author of the comment? :raised_eyebrow:

tests/prepare/artifact/lib/common.sh
```diff
@@ -19,7 +19,8 @@ setup_distro_environment() {
         rlDie "Test requires Fedora"
     fi
 
-    fedora_release="${1:-43}"
+    # TODO: Temporary hardcoded release - should be taken from function input
+    fedora_release=43
```
vaibhavdaren, line 23: [Optional][Nit] We simply, can undo this change. if param is not passed the value would be 43. Comment is fine.

psss:
@AthreyVinay, tests seem to be failing with:

    tee: /etc/yum.repos.d/mariobl/pyspread.repo: No such file or directory

tests/prepare/artifact/lib/common.sh
AthreyVinay: This already existed on main - do not want to make any unrelated changes here.

LecrisUT: Please check again with the current state of `main`. This is a rebase issue I believe
https://github.com/teemtee/tmt/blob/a18f8059fd181d9bfe62cfbd56af8ef8fb12521f/tests/prepare/artifact/lib/common.sh#L17-L25
And as pointed previously this overlaps with #4723

AthreyVinay: True  - but wasnt the consensus to leave it as-is i.e. revert the above change?

AthreyVinay:
> @AthreyVinay, tests seem to be failing with:
> 
> ```
> tee: /etc/yum.repos.d/mariobl/pyspread.repo: No such file or directory
> ```

Yes, apparently, the rebase had reverted my commit for this fix.

tests/prepare/artifact/lib/common.sh
LecrisUT: Yes, and isn't that also what @vaibhavdaren is suggesting? :raised_eyebrow:

AthreyVinay: I do understand the default would be 43.... however this patch was introduced as part of a fix that I thought would be for a failure - however, the state before this was just:
```python
# TODO: Temporary hardcoded release - should be taken from function input
fedora_release=43
```

LecrisUT: > I do understand the default would be 43.... however this patch was introduced as part of a fix that I thought would be for a failure - however, the state before this was just:
> 
> ```python
> # TODO: Temporary hardcoded release - should be taken from function input
> fedora_release=43
> ```

So some questions to be confirmed
- Do the tests fail if these are reverted?
- You are saying that you want to revert a previous commit of yours https://github.com/teemtee/tmt/commit/0e31990f5e6aec18070669f00d7767863b35878b. But what would be the issue if the revert is happening in #4723 instead?

AthreyVinay: No & yes (simply because I introduced the change, nothing other than that!) & also tired of rebasing(mightve done at-least 10 times since last sprint :D )

LecrisUT: > No & yes (simply because I introduced the change, nothing other than that!)

Well, let's just drop that one, it will be addressed in #4723. But ultimately I am fine with either way since it would just mean a minor rebase when it's that PR's turn, so feel free to resolve this discussion.

>also tired of rebasing(mightve done at-least 10 times since last sprint :D )

What are you using for the rebase? On my side, PyCharm is fairly consistent to account for commits that were already merged. Seems it is not as reliable on your IDE?

AthreyVinay: No fancy IDE - manually rebasing :/

tmt/steps/prepare/artifact/providers/__init__.py
thrix: @LecrisUT yes, but I would expect a comment from the author :)

thrix: that something was done

AthreyVinay: @thrix In fairness, a [question](https://github.com/teemtee/tmt/pull/4703#discussion_r2981288940) was asked but was not answered. The conversation got extended as part of another thread which might've caused this confusion. All good.

PR #4704, opened 2026-03-16
Document tree handling and workdirs

### LINKED ISSUE #4663: Document `user tree`, `work tree` and `test tree` in the Glossary
As mentioned in #4602, there's a lot of fmf trees being used all around, let's give them some clear names so that we can clearly distinguish what we're speaking about. These should be documented in the Glossary, depends on #4650.

### user tree

The initial `fmf` tree from which `tmt` commands are executed is called **user tree**. For example:

    /home/psss/git/tests/nested

By default it is detected from the current working directory. It can be directly `$PWD` or the closest parent which contains `.fmf` directory marking root of the `fmf` tree as implemented by `fmf`. Alternatively it can be provided using the `--root` option.

All the following cases result in the `fmf` root mentioned above:

    tmt --root /home/psss/git/tests/nested
    tmt --root /home/psss/git/tests/nested/deeper
    cd /home/psss/git/tests/nested && tmt
    cd /home/psss/git/tests/nested/deeper && tmt

### work tree

When a new run is created, the *user tree* is copied into the run workdir under the `tree` folder and its path is made available in the `TMT_TREE` environment variable. For example this is a **runner work tree**:

    runner:/var/tmp/tmt/run-123/plan/tree

Runner work tree is synced to the **guest work tree** so that tests defined by the discover shell plugin or prepare shell phases can use files from the *user tree*.

    guest:/var/tmp/tmt/run-123/plan/tree

### test tree

The discover plugin prepares `tests` directory where test scripts and files are available for test execution, let's call it **test tree**.

The `fmf` discover plugin copies *user tree* to *test tree* when `url` is not provided. Otherwise the whole git repository is cloned into it.

The `shell` discover plugin creates a symlink to the *work tree*:

```
    ├── discover
    │   ├── default-0
    │   │   └── tests -> ../../tree
    │   ├── step.yaml
    │   └── tests.yaml
```

psss:
Define `user tree`, `work tree` and `test tree`.
Document `run workdir` and `plan workdir` as well.
Add the `Tree Handling` section for contributors.

Fix #4663.

Pull Request Checklist

* [x] write the documentation
* [x] include a release note

### DIFF
--- a/docs/code/trees.rst
+++ b/docs/code/trees.rst
@@ -16,7 +16,7 @@ User Tree
 ------------------------------------------------------------------
 
 The initial ``fmf`` tree from which ``tmt`` commands are executed
-is called **user tree**. For example:
+is called :term:`user tree`. For example:
 
 .. code-block::
 
@@ -44,18 +44,18 @@ above:
 Work Tree
 ------------------------------------------------------------------
 
-When a new run is created, the **user tree** is copied into the
-plan workdir under the ``tree`` folder and its path is made
+When a new run is created, the :term:`user tree` is copied into
+the plan workdir under the ``tree`` folder and its path is made
 available in the ``TMT_TREE`` environment variable. For example
-this is a **runner work tree**:
+this is a :term:`runner` :term:`work tree`:
 
 .. code-block::
 
     runner:/var/tmp/tmt/run-123/plan/tree
 
-Runner work tree is synced to the **guest work tree** so that
-discovered tests, prepare and finish steps can use files and
-scripts from there.
+Runner work tree is synced to the :term:`guest` :term:`work tree`
+so that discovered tests, prepare and finish steps can use files
+and scripts from there.
 
 .. code-block::
 
@@ -69,15 +69,15 @@ Test Tree
 
 The discover plugin prepares a ``tests`` directory where test
 scripts and files are available for test execution. This directory
-is called a **test tree**.
+is called a :term:`test tree`.
 
-The :ref:`/plugins/discover/fmf` discover plugin copies **user
-tree** to **test tree** when ``url`` is not provided. Otherwise
-the whole git repository is cloned into it.
+The :ref:`/plugins/discover/fmf` discover plugin copies
+:term:`user tree` to :term:`test tree` when ``url`` is not
+provided. Otherwise the whole git repository is cloned into it.
 
 The :ref:`/plugins/discover/shell` discover plugin creates a
-symlink to the **work tree** so the resulting structure looks like
-this:
+symlink to the :term:`work tree` so the resulting structure looks
+like this:
 
 .. code-block::
 


docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+workdir under the ``tree`` folder and its path is made available
+in the ``TMT_TREE`` environment variable. For example this is a
+**runner work tree**:
+
+.. code-block::
+
+    runner:/var/tmp/tmt/run-123/plan/tree
+
+Runner work tree is synced to the **guest work tree** so that
+tests defined by the discover shell plugin or prepare shell phases
```
mcasquer: If the plugin is explicitly mentioned... what about?
```suggestion
tests defined by the :ref:`discover shell</plugins/discover/shell>` plugin or prepare shell phases
```
Ps: not sure if that's the correct syntax 😅

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+
+The discover plugin prepares ``tests`` directory where test
+scripts and files are available for test execution, let's call it
+a **test tree**.
+
+The ``fmf`` discover plugin copies *user tree* to *test tree* when
+``url`` is not provided. Otherwise the whole git repository is
+cloned into it.
+
+The ``shell`` discover plugin creates a symlink to the *work
```
mcasquer: Perhaps another mention to :ref:\`/plugins/discover/shell\` ?

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+.. _test-tree:
+
+Test Tree
+------------------------------------------------------------------
+
+The discover plugin prepares ``tests`` directory where test
+scripts and files are available for test execution, let's call it
+a **test tree**.
+
+The ``fmf`` discover plugin copies *user tree* to *test tree* when
```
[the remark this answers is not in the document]
psss: No, that is not the case.

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+**runner work tree**:
+
+.. code-block::
+
+    runner:/var/tmp/tmt/run-123/plan/tree
+
+Runner work tree is synced to the **guest work tree** so that
+tests defined by the discover shell plugin or prepare shell phases
+can use files from the *user tree* when scripts are executed on
+the guest.
```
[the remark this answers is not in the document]
psss: Hm, makes sense. Shortened a bit in ac7d917d.

docs/glossary.rst
```diff
@@ -14,3 +14,34 @@ Machines
...
+Directories
+-----------
+
+.. glossary::
+
+    test tree
+        Directory ``tests`` containing test code prepared by the
+        ``discover`` step, e.g.
+        ``/var/tmp/tmt/run-123/plan/discover/default-0/tests``.
+        See :ref:`test-tree` for details.
```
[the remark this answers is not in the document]
psss: Used consistent plan name `/plans/core` in ac7d917d.

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
+.. _trees:
+
+======================
+    Tree Handling
+======================
+
+The ``fmf`` trees holding test, plan and story metadata are used
+at several places of the code and their data is copied to multiple
+locations. Here we give them clear names so that we can clearly
+distinguish what we're speaking about.
```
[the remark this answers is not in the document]
psss: Sounds ok: ac7d917d

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+
+User Tree
+------------------------------------------------------------------
+
+The initial ``fmf`` tree from which ``tmt`` commands are executed
+is called **user tree**. For example:
+
+.. code-block::
+
+    /home/psss/git/tests/nested
```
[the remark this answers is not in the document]
psss: Makes sense: ac7d917d

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+
+
+.. _test-tree:
+
+Test Tree
+------------------------------------------------------------------
+
+The discover plugin prepares ``tests`` directory where test
+scripts and files are available for test execution, let's call it
+a **test tree**.
```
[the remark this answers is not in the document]
psss: Ack: ac7d917d.

docs/glossary.rst
```diff
@@ -14,3 +14,34 @@ Machines
...
+        :ref:`work-tree` for details.
+
+    run workdir
+        Working directory created for each run used for storing various
+        temporary files needed during the test discovery and execution,
+        e.g. ``/var/tmp/tmt/run-123/``.
+
+    plan workdir
+        Directory used for storing temporary files related to particular
+        plan, e.g. ``/var/tmp/tmt/run-123/plan/one``.
```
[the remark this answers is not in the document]
psss: Ack, consistent plan names added in ac7d917d.

docs/code/trees.rst
psss: Yeah, more cross-referencing never hurts: 8fbcc3cc

docs/code/trees.rst
psss: Ack: 8fbcc3cc

docs/code/trees.rst
```diff
@@ -0,0 +1,89 @@
...
+in the ``TMT_TREE`` environment variable. For example this is a
+**runner work tree**:
+
+.. code-block::
+
+    runner:/var/tmp/tmt/run-123/plan/tree
+
+Runner work tree is synced to the **guest work tree** so that
+tests defined by the :ref:`/plugins/discover/shell` discover
+plugin or prepare :ref:`/plugins/prepare/shell` phases can use
```
happz: These two mentions look too much like the only options. How about something along the lines of "... tests defined by discover plugins (like shell) or prepare plugins (like another shell) ..."?

docs/code/trees.rst
```diff
@@ -0,0 +1,89 @@
...
+    cd /home/user/git/tests/nested && tmt
+    cd /home/user/git/tests/nested/deeper && tmt
+
+
+.. _work-tree:
+
+Work Tree
+------------------------------------------------------------------
+
+When a new run is created, the *user tree* is copied into the run
```
happz:
```suggestion
When a new run is created, the **user tree** is copied into the run
```

docs/code/trees.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Test Tree
+------------------------------------------------------------------
+
+The discover plugin prepares a ``tests`` directory where test
+scripts and files are available for test execution. This directory
+is called a **test tree**.
+
+The :ref:`/plugins/discover/fmf` discover plugin copies *user
+tree* to *test tree* when ``url`` is not provided. Otherwise the
```
happz:
```suggestion
tree* to **test tree** when ``url`` is not provided. Otherwise the
```

docs/code/trees.rst
```diff
@@ -0,0 +1,89 @@
...
+whole git repository is cloned into it.
+
+The :ref:`/plugins/discover/shell` discover plugin creates a
+symlink to the *work tree* so the resulting structure looks like
```
happz:
```suggestion
symlink to the **work tree** so the resulting structure looks like
```

docs/glossary.rst
```diff
@@ -14,3 +14,34 @@ Machines
...
+        to this directory is available in the ``TMT_TREE`` environment
+        variable, e.g. ``/var/tmp/tmt/run-123/plans/core/tree``. See
+        :ref:`work-tree` for details.
+
+    run workdir
+        Working directory created for each run used for storing various
+        temporary files needed during the test discovery and execution,
+        e.g. ``/var/tmp/tmt/run-123/``.
+
+    plan workdir
```
happz: Can the entries be sorted?

docs/code/trees.rst
psss: I intentionally used bold when the term is defined and italics when the term is referenced. Otherwise the text would be full of bold words. Do you find this confusing?

happz: Kind of, yes. Or more like hiding the referenced trees in the text, harder to spot them.

docs/glossary.rst
psss: Sure: 02ffd283

docs/code/trees.rst
psss: Hm, yeah, right, even `fmf` discovered tests or `ansible` prepare phases can use some file from the work tree. Updated the description to be more general: 02ffd283

docs/code/trees.rst
psss: Ok, let's try: a40c579e

docs/code/trees.rst
psss: Done: a40c579e

docs/code/trees.rst
psss: Done: a40c579e

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+workdir under the ``tree`` folder and its path is made available
+in the ``TMT_TREE`` environment variable. For example this is a
+**runner work tree**:
+
+.. code-block::
+
+    runner:/var/tmp/tmt/run-123/plan/tree
+
+Runner work tree is synced to the **guest work tree** so that
+discovered tests, prepare and finish steps can can use files and
```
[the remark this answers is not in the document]
psss: Fixed: 2e2bd445

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
...
+
+.. _work-tree:
+
+Work Tree
+------------------------------------------------------------------
+
+When a new run is created, the **user tree** is copied into the
+plan workdir under the ``tree`` folder and its path is made
+available in the ``TMT_TREE`` environment variable. For example
+this is a **runner work tree**:
```
LecrisUT: The main benefit of the glossary is if it is combined with `:term:`, e.g. here it would be

```suggestion
this is a :term:`runner` **work tree**:
```

Giving the back-references.

docs/code/trees.rst
```diff
@@ -0,0 +1,88 @@
+.. _trees:
+
+======================
+    Tree Handling
+======================
```
LecrisUT, line 5: I wonder if we should be making this split into individual sections instead of putting it all in the glossary. I am torn because in one case we have short-concise definitions for the `glossary` and in the other we have a shorter TOC and less clicks to navigate from the `:term:` which should be the main entry-point for these.

Maybe we can redefine the glossary to be broader in scope? At the same time we have all of the `spec` that also has definitions 🤔

docs/glossary.rst
```diff
@@ -14,3 +14,34 @@ Machines
...
       The machine that is running the ``tmt`` command.
+
+Directories
+-----------
+
+.. glossary::
+
+    plan workdir
+        Directory used for storing temporary files related to particular
+        plan, e.g. ``/var/tmp/tmt/run-123/plans/core``.
```
LecrisUT, line 25: For completeness there should also be `test workdir` which just states that it is the same as `test tree`

docs/code/trees.rst
psss: Nice! Let's use the terms: cf6e1c66

docs/code/trees.rst
psss: Do you suggest to merge the code implementation details into the Glossary so that everything's at one place? I'm not sure about that: I'd say that for a common user, the concise definition is sufficient. I'm afraid having all the implementation details would not help but rather confuse them.

docs/glossary.rst
psss: Hm, I don't think it makes sense to duplicate definitions. Furthermore, `test workdir` sounds a bit more like the current working directory used when a particular test is executed, e.g. `test tree` + `test.path`.

happz: It'd be nice to define these paths too, and show examples. I'm dealing with tmt for quite some time, and I still can't get my head around various test-related paths, test data dir, plan dir, etc. I'm not necessarily insisting on doing it now, in this PR, but "`test tree` + `test.path`" should have a name, and should be demonstrated on examples, together "test datadir" and others.

LecrisUT: > Furthermore, `test workdir` sounds a bit more like the current working directory used when a particular test is executed, e.g. `test tree` + `test.path`.

That is precisely why I think we should be documenting the `test workdir` like that. The test workdir is by default the `test tree`, but it can be changed with `test.path`. Crucially we are stating that that is consistent expectation (looking at you `dist-git-*`) that the user can rely on.

> "test datadir" and others.

Precisely, that is what makes the workdir vs tree confusing. The datadir is unique to each test, but the workdir and test tree is not, and is only unique to the discover (and what about upgrade? :/)

docs/code/trees.rst
LecrisUT: Yeah, I am quite torn on this, I can see it in both ways. Maybe it's the split across all the `spec`, `code`, `stories` etc. I am not sure what a good split would be, but I guess it would involve a few main TOC sections:
- How it works under the hood
  - The basic split/flow of `user tree` + (user config) -> `run` -> `step` -> `phase` -> `test`
  - The various trees and directories
  - The split between guest
  - Plugins
  - (Recipe eventually)
  - Glossary
  - Advance topics:
    - Phases that create other phases
    - Profiles/Policies
    - Phases that jump across steps, e.g. `dist-git` discover and `upgrade`
- What feature is available right now/How to use `tmt`
  - Specification/Features what you can do: multi-host, restart, hardware, etc.
  - External plugins
  - Examples and guides
  - Projects that use tmt and their feedback
- Code details
  - Python API
  - Config reference/index

Introspecting my thoughts, maybe the glossary being more visible than the breakdown is what feels off. The former is currently as a simple index

docs/glossary.rst
psss: > The test workdir is by default the test tree, but it can be changed with test.path.

Actually, this does not apply for `fmf` discovered tests: Their current working directory defaults to the `test tree` + `path` and `path` defaults to the directory where test's metadata are defined.

> I'm not necessarily insisting on doing it now, in this PR, but "`test tree` + `test.path`" should have a name, and should be demonstrated on examples, together "test datadir" and others.

Completely agree. Would be good to have them clearly documented as well. Let's do it in a follow-up pull request not to extend the scope of this more and more.

docs/code/trees.rst
psss: As agreed this will be covered in #4651.

psss:

PR #4705, opened 2026-03-16
Control which runner devices can be exposed to guests

happz:
As requested in [1], `container` plugin would expose some runner devices to containers representing guests. But we do not want to expose any device requested, it seems wise to give runner control over which devices would be accessible. The patch adds an option/envvar to control this, by giving provision plugins a way to consult a list of regular expressions - only devices whose path matches a pattern would be accessible should guest ask.

[1] https://github.com/teemtee/tmt/pull/4333

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note

### DIFF
--- a/tmt/options.py
+++ b/tmt/options.py
@@ -247,7 +247,7 @@ AGAIN_OPTION: list[ClickOptionDecoratorType] = [
     ),
 ]
 
-FEELING_SAFE_OPTIONS: list[ClickOptionDecoratorType] = [
+SECURITY_OPTIONS: list[ClickOptionDecoratorType] = [
     option(
         '--feeling-safe',
         metavar='FEELING_SAFE',
--- a/tmt/cli/_root.py
+++ b/tmt/cli/_root.py
@@ -52,7 +52,7 @@ dry_options = create_options_decorator(tmt.options.DRY_OPTIONS)
 force_dry_options = create_options_decorator(tmt.options.FORCE_DRY_OPTIONS)
 again_option = create_options_decorator(tmt.options.AGAIN_OPTION)
 fix_options = create_options_decorator(tmt.options.FIX_OPTIONS)
-feeling_safe_options = create_options_decorator(tmt.options.FEELING_SAFE_OPTIONS)
+security_options = create_options_decorator(tmt.options.SECURITY_OPTIONS)
 import_before_name_filter_option = create_options_decorator(
     tmt.options.IMPORT_BEFORE_NAME_FILTER_OPTION
 )
@@ -139,7 +139,7 @@ def _load_policies(
          """,
 )
 @verbosity_options
-@feeling_safe_options
+@security_options
 @import_before_name_filter_option
 @option(
     '--show-time',
--- a/tmt/steps/__init__.py
+++ b/tmt/steps/__init__.py
@@ -1663,7 +1663,7 @@ class BasePlugin(
             tmt.options.VERBOSITY_OPTIONS
             + tmt.options.FORCE_DRY_OPTIONS
             + tmt.options.AGAIN_OPTION
-            + tmt.options.FEELING_SAFE_OPTIONS
+            + tmt.options.SECURITY_OPTIONS
         )
 
     @classmethod


tmt/options.py
```diff
@@ -261,7 +261,18 @@ def option(
              Do not use this option unless you trust metadata consumed
              by tmt, or unless you know what you are doing.
              """,
-    )
+    ),
+    option(
+        '--acessible-runner-devices',
```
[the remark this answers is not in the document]
happz: Addressed in 4f454617cdef23592bdce9f01d892a38b4575641

tmt/options.py
```diff
@@ -261,7 +261,18 @@ def option(
...
+        metavar='PATTERN',
+        envvar='TMT_ACCESSIBLE_RUNNER_DEVICES',
+        multiple=True,
+        help="""
+             Guests may require access to devices of the runner, and
+             only devices whose path matches this regular expression
+             would be made accessible.
```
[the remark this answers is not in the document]
happz: Addressed in 4f454617cdef23592bdce9f01d892a38b4575641.

tmt/utils/__init__.py
```diff
@@ -2171,6 +2171,14 @@ def import_before_name_filter(self) -> bool:
         # support missing `key` parameter
         return self._get_cli_flag('import_before_name_filter', 'import_before_name_filter', False)
 
+    @property
+    def accessible_runner_device_patterns(self) -> list[Pattern[str]]:
+        """
+        A list of patterns accessible runner devices must conform to.
+        """
+
+        return [re.compile(pattern) for pattern in self.opt('accessible_runner_devices')]
```
skycastlelily: You might want to add regex pattern validation with early feedback here

docs/releases/pending/4705.fmf
```diff
@@ -0,0 +1,7 @@
+description: |
+  Adding new option of tmt command, ``--allowed-runner-devices``, and a
```
skycastlelily: Adding new option of tmt command, ``--accessible-runner-devices``, and a

skycastlelily (approved):
LGTM,other than the comments^^

skycastlelily:
This mr blocks #4333 ,right? If so, we need add it to 1.70:)

tmt/utils/__init__.py
happz: Added a better exception, but no early detection. We can do that via Click for all options later.

docs/releases/pending/4705.fmf
happz: Fixed. I even renamed the option to better match its purpose.

tmt/utils/__init__.py
```diff
@@ -2171,6 +2171,22 @@ def import_before_name_filter(self) -> bool:
...
+        A list of patterns exposable runner devices must conform to.
+        """
+
+        try:
+            return [re.compile(pattern) for pattern in self.opt('exposable_runner_devices')]
+
+        except re.error as exc:
+            faulty_pattern = (
+                exc.pattern.decode() if isinstance(exc.pattern, bytes) else exc.pattern
+            )
```
LecrisUT, line 2186: Don't think this part will get triggered, but fine either way with it

happz: Probably not, but mypy sees `exc.pattern` as either `str` or `bytes`, and demands we handle `bytes` in one way or the other.

tmt/options.py
```diff
@@ -247,7 +247,7 @@ def option(
     ),
 ]
 
-FEELING_SAFE_OPTION: list[ClickOptionDecoratorType] = [
+FEELING_SAFE_OPTIONS: list[ClickOptionDecoratorType] = [
```
psss: Hm, one more thought. Maybe we should call the group something like `SECURITY_OPTIONS`?

happz: Sure, why not. Renamed in 15d4d2ce.

tmt/utils/__init__.py
LecrisUT: Weird, it should have been typed by `Pattern[str]` 🤷. It's probably because `self.opt` is `Any` in there

tmt/options.py
psss: Thanks!

PR #4710, opened 2026-03-18
Rename "Pass" Github action

happz:
"Pass" alone is not clear, renaming the action to give it more explanative name among other checks.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
     needs: [ pre-commit ]
 
   overall-outcome:
-    name: ✅ GitHub Actions Overall Outcome
+    name: GitHub Actions Overall Outcome
     if: always()
     needs: [ pre-commit, build-wheel, shellcheck, doc-tests ]
     runs-on: ubuntu-slim


.github/workflows/ci.yml
```diff
@@ -45,8 +45,8 @@ jobs:
       ref: ${{ inputs.ref }}
     needs: [ pre-commit ]
 
-  pass:
-    name: ✅ Pass
+  overall-outcome:
+    name: ✅ GitHub Actions Overall Outcome
```
LecrisUT:
```suggestion
    name: ✅ Overall Outcome
```
Think we can drop the `GitHub Actions` part. This is always prefixed by `CI / ` and `CI / ✅ Overall Outcome` seems less redundant?

happz: I'd like to keep the "Gihub action" bit part of the name. My reasoning is that most of the checks are part of the "CI suite" that runs for each PR, to test its content in some way. "CI / Overall Outcome" suggests it's out for all CI jobs there might be, possibly including TF jobs. Which I find puzzling.

happz: I'm open to suggestions but I'd like to express that this "Overall Outcome" speaks for just a subset of tests, not for all of them.

LecrisUT: Fair enough point. And actually I wonder if it would not actually be better to have this aggregator for those as well. Reasoning being that with GH actions we actually get notifications of failing jobs. Well a topic for later though

LecrisUT (approved):

thrix (approved):
Lgtm

.github/workflows/ci.yml
psss: Could we also remove the tick? It's a bit confusing as another tick or red cross is provided by GitHub itself.

happz: It was added in one of the previous patches by @LecrisUT , together with a similar sign in `doc-tests`:

```yaml
doc-tests:
  name: 📘 docs
```

I don't care about the tick mark or 📘, I just see it as another issue, whether we want this kind of extra signaling.

LecrisUT:
```suggestion
    name: GitHub Actions Overall Outcome
```
No strong opinions on with or without. I generally use them in other projects because it helps with navigability.

psss:
Just github action changes, works as expected, merging.

PR #4712, opened 2026-03-18
Use `ruamel.yaml.clib` as default

LecrisUT:
It seems the only thing we need to do to enable this is to pull in this dependency and switch to using `safe` type by default. This would be a better apple-to-apple comparison with #4486 since that one uses a C library for handling json.

Note that the clib does not respect `.indent` formatting hints

> ### Indenting using typ="safe"
>
> The C based emitter doesn’t have the fine control, distinguishing between block mappings and sequences. Do only use the pure Python versions of the dumper if you want to have that sort of control.

Reference:
- Upstream documentation: https://yaml.dev/doc/ruamel.yaml/
- Note about `yaml.indent` https://yaml.dev/doc/ruamel.yaml/detail

---

Pull Request Checklist

* [x] implement the feature
* [ ] include a release note

Related to #4406

### DIFF
--- a/pylock.toml
+++ b/pylock.toml
@@ -1924,6 +1924,7 @@ name = "pylero"
 version = "0.1.1"
 index = "https://pypi.org/simple"
 sdist = { url = "https://files.pythonhosted.org/packages/88/38/0c0ccaafbb8594cf50af8d2376a5afee9e7279b7715a928558e7b52eb6f6/pylero-0.1.1.tar.gz", upload-time = 2025-05-30T13:38:52Z, size = 309121, hashes = { sha256 = "de0ccd37da69e50993fe403eca5d093d70c57319640d6af6403ab9a3496ae16c" } }
+wheels = [{ url = "https://files.pythonhosted.org/packages/0b/79/4a7ff895325f7226f846cf4e274be9bbeedcd2d9804027c5025e72134ff4/pylero-0.1.1-py3-none-any.whl", upload-time = 2025-12-19T15:11:28Z, size = 101338, hashes = { sha256 = "ada04668e36adaaed950e213699f8442466994142c127a3f07b5e4d19fc9709f" } }]
 
 [[packages]]
 name = "python-bugzilla"
@@ -2402,7 +2403,6 @@ wheels = [{ name = "ruamel_yaml-0.18.16-py3-none-any.whl", url = "https://files.
 [[packages]]
 name = "ruamel-yaml-clib"
 version = "0.2.15"
-marker = "python_full_version < '3.14' and platform_python_implementation == 'CPython'"
 index = "https://pypi.org/simple"
 sdist = { url = "https://files.pythonhosted.org/packages/ea/97/60fda20e2fb54b83a61ae14648b0817c8f5d84a3821e40bfbdae1437026a/ruamel_yaml_clib-0.2.15.tar.gz", upload-time = 2025-11-16T16:12:59Z, size = 225794, hashes = { sha256 = "46e4cc8c43ef6a94885f72512094e482114a8a706d3c555a34ed4b0d20200600" } }
 wheels = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ dependencies = [
     "pygments>=2.7.4",
     "requests>=2.25.1",
     "ruamel.yaml>=0.16.6",
+    "ruamel-yaml-clib",
     "urllib3>=1.26.5, <3.0",
     "typing-extensions>=4; python_version < '3.13'",
     ]
--- a/tests/plan/export/test.sh
+++ b/tests/plan/export/test.sh
@@ -16,11 +16,11 @@ rlJournalStart
 
     rlPhaseStartTest
         rlRun -s "tmt plan export ." 0 "Export plan"
-        rlAssertGrep "- name: /plan/basic" $rlRun_LOG
-        rlAssertGrep "- name: /plan/context" $rlRun_LOG
-        rlAssertGrep "- name: /plan/environment" $rlRun_LOG
-        rlAssertGrep "- name: /plan/gate" $rlRun_LOG
-        rlAssertGrep "- name: /plan/extra-keys" $rlRun_LOG
+        rlAssertGrep "name: /plan/basic" $rlRun_LOG
+        rlAssertGrep "name: /plan/context" $rlRun_LOG
+        rlAssertGrep "name: /plan/environment" $rlRun_LOG
+        rlAssertGrep "name: /plan/gate" $rlRun_LOG
+        rlAssertGrep "name: /plan/extra-keys" $rlRun_LOG
         rlAssertGrep "discover:" $rlRun_LOG
         rlAssertGrep "execute:" $rlRun_LOG
         assert_internal_fields "$rlRun_LOG"
@@ -28,7 +28,7 @@ rlJournalStart
 
     rlPhaseStartTest "tmt plan export /plan/basic"
         rlRun -s "tmt plan export /plan/basic" 0 "Export plan"
-        rlAssertGrep "- name: /plan/basic" $rlRun_LOG
+        rlAssertGrep "name: /plan/basic" $rlRun_LOG
         rlAssertGrep "summary: Just basic keys." $rlRun_LOG
         rlAssertGrep "discover:" $rlRun_LOG
         rlAssertGrep "execute:" $rlRun_LOG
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -48,6 +48,7 @@ dependencies = [
     "pygments>=2.7.4",
     "requests>=2.25.1",
     "ruamel.yaml>=0.16.6",
+    "ruamel-yaml-clib",
```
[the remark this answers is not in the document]
LecrisUT: Interesting observation. We are implying CPython in many other places though due to other dependencies. Will keep it open for discussions

tmt/utils/__init__.py
```diff
@@ -3522,6 +3506,8 @@ def _yaml(
     :returns: a loader/dumper instance.
     """
 
+    # We use the safe mode as default in order to use CParser from ruamel.yaml.clib
+    yaml_type = yaml_type or "safe"
```
happz: `yaml_type: Optional[YamlTypType] = 'safe'` seems like a better change to me, the default value clearly visible in the signature.

LecrisUT: I have considered it initially. It requires to propagate this to half-dozen functions, but if the bigger change would be ok, I can change it to that.

LecrisUT: Done in 99b6d3ff90bdb365234360fb006e5ae5e8e16441

Cases like `/tests/lint/test#/fix` are unhappy with this though :thinking:

LecrisUT: Fixed the `rt` issue with d0623b7 and there were formatting differences 39cea2e which I suspect will conflict with #4338. Shall we make this blocked on that PR?

happz: > Shall we make this blocked on that PR?

Probably, sounds like a right move.

LecrisUT:
Blocked by https://github.com/teemtee/tmt/pull/4338

tmt/base/core.py
```diff
@@ -1712,12 +1712,12 @@ def lint_legacy_relevancy_rules(self) -> LinterReturn:
             return
 
         filename = self.fmf_sources[-1]
-        metadata: dict[str, Any] = tmt.utils.yaml_to_dict(self.read(filename))
+        metadata: dict[str, Any] = tmt.utils.yaml_to_dict(self.read(filename), yaml_type="rt")
```
happz, line 1716: Is `rt` working better than the default `safe`?

LecrisUT: This part is necessary there because it is doing a lint fix and without it the comments would not be preaerved

happz: A comment would be nice, so our future selves would not need to rely on their memory. Or we can add a parameter to `yaml*` helpers, e.g. `preserve_comments: bool`, which would enforce `yaml_type='rt' when set.

tests/plan/export/test.sh
```diff
@@ -16,19 +16,19 @@ rlJournalStart
...
-        rlAssertGrep "- name: /plan/basic" $rlRun_LOG
-        rlAssertGrep "- name: /plan/context" $rlRun_LOG
-        rlAssertGrep "- name: /plan/environment" $rlRun_LOG
-        rlAssertGrep "- name: /plan/gate" $rlRun_LOG
-        rlAssertGrep "- name: /plan/extra-keys" $rlRun_LOG
+        rlAssertGrep "name: /plan/basic" $rlRun_LOG
+        rlAssertGrep "name: /plan/context" $rlRun_LOG
+        rlAssertGrep "name: /plan/environment" $rlRun_LOG
+        rlAssertGrep "name: /plan/gate" $rlRun_LOG
+        rlAssertGrep "name: /plan/extra-keys" $rlRun_LOG
```
thrix, line 23: How are these changes related to the PR, pls add a comment to the MR description. Smells like unrealted change to me

pylock.toml
```diff
@@ -1924,6 +1924,7 @@ name = "pylero"
 version = "0.1.1"
 index = "https://pypi.org/simple"
 sdist = { url = "https://files.pythonhosted.org/packages/88/38/0c0ccaafbb8594cf50af8d2376a5afee9e7279b7715a928558e7b52eb6f6/pylero-0.1.1.tar.gz", upload-time = 2025-05-30T13:38:52Z, size = 309121, hashes = { sha256 = "de0ccd37da69e50993fe403eca5d093d70c57319640d6af6403ab9a3496ae16c" } }
+wheels = [{ url = "https://files.pythonhosted.org/packages/0b/79/4a7ff895325f7226f846cf4e274be9bbeedcd2d9804027c5025e72134ff4/pylero-0.1.1-py3-none-any.whl", upload-time = 2025-12-19T15:11:28Z, size = 101338, hashes = { sha256 = "ada04668e36adaaed950e213699f8442466994142c127a3f07b5e4d19fc9709f" } }]
```
thrix, line 1927: hmm, any idea how is this change related to the PR?

tmt/base/core.py
LecrisUT: Oh, totally my bad. I commented on the need of consistency in `_yaml`, but did not clarify it here also. Added comment in 99753b201b748717cc635af98a7cd3fcbef3a131

pyproject.toml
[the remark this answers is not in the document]
thrix: in my POW the tmt documentation should mention the projects currently supports only CPython. I would be fine with saying that.

tests/plan/export/test.sh
LecrisUT: `ruamel.yaml.clib` uses a different method of constructing the yaml, in this case the order of the items change so
```yaml
hello:
  - foo: 1
    bar: 2
```
became
```yaml
hello:
  - bar: 2
    foo: 1
```
and the regex failed because it was expecting the `-`.

As for documenting in the MR description,

> > ### Indenting using typ="safe"
> >
> > The C based emitter doesn’t have the fine control, distinguishing between block mappings and sequences. Do only use the pure Python versions of the dumper if you want to have that sort of control.

was support to cover that

pylock.toml
LecrisUT: New dependency -> `uv.lock` and `pylock.toml` change

pyproject.toml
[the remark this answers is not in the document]
LecrisUT: I do not think it is relevant right now since we do not know what other implications we hard-coded that makes it CPython specific

> We are implying CPython in many other places though due to other dependencies.

CPython is the default implementation and we do not actively package this for other python implementations or advertise that with the identifiers. I would rather keep it as-is until someone comes to complain

happz (approved):

tmt/utils/__init__.py
```diff
@@ -3592,7 +3594,9 @@ def _sanitize_yaml_tree(value: Any, sort_keys: bool) -> Any:
     if isinstance(value, MutableMapping):
         if sort_keys:
             # Sort the data https://stackoverflow.com/a/40227545
-            sorted_value = CommentedMap()
+            # TODO: This needs to be CommentedMap if using roundtrip, but it is incompatible with
+            #  safe yaml mode.
+            sorted_value = {}
```
psss, line 3599: What about passing `yaml_type` to `_sanitize_yaml_tree` and use the appropriate type here?

psss (commented):
Looks good, just one question and one thought: There are a couple more places where `tmt` is storing modified `fmf` files to disk:

* `tmt link`
* `tmt test id`
* `tmt test export --nitrate`
* `tmt test export --polarion`

But seems all those cases are actually using the `fmf` context manager approach:

```python
with tmt_object.node as data: 
    data["link"] = [link_data]
```

Would be good to fix on the `fmf` side as well. Filed here:

* teemtee/fmf/issues/292

tmt/utils/__init__.py
LecrisUT: Yes, but that requires a bit more refactoring together with refactoring ideas in https://github.com/teemtee/tmt/pull/4486

psss: Ok.

psss (approved):

vaibhavdaren (approved):

tests/plan/export/test.sh
thrix: thank you

pylock.toml
thrix: yeah, it looks weird, but I guess that is how ti is, it is a generated file anyway.

thrix (approved):
LGTM

Issue #4557, opened 2026-02-05
Review flaky tests and make them reliable


psss:
Some tests from time to time fail. Let's gather them here, review and improve.

psss:
`/tests/try/basic` today [failed](https://artifacts.dev.testing-farm.io/b1a8ab04-7b8e-4619-bd0c-326c2a397bc0/work-basicbkpo55dz/plans/features/basic/execute/data/guest/default-0/tests/try/basic-83/output.txt) with:

```
What do we do next?

    test       rediscover tests and execute them again
    login      log into the guest for experimenting
    host       run command on the host
    lcd        change directory on the local host, discover tests there
    verbose    set the desired level of verbosity.
    debug      choose a different debugging level

    discover   gather information about tests to be executed
    prepare    prepare the environment for testing
    execute    run tests using the specified executor
    report     provide test results overview and send reports
    finish     perform the user defined finishing tasks
    cleanup    clean up guests and prune the workdir

    keep       exit the session but keep the run for later use
    quit       clean up the run and quit the session

b
1
> 
Which debug level would you like?
Choose 0-3, hit Enter to keep 0> Switched to debug level 1.

/plans/multi/plan/one
:: [ 13:27:48 ] :: [   PASS   ] :: Try with local (Expected 0, got 0)
:: [ 13:27:48 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.Jx4WqIgh' should contain 'Switched to debug level 1' 
:: [ 13:27:48 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.Jx4WqIgh' should not contain 'Choose 0-3, hit Enter to keep 1' 
:: [ 13:27:48 ] :: [  BEGIN   ] :: Running 'grep -A 100 'What do we do next?' /var/tmp/rlRun_LOG.Jx4WqIgh | grep /plans/multi/plan/one'
/plans/multi/plan/one
:: [ 13:27:48 ] :: [   PASS   ] :: Command 'grep -A 100 'What do we do next?' /var/tmp/rlRun_LOG.Jx4WqIgh | grep /plans/multi/plan/one' (Expected 0, got 0)
:: [ 13:27:48 ] :: [  BEGIN   ] :: Running 'grep -A 100 'What do we do next?' /var/tmp/rlRun_LOG.Jx4WqIgh | grep /plans/multi/plan/two'
[...]

PR #4717, opened 2026-03-18
Fix flaky test, add retry to testcloud

bajertom:
Pull Request Checklist

* [x] implement the feature

Fixes https://github.com/teemtee/tmt/issues/4557 

First test mentioned above sometimes fail due to the timeout being too fast and cuts off right before `/plans/multi/plan/one` gets through, so I've added another expect phrase that comes after that line and increased timeout to 3 seconds.

Second test imho failed due to messed up download of the qcow2 image.
```
        progress: downloading...
Unknown download size.
        effective hardware: {}
```
The unindented  "Unknown download size."  error seems odd. I've checked successful runs of that test in other PRs and "Unknown download size" does not appear there. Another line says: 
```
qemu-img: /tmp/tmp.9726GhCN02/testcloud/instances/tmt-001-RuKJKCkN/tmt-001-RuKJKCkN-local.qcow2: Image is not in qcow2 format
```
So I think that something did get downloaded but the file was corrupted.

I wrapped the responsible code part into a function `prepare_image` and added a validation of the qcow2. If that fails, the downloaded file will be removed and `prepare_image` will be retried. 

Hard to fix flaky test when they're flaky and can't be reproduced at will 😐

Co-generated by Claude Code.

### DIFF
--- a/tmt/steps/provision/testcloud.py
+++ b/tmt/steps/provision/testcloud.py
@@ -1162,7 +1162,7 @@ class GuestTestcloud(tmt.GuestSsh):
             # Validate the base image is a proper qcow2 file
             image_path = Path(self._image.local_path)
             try:
-                Command("qemu-img", "info", str(image_path)).run(
+                Command("qemu-img", "check", str(image_path)).run(
                     cwd=Path.cwd(),
                     logger=self._logger,
                 )


tmt/steps/provision/testcloud.py
```diff
@@ -1113,22 +1118,56 @@ def start(self) -> None:
...
+                Command("qemu-img", "info", str(image_path)).run(
+                    cwd=Path.cwd(),
+                    logger=self._logger,
+                )
+            except tmt.utils.RunError as error:
+                if image_path.exists():
+                    self.debug(f"Removing corrupt image '{image_path}'.")
+                    image_path.unlink()
+                raise ProvisionError(
+                    f"Image '{image_path}' is not a valid qcow2 image.") from error
```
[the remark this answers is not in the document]
bajertom: Your suggestion is exactly the same as what is in the code.

tmt/steps/provision/testcloud.py
```diff
@@ -218,6 +219,10 @@ def import_testcloud(logger: tmt.log.Logger) -> None:
 IMAGE_URL_FETCH_RETRY_ATTEMPTS = 5
 IMAGE_URL_FETCH_RETRY_INTERVAL = 5
 
+#: Image prepare retry attempts and interval
+IMAGE_PREPARE_RETRY_ATTEMPTS = 3
+IMAGE_PREPARE_RETRY_INTERVAL = 5
```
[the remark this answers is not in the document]
bajertom: You say "be concise and avoid conversational filler" and yet your suggestion is way longer that my comment which is.. concise.

tmt/steps/provision/testcloud.py
```diff
@@ -1113,22 +1118,54 @@ def start(self) -> None:
...
+                self._image.prepare()
+            except FileNotFoundError as error:
+                raise ProvisionError(f"Image '{self._image.local_path}' not found.") from error
+            except (testcloud.exceptions.TestcloudPermissionsError, PermissionError) as error:
+                raise ProvisionError(
+                    f"Failed to prepare the image. Check the "
+                    f"'{self.testcloud_image_dirpath}' directory permissions."
+                ) from error
+            except KeyError as error:
+                raise ProvisionError(f"Failed to prepare image '{self.image_url}'.") from error
```
tcornell-bus, line 1160: If any of these exceptions occur, the function will get retried by rety(). Do they have a chance of succeeding if retried? If not, only 15 seconds are wasted, so it doesn't seem like a huge deal. Just wondering if it was intended that these would also cause retries.

bajertom: They're unlikely to succeed on a retry, but retry has to cover for both the download and validation of the image together. I don't know how to separate those exceptions tied to the `try: self._image.prepare()` from the exception tied to `try: validation command` in a nice way. I can only think of adjusting the `tmt.utils.retry()` itself to give it ability to retry only on a user-defined exceptions and do nothing on other exceptions (or vice versa). (Which would deserve its own PR)

tmt/steps/provision/testcloud.py
```diff
@@ -1113,22 +1118,54 @@ def start(self) -> None:
...
+
+            # Validate the base image is a proper qcow2 file
+            image_path = Path(self._image.local_path)
+            try:
+                Command("qemu-img", "info", str(image_path)).run(
+                    cwd=Path.cwd(),
+                    logger=self._logger,
+                )
```
psss, line 1168: Hm, just tried some dummy test and `qemu-img info` seems to be happy:

```
> echo hi > Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2

> qemu-img info Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2
image: Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2
file format: raw
virtual size: 512 B (512 bytes)
disk size: 4 KiB
Child node '/file':
    filename: Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2
    protocol type: file
    file length: 512 B (512 bytes)
    disk size: 4 KiB

> echo $?
0
```

Perhaps we need some more thorough check for this?

bajertom: Changed `info` subcommand to `check` subcommand, which exits with 63 on a non-qcow2 file.

psss: Great, now it works nicely!

    provision
        how: virtual
        order: 50
        ansible: GuestAnsible(group=None, vars={})
        image: fedora-42
        fail: Image '/var/tmp/tmt/testcloud/images/Fedora-Cloud-Base-Generic-42-1.1.x86_64.qcow2' is not a valid qcow2 image.
        progress: downloading...

PR #4719, opened 2026-03-19
Improve test coverage for `prepare` plugins in image mode

### LINKED ISSUE #4606: Improve test coverage for `prepare` plugins in image mode
In order to prevent breakages like #4598 it would be good to extend the test coverage to exercise relevant `prepare` plugins under image mode as well.

thrix:
Incorporate the image mode testing into the plugins' tests and remove the separate `bootc` specific testing.
Running the plugins test exercises all the cases as in package mode and provides good confidence about the
image mode testing support.

The testing is done against F44 and CS10 image mode cloud base images provided by Testing Farm.

Run the testing in a separate `/provision/virtual-image-mode` plan, because the testing will similar
time to run as `/provision/virtual`.

Fixes teemtee/tmt#4606

Pull Request Checklist

* [x] implement the feature

Assisted-by: Claude Code


plans/provision/virtual.fmf
```diff
@@ -85,3 +85,25 @@ adjust+:
 /the-rest:
     discover+:
         test: "^/tests/(?!provision|prepare|execute)"
+
+/image-mode:
+    summary: Image mode guests via testcloud
+    description: |
+        Verify functionality of prepare plugins on image mode guests
+        provisioned via the virtual (testcloud) plugin with bootc
+        qcow2 images.
```
thrix:
```suggestion
        Verify functionality of plugins on image mode guests
        provisioned via the virtual (testcloud) plugin with Testing Farm
        provided qcow2 images.
```

plans/provision/virtual.fmf
```diff
@@ -85,3 +85,25 @@ adjust+:
...
+
+/image-mode:
+    summary: Image mode guests via testcloud
+    description: |
+        Verify functionality of plugins on image mode guests
+        provisioned via the virtual (testcloud) plugin with Testing Farm
+        provided qcow2 images.
+
+    discover+:
+        filter: 'tag:provision-virtual-image-mode'
```
LecrisUT: Would prefer if it compounded the tag filter like `tag:provision-virtual & tag:image-mode`

plans/provision/virtual.fmf
```diff
@@ -85,3 +85,25 @@ adjust+:
...
+
+    environment+:
+        IS_IMAGE_MODE: "yes"
```
LecrisUT:
```suggestion
        IMAGE_MODE: "yes"
```
Name nitpick. `IS_IMAGE_MODE` would indicate that something is image mode, which 🤷 on what that something it could refer to. This would rather indicate that we are testing image mode.

plans/provision/virtual.fmf
```diff
@@ -85,3 +85,25 @@ adjust+:
...
+
+    /prepare:
+        discover+:
+            test: "^/tests/prepare"
+            exclude: "^/tests/prepare/install"
+
+    /install:
+        discover+:
+            test: "^/tests/prepare/install"
```
LecrisUT: I don't like this duplication. Should be able to avoid it if the top-level `image-mode` is changed into an `adjust`

tests/prepare/install/test.sh
```diff
@@ -6,18 +6,25 @@ function fetch_downloaded_packages () {
...
-                                                    && mv /tmp/diffutils*.rpm /tmp/diffutils.rpm\""
-        rlRun "podman cp $container_id:/tmp/tree.rpm $package_cache/"
-        rlRun "podman cp $container_id:/tmp/diffutils.rpm $package_cache/"
-        rlRun "podman kill $container_id"
-        rlRun "podman rm $container_id"
+        if [ "$IS_IMAGE_MODE" = "yes" ]; then
+            # For image mode, download RPMs directly on the runner
+            rlRun "dnf download --destdir $package_cache tree diffutils"
+            rlRun "mv $package_cache/tree-*.rpm $package_cache/tree.rpm"
+            rlRun "mv $package_cache/diffutils-*.rpm $package_cache/diffutils.rpm"
```
LecrisUT: This does not do the same thing as the podman counterpart does. Primarily it does not guarantee that the rpms have the same distgit tag. If you can establish a conversion back-and-forth with the container image names, then you can just reuse the podman approach.

.packit.yaml
```diff
@@ -174,7 +174,11 @@ jobs:
 
   - <<: *provision
     identifier: provision-virtual
-    tmt_plan: /plans/provision/virtual
+    tmt_plan: '/plans/provision/virtual/(?!image-mode)'
+
+  - <<: *provision
+    identifier: provision-virtual-image-mode
+    tmt_plan: /plans/provision/virtual/image-mode
```
LecrisUT: Pending my other comment in `plans/provision/virtual.fmf`, this can instead be changed to a context adjustment

tmt/package_managers/bootc.py
```diff
@@ -381,6 +388,55 @@ def install_local(
...
+            )
+
+    def create_repository(self, directory: Path) -> CommandOutput:
+        """
+        Create repository metadata by delegating to the underlying engine.
+
+        Runs on the live system, writing to the workdir which is writable
+        in image mode.
+        """
+        return self.guest.execute(self.engine.aux_engine.create_repository(directory))
```
LecrisUT: Please separate these changes

tests/prepare/shell/test.sh
```diff
@@ -4,28 +4,51 @@
 rlJournalStart
     rlPhaseStartSetup
         rlRun "PROVISION_HOW=${PROVISION_HOW:-local}"
+        rlRun "IS_IMAGE_MODE=${IS_IMAGE_MODE:-no}"
+        rlRun "PROVISION_OPTS='--how=$PROVISION_HOW'"
+        if [ "$IS_IMAGE_MODE" = "yes" ]; then
+            . ../../images.sh || exit 1
+            rlRun "IMAGE_MODE_IMAGE=$(echo "$TEST_IMAGE_MODE_IMAGES" | head -1)"
```
LecrisUT: In the places where we consume `TEST_VIRTUAL_IMAGES` we do not consume only the top item. Let's make this equivalent with that one so that it is simpler to patch these instances when we convert to pytest.

tests/prepare/recommend/test.sh
LecrisUT, line 46: There are so many ifs in this file. It would be high time to consolidate them. Something like `/prepare/install` approach would be better

tests/prepare/install/test.sh
```diff
@@ -131,7 +151,7 @@ rlJournalStart
                 rlFail "Cannot infer distro for image $image"
             fi
 
-            tmt_run="tmt -vvv -c distro=$distro run --id $run --scratch"
+            tmt_run="tmt -vvv -c distro=$distro -c package_manager=$package_manager run --id $run --scratch"
```
LecrisUT: Can you elaborate on the why? I do not see where this is newly used

tests/prepare/install/test.sh
```diff
@@ -144,14 +164,30 @@ rlJournalStart
...
 
-            if is_ubuntu "$image" || is_debian "$image"; then
+            if is_image_mode "$image"; then
+                # Image mode uses Containerfile build, image switch and reboot.
+                # The prepare step itself handles the reboot — if it completes
+                # successfully, packages are baked into the new image.
+                rlAssertGrep "package: building container image with dependencies" $rlRun_LOG
+                rlAssertGrep "switching to new image" $rlRun_LOG
+                rlAssertGrep "rebooting to apply new image" $rlRun_LOG
+            elif is_ubuntu "$image" || is_debian "$image"; then
```
LecrisUT: Why does it skip the rest of the `elif`?

tests/prepare/install/test.sh
```diff
@@ -269,6 +305,9 @@ rlJournalStart
             elif is_alpine "$image"; then
                 rlAssertGrep "stderr:   tree-but-spelled-wrong (no such package)" $rlRun_LOG
 
+            elif is_image_mode_fedora "$image"; then
+                rlAssertGrep "No match for argument: tree-but-spelled-wrong" $rlRun_LOG
+
```
LecrisUT: I don't see why this would not have been caught by the previous ones

tests/prepare/install/test.sh
```diff
@@ -174,7 +210,7 @@ rlJournalStart
             fi
         rlPhaseEnd
 
-        if [ "$PROVISION_HOW" = "container" ] && rlIsFedora 43 && is_fedora_43 "$image"; then
+        if ([ "$PROVISION_HOW" = "container" ] && rlIsFedora 43 && is_fedora_43 "$image") || is_image_mode "$image"; then
```
LecrisUT: Why should a special `is_image_mode` here be done?

tests/prepare/install/test.sh
```diff
@@ -364,7 +409,7 @@ rlJournalStart
 
         # TODO: at least copr is RH-specific, but package name escaping and debuginfo should be
         # possible to extend to other distros.
-        if (is_fedora "$image" && ! is_fedora_coreos "$image") || is_centos "$image" || is_ubi "$image"; then
+        if (is_fedora "$image" && ! is_fedora_coreos "$image") || is_centos "$image" || is_ubi "$image" || is_image_mode "$image"; then
```
LecrisUT: Here especially, why shouldn't the previous steps have been satisfied?

thrix:

tests/prepare/install/test.sh
thrix: @LecrisUT if you insist to do a proper match, I am fine with that, seems just too much complication for not much benefit to me, if this works

thrix: But I am happy to be wrong, as usual :yawning_face:

tmt/package_managers/bootc.py
thrix: @thrix-bot ok, but then why you kept it here! remove it from this change and mention in #4748 that what was the motivation of the change (mention this comment?)

tests/prepare/install/test.sh
thrix: @thrix-bot maybe we could lower the cases in is_fedora and call it a day?? then it would not be needed?

tests/prepare/install/test.sh
thrix: @thrix-bot again, maybe then we could drop the `stderr: ` checking and as mentioned in other comment, we could then get rif of `is_image_mode_fedora` maybe?

tests/prepare/install/test.sh
thrix: @thrix-bot I think he meant why do we not assert the 

```
 rlAssertGrep "summary: 2 preparations applied" $rlRun_LOG
```

can you recheck?

tmt/package_managers/bootc.py
thrix: it is now :D

tests/prepare/install/test.sh
thrix: @thrix-bot you should see the output here: https://artifacts.dev.testing-farm.io/ed9ce5fd-52bc-4417-8508-86b3b846cf78/ (once it finishes)

tests/prepare/install/test.sh
thrix: @thrix-bot isn't the difference only the `stder: ` prefix? You can check here once the testing is done:

https://artifacts.dev.testing-farm.io/ed9ce5fd-52bc-4417-8508-86b3b846cf78/

.packit.yaml
```diff
@@ -176,6 +176,17 @@ jobs:
...
+  - <<: *provision
+    identifier: provision-virtual-image-mode
+    tmt_plan: /plans/provision/virtual
+    tf_extra_params:
+      environments:
+        - tmt:
+            context:
+              how: provision
+              image_mode: "yes"
+          <<: *tmt-cloud-resources
```
LecrisUT: Define this in the `_` and use inheritance

tests/prepare/install/test.sh
LecrisUT: > For image mode there's no container to exec into (the images are qcow2, not OCI)

Huh? what is this thing on about. The whole point of bootc is to be able to have a containerfile and replicate the environment in that container before finalizing it.

> @LecrisUT if you insist to do a proper match, I am fine with that, seems just too much complication for not much benefit to me, if this works

Yes I insist

tests/prepare/recommend/test.sh
LecrisUT: Bots do not have memory. I refuse to have the comment resolved without the human doing something about the "would be a nice follow-up".

The bot commenting on this is meaningless and deceiving in this case, teach it to not do it.

tests/prepare/install/test.sh
LecrisUT: What's the point in commenting on an issue of its own making ....... Do something about the unnecessary verbosity of this thing.

tests/prepare/install/test.sh
LecrisUT: Do something?

tests/prepare/install/test.sh
LecrisUT: ....

tests/prepare/install/test.sh
LecrisUT: For a "Largue language model" this thing has no grasp of English or human language. @thrix figure this thing out.

LecrisUT:
Change the PR description. It is not in a good form to be useful as a commit description.

tests/images.sh
```diff
@@ -51,6 +51,12 @@ fedora-coreos}"
 # combinations, just make sure the basic functionality works.
 TEST_VIRTUAL_IMAGES_SECONDARY="${TEST_VIRTUAL_IMAGES_SECONDARY:-fedora-42}"
 
+
+# Image mode (bootc) images for virtual provision.
+# These are qcow2 images with bootc/image mode enabled.
+TEST_IMAGE_MODE_IMAGES="${TEST_IMAGE_MODE_IMAGES:-https://artifacts.dev.testing-farm.io/images/CentOS-Stream-10-image-mode-x86_64.qcow2
+https://artifacts.dev.testing-farm.io/images/Fedora-44-image-mode-x86_64.qcow2}"
```
thrix: Maybe move the common URL to a variable for better readability

.packit.yaml
thrix: ok, but was not expecting to reuse this anywhere, so seemed like an overkill.

tests/prepare/install/test.sh
thrix: Well, I am not sure where the confusion is, but the below tests run only against a container provisioner now. I want to run them on image mode, because they handling is special there and it is worth testing. Also, one of the cases tested should be testing the case we fixed in https://github.com/teemtee/tmt/pull/4598/files.

tests/prepare/install/test.sh
thrix: It is true it might break very easily, so it is brittle, F44 rpms can be installed on CS10, but yeah, adding CS9 it would burn right there on glibc :D

Adding an associative array to do the mapping from image mode qcow2 url to a container image.

thrix: It is there.

tests/prepare/install/test.sh
thrix: It is trying to live :)

thrix: I will omit comenting if other reviewers involved.

tests/prepare/install/test.sh
thrix:
```
 rlAssertGrep "summary: 2 preparations applied" $rlRun_LOG
```

Should had been asserted. Moved the image mode specific assertions under a dedicated image mode only if condition.

Kept the others as is.

tests/prepare/install/test.sh
thrix: the condition is needed, as the failed output looks like in with `is_ostree` ...

```
            stderr: Copying blob sha256:d34da7e1e73b79af325c61855b24d99131eca28e5fd17ed208173a193876d489
            stderr: Copying blob sha256:12787d84fa137cd5649a9005efe98ec9d05ea46245fdc50aecb7dd007f2035b1
            stderr: Copying blob sha256:140e7dafa8cd03e921c2c6e87226aadf269b29be44dd2e7cea5884324b324ca6
            stderr: Copying blob sha256:a8fed5e89afb90df86445c2091c5931da38a26b1ce9fce030a6ab60daaed96fb
            stderr: Copying config sha256:037355d2399ca61afefbc5c47f2450849b4ffb1dd49be40f66f7c1563c7bd52d
            stderr: Writing manifest to image destination
            stdout: STEP 2/2: RUN rpm -q --whatprovides tree-but-spelled-wrong || dnf5 install -y  tree-but-spelled-wrong 
            stdout: no package provides tree-but-spelled-wrong
            stderr: Updating and loading repositories:
            stderr: Repositories loaded.
            stderr: Failed to resolve the transaction:
            stderr: No match for argument: tree-but-spelled-wrong
            stderr: You can try to add to command line:
            stderr:   --skip-unavailable to skip unavailable packages
            stderr: Error: building at STEP "RUN rpm -q --whatprovides tree-but-spelled-wrong || dnf5 install -y  tree-but-spelled-wrong": while running runtime: exit status 1
```

Moved next to it add `|| is_image_mode ...`

.packit.yaml
thrix: Moved there, pls recheck

tests/images.sh
thrix: Done

thrix:
> Change the PR description. It is not in a good form to be useful as a commit description.

@LecrisUT description was updated, no AI involved

tests/prepare/install/test.sh
thrix: Resolving, `is_fedora` now matches also image mode composes, so `is_image_mode was removed from the condition

tests/prepare/recommend/test.sh
```diff
@@ -50,8 +56,21 @@ rlJournalStart
...
+    if [[ "$IMAGE_MODE" == "yes" ]]; then
+        while IFS= read -r image; do
+            rlPhaseStartTest "Test $image (image-mode)"
+                rlRun -s "$tmt --image $image $basic"
+                rlAssertGrep "Recommended packages failed to install, continuing regardless:" $rlRun_LOG
+                rlAssertGrep "forest: recommended by: /test/mixed" $rlRun_LOG
+                rlAssertGrep "weird-package: recommended by: /test/weird" $rlRun_LOG
+                rlAssertNotGrep "dconf: recommended by: /test/mixed" $rlRun_LOG
+            rlPhaseEnd
+        done <<< "$TEST_IMAGE_MODE_IMAGES"
```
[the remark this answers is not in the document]
thrix: yup, kept the newlines

tests/prepare/shell/test.sh
```diff
@@ -4,28 +4,61 @@
...
+                rlRun "tmt run -arv provision $provision_opts plan -n custom \
+                    prepare -h shell -s './prepare.sh'" 0 "Prepare using a custom script from cmdline"
+            rlPhaseEnd
+
+            # TODO: On image mode, /tmp is a tmpfs cleared on reboot. The inner
+            # test plan verifies files in /tmp which don't persist. Similarly,
+            # the remote script URL plan's $TMT_PREPARE_SHELL_URL_REPOSITORY
+            # env var is empty during the Containerfile RUN. Both need image
+            # mode support in tmt core before these tests can pass.
+        done <<< "$IMAGES"
```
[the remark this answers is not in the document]
thrix: repeating yourself are you :yo_yo:

tests/prepare/install/test.sh
```diff
@@ -364,7 +404,8 @@ rlJournalStart
 
         # TODO: at least copr is RH-specific, but package name escaping and debuginfo should be
         # possible to extend to other distros.
-        if (is_fedora "$image" && ! is_fedora_coreos "$image") || is_centos "$image" || is_ubi "$image"; then
+        # TODO: image mode copr support depends on #4748
+        if ! is_image_mode "$image" && ((is_fedora "$image" && ! is_fedora_coreos "$image") || is_centos "$image" || is_ubi "$image"); then
```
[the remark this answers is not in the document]
thrix: well, you might be correct, but the "fix" is wrong

tests/prepare/install/test.sh
thrix: But not everywhere 😭  so a dedicated condition might be needed

tests/prepare/install/test.sh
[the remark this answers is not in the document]
thrix: I was testing this manually and `((` does not seems to be an issue.

I tested various combos if

```
❯ bash -c "true && (( true && false ) && false)"; echo $?
1

❯ bash -c "true || (( true && false ) && false)"; echo $?
0

❯ bash -c "true || (( true && false ) || false)"; echo $?
0

❯ bash -c "true && (( true && false ) || false)"; echo $?
1

❯ bash -c "true && (( true && false ) || true)"; echo $?
0
```

No sign of trouble.

tests/prepare/install/test.sh
LecrisUT: > Adding an associative array to do the mapping from image mode qcow2 url to a container image.

Why? Why can't we build the bootc images directly from `quay.io/fedora/fedora-bootc:...` and use the same naming structure that we have for the other parts?

tests/prepare/install/test.sh
LecrisUT:
```bash
            if is_centos_7 "$image"; then
                rlAssertGrep "stdout: no package provides tree-but-spelled-wrong" $rlRun_LOG

            elif is_ostree "$image"; then
                if [ "$PROVISION_HOW" = "virtual" ]; then
                    rlAssertGrep "stderr: No match for argument: tree-but-spelled-wrong" $rlRun_LOG
                else
                    rlAssertGrep "stderr: error: Packages not found: tree-but-spelled-wrong" $rlRun_LOG
                fi

            elif is_fedora_coreos "$image"; then
                rlAssertGrep "stderr: No match for argument: tree-but-spelled-wrong" $rlRun_LOG

            elif is_fedora_rawhide "$image"; then
                rlAssertGrep "stderr: No match for argument: tree-but-spelled-wrong" $rlRun_LOG

            elif is_fedora_eln "$image"; then
                rlAssertGrep "stderr: No match for argument: tree-but-spelled-wrong" $rlRun_LOG

            elif is_fedora_43 "$image"; then
                rlAssertGrep "stderr: No match for argument: tree-but-spelled-wrong" $rlRun_LOG
```
It is not matching `is_ostree`, all of them are the same and can be substituted with a plain `is_fedora`. Similarly would apply for image_mode for centos

tests/prepare/install/test.sh
LecrisUT: Then put it next to the `PROVSION_HOW` check, not globally `or`-ed

tests/prepare/install/test.sh
```diff
@@ -84,7 +94,11 @@ rlJournalStart
 
             elif is_centos_stream_10 "$image"; then
                 rlRun "distro=centos-stream-10"
-                rlRun "package_manager=dnf"
+                if is_image_mode "$image"; then
+                    rlRun "package_manager=bootc"
```
LecrisUT: Move this outside since it is common for all of the `is_image_mode`

tests/prepare/install/test.sh
```diff
@@ -3,17 +3,22 @@
...
         # For some reason, this command will get stuck in rlRun...
-        container_id="$(podman run -d $1 sleep 3600)"
+        container_id="$(podman run -d $image sleep 3600)"
 
         rlRun "podman exec $container_id bash -c \"set -x; \
                                                     dnf install -y 'dnf-command(download)' \
                                                     && dnf download --destdir /tmp tree diffutils \
                                                     && mv /tmp/tree*.rpm /tmp/tree.rpm \
                                                     && mv /tmp/diffutils*.rpm /tmp/diffutils.rpm\""
+
```
LecrisUT: Unrelated change

tests/prepare/recommend/test.sh
```diff
@@ -5,25 +5,31 @@
...
+        if [ "$IMAGE_MODE" != "yes" ]; then
+            build_container_image "ubi/8/upstream\:latest"
+            build_container_image "centos/7/upstream\:latest"
+            build_container_image "fedora/latest/upstream\:latest"
+        fi
 
-        build_container_image "ubi/8/upstream\:latest"
-        build_container_image "centos/7/upstream\:latest"
-        build_container_image "fedora/latest/upstream\:latest"
+        rlRun "pushd data"
```
LecrisUT, line 16: Existing issue even with `PROVISION_HOW=virtual`

tests/prepare/recommend/test.sh
```diff
@@ -5,25 +5,31 @@
...
 
-    # Verify against the default provision image
-    rlPhaseStartTest "Test the default image ($PROVISION_HOW)"
-        rlRun -s "$tmt $basic"
-	rlAssertGrep "Recommended packages failed to install, continuing regardless:" $rlRun_LOG
-	rlAssertGrep "forest: recommended by: /test/mixed" $rlRun_LOG
-	rlAssertGrep "weird-package: recommended by: /test/weird" $rlRun_LOG
-	rlAssertNotGrep "dconf: recommended by: /test/mixed" $rlRun_LOG
-    rlPhaseEnd
+    # Verify against the default provision image (skip for image mode, no default image)
```
LecrisUT, line 23: Don't see why this should not be available in the future. Add a todo note if the intention will be to be supported in the future with a `provision: bootc`-like wrapper

tests/prepare/shell/test.sh
```diff
@@ -4,28 +4,61 @@
...
+            rlPhaseStartTest "[$image] Commandline Script"
+                rlRun "tmt run -arv provision $provision_opts plan -n custom \
+                    prepare -h shell -s './prepare.sh'" 0 "Prepare using a custom script from cmdline"
+            rlPhaseEnd
+
+            # TODO: On image mode, /tmp is a tmpfs cleared on reboot. The inner
+            # test plan verifies files in /tmp which don't persist. Similarly,
+            # the remote script URL plan's $TMT_PREPARE_SHELL_URL_REPOSITORY
+            # env var is empty during the Containerfile RUN. Both need image
+            # mode support in tmt core before these tests can pass.
```
LecrisUT: I don't see how that is unique to image mode

tests/prepare/shell/test.sh
```diff
@@ -4,28 +4,61 @@
...
+        fi
         rlRun "pushd data"
     rlPhaseEnd
 
-    rlPhaseStartTest "Custom Script"
-        rlRun "tmt run -arv provision --how=$PROVISION_HOW plan -n custom" 0 "Prepare using a custom script"
-    rlPhaseEnd
+    if [ "$IMAGE_MODE" = "yes" ]; then
+        while IFS= read -r image; do
+            provision_opts="--how=virtual --image=$image"
```
LecrisUT: The ordering of the `if` here makes the diff unnecessarily complicated. Invert the order

tests/prepare/shell/test.sh
```diff
@@ -4,28 +4,61 @@
...
+            # TODO: On image mode, /tmp is a tmpfs cleared on reboot. The inner
+            # test plan verifies files in /tmp which don't persist. Similarly,
+            # the remote script URL plan's $TMT_PREPARE_SHELL_URL_REPOSITORY
+            # env var is empty during the Containerfile RUN. Both need image
+            # mode support in tmt core before these tests can pass.
+        done <<< "$IMAGES"
+    else
+        rlPhaseStartTest "Custom Script"
+            rlRun -s "tmt run -arv provision --how=$PROVISION_HOW plan -n custom" 0 "Prepare using a custom script"
+        rlPhaseEnd
```
LecrisUT, line 30: I don't see why these cases cannot be merged

tests/prepare/shell/test.sh
```diff
@@ -4,28 +4,61 @@
...
+        rlPhaseStartTest "Multiple Commandline Scripts"
+            rlRun "tmt run -arv provision --how=$PROVISION_HOW plans -n multiple \
+                prepare -h shell -s 'touch /tmp/first' -s 'touch /tmp/second'"
+        rlPhaseEnd
+
+        rlPhaseStartTest "Remote Script"
+            rlRun -s "tmt -vvv run provision --how=$PROVISION_HOW prepare finish cleanup plan -n url" 0 "Prepare using a remote script"
+            rlAssertGrep "Hello world" "$rlRun_LOG" #check for the prepare script
+            rlAssertGrep "third" "$rlRun_LOG" # check for the finish script
+        rlPhaseEnd
```
LecrisUT: Why is this case not covered?

.packit.yaml
```diff
@@ -176,6 +186,11 @@ jobs:
     identifier: provision-virtual
     tmt_plan: /plans/provision/virtual
 
+  - <<: *provision
+    identifier: provision-virtual-image-mode
+    tmt_plan: /plans/provision/virtual
+    <<: *image-mode
```
LecrisUT, line 190:
```suggestion
  - <<: *provision
    <<: *image-mode
    identifier: provision-virtual-image-mode
    tmt_plan: /plans/provision/virtual
```

.packit.yaml
```diff
@@ -127,6 +127,16 @@ _:
...
           <<: *tmt-cloud-resources
 
+  # Image mode
+  - &image-mode
+    tf_extra_params:
+      environments:
+        - tmt:
+            context:
+              how: provision
+              image_mode: "yes"
```
LecrisUT, line 137: This is inheriting from the provision context, indicate so with an anchor

tests/prepare/install/test.sh
thrix: @LecrisUT Not sure I follow, we need to use QCOW2 images. Not the container images for the `virtual` provisioner. I used package mode containers as they are smaller then the fedora-bootc ones.

tests/prepare/install/test.sh
thrix: it would not match `rlIsFedora 43 && is_fedora_43 "$image"` ...

tests/prepare/install/test.sh
LecrisUT: But that is not a typical setup that the user would use. Please add a comment on why that image is being used. I would like it to instead build the image locally here in tmt the same with how we do the normal containers, but can deffer it if necessary

tests/prepare/install/test.sh
LecrisUT: > it would not match `rlIsFedora 43 && is_fedora_43 "$image"` ...

Yes, I'm aware, and that's a separate issue to be resolved for the exact same reasons.

psss:
The new [provision-virtual-image-mode job](https://artifacts.dev.testing-farm.io/9d7baf08-3526-49c8-8c50-ac28a160fe83/) contains four empty plans. Let's disable them if they are not relevant?

LecrisUT:
Btw, what do we do with the diverging naming conventions `bootc` vs `image-mode`. Shouldn't this be `provision-virtual-bootc`?

tests/prepare/install/test.sh
thrix: Removed

tests/prepare/recommend/test.sh
thrix: Filled under #4786

tests/prepare/install/test.sh
thrix: Filed under #4787, refusing to do too many separate refactoring asks

thrix:
> Btw, what do we do with the diverging naming conventions `bootc` vs `image-mode`. Shouldn't this be `provision-virtual-bootc`?

I would opt calling this `image-mode` to mitigate the confusion with the `bootc` provision plugin. It already caused that confusion at the beginning of this review.

LecrisUT:
> > Btw, what do we do with the diverging naming conventions `bootc` vs `image-mode`. Shouldn't this be `provision-virtual-bootc`?
> 
> I would opt calling this `image-mode` to mitigate the confusion with the `bootc` provision plugin. It already caused that confusion at the beginning of this review.

Did it create a confusion? Isn't the whole point of `virtual-(bootc|image-mode)` to disambiguate it. To me it would feel more confusing down the line as if we are distinguishing between bootc and image-mode. Another option is `virtual-with-bootc`

tests/prepare/install/test.sh
thrix: I am lost what do you want here, please provide a snippet. We want to run this:

1. one of the container images (fedora 43 now)
2. on image mode

tests/prepare/install/test.sh
thrix: was moved outside

tests/prepare/recommend/test.sh
thrix: This was refactored to build the container images only for `PROVISION = container`

tests/prepare/recommend/test.sh
thrix: Not sure how it would be supported, we use the same provisioning plugin `virtual` and it can have only a single default image. Not sure what would be the `provision: bootc`-like wrapper, please expand.

tests/prepare/shell/test.sh
thrix: This was removed, resolving

tests/prepare/shell/test.sh
thrix: This was refactored. Resolving.

tests/prepare/shell/test.sh
thrix: This test was refactored, please reecheck. 

Also the comment is on a single test, so maybe your comment just needs clarification.

tests/prepare/shell/test.sh
thrix: Filed #4785, this is broken in image mode.

.packit.yaml
thrix: Does not help much, but ok:

```
.packit.yaml --- 1/2 --- YAML
122       - fedora-latest-stable             122       - fedora-latest-stable
123     tf_extra_params:                     123     tf_extra_params:
124       environments:                      124       environments:
125         - tmt:                           125           tmt:
126             context:                     126             context: &provision-context
127               how: provision             127               how: provision
128           <<: *tmt-cloud-resources       128           <<: *tmt-cloud-resources
129                                          129 

.packit.yaml --- 2/2 --- YAML
133       environments:                      133       environments:
134         - tmt:                           134         - tmt:
135             context:                     135             context: 
136               how: provision             136               <<: *provision-context
137               image_mode: "yes"          137               image_mode: "yes"
138           <<: *tmt-cloud-resources       138           <<: *tmt-cloud-resources
139                                          139 
```

tests/prepare/install/test.sh
thrix: added a ocomment pls recheck

tests/prepare/install/test.sh
LecrisUT: - Have the test run solely on `rlIsFedora 43 && is_fedora_43 "$image"` check (or a better equivalent one that we should add)
  - The `rlIsFedora` in here does not seem right, it would make more sense to gate it on the packit/plan side, in line with https://github.com/teemtee/tmt/pull/4400
  - The time-saved is miniscule when you compare https://artifacts.dev.testing-farm.io/c209da29-bb97-43e7-9636-bbe55a46db04/ with https://artifacts.dev.testing-farm.io/e5345cba-e5d8-4f95-be6c-b7cb4639b2a1/
- `[[ "$PROVISION_HOW" = "container" || is_image_mode "$image" ]]`
- The comments indicate that both branches are intentional and should be kept as-is in a future refactoring

Overall the intent in this section is unclear and the change looks more like slapping a duct-tape on top of a pile of duct-tapes. If we come back to re-organizing this test, how should this be split, what is meaningful to run, etc., it would make our future selves jobs easier.

tests/prepare/recommend/test.sh
LecrisUT: Just that the user might want a simpler interface like
```yaml
provision:
  how: virtual
  image_mode: true
```
and have a translation to pick up the image mode variant of whatever short-name image they provide `fedora-rawhide` for example.

thrix: Adding comment in next update referencing newly created issue #4806

tests/prepare/shell/test.sh
```diff
@@ -1,31 +1,59 @@
...
 rlJournalStart
     rlPhaseStartSetup
         rlRun "PROVISION_HOW=${PROVISION_HOW:-local}"
+        rlRun "IMAGE_MODE=${IMAGE_MODE:-no}"
+        rlRun "FIRST=/tmp/first SECOND=/tmp/second"
+        if [ "$IMAGE_MODE" = "yes" ]; then
+            rlRun "IMAGES='$TEST_IMAGE_MODE_IMAGES'"
+            # In image mode use paths which survive a reboot
+            rlRun "FIRST=/first SECOND=/second"
+        fi
```
LecrisUT: Why does `/var/*` not work?

tests/prepare/install/test.sh
thrix: ok, changing to this

```
          # Limit these test cases to:                                                                                                                                                         
          # * container provisioner - to save resources, they do not provide additional value with the virtual provisioner                                                                     
          # * image mode - the code handling is different in this case, so we need to make sure these cases work well                                                                          
          if [ "$PROVISION_HOW" = "container" ] || is_image_mode "$image"; then                                                                                                                
              rlPhaseStartTest "$phase_prefix Install downloaded packages from current directory (plan)"                                                                                       
                  fetch_downloaded_packages "$image"                                                                                                                                           
                                                                                                                                                                                               
                  rlRun -s "$tmt plan --name /downloaded/in-cwd"                                                                                                                               
                                                                                                                                                                                               
I                 rlAssertGrep "package manager: $package_manager$" $rlRun_LOG   ● Double quote to prevent globbing and word splitting.                                                        
                                                                                                                                                                                               
I                 rlAssertGrep "summary: 2 preparations applied" $rlRun_LOG   ● Double quote to prevent globbing and word splitting.                                                           
              rlPhaseEnd   
```

tests/prepare/shell/test.sh
thrix: I guess I can unify with a comment ok

thrix: can be the same for both

thrix: Left the variables, moved just in the test block and used `/var/first` and `/var/second` for both scenarious (non-image mode and image mode)

thrix: updated, pls recheck

tests/prepare/install/test.sh
thrix: updated ^

thrix:
> > > Btw, what do we do with the diverging naming conventions `bootc` vs `image-mode`. Shouldn't this be `provision-virtual-bootc`?
> > 
> > 
> > I would opt calling this `image-mode` to mitigate the confusion with the `bootc` provision plugin. It already caused that confusion at the beginning of this review.
> 
> Did it create a confusion? Isn't the whole point of `virtual-(bootc|image-mode)` to disambiguate it. To me it would feel more confusing down the line as if we are distinguishing between bootc and image-mode. Another option is `virtual-with-bootc`

FTR, we agreed on packit job neing called `image-mode-virtual`. Otherwise keepin stuff as is.

thrix:
> > > > Btw, what do we do with the diverging naming conventions `bootc` vs `image-mode`. Shouldn't this be `provision-virtual-bootc`?
> > > 
> > > 
> > > I would opt calling this `image-mode` to mitigate the confusion with the `bootc` provision plugin. It already caused that confusion at the beginning of this review.
> > 
> > 
> > Did it create a confusion? Isn't the whole point of `virtual-(bootc|image-mode)` to disambiguate it. To me it would feel more confusing down the line as if we are distinguishing between bootc and image-mode. Another option is `virtual-with-bootc`
> 
> FTR, we agreed on packit job neing called `image-mode-virtual`. Otherwise keepin stuff as is.

edited

tests/prepare/shell/test.sh
LecrisUT: Huh?
```
            # NOTE: These paths need to persist from the image and survive a reboot
            # See https://developers.redhat.com/articles/2025/08/25/what-image-mode-3-way-merge
            rlRun "FIRST=/usr/share/first SECOND=/usr/share/second"
```
am I blind or does this conflict with your statement

> Left the variables, moved just in the test block and used `/var/first` and `/var/second` for both scenarious (non-image mode and image mode)

am I missing context? I am completely lost :worried:

thrix: @LecrisUT yes, `/var` does not work, I was wrong see:

https://developers.redhat.com/articles/2025/08/25/what-image-mode-3-way-merge

> /var → local state: Contents of the image will be ignored after the initial installation of any image

I am preparing a documentation update so we write nasty details like this one down.

LecrisUT: I don't get it. Isn't it that all `/var` is persistent so if the user changed it in a test, that survives a reboot.

From what I read in the quote provided. The issue is in the running of the scripts at build time instead of immediately? It seems no matter how we choose to run the scripts in prepare at bootc we are causing trouble, and wouldn't it be better to allow it to fail as it would be an intended failure?

thrix: That all should be documented. The shell prepare step in image mode is run in the container build. The data created in the container build in `/var` does not get applied to the live system during bootc switch. It is applied only during intial provisioning. The purpose of the test is to verify shell step works well with this current implementation. I can create a negative test, but it would be just testing the documented behaviour of bootc. 

It would be good if user would be able to select where the shell step should run, in the container build or on the live system.

thrix: Anyway, the documentation very very raw draft: https://github.com/teemtee/tmt/pull/4818

The issue selecting where shell step should run:  https://github.com/teemtee/tmt/issues/4817

happz (approved):

plans/provision/virtual.fmf
```diff
@@ -60,9 +60,18 @@ adjust+:
...
+    environment+:
+        IMAGE_MODE: "yes"
+    when: image_mode == yes
+
 /provision:
     discover+:
         test: "^/tests/provision"
+    adjust+:
+      - enabled: false
+        when: image_mode == yes
```
LecrisUT, line 74: Can squeeze in a `because` linking to the tracker? Right now it reads like it is always supposed to be disabled.

LecrisUT (approved):
Lgtm now, just a quick point if you can squeeze in

happz:
Linkcheck failed because of target site being down, merging.

Issue #4714, opened 2026-03-18
Link the Artifact Verification Plugin for discover only Providers


vaibhavdaren:
## Problem                                                                                                           
                                                                                                                       
  `ArtifactInfo.location` is currently overloaded — it holds a download URL for download providers (e.g. `koji.build`, 
  `copr.build`) and a repository name for repository providers (e.g. `repository-url`, `repository-file`,              
  `copr.repository`). This ambiguity makes it impossible to reliably associate an artifact with its source repository  
  in the verify-installation logic.                                                                                    
   
  The workaround currently in place in PR #4726 uses `provider.get_repositories()` as a proxy to extract `repo_ids`    
  directly from the `Repository` objects, bypassing `artifact.location` entirely:

Once this is done the tests for NVR priority , should work correctly with auto verification https://github.com/teemtee/tmt/pull/4726/changes/0b543fae8a9bf0da0bfde59088513652feddee5c. 

  ```python                                                                                                            
  # TODO: ``artifact.location`` is overloaded: a download URL for download
  # providers and a repo name for repository providers. We use                                                         
  # ``provider.get_repositories()`` as a proxy until ``ArtifactInfo`` gains
  # a dedicated ``repo`` field. See https://github.com/teemtee/tmt/issues/4714.
```

vaibhavdaren:
It will be good to get in https://github.com/teemtee/tmt/pull/4765 first.

vaibhavdaren:
[Discuss] Priority Resolution 

   - `tmt-artifact-shared` is created with `priority=default_repository_priority` (default: 50)                                                             
  - `repository-file` and `copr-repository` are installed as-is from the .repo file — tmt never touches the priority= field in that file                                 
  - If` dummy-1.0` is in `tmt-artifact-shared` (priority 50) and `dummy-1.1` is in the `repo-file` (say priority 99 or unset), DNF picks `tmt-artifact-shared`
  because 50 < 99.  but this is accidental, not guaranteed.                                                                                             
  - If the `repo-file` declares priority 30, DNF picks from there instead, and the verify phase fails because the package came from the wrong repo.
                                                                                                                                                       
  tmt has no way to know or control this without inspecting or rewriting the .repo file content at install time.

vaibhavdaren:
## Analysis & Implementation Plan                                                                                                                    
                                                                                                                                                       
  ### Root Cause                                                                                                                                       
                                                                                                                                                       
[...]

PR #4723, opened 2026-03-20
Simpler branching for verify-installation tests

vaibhavdaren:
Using centpkg as a common artifact works for fedora and centos both.
Related to #4646

### DIFF
--- a/tests/prepare/verify-installation/test.sh
+++ b/tests/prepare/verify-installation/test.sh
@@ -13,7 +13,7 @@ rlJournalStart
 
         rlRun "image=$TEST_IMAGE_PREFIX/$image_name"
 
-        rlRun "data_dir=$(mktemp -d)" 0 "Create temp data directory"
+        rlRun "data_dir=\$(mktemp -d)" 0 "Create temp data directory"
         rlRun "cp -r data/. $data_dir/" 0 "Copy test data"
 
         get_koji_build_id "centpkg" "$koji_tag"


tests/prepare/verify-installation/test.sh
```diff
@@ -7,26 +7,22 @@ rlJournalStart
...
-            get_koji_build_id "make" "f\${fedora_release}"
+        rlRun "data_dir=\$(mktemp -d)" 0 "Create temp data directory"
 
-            # Copy plan data and substitute the dynamic build ID
+        if rlIsFedora; then
+            get_koji_build_id "centpkg" "f\${fedora_release}"
             rlRun "cp -r data-fedora/. \$data_dir/" 0 "Copy test data"
-            rlRun "sed -i 's/KOJI_BUILD_ID/${KOJI_BUILD_ID}/g' \$data_dir/main.fmf" 0 "Substitute koji build ID"
-            rlRun "pushd \$data_dir"
         else
```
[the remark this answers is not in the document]
vaibhavdaren: Addressed in https://github.com/teemtee/tmt/pull/4723/commits/d64237d7392546a3413a85dc5b0c1d9c4eaa5de5. The code becomes more readable.

vaibhavdaren:
Created an issue while working on this : https://github.com/teemtee/tmt/issues/4742 . There is no way to setup a koji stream profile from inside tmt.

tests/prepare/artifact/lib/common.sh
```diff
@@ -7,32 +7,40 @@
...
 #
 # Usage: setup_distro_environment
 #
-# TODO: Add CentOS/RHEL support when needed
-#
 setup_distro_environment() {
-    if ! rlIsFedora; then
-        rlDie "Test requires Fedora"
+    if rlIsFedora; then
+        # TODO: Temporary hardcoded release - should be taken from function input
```
LecrisUT: We can relax this hard-coding and instead match the current distro that is currently running the `full` tests. This might require some parsing of the context

tests/prepare/artifact/lib/common.sh
```diff
@@ -7,32 +7,40 @@
...
 }
 
 # Get koji build ID from package name (fetch latest for tag)
 #
 # Usage:
-#   get_koji_build_id "make" "f43"
+#   get_koji_build_id "centpkg" "f43"
+#   echo "$KOJI_BUILD_ID"
+#
+#   get_koji_build_id "centpkg" "c10s" "stream"
```
LecrisUT: Fine with the addition of `profile`, but `centpkg` is in Fedora's koji: https://koji.fedoraproject.org/koji/buildinfo?buildID=2958084

tests/prepare/artifact/providers/copr-repository/test.sh
```diff
@@ -14,7 +14,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Test copr-repository provider with command-line override"
-        rlRun "tmt run -i $run --scratch -vv --all \
+        rlRun "tmt run -i $run --scratch -vvv --all \
```
LecrisUT: What is the purpose of the increased verbosity if you do not assertgrep anything from it?

tests/prepare/verify-installation/data/main.fmf
LecrisUT, line 1: Why are these files still diverging?

tests/prepare/verify-installation/data-fedora/main.fmf
```diff
@@ -15,22 +15,22 @@
     prepare+:
       - how: verify-installation
         verify:
-            make: tmt-artifact-shared
-            make-devel: tmt-artifact-shared
+            diffstat: tmt-artifact-shared
+            make: fedora
```
LecrisUT: Why the change to using `diffstat`?

tests/prepare/verify-installation/test.sh
```diff
@@ -7,26 +7,29 @@ rlJournalStart
...
+        rlRun "data_dir=\$(mktemp -d)" 0 "Create temp data directory"
 
-            # Fetch the latest koji build ID for make dynamically
-            get_koji_build_id "make" "f\${fedora_release}"
+        # FIXME: https://github.com/teemtee/tmt/issues/4742
+        # Both distros use Fedora koji - the koji.build provider always reads
+        # the default 'koji' profile which points to Fedora koji.
+        # Fedora-built diffstat (1.68) is newer than CentOS baseos (1.66)
+        # so the artifact version is always preferred on both distros.
+        get_koji_build_id "diffstat" "f\${fedora_release:-43}"
```
LecrisUT: The whole point of my recommendation of using `epel` is because it is the same koji profile as fedora and would require the least patching right now

tests/prepare/artifact/providers/copr-repository/test.sh
vaibhavdaren: @LecrisUT : Once we release the auto-verify plugin, the increased verbosity should help us catch issues early. We can revert these changes once the new plugin is stable

LecrisUT: A. Then add it in that PR. B. that pr does not use verbose level.

vaibhavdaren: The other PR was big, i wanted to dillute that a bit. Let me move these out. Makes sense.

LecrisUT: > B. that pr does not use verbose level.

Still applies, what exactly are you going to get extra from it?

vaibhavdaren: I’d like to review the verification logs after [PR #4726](https://github.com/teemtee/tmt/pull/4726) tmt repository. Verification should work automatically for providers that download packages, and after [Issue #4714](https://github.com/teemtee/tmt/issues/4714) tmt repository, it should also support discover-only providers.

Since we first create the package and source_repository combination and then validate whether it’s the expected pairing, we effectively control both the input and the expected output. This makes it relatively easy to encounter false negatives

LecrisUT: You did not answer my question. Which verbose message level 3 are you interested in seeing? I.e. does it make sense for that message to be hidden in a level 3 verbose?

vaibhavdaren: The plan was to see these, no point to do it this way. I will change all the important assertions to info. 
https://github.com/teemtee/tmt/pull/4726/changes#diff-a31fc4e29f108169d2d771f6e8bab29d974739e9694785234508c72c9afdbdc3R434-R442

LecrisUT: That is a debug. No level of `-v` would have helped you, you needed a single `-d` instead

tests/prepare/artifact/lib/common.sh
vaibhavdaren: https://artifacts.dev.testing-farm.io/4f16202a-9d39-4659-a62b-a5a17cd00931/work-installh1qyyuyd/plans/provision/container/install/execute/data/guest/default-0/tests/prepare/verify-installation-2/output.txt

I see that centpkg needs other dependencies as well.

LecrisUT: Yes, enabling `epel`

LecrisUT: Also, you are downloading and installing the f43 centpkg for centos. The binary rpms are not interchangeable, use the appropriate `elX` tag

tests/prepare/verify-installation/test.sh
```diff
@@ -10,21 +10,27 @@ rlJournalStart
...
         if rlIsFedora; then
+            # Use the host's actual Fedora release for verify-installation
+            fedora_release="$release"
+            image_name="fedora/${fedora_release}:latest"
+            get_koji_build_id "diffstat" "f${fedora_release}"
             rlRun "cp -r data-fedora/. \$data_dir/" 0 "Copy Fedora test data"
         elif rlIsCentOS; then
+            # Use f40 for CentOS Stream 9 - f43+ packages require newer glibc
+            fedora_release="40"
+            get_koji_build_id "diffstat" "f40"
```
LecrisUT: No. Do not mix fxx tags with centos

tests/prepare/verify-installation/data/main.fmf
vaibhavdaren: the files should be mostly the same now, with minimal differences.

tests/prepare/verify-installation/data-fedora/main.fmf
vaibhavdaren: chnaged back to centpkg.

tests/prepare/verify-installation/data/main.fmf
LecrisUT: Well, then properly unify them. Use a `when` or `adjust` as necessary

tests/prepare/artifact/lib/common.sh
```diff
@@ -24,6 +22,30 @@ setup_distro_environment() {
...
+
+    centos_release=$(rlGetDistroRelease)
+    image_name="centos/stream${centos_release}/upstream:latest"
+    build_container_image "$image_name"
+
+    # For CentOS 10+, packages are tagged under epelX.Y (e.g., epel10.3).
+    # For CentOS 9 and earlier, the plain epelX tag is used (e.g., epel9).
+    epel_tag=$(koji list-tags 2>/dev/null | grep -E "^epel${centos_release}\.[0-9]+$" | sort -V | tail -1)
+    epel_tag="${epel_tag:-epel${centos_release}}"
+}
```
LecrisUT: Move this into `setup_distro_environment`

tests/prepare/artifact/lib/common.sh
LecrisUT: As it is being discussed in https://github.com/teemtee/tmt/pull/4703#discussion_r3016013237, it is a good time to move this to also be automatic similar to the centos handling.

tests/prepare/artifact/lib/common.sh
```diff
@@ -24,6 +22,30 @@ setup_distro_environment() {
...
+    fi
+
+    centos_release=$(rlGetDistroRelease)
+    image_name="centos/stream${centos_release}/upstream:latest"
+    build_container_image "$image_name"
+
+    # For CentOS 10+, packages are tagged under epelX.Y (e.g., epel10.3).
+    # For CentOS 9 and earlier, the plain epelX tag is used (e.g., epel9).
+    epel_tag=$(koji list-tags 2>/dev/null | grep -E "^epel${centos_release}\.[0-9]+$" | sort -V | tail -1)
+    epel_tag="${epel_tag:-epel${centos_release}}"
```
LecrisUT: I need to digest this thing to see if this is the best approach. Definitely want to avoid all of the `grep` -> `sort` -> `tail`

LecrisUT: Generalize the name of `epel_tag` so that it is reusable for either cases.

tests/prepare/verify-installation/test.sh
```diff
@@ -5,71 +5,48 @@
...
+        rlRun "sed -i 's/KOJI_BUILD_ID/${KOJI_BUILD_ID}/g' $data_dir/main.fmf" 0 "Substitute koji build ID"
+        rlRun "pushd $data_dir"
     rlPhaseEnd
 
     rlPhaseStartTest "Test successful verification"
-        rlRun -s "tmt run -i \$run/success --scratch -vvv --all \
+        rlRun -s "tmt run -i $run/success --scratch -vvv --all \
             plan --name /plan/success \
-            provision -h \$PROVISION_HOW --image \$image" \
+            provision -h $PROVISION_HOW --image $image" \
```
LecrisUT, line 27: The change is correct, but I do not like that it is completely random that you or the "assistant" is changing their mind. Please make sure this is avoided in the future.

tests/prepare/artifact/lib/common.sh
LecrisUT: Ok, basically we do not care of `epel10.X`, epel10 will do just fine. There is some minor issues with `epel9` vs `epel9-next`, but I do not expect these to be relevant in these tests. So, can you instead use `rlGetDistroRelease` and just add either `fXX` or `epelXX`. Tested that it works for `rawhide` as well since those builds are tagged with both `rawhide` and `f45`

tests/prepare/verify-installation/test.sh
```diff
@@ -5,71 +5,48 @@
 rlJournalStart
     rlPhaseStartSetup
         rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
-        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
+        rlRun "run=$(mktemp -d)" 0 "Create run directory"
```
thrix: @vaibhavdaren what is the reason for these changes, seems for consistency the first form is used more in tests?

tests/prepare/artifact/lib/common.sh
LecrisUT: https://github.com/teemtee/tmt/pull/4723#discussion_r3016208106

tests/prepare/artifact/lib/common.sh
vaibhavdaren: @LecrisUT : Why i added the special tag logic.
```
(dev) ➜  ~ koji list-tagged --latest epel10 centpkg
Build                                     Tag                   Built by
----------------------------------------  --------------------  ----------------
(dev) ➜  ~ koji list-tagged --latest epel9 centpkg
Build                                     Tag                   Built by
----------------------------------------  --------------------  ----------------
centpkg-0.10.3-1.el9                      epel9                 tdawson
```

LecrisUT: Hmm, then either find a pattern to get it in one command or go through `koji list-targets`. When we convert this to python, processing the latter will be easier

vaibhavdaren: Thanks for the reply. I can use `koji list-tagged --latest --inherit` it works. Making the changes.

tests/prepare/artifact/lib/common.sh
vaibhavdaren: [Note] The tests now run dynamically on Fedora based on host distro. May see failures due to provider compatibility (copr/koji/repo availability varies by distro versions). Will fix as they come up.

LecrisUT: Don't expect any failures on that front. But we should be consolidating the test cases even more, but fine to deffer.

tests/prepare/artifact/lib/common.sh
```diff
@@ -3,29 +3,40 @@
...
+        release=$(rlGetDistroRelease)
+        koji_tag="f${release}"
+        distro="fedora-${release}"
+        # Rawhide reports a numeric release (e.g. 45) but the Makefile
+        # target uses the stable name 'rawhide', not the version number.
+        if grep -qi "rawhide" /etc/os-release; then
+            image_name="fedora/rawhide:latest"
+        else
+            image_name="fedora/${release}:latest"
+        fi
```
LecrisUT: Seems like we missed to do the equivalent bump of https://github.com/teemtee/tmt/pull/4250, I've recreated it in https://github.com/teemtee/tmt/pull/4775.

tests/prepare/artifact/providers/copr-repository/main.fmf
```diff
@@ -2,3 +2,11 @@ summary: Test copr-repository artifact provider
 description: |
     Test copr-repository artifact provider which makes packages from Copr
     repositories available for installation.
+
+# FIXME: mariobl/pyspread only has chroots for Fedora 43+, pick a stable
+# community COPR available on all Fedora releases (including rawhide) and
+# remove this gate once the test data is updated.
```
LecrisUT: Copr can auto-branch. What we actually need is to create our own copr for testing that is not dependent on the distro at all.

tests/prepare/verify-installation/test.sh
vaibhavdaren: @thrix : It worked without the `\` for me. Reading code with lots of `\` escape sequences is difficult for me.

tests/prepare/artifact/providers/copr-repository/main.fmf
vaibhavdaren: Created https://github.com/teemtee/tmt/issues/4776 for tracking.

tests/prepare/verify-installation/test.sh
LecrisUT: Both are valid, the difference is in how it is reported in beakerlib as `run=$(mktemp -d)` or `run=tmp-sij249s`. Not enough of a difference to bother me either way

psss: The main difference is that upon failure the assert is not failing:

```
./test.sh: line 7: mktempp: command not found
:: [ 09:47:30 ] :: [  BEGIN   ] :: Create tmp directory :: actually running 'tmp='
:: [ 09:47:30 ] :: [   PASS   ] :: Create tmp directory (Expected 0, got 0)
```

Not a big deal for `mktemp`, but I'd suggest to keep things consistent.

tests/prepare/artifact/providers/copr-repository/main.fmf
```diff
@@ -2,3 +2,11 @@ summary: Test copr-repository artifact provider
 description: |
     Test copr-repository artifact provider which makes packages from Copr
     repositories available for installation.
+
+# FIXME: mariobl/pyspread only has chroots for Fedora 43+. Replace with a
+# tmt-owned COPR project with auto-branching so the gate can be removed.
+# See https://github.com/teemtee/tmt/issues/4776
+adjust+:
+  - when: distro ~< fedora-43
```
psss: Why using the `~<` operator? It's supposed to be used for comparing [within the same major version](https://fmf.readthedocs.io/en/stable/context.html#major-version), e.g. not relevant for Fedora.

tests/prepare/artifact/providers/copr-repository/main.fmf
```diff
@@ -2,3 +2,11 @@ summary: Test copr-repository artifact provider
...
+    enabled: false
+    because: mariobl/pyspread COPR only has chroots for Fedora 43+
```
psss, line 12: Do we need to use `mariobl/pyspread` which has this limitation?

tests/prepare/artifact/providers/multi/main.fmf
```diff
@@ -2,3 +2,10 @@ summary: Test multiple artifact providers
 description: |
     Test using multiple artifact providers together in a single
     prepare step.
+
+# FIXME: docker-ce.repo does not support Fedora 44+, pick a stable
+# repository-file source available on all Fedora releases and remove this gate.
+adjust+:
+  - when: distro ~> fedora-43
```
psss: Same as above.

tests/prepare/artifact/providers/repository-file/main.fmf
```diff
@@ -2,3 +2,10 @@ summary: Test repository-file artifact provider
 description: |
     Test repository-file artifact provider which configures external
     package repositories from .repo files or URLs.
+
+# FIXME: docker-ce.repo does not support Fedora 44+, pick a stable
+# repository-file source available on all Fedora releases and remove this gate.
+adjust+:
+  - when: distro ~> fedora-43
+    enabled: false
+    because: docker-ce.repo does not support Fedora 44+
```
psss, line 11: What about just choosing a repo which is available across all supported fedoras? Perhaps something small like https://copr.fedorainfracloud.org/coprs/psss/did/ ?

tests/prepare/artifact/providers/copr-repository/main.fmf
vaibhavdaren: This will be addressed in https://github.com/teemtee/tmt/issues/4776

tests/prepare/verify-installation/test.sh
thrix: yeah, let's be ideally consistent

tests/prepare/artifact/providers/copr-repository/main.fmf
vaibhavdaren: Using `~< ` is semantically incorrect. Addressed in af37614d8e953d3c242eb26f040ed0714fd669d1

tests/prepare/verify-installation/data/main.fmf
```diff
@@ -11,26 +11,27 @@
         provide:
           - koji.build:KOJI_BUILD_ID
 
+    adjust+:
+      - when: distro == centos
```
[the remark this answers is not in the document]
AthreyVinay: @LecrisUT Why the 👎 ? Just trying to understand.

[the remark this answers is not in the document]
LecrisUT: `distro == centos` already covers any `centosXYZ` variants. The handling is quite a mess right now https://github.com/teemtee/fmf/blob/39af408279b305bd8744e1a74c4ad09958394587/fmf/context.py#L34-L37, but hopefully we can move these and refactor it, but at least the documentation marks it as the intended behavior https://fmf.readthedocs.io/en/stable/context.html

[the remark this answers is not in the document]
AthreyVinay: Thanks Cris - explains it much better!

tests/prepare/artifact/providers/copr-repository/main.fmf
psss: Thanks!

tests/prepare/verify-installation/test.sh
psss: Thanks!

tests/prepare/verify-installation/test.sh
```diff
@@ -7,69 +7,40 @@ rlJournalStart
...
-            setup_distro_environment
-            rlRun "image=\$TEST_IMAGE_PREFIX/\$image_name"
+        setup_distro_environment
 
-            rlRun "data_dir=\$(mktemp -d)" 0 "Create temp data directory"
+        rlRun "image=$TEST_IMAGE_PREFIX/$image_name"
 
-            # Fetch the latest koji build ID for make dynamically
-            get_koji_build_id "make" "f\${fedora_release}"
+        rlRun "data_dir=$(mktemp -d)" 0 "Create temp data directory"
```
psss:
```suggestion
        rlRun "data_dir=\$(mktemp -d)" 0 "Create temp data directory"
```

PR #4724, opened 2026-03-20
Fix for `pint` fixing dimensionless units

LecrisUT:
Pint 0.25.3 finally fixed its return when using output without unit, but now it unconvered that we were not ready for their fix because we did not check if `raw_value` was unitless or not.

---

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/hardware.py
+++ b/tmt/hardware.py
@@ -584,15 +584,16 @@ class Constraint(BaseConstraint):
         if as_quantity:
             value = UNITS(raw_value)
 
+            # pint < 0.25.3:
             # Number-like raw_value, without units, get converted into
             # pure `int` or `float`. Force `Quantity` for quantities by
             # explicitly wrapping built-in types with `Quantity`.
-            #
-            # reportUnnecessaryIsInstance: pyright cannot infere this
-            # little trick Pint plays with us, and considers the `isinstance`
-            # call to be pointless. But it's not. mypy does not care...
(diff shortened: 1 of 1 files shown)


tmt/hardware.py
```diff
@@ -584,16 +584,17 @@ def _from_specification(
...
-            # reportUnnecessaryIsInstance: pyright cannot infere this
-            # little trick Pint plays with us, and considers the `isinstance`
-            # call to be pointless. But it's not. mypy does not care...
-            if not isinstance(value, pint.Quantity):  # type: ignore[reportUnnecessaryIsInstance,unused-ignore]
+            if not isinstance(value, pint.Quantity):
                 value = pint.Quantity(value, default_unit)
 
+            # Make sure the value has appropriate units if it was not provided
+            if value.unitless and default_unit:
+                value *= UNITS(default_unit)
```
happz, line 596: Would the same operation as the one above work? `value = pint.Quantity(value, default_unit)` ? I had to look really closely to spot the multiplication `*`.

LecrisUT: Good question, apparently not, it gives the error
> pint.errors.DimensionalityError: Cannot convert from 'dimensionless' (dimensionless) to 'megahertz' (1 / [time])

happz: Good to know.

psss (approved):

PR #4726, opened 2026-03-20
Auto-verify artifact packages against require/recommend

vaibhavdaren:
When the artifact plugin is present with verify=True (default), a verify-installation phase is automatically injected at order=79. It intersects require/recommend packages for the current guest with packages listed in artifacts.yaml, and verifies that the overlapping packages were installed from the tmt-artifact-shared repository.

Related to #4646

### DIFF
--- a/tests/prepare/artifact/nvr-priority/data/tc11.fmf
+++ b/tests/prepare/artifact/nvr-priority/data/tc11.fmf
@@ -14,6 +14,7 @@ summary: TC 1.1 - Repository priority overrides package version
           script: mkdir -p /tmp/nvr-test && cp -r tc11-high tc11-low /tmp/nvr-test/
         - name: Install repo files
           how: artifact
+          verify: false
           provide:
             - repository-file:file:tc11-high.repo
             - repository-file:file:tc11-low.repo


tmt/steps/prepare/__init__.py
```diff
@@ -370,6 +370,44 @@ def _emit_phase(
...
+                )
+            verify_data = PrepareVerifyInstallationData(
+                name='verify-artifact-packages',
+                how='verify-installation',
+                summary='Verify packages were installed from artifact repositories',
+                order=tmt.steps.PHASE_ORDER_PREPARE_VERIFY_INSTALLATION,
+                auto=True,
+                where=verify_where,
+            )
+            self._phases.append(PreparePlugin.delegate(self, data=verify_data))
```
LecrisUT: Uph, please not like this. Would rather it be created inside the `artifact` plugin instead, where we would also control how it is created. The questionable thing is if can be created while in the `prepare.go` phase, but that needs to be solved regardless because we do not know the packages and their origins until `PrepareArtifact` (probably all of them) complete.

LecrisUT: For reference, see Discover dist-git in how it inserts the phase dynamically

happz: It does, but it's a `discover` phase adding to `prepare` step that haven't started yet. `prepare` phase adding to `prepare` step, after the queue has been established and started is not supported. Right now, I would do it in the `prepare` step, just as test requirements are sorted out there. It's not ideal, but it's available today, and can be improved independently later.

LecrisUT: Uph. At the minimum, encapsulate it in a function, and document to get rid of it as soon as possible.

tmt/steps/prepare/verify_installation.py
```diff
@@ -58,6 +71,76 @@ class PrepareVerifyInstallation(PreparePlugin[PrepareVerifyInstallationData]):
...
+    def _build_auto_verify(self, guest: Guest) -> dict[str, str]:
+        """
+        Build a verify mapping from artifact metadata and test requirements.
+
+        Reads ``artifacts.yaml`` from the plan workdir, collects all
+        package names provided by artifact providers, and intersects them
+        with packages listed in ``require``/``recommend`` of tests enabled
+        on this guest.  Returns a mapping of intersecting package names to
+        the artifact shared repository name.
+        """
```
LecrisUT: Well, the second issue with not being able to inject the phase is having this magic method. Didn't like the branching in discover, and this just adds another flavor of it. This wouldn't even work if you have one artifact provider with verify and another without since it cannot discriminate between them, unless there is even more introspection between these phases

tmt/steps/prepare/__init__.py
LecrisUT: On second thought, I still don't like it. What does it take to inject the phase in the queue properly and improperly?

happz: Changing how queues are populated in steps, adding API to support dynamic additions while the queue is already running, which most likely requires reordering of the queued tasks as tasks were enqueued according to phases' `order` keys. I don't mind working on that, already some needed work landed, it just won't be ready today.

happz: And a unit test, to make sure it works because this would be rather a rare use case. If there's a bug in how queue works, we'd see tmt explode everywhere; if there's a bug in how tasks are enqueued from another tasks of the same step, only a handful of tests would fail, so this calls for having a test as well.

LecrisUT: We did create the interface in preparation for it at least
https://github.com/teemtee/tmt/blob/b32fa85ce8ecbd9f2656ebabc1a3e5a282c1f998/tmt/steps/__init__.py#L1322-L1330

happz: Added #4731 and #4729 to extend how queue can be modified, connecting the dots will come in a separate patch.

LecrisUT: Can we do the dots connecting with `PrepareDistgit`, that should be a good example on how to do that later on

happz: Sure, but most likely the change would be minimal in that particular case, because `PrepareDistgit` is different as it adds new phase to future step, while here we wish to add phase to the current step, one that's already running its phases. There is no `prepare` queue yet when `discover/fmf` adds `PrepareDistgit` phase.

LecrisUT: > Sure, but most likely the change would be minimal in that particular case, because `PrepareDistgit` is different as it adds new phase to future step, while here we wish to add phase to the current step, one that's already running its phases. There is no `prepare` queue yet when `discover/fmf` adds `PrepareDistgit` phase.

Not quite. `discover/fmf` adds all of the 3 phases in the prepare step, but the change I have in mind is for it to only add `prepare/distgit` and in turn, `prepare/distgit` adds the two installation phases later in the future as opposed to the current approach of manipulating the data
https://github.com/teemtee/tmt/blob/a42b13c19bd2bdeb8ba00de060851fdba8de6d37/tmt/steps/prepare/distgit.py#L341-L346

happz: I see. Ok, I can do that, that would be 4th PR, and it does not seem like a blocker for this PR. 3 patches to allow adding phases to already humming steps, then this one can continue, and 4th PR would update `prepare/distgit` workflow to use the same approach. How about this plan?

LecrisUT: > How about this plan?

Sounds good with me. But the changes are not strictly blocking this PR. I am ok if a similar design as the `PrepareDistgit` is used as an intermediate step.

If you are ok with that as well I will leave it to @vaibhavdaren to choose between these 2 options: depend on #4729 and #4731 or use a `future_*` plugin and edit its data

happz: https://github.com/teemtee/tmt/pull/4750 enhances the step API. As mentioned above, distgit would be a separate patch, and we might have a 5th one depending on whether this PR gets merged before the #4750 series, or after.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -37,6 +98,17 @@ class PrepareArtifactData(PrepareStepData):
...
+    auto_verify: bool = field(
+        default=True,
+        option='--auto-verify/--no-auto-verify',
+        is_flag=True,
+        help="""
+            Automatically verify that packages from require/recommend that are
+            present in the artifact metadata were installed from the artifact
+            repository. Enabled by default.
+            """,
+    )
```
LecrisUT: Should be named `verify`. Drop the "auto" part

tests/prepare/artifact/providers/koji-build/data/main.fmf
```diff
@@ -12,6 +12,17 @@
...
+/plan/auto-verify:
+    summary: Verify auto-inject of verify-installation phase when auto_verify=true (default)
+
+/plan/no-verify:
+    summary: Verify auto_verify=false suppresses auto-inject of verify-installation phase
+    prepare:
+      - how: artifact
+        provide:
+          - koji.build:DUMMY_KOJI_BUILD_ID
+        auto_verify: false
```
LecrisUT: What is the purpose of this test? If you want a test checking the non-verify, then put it somewhere outside of the sub-provider specific test. The verification is not provider specific.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -15,6 +18,64 @@
...
 from tmt.utils import Environment, Path
 
+if TYPE_CHECKING:
+    import tmt.steps.prepare
+
+#: Name of the shared repository created by the artifact plugin.
+ARTIFACT_SHARED_REPO_NAME: Final[str] = 'tmt-artifact-shared'
+
+#: Filename of the artifact metadata file written by the artifact plugin.
+ARTIFACT_METADATA_FILENAME: Final[str] = 'artifacts.yaml'
```
LecrisUT: Why did  you move these outside?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -15,6 +18,64 @@
...
+#: Name of the auto-injected verify-installation phase.
+VERIFY_PHASE_NAME: Final[str] = 'verify-artifact-packages'
+
+#: Summary of the auto-injected verify-installation phase.
+VERIFY_PHASE_SUMMARY: Final[str] = 'Verify packages were installed from artifact repositories'
+
+
+def inject_verify_phase(
+    prepare_step: 'tmt.steps.prepare.Prepare', where: list[str]
+) -> 'PreparePlugin[Any]':
```
LecrisUT: Not needed to be at module-level. Can be owned by `PrepareArtifact`

tmt/steps/prepare/__init__.py
```diff
@@ -370,6 +370,57 @@ def _emit_phase(
...
+                    {dest for phase in artifact_phases_with_verify for dest in phase.data.where}
+                )
+
+            # Inject the verify phase via artifact plugin
+            future_verify = inject_verify_phase(self, verify_where)
+
+            # Give each artifact phase a reference so go() can populate the verify dict
+            for phase in artifact_phases_with_verify:
+                phase._future_verify = future_verify
+
```
LecrisUT: Please no specific implementations in here. Move this in the `__init__` of the `PrepareArtifact`. If you are missing something, or you find an issue with the order of operations, please voice them instead

tmt/steps/prepare/verify_installation.py
```diff
@@ -71,17 +71,19 @@ def go(
         if self.is_dry_run:
             return outcome
 
-        if not self.data.verify:
+        expected_origins: dict[str, str] = dict(self.data.verify)
+
+        if not expected_origins:
```
LecrisUT: Leftover from refactoring?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -268,6 +357,9 @@ def go(
         # Persist artifact metadata to YAML
         self._save_artifacts_metadata(providers)
 
+        # Populate the future verify phase with package mappings
+        self._populate_verify_from_providers(providers, guest)
```
LecrisUT: Gate this by the `verify` key

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+        Computes the intersection of artifact package names and package names
+        collected from test require/recommend (filtered to those enabled on *guest*)
+        and explicit prepare install phases, then updates :py:attr:`_future_verify`
+        with ``{package: ARTIFACT_SHARED_REPO_NAME}`` entries for each package in
+        the intersection.
+
+        Called from :py:meth:`go` after providers have been initialized and artifacts
+        are known. Each artifact phase contributes only its own providers' packages.
+        Different artifact phases run sequentially in the queue so no cross-phase race
+        is possible.
```
LecrisUT: What is the key point you want to say here?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+
+        TODO: After #4729/#4731, this should create and inject the verify phase
+        directly using ``self.step.add_phase()`` instead of populating a
+        pre-registered phase via ``_future_verify``.
+
+        Computes the intersection of artifact package names and package names
+        collected from test require/recommend (filtered to those enabled on *guest*)
+        and explicit prepare install phases, then updates :py:attr:`_future_verify`
+        with ``{package: ARTIFACT_SHARED_REPO_NAME}`` entries for each package in
+        the intersection.
```
LecrisUT: Seems unnecessarily technical

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+        """
+        if self._future_verify is None:
+            return
+
+        from typing import cast
+
+        # @provides_method erases PrepareInstall's type to PluginClass; the pyright
+        # suppression covers the resulting reportUnknownVariableType on the import and
+        # reportUnknownArgumentType on its use as a classes= argument.  The cast is
+        # needed by Pyright but redundant for mypy (hence type: ignore[redundant-cast]).
```
LecrisUT: If you are confident that this is what's going on, please make the issue so we can correct these

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+
+        Called from :py:meth:`go` after providers have been initialized and artifacts
+        are known. Each artifact phase contributes only its own providers' packages.
+        Different artifact phases run sequentially in the queue so no cross-phase race
+        is possible.
+        """
+        if self._future_verify is None:
+            return
+
+        from typing import cast
```
LecrisUT: redundnat

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+
+        # @provides_method erases PrepareInstall's type to PluginClass; the pyright
+        # suppression covers the resulting reportUnknownVariableType on the import and
+        # reportUnknownArgumentType on its use as a classes= argument.  The cast is
+        # needed by Pyright but redundant for mypy (hence type: ignore[redundant-cast]).
+        from tmt.steps.prepare.install import (
+            PrepareInstall,  # pyright: ignore[reportUnknownVariableType]
+        )
```
LecrisUT: Cannot be moved to the top?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+            f"Auto-verifying {fmf.utils.listed(sorted(intersection), 'package')} "
+            f"against '{ARTIFACT_SHARED_REPO_NAME}'."
+        )
+
+        uncovered = pkg_names - artifact_pkg_names
+        if uncovered:
+            self.debug(
+                f"{fmf.utils.listed(sorted(uncovered), 'package')} "
+                f"from require/recommend/install not provided by this artifact phase."
+            )
```
LecrisUT: Why do we care about this. It is also duplicated across the calls

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +368,89 @@ def go(
...
+
+        self._future_verify.data.verify.update(
+            dict.fromkeys(intersection, ARTIFACT_SHARED_REPO_NAME)
+        )
```
LecrisUT: Forcing only `ARTIFACT_SHARED_REPO_NAME` is incorrect. The `provider.artifacts` have artifacts in other repos as well. The artifacts must own a `repo` field or equivalent that would specify where they come from.

tmt/steps/prepare/verify_installation.py
vaibhavdaren: If the user specifies empty dict. Which is a vaild input as well. we return early.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Added  https://github.com/teemtee/tmt/issues/4766 , with some prework.

tmt/steps/prepare/artifact/__init__.py
LecrisUT: The help text still needs adjustments. Drop the `Automatically`. And similaraly some other occurrences of `Auto-` still remain

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -172,6 +186,86 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
...
+        self._future_verify = None
+
+        if not self.data.verify:
+            return
+
+        # Avoid circular import with verify_installation.
+        from tmt.steps.prepare.verify_installation import (
+            PrepareVerifyInstallation,  # pyright: ignore[reportUnknownVariableType]
+            PrepareVerifyInstallationData,
+        )
```
LecrisUT: There is no circular dependencies anymore

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -172,6 +186,86 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
...
+
+        # Skip if the user already has a verify-installation phase.
+        user_verify_exists = any(
+            isinstance(phase, PrepareVerifyInstallation)  # pyright: ignore[reportArgumentType,reportUnknownArgumentType]
+            and phase.data.name != self.VERIFY_PHASE_NAME
+            for phase in self.step._phases
+        )
+        if user_verify_exists:
+            return
```
LecrisUT: Why? What scenario would the user verify be intended as overriding instead of complementary?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -172,6 +186,86 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
...
+
+        # Reuse verify phase if already injected by an earlier artifact phase.
+        existing_verify: Optional[PreparePlugin[Any]] = cast(
+            Optional[PreparePlugin[Any]],
+            next(
+                (  # pyright: ignore[reportUnknownArgumentType]
+                    phase
+                    for phase in self.step._phases
+                    if isinstance(phase, PrepareVerifyInstallation)  # pyright: ignore[reportArgumentType]
```
LecrisUT: Use `self.step.phases()` here to avoid the `isinstance`. Would like to have this expanded to cover the `name` and such, but that's for a later change

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -172,6 +186,86 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
...
+                summary=self.VERIFY_PHASE_SUMMARY,  # pyright: ignore[reportAttributeAccessIssue,reportUnknownArgumentType]
+                order=tmt.steps.PHASE_ORDER_PREPARE_VERIFY_INSTALLATION,
+                where=list(self.data.where),
+                verify={},
+            )
+            verify_phase: PreparePlugin[Any] = cast(
+                PreparePlugin[Any],
+                PreparePlugin.delegate(self.step, data=verify_data),
+            )
+            self.step._phases.append(verify_phase)
```
```suggestion
            self.step.add_phase(verify_phase)
```

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -172,6 +186,86 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
...
+            self._future_verify = verify_phase
+        else:
+            # Expand where= to include this phase's guests.
+            if not self.data.where or not existing_verify.data.where:
+                existing_verify.data.where = []
+            else:
+                existing_verify.data.where = list(
+                    set(existing_verify.data.where) | set(self.data.where)
+                )
```
LecrisUT: Does not seem correct. This would inject more artifact checks that a guest would not have had access to. Better to make the `existing_verify` aware of `where` as well.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+    def _populate_verify_from_providers(
+        self, providers: list[ArtifactProvider], guest: Guest
+    ) -> None:
+        """
+        Populate the pre-registered verify phase with the package→repo mapping.
+
+        TODO: After #4729/#4731, replace this with self.step.add_phase() in go().
+        """
+        if self._future_verify is None:
+            return
```
```suggestion
        assert self._future_verify is not None  # Narrow type
```

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+
+        # Inline import to avoid circular dependency at module level.
+        from tmt.steps.prepare.install import (
+            PrepareInstall,  # pyright: ignore[reportUnknownVariableType]
+        )
+
+        pkg_names: set[str] = set()
```
LecrisUT, line 348:
```suggestion
        pkgs_to_verify: dict[str, list[str]] = {}
```
Clearer name and includes the information of the repo owning the package: key is the required/recommended package, value are the repositories it must validate to have come from. Unavoidably you will have multiple repo locations where a package can originate from (for example there can be multiple non-clashing NVRs). What to do when multiple repos are encountered has to be defined in this plugin.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+
+        # Collect packages from test require/recommend enabled on this guest.
+        for test_origin in self.step.plan.discover.tests(enabled=True):
+            test = test_origin.test
+            if not test.enabled_on_guest(guest):
+                continue
+            for dep in (*test.require, *test.recommend):
+                if isinstance(dep, DependencySimple):
+                    pkg_names.add(str(dep))
```
LecrisUT: Why is this needed. It **has** to be covered in the `_install_phases` otherwise something is wrong

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+        # Collect packages explicitly listed in prepare install phases.
+        _install_phases = cast(  # type: ignore[redundant-cast]
+            list[Any],
+            self.step.phases(
+                classes=PrepareInstall,  # pyright: ignore[reportUnknownArgumentType]
+            ),
+        )
+        for install_phase in _install_phases:
+            for pkg in install_phase.data.package:
+                pkg_names.add(str(pkg))
```
LecrisUT: Check that these belong to the relevant guest first

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+        # FIXME: Providers that manage their own repositories (repository-file,
+        # repository-url, copr.repository) should not be verified against
+        # tmt-artifact-shared. See https://github.com/teemtee/tmt/issues/4714
+        verify_skip_providers: frozenset[str] = frozenset(
+            {
+                'repository-file',
+                'repository-url',
+                'copr.repository',
+            }
+        )
```
LecrisUT: The current agreement is that they must still be verified from their origin repository that populated the `artifacts` entry. There will be issues with overlapping `name`, but that will be yet another round of debates.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+        verifiable_providers = [
+            p for p in providers if p.raw_id.split(':', maxsplit=1)[0] not in verify_skip_providers
+        ]
+
+        # Artifact package names from providers in this phase only
+        artifact_pkg_names: set[str] = {
+            artifact.version.name
+            for provider in verifiable_providers
+            for artifact in provider.artifacts
+        }
```
LecrisUT: Please see the recent changes with `enumerate_artifacts`. This needs a refactoring to drop the `verify_skip_providers`. Basically the `location` for the artifacts is the `repository.name` that needs to be used for the verification (not ideal, it should have a dedicated field to disambiguate)

This is a blocker for this PR because `copr.build` has both type of artifacts that need to be accounted for.

tests/prepare/artifact/verify-installation/data/main.fmf
```diff
@@ -0,0 +1,31 @@
...
+    prepare:
+      - how: artifact
+        provide:
+          # DUMMY_KOJI_BUILD_ID placeholder - replace with actual koji build ID
+          # Example: koji.build:2761810
+          - koji.build:DUMMY_KOJI_BUILD_ID
+    execute:
+        how: tmt
+
+/plan/auto-verify:
```
psss:
```suggestion
    /auto-verify:
```

Indenting the child plan content instead of the the full name can be a bit more readable as it clearly shows the parent-child relation.

vaibhavdaren:
Thanks for the review @LecrisUT.
This looks a lot better and simpler with the new queue functions added by @happz. Thankyou ^_^ for adding the support. See https://github.com/teemtee/tmt/commit/0d2b4c9ae665080f876063e05ae71e397dfa395a. Will rebase this PR  and add new changes  Once https://github.com/teemtee/tmt/pull/4750 is merged.

tests/prepare/artifact/verify-installation/data/main.fmf
psss: Indenting the child plan instead of the the full name should be a bit more readable.

tests/prepare/artifact/verify-installation/data/main.fmf
```diff
@@ -0,0 +1,31 @@
...
+    summary: Verify auto-inject of verify-installation phase when verify=true (default)
+
+/plan/no-verify:
+    summary: Verify verify=false suppresses auto-inject of verify-installation phase
+    prepare:
+      - how: artifact
+        provide:
+          - koji.build:DUMMY_KOJI_BUILD_ID
```
psss:
```suggestion
    prepare+:
```

Or just extend the `prepare` definition and prevent duplication?

tests/prepare/artifact/verify-installation/test.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+# Test verify-installation auto-inject behaviour for artifact plugin
```
psss:
```suggestion
```

No need to duplicate test metadata already in `main.fmf`?

tests/prepare/artifact/verify-installation/test.sh
```diff
@@ -0,0 +1,52 @@
...
+        setup_distro_environment
+
+        # Get koji build ID for make
+        get_koji_build_id "make" "f${fedora_release}"
+
+        # Create a substituted copy of the data dir for plan-based tests
+        rlRun "data_dir=\$(mktemp -d)" 0 "Create temp data directory"
+        rlRun "cp -r . $data_dir/" 0 "Copy test data"
+        rlRun "sed -i 's/DUMMY_KOJI_BUILD_ID/${KOJI_BUILD_ID}/g' $data_dir/main.fmf" \
+            0 "Substitute koji build ID"
```
psss: Hm, thinking about it a bit more: Wouldn't it be easier in this case to just define the `prepare` step from the command line? We would save ourselves from the need to copy and modify the plan data.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -276,6 +374,77 @@ def go(
...
+
+        intersection = artifact_pkg_names & pkg_names
```
psss: Wasn't the agreement to use `rpm -q --whatprovides` to get the list of actual rpm packages providing the requested stuff? This should cover cases when the `require` key contains things like `/usr/bin/something`.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -37,6 +44,17 @@ class PrepareArtifactData(PrepareStepData):
...
     )
 
+    verify: bool = field(
+        default=True,
+        option='--verify/--no-verify',
+        is_flag=True,
+        help="""
+            Verify that packages from install phases that are present in the
+            artifact metadata were installed from the provider artifact
+            repository. Enabled by default.
```
happz: This message is puzzling to me.
 
1. it's not *all* install phases, just those representing test requirements (and essential requirements and recommended packages) - we should probably ignore `prepare/install` phases added by user.
2. "phases that are present in the artifact metadata" - this I don't understand at all. Maybe "... phases installing test requirements..."?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -172,6 +190,14 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
     SHARED_REPO_NAME: ClassVar[str] = 'tmt-artifact-shared'
     ARTIFACTS_METADATA_FILENAME: ClassVar[str] = 'artifacts.yaml'
 
+    #: Name of the auto-injected verify-installation phase.
+    VERIFY_PHASE_NAME: ClassVar[str] = 'verify-artifact-packages'
+
+    #: Summary of the auto-injected verify-installation phase.
+    VERIFY_PHASE_SUMMARY: ClassVar[str] = (
+        'Verify packages were installed from artifact repositories'
+    )
```
happz, line 210: Both texts suffer same point I mentioned above, both suggest they apply to everything while we should focus on test requirements only.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +310,75 @@ def go(
...
+        Inject a verify-installation phase for packages from these providers.
+
+        If a verify phase already exists for the same where= group, merge
+        the packages into it. Otherwise, create and add a new phase.
+        """
+        # Collect packages explicitly listed in prepare install phases
+        # that are enabled for this guest.
+        pkg_names: set[str] = set()
+        _install_phases = self.step.phases(classes=PrepareInstall)  # pyright: ignore[reportUnknownVariableType,reportUnknownArgumentType]
+        for install_phase in _install_phases:  # pyright: ignore[reportUnknownVariableType]
```
happz, line 350: Yep, doing this for all phases, and I don't think that's the right thing to do at this point, we should care about phases added by us on behalf of test and essential requirements.

tmt/steps/prepare/artifact/__init__.py
happz: I'd also drop the "Enabled by default" bit, all help text rendering should be able to display `default=True` automagically.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +310,75 @@ def go(
...
+            if not install_phase.enabled_on_guest(guest):  # pyright: ignore[reportUnknownMemberType]
+                continue
+            for pkg in install_phase.data.package:  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+                pkg_names.add(str(pkg))  # pyright: ignore[reportUnknownArgumentType]
+
+        # Build package → origin repo mapping from all providers.
+        # FIXME: ``artifact.location`` is overloaded: a download URL for download
+        # providers and a repo name for repository providers. We use
+        # ``provider.get_repositories()`` as a proxy until ``ArtifactInfo`` gains
+        # a dedicated ``verification_repo`` field. See https://github.com/teemtee/tmt/issues/4714.
```
LecrisUT: The linked issue is empty. Please add context into it. Also my vote is on `repo`, not `verification_repo` for the name.

tests/prepare/verify-installation/test.sh
```diff
@@ -74,7 +74,6 @@ rlJournalStart
         rlAssertGrep "fail .* / random-non-existent-package" $rlRun_LOG
         rlAssertGrep "random-non-existent-package.*not installed" $rlRun_LOG
         rlAssertGrep "Package source verification failed for:" $rlRun_LOG
-        rlAssertNotGrep "All packages verified successfully." $rlRun_LOG
```
LecrisUT, line 48: Why is this one lost?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +310,75 @@ def go(
...
+            else:
+                # Download provider: packages land in the shared repo.
+                for artifact in provider.artifacts:
+                    pkg_to_repo[artifact.version.name] = self.SHARED_REPO_NAME
+
+        # Only verify packages that are both required and from a known artifact.
+        pkgs_to_verify = {pkg: repo for pkg, repo in pkg_to_repo.items() if pkg in pkg_names}
+
+        if not pkgs_to_verify:
+            self.debug('No overlap between artifact packages and install phases.')
```
```suggestion
            self.verbose('No packages to be installed were found in the provided artifacts')
```
Too technical (avoid the `overlap` part because that only makes sense if you know the inner implementation), and better to move it outside of `debug`

tmt/steps/prepare/artifact/__init__.py
LecrisUT: I don't know about that. It would require quite a more complex implementation to wrap our heads around, not to mention it would be even more incompatible with bootc mode (because we cannot query anything live from there). I would say to just use a `TODO` for now.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +310,75 @@ def go(
...
+        pkg_names: set[str] = set()
+        _install_phases = self.step.phases(classes=PrepareInstall)  # pyright: ignore[reportUnknownVariableType,reportUnknownArgumentType]
+        for install_phase in _install_phases:  # pyright: ignore[reportUnknownVariableType]
+            if not install_phase.enabled_on_guest(guest):  # pyright: ignore[reportUnknownMemberType]
+                continue
+            for pkg in install_phase.data.package:  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+                pkg_names.add(str(pkg))  # pyright: ignore[reportUnknownArgumentType]
+
+        # Build package → origin repo mapping from all providers.
+        # FIXME: ``artifact.location`` is overloaded: a download URL for download
```
LecrisUT: `FIXME` -> `TODO`. To me `FIXME` indicates that it is something that should be fixed within the current PR, not at a later date.

tmt/steps/prepare/artifact/__init__.py
LecrisUT: > Yep, doing this for all phases, and I don't think that's the right thing to do at this point, we should care about phases added by us on behalf of test and essential requirements.

I disagree here. The user can use either a `prepare/install` or `requires`/`recommend` for the same purpose, and we should not have diverging behavior between them. The only additional check I would say we should have is to check for the order (do nothing if the install prepare is scheduled after this verification phase)

happz: > The user can use either a `prepare/install` or `requires`/`recommend` for the same purpose, and we should not have diverging behavior between them.

And what if the user did not want this verification? Sure they can, but what if they used `prepare/install` and get this extra step with no chance to avoid it? I think we are overstepping the scope of "install test requirements and make sure they come from artifact providers" assignment.

LecrisUT: > And what if the user did not want this verification?

Either we add a flag `verify` in the `prepare/install` for them to indicate so, or they use the ordering. Otherwise the arguments works the other way around as well.

happz: > Otherwise the arguments works the other way around as well.

Except that not forcing the verification on everything the risk of making user angry or breaking their workflow is smaller, as we wouldn't impose something new on them they cannot disable unless they modify their plans.

LecrisUT: > > Otherwise the arguments works the other way around as well.
> 
> Except that not forcing the verification on everything the risk of making user angry or breaking their workflow is smaller, as we wouldn't impose something new on them they cannot disable unless they modify their plans.

The trade-off here would be that it now silently fails because the user is not aware of this difference. My opinion is that all Fedora-CI uesrs would not have this requirement/expectation as they are only working with one artifact provider. For the users that have the other expectations, I would say that it is better to dive deeper in the installation of the `prepare/artifacts` instead, e.g. wouldn't the add artifacts as disabled be what they expect instead, i.e. do the opposite of the verify and make sure the artifacts do not come from us?

happz: > > > Otherwise the arguments works the other way around as well.
> > 
> > 
> > Except that not forcing the verification on everything the risk of making user angry or breaking their workflow is smaller, as we wouldn't impose something new on them they cannot disable unless they modify their plans.
> 
> The trade-off here would be that it now silently fails because the user is not aware of this change. My opinion is that all Fedora-CI uesrs would not have this requirement/expectation as they are only working with one artifact provider. For the users that have the other expectations, I would say that it is better to dive deeper in the installation of the `prepare/artifacts` instead, e.g. wouldn't the add artifacts as disabled be what they expect instead, i.e. do the opposite of the verify and make sure the artifacts do not come from us?

I don't have the answer. I just don't want to look for it under the pressure of tmt changing its behaviour for everyone, even for users we not necessarily meant to affect.

tests/prepare/verify-installation/test.sh
vaibhavdaren: the assertion was intentionally removed because the auto-injected verify phase now runs alongside the failure test and prints "All packages verified successfully."

tmt/steps/prepare/artifact/__init__.py
happz: From the hacking session: verify only the phases added by tmt, i.e. those based on test `require` and `recommend` keys, and essential requirements.

vaibhavdaren: Addressed in a4de55ba5e231de72252fddce2375334412e68da. I have created the issue for the follow up https://github.com/teemtee/tmt/issues/4781.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Added a todo in a subsequent ca240079f413116f05ad5667e001453cfeb4b435.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Addressed in ca240079f413116f05ad5667e001453cfeb4b435.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Addressed in https://github.com/teemtee/tmt/pull/4726/changes/ca240079f413116f05ad5667e001453cfeb4b435.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +310,79 @@ def go(
...
+        """
+        Inject a verify-installation phase for packages from these providers.
+
+        If a verify phase already exists for the same where= group, merge
+        the packages into it. Otherwise, create and add a new phase.
+        """
+        # Collect packages from the install phases injected by tmt on behalf of
+        # test/essential requirements (names set in Prepare._go). User-defined
+        # prepare/install phases are intentionally excluded.
+        _tmt_install_phase_names = {'essential-requires', 'requires', 'recommends'}
```
LecrisUT: Missing `requires (dist-git)` and `recommends (dist-git)`. Please also specify where these are coming from

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -37,6 +44,17 @@ class PrepareArtifactData(PrepareStepData):
...
     )
 
+    verify: bool = field(
+        default=True,
+        option='--verify/--no-verify',
+        is_flag=True,
+        help="""
+        Verify that packages from tmt-injected install phases (test require/recommend keys,
+        their dist-git equivalents, and essential-requires) were installed from the correct
+        provider artifact repository. User-defined prepare/install phases are not covered.
```
happz:
```suggestion
        Verify that packages from tmt-injected ``prepare/install`` phases (test ``require``/``recommend`` keys,
        their dist-git equivalents, and essential requires) were installed from the correct
        provider artifact repository. User-defined ``prepare/install`` phases are not covered.
```

Or something like that.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +316,92 @@ def go(
...
+        #   'recommends'            — Prepare._go() in tmt/steps/prepare/__init__.py
+        #   'requires (dist-git)'   — tmt/steps/prepare/distgit.py
+        #   'recommends (dist-git)' — tmt/steps/prepare/distgit.py
+        _tmt_install_phase_names = {
+            'essential-requires',
+            'requires',
+            'recommends',
+            'requires (dist-git)',
+            'recommends (dist-git)',
+        }
```
happz, line 347: File a follow-up issue, please, we should add some form of subscription mechanism so plugins can register for this check. Something rather simpler, in case all `prapre/install` would end up covered by the verification, but static lists are hard to maintain. Not a blocker for now.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -13,6 +15,11 @@
     ArtifactProvider,
     Repository,
 )
+from tmt.steps.prepare.install import PrepareInstall  # pyright: ignore[reportUnknownVariableType]
```
happz, line 22: Yeah, I'd add a comment before this line, to explain why so many `pyright` waivers ended up in the code.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Added a new issue: https://github.com/teemtee/tmt/issues/4784.

tests/prepare/artifact/verify-installation/data/main.fmf
```diff
@@ -0,0 +1,20 @@
...
+    provision:
+        how: container
+    execute:
+        how: tmt
+
+    /verify:
+        summary: Verify injection of verify-installation phase when verify=true (default)
+
+    /no-verify:
+        summary: Verify verify=false suppresses verify-installation phase injection
```
psss: Now when the parameters are coming from the command line there is no need to have two (identical) plans, right?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -280,6 +321,92 @@ def go(
...
+        pkg_to_repo: dict[str, str] = {}
+        for provider in providers:
+            if provider.get_repositories():
+                # Repository provider: location is the repo name set by enumerate_artifacts().
+                for artifact in provider.artifacts:
+                    pkg_to_repo[artifact.version.name] = artifact.location
+            else:
+                # Download provider: packages land in the shared repo.
+                for artifact in provider.artifacts:
+                    pkg_to_repo[artifact.version.name] = self.SHARED_REPO_NAME
```
psss, line 377: I did a quick test with the following config:

```yaml
/plan:
    discover:
        how: fmf
    provision:
        how: container
        image: fedora:42
    prepare:
        how: artifact
        #provide: koji.build:2970624
        #provide: repository-url:https://download.copr.fedorainfracloud.org/results/psss/did/fedora-42-x86_64/
        provide: repository-file:https://copr.fedorainfracloud.org/coprs/psss/did/repo/fedora-42/psss-did-fedora-42.repo
    execute:
        how: tmt

/test:
    test: rpm -q did
    require: did
```

Works fine for `koji.build` and `repository-url` but fails for `repository-file` as the repo name does not match:

```
fail verify-artifact-packages / did
Note: Package 'did': expected repo 'psss-did-fedora-42', actual 'copr:copr.fedorainfracloud.org:psss:did'.
```

Any quick fix to make the `repository-file` provider working as well?

tests/prepare/artifact/verify-installation/data/main.fmf
vaibhavdaren: Addressed in cc9f9bdd4d4de45d4790a11eff459cbf8c2ff9c6.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Addressed in f2ac796d5ac4ad108c1c47bb446e6b4e940fd188.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Added context to https://github.com/teemtee/tmt/issues/4714

vaibhavdaren:
All the tests passed :) ^_^.  Rebasing for the final merge

PR #4727, opened 2026-03-20
Add additional conversion methods to step & guest dataclasses

therazix:
This PR changes some `to_spec/to_minimal_spec` conversion methods in the step and guest dataclasses and also adds additional ones that were missing. Steps should now be correctly converted, and it should get rid of `cannot represent an object` errors.

Blocks: #4661 

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -1326,7 +1326,6 @@ class GuestData(
         return super().from_spec(raw_data, logger)  # type: ignore[call-arg]
 
     def to_spec(self) -> tmt.steps._RawStepData:
-        """Convert to a form suitable for saving in a specification file"""
         spec = super().to_spec()
 
         spec.pop('facts', None)  # type: ignore[typeddict-item]
@@ -1338,10 +1337,6 @@ class GuestData(
         return spec
 
     def to_minimal_spec(self) -> tmt.steps._RawStepData:
-        """
-        Convert to a form suitable for saving in a specification
-        file while skipping empty values.
-        """
         spec = {**super().to_minimal_spec()}
 
         spec.pop('facts', None)


tmt/steps/discover/shell.py
```diff
@@ -185,7 +191,29 @@ def to_spec(self) -> dict[str, Any]:
...
+            value = getattr(self, option_to_key(key), None)
+            if value is not None:
+                value = transform(value)
+            # Do not include empty values
+            if value in (None, [], {}):
+                data.pop(key, None)
+            else:
+                data[key] = value
+
+        return data
```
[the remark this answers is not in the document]
therazix: `getattr` needs keys with underscores, while `data[key]` needs keys with dashes. I would keep `option_to_key` in case we add additional keys to `field_map` in the future that contain dashes.

tmt/steps/provision/connect.py
```diff
@@ -63,6 +63,48 @@ class ConnectGuestData(tmt.guest.GuestSshData):
...
+            value = getattr(self, option_to_key(key), None)
+            if value is not None:
+                value = transform(value)
+            # Do not include empty values
+            if value in (None, [], {}):
+                spec.pop(key, None)
+            else:
+                spec[key] = value
+
+        return cast(tmt.steps._RawStepData, spec)
```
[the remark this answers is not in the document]
therazix: https://github.com/teemtee/tmt/pull/4727#discussion_r2967353198

tmt/guest/__init__.py
```diff
@@ -1325,13 +1326,46 @@ def from_spec(  # type: ignore[override]
         return super().from_spec(raw_data, logger)  # type: ignore[call-arg]
 
     def to_spec(self) -> tmt.steps._RawStepData:
+        """Convert to a form suitable for saving in a specification file"""
```
LecrisUT: Let's be consistent and not docstrings especially if they are inherited.

tmt/guest/__init__.py
```diff
@@ -1325,13 +1326,46 @@ def from_spec(  # type: ignore[override]
...
+        spec['hardware'] = self.hardware.to_spec() if self.hardware else None  # type: ignore[typeddict-unknown-key]
 
         return spec
 
+    def to_minimal_spec(self) -> tmt.steps._RawStepData:
+        """
+        Convert to a form suitable for saving in a specification
+        file while skipping empty values.
+        """
+        spec = {**super().to_minimal_spec()}
```
LecrisUT, line 1340: Why not inherit from `to_spec` and do the cleanup from there? Seems cleaner then the duplication

therazix: Because we need to ensure that some fields also call `to_minimal_spec`. If we inherit from `self.to_spec()`, there may be some fields that were converted using the `to_spec` method instead of `to_minimal_spec`. For example, in this case, the `hardware` field would call `to_spec`, not `to_minimal_spec`.

LecrisUT: :sob:

happz: Yep, nested items and their "unset" fields... I'd very much like to get rid of one of the method chains, maybe the "minimal" can be the one and only, and we'd replace `to_spec` with it, or vice versa. But yes, over time I think we don't need both. I might be wrong though.

tmt/guest/__init__.py
therazix: Removed in a25ecd7de0079c8ead25de0870d2aefd77723b02.

tmt/guest/__init__.py
happz: Added to meeting agenda.

LecrisUT:
Linkcheck failure is a known issue

PR #4729, opened 2026-03-23
Expose step-level queue of phases

happz:
Queues are "buried" inside `go()` methods, but if we want to support addition of new tasks from outside of the step, the queue needs to become step-level instance attribute so it can be modified from outside of `go()`.

Related to #4726

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/queue.py
+++ b/tmt/queue.py
@@ -418,11 +418,19 @@ class Queue(list[TaskT]):
         super().__init__()
 
         self.name = name
-        self._keep_running = True
         self._logger = logger
-
-        self._invoked_tasks = 0
         self._queue_lock = threading.Lock()
+
+        self.reset()
+
+    def reset(self) -> None:
+        """
+        Reset queue content and properties as if it was just created.
+        """
+
+        self[:] = []
+        self._keep_running = True
+        self._invoked_tasks = 0
 
     # We only need to track the head and tail to properly calculate the
     # task queue number dynamically. Adding task number to each task
--- a/tmt/steps/cleanup/__init__.py
+++ b/tmt/steps/cleanup/__init__.py
@@ -149,6 +149,8 @@ class Cleanup(tmt.steps.StepWithQueue[CleanupStepData, PluginOutcome]):
         # expected in the cleanup step
         phases: list[CleanupPlugin[CleanupStepData]] = self.phases(classes=(CleanupPlugin,))
 
+        self._queue.reset()
+
         for phase in phases:
             if phase.enabled_by_when:
                 self._queue.enqueue_plugin(
--- a/tmt/steps/execute/__init__.py
+++ b/tmt/steps/execute/__init__.py
@@ -1228,6 +1228,8 @@ class Execute(tmt.steps.StepWithQueue[ExecuteStepData, None]):
         # Clean up possible old results
         execute_phases[0]._results.clear()
 
+        self._queue.reset()
+
         for phase in self.phases(classes=(Action, ExecutePlugin)):
             if isinstance(phase, Action):
                 self._queue.enqueue_action(phase=phase)
--- a/tmt/steps/finish/__init__.py
+++ b/tmt/steps/finish/__init__.py
@@ -137,6 +137,8 @@ class Finish(tmt.steps.StepWithQueue[FinishStepData, PluginOutcome]):
 
                 guest_copies.append(guest_copy)
 
+            self._queue.reset()
+
             for phase in self.phases(classes=(Action, FinishPlugin)):
                 if isinstance(phase, Action):
                     self._queue.enqueue_action(phase=phase)
--- a/tmt/steps/prepare/__init__.py
+++ b/tmt/steps/prepare/__init__.py
@@ -390,6 +390,8 @@ class Prepare(tmt.steps.StepWithQueue[PrepareStepData, PluginOutcome]):
             # To separate "push" from "prepare" queue visually
             self.info('')
 
+        self._queue.reset()
+
         for prepare_phase in self.phases(classes=(Action, PreparePlugin)):
             if isinstance(prepare_phase, Action):
                 self._queue.enqueue_action(phase=prepare_phase)


tmt/steps/__init__.py
```diff
@@ -1378,6 +1378,21 @@ def prune(self, logger: tmt.log.Logger) -> None:
         prune_directory(self.workdir, preserved_members, self._logger)
 
 
+class StepWithQueue(Step, Generic[StepDataT, PluginReturnValueT]):
```
LecrisUT, line 1385: Could we not make all steps have a queue? Might not be used by some atm, but could they not be converted to use one?

tmt/steps/__init__.py
```diff
@@ -1378,6 +1378,21 @@ def prune(self, logger: tmt.log.Logger) -> None:
...
+    """
+    Common parent of all steps that use queue to run phases in parallel.
+    """
+
+    _queue: 'PhaseQueue[StepDataT, PluginReturnValueT]'
```
LecrisUT, line 1390: Seems fine to make it public

happz: I disagree, I don't want code external to a step have direct access to its queue. For example, `discover` phase adding new `prepare` phases is not just a question of adding them to queue, but also step's `_data` needs to be modified, there should be same message emitted to logs, and so on. I see this action as consisting of multiple subactions, hence step-level API seems more fitting.

Is there a use case for e.g. `discover` accessing `prepare` queue directly? I don't have such use case at hand.

tmt/steps/__init__.py
happz: Yes, we can, but it will open gate to many, many complaints by mypy and pyright. By merging `StepWithQueue` into `Step`, `Step` would become generic, it would need to accept 2 types in all places where it's used, even in unrelated annotations. It's something I would like to do eventually, but the fallout would be huge.

tmt/steps/__init__.py
LecrisUT: In that case, would like this one to own (some?) interface for interacting with the `_queue`

tmt/steps/__init__.py
LecrisUT: Oh I thought it was already generic :thinking:. Maybe can revisit when we can revisit the output `PluginReturnValueT` (making all steps have outcomes)

LecrisUT: Could just add a note on why we need it and when we could move this in `Step` proper?

tmt/steps/__init__.py
happz: I don't expect a `Step`-level API for manipulation of its queue alone. We do have a method that handles the "add a phase to step" task, `add_phase()`, and in the "connect the dots" PR I want to extend it to not just update `self.data`, but also `self._queue` if it's already running. As mentioned above, so far I don't see a use case for code outside of the step to work with the queue as a standalone entity, it seems to me the queue would always be part of something bigger - "add this phase to step", or "run the step", and so on.

Edit: not `self.data`, but `self._phases` is extended with new phase.

tmt/steps/__init__.py
happz: Added in 7586db87.

tmt/steps/__init__.py
happz: Added https://github.com/teemtee/tmt/pull/4750 - a draft - that should be enough for `prepare/artifact` to be able to add a phase dynamically, to the same step, while the step is already running. That should drop the relevant code from `prepare/__init__.py`. Needs tests, but the basic idea is in place.

tmt/queue.py
```diff
@@ -354,25 +397,78 @@ class Queue(list[TaskT]):
...
+                    f'queued {self.name} task #{self._tail_task_number}, caused queue reordering',
+                    task.name,
+                    color='cyan',
+                )
+
+                # PLR1704: redefining `task` - on purpose, parameter is
+                # no longer needed for anything.
+                for task_number, task in enumerate(self, start=self._head_task_number):  # noqa: PLR1704
+                    self._logger.info(
+                        f'queued {self.name} task #{task_number}',
```
tcornell-bus: minor, if reordering happens before invoking any tasks, this could print "task #0", which does not match how we were printing "task #" before.

tmt/steps/__init__.py
```diff
@@ -1378,6 +1378,25 @@ def prune(self, logger: tmt.log.Logger) -> None:
...
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+
+        self._queue = PhaseQueue(
```
tcornell-bus, line 1395: Since this was moved from being freshly made every time in the go() methods to existing in this class, it's persistent. Which could cause problems when a step is re-run, using something like `tmt try`.
Example:
my-plan.fmf:
```
discover:
    how: fmf
prepare:
  - name: first-task
    how: shell
    script: |
      echo "first-task ran"
      /bin/false
  - name: second-task
    how: shell
    script: |
      echo "second-task ran"
execute:
  how: tmt

/test:
    test: 'true'
```
then run: `tmt try fedora@virtual --plan my-plan`
```
    prepare
        
        queued prepare task #2: first-task on default-0
        queued prepare task #3: second-task on default-0
        
        how: install
        name: essential-requires
        order: 30
        where: default-0
        package: 1 package requested
            /usr/bin/flock
            cmd: rpm -q --whatprovides /usr/bin/flock || dnf5 install -y  /usr/bin/flock
            stdout: util-linux-core-2.41.1-17.fc43.x86_64
        
        prepare task #2: first-task on default-0
        how: shell
        name: first-task
        order: 50
        overview: 1 script found
        script:
            echo "first-task ran"
            /bin/false
            cmd:
                set -eo pipefail; echo "first-task ran"
                /bin/false
            stdout: first-task ran
        fail:
            Command 'set -eo pipefail; echo "first-task ran"
            /bin/false
            ' returned 1.

prepare step failed

Cause number 1:

    Command 'set -eo pipefail; echo "first-task ran"
    /bin/false
    ' returned 1.

    # stdout (1 lines)
    first-task ran
    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
first prepare step fails, then change the `bin/false` to `bin/true` and re-run the prepare step:
```
/data2/my-plan/test
    prepare
        
        queued prepare task #4, caused queue reordering: essential-requires on default-0
        queued prepare task #2: essential-requires on default-0
        queued prepare task #3: second-task on default-0
        queued prepare task #5, caused queue reordering: essential-requires on default-0
        queued prepare task #2: essential-requires on default-0
        queued prepare task #3: essential-requires on default-0
        queued prepare task #4: second-task on default-0
        queued prepare task #6: first-task on default-0
        queued prepare task #7: second-task on default-0
        
        prepare task #3: essential-requires on default-0
        how: install
        name: essential-requires
        order: 30
        where: default-0
        package: 1 package requested
            /usr/bin/flock
            cmd: rpm -q --whatprovides /usr/bin/flock || dnf5 install -y  /usr/bin/flock
            stdout: util-linux-core-2.41.1-17.fc43.x86_64
    
```
some weirdness

tmt/queue.py
happz: Hopefully addressed in 2bef4c28.

tmt/steps/__init__.py
happz: Added a queue reset in 32377ab9.

LecrisUT: I am not sure if we should do the reset every time we enter the steps. E.g. what if we run a step until a phase and then we continue?

tcornell-bus: I am now getting an odd buildup of the essential requires task:
```
        queued prepare task #2: essential-requires on default-0
        queued prepare task #3: essential-requires on default-0
        queued prepare task #4: essential-requires on default-0
        queued prepare task #5: first-task on default-0
        queued prepare task #6: second-task on default-0
```
everytime I run 'prepare' again in `tmt try` it adds another. This is with the failing prepare step.

happz: > I am now getting an odd buildup of the essential requires task:
> 
> ```
>         queued prepare task #1: essential-requires on default-0
>         queued prepare task #2: essential-requires on default-0
>         queued prepare task #3: essential-requires on default-0
>         queued prepare task #4: essential-requires on default-0
>         queued prepare task #5: first-task on default-0
>         queued prepare task #6: second-task on default-0
> ```
> 
> everytime I run 'prepare' again in `tmt try` it adds another. This is with the failing prepare step.

Yeah, but that's not a regression in my patch, exactly the same happens with `main` branch :)

When `Prepare.go()` runs, it collects essential requirements and test requirements, and adds phases to `Prepare._phases` to install them. When `tmt try` reruns `prepare` step, `Prepare.go()` is invoked and adds those phases once again. `Prepare._phases` is not reset, and `Prepare.go()` is not idempotent in this regard.

Added queue reset helps because where `Prepare.go()` takes `_phases` list, and adds one task for each phase. Queue reset before this step makes sure there are no tasks left from previous `prepare` invocation. But the issue with `Prepare.go()` adding more and more tasks is not affected, and queue will be populated by tasks mirroring growing `_phases`. It's a bug for sure, just not a one added in this PR :)

happz: > I am not sure if we should do the reset every time we enter the steps. E.g. what if we run a step until a phase and then we continue?

In that case yes, it would be a problem. But steps are not interleavable in this way, they are barely restartable as just demonstrated above, so way more things would need to that.

The reset applies only when step is started, by a plan or `tmt try`, not necessarily every time we "enter" the step. I think we could even step aside into another step today, e.g. if we call `Finish.go()` from a `prepare` phase - as long as we don't call `Prepare.go()` once again, we should get back to the same phase and continue with remaining `preparing` tasks. But it's just a theory.

happz: Filed https://github.com/teemtee/tmt/issues/4761

PR #4730, opened 2026-03-23
Cache linkcheck results for the rough duration of the sprint

LecrisUT:
Seems like we do get linkcheck failure roughly every day due to freedesktop. We could put it explicitly in the ignore list, but it recovers in the subsequent run, so would prefer not to.

On the other hand we would not be addressing these issues on a daily basis, so should be fine to check up on the state every sprint instead.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -331,7 +331,7 @@ linkcheck_request_headers = {
 }
 
 # custom linkcheck cache variables
-linkcheck_cache_period = 1.0
+linkcheck_cache_period = 14.0
 
 # -- Options for sphinx_reredirects ---------------------------------------
 


docs/conf.py
```diff
@@ -325,7 +325,7 @@ def _load_theme(theme_package_name: str, theme_name: str) -> bool:
 }
 
 # custom linkcheck cache variables
-linkcheck_cache_period = 1.0
+linkcheck_cache_period = 14.0
```
psss, line 334: Makes sense. Maybe... what about refreshing the cache once per week? In this way we would get two real attempts per sprint.

psss: Btw, should this help also with the ReportPortal failures?

    (releases/index: line  306) broken
    https://reportportal.io/docs/log-data-in-reportportal/ImportDataToReportPortal/
    - 403 Client Error: Forbidden for url:
    https://reportportal.io/docs/log-data-in-reportportal/ImportDataToReportPortal/

LecrisUT: > Makes sense. Maybe... what about refreshing the cache once per week? In this way we would get two real attempts per sprint.

Fine with either. I put it as once per sprint since I would think the latter half would be more chaotic for the release lead compared to the beginning. But I think the github caching is also getting in the way here.

> Btw, should this help also with the ReportPortal failures?
> 
> ```
> (releases/index: line  306) broken
> https://reportportal.io/docs/log-data-in-reportportal/ImportDataToReportPortal/
> - 403 Client Error: Forbidden for url:
> https://reportportal.io/docs/log-data-in-reportportal/ImportDataToReportPortal/
> ```

Unfortunately not. Once the cached result expires this will continue to complain. It can silent it for this sprint I believe since that one started somewhere in the middle of the current sprint

vaibhavdaren: @LecrisUT This cloudfare waf, bot protected domain. I think we would need to put in ignore.

LecrisUT: It is `cloudfront` (AWS) not cloudflare, and it is already resolved https://github.com/teemtee/tmt/pull/4745. Please rebase.

thrix: Looks resolved.

PR #4731, opened 2026-03-23
Resort queue after adding new task

happz:
Adding concept of ordering to queue tasks. Most of the time, tasks will inherit their order from phases they represent. This will allow reordering of queue after adding new tasks while the queue is already running - today, the queue is sorted at the beginning, because steps already add tasks in the order of phases they represent.

Related to #4726.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/queue.py
+++ b/tmt/queue.py
@@ -434,7 +434,7 @@ class Queue(list[TaskT]):
 
         For example, for a queue with three tasks, none invoked yet, the
         head task number would be ``1``. After invoking the first task,
-        the head task number would be ``1``.
+        the head task number would be ``2``.
 
         :returns: task number of the first task in the queue, or
             ``None`` if the queue is empty.


tmt/steps/__init__.py
```diff
@@ -3361,7 +3361,10 @@ def enqueue_plugin(
                 f'No guests queued for phase "{phase}". A typo in "where" key?'
             )
 
-        self.enqueue_task(PluginTask(phase, guests, phase._logger))
+        task = PluginTask(phase, guests, phase._logger)
+        task.order = phase.order
```
LecrisUT: Why not move this to `PluginTask` (or also `Task`) constructor?

tmt/queue.py
```diff
@@ -34,6 +36,12 @@ class Task(abc.ABC, Generic[TaskResultT]):
     #: A logger to use for logging events related to the outcome.
     logger: Logger
 
+    #: Order of this task. Follow the semantics of the
+    #: :tmt:story:`/spec/test/order` key, the lower the number, the
+    #: earlier the task runs. Tasks with ``order`` left unset will be
+    #: invoked last, in no guaranteed order.
+    order: Optional[int] = None
```
LecrisUT, line 45: Isn't `phase.order` always defined as an `int`?

tmt/queue.py
```diff
@@ -49,6 +57,47 @@ class Task(abc.ABC, Generic[TaskResultT]):
     def __init__(self, logger: Logger) -> None:
         self.logger = logger
 
+    def __repr__(self) -> str:
+        return f'<{self.__class__.__name__} name={self.name} order={self.order}>'
```
LecrisUT: I don't like how this deviates from `__repr__` being equivalent with the constructor. Where would you expect this to be used?

tmt/queue.py
```diff
@@ -49,6 +57,47 @@ class Task(abc.ABC, Generic[TaskResultT]):
...
+
+    # Magic methods to support queue sorting: sort tasks by their `order`,
+    # `order=None` means "undefined" and comes last.
+    def __gt__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
```
LecrisUT: This should raise instead of return. Otherwise, when do you expect this to be a valid path?

tmt/queue.py
```diff
@@ -49,6 +57,47 @@ class Task(abc.ABC, Generic[TaskResultT]):
...
+
+        if self.order is not None and other.order is not None:
+            return self.order > other.order
+
+        raise NotImplementedError
+
+    def __gte__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return True
```
LecrisUT: How about using `@total_ordering` and define only 2 magic operators?

tmt/queue.py
```diff
@@ -354,25 +403,80 @@ class Queue(list[TaskT]):
...
+            self.append(task)
+            self._task_counter += 1
 
-        self._logger.info(
-            f'queued {self.name} task #{len(self)}',
-            task.name,
-            color='cyan',
-        )
+            self._logger.info(
+                f'queued {self.name} task #{self._task_counter}',
```
LecrisUT: The usage of _task_counter here feels more like a _current_task_number. Wonder if this could actually be set and owned by the task. If you squint in the right way, you can find that you can even show how the queue moved.

tmt/queue.py
```diff
@@ -354,25 +403,80 @@ class Queue(list[TaskT]):
...
+        Index of the first task in the queue, including all completed tasks.
+
+        :raises GeneralError: when accessed without holding the queue
+            lock, :py:attr:`_queue_lock`.
+        """
+
+        if not self._queue_lock.locked():
+            raise GeneralError("Cannot access queued tasks without holding a queue lock.")
+
+        return self._task_counter - len(self)
```
LecrisUT: This is confusing. Maybe the terminology of top of the queue is more fitting?

tmt/queue.py
```diff
@@ -354,25 +403,80 @@ class Queue(list[TaskT]):
...
+
+            self.sort()
+
+            new_order = [task.name for task in self]
+
+            if current_order != new_order:
+                self._logger.info(
+                    f'queued {self.name} changed order',
+                    color='cyan',
+                )
```
LecrisUT: The logging in combination with the previous `queued` feels confusing: `queued (task) -> changed order -> queued (all)`. Why not something like branching to either `queued` or "task caused queue to reshuffle" -> `queued (all)`

tmt/queue.py
happz: Yes, but sometimes the order is not important, push/pull tasks are not backed by any phase and therefore have no specific order. Tasks based on phases shall have meaningful order, but not all tasks will have such luck.

tmt/queue.py
happz: Hmm, probably remnant of debugging, I don't think this is used anywhere in my patch.

tmt/steps/__init__.py
happz: Moved in ee8b075d.

tmt/queue.py
happz: Dropped in ee8b075d.

tmt/queue.py
happz: Addressed in ee8b075d.

tmt/queue.py
happz: Addressed in ee8b075d. Hopefully I got it right, first time using `@total_ordering`.

tmt/queue.py
happz: Changed in ee8b075d, now there's a tracker for invoked tasks, and properties for the numbers assigned to the head/tail tasks. I tried to make them owned by tasks, but then I had to change them after sorting. The number is pretty much for users' convenience, it has no use in the queue implementation, this seems simpler and still fairly clean.

tmt/queue.py
happz: Head/tail after ee8b075d.

tmt/queue.py
happz: Changed in ee8b075d.

tmt/queue.py
LecrisUT: Ok, the docstring clarifies it a bit. Can you narrow-type `order` in `ActionTask` and `PluginTask` explicitly and move the default definition outside of the base class?

tmt/queue.py
```diff
@@ -49,6 +59,39 @@ class Task(abc.ABC, Generic[TaskResultT]):
...
+            return False
+
+        if self.order is None and other.order is not None:
+            return True
+
+        if self.order is not None and other.order is None:
+            return False
+
+        if self.order is not None and other.order is not None:
+            return self.order > other.order
```
LecrisUT, line 92: These test cases would need documentation:

```suggestion
        if self.order is None and other.order is None:
            # Both self and other are unordered: the result does not matter,
            # __eq__ will be true, preventing reordering via `sorted`
            return False

        if self.order is None:
            # self is unordered, other is ordered: self should go at the end
            return True

        if other.order is None:
            # self is ordered, other is unordered: other should go at the end
            return False

        # Both self and other are ordered: respect their value
        return self.order > other.order
```

tmt/queue.py
LecrisUT: Sounds, good, can you just add a note about that near those properties? Like

> we only need to track the head and tail to properly calculate the task queue number dynamically; tracking each task's number would require re-defining the task number whenever it is resorted.

tmt/queue.py
happz: Hmm, I can imagine action and plugin tasks without a specific order, I'd prefer keeping the default where it is.

tmt/queue.py
happz: Added in bd1b0471.

tmt/queue.py
happz: Added in 378562d4

tmt/queue.py
```diff
@@ -354,25 +407,81 @@ class Queue(list[TaskT]):
...
+            else:
+                self._logger.info(
+                    f'queued {self.name} task #{self._tail_task_number}, caused queue reordering',
+                    task.name,
+                    color='cyan',
+                )
+
+                # PLR1704: redefining `task` - on purpose, parameter is
+                # no longer needed for anything.
+                for task_number, task in enumerate(self, start=self._head_task_number):  # noqa: PLR1704
```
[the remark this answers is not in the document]
happz: Mixing number and index.

tmt/queue.py
LecrisUT: Well actually things are complicated. I see that `Phase` always has an `int` order, but we just don't happen to use it in the `Action` case? This smells like a landmine :worried:. Guess it's fine in the current form, but we need to be careful if we ever refactor `ActionTask`

happz: Which we will have to anyway, as actions are handled in a very different way when it comes to on which guests they run. Eventually there should be no difference, just one "phase task".

tmt/steps/__init__.py
```diff
@@ -3322,6 +3322,7 @@ def __init__(
         super().__init__(guests, logger)
 
         self.phase = phase
+        self.order = phase.order
```
tcornell-bus, line 3326: Does `ActionTask` need the same treatment?

LecrisUT: Discussed in a resolved thread: https://github.com/teemtee/tmt/pull/4731#discussion_r3000158825

happz: Right, added in 2bef4c28.

happz: > Discussed in a resolved thread: [#4731 (comment)](https://github.com/teemtee/tmt/pull/4731#discussion_r3000158825)

Yes, and I misunderstood the point there, it was indeed missing from `ActionTask`.

tmt/queue.py
```diff
@@ -382,16 +516,26 @@ def run(self) -> Iterator[TaskT]:
...
+        # and `pop()` are atomic - but also 2. sorting of the queue
+        # after addition, which is not atomic.
         while self:
-            task_index = task_count - len(self)
-            task = self.pop(0)
+            with self._queue_lock:
+                self._invoked_tasks += 1
+
+                task_number = self._head_task_number
+                task = self.pop(0)
```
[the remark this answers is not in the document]
happz: Fixed in 9f4f0338

tmt/steps/__init__.py
LecrisUT: I was worried that this might affect the placement of the `login` action, but it seems that that one is using integer phase as well.

Wait, but now, when exactly do we have a `None` order?

tmt/queue.py
```diff
@@ -354,25 +407,106 @@ class Queue(list[TaskT]):
...
+                    color='cyan',
+                )
+
+            else:
+                # Narrow type: this shall no longer be undefined as we
+                # do have at least one item in the queue.
+                assert self._head_task_number is not None
+
+                self._logger.info(
+                    f'queued {self.name} task #{self._tail_task_number}, caused queue reordering',
```
LecrisUT:
```suggestion
                    f'queuing {self.name} task, caused queue reordering',
```

Hopefully more navigable than
```
        queued prepare task #4, caused queue reordering: essential-requires on default-0
        queued prepare task #2: essential-requires on default-0
        queued prepare task #3: second-task on default-0
        queued prepare task #5, caused queue reordering: essential-requires on default-0
        queued prepare task #2: essential-requires on default-0
        queued prepare task #3: essential-requires on default-0
        queued prepare task #4: second-task on default-0
        queued prepare task #6: first-task on default-0
        queued prepare task #7: second-task on default-0
```

tmt/steps/__init__.py
happz: > Wait, but now, when exactly do we have a `None` order?

IIRC, provisioning task does not need `order`, and push/pull tasks also don't use it.

tmt/queue.py
happz: I tried to improve logging in 445e4a33, and I think I'm out of ideas, everything I tried looks underwhelming, and it tends to introduce - or drop - empty lines where they should not be.

```
        prepare queue: reordering after task default-0 on default-0
            #1: default-0 on default-0
            #2: essential-requires on default-0
        prepare queue: reordering after task default-1 on default-0
            #1: default-0 on default-0
            #2: default-1 on default-0
            #3: essential-requires on default-0
        prepare queue: reordering after task requires on default-0
            #1: default-0 on default-0
            #2: default-1 on default-0
            #3: requires on default-0
            #4: essential-requires on default-0
        queued prepare tasks
            #1: default-0 on default-0
            #2: default-1 on default-0
            #3: requires on default-0
            #4: essential-requires on default-0

        prepare task #1: default-0 on default-0
        how: install
        package: 5 packages requested
            jq
            make
            python3-pip
            python3-pytest
            yq
            cmd: rpm -q --whatprovides jq yq make python3-pip python3-pytest || sudo dnf5 install -y  jq yq make python3-pip python3-pytest
```

tests/unit/test_queue.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def test_reordering(root_logger: Logger, caplog) -> None:
+    """
+    Test whether tasks are correctly ordered.
+    """
+
+    queue: Queue[Task] = Queue('dummy queue', root_logger)
+
+    tasks: Task[None] = [
```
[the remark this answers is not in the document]
happz: Fixed in 445e4a33

happz:
Tests are failing because of https://github.com/teemtee/tmt/pull/4731#discussion_r3002859225 - once the dust settles on logging format, I will fix the tests.

tmt/steps/__init__.py
LecrisUT: Ah, ok. That will be a source of confusion :upside_down_face:

tmt/queue.py
LecrisUT: I guess it is fine. The only thing that would be worth considering is the verbosity level:
- `queued prepare tasks`: fine as `info`
- `reordering after task`: seems more of a verbose output now
  - unless it is done after it is in `running` state

happz: Added some final touches in 05986eda. Logging is less verbose now, just reporting queued tasks, plus the dump after reordering - that one will turn into "only when running" in https://github.com/teemtee/tmt/pull/4729, since this PR has no flag for "queue is running".

tmt/queue.py
```diff
@@ -354,25 +407,113 @@ class Queue(list[TaskT]):
...
+    # would require re-defining it whenever the queue get reordered.
+    @property
+    def _head_task_number(self) -> Optional[int]:
         """
-        Put new task into a queue
+        Task number of the task currently at the beginning of the queue.
+
+        For example, for a queue with three tasks, none invoked yet, the
+        head task number would be ``1``. After invoking the first task,
+        the head task number would be ``1``.
```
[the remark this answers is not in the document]
happz: Fixed in c912ee4b

tmt/queue.py
LecrisUT: Sounds reasonable :+1:

tests/unit/test_queue.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__(logger)
+
+        self._name = name
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def go(self) -> Iterator[Self]:
+        return
```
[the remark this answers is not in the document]
happz: Task is not invoked, it's just a mock task for testing purposes. If we ever need it to be invoked, we would have nicer code to run.

PR #4734, opened 2026-03-23
Restructure rpm types across package_managers

AthreyVinay:
Move `Version`, `RpmVersion` and `Repository` to `tmt.package_managers.rpm`. These types are shared between package managers and artifact providers.  Keeping them in artifact providers created an awkward dependency direction and required `TYPE_CHECKING` hacks in `package_managers/__init__.py` and `dnf.py`

more importantly, I would like `tmt.package_managers.list_packages()` to return a `list[RpmVersion]` instead of `list[str]` and this would be the first step in the process.

### DIFF
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -13,6 +13,16 @@ import tmt.utils
 from tmt.container import container, simple_field
 from tmt.utils import Command, CommandOutput, GeneralError, Path, PrepareError, ShellScript
 
+if TYPE_CHECKING:
+    from tmt._compat.typing import TypeAlias
+    from tmt.guest import Guest
+
+    #: A type of package manager names.
+    GuestPackageManager: TypeAlias = str
+
+    #: A package origin: either an actual repository name or a :class:`SpecialPackageOrigin`.
+    PackageOrigin: TypeAlias = Union[str, 'SpecialPackageOrigin']
+
 
 @container(frozen=True)
 class Version:
@@ -169,17 +179,6 @@ class SpecialPackageOrigin(str, enum.Enum):
     #: Package is installed but its source repository cannot be determined
     #: (e.g. pre-installed in a container image).
     UNKNOWN = '<unknown>'
-
-
-if TYPE_CHECKING:
-    from tmt._compat.typing import TypeAlias
-    from tmt.guest import Guest
-
-    #: A type of package manager names.
-    GuestPackageManager: TypeAlias = str
-
-    #: A package origin: either an actual repository name or a :class:`SpecialPackageOrigin`.
-    PackageOrigin: TypeAlias = Union[str, SpecialPackageOrigin]
 
 
 #
--- a/tmt/steps/prepare/artifact/providers/__init__.py
+++ b/tmt/steps/prepare/artifact/providers/__init__.py
@@ -16,7 +16,6 @@ from tmt.container import container
 from tmt.guest import Guest
 from tmt.package_managers import Repository, Version
 from tmt.plugins import PluginRegistry
-from tmt.utils import GeneralError as GeneralError
 from tmt.utils import Path, ShellScript
 
 


tmt/package_managers/rpm.py
```diff
@@ -0,0 +1,212 @@
...
+        arch = arch_parts[0] if arch_parts else "noarch"
+        parts = nvr_part.rsplit("-", 2)
+        if len(parts) != 3:
+            raise ValueError(
+                f"Invalid RPM filename format: '{filename}'. "
+                f"Expected name-version-release.arch.rpm"
+            )
+        name, version, release = parts
+
+        return cls(name=name, version=version, release=release, arch=arch, epoch=0)
```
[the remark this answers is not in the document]
AthreyVinay: this is just a move - no changes :(

AthreyVinay:
> LGTM. Only change worth discussing is if it should be public or private because all of the other modules represent a real `package_manager`, but this one is more an utility to it

Makes sense - maybe `_rpm.py` ?

happz:
Proposal of slightly different locations:

* `Version` is not RPM-only thing, so `package_managers/__init__.py` would be better.
* Same for `Repository`, also not an RPM-specific concept.
* Could `RpmVersion` live in `package_managers/dnf.py` after this? Seems to be the only remaining, rpm-specific piece.

tmt/package_managers/rpm.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+@container
```
vaibhavdaren: I’m a bit concerned about the domain relationship being modeled here.
repositories typically contain RPMs, but placing Repository in the`rpm.py` suggests the opposite relationship.

AthreyVinay:
> Proposal of slightly different locations:
> 
>     * `Version` is not RPM-only thing, so `package_managers/__init__.py` would be better.
> 
>     * Same for `Repository`, also not an RPM-specific concept.
> 
>     * Could `RpmVersion` live in `package_managers/dnf.py` after this? Seems to be the only remaining, rpm-specific piece.

Partially. `Version` and `Repository` in `package_managers/__init__.py` makes sense as they can apply agnostically to any package manager.

However, having `RpmVersion` in  `package_managers/dnf.py` implies a depdency from  specific manager implementation from unrelated providers. My best guess for it is to stay in `_rpm.py` or if you dont want to introduce a new module file, it can also move to `package_managers/__init__.py`.

LecrisUT:
> - `Version` is not RPM-only thing, so `package_managers/__init__.py` would be better.

Putting stuff that will be inherited in submodules in the parent sounds like a recipe for circular dependency issues. Could it be `_base.py` instead (and re-exported to the parent)?

> However, having `RpmVersion` in `package_managers/dnf.py` implies a depdency from specific manager

I concur. Yes it makes logical sense for it to be linked to `dnf`, but in this context `dnf.py` relates to the packagemanager and engine which is not shared with `rpm_ostree.py` for example. My vote here would be `_rpm.py` for this part

AthreyVinay:
So good to go with a `_base.py` & `_rpm.py`?

happz:
> > * `Version` is not RPM-only thing, so `package_managers/__init__.py` would be better.
> 
> Putting stuff that will be inherited in submodules in the parent sounds like a recipe for circular dependency issues. Could it be `_base.py` instead (and re-exported to the parent)?

Hm, this is already a thing, every single package manager module imports from `tmt/package_managers/__init__.py`. It does not serve as a parent but as the "base" source. I think it's fine to add more classes there for package managers to consume.

```python
# tmt/package_managers/dnf.py:
from tmt.package_managers import (
    FileSystemPath,
    Installable,
    Options,
    Package,
    PackageManager,
    PackageManagerEngine,
    escape_installables,
    provides_package_manager,
)
```

> > However, having `RpmVersion` in `package_managers/dnf.py` implies a depdency from specific manager
> 
> I concur. Yes it makes logical sense for it to be linked to `dnf`, but in this context `dnf.py` relates to the packagemanager and engine which is not shared with `rpm_ostree.py` for example. My vote here would be `_rpm.py` for this part

Agreed, `tmt/package_managers/_rpm.py` sounds good.

LecrisUT:
> Hm, this is already a thing, every single package manager module imports from `tmt/package_managers/__init__.py`. It does not serve as a parent but as the "base" source. I think it's fine to add more classes there for package managers to consume.

Yes, but also how many `import` do we have that cannot be done on the module level and have to be in the functions. It's not a good design imo

happz:
> > Hm, this is already a thing, every single package manager module imports from `tmt/package_managers/__init__.py`. It does not serve as a parent but as the "base" source. I think it's fine to add more classes there for package managers to consume.
> 
> Yes, but also how many `import` do we have that cannot be done on the module level and have to be in the functions. It's not a good design imo

One under `tmt.package_managers`, `from tmt.package_managers.dnf import Dnf5` in `rpm_ostree.py`. So it's probably not a big issue in this particular case.

But I get the point. I just wouldn't introduce such a change in this PR, as package manager implementations already happily import from `__init__.py`, and few more classes won't make it worse. Whether we want to turn `__init__.py` files throughout the code into less populated ones, adding `_base.py` into packages that offer code to modules of said package, that seems like a distinct discussion topic to me. Add it to the hacking session agenda, or file an issue so it can be discussed, and such a refactoring can take its own pace.

tmt/package_managers/rpm.py
AthreyVinay: This should be answered in the follow-up discussions below.

tmt/package_managers/__init__.py
```diff
@@ -12,6 +14,150 @@
 from tmt.utils import Command, CommandOutput, GeneralError, Path, PrepareError, ShellScript
 
 
+@container(frozen=True)
```
happz, line 27: Eeee, I think we missed the positioning: this new addition should be *after* the `if TYPE_CHECKING:` block below, something that we missed in the case of `SpecialPackageOrigin`.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -2,22 +2,22 @@
...
 import tmt.utils
-from tmt._compat.typing import Self, TypeAlias
-from tmt.container import container, simple_field
+from tmt._compat.typing import TypeAlias
+from tmt.container import container
 from tmt.guest import Guest
+from tmt.package_managers import Repository, Version
 from tmt.plugins import PluginRegistry
-from tmt.utils import GeneralError, Path, ShellScript
+from tmt.utils import GeneralError as GeneralError
```
happz: This looks like a reimport to silence unused import, the trick ruff plays with us.

LecrisUT: Maybe setting `__all__` would nulify it? For `ruff` check at least

happz: Could be.

tmt/package_managers/__init__.py
vaibhavdaren: Yup, these two (`Version`  and `Repository`) along with `SpecialPackageOrigin` should be moved down to after `TYPE_CHECKING` block.

LecrisUT: What? No. `if TYPE_CHECKING` should never have been down there. It should always be near the top near the global imports since it is doing the same thing.

Iiuc it was added there just for the usage together with `SpecialPackageOrigin`, but that should have been changed to use a string instead.

happz: Aren't we saying the same thing? That `if TYPE_CHECKING` should come before these classes?

AthreyVinay: Ha ha - agree. Both saying the same :)

LecrisUT: Ok, my brain does not seem to comprehend the difference between move up and move down. Yeah, seems we all on the same page.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: I have removed it as this will be imported as part of the utils module.

AthreyVinay: Addressed here: [2fce056](https://github.com/teemtee/tmt/pull/4734/commits/2fce0560d553ad4b5d5383143144163554983de7)

tmt/package_managers/__init__.py
AthreyVinay: Addressed here: [2fce056](https://github.com/teemtee/tmt/pull/4734/commits/2fce0560d553ad4b5d5383143144163554983de7)

PR #4736, opened 2026-03-23
Change `list_packages` return type from `list[str]` to `list[RpmVersion]`

AthreyVinay:


tests/prepare/artifact/lib/common.sh
LecrisUT: Overlaps with #4723. Can drop it from this one?

AthreyVinay: Makes sense - as this was more of a probable fix for a test failure - shall revert, thanks,

tmt/package_managers/__init__.py
```diff
@@ -361,21 +362,24 @@ def install_repository(self, repository: "Repository") -> CommandOutput:
         """
         return self.guest.execute(self.engine.install_repository(repository))
 
-    def list_packages(self, repository: "Repository") -> list[str]:
+    def list_packages(self, repository: "Repository") -> "list[RpmVersion]":
```
psss: So this is method of `PackageManager`, thus not related specifically to `dnf` and `rpm` packages. Should it return `list[Version]` instead?

AthreyVinay: Addressed: [5854629](https://github.com/teemtee/tmt/pull/4736/commits/58546295d056513babdd2637ae907d47e9f7f4b5)

psss: Thanks, but the implementation which is closely related to `rpm` remained:

```yaml
return [RpmVersion.from_nevra(line) for line in stripped_lines if line]
```

Shouldn't this go under `Dnf` package manager where all rpm-related stuff is to be handled?

AthreyVinay: Thanks, addressed here: [4e0aa9a](https://github.com/teemtee/tmt/pull/4736/commits/4e0aa9aea623d6dfe3461c7c6943fc57ab9252c6). Looking at the fails.

tmt/steps/prepare/artifact/providers/__init__.py
psss: Can be dropped as reported by `pre-commit`.

psss: Done in 6a880936.

thrix:
> Looks good, except for the rebasing issues, will follow up again after #4703 and #4723 are resolved.
> 
> About the copr repository being tested, these should be changed with a bespoke copr project for these tests.

@LecrisUT so is this blocked on #4723?

LecrisUT:
> > Looks good, except for the rebasing issues, will follow up again after #4703 and #4723 are resolved.
> > About the copr repository being tested, these should be changed with a bespoke copr project for these tests.
> 
> @LecrisUT so is this blocked on #4723?

Related to https://github.com/teemtee/tmt/pull/4736#discussion_r3000845211 that has been resolved. No blockers

PR #4745, opened 2026-03-25
Specify a different User-Agent for pages behind Cloudfront

LecrisUT:
This should fix the linkcheck issue for reportportal


LecrisUT:
Github actions only PR and tests pass, merging

PR #4750, opened 2026-03-25
Phases added dynamically to a step are added to its queue too

happz:
If the queue is running, phases is added there. If the queue is not running, we shall not touch the queue: it's either already done, or it will start at some point in the future, in which case step would add the phase just as any other "static" phase.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/tests/unit/test_steps.py
+++ b/tests/unit/test_steps.py
@@ -99,7 +99,7 @@ def fixture_mocked_queue(
     return [step, step._queue, plugin]
 
 
-def test_add_phase_addeed_to_queue(
+def test_add_phase_added_to_queue(
     mocked_queue: tuple[tmt.steps.StepWithQueue, tmt.queue.Queue, tmt.steps.Plugin],
 ) -> None:
     step, queue, plugin = mocked_queue


tmt/steps/__init__.py
```diff
@@ -1396,6 +1402,55 @@ def __init__(self, **kwargs: Any) -> None:
             self.step_name, self._logger.descend(logger_name=f'{self.step_name}.queue')
         )
 
+    @functools.cached_property
+    def _guest_copies(self) -> list['Guest']:
+        # Create a guest copy and change its parent so that the operations
+        # inside this step's phases on the guest use this step's config
+        # rather than provision step config.
```
LecrisUT: Can make this a docstring?

happz: Yeah, most likely. I also don't like the name, "guest copies" was fine in a method, but here's it's just wrong. So this will see some changes for sure.

tmt/queue.py
```diff
@@ -564,9 +522,13 @@ def run(self) -> Iterator[TaskT]:
...
             # part of https://github.com/teemtee/tmt/issues/4668.
             if failed_tasks:
+                self.is_running = False
+
                 return
 
             if not self._keep_running:
+                self.is_running = False
+
                 return
```
AthreyVinay, line 587: Missing setting `is_running = False` after normal execution?

happz: Fixed in 1d33b157

tmt/steps/__init__.py
happz: Renamed in 0ddd54e3.

tests/unit/test_steps.py
```diff
@@ -51,3 +58,66 @@ def test_assert_feeling_safe_feeling_safe(self):
...
+        step._queue, 'enqueue_plugin', MagicMock(name='<mock>queue.enqueue_plugin')
+    )
+
+    return [step, step._queue, plugin]
+
+
+def test_add_phase_addeed_to_queue(
+    mocked_queue: tuple[
+        tmt.steps.StepWithQueue, tmt.queue.Queue, tmt.steps.Plugin, tmt.steps.Action
+    ],
```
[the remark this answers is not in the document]
happz: Fixed in 6fd141cc.

PR #4759, opened 2026-03-27
Clear the library cache when re-running a tmt try

### LINKED ISSUE #4735: Rerunning tests with libraries broken in tmt-1.69
Running [https://gitlab.com/redhat/centos-stream/tests/kea/-/tree/main/Sanity/dhcp4-network-boot?ref_type=heads](https://gitlab.com/redhat/centos-stream/tests/kea/-/tree/main/Sanity/dhcp4-network-boot) test

with `tmt try rhel10@minute`

First run is OK, but when I choose "t" for retest, the second run fails  with

```
::   import virtual-network library

:: [ 06:17:32 ] :: [  ERROR   ] :: rlImport: Could not find library virtual-network/virtual-network
:: [ 06:17:32 ] :: [   FAIL   ] :: Command 'rlImport virtual-network/virtual-network' (Expected 0, got 1)
:: [ 06:17:32 ] :: [  FATAL   ] :: Import of the library required for the test failed
:: [ 06:17:32 ] :: [   FAIL   ] :: Import of the library required for the test failed (Assert: expected 0, got 1)
::   Duration: 1s
::   Assertions: 0 good, 2 bad
::   RESULT: WARN (import virtual-network library)
```

LecrisUT:
When in tmt try we run `test` the discover workdir directory is being cleared effectively loosing the files that were cached in the beakerlib library.

For a long term solution these cache should be owned by one of the library parents (discover step/plugin). But that requires quite a bit more refactoring including patches for `wow` which is rather difficult to maintain the cross tmt version compatibility, so will postpone it for a more concrete overall refactoring

---

Fixes #4735

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tests/try/basic/test.sh
+++ b/tests/try/basic/test.sh
@@ -44,6 +44,8 @@ rlJournalStart
         rlRun -s "./rerun.exp"
         rlAssertGrep "Let's try.*/plans/rerun" $rlRun_LOG
         rlAssertGrep "pass /tests/library" $rlRun_LOG
+        # Make sure `/tests/library` does not fail on second run
+        rlAssertNotGrep "fail /tests/library" $rlRun_LOG
         rlAssertGrep "Run .* successfully finished. Bye for now!" $rlRun_LOG
     rlPhaseEnd
 


tmt/steps/discover/__init__.py
```diff
@@ -593,6 +593,15 @@ def __init__(
...
         self.extract_tests_later: bool = False
 
+    def _workdir_cleanup(self, path: Optional[Path] = None) -> None:
+        from tmt.libraries.beakerlib import BeakerLib, BeakerLibFromUrl
+
+        super()._workdir_cleanup(path)
+        # TODO: for now we are cleaning up the whole library cache, probably
+        #  another class should own it instead.
+        BeakerLib._library_cache.clear()
+        BeakerLibFromUrl._git_clone_cache.clear()
```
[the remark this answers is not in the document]
LecrisUT: Hell no, there is even a comment of how this is a temporary fixup until refactoring

Also the reference is completely idiotic

tests/try/basic/test.sh
```diff
@@ -40,6 +40,13 @@ rlJournalStart
         rlAssertGrep "Run .* successfully finished. Bye for now!" $rlRun_LOG
     rlPhaseEnd
 
+    rlPhaseStartTest "Local Plan (rerun)"
+        rlRun -s "./rerun.exp"
+        rlAssertGrep "Let's try.*/plans/rerun" $rlRun_LOG
+        rlAssertGrep "pass /tests/library" $rlRun_LOG
+        rlAssertGrep "Run .* successfully finished. Bye for now!" $rlRun_LOG
```
psss, line 49: Let's `rlAssertNotGrep` for `FAIL`, otherwise we would not catch the problem as `tmt try` exits with 0 after a successful session.

LecrisUT: Ah ok. How about `rlAssertNotGrep fail /tests/library` for consistency of what is being checked with the above?

psss: Yeah, sounds good.

LecrisUT: Done, ca76b95

psss: Thanks, now works as expected.

PR #4765, opened 2026-03-30
Convert ArtifactProvider to dataclass

LecrisUT:
Making changes to the constructor is a pain otherwise

### DIFF
--- a/tests/unit/artifact/test_copr_repository.py
+++ b/tests/unit/artifact/test_copr_repository.py
@@ -6,18 +6,6 @@ from tmt.package_managers._rpm import RpmVersion
 
 
 @pytest.mark.parametrize(
-    ("raw_id", "expected"),
-    [
-        ("copr.repository:packit/packit-dev", "packit/packit-dev"),
-        ("copr.repository:@teemtee/stable", "@teemtee/stable"),
-    ],
-)
-def test_valid_repository_id(raw_id, expected, artifact_provider):
-    provider = artifact_provider(raw_id)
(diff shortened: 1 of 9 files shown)


vaibhavdaren (approved):

vaibhavdaren:
Shows test cancelled. Running Again.   The change is LGTM.

/packit retest-failed

tests/unit/artifact/test_copr_repository.py
```diff
@@ -5,18 +5,6 @@
...
-@pytest.mark.parametrize(
-    ("raw_id", "expected"),
-    [
-        ("copr.repository:packit/packit-dev", "packit/packit-dev"),
-        ("copr.repository:@teemtee/stable", "@teemtee/stable"),
-    ],
-)
-def test_valid_repository_id(raw_id, expected, artifact_provider):
-    provider = artifact_provider(raw_id)
-    assert provider.copr_repo == expected
```
tcornell-bus, line 17:
```
assert provider.id == expected
```
Could `copr_repo` var just be renamed to `id` instead of removing the test coverage? Unless this copr_repo specific test is covered in some other way.

LecrisUT: I had it like that before, but felt like this test is no longer testing something significant since it is the same for all other artifact providers, just a different check and a different error. That is already handled by the test below this one.

tcornell-bus (approved):

happz (approved):

Issue #4221, opened 2025-10-21
Follow-up tasks before calling the `artifact` plugin done


happz:
* mark "abstract" methods from package manager and engine as `@abstractmethod`s - will require at least dummy implementations in the rest of package managers

thrix:
From triage meeting: will be splitted into multiple issues later

PR #4773, opened 2026-04-01
Introduce  `guest.download()` to centralise downloads on guest

AthreyVinay:
Added `download(url, destination)` method to Guest - this de-duplicates the `_download_artifact()`  on koji.py, copr_build.py & calls it from with file.py. Also, the repository.py now follows the same pattern as copr_repository.py

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -541,8 +541,8 @@ TMT_DOWNLOAD_ATTEMPTS
     By default, 3 attempts are made.
 
 TMT_DOWNLOAD_INTERVAL
-    Number of seconds to wait between download attempts.
-    By default, it is 5 seconds.
+    Number of seconds to wait before retrying after an unsuccessful
+    attempt to download a file from a URL to guest.
 
 TMT_RETRY_SESSION_RETRIES
     The number of retries for HTTP/HTTPS requests when encountering
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -543,6 +543,7 @@ TMT_DOWNLOAD_ATTEMPTS
 TMT_DOWNLOAD_INTERVAL
     Number of seconds to wait before retrying after an unsuccessful
     attempt to download a file from a URL to guest.
+    By default, 5 seconds.
 
 TMT_RETRY_SESSION_RETRIES
     The number of retries for HTTP/HTTPS requests when encountering
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -536,6 +536,14 @@ TMT_REBOOT_TIMEOUT
     How many seconds to wait for a connection to succeed after
     guest reboot. By default, it is 10 minutes.
 
+TMT_DOWNLOAD_ATTEMPTS
+    Number of attempts to download a file from a URL to guest.
+    By default, 3 attempts are made.
+
+TMT_DOWNLOAD_INTERVAL
+    Number of seconds to wait between download attempts.
+    By default, it is 5 seconds.
+
 TMT_RETRY_SESSION_RETRIES
     The number of retries for HTTP/HTTPS requests when encountering
     retriable errors (such as 503 Service Unavailable). By default,
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -116,9 +116,22 @@ DEFAULT_REBOOT_TIMEOUT: int = 10 * 60
 REBOOT_TIMEOUT: int = configure_constant(DEFAULT_REBOOT_TIMEOUT, 'TMT_REBOOT_TIMEOUT')
 
 #: Number of attempts to download a file from a URL.
-DOWNLOAD_ATTEMPTS = 3
+#: This is the default value tmt would use unless told otherwise.
+DEFAULT_DOWNLOAD_ATTEMPTS: int = 3
+
+#: Number of attempts to download a file from a URL.
+#: This is the effective value, combining the default and optional envvar,
+#: ``TMT_DOWNLOAD_ATTEMPTS``.
+DOWNLOAD_ATTEMPTS: int = configure_constant(DEFAULT_DOWNLOAD_ATTEMPTS, 'TMT_DOWNLOAD_ATTEMPTS')
+
 #: Number of seconds to wait between download attempts.
-DOWNLOAD_INTERVAL = 5
+#: This is the default value tmt would use unless told otherwise.
+DEFAULT_DOWNLOAD_INTERVAL: int = 5
+
+#: Number of seconds to wait between download attempts.
+#: This is the effective value, combining the default and optional envvar,
+#: ``TMT_DOWNLOAD_INTERVAL``.
+DOWNLOAD_INTERVAL: int = configure_constant(DEFAULT_DOWNLOAD_INTERVAL, 'TMT_DOWNLOAD_INTERVAL')
 
 
 def default_reboot_waiting() -> Waiting:
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -115,6 +115,11 @@ DEFAULT_REBOOT_TIMEOUT: int = 10 * 60
 #: ``TMT_REBOOT_TIMEOUT``.
 REBOOT_TIMEOUT: int = configure_constant(DEFAULT_REBOOT_TIMEOUT, 'TMT_REBOOT_TIMEOUT')
 
+#: Number of attempts to download a file from a URL.
+DOWNLOAD_ATTEMPTS = 3
+#: Number of seconds to wait between download attempts.
+DOWNLOAD_INTERVAL = 5
+
 
 def default_reboot_waiting() -> Waiting:
     """
@@ -2399,19 +2404,24 @@ class Guest(
 
         :param url: URL to download from.
         :param destination: path on the guest to save the file to.
-        :raises GeneralError: if the download fails.
+        :raises DownloadError: if the download fails after all attempts.
         """
 
-        try:
-            self.execute(
-                ShellScript(
-                    f"{self.facts.sudo_prefix} curl -L --fail"
-                    f" -o {quote(str(destination))} {quote(url)}"
-                ),
-                silent=True,
-            )
-        except tmt.utils.RunError as error:
-            raise DownloadError(f"Failed to download '{url}' to '{destination}'.") from error
+        def _do_download() -> None:
+            try:
+                self.execute(
+                    ShellScript(
+                        f"{self.facts.sudo_prefix} curl -L --fail"
+                        f" -o {quote(str(destination))} {quote(url)}"
+                    ),
+                    silent=True,
+                )
+            except tmt.utils.RunError as error:
+                raise DownloadError(f"Failed to download '{url}' to '{destination}'.") from error
+
+        tmt.utils.retry(
+            _do_download, DOWNLOAD_ATTEMPTS, DOWNLOAD_INTERVAL, f"Download '{url}'", self._logger
+        )
 
     @abc.abstractmethod
     def stop(self) -> None:
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -2398,7 +2398,10 @@ class Guest(
 
         try:
             self.execute(
-                ShellScript(f"curl -L --fail -o {quote(str(destination))} {quote(url)}"),
+                ShellScript(
+                    f"{self.facts.sudo_prefix} curl -L --fail"
+                    f" -o {quote(str(destination))} {quote(url)}"
+                ),
                 silent=True,
             )
         except tmt.utils.RunError as error:
--- a/tmt/guest/__init__.py
+++ b/tmt/guest/__init__.py
@@ -133,37 +133,6 @@ DEFAULT_DOWNLOAD_INTERVAL: int = 5
 #: ``TMT_DOWNLOAD_INTERVAL``.
 DOWNLOAD_INTERVAL: int = configure_constant(DEFAULT_DOWNLOAD_INTERVAL, 'TMT_DOWNLOAD_INTERVAL')
 
-#: Seconds curl waits for the connection phase to succeed during a download.
-#: This is the default value tmt would use unless told otherwise.
-DEFAULT_DOWNLOAD_CONNECT_TIMEOUT: int = 30
-
-#: Seconds curl waits for the connection phase to succeed during a download.
-#: This is the effective value, combining the default and optional envvar,
-#: ``TMT_DOWNLOAD_CONNECT_TIMEOUT``.
-DOWNLOAD_CONNECT_TIMEOUT: int = configure_constant(
-    DEFAULT_DOWNLOAD_CONNECT_TIMEOUT, 'TMT_DOWNLOAD_CONNECT_TIMEOUT'
-)
-#: Minimum transfer rate (bytes/sec) below which a download is considered stalled.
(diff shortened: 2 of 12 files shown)


tmt/steps/prepare/artifact/providers/repository.py
```diff
@@ -76,18 +76,23 @@ def fetch_contents(
...
+            repo_filename = parsed_path.name
+            remote_path = Path(f"/etc/yum.repos.d/{repo_filename}")
+            guest.download(self.id, remote_path)
+            try:
+                output = guest.execute(tmt.utils.Command("cat", remote_path))
+            except tmt.utils.RunError as error:
+                raise PrepareError(f"Failed to read '{repo_filename}' from the guest.") from error
+            self.repository = Repository.from_content(
+                output.stdout or '', repo_filename.removesuffix('.repo'), self.logger
+            )
```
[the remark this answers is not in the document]
AthreyVinay: This is fine - it needs to go there.

tmt/guest/__init__.py
```diff
@@ -2374,6 +2374,28 @@ def pull(
...
+
+        try:
+            self.execute(
+                ShellScript(
+                    f"{self.facts.sudo_prefix} curl -L --fail"
+                    f" -o {quote(str(destination))} {quote(url)}"
+                ),
+                silent=True,
+            )
+        except tmt.utils.RunError as error:
```
LecrisUT, line 2431: Can also move `DownloadError` to be owned here or in `utils`.

tmt/steps/prepare/artifact/providers/repository.py
```diff
@@ -76,18 +77,28 @@ def fetch_contents(
         self.logger.info(f"Initializing repository provider with URL: {self.id}")
 
         parsed = urlparse(self.id)
+        parsed_path = Path(parsed.path)
```
LecrisUT, line 79: `parsed` is not used anywhere else, might as well unify it

LecrisUT (approved):

tmt/steps/prepare/artifact/providers/repository.py
AthreyVinay: It is indeed used here: 
https://github.com/teemtee/tmt/blob/4107506b4f353fa23a482cf424b7cb547e951d1c/tmt/steps/prepare/artifact/providers/repository.py#L81

or am I missing something?

tmt/guest/__init__.py
AthreyVinay: Done: [ac65764](https://github.com/teemtee/tmt/pull/4773/commits/ac657640d0d97d993cbf0355a6508109035781cd)

tmt/steps/prepare/artifact/providers/repository.py
LecrisUT: Ah indeed, am blind

tmt/guest/__init__.py
```diff
@@ -2374,6 +2380,26 @@ def pull(
 
         raise NotImplementedError
 
+    def download(self, url: str, destination: Path) -> None:
```
vaibhavdaren, line 2414: Let’s discuss the possibility of adding retry and backoff logic to the download function.

While running the integration tests, I noticed that Koji URLs would frequently time out. Introducing a retry mechanism with exponential backoff could make the process more resilient and reduce transient failures.

LecrisUT: :+1: but would propose for a different PR

happz: It could be done in this PR, we already have `tmt.utils.retry`, the change would not be that big: wrap the `try` below into a nested function so it could be called, define attempts and interval numbers, and that should be enough.

LecrisUT: I was thinking that it would have to be embedded in the script to satisfy image-mode, but if we do not try to address that part yet, no objections with the `tmt.utils.retry` approach.

AthreyVinay: Yes, the idea was to do this a a two-part PR. But as noted above, it can be a small change in the same one. This was the TODO task: Add a Guest method which would "download file(s) from a given URL" - we have plenty of curl calls, let's have a helper with retries and timeouts ready from this issue: #4221

AthreyVinay: Implemented here: [dcd436c](https://github.com/teemtee/tmt/pull/4773/commits/dcd436cb25fc99a3b2a100164939bd33fccb9bfe)

tmt/guest/__init__.py
```diff
@@ -2401,19 +2406,24 @@ def download(self, url: str, destination: Path) -> None:
...
+            try:
+                self.execute(
+                    ShellScript(
+                        f"{self.facts.sudo_prefix} curl -L --fail"
+                        f" -o {quote(str(destination))} {quote(url)}"
+                    ),
+                    silent=True,
+                )
+            except tmt.utils.RunError as error:
+                raise DownloadError(f"Failed to download '{url}' to '{destination}'.") from error
```
LecrisUT: Uph, move the `try ... except` outside please (covering the `tmt.utils.retry` instead)

AthreyVinay: Done.

AthreyVinay:
@copilot resolve the merge conflicts in this pull request

tmt/guest/__init__.py
```diff
@@ -114,6 +114,11 @@ def default_connect_waiting() -> Waiting:
 #: ``TMT_REBOOT_TIMEOUT``.
 REBOOT_TIMEOUT: int = configure_constant(DEFAULT_REBOOT_TIMEOUT, 'TMT_REBOOT_TIMEOUT')
 
+#: Number of attempts to download a file from a URL.
+DOWNLOAD_ATTEMPTS = 3
+#: Number of seconds to wait between download attempts.
+DOWNLOAD_INTERVAL = 5
```
happz: I strongly recommend to allow configuration of these values. Have a `DEFAULT_*` to hold the default values, and use `configure_constant` with environment variable to allow customization. Don't forget to update the docs in `overview.rst`.

tmt/guest/__init__.py
```diff
@@ -2389,6 +2400,35 @@ def pull(
...
+            self.execute(
+                ShellScript(
+                    f"{self.facts.sudo_prefix} curl -L --fail"
+                    f" -o {quote(str(destination))} {quote(url)}"
+                ),
+                silent=True,
+            )
+
+        try:
+            tmt.utils.retry(
```
vaibhavdaren: [Question] Do we want to retry all errors can we make this smarter and retry only retryable errors somehow?

tmt/guest/__init__.py
```diff
@@ -2389,6 +2400,35 @@ def pull(
...
+
+        :param url: URL to download from.
+        :param destination: path on the guest to save the file to.
+        :raises DownloadError: if the download fails after all attempts.
+        """
+
+        def _do_download() -> None:
+            self.execute(
+                ShellScript(
+                    f"{self.facts.sudo_prefix} curl -L --fail"
```
vaibhavdaren, line 2424: I think we can probably use `--connect-timeout`  and `--max-time` args with the curl command to make it bound. Finish it in a set time.

vaibhavdaren (commented):
Thanks for refactor!! Added two suggestions for resilience.

tmt/guest/__init__.py
AthreyVinay: Done: [address comments...](https://github.com/teemtee/tmt/pull/4773/commits/f1be83f1c6195f7ed15217232f6a589307cc2514)

AthreyVinay:
> Added two suggestions for resilience.

Both fair points - Im just wondering.... why not just use `curl`s out-of-the-box `retry/timeout` flags rather than wrapping it in `tmt.utils.retry()` which kinda solves both the concerns -- 1. Only transient errors are retried on & 2. the bounds can be set to. WDYT @happz as `tmt.utils.retry()` was suggested by you?

tmt/guest/__init__.py
LecrisUT: Max time might be tricky. There are some rpms and srpms that are multiple gigs

AthreyVinay: Nice - what about using these flags instead?
 `--speed-time <seconds>` — https://curl.se/docs/manpage.html#-y
 `--speed-limit <speed>` — https://curl.se/docs/manpage.html#-Y
 
 Setting the time and speed might get over this blocker.

AthreyVinay: An attempt to demo the above: [47194e6](https://github.com/teemtee/tmt/pull/4773/commits/47194e6588087cdc49273973dc0e1afab8be3777)

happz: I don't think we should deal with time/speed limits at the moment. First, so far no user reported an issue with how fast is curl fetching files, or how long it takes to fetch them. Second, it would be problematic code-wise: `Guest.download()` seems to be invoked under the hood of `prepare/artifact` phases, and these phases have no time limit, unlike tests and their `duration`. Also, we don't impose any such limits on package manager actions, for example, we don't tell `dnf install` to finish until some deadline.

I think it's a good idea to have `duration` for `prepare` phases, from `ansible` to `shell`; `Guest.download` would then be given a `Deadline` instance from which it could derive the values for `--connect-timeout` and other options. Without such a overall solution I would not impose speed or time limits on curl.

AthreyVinay: This looks like it deserves a separate PR of it own.... no? I plan to unblock this.

LecrisUT: From the curl documentation
>        --retry <num>
>              If a transient error is returned when curl tries to perform a transfer, it retries this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default). Tran‐
>              sient error means either: a timeout, an FTP 4xx response code or an HTTP 408, 429, 500, 502, 503 or 504 response code.

I would say if we implement it, we just use the `--retry` in the default form and that's it. Anything beyond that is out-of-scope for this PR

AthreyVinay: Agreed - precisely what I was about to do too.

happz: > This looks like it deserves a separate PR of it own.... no? I plan to unblock this.

Yes, absolutely, hence me trying to make this PR simpler by not including any speed or time constraints.

AthreyVinay: Done: [revert to only having curl retry...](https://github.com/teemtee/tmt/pull/4773/commits/1c134dea4afeabbd5bf203230073ad68adf8e58d)

tmt/guest/__init__.py
AthreyVinay: [revert to only having curl retry...](https://github.com/teemtee/tmt/pull/4773/commits/1c134dea4afeabbd5bf203230073ad68adf8e58d)

vaibhavdaren (approved):

happz:
> > Added two suggestions for resilience.
> 
> Both fair points - Im just wondering.... why not just use `curl`s out-of-the-box `retry/timeout` flags rather than wrapping it in `tmt.utils.retry()` which kinda solves both the concerns -- 1. Only transient errors are retried on & 2. the bounds can be set to. WDYT @happz as `tmt.utils.retry()` was suggested by you?

Both would work, I don't mind using `curl`'s native controls.

docs/overview.rst
```diff
@@ -536,6 +536,14 @@ TMT_REBOOT_TIMEOUT
     How many seconds to wait for a connection to succeed after
     guest reboot. By default, it is 10 minutes.
 
+TMT_DOWNLOAD_ATTEMPTS
+    Number of attempts to download a file from a URL to guest.
+    By default, 3 attempts are made.
+
+TMT_DOWNLOAD_INTERVAL
+    Number of seconds to wait between download attempts.
```
happz: "Number of seconds to wait before retrying after an unsuccessful attempt to download a file from a URL to guest."

tmt/guest/__init__.py
```diff
@@ -2390,6 +2414,26 @@ def pull(
...
+        Download a file from a URL to a path on the guest.
+        :param url: URL to download from.
+        :param destination: path on the guest to save the file to.
+        :raises DownloadError: if the download fails after all attempts.
+        """
+        try:
+            self.execute(
+                ShellScript(
+                    f"{self.facts.sudo_prefix} curl -L --fail"
+                    f" --retry {DOWNLOAD_ATTEMPTS - 1}"
```
psss: Perhaps we could make sure that `DOWNLOAD_ATTEMPTS` is always greater than `0`? Not a big deal though.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -147,18 +141,22 @@ def artifacts(self) -> Sequence[ArtifactInfo]:
...
         :param guest: the guest on which the artifact should be downloaded.
         :param destination: path into which the artifact should be downloaded.
+        :raises DownloadError: if the download fails.
         """
 
-        raise NotImplementedError
+        try:
+            guest.download(artifact.location, destination)
+        except tmt.utils.GeneralError as error:
+            raise DownloadError(f"Failed to download '{artifact}'.") from error
```
psss: Maybe mention the `guest` as well in the error message? Btw, isn't the `DownloadError` already fully covered in `guest.download()`? Do we need to re-raise it here?

tmt/guest/__init__.py
```diff
@@ -2390,6 +2414,26 @@ def pull(
...
+                    f" --retry-delay {DOWNLOAD_INTERVAL}"
```
psss, line 2426: Shall we possibly include `--retry-connrefused` or `--retry-all-errors` as well? To catch connection refused (and similar)?

docs/overview.rst
psss: Yeah, this sounds like a more precise definition.

tmt/steps/prepare/artifact/providers/__init__.py
AthreyVinay: Maybe just this: [drop the wrap](https://github.com/teemtee/tmt/pull/4773/commits/28eeb7c5e18ace02190afb65d813eee1af70434a)

We would still not what failed by the `url` and thew guest name should be shown in the log?

docs/overview.rst
AthreyVinay: Updated here: [390a5f0](https://github.com/teemtee/tmt/pull/4773/commits/390a5f054e2cda08f014c2b76ae84ac95e0a905a)

tmt/guest/__init__.py
AthreyVinay: Done: [aebd55f](https://github.com/teemtee/tmt/pull/4773/commits/aebd55f7e23ef454a2ac24b1004d8e6739da1ed6)

tmt/guest/__init__.py
AthreyVinay: Yep, this was somewhat discussed here to use `--retry` in a default form for this pr here: https://github.com/teemtee/tmt/pull/4773#discussion_r3160606557. A discussion followed by a pr should tackle this.

However, `--retry-all-errors` is [what I decided against](https://github.com/teemtee/tmt/pull/4773#issuecomment-4338018794) as we do not want it to retry against `4xx` errors.

psss: Ok.

psss (approved):

happz (approved):

PR #215, opened 2020-04-21
Add 'link' as a core attribute

lukaszachy:

PR #4792, opened 2026-04-08
Remove redundant `install_*` methods

AthreyVinay:
originated from a comment here: https://github.com/teemtee/tmt/issues/4605#issuecomment-4031054446. Good to align before artifacts call various install methods.

also, a slight functional change in `RpmOstree.sort_packages()` which now batches the presence check.

### DIFF
--- a/tmt/package_managers/bootc.py
+++ b/tmt/package_managers/bootc.py
@@ -311,7 +311,6 @@ class Bootc(PackageManager[BootcEngine]):
 
         if missing_installables:
             self.engine.install(*missing_installables, options=options)
-            return self.build_container() or CommandOutput(stdout=None, stderr=None)
 
         return CommandOutput(stdout=None, stderr=None)
 
@@ -322,7 +321,7 @@ class Bootc(PackageManager[BootcEngine]):
     ) -> CommandOutput:
         self.engine.reinstall(*installables, options=options)
 
-        return self.build_container() or CommandOutput(stdout=None, stderr=None)
+        return CommandOutput(stdout=None, stderr=None)
 
     def install_debuginfo(
         self,
@@ -330,35 +329,6 @@ class Bootc(PackageManager[BootcEngine]):
         options: Optional[Options] = None,
     ) -> CommandOutput:
         self.engine.install_debuginfo(*installables, options=options)
-        return CommandOutput(stdout=None, stderr=None)
-
-    def install_from_repository(
-        self,
-        *installables: Installable,
-        options: Optional[Options] = None,
-    ) -> CommandOutput:
-
-        # Check presence to avoid unnecessary container rebuilds
-        presence = self.check_presence(*installables)
-
-        missing_installables = {i for i, present in presence.items() if not present}
-        if missing_installables:
-            self.engine.install(*missing_installables, options=options)
-
-        return CommandOutput(stdout=None, stderr=None)
-
-    def install_from_url(
-        self,
-        *installables: Installable,
-        options: Optional[Options] = None,
-    ) -> CommandOutput:
-
-        presence = self.check_presence(*installables)
-
-        missing_installables = {i for i, present in presence.items() if not present}
-        if missing_installables:
-            self.engine.install(*missing_installables, options=options)
-
         return CommandOutput(stdout=None, stderr=None)
 
     def install_local(
--- a/tmt/package_managers/rpm_ostree.py
+++ b/tmt/package_managers/rpm_ostree.py
@@ -247,7 +247,7 @@ class RpmOstree(PackageManager[RpmOstreeEngine]):
                 self.warn(f"Unable to install recommended package '{package}'.")
 
         if required:
-            return super().install(*required)
+            return super().install(*required, options=options)
 
         return CommandOutput(stdout=None, stderr=None)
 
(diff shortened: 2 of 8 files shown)


LecrisUT (approved):
Sure, why not.

tmt/package_managers/bootc.py
```diff
@@ -311,7 +311,6 @@ def install(
 
         if missing_installables:
             self.engine.install(*missing_installables, options=options)
-            return self.build_container() or CommandOutput(stdout=None, stderr=None)
```
happz, line 314: `self.build_container()` call being dropped is unexpected, what is the reason for it?

AthreyVinay: Reason (if I remember...): `PrepareInstall` called `install_from_repository() `and `install_from_url()`, which didn't call build_container(). The build happened through `finalize_installation() `at the end. Are you hinting at other callers?

tcornell-bus: Asked AI to investigate, it found these other call sites that expect the package to be immediately available, and I agree with its assessment: 

1. tmt/guest/__init__.py in `setup` line ~3303
2. tmt/guest/__init__.py in `_assert_rsync` line ~3489
3. tmt/steps/prepare/distgit.py in `go` line ~215
4. tmt/steps/prepare/feature/crb.py in `_manage_repo` line ~62

tmt/package_managers/rpm_ostree.py
```diff
@@ -239,13 +241,13 @@ def install_from_repository(
...
             try:
-                self.install(package)
+                super().install(package, options=options)
             except RunError as error:
                 self.debug(f"Package installation failed: {error}")
                 self.warn(f"Unable to install recommended package '{package}'.")
 
         if required:
-            return self.install(*required)
+            return super().install(*required)
```
tcornell-bus: should this also get `options` added?

tmt/package_managers/bootc.py
happz: > Reason (if I remember...): `PrepareInstall` called `install_from_repository() `and `install_from_url()`, which didn't call build_container(). The build happened through `finalize_installation() `at the end. Are you hinting at other callers?

I wasn't, it just felt unexpected, maybe an unexpected mistake. But @tcornell-bus found some callers that might be surprised by the change.

LecrisUT: The change looks legit because afaiu the `build_container` is not meant to be called at each an every phase immediately, otherwise we have a big fracturing in the understanding of how this whole workflow works and we need to resync.

Re @tcornell-bus indeed it's worth going over them
- `tmt/guest/__init__.py:setup()#3303`: `self.execute` does not use the output so it seems fine to not run immediately
- `tmt/guest/__init__.py:_assert_rsync()#3489`: Legit issue because we need rsync for now. Best would be to have these in a dedicated list of dependencies. Related to https://github.com/teemtee/tmt/issues/4839
- `tmt/steps/prepare/distgit.py:go()#215`: Complicated and fwict it doesn't work independent of this change because it needs everything to be live when it calls `guest.pull(source_dir)`
- `tmt/steps/prepare/feature/crb.py`: Looks fine for the same reason as the first one

Afaict these are indeed the only usages of `PackageManager.install` (that my IDE found)

tmt/package_managers/rpm_ostree.py
AthreyVinay: Fixed in: [0107b5c](https://github.com/teemtee/tmt/pull/4792/commits/0107b5c859e35dac4a1a0d0f096520ca65647732)

tmt/package_managers/bootc.py
AthreyVinay: Done: [6bad13e](https://github.com/teemtee/tmt/pull/4792/commits/6bad13e490bd8f7d7927dd9dd1c473f51667703e)

tcornell-bus: > - `tmt/guest/__init__.py:setup()#3303`: self.execute does not use the output so it seems fine to not run immediately
> - `tmt/steps/prepare/feature/crb.py`: Looks fine for the same reason as the first one

I am confused; it attempts to execute the just-installed package right after `package_manager.install()` for both of those. By default self.execute() has `immediately = true`, [here](https://github.com/teemtee/tmt/blob/ee7b865b111c6f6d41114905431286ffd07ab807/tmt/guest/__init__.py#L3362), so the command will not be deferred. The self.execute() call will fail due to "command not found", if the provision is bootc. Am I misunderstanding something?

LecrisUT: Oh good catch. Yes, perfectly valid, I thought we had the defaults inverted. Well that opens up yet another can of worms and failure scenarios 🙃. Good luck to whoever handles the https://github.com/teemtee/tmt/pull/4719 follow-up

tcornell-bus: I think unless @AthreyVinay can manually verify that those 4 call spots are not broken when used with bootc they should still be considered a problem and will need to be addressed.  I think if `self.package_manager.finalize_installation()` is added after each of the `install()` calls in those 4 spots then it will guarantee there is no behavior regression with bootc.

AthreyVinay: Done: [ea4781d](https://github.com/teemtee/tmt/pull/4792/commits/ea4781dc8eb161f21f0bb04a1577ef3d05652ea1)

LecrisUT: LGTM

tcornell-bus: LGTM too, I'll let @happz close the thread

tcornell-bus (approved):

therazix (approved):

Issue #4518, opened 2026-01-22
SSH execution fails with "Argument list too long" when environment variables are very large


thrix:
## Problem

When running tests via SSH, tmt fails with the error:

```
[Errno 7] Argument list too long: 'ssh'
```

This occurs when the total size of environment variables exceeds the Linux kernel's `ARG_MAX` limit.

## Root Cause

tmt passes environment variables to the remote host by converting them to shell `export` statements and including them directly in the SSH command line:

https://github.com/teemtee/tmt/blob/7686aaac35cc3d5d2ba95e4faa6a54f1b23db9a8/tmt/steps/provision/__init__.py#L3115-L3117

```python
remote_commands: ShellScript = ShellScript.from_scripts(
    self._prepare_environment(env).to_shell_exports()
)
```

The entire environment is then passed as a single argument to SSH:

```python
ssh_command += [self._ssh_guest, remote_command]
```

When environment variables are very large (e.g., a `SNAPSHOT_b64` variable containing Base64-encoded JSON with 60+ container image references), the total command line exceeds the OS limit and Python's subprocess raises `OSError: [Errno 7] Argument list too long`.

## Reproduction

This was observed in a Testing Farm run where the `SNAPSHOT_b64` environment variable contained a massive Base64-encoded JSON with many container image references:
- Pipeline log: https://artifacts.osci.redhat.com/testing-farm/b7a6ea45-3218-49ee-8802-70c68ffed5f1/pipeline.log

The relevant error from the log:
```
provision
    ...
    fail: [Errno 7] Argument list too long: 'ssh'
    
The exception was caused by 1 earlier exceptions

Cause number 1:

    provision step failed

    The exception was caused by 1 earlier exceptions

    Cause number 1:

        [Errno 7] Argument list too long: 'ssh'
```

## Potential Solutions

This is a fundamental OS-level limitation. Possible approaches to address it:

1. **Write environment to a file and transfer it**: Instead of passing environment in the command line, write exports to a temporary file, `rsync` it to the guest, and `source` it before running commands.

2. **Use SSH's `SendEnv`/`AcceptEnv`**: SSH can forward specific environment variables, but this requires sshd configuration on the guest which may not be available.

3. **Stream environment via stdin**: Pass large environment variables through SSH's stdin instead of command line arguments.

4. **Detect and warn about oversized environments**: Add a check for environment size and provide a clear error message suggesting users reduce the environment size.

[...]

PR #4793, opened 2026-04-08
Avoid `ARG_MAX` overflow in `bootc` image-mode builds

Lorquas:
## Summary

When plan environment contains large values (e.g. base64-encoded SSH keys, certificates, and API tokens from CI systems like Testing Farm), bootc image-mode builds fail with `OSError: [Errno 7] Argument list too long: 'ssh'`. Two patterns combine to hit the OS `ARG_MAX` limit:

1. `collect_command()` inlines the full environment as `export` statements into **every** Containerfile `RUN` directive, multiplying the size
2. `build_container()` passes the entire generated Containerfile as a `cat <<EOF` SSH command-line argument

## Changes

**`tmt/package_managers/bootc.py`**: Transfer the generated Containerfile to the guest via `push()` (rsync/SCP) instead of inlining it in a `cat` heredoc SSH command, removing the command-line size ceiling entirely. Uses `self.guest.guest_workdir / 'Containerfile'` as the local staging path.

Relates to #4518

## Testing

- **Integration test** (`tests/prepare/large-env/`): End-to-end BeakerLib test that dynamically sizes a base64 environment variable based on the system's `MAX_ARG_STRLEN` (`PAGE_SIZE * 32`). The test plan uses two `prepare: how: shell` steps so the full environment is inlined via `collect_command()` into two separate `RUN` directives — making the combined Containerfile exceed `MAX_ARG_STRLEN`. This reliably triggers the old `cat <<EOF` failure while passing with the `push()` fix.
- Verified locally: old code fails with `[Errno 7] Argument list too long: 'ssh'`, new code passes.
- All pre-commit hooks pass (mypy, pyright, ruff, shellcheck, tmt lint, codespell, flake8).

## Pull Request Checklist

* [x] implement the feature
* [x] include a release note


tmt/package_managers/bootc.py
```diff
@@ -265,9 +268,27 @@ def build_container(self) -> Optional[CommandOutput]:
...
                     )
                 )
-                self.guest.execute(
-                    ShellScript(f'cat <<EOF > {containerfile_path!s} \n{containerfile} \nEOF')
-                )
+                # Write Containerfile via push() to avoid exceeding
+                # the OS ARG_MAX limit on the SSH command line.
+                import tempfile
+
+                from tmt.guest import TransferOptions
```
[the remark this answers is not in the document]
Lorquas: Done — moved both `import tempfile` and `from tmt.guest import TransferOptions` to the top-level imports.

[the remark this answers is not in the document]
Lorquas: fixed

tmt/guest/__init__.py
```diff
@@ -2868,11 +2868,25 @@ def collect_command(
...
-            self._prepare_environment(env).to_shell_exports()
-        )
+        self.package_manager.engine.open_containerfile_directives()
+
+        # Emit ENV directives once rather than inlining export statements
+        # into every RUN directive. This avoids exceeding the OS ARG_MAX
+        # limit when plan environment contains large values (e.g. base64-
+        # encoded secrets).
+        environment = self._prepare_environment(env)
+        if environment and not getattr(
```
happz: No `getattr`, please. It loses type awareness, it's prone to typos, and it looks for "private" attribute. Maybe `open_containerfile_directives()` method could gain an optional parameter, `environment`, and the method itself would then take care of emitting `ENV` directives.

happz:
`env` being a parameter of `collect_command` method: is it guaranteed that two calls to `collect_command()`, contributing two commands to the same "session" of future `Containerfile` commands, would have the same `env` content? For example, `collect_command(..., env={'FOO': 'bar'})` followed by `collect_command(..., env={'FOO': 'baz'})` would mean the second command runs with unexpected environment, correct?

Would it make sense to simply `ENV` directives for command's `env` before each command as they are getting collected, to prevent this behaviour?

Is it possible to somehow clear the environment before emitting new directives for new command? I'm thinking about `collect_command(..., env={'DEBUG': '1'})` followed by `collect_command(...)` - the latter command would inherit environment from the first command, running with `DEBUG=1` which was not what the command asked for.

happz:
> ## Summary
> When plan environment contains large values (e.g. base64-encoded SSH keys, certificates, and API tokens from CI systems like Testing Farm), bootc image-mode builds fail with `OSError: [Errno 7] Argument list too long: 'ssh'`. Two patterns combine to hit the OS `ARG_MAX` limit:
> 
> 1. `collect_command()` inlines the full environment as `export` statements into **every** Containerfile `RUN` directive, multiplying the size
> 2. `build_container()` passes the entire generated Containerfile as a `cat <<EOF` SSH command-line argument
> 
> ## Changes
> **Fix A** (`tmt/guest/__init__.py`): Emit plan environment variables as Containerfile `ENV` directives once at the top, rather than repeating `export` statements in every `RUN` directive. This is both smaller and more idiomatic for container builds.

This one is problematic, see https://github.com/teemtee/tmt/pull/4793#issuecomment-4209444396 - as long as `Containerfile` can host multiple commands, collected by a sequence of `collect_command()`, `ENV` can cause commands to see unexpected environment.

> **Fix B** (`tmt/package_managers/bootc.py`): Transfer the generated Containerfile to the guest via `push()` (rsync/SCP) instead of inlining it in a `cat` heredoc SSH command, removing the command-line size ceiling entirely.

This one seems like a no-brainer, we absolutely can and should transfer the file via `push()`. I think you can use `self.workdir / Containerfile` instead of the temporary file, as the guest should be owned by one thread at one point in time, no other thread would be collecting commands or even thinking about rebuilding the guest, so there's no contest over such a path.

Lorquas:
@happz  Thanks for the hints, that simplifies things tremendously.  I tried something like

```
                local_containerfile = self.workdir / 'Containerfile'
                local_containerfile.write_text(containerfile)
```

but it seems like by then workdir never exists because when either `PackageManager` or `PackageManagerEngine` run their `super().__init__(logger=logger)` the `__init__`'s only accept kwargs `guest` and `logger`, not `parent`. So `Common.workdir` returns None if parent is None.  It seems like package managers are kind of detached from the run > plan > step.. workflow?  I could attempt to wire that in but it would touch way more moving parts than I am comfortable handling.

tcornell-bus:
Hi @Lorquas, we would like to propose this for the sprint that starts on May 7th. Would you have the capacity to continue work on this based on your and @happz discussion?

Lorquas:
@tcornell-bus Yes, I've fixed everything except the super()__init__ problem.   I can look at that this week.

Lorquas:
Updated the branch to wire `parent=guest` through the `PackageManager` / `PackageManagerEngine` constructors, so `self.workdir` is available in `BootcEngine.build_container()` — replacing the tempfile approach with `self.workdir` / 'Containerfile' as suggested.

This means all PackageManager instances now technically have access to workdir, not just the bootc engine. This should be safe: `Common.workdir` is lazy and only creates its directory on first access. Since no other package manager code references self.workdir that I could find.

Other Note: `BootcEngine` forwards constructor kwargs to its internal `aux_engine` (e.g. DNF5), so `parent=guest` gets passed there too. The aux engine simply gains an unused parent reference.  Let me know if you want me to strip that out.

tmt/guest/__init__.py
Lorquas: removed.

tmt/package_managers/__init__.py
```diff
@@ -308,8 +308,14 @@ class PackageManagerEngine(tmt.utils.Common):
...
-    def __init__(self, *, guest: 'Guest', logger: tmt.log.Logger) -> None:
-        super().__init__(logger=logger)
+    def __init__(
+        self,
+        *,
+        guest: 'Guest',
+        parent: Optional['tmt.utils.Common'] = None,
+        logger: tmt.log.Logger,
+    ) -> None:
+        super().__init__(parent=parent, logger=logger)
```
LecrisUT: Unnecessary change afaict. And if it was needed, a more narrow typing should be used.

tests/prepare/bootc-large-env/test.sh
```diff
@@ -0,0 +1,41 @@
...
+    rlPhaseStartSetup
+        rlRun "pushd data"
+        rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
+
+        # Generate a ~67KB base64 string simulating an encoded secret.
+        # This exceeds ARG_MAX when multiplied across Containerfile RUN
+        # directives (the bug), but is small enough to pass through
+        # individual SSH execute() calls.
+        rlRun "python3 -c 'import base64; print(\"LARGE_SECRET: \" + base64.b64encode(b\"x\" * 50000).decode())' > large-env.yaml"
+        rlRun "echo \"LARGE_SECRET size: \$(wc -c < large-env.yaml) bytes\""
```
LecrisUT: Use [`rlLogInfo`](https://beakerlib.readthedocs.io/en/latest/manual.html#rlloginfo) instead here.

tests/prepare/bootc-large-env/test.sh
```diff
@@ -0,0 +1,41 @@
...
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "pushd data"
+        rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
+
+        # Generate a ~67KB base64 string simulating an encoded secret.
+        # This exceeds ARG_MAX when multiplied across Containerfile RUN
+        # directives (the bug), but is small enough to pass through
+        # individual SSH execute() calls.
+        rlRun "python3 -c 'import base64; print(\"LARGE_SECRET: \" + base64.b64encode(b\"x\" * 50000).decode())' > large-env.yaml"
```
LecrisUT: This creates the file into `data`, put it somewhere else. Maybe copying `data` to that workdir as well.

tests/prepare/bootc-large-env/test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "pushd data"
+        rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
```
LecrisUT: Please not `/var/tmp/tmt`, we do not do it for any other tests right now.

tests/prepare/bootc-large-env/main.fmf
```diff
@@ -0,0 +1,15 @@
+summary: Test bootc image-mode with large plan environment
+description: |
+    Verify that prepare/shell on bootc image-mode guests succeeds
+    when plan environment contains large values (e.g. base64-encoded
+    secrets). Without the push() fix, the generated Containerfile
+    was passed as a cat heredoc SSH command-line argument, exceeding
+    the OS ARG_MAX limit.
```
LecrisUT: Originally this is a bootc issue, but I don't see an issue to not extend the testing to all other provision states. If it fails for other provisions, then we can skip, mark it as xfail or something in between.

Also we recently reworked some of the tests, please follow the format in
https://github.com/teemtee/tmt/blob/fd655e5e76dc5278109835d7ea8e1110e7667f22/tests/prepare/install/test.sh#L41-L60
It is fine if you just filter all of the images  so that it is run only once.

tests/prepare/bootc-large-env/data/test.fmf
```diff
@@ -0,0 +1,4 @@
+summary: Verify prepare/shell with large environment succeeded
+test: |
+    tree --version
+    bootc status --booted --format humanreadable
```
LecrisUT: The explicit test here is not important right? Please in-line it with `discover/shell` and use something like `/usr/bin/true` instead.

tests/prepare/bootc-large-env/data/plans.fmf
```diff
@@ -0,0 +1,19 @@
...
+/centos-stream-10:
+    provision+:
+        image: https://artifacts.dev.testing-farm.io/images/CentOS-Stream-10-image-mode-x86_64.qcow2
+
+    /large-env:
+        summary: "Prepare/shell with large environment on bootc guest"
+        prepare:
+            how: shell
+            script:
+              - dnf -y install tree
```
LecrisUT: Why the manual dnf install command?

tests/prepare/bootc-large-env/test.sh
happz: Yes, we do, it is needed when tmt interacts with image mode guests, as `/tmp` would not persist across reboots. Rundir under `/tmp` then disappears, and tmt can't pull workdir from the guest.

LecrisUT: Well, we are definitely not consistent on the format then
```bash
        rlRun "run=\$(mktemp -d --tmpdir=/var/tmp/tmt)" 0 "Create run directory"
        rlRun "run=\$(mktemp -d -p /var/tmp)" 0 "Create run directory"
[ "$PROVISION_HOW" = "container" ] && TMP_DIR=${TMT_WORKDIR_ROOT:-/var/tmp} || TMP_DIR=/var/tmp
        rlRun "run=\$(mktemp -d -p $TMP_DIR)" 0 "Create run directory"
        rlRun "run=\$(mktemp -d)" 0 "Create run directory"
```
And it is still hiding the issue on our side that we still haven't addressed this issue.

tmt/package_managers/__init__.py
Lorquas: reverted.

tests/prepare/bootc-large-env/test.sh
Lorquas: done.

tests/prepare/bootc-large-env/data/plans.fmf
Lorquas: This is how i found the bug in the first place, but no need for it here;  switching to `package:`

tests/prepare/bootc-large-env/data/test.fmf
Lorquas: done.

tests/prepare/bootc-large-env/test.sh
Lorquas: Fixed. the generated large-env.yaml is now created after `pushd data`  and cleaned up in the cleanup phase with `rm -f large-env.yaml`.   afaict this is how other tests are doing this?

tests/prepare/bootc-large-env/main.fmf
Lorquas: Thanks for the reference example!   I tried to rework this using that if/else pattern.  I think this should only now trigger on image mode provisions.

tests/prepare/bootc-large-env/data/plans.fmf
```diff
@@ -0,0 +1,22 @@
...
+    how: virtual
+
+execute:
+    how: tmt
+
+/centos-stream-10:
+    provision+:
+        image: https://artifacts.dev.testing-farm.io/images/CentOS-Stream-10-image-mode-x86_64.qcow2
+
+    /large-env:
```
LecrisUT: Can drop these, it is handled in `test.sh` now.

tests/prepare/bootc-large-env/main.fmf
```diff
@@ -0,0 +1,16 @@
+summary: Test bootc image-mode with large plan environment
+description: |
+    Verify that prepare/shell on bootc image-mode guests succeeds
```
LecrisUT: Can drop the `bootc`, `image-mode`, etc. references and put it in a more generic `/tests/prepare/large-env`?

tmt/package_managers/__init__.py
LecrisUT: Still relevant. There are many other places with similar change

tests/prepare/bootc-large-env/test.sh
LecrisUT: It still writes to the source `data` instead of temporary directory. Not a strict blocker, just an unnecessary annoyance to keep track of when things go haywire.

tests/prepare/bootc-large-env/main.fmf
LecrisUT: Cool, can revisit enabling these for the other cases in a subsequent PR. If you could also do a quick check running these for container and virtual that would be much appreciated.

happz:
> Updated the branch to wire `parent=guest` through the `PackageManager` / `PackageManagerEngine` constructors, so `self.workdir` is available in `BootcEngine.build_container()` — replacing the tempfile approach with `self.workdir` / 'Containerfile' as suggested.
> 
> This means all PackageManager instances now technically have access to workdir, not just the bootc engine. This should be safe: `Common.workdir` is lazy and only creates its directory on first access. Since no other package manager code references self.workdir that I could find.
> 
> Other Note: `BootcEngine` forwards constructor kwargs to its internal `aux_engine` (e.g. DNF5), so `parent=guest` gets passed there too. The aux engine simply gains an unused parent reference. Let me know if you want me to strip that out.

I had to think quite hard to recall why `parent` becomes needed, and it seems like an unnecessary complication. And I don't remember whether I proposed it, or whether it's based on your own discovery. Would you mind using `guest.guest_workdir` instead of `self.workdir`? A package manager is tied to the guest, as such it has no workdir of its own, it will be puzzling in the future.

Lorquas:
Updated the branch addressing all feedback and rebased on latest main. Forgive me for not replying in each thread, I figure this is better than the notification spam. 

> I had to think quite hard to recall why parent becomes needed

I also forgot what the original PR even looked like at this point.  Here's what I got so far:

@happz — Switched to `self.guest.guest_workdir / 'Containerfile'` instead of `self.workdir`, and reverted all the `parent=` constructor plumbing from `PackageManager.__init__` and the call sites in `tmt/guest/__init__.py`. 

@LecrisUT —

- `parent=` changes fully reverted
- Renamed `tests/prepare/bootc-large-env/` to `tests/prepare/large-env/`, stripped refrences to bootc/image-mode 
- Flattened plans.fmf and dropped the `/centos-stream-10:` subplan structure since test.sh handles things
- `large-env.yaml` now generated into `$run/` temp directory instead of the source data/ tree

tests/prepare/large-env/test.sh
```diff
@@ -0,0 +1,53 @@
...
+        rlRun "pushd data"
+        rlRun "export TMT_BOOT_TIMEOUT=300"
+        rlRun "export TMT_CONNECT_TIMEOUT=300"
+    rlPhaseEnd
+
+    while IFS= read -r image; do
+        phase_prefix="$(test_phase_prefix "$image")"
+
+        rlPhaseStartTest "$phase_prefix Prepare/install with large environment"
+            rlRun -s "tmt -vvv run --scratch -i \$run --environment-file \$run/large-env.yaml provision --how virtual --image $image"
```
LecrisUT:
```suggestion
            rlRun -s "tmt -vvv run --scratch -i \$run --environment-file \$run/large-env.yaml provision --how $PROVISION_HOW --image $image"
```
🙂

thrix:
Related to #4871 #4875., the fixes there should be revisited on top of this

tmt/package_managers/bootc.py
LecrisUT, line 239: The gymnastics of this temporary dir, and then the `self.guest.guest_workdir / 'Containerfile'` is quite funny 😄. Not a thing that needs to be fixed here though

LecrisUT (approved):
> I figure this is better than the notification spam.

Don't worry about that, we have plenty of notification "spam" we deal on a daily basis :sweat_smile:. Also incremental commits are fine, we do squash commits.

happz:
@Lorquas please, check out the tests, they fail in our CI: https://artifacts.dev.testing-farm.io/2689664c-7b34-4464-a180-d2a7d5f80fc2/work-preparez7vj16yz/plans/provision/virtual/prepare/execute/data/guest/default-0/tests/prepare/large-env-3/output.txt. If I'm not mistaken, the large env file needs to be injected into the tree, tmt has a limitation in place to stick to the tree when it comes to loading files.

happz (approved):

psss:
@Lorquas, could you please update the pull request description so that it can be used as the git commit message when squashing? Currently it contains some outdated stuff like Fix A or unit tests. Could you also add a short [release note](https://github.com/teemtee/tmt/tree/main/docs/releases/pending)? Thanks.

tests/prepare/large-env/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+        elif [ "$PROVISION_HOW" = "container" ]; then
+            rlRun "IMAGES="
+
+        elif [ "$PROVISION_HOW" = "virtual" ]; then
+            rlRun "IMAGES="
+
+        else
+            rlRun "IMAGES="
+        fi
```
psss, line 23: The test is expected to be run only in `image-mode`, so we could get rid of these?

tests/prepare/large-env/main.fmf
```diff
@@ -0,0 +1,15 @@
+summary: Test prepare with large plan environment
+description: |
+    Verify that prepare succeeds when plan environment contains large
+    values (e.g. base64-encoded secrets). Without the push() fix,
+    the generated Containerfile was passed as a cat heredoc SSH
+    command-line argument, exceeding the OS ARG_MAX limit.
+link:
+  - verifies: https://github.com/teemtee/tmt/issues/4518
```
psss: Does the test full verify the issue in question? @thrix, I thought the issue scope was broader. Or does the pull request fix it completely?

tests/prepare/large-env/test.sh
LecrisUT: My previous comment is to add them so that we would enable these for the others. The original issue of big env variable applies to the others as well, and I will want this covered for those as well asap

psss: I see, makes sense. What about including a short `TODO` comment mentioning this plan?

tests/prepare/large-env/test.sh
```diff
@@ -0,0 +1,55 @@
...
+
+        rlRun "run=\$(mktemp -d -p /var/tmp)" 0 "Create run directory"
+        rlRun "pushd data"
+
+        # Generate a ~67KB base64 string simulating an encoded secret.
+        # Must live inside the fmf tree (data/) so tmt accepts it as
+        # an --environment-file source.
+        rlRun "python3 -c 'import base64; print(\"LARGE_SECRET: \" + base64.b64encode(b\"x\" * 50000).decode())' > large-env.yaml"
```
psss: Seems the test does not reproduce the problem. With `50000` it happily passes with the old `tmt` as well, with `100000` it fails for both old and new `tmt`.

psss: After the adjustment test works as expected. Thanks!

psss (approved):

PR #4797, opened 2026-04-13
Remove custom dictionary from `beakerlib` template

psss:
Seems because of CVE-2026-34982 enabling a custom dictionary from the modeline is now forbidden in vim by default as it could present a security risk. Let's remove the line from the test template and our tests as well.

### DIFF
--- a/examples/convert/runtest.sh
+++ b/examples/convert/runtest.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-# vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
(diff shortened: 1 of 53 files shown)


psss:
For reference, here's the related [chat](https://redhat-internal.slack.com/archives/C04LRPNDZ4Y/p1776068975532289).

psss:
@mrc0mmand, do you agree with this approach? I'd say it does not make sense to keep the dictionary path either as without the `k` modifier it has no effect. I guess the recommended way should be add the custom dictionary to the user config.

mrc0mmand:
> @mrc0mmand, do you agree with this approach? I'd say it does not make sense to keep the dictionary path either as without the `k` modifier it has no effect. I guess the recommended way should be add the custom dictionary to the user config.

I definitely agree, there's no point in leaving that in, so this change looks good to me. Thank you!

LecrisUT (approved):

tcornell-bus (approved):

vaibhavdaren (approved):

PR #4798, opened 2026-04-14
Resolve ArtifactInfo.location overloading for verify-installation

### LINKED ISSUE #4714: Link the Artifact Verification Plugin for discover only Providers

vaibhavdaren:
Add a dedicated ``repo`` field to ``ArtifactInfo`` holding the DNF repo_ids for repository providers (``None`` for download providers whose artifacts land in the shared repo). This replaces the workaround in ``_inject_verify_phase()`` that used ``provider.get_repositories()`` as a proxy and stored a comma-joined string of repo_ids, fixing the last-writer-wins bug where the same package in multiple providers would silently drop all but one valid source repo.

The ``verify`` field in ``PrepareVerifyInstallationData`` is changed from ``dict[str, str]`` to ``dict[str, list[str]]`` so that a package can be expected from any of multiple repos (OR semantics). A new ``normalize_string_list_dict`` normalizer coerces single string values to one-element lists, keeping existing fmf plans working without changes.

The ``verify: false`` workaround in the nvr-priority TC 1.1 test is removed as it is no longer needed.

Closes: https://github.com/teemtee/tmt/issues/4714

Assisted-by: Claude

### DIFF
--- a/tmt/package_managers/dnf.py
+++ b/tmt/package_managers/dnf.py
@@ -192,7 +192,7 @@ class DnfEngine(PackageManagerEngine):
                 '--disablerepo=*',
                 *[f'--enablerepo={repo_id}' for repo_id in repository.repo_ids],
                 '--queryformat',
-                r'%{repoid} %{name}-%{epoch}:%{version}-%{release}.%{arch}\n',
+                r'%{repoid}\t%{name}-%{epoch}:%{version}-%{release}.%{arch}\n',
             )
         ).to_script()
 
@@ -262,7 +262,7 @@ class Dnf(PackageManager[DnfEngine]):
             line = line.strip()
             if not line:
                 continue
-            repo_id, _, nevra = line.partition(' ')
+            repo_id, nevra = line.split('\t', maxsplit=1)
             result.append(RpmVersion.from_nevra(nevra, repo_id=repo_id))
         return result
 
--- a/tmt/package_managers/dnf.py
+++ b/tmt/package_managers/dnf.py
@@ -192,7 +192,7 @@ class DnfEngine(PackageManagerEngine):
                 '--disablerepo=*',
                 *[f'--enablerepo={repo_id}' for repo_id in repository.repo_ids],
                 '--queryformat',
-                r'%{repoid}\t%{name}-%{epoch}:%{version}-%{release}.%{arch}\n',
+                r'%{repoid};%{name}-%{epoch}:%{version}-%{release}.%{arch}\n',
             )
         ).to_script()
 
@@ -262,7 +262,7 @@ class Dnf(PackageManager[DnfEngine]):
             line = line.strip()
             if not line:
                 continue
-            repo_id, nevra = line.split('\t', maxsplit=1)
+            repo_id, nevra = line.split(';', maxsplit=1)
             result.append(RpmVersion.from_nevra(nevra, repo_id=repo_id))
         return result
 
--- a/tmt/package_managers/dnf.py
+++ b/tmt/package_managers/dnf.py
@@ -185,12 +185,16 @@ class DnfEngine(PackageManagerEngine):
         )
 
     def list_packages(self, repository: Repository) -> ShellScript:
-        repo_ids = " ".join(f"--enablerepo={repo_id}" for repo_id in repository.repo_ids)
-        return ShellScript(
-            f"""
-            {self.command.to_script()} repoquery --disablerepo='*' {repo_ids}
-            """
-        )
+        return (
+            self.command
+            + Command(
+                'repoquery',
+                '--disablerepo=*',
+                *[f'--enablerepo={repo_id}' for repo_id in repository.repo_ids],
+                '--queryformat',
+                r'%{repoid} %{name}-%{epoch}:%{version}-%{release}.%{arch}\n',
+            )
+        ).to_script()
 
     def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
         return (
@@ -253,8 +257,14 @@ class Dnf(PackageManager[DnfEngine]):
         if stdout is None:
             raise GeneralError("Repository query provided no output")
 
-        stripped_lines = (line.strip() for line in stdout.strip().splitlines())
-        return [RpmVersion.from_nevra(line) for line in stripped_lines if line]
+        result: list[Version] = []
+        for line in stdout.strip().splitlines():
+            line = line.strip()
+            if not line:
+                continue
+            repo_id, _, nevra = line.partition(' ')
+            result.append(RpmVersion.from_nevra(nevra, repo_id=repo_id))
+        return result
 
     def check_presence(self, *installables: Installable) -> dict[Installable, bool]:
         try:
--- a/tmt/steps/prepare/artifact/__init__.py
+++ b/tmt/steps/prepare/artifact/__init__.py
@@ -199,7 +199,6 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
 
     # Shared repository configuration
     SHARED_REPO_DIR_NAME: ClassVar[str] = 'artifact-shared-repo'
-    SHARED_REPO_NAME: ClassVar[str] = SHARED_REPO_NAME
     ARTIFACTS_METADATA_FILENAME: ClassVar[str] = 'artifacts.yaml'
 
     #: Name of the auto-injected verify-installation phase.
@@ -229,7 +228,7 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
             artifact_dir=shared_repo_dir,
             guest=guest,
             logger=logger,
-            repo_name=self.SHARED_REPO_NAME,
+            repo_name=SHARED_REPO_NAME,
             priority=self.data.default_repository_priority,
         )
 
@@ -356,18 +355,15 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
             for pkg in install_phase.data.package:  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
                 pkg_names.add(str(pkg))  # pyright: ignore[reportUnknownArgumentType]
 
-        # Build package → list of valid repo_ids mapping from all providers.
-        pkg_to_repos: dict[str, list[str]] = {}
+        # Build package → set of valid repo_ids mapping from all providers.
+        pkg_to_repos: dict[str, set[str]] = {}
         for provider in providers:
             for artifact in provider.artifacts:
-                existing = pkg_to_repos.setdefault(artifact.version.name, [])
-                existing.extend(
-                    repo_id for repo_id in artifact.repo_ids if repo_id not in existing
-                )
+                pkg_to_repos.setdefault(artifact.version.name, set()).update(artifact.repo_ids)
 
         # Only verify packages that are both required and from a known artifact.
         pkgs_to_verify = {
-            pkg: repo_ids for pkg, repo_ids in pkg_to_repos.items() if pkg in pkg_names
+            pkg: sorted(repo_ids) for pkg, repo_ids in pkg_to_repos.items() if pkg in pkg_names
         }
 
         if not pkgs_to_verify:
--- a/tmt/steps/prepare/artifact/__init__.py
+++ b/tmt/steps/prepare/artifact/__init__.py
@@ -355,11 +355,11 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
             for pkg in install_phase.data.package:  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
                 pkg_names.add(str(pkg))  # pyright: ignore[reportUnknownArgumentType]
 
-        # Build package → set of valid repo_ids mapping from all providers.
+        # Build package → set of valid repo_id mapping from all providers.
         pkg_to_repos: dict[str, set[str]] = {}
         for provider in providers:
             for artifact in provider.artifacts:
-                pkg_to_repos.setdefault(artifact.version.name, set()).update(artifact.repo_ids)
+                pkg_to_repos.setdefault(artifact.version.name, set()).add(artifact.repo_id)
 
         # Only verify packages that are both required and from a known artifact.
         pkgs_to_verify = {
--- a/tmt/steps/prepare/artifact/__init__.py
+++ b/tmt/steps/prepare/artifact/__init__.py
@@ -360,13 +360,15 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
         pkg_to_repos: dict[str, list[str]] = {}
         for provider in providers:
             for artifact in provider.artifacts:
-                repo_ids = artifact.repo_ids
-                pkg_name = artifact.version.name
-                existing = pkg_to_repos.setdefault(pkg_name, [])
-                existing.extend(rid for rid in repo_ids if rid not in existing)
+                existing = pkg_to_repos.setdefault(artifact.version.name, [])
+                existing.extend(
+                    repo_id for repo_id in artifact.repo_ids if repo_id not in existing
+                )
 
         # Only verify packages that are both required and from a known artifact.
-        pkgs_to_verify = {pkg: repos for pkg, repos in pkg_to_repos.items() if pkg in pkg_names}
+        pkgs_to_verify = {
+            pkg: repo_ids for pkg, repo_ids in pkg_to_repos.items() if pkg in pkg_names
+        }
 
         if not pkgs_to_verify:
             self.verbose('No packages to be installed were found in the provided artifacts.')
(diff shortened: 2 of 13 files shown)


tmt/steps/prepare/verify_installation.py
```diff
@@ -115,10 +120,10 @@ def go(
             return outcome
 
         failed_packages: list[str] = []
-        for package, expected_repo in self.data.verify.items():
+        for package, expected_repos in self.data.verify.items():
             actual_origin = package_origins[package]
 
-            if actual_origin in expected_repo:
+            if actual_origin in expected_repos:
```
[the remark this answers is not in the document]
vaibhavdaren: SpecialPackageOrigin is defined as class SpecialPackageOrigin(str, enum.Enum), making its members full str          Instances. The in operator already compares by string value. Hence, no cast needed.

tmt/utils/__init__.py
```diff
@@ -5900,6 +5900,42 @@ def normalize_string_dict(
...
+    if not isinstance(raw_value, dict):
+        raise tmt.utils.NormalizationError(
+            key_address,
+            raw_value,
+            'a dictionary mapping string keys to string values or lists of strings',
+        )
+
+    return {
+        str(k).strip(): (
+            [str(v).strip() for v in val] if isinstance(val, list) else [str(val).strip()]
```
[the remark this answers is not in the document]
vaibhavdaren: Tuples are not a supported input type in this context.

tmt/utils/__init__.py
```diff
@@ -5900,6 +5900,42 @@ def normalize_string_dict(
...
+        )
+        for k, val in raw_value.items()
```
happz: `k, val` is unexpected - `k, v` or `key, value` would be more easily understood. Seeing `k` in the code, I would automatically expect it to be corresponding to `v`, not `val`.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -45,6 +48,11 @@ class ArtifactInfo:
     version: Version
     location: str
     provider: "ArtifactProvider"
+    #: DNF repo section IDs this artifact is available from.
+    #: Download providers set this to ``[ArtifactProvider.SHARED_REPO_NAME]``
+    #: since their RPMs land in the shared repo after being downloaded.
+    #: Repository providers set this to the section IDs from their ``.repo`` file.
```
happz: * No "DNF", please, the message should be package manager-agnostic.
* I would stay away from what various groups of providers store in this field - if that changes in one of them, you would need to go here and make sure this docstring is still accurate. Instead, describe the purpose of this field, and that should be all that's needed here. The note about the shared repo can be a comment in the corresponding providers ("well, we are contributing to the shared repo, artifacts we produce therefore shall have shared repo in their repo ID field").

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,29 +356,17 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
-                )
-                for artifact in provider.artifacts:
-                    pkg_to_repo[artifact.version.name] = all_repo_ids
-            else:
-                # Download provider: packages land in the shared repo.
-                for artifact in provider.artifacts:
-                    pkg_to_repo[artifact.version.name] = self.SHARED_REPO_NAME
+            for artifact in provider.artifacts:
+                repo_ids = artifact.repo_ids
+                pkg_name = artifact.version.name
```
happz: These two variables are used just once. There is no harm in `pkg_to_repos.setdefault(artifact.version.name, [])` and `existing.extend(rid for rid in artifact.repo_ids if rid not in existing)`.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,29 +356,17 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
+                existing = pkg_to_repos.setdefault(pkg_name, [])
+                existing.extend(rid for rid in repo_ids if rid not in existing)
```
happz: `repo_id` instead of `rid`, please. It's an item of `repo_ids` list, and we can afford using more characters to be clearer about value names.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,29 +356,17 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
 
         # Only verify packages that are both required and from a known artifact.
-        pkgs_to_verify = {pkg: repo for pkg, repo in pkg_to_repo.items() if pkg in pkg_names}
+        pkgs_to_verify = {pkg: repos for pkg, repos in pkg_to_repos.items() if pkg in pkg_names}
```
happz: `pkg: repo_ids`?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -397,8 +386,10 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
         )
 
         if existing_verify is not None:
-            # Merge into existing verify phase.
-            existing_verify.data.verify.update(pkgs_to_verify)  # pyright: ignore[reportUnknownMemberType]
+            # Merge into existing verify phase, extending repo lists rather than replacing them.
+            for verify_pkg, verify_repos in pkgs_to_verify.items():
+                existing = existing_verify.data.verify.setdefault(verify_pkg, [])  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
+                existing.extend(r for r in verify_repos if r not in existing)  # pyright: ignore[reportUnknownMemberType]
```
happz: Yeah, `repo_id`, please. The same value has been called `rid` and `r`, and it comes from `repo_ids` field of artifacts, that's inconsistent.

tmt/steps/prepare/verify_installation.py
```diff
@@ -20,17 +20,15 @@ class PrepareVerifyInstallationData(PrepareStepData):
...
-    # and the semantics must be documented: a package passes verification if its
-    # actual source repo matches ANY of the listed repos (OR semantics).
-    verify: dict[str, str] = field(
+    verify: dict[str, list[str]] = field(
         default_factory=dict,
-        help="Mapping of package names to expected source repository names.",
-        normalize=tmt.utils.normalize_string_dict,
+        help="""
+            Mapping of package names to expected source repository names.
+            A package passes verification if its actual source repo matches
```
happz: A package passes verification if it is installed and it was installed from one of the listed repositories.

tmt/steps/prepare/verify_installation.py
```diff
@@ -132,15 +137,15 @@ def go(
                 continue
 
             failed_packages.append(package)
+            expected_str = ' or '.join(f"'{r}'" for r in expected_repos)
```
happz: * `_str`, encoding of type in the variable name - consider calling `fmf.utils.listed()` in the f-string, or prepare it before, but give it a name expressing the purpose, e.g. `expected_repos_formatted` or something similar.
* "repo ID" called `r` this time ;)

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -45,6 +48,11 @@ class ArtifactInfo:
     version: Version
     location: str
     provider: "ArtifactProvider"
+    #: DNF repo section IDs this artifact is available from.
+    #: Download providers set this to ``[ArtifactProvider.SHARED_REPO_NAME]``
```
AthreyVinay: This is incorrect as `ArtifactProvider.SHARED_REPO_NAME` doesn't exist.

tmt/steps/prepare/artifact/providers/copr_build.py
```diff
@@ -220,7 +221,10 @@ def make_rpm_artifact(self, rpm_meta: dict[str, str]) -> ArtifactInfo:
             base_url = self.result_url.rstrip("/")
 
         return ArtifactInfo(
-            version=version_info, location=urljoin(base_url + "/", filename), provider=self
+            version=version_info,
+            location=urljoin(base_url + "/", filename),
+            provider=self,
+            repo_ids=[SHARED_REPO_NAME],
```
AthreyVinay: You could give `repo_ids` a default value of `[SHARED_REPO_NAME] ` on `ArtifactInfo` and then only the repository provider path in enumerate_artifacts would need to explicitly pass it.  This will avoid importing and boilerplating it at 6 places. Also, worth adding a comment to future additions of repo providers to override this.

vaibhavdaren: Thanks for suggestion, looks better. Addressed in 3e8e29e9ff5352b3e26fcbfa47e11c7a001adf87.

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: Addressed in 3e8e29e9ff5352b3e26fcbfa47e11c7a001adf87.

happz (approved):

AthreyVinay (approved):

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -45,6 +48,9 @@ class ArtifactInfo:
     version: Version
     location: str
     provider: "ArtifactProvider"
+    #: Repository IDs this artifact is available from. Used during verification
+    #: to confirm the artifact was installed from an expected repository.
+    repo_ids: list[str] = simple_field(default_factory=lambda: [SHARED_REPO_NAME])
```
```suggestion
    repo_ids: list[str] = simple_field(default=[SHARED_REPO_NAME])
```
Use `default` if you mean to use a default. But also, why is this a list. An artifactInfo coming from a single artifact provider with as specific NVR can only come from one repo

LecrisUT (changes requested):

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -198,7 +199,7 @@ class PrepareArtifact(PreparePlugin[PrepareArtifactData]):
 
     # Shared repository configuration
     SHARED_REPO_DIR_NAME: ClassVar[str] = 'artifact-shared-repo'
-    SHARED_REPO_NAME: ClassVar[str] = 'tmt-artifact-shared'
+    SHARED_REPO_NAME: ClassVar[str] = SHARED_REPO_NAME
```
LecrisUT: Does not make sense to be `ClassVar` anymore.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,29 +356,19 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
-            repositories = provider.get_repositories()
-            if repositories:
-                # Repository provider: collect all repo_ids from all repositories.
-                # A .repo file may declare multiple sections; any of them is a valid origin.
-                all_repo_ids = ','.join(
-                    repo_id for repo in repositories for repo_id in repo.repo_ids
+            for artifact in provider.artifacts:
+                existing = pkg_to_repos.setdefault(artifact.version.name, [])
+                existing.extend(
+                    repo_id for repo_id in artifact.repo_ids if repo_id not in existing
```
LecrisUT: The `if` conditional is unnecessary if you change this to be a `set`

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -259,6 +265,7 @@ def enumerate_artifacts(self, guest: Guest) -> None:
                         version=rpm_version,
                         provider=self,
                         location=repository.name,
+                        repo_ids=list(repository.repo_ids),
                     )
                 )
```
LecrisUT, line 282: This part does not seem right. During the `list_packages` the specific repo_id must be saved in the package metadata. Please adjust `RpmVersion` to save that data

LecrisUT (changes requested):

tmt/steps/prepare/artifact/providers/__init__.py
happz: > Use `default` if you mean to use a default.

Mutable default - `default_factory` is correct in this case, each instance gets its own list rather than all instances sharing one list.

LecrisUT: Ah right, that's nasty ...

tmt/steps/prepare/artifact/__init__.py
AthreyVinay: +1

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: Yes, In the artifact domain  one `package` can come only from one `repo-id`.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: This should be set. Changed.

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -45,6 +48,9 @@ class ArtifactInfo:
     version: Version
     location: str
     provider: "ArtifactProvider"
+    #: Repository ID this artifact is available from. Used during verification
+    #: to confirm the artifact was installed from the expected repository.
+    repo_id: str = simple_field(default=SHARED_REPO_NAME)
```
```suggestion
    repo_id: str = SHARED_REPO_NAME
```
Equivalent and more readable

tmt/steps/prepare/artifact/providers/__init__.py
```diff
@@ -254,11 +260,16 @@ def enumerate_artifacts(self, guest: Guest) -> None:
                 )
                 continue
             for rpm_version in packages:
+                from tmt.package_managers._rpm import RpmVersion
+
+                if not isinstance(rpm_version, RpmVersion) or rpm_version.repo_id is None:
+                    continue
```
LecrisUT: It is unclear why this should skip if `repo_id` was not set. Rather know about the case when this happens

tmt/package_managers/dnf.py
```diff
@@ -185,12 +185,16 @@ def install_repository(self, repository: Repository) -> ShellScript:
...
-            """
-        )
+        return (
+            self.command
+            + Command(
+                'repoquery',
+                '--disablerepo=*',
+                *[f'--enablerepo={repo_id}' for repo_id in repository.repo_ids],
+                '--queryformat',
+                r'%{repoid} %{name}-%{epoch}:%{version}-%{release}.%{arch}\n',
```
LecrisUT: A split by space seems fragile. Can use anything else?

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,29 +355,16 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
-                for artifact in provider.artifacts:
-                    pkg_to_repo[artifact.version.name] = self.SHARED_REPO_NAME
+            for artifact in provider.artifacts:
+                pkg_to_repos.setdefault(artifact.version.name, set()).add(artifact.repo_id)
 
         # Only verify packages that are both required and from a known artifact.
-        pkgs_to_verify = {pkg: repo for pkg, repo in pkg_to_repo.items() if pkg in pkg_names}
+        pkgs_to_verify = {
+            pkg: sorted(repo_ids) for pkg, repo_ids in pkg_to_repos.items() if pkg in pkg_names
+        }
```
LecrisUT: These seem rather unnecessarily duplicating. You can just do the if check when going through the `provider.artifacts`

tmt/steps/prepare/verify_installation.py
```diff
@@ -132,15 +137,16 @@ def go(
                 continue
 
             failed_packages.append(package)
+            expected_repos_formatted = ' or '.join(f"'{repo_id}'" for repo_id in expected_repos)
```
LecrisUT: We have `fmf.utils.listed` that does this.

LecrisUT (commented):
Overall lgtm

tmt/package_managers/dnf.py
vaibhavdaren: `\t` is the standard convention for structured command output just like git.

tmt/steps/prepare/artifact/providers/__init__.py
vaibhavdaren: Changed to an explicit raise.

tmt/package_managers/dnf.py
LecrisUT: Rather something like `,` or `;`. Whitespace is fragile

LecrisUT (approved):

Issue #4121, opened 2025-09-26
Add support for RHEL/CentOS-Stream 7.9 to epel feature


tcornell-bus:
When I try to use the `prepare --how feature --epel enabled` with RHEL-7.9, it is skipped with a warning.
```
warn: EPEL prepare feature is supported on RHEL/CentOS-Stream 8+.
```

Would it be possible to provide support for RHEL-7.9?

psss:
The `epel-7` repository is no longer supported:

* https://docs.fedoraproject.org/en-US/epel/getting-started/#end_of_life_releases

I'm not sure we would be able to make it working.

psss:
The repositories seems to be still available, let's try to enable it.
Probably show a warning about the deprecation.
Essential for RHEL-7 testing.

tcornell-bus:
After a quick investigation, this will likely need to wait until after the all-mighty install (artifact) plugins are implemented. The reason is that Ansible does not support < python 3.7 and to get python 3.8 on RHEL 7.9 we would need the RH Software Collections (SCL) repository to be installed. 

It seems like the prepare artifacts plugins could help with adding the SCL repo? Though I might be misunderstanding something.

thrix:
@tcornell-bus for RHEL7 we should somehow make use of the testing farm solution - https://gitlab.com/testing-farm/standalone-python

PR #4800, opened 2026-04-15
Support `epel` and `crb` features in Image Mode

### LINKED ISSUE #4625: Support Ansible-based feature plugins in image mode

thrix:
Support `epel` and `crb` features in Image Mode
    
Replace Ansible playbooks with native Python implemetnation
for these feature plugins and support the plugins in Image Mode.
    
For FIPS, just inform the user we do not support the feature
plugin in image mode. This was discussed internally and currently
not supported:
    
    https://redhat-internal.slack.com/archives/C03BRN71JAF/p1776756406788799
    
Other notable changes:
   
    - Add `distro_id` and `distro_major_version` to guest facts (from
      `/etc/os-release`)
    - Drop Python 3.9 workaround, not needed anymore as we do not use Ansible
    - Update and extend tests to support `virtual` provisioner and both package
      mode and image mode variants
    - Do not install epel-next on RHEL-9 to conform to the official
      installation instructions: https://docs.fedoraproject.org/en-US/epel/getting-started/
    
Assisted-by: Claude Code

Pull Request Checklist

* [x] implement the feature
* [x] adjust plugin docstring
* [ ] include a release note


tmt/steps/prepare/feature/epel.py
```diff
@@ -64,8 +65,57 @@ def enable(cls, guest: Guest, logger: tmt.log.Logger) -> None:
...
+                    )
+        elif distro == 'centos':
+            # CentOS has epel-release in its default repos
+            guest.package_manager.install(Package("epel-release"))
+            if version == 7:
+                guest.package_manager.install(Package("yum-utils"))
+            else:
+                guest.package_manager.install(Package("dnf-command(config-manager)"))
+                # EPEL Next is available for CentOS Stream 9
+                # Enable for Stream 10 once epel-next is available
```
LecrisUT:
```suggestion
```
What? This would never happen. Seems like legacy comment

tmt/steps/prepare/feature/epel.py
```diff
@@ -64,8 +65,57 @@ def enable(cls, guest: Guest, logger: tmt.log.Logger) -> None:
...
+                if version == 9:
+                    guest.package_manager.install(Package("epel-next-release"))
+
         cls._run_playbook('enable', "epel-enable.yaml", guest, logger)
```
LecrisUT: Is there value of keeping the remaining playbook?

thrix: @LecrisUT yes, I do not want to implement more spagethi fun in the code only the amount is needed. Also the same for FIPS.

tmt/steps/prepare/feature/epel.py
thrix: This was addressed

tmt/steps/prepare/feature/epel.py
LecrisUT: Let me rephrase it. What is being done in the playbook now that there is the logic section above?

Edit: or let me put it differently, would this even do what you expect it to do since the `install` commands are done at build time and the playbook would run immediately without even the plugin available to handle `config-manager`

thrix: > Let me rephrase it. What is being done in the playbook now that there is the logic section above?

In the playbook we mostly play around with repository enablement:

https://github.com/teemtee/tmt/blob/c0a3c8994f6be4829632b58e3c13f74dec35e0a8/tmt/steps/prepare/feature/epel-enable.yaml

> Edit: or let me put it differently, would this even do what you expect it to do since the install commands are done at build time and the playbook would run immediately without even the plugin available to handle config-manager

The `bootc` package manager `install` function should install the missing dependencies to a derived container image, build the new container image and switch to it:

https://github.com/teemtee/tmt/blob/c0a3c8994f6be4829632b58e3c13f74dec35e0a8/tmt/package_managers/bootc.py#L301-L316

So not sure I follow how it would run immediately. It runs after the missing dependencies are installed the image mode way.

LecrisUT: > > Let me rephrase it. What is being done in the playbook now that there is the logic section above?
> 
> In the playbook we mostly play around with repository enablement:
> 
> https://github.com/teemtee/tmt/blob/c0a3c8994f6be4829632b58e3c13f74dec35e0a8/tmt/steps/prepare/feature/epel-enable.yaml

That is an answer to the litteral question, not the spirit of the question. The purpose of the question is to prod on whether it makes sense to keep that or not, which is not addressed.

> > Edit: or let me put it differently, would this even do what you expect it to do since the install commands are done at build time and the playbook would run immediately without even the plugin available to handle config-manager
> 
> The `bootc` package manager `install` function should install the missing dependencies to a derived container image, build the new container image and switch to it:
> 
> https://github.com/teemtee/tmt/blob/c0a3c8994f6be4829632b58e3c13f74dec35e0a8/tmt/package_managers/bootc.py#L301-L316
> 
> So not sure I follow how it would run immediately. It runs after the missing dependencies are installed the image mode way.

Please read my comment again and more carefully!

thrix: > > > Let me rephrase it. What is being done in the playbook now that there is the logic section above?
> > 
> > 
> > In the playbook we mostly play around with repository enablement:
> > https://github.com/teemtee/tmt/blob/c0a3c8994f6be4829632b58e3c13f74dec35e0a8/tmt/steps/prepare/feature/epel-enable.yaml
> 
> That is an answer to the litteral question, not the spirit of the question. The purpose of the question is to prod on whether it makes sense to keep that or not, which is not addressed.

It does for me, I do not want to introduce more python code as necessary. The ansible code alreaedy does decent job doing what is needed except the packages installation.

> 
> > > Edit: or let me put it differently, would this even do what you expect it to do since the install commands are done at build time and the playbook would run immediately without even the plugin available to handle config-manager
> > 
> > 
> > The `bootc` package manager `install` function should install the missing dependencies to a derived container image, build the new container image and switch to it:
> > https://github.com/teemtee/tmt/blob/c0a3c8994f6be4829632b58e3c13f74dec35e0a8/tmt/package_managers/bootc.py#L301-L316
> > 
> > So not sure I follow how it would run immediately. It runs after the missing dependencies are installed the image mode way.
> 
> Please read my comment again and more carefully!

Which one, this below? Man, I read it like 10 times already :) and I thought I replied well.

> Edit: or let me put it differently, would this even do what you expect it to do since the install commands are done at build time 

yes, they are done at build time, but they are also applid to the live image right after each install command.

> and the playbook would run immediately without even the plugin available to handle config-manager

No, the missing deps are installed and applied on the running system BEFORE the playbook runs.

I just tested it with injecting another dep, because all those ar actually alraedy baked into our images:

```
❯ tmt -vvv run -a plan --name '/epel/enabled/with-epel-preinstalled' provision --how virtual --image https://artifacts.dev.testing-farm.io/images/CentOS-Stream-9-image-mode-x86_64.qco
w2
...
        how: feature
        order: 50
> /var/home/thrix/git/github.com/teemtee/tmt/worktree/feature-plugins-image-mode/tmt/steps/prepare/feature/epel.py(76)enable()
-> import pdb; pdb.set_trace()
(Pdb)  distro
'centos'
(Pdb) n
> /var/home/thrix/git/github.com/teemtee/tmt/worktree/feature-plugins-image-mode/tmt/steps/prepare/feature/epel.py(77)enable()
-> guest.package_manager.install("neovim")
(Pdb) n
            cmd: rpm -q --whatprovides neovim
            stdout: no package provides neovim
            cmd: mktemp --directory tmp.XXXXXXXXXX
            stdout: tmp.KY02SByTo0
            cmd: /bin/bash -c "(  ( podman pull quay.io/testing-farm/centos-bootc:stream9 || podman pull containers-storage:quay.io/testing-farm/centos-bootc:stream9 )  || bootc image copy-to-storage --target quay.io/testing-farm/centos-bootc:stream9)"
            stderr: Trying to pull quay.io/testing-farm/centos-bootc:stream9...
            stderr: Getting image source signatures
            stderr: Copying blob sha256:9f2741e8a325eabbd2d2fdd05a403342e367cba176db17ad0cee655fd76c2620
            stderr: Copying blob sha256:29b667a8b294984efb1c0aca524ff788af33ee241b4adf543a64e13e23d134ba
            stderr: Copying blob sha256:0f6aaa63cc93223316d8174f136b292152af143f07236da94146f0489a38ea05
....
....
(neovim is installed before the playbook runs)
```

tmt/steps/prepare/feature/epel.py
```diff
@@ -64,8 +65,47 @@ def enable(cls, guest: Guest, logger: tmt.log.Logger) -> None:
...
+                PackageUrl(
+                    f"https://dl.fedoraproject.org/pub/epel/"
+                    f"epel-release-latest-{version}.noarch.rpm"
+                ),
+                options=Options(allow_untrusted=True),
+            )
+            if version == 7:
+                guest.package_manager.install(Package("yum-utils"))
+            else:
+                guest.package_manager.install(Package("dnf-command(config-manager)"))
```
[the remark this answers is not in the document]
thrix: Yes, and that is actually a bug we are fixing here:

https://docs.fedoraproject.org/en-US/epel/getting-started/#_el9

See? no epel-nxt installation for RHEL 9!

I will mention it in the description

tmt/steps/prepare/feature/epel.py
LecrisUT: > > > > Edit: or let me put it differently, would this even do what you expect it to do since the install commands are done at build time and the playbook would run immediately without even the plugin available to handle config-manager
>
> ...
> 
> > Please read my comment again and more carefully!
> 
> Which one, this below? Man, I read it like 10 times already :) and I thought I replied well.

The conversation that I quoted. My comment there is to indicate that you have misinterpreted what I was saying, i.e. re-read what I wrote and try to find if we are misaligned between what you think I wrote, and what I was trying to write, *not* about your reply!

To be more precise:
- I did not say that the `install` commands are done immediately, nor will they ever be able to because `dnf install` fails. These are done at build time
- I said that the playbooks are the things that are run immediately
- There is discrepancy between these two

> yes, they are done at build time, but they are also applid to the live image right after each install command.

What? How can `dnf install` be applied to a live bootc image? Also be more specific on what you are referencing there, the `install` step, the `ansible` step, or something completely different. I have interpreted that you were referring to the `install` step.

> > and the playbook would run immediately without even the plugin available to handle config-manager
> 
> No, the missing deps are installed and applied on the running system BEFORE the playbook runs.

Huh, but then that means that we are doing the rebuild every time we hit an ansible step? E.g. if we have a `prepare/install` -> `prepare/ansible` -> `prepare/shell`, then it does: add install step to deferred commands -> build container -> run ansible -> add shell step to deferred commands -> build container? That does not fit with the understanding of what was done to support `prepare/install` + `prepare/shell` for bootc image. There is a big misalignment in how the bootc support is going on here if this is the case. Please prepare the design document more clearly in https://github.com/teemtee/tmt/pull/4818 with some references in the PR so that we can all be on the same page of what is going on here.

> I just tested it with injecting another dep, because all those ar actually alraedy baked into our images:
> 
> ```
> ❯ tmt -vvv run -a plan --name '/epel/enabled/with-epel-preinstalled' provision --how virtual --image https://artifacts.dev.testing-farm.io/images/CentOS-Stream-9-image-mode-x86_64.qco
> w2
> ...
>         how: feature
>         order: 50
> > /var/home/thrix/git/github.com/teemtee/tmt/worktree/feature-plugins-image-mode/tmt/steps/prepare/feature/epel.py(76)enable()
> -> import pdb; pdb.set_trace()
> (Pdb)  distro
> 'centos'
> (Pdb) n
> > /var/home/thrix/git/github.com/teemtee/tmt/worktree/feature-plugins-image-mode/tmt/steps/prepare/feature/epel.py(77)enable()
> -> guest.package_manager.install("neovim")
> (Pdb) n
>             cmd: rpm -q --whatprovides neovim
>             stdout: no package provides neovim
>             cmd: mktemp --directory tmp.XXXXXXXXXX
>             stdout: tmp.KY02SByTo0
>             cmd: /bin/bash -c "(  ( podman pull quay.io/testing-farm/centos-bootc:stream9 || podman pull containers-storage:quay.io/testing-farm/centos-bootc:stream9 )  || bootc image copy-to-storage --target quay.io/testing-farm/centos-bootc:stream9)"
>             stderr: Trying to pull quay.io/testing-farm/centos-bootc:stream9...
>             stderr: Getting image source signatures
>             stderr: Copying blob sha256:9f2741e8a325eabbd2d2fdd05a403342e367cba176db17ad0cee655fd76c2620
>             stderr: Copying blob sha256:29b667a8b294984efb1c0aca524ff788af33ee241b4adf543a64e13e23d134ba
>             stderr: Copying blob sha256:0f6aaa63cc93223316d8174f136b292152af143f07236da94146f0489a38ea05
> ....
> ....
> (neovim is installed before the playbook runs)
> ```

Please provide a simple tmt plan to properly follow. I do not follow what you showed there

tmt/steps/prepare/feature/epel.py
```diff
@@ -64,8 +65,48 @@ def enable(cls, guest: Guest, logger: tmt.log.Logger) -> None:
...
+                if version == 9:
+                    guest.package_manager.install(Package("epel-next-release"))
```
LecrisUT: This is wrong. `epel-next` is only for centos

tmt/steps/prepare/feature/epel.py
```diff
@@ -64,8 +65,48 @@ def enable(cls, guest: Guest, logger: tmt.log.Logger) -> None:
...
+                PackageUrl(
+                    f"https://dl.fedoraproject.org/pub/epel/"
+                    f"epel-release-latest-{version}.noarch.rpm"
+                ),
+                options=Options(allow_untrusted=True),
+            )
+            if version == 7:
+                guest.package_manager.install(Package("yum-utils"))
+            else:
+                guest.package_manager.install(Package("dnf-command(config-manager)"))
```
LecrisUT: This should be handled similarly with `copr_plugin`. We also have a handling of `_base_debuginfo_command` that can be generalized or implemented in a similar way

tmt/steps/prepare/feature/epel.py
thrix: Resolving, this is not outdated

tmt/steps/prepare/feature/epel.py
thrix: This was addressed.

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,136 @@ class Epel(ToggleableFeature):
...
             guest.facts.distro
             and any(pattern.match(guest.facts.distro) for pattern in SUPPORTED_DISTRO_PATTERNS)
-        ):
+        )
+
+    @classmethod
+    def _install_config_manager(cls, guest: Guest, version: int) -> None:
+        """Install the appropriate config-manager tool for the distro version."""
+        # NOTE: RHEL/CentOS 7 is not matched by SUPPORTED_DISTRO_PATTERNS,
+        # this branch is currently unreachable (EPEL 7 archived June 2024).
```
thrix: @tcornell-bus currently these plugins do not work for RHEL7. The original code had this dead part there. Do we want to make it run, or SE does not need it to work? Can you try to find out?

thrix: I would like to ideally not have dead code in the codebase

tmt/steps/prepare/feature/epel.py
thrix: Yes, but it is a larger refactoring I will not do in this PR. Filed #4847, with a draft PR under #4848

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,136 @@ class Epel(ToggleableFeature):
...
+            guest.package_manager.install(Package("dnf-command(config-manager)"))
+
+    @classmethod
+    def _enable_repos(
+        cls, guest: Guest, distro: str, version: int, logger: tmt.log.Logger
+    ) -> None:
+        """Enable EPEL repositories via config-manager."""
+        # NOTE: RHEL/CentOS 7 is not matched by SUPPORTED_DISTRO_PATTERNS,
+        # this branch is currently unreachable (EPEL 7 archived June 2024).
+        if version == 7:
```
happz: This should be handled by package managers. We probably don't have such API yet, but enabling a repo is something package managers should know how to achieve.

tests/prepare/feature/crb/test.sh
```diff
@@ -2,51 +2,68 @@
 . /usr/share/beakerlib/beakerlib.sh || exit 1
 . ../../../images.sh || exit 1
 
+CONTAINER_IMAGES="$TEST_IMAGE_PREFIX/centos/stream9/upstream:latest
+$TEST_IMAGE_PREFIX/ubi/8/upstream:latest
+ubi9
```
LecrisUT: What's up with this one?

tests/prepare/feature/crb/test.sh
```diff
@@ -2,51 +2,68 @@
...
         rlRun "PROVISION_HOW=${PROVISION_HOW:-container}"
+        rlRun "IMAGE_MODE=${IMAGE_MODE:-no}"
 
-        build_container_image "centos/stream9/upstream\:latest"
-        build_container_image "ubi/8/upstream\:latest"
-        build_container_image "fedora/latest\:latest"
+        if [ "$PROVISION_HOW" = "container" ]; then
+            build_container_image "centos/stream9/upstream\:latest"
+            build_container_image "ubi/8/upstream\:latest"
+            build_container_image "fedora/latest\:latest"
```
LecrisUT: If you defined the `CONTAINER_IMAGED` in the part just above, why keep a manual list here?

tests/prepare/feature/crb/test.sh
```diff
@@ -2,51 +2,68 @@
...
+
+            # Run the '/crb/disabled' plan, overriding the provision image.
+            # The plan itself contains the checks (! dnf repolist enabled ... && dnf repolist disabled ...).
+            # Expecting tmt run to succeed (exit code 0) as the checks should pass.
+            rlPhaseStartTest "Test /crb/disabled on $image"
+                rlRun "tmt run --all provision --how $PROVISION_HOW --image $image plan --name /crb/disabled" 0 "Run /crb/disabled plan for $image"
+            rlPhaseEnd
+
+            # Run the '/crb_package' only on c9s
+            if is_centos_stream_9 "$image"; then
```
LecrisUT: Does it still make sense to gate it here?

tests/prepare/feature/crb/main.fmf
```diff
@@ -3,3 +3,5 @@ description: Make sure enable/disable CRB works.
 tag+:
   - provision-container
   - provision-only
+  - provision-virtual
```
LecrisUT, line 6: Looks weird for `provision-only` to be in the middle

tests/prepare/feature/epel/test.sh
```diff
@@ -2,55 +2,67 @@
...
-    images="$TEST_IMAGE_PREFIX/centos/stream9/upstream:latest $TEST_IMAGE_PREFIX/ubi/8/upstream:latest ubi9"
-
     # EPEL
-    for image in $images; do
-        if rlIsFedora ">=42" && is_centos_7 "$image"; then
-            rlLogInfo "Skipping because Ansible shipped with Fedora does not support Python 3.6"
-
+    while IFS= read -r image; do
+        if is_fedora "$image"; then
+            rlLogInfo "Skipping Fedora for testing EPEL"
```
LecrisUT: The comment in the previous part does not really make sense to me. Better to align it with CRB approach

tmt/steps/prepare/feature/__init__.py
```diff
@@ -257,21 +257,9 @@ class PrepareFeature(tmt.steps.prepare.PreparePlugin[PrepareFeatureData]):
...
-       the fact it uses Ansible to implement some of the features:
-
-       * Ansible 2.17+ no longer supports Python 3.6 and older. Guests
-         where Python 3.7+ is not available cannot be prepared with the
-         ``feature`` plugin. This has been observed when Fedora Rawhide
-         runner is used with CentOS 7 or CentOS Stream 8 guests. Possible
-         workarounds: downgrade Ansible tmt uses, or install Python 3.7+
-         before using ``feature`` plugin from an alternative repository
-         or local build.
+       Some features (such as ``fips``) require a working Ansible
```
LecrisUT:
```suggestion
       ``fips`` feature currently require a working Ansible
```
I do not see a reason why this cannot be converted as well in the future to preserve consistency.

Should also indicate that the python method is preferred to avoid the injection of python dependency

tests/prepare/feature/crb/data/plans.fmf
```diff
@@ -1,5 +1,5 @@
 provision:
-  how: container
+  how: $PROVISION_HOW
```
LecrisUT: The environment variable is not passed, why not delete it instead?

LecrisUT:
Test failures are relevant, please check

tmt/steps/prepare/feature/epel.py
tcornell-bus: We do need EPEL for rhel-7.9, I had made an [issue](https://github.com/teemtee/tmt/issues/4121) and did a brief investigation and the Python version was a blocker there. I would be lovely if this worked for RHEL-7.9 :smile: 

I think CRB is not a thing for rhel-7.9 so no worries there.

tcornell-bus: I could understand if it needs to be a separate PR, though. I think RHEL-7 epel has been archived, so there is no epel rpm you can just install anymore.

LecrisUT: It should be testing-farm profile instead imo. If epel is to be enabled there are other steps like pointing it to the archive which should not be tmt's responsibility

thrix: We already fix the repositories in the guest setup for eoled instance on TF side, so definitely outside of tmt.

thrix: @tcornell-bus like I can enable it,but to test it properly we would need to wait for profiles to land or implement a prepare steps that applies these few wokarounds. I am fine doing that if you want. Just let me know

tmt/steps/prepare/feature/epel.py
thrix: yep, see #4847 #4848

tmt/steps/prepare/feature/epel.py
tcornell-bus: Let's enable RHEL-7 and RHEL-6 and then wait for profiles to land please.

thrix: ok, will do

tmt/steps/prepare/feature/epel.py
thrix: Resolved, it is based on top of #4848 now

tests/prepare/feature/crb/test.sh
thrix: I am not sure why it was not enabled, testing on ubi8 and ubi9, according to the code it should be supported ...

thrix: as the package of CRB repos is different, it will need the installed package adjustment, but otherwise it will work ...

tmt/steps/prepare/feature/epel.py
thrix: RHEL-6 an RHEL-7 allowed, but not tested as it cannot be reasonably tested without profiles and standalone python3

tests/prepare/feature/crb/test.sh
thrix: condition should be gone now.

tests/prepare/feature/crb/main.fmf
thrix: reordered, `provision-only` is now first.

tests/prepare/feature/epel/test.sh
thrix: Removed the `# EPEL` comment as it is unneeded. But I am keeping the code as is, it looks fine to me this way.

tmt/steps/prepare/feature/__init__.py
thrix: I have filed #4899, fixed docs to indicate what is preferred

tests/prepare/feature/crb/data/plans.fmf
thrix: Indeed, it was removed

thrix:
> Test failures are relevant, please check

2 failures addresed, waiting for fresh run.

tests/prepare/feature/crb/data/plans.fmf
```diff
@@ -32,5 +29,5 @@ execute:
         crb: enabled
     execute:
       script: |
-        dnf install -y sgx-rpm-macros
-        rpm -q sgx-rpm-macros
+        dnf install -y gdbm
+        rpm -q gdbm
```
thrix, line 33: FTR, this was changes to packages which work also for CRB for UBI images.

tmt/steps/prepare/feature/__init__.py
```diff
@@ -257,21 +257,10 @@ class PrepareFeature(tmt.steps.prepare.PreparePlugin[PrepareFeatureData]):
...
-         where Python 3.7+ is not available cannot be prepared with the
-         ``feature`` plugin. This has been observed when Fedora Rawhide
-         runner is used with CentOS 7 or CentOS Stream 8 guests. Possible
-         workarounds: downgrade Ansible tmt uses, or install Python 3.7+
-         before using ``feature`` plugin from an alternative repository
-         or local build.
+       Python is the preferred implementation language for feature
+       plugins as it avoids injecting an Ansible dependency on the
+       test runner. The ``fips`` feature currently still requires a
+       working Ansible.
```
happz: As a user, I don't care about your language of choice, not interested in the legend and myth. What I care about is that the `fips` plugin still requires Ansible, and therefore the removed warning and note still apply. They could get an extra word or two mentioning that now it's just the `feature` + `fips` rather than "cannot be prepared with the `feature` plugin", but it's still valid.

thrix: Ok, reverting, @LecrisUT fyi

tmt/steps/prepare/feature/fips.py
```diff
@@ -46,6 +46,10 @@ class Fips(ToggleableFeature):
        In order to prevent issues with installation of packages signed by
        non-FIPS-compliant algorithms we recommend enabling FIPS mode after
        package installation prepare steps. Use ``order:`` to enforce that.
+
+    .. note::
+
+       This feature plugin is not supported in Image Mode.
```
happz: And container and ostree systems, according to the exception message below.

tmt/steps/prepare/feature/__init__.py
LecrisUT: Fine with dropping the language part. It is mostly for the developer's note, which we do not really have a way to add that.

tmt/steps/prepare/feature/epel.py
happz: Unresolving: yes, this PR is now based on code with #4848 merged, but `epel.py` still does the whole "if version in [6, 7], then yum-config-manager" dance instead of calling some form of `guest.package_manager.enable/disable_repo()`. It's the purpose of the package manager layer to protect common code from the burden of guessing which command is the right one to install a package or enable a repository. `guest.package_manager.enable_repo('epel', 'epel-debuginfo', 'epel-source')`, for example.

tmt/steps/prepare/feature/fips.py
thrix: thanks, fixed

tmt/steps/prepare/feature/__init__.py
thrix: Dropped the language choice note and added back the warning. Please recheck.

tmt/steps/prepare/feature/epel.py
thrix: @happz creating a new PR.

tests/prepare/feature/epel/test.sh
LecrisUT: What I mean is that the previous skip reason

> Skipping because Ansible shipped with Fedora does not support Python 3.6

Did not really make sense then, and now it makes even less sense. CRP on the other hand was running even on Fedora even though it would be a no-op and it covered its behavior:

```
        # Run a CRB plan (just provision, prepare and finish) on fedora and verify the warning is shown
        # We expect the tmt run to succeed (exit code 0) because it's a warning, not an error.
        rlRun -s "tmt run provision --how container --image $TEST_IMAGE_PREFIX/fedora/latest:latest prepare finish plan --name /crb/enabled" 0 "Run plan on fedora and capture output"
        rlAssertGrep "CRB prepare feature is supported on RHEL/CentOS-Stream 8, 9 or 10." $rlRun_LOG
```

tmt/steps/prepare/feature/crb.py
LecrisUT: Note about this change. As @tcornell-bus noted, `guest.execute` runs immediately by default which conflicts with the `dnf config-manager` part here. Either make it deferred or make sure it is built at each of these steps

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,127 @@ class Epel(ToggleableFeature):
...
+
+        elif distro == 'centos':
+            guest.package_manager.install(Package("epel-release"))
+            # EPEL Next is available for CentOS Stream 9 only
+            if version == 9:
+                guest.package_manager.install(Package("epel-next-release"))
+
+        guest.package_manager.assert_config_manager()
+
+        logger.info('Enable EPEL')
```
LecrisUT: Shouldn't this be a `logger.verbose` level?

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,127 @@ class Epel(ToggleableFeature):
...
+    def _get_distro_facts(cls, guest: Guest) -> tuple[str, int]:
+        """Return validated distro id and major version from guest facts."""
+        distro = guest.facts.distro_id
+        version = guest.facts.distro_major_version
+
+        if distro is None:
+            raise tmt.utils.GeneralError('Cannot determine distro id from guest facts.')
+
+        if version is None:
+            raise tmt.utils.GeneralError('Cannot determine distro major version from guest facts.')
```
LecrisUT, line 77: The purpose of this seems more like a `assert_distro_facts` function

tmt/steps/prepare/feature/__init__.py
happz: Please, bring back the note as well: the plugin still requires Ansible to work on the guest when some of the feature plugins are used. This is still true.

tmt/steps/prepare/feature/crb.py
thrix: yes, this one should not run immediately, fixed

tmt/steps/prepare/feature/epel.py
thrix: @LecrisUT I used info here, as that is what CRB uses:
https://github.com/teemtee/tmt/blob/main/tmt/steps/prepare/feature/crb.py#L64C1-L65C1

tmt/steps/prepare/feature/epel.py
thrix: @LecrisUT it is `_assert_distro_facts`, did not move it elsewehre as seems we do not have another usage as in this module

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,129 @@ class Epel(ToggleableFeature):
...
+                        " --enable epel-next epel-next-debuginfo epel-next-source"
+                    )
+                )
+
+            # Enable CRB repository (needed for EPEL dependencies).
+            # FORCE_DNF=1 skips subscription-manager, not configured in test environments.
+            logger.info('Enable CRB (EPEL dependency)')
+            guest.execute(
+                ShellScript(f"FORCE_DNF=1 {guest.facts.sudo_prefix} crb enable"), immediately=False
+            )
```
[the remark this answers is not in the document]
thrix: I do not think this is true

tmt/steps/prepare/feature/crb.py
thrix: this should be outdated now

tmt/steps/prepare/feature/__init__.py
thrix: @happz note should be back

tmt/steps/prepare/feature/epel.py
thrix: Resolving this one.

tmt/steps/prepare/feature/epel.py
thrix: @happz this MR is now based on #4903

tests/prepare/feature/epel/test.sh
thrix: ok, I see thx, checking

tmt/steps/prepare/feature/epel.py
LecrisUT: Doesn't mean that that one is aligned, and this one has a different behavior to the previous ansible one.

tmt/steps/prepare/feature/epel.py
LecrisUT: As you can see in github UI, it was not when I commented. And assert functions do not return

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,96 @@ class Epel(ToggleableFeature):
...
     def disable(cls, guest: Guest, logger: tmt.log.Logger) -> None:
-        cls._run_playbook('disable', "epel-disable.yaml", guest, logger)
+        if not cls._is_supported(guest):
+            logger.warning('EPEL prepare feature is supported on RHEL/CentOS-Stream 8+.')
+            return
+
+        distro, version = cls._assert_distro_facts(guest)
+
+        logger.info('Disable EPEL')
+        cls._disable_if_installed(guest, "epel-release", "epel", "epel-debuginfo", "epel-source")
```
LecrisUT: The interface of this one is so confusing.

tmt/steps/prepare/feature/epel.py
thrix: agreed, both are verbose now

tmt/steps/prepare/feature/epel.py
thrix: @LecrisUT refactored, what about now?

tests/prepare/feature/epel/test.sh
```diff
@@ -2,55 +2,74 @@
...
-    for image in $images; do
-        if rlIsFedora ">=42" && is_centos_7 "$image"; then
-            rlLogInfo "Skipping because Ansible shipped with Fedora does not support Python 3.6"
+        if is_fedora "$image"; then
+            # Test Fedora for the warning message
+            rlPhaseStartTest "Test warning on $image"
+                # Run a epel plan (just provision, prepare and finish) on fedora and verify the warning is shown
+                # We expect the tmt run to succeed (exit code 0) because it's a warning, not an error.
+                rlRun -s "tmt run provision --how $PROVISION_HOW --image $image prepare finish plan --name /epel/enabled/default" 0 "Run plan on fedora and capture output"
+                rlAssertGrep "EPEL·prepare·feature·is·supported·on·RHEL/CentOS-Stream·8+." $rlRun_LOG
```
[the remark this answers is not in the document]
thrix: yeah, copy paste error :face_exhaling:

tmt/steps/prepare/feature/epel.py
```diff
@@ -51,21 +54,92 @@ class Epel(ToggleableFeature):
...
+        if distro is None:
+            raise tmt.utils.GeneralError('Cannot determine distro id from guest facts.')
+
+        if version is None:
+            raise tmt.utils.GeneralError('Cannot determine distro major version from guest facts.')
+
+        return distro, version
+
+    @classmethod
+    def _is_installed(cls, guest: Guest, package: str) -> bool:
```
happz: I recommend using the existing `guest.package_manager.check_presence()`, you probably are not aware of it.

thrix: seems so

thrix: @happz hopefully fixed, thx!

tests/main.fmf
```diff
@@ -15,7 +15,7 @@ adjust:
         we don't want to run the avc check for local testing as it
         needs root and we're executing tests under a regular user
   - when: provision_how == virtual
-    duration+: "*2"
+    duration+: "*3"
```
thrix:
```suggestion
    duration+: "*2"
```

happz (approved):

LecrisUT (approved):

psss (approved):
Looks good, just missing release note, but I'll add it in the release commit.

PR #4801, opened 2026-04-15
Running `tmt clean` will now show disk space freed

### LINKED ISSUE #4593: `tmt clean --dry` could show total disk space that would be freed
Current behavior:

```bash
> tmt clean -v --dry
clean
guests
workdir root: /var/tmp/tmt
runs
workdir root: /var/tmp/tmt
    Would remove workdir '/var/tmp/tmt/run-001'
```

proposed:
```bash
> tmt clean -v --dry
clean
guests
workdir root: /var/tmp/tmt
runs
workdir root: /var/tmp/tmt
    Would remove workdir '/var/tmp/tmt/run-001' (0.28MB)
    ... (5 more)

Summary: Would free <insert_total_freed_space> of disk space
```

Alternatively, even without `--dry`, showing a summary after cleanup would be helpful. Just an indication if cleanup is worth the time.

AthreyVinay:
resolves: #4593

Pull Request Checklist

* [X] implement the feature
* [x] include a release note

### DIFF
--- a/tmt/base/core.py
+++ b/tmt/base/core.py
@@ -40,6 +40,7 @@ import tmt.convert
 import tmt.export
 import tmt.frameworks
 import tmt.guest
+import tmt.hardware
 import tmt.identifier
 import tmt.lint
 import tmt.log
@@ -3009,7 +3010,7 @@ CleanCallback = Callable[[], bool]
 
 def _dir_size(path: Path) -> int:
     """Return the total size in bytes of all files under path."""
-    return sum(f.stat().st_size for f in path.rglob('*') if f.is_file())
+    return sum(f.lstat().st_size for f in path.rglob('*'))
 
 
 class Clean(tmt.utils.Common):
@@ -3159,11 +3160,12 @@ class Clean(tmt.utils.Common):
                 successful = False
         return successful
 
-    def _clean_workdir(self, path: Path, size: int) -> bool:
+    def _clean_workdir(self, path: Path) -> tuple[bool, int]:
         """
         Remove a workdir (unless in dry mode)
         """
-        formatted_size = tmt.utils.format_size(size)
+        size = _dir_size(path)
+        formatted_size = round(tmt.hardware.UNITS(f'{size} bytes').to_compact(), 1)
         if self.is_dry_run:
             self.verbose(f"Would remove workdir '{path}' ({formatted_size}).", shift=1)
         else:
@@ -3172,8 +3174,8 @@ class Clean(tmt.utils.Common):
                 shutil.rmtree(path)
             except OSError as error:
                 self.warn(f"Failed to remove '{path}': {error}.", shift=1)
-                return False
-        return True
+                return False, 0
+        return True, size
 
     def runs(self, id_: tuple[str, ...], keep: Optional[int]) -> bool:
         """
@@ -3190,11 +3192,10 @@ class Clean(tmt.utils.Common):
             # the correct one.
             last_run = Run(logger=self._logger, cli_invocation=self.cli_invocation)
             last_run.load_workdir(with_logfiles=False)
-            size = _dir_size(last_run.run_workdir)
-            success = self._clean_workdir(last_run.run_workdir, size)
+            success, size = self._clean_workdir(last_run.run_workdir)
             self.verbose(
                 f"Summary: {'Would free' if self.is_dry_run else 'Freed'} "
-                f"{tmt.utils.format_size(size)} of disk space.",
+                f"{round(tmt.hardware.UNITS(f'{size} bytes').to_compact(), 1)} of disk space.",
                 shift=1,
             )
             return success
@@ -3207,14 +3208,14 @@ class Clean(tmt.utils.Common):
         successful = True
         total_size = 0
         for workdir in all_workdirs:
-            size = _dir_size(workdir)
-            total_size += size
-            if not self._clean_workdir(workdir, size):
+            success, size = self._clean_workdir(workdir)
+            if not success:
                 successful = False
+            total_size += size
 
         self.verbose(
             f"Summary: {'Would free' if self.is_dry_run else 'Freed'} "
-            f"{tmt.utils.format_size(total_size)} of disk space.",
+            f"{round(tmt.hardware.UNITS(f'{total_size} bytes').to_compact(), 1)} of disk space.",
             shift=1,
         )
         return successful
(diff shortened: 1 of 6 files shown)


tmt/utils/__init__.py
```diff
@@ -86,6 +86,15 @@
     from tmt.hardware import Size
 
 
+def format_size(size: int) -> str:
```
happz: Use `tmt.hardware.UNITS` instead: `tmt.hardware.UNITS(f'{size} bytes').to_compact()`

tmt/base/core.py
```diff
@@ -3751,14 +3756,15 @@ def guests(self, run_ids: tuple[str, ...], keep: Optional[int]) -> bool:
                 successful = False
         return successful
 
-    def _clean_workdir(self, path: Path) -> bool:
+    def _clean_workdir(self, path: Path, size: int) -> bool:
```
happz: Computing the size outside of `_clean_workdir` and just let it print out is very weird. We can change the return value of `_clean_workdir`, to e.g. a tuple of `(success, size)`. `_clean_workdir` would collect the size, and just tell its caller how much data it did/would remove.

AthreyVinay: Absolutely - that was my intuition too. The only reason was I was trying to avoid returning a tuple - and I thought dataclass would be an overkill. Will revert to returning a tuple instead.

AthreyVinay: Done: [c4b334a](https://github.com/teemtee/tmt/pull/4801/commits/c4b334ac850d37eecc72b564f3a1f8c7956b2a93)

tmt/utils/__init__.py
AthreyVinay: Done: [c4b334a](https://github.com/teemtee/tmt/pull/4801/commits/c4b334ac850d37eecc72b564f3a1f8c7956b2a93)

tmt/base/core.py
```diff
@@ -3613,6 +3613,11 @@ def show(self) -> None:
 CleanCallback = Callable[[], bool]
 
 
+def _dir_size(path: Path) -> int:
+    """Return the total size in bytes of all files under path."""
+    return sum(f.stat().st_size for f in path.rglob('*') if f.is_file())
```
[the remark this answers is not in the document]
AthreyVinay: Done: [c4b334a](https://github.com/teemtee/tmt/pull/4801/commits/c4b334ac850d37eecc72b564f3a1f8c7956b2a93)

tmt/base/core.py
```diff
@@ -3777,18 +3785,32 @@ def runs(self, id_: tuple[str, ...], keep: Optional[int]) -> bool:
...
             all_workdirs = all_workdirs[keep:]
 
         successful = True
+        total_size = 0
         for workdir in all_workdirs:
-            if not self._clean_workdir(workdir):
+            success, size = self._clean_workdir(workdir)
+            if not success:
                 successful = False
+            total_size += size
```
LecrisUT: Can move this in an `else` and avoid defining the same `self.verbose` message.

LecrisUT (approved):

therazix (approved):

tmt/base/core.py
AthreyVinay: Addressed: [0e495b4](https://github.com/teemtee/tmt/pull/4801/commits/0e495b4edeb02868c107d49340343237f5bbaded)

tmt/steps/provision/testcloud.py
```diff
@@ -1589,16 +1589,25 @@ def clean_images(cls, clean: 'tmt.base.core.Clean', dry: bool, workdir_root: Pat
             clean.warn(f"Directory '{testcloud_images}' does not exist.", shift=2)
             return True
         successful = True
+        total_size = 0
         for image in testcloud_images.iterdir():
+            size = image.stat().st_size
+            total_size += size
+            formatted_size = round(tmt.hardware.UNITS(f'{size} bytes').to_compact(), 1)
```
happz: `round(tmt.hardware.UNITS(...).to_compact(), 1)` appears 4 times in the patch, that sounds like a good candidate for a helper. Maybe under the `tmt.hardware` package:

```python
def format_compact(quantity: 'Quantity', digits: int = 1) -> str:
    return str(round(quantity.to_compact(), digits))
```

```python
f"{format_compact(tmt.hardware.UNITS(f'{total_size} bytes')} of disk space.",
```

AthreyVinay: Fair enough!

AthreyVinay: Addressed:: [4fd1eb2](https://github.com/teemtee/tmt/pull/4801/commits/4fd1eb27c7e04d7d8a20b41d32c6448f31451101)

tmt/base/core.py
```diff
@@ -3007,6 +3008,11 @@ def show(self) -> None:
 CleanCallback = Callable[[], bool]
 
 
+def _dir_size(path: Path) -> int:
```
happz: Also, consider returning `Quantity` here as well - it's not just any integer, it's number of bytes, there's a unit. Arithmetic should work, and the amount of `tmt.hardware.UNITS` mentions would lower. The same for `_clean_workdir()`, it also returns not just any unit, but the size of something.

AthreyVinay: Addressed: [d98314f](https://github.com/teemtee/tmt/pull/4801/commits/d98314f2326658d38506ac39bd3bfbb14d4259b4). Arithmetic doesnt seem to work, hence the `# type: ignore[misc]`

happz: Doesn't work, or it's just a linter complaining? Are you getting right answers in runtime?

AthreyVinay: > linter complaining

Apologies, shoudve been clearer.

happz (approved):

tmt/base/core.py
```diff
@@ -3184,18 +3194,28 @@ def runs(self, id_: tuple[str, ...], keep: Optional[int]) -> bool:
...
+
+            successful = True
+            total_size = tmt.hardware.UNITS('0 bytes')
+            for workdir in all_workdirs:
+                success, size = self._clean_workdir(workdir)
+                if not success:
+                    successful = False
+                total_size += size  # type: ignore[misc]
+
+        self.verbose(
```
psss:
```suggestion
        self.info(
```

Wouldn't `info` be more appropriate? I tried just `tmt clean --dry` and saw nothing, only to realize that verbose mode is needed.

AthreyVinay: Makes sense, addressed here: [c13bcb5](https://github.com/teemtee/tmt/pull/4801/commits/c13bcb5ca13ce22f6c1bd635cb0990ed10429f13)

psss: Thanks!

psss (approved):
Works nice! Thanks for implementing this.

PR #4802, opened 2026-04-15
Handle test framework/check dependencies in required packages

### LINKED ISSUE #4796: Unable to find beakerlib error in centos-stream-9
Hi,

We have in the compose-tests [documentation](https://gitlab.com/CentOS/Integration/compose-tests) for centos-stream that the following command can be used to test things locally:

```sh
tmt -vv -c distro=centos-stream-9 run -a provision --how=virtual --image centos-stream-9 prepare --how=install --package=epel-release test --name /tests/sysstat
```

However, it no longer works. `tmt` tries to install `beakerlib` before `epel-release`, so it fails with "unable to find match: beakerlib" error. Is there a way to get `epel-release` installed before `beakerlib`? The tests do have `framework: beakerlib` in their `main.fmf`.

Thanks!

psss:
Instead of installing test framework and test check dependencies during the essential requires, let's handle them under the required packages as they are not needed before the actual test execution. This prevents having to provide an explicit `order` when enabling the `epel` repository on `centos-stream` to run `beakerlib` tests.

Fix #4796.

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation
* [x] include a release note


tmt/steps/prepare/__init__.py
LecrisUT: - Nit: Can move the snippet next to the `collected_requires`?
- What about the checks?

psss:
Simple plan and test to verify:

```yaml
/plan:
    discover:
        how: fmf
    provision:
        how: container
        image: centos:stream9
    prepare:
        how: feature
        epel: enabled
    execute:
        how: tmt

/test:
    test: . /usr/share/beakerlib/beakerlib.sh && rlJournalStart && rlJournalEnd
    framework: beakerlib
```

tmt/steps/prepare/__init__.py
```diff
@@ -279,9 +279,9 @@ def go(self, force: bool = False) -> None:
                     self._logger,
                 )
 
-                collected_essential_requires[
-                    guest
-                ].dependencies += test.test_framework.get_requirements(test, self._logger)
+                collected_requires[guest].dependencies += test.test_framework.get_requirements(
+                    test, self._logger
+                )
```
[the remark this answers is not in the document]
psss: Not needed, frameworks do not use `library` requirements.

tmt/steps/prepare/__init__.py
psss: > * Nit: Can move the snippet next to the `collected_requires`?

Yeah, makes sense grouped in 71eaaec75.

> * What about the checks?

Hm, good point. I believe they should be handled in the very same way. Included in 71eaaec75 as well.

tests/prepare/framework/test.sh
```diff
@@ -0,0 +1,23 @@
...
+    rlPhaseStartSetup
+        rlRun "run=\$(mktemp -d)" 0 "Create a run directory"
+        rlRun "pushd data"
+        rlRun "PROVISION_HOW=container"
+        build_container_image "centos/stream10/upstream:latest"
+    rlPhaseEnd
+
+    rlPhaseStartTest
+        image="$TEST_IMAGE_PREFIX/centos/stream10/upstream:latest"
+        rlRun "tmt run --id $run -vvv --all provision --how container --image $image"
```
LecrisUT: Can you add a comment short comment here as well about what this is trying to catch

tests/prepare/framework/main.fmf
```diff
@@ -0,0 +1,5 @@
+summary: Verify framework installation on centos-stream
+description: |
+    Make sure that a beakerlib test can be easily executed on
+    `centos-stream` with `epel` enabled without need for `order`
+    in the `prepare` step.
```
LecrisUT: The implementation is very specific to epel failure, but can this be generalized in what it is supposed to catch? Something like
```yaml
summary: Make sure framework dependencies are handled at the proper step/phase
```

tests/prepare/framework/data/main.fmf
```diff
@@ -0,0 +1,15 @@
...
+        how: container
+        image: centos:stream10
+    prepare:
+        how: feature
+        epel: enabled
+    execute:
+        how: tmt
+
+/test:
+    test: . /usr/share/beakerlib/beakerlib.sh && rlJournalStart && rlJournalEnd
```
LecrisUT:
```suggestion
    test: |
      . /usr/share/beakerlib/beakerlib.sh
      rlJournalStart
      rlJournalEnd
```

tmt/steps/prepare/__init__.py
```diff
@@ -272,12 +272,25 @@ def go(self, force: bool = False) -> None:
...
                 )
 
+                collected_requires[guest].dependencies += test.test_framework.get_requirements(
+                    test, self._logger
+                )
+
+                for check in test.check:
+                    collected_requires[guest].dependencies += check.plugin.essential_requires(
+                        guest, test, self._logger
+                    )
```
LecrisUT, line 290: The `essential_requires` name there is a bit divergent now

LecrisUT (approved):
Just some non-blocker suggestions

tests/prepare/framework/main.fmf
psss: Ok, done in f9518d4fa.

tests/prepare/framework/test.sh
psss: Sure, f9518d4fa.

tests/prepare/framework/data/main.fmf
psss: Updated in f9518d4fa.

tmt/steps/prepare/__init__.py
psss: Hm, right, `requires` would fit here better: f901e1c4f

happz: -1 from me, “essential” is, hm, essential, as in “not to be confused with your usual requirements”. They are essential for the tmt itself, for the given check to work, and they remain essential even when installed together with regular, ordinary, run-of-the-mill requirements.

LecrisUT: But we have a naming colision with the essential requirements that is defined in the non-test places. Any other non-conflicting names to choose from?

Behaviourly though they do act like any other run-of-the-mill requirements, it's just injected more automatically by us

happz: > But we have a naming colision with the essential requirements that is defined in the non-test places. Any other non-conflicting names to choose from?

I don't understand, where is the conflict, and what pieces are in conflict? Essential requirements are supposed to be defined in the non-test places, that's the point, "essential" requirements allowing tmt to even work, vs. "test" requirements.

LecrisUT: > Essential requirements are supposed to be defined in the non-test places, that's the point, "essential" requirements allowing tmt to even work, vs. "test" requirements.

Well yes, but aren't the `check` tied to the tests/execute step? From what I can see it is, and therefore the naming `essential_requires` does not fit as it collides with the essential requires in

https://github.com/teemtee/tmt/blob/b11683370af33caf11cba496ade5e5938c756c4e/tmt/guest/__init__.py#L2578-L2587

https://github.com/teemtee/tmt/blob/b11683370af33caf11cba496ade5e5938c756c4e/tmt/steps/__init__.py#L108

As opposed to the check one with now the same name but not tied to the above

https://github.com/teemtee/tmt/blob/b11683370af33caf11cba496ade5e5938c756c4e/tmt/checks/__init__.py#L247-L261

happz: > > Essential requirements are supposed to be defined in the non-test places, that's the point, "essential" requirements allowing tmt to even work, vs. "test" requirements.
> 
> Well yes, but aren't the `check` tied to the tests/execute step? From what I can see it is, and therefore the naming `essential_requires` does not fit as it collides with the essential requires in

As I see it, the "essential" bit is there for tmt components to let the machinery know what they absolutely need to work properly. Change of ordering does not change the semantics of `essential_requires`. To me, such a change means the documentation of those constants should be updated, to reflect "yeah, essential, but checks we decided to tie with test requirements", but it does not mean the check requirements are no longer essential just because they install in different point of workflow. They still are - in fact, I would also rename `TestFramework.get_requirements` to `TestFramework.essential_requires` because it's the same thing. I'm fine with changing the definition of the `PHASE_ORDER_PREPARE_INSTALL_ESSENTIAL_REQUIRES`, as it no longer includes checks, but IMO check requirements are still "essential", as in "tmt's own", as distinct from "test" requirements.

psss: Yeah, seems like the *required packages* terminology might welcome some clarification and cleanup. When I was thinking a bit more about the required packages categories, I see group related to `tmt` implementation itself:

* `essential_requires` ... required for essential `tmt` functionality, e.g. `/usr/bin/flock` or `bash`
* `plugin_requires` ... needed by enabled step plugins, e.g. `python3` for `prepare.ansible`

The rest seems to be closely related to tests:

* `test_requires` ... packages requires by tests themselves
* `framework_requires` ... test framework requires
* `check_requires` ... packages required by test checks

What about using the above-mentioned naming (or similar) to make the area more clear? I'd revert the last commit not to block this pull request and we could handle that in a separate one.

psss: Reverted name change in 9d610711c and filed issue as:

* #4839

happz (approved):

PR #4803, opened 2026-04-16
Implement the `sprint-sync` script

### LINKED ISSUE #4671: Sync github sprint to Jira sprint
Second part of the syncing automation. To be finished after the Jira Cloud migration.

psss:
Implement an easy way to sync github sprint items into the Jira sprint. Removed items are dropped, new items are added. Search for the right Jira issue is done using `remoteLinkUrl` but since it does not work always (e.g. when the index it outdated) fallback using the summary is included. Duplicate issues, when detected during the search, are reported as well.

Fix #4671.

Assisted-by: Claude

Pull Request Checklist

* [x] implement the feature
* [x] write the documentation

### DIFF
--- a/scripts/sprint-overview
+++ b/scripts/sprint-overview
@@ -24,6 +24,7 @@ from typing import Any, Optional
 import click
 import jinja2
 import requests
+from common import Item
 from ruamel.yaml import YAML
 
 # https://docs.github.com/en/rest/about-the-rest-api/breaking-changes
@@ -73,26 +74,6 @@ DISPLAY_TEMPLATE = jinja2.Template(
 ================================================================================
 """,
 )
-
-
-@dataclasses.dataclass
-class Item:
-    """
-    A sprint item (issue or pull request).
-    """
-
-    id: int
-    type: str
-    repo: str
-    status: str
-    size: Optional[int]
-    url: str
-    title: str
-
-    def __str__(self) -> str:
-        size_str = f"[{self.size}]" if self.size is not None else "[-]"
-        identifier = f"{self.repo}#{self.id}"
-        return f"{identifier:<10} {size_str:>4}  {self.title}"
 
 
 def github_api_get(
--- a/scripts/sprint-overview
+++ b/scripts/sprint-overview
@@ -235,14 +235,14 @@ def main(sprint: str, output_yaml: bool) -> None:
     """
     items = fetch_sprint_items(sprint)
 
-    if not items:
-        click.echo(f"No items found in sprint '{sprint}'.")
-        return
-
     if output_yaml:
         YAML().dump([dataclasses.asdict(item) for item in items], sys.stdout)
-    else:
+        return
+
+    if items:
         display_items(sprint, items)
+    else:
+        click.echo(f"No items found in sprint '{sprint}'.", err=True)
 
 
 if __name__ == "__main__":
--- a/scripts/sprint-overview
+++ b/scripts/sprint-overview
@@ -24,8 +24,9 @@ from typing import Any, Optional
 import click
 import jinja2
 import requests
-from common import Item
 from ruamel.yaml import YAML
+
+from common import Item  # isort: skip
 
 # https://docs.github.com/en/rest/about-the-rest-api/breaking-changes
 GITHUB_API_URL = "https://api.github.com"
--- a/scripts/sprint-sync
+++ b/scripts/sprint-sync
@@ -1,0 +1,309 @@
+#!/usr/bin/python3
+#
+# /// script
+# dependencies = [
+#   "click",
+#   "jinja2",
+#   "jira",
+#   "ruamel.yaml",
+# ]
+# ///
+
+"""
+Sync a Jira sprint with GitHub sprint items.
+"""
+
+import os
+import re
+import sys
+from typing import Optional, cast
+
+import click
+import jinja2
+from common import Item
+from jira import JIRA
+from jira.resources import Issue
+from ruamel.yaml import YAML
+
+ITEM_TEMPLATE = jinja2.Template(
+    source="""\
+  {{ key      }}   {{ summary }}
+  {{ decision }}   {{ url }}
+{% if duplicates %}\
+             duplicates found: {{ duplicates }}
+{% endif %}"""
+)
+
+DECISION_COLORS = {
+    "keep": "blue",
+    "add ": "green",
+    "drop": "red",
+    "miss": "yellow",
+}
+
+JIRA_SERVER = "https://redhat.atlassian.net/"
+JIRA_PROJECT = "TMT"
+JIRA_BOARD_ID = 1411
+
+
+def connect_to_jira() -> JIRA:
+    """
+    Connect to Jira using environment variables for authentication.
+
+    :returns: an authenticated :py:class:`JIRA` client.
+    """
+
+    server = os.environ.get("JIRA_SERVER", JIRA_SERVER)
+    email = os.environ.get("JIRA_EMAIL")
+    token = os.environ.get("JIRA_TOKEN")
+
+    if not email:
+        click.echo("Error: JIRA_EMAIL environment variable is not set.")
+        sys.exit(1)
+
+    if not token:
+        click.echo("Error: JIRA_TOKEN environment variable is not set.")
+        sys.exit(1)
+
+    return JIRA(server=server, basic_auth=(email, token))
+
+
+def fetch_jira_items(jira: JIRA, sprint_name: str) -> list[Issue]:
+    """
+    Fetch all items from the given sprint.
+
+    :param jira: authenticated Jira client.
+    :param sprint_name: name of the sprint.
+    :returns: list of Jira items
+    """
+
+    query = f'sprint = "{sprint_name}" ORDER BY key ASC'
+    items = cast(list[Issue], jira.search_issues(query, maxResults=False))
+    click.echo(f"Fetched {len(items)} items.\n")
+
+    return items
+
+
+def load_github_items(path: Optional[str] = None) -> list[Item]:
+    """
+    Load GitHub sprint items in sprint-overview --yaml format.
+
+    :param path: path to the YAML file, or ``None`` to read from stdin.
+    :returns: a list of :py:class:`Item` instances.
+    """
+
+    if path:
+        with open(path) as f:
+            data = YAML().load(f)
+    else:
+        data = YAML().load(sys.stdin)
+
+    if not data:
+        return []
+
+    return [Item(**item) for item in data]
+
+
+def get_upstream_url(jira: JIRA, issue: Issue) -> Optional[str]:
+    """
+    Get the "Upstream issue" remote link URL for a Jira issue.
+
+    :param jira: authenticated Jira client.
+    :param issue: a Jira issue object.
+    :returns: the upstream URL, or ``None`` if not found.
+    """
+
+    for link in jira.remote_links(issue):
+        if link.object.title == "Upstream issue":
+            return link.object.url
+
+    return None
+
+
+def find_sprint_id(jira: JIRA, sprint_name: str) -> Optional[int]:
+    """
+    Find the sprint ID by name on the TMT board.
+
+    :param jira: authenticated Jira client.
+    :param sprint_name: name of the sprint.
+    :returns: the sprint ID, or ``None`` if not found.
+    """
+
+    for sprint in jira.sprints(JIRA_BOARD_ID):
+        if sprint.name == sprint_name:
+            return sprint.id
+
+    return None
+
+
+def find_jira_issue(
+    jira: JIRA,
+    github_item: Item,
+) -> tuple[Optional[Issue], Optional[str]]:
+    """
+    Find a Jira issue that has a remote link matching the given GitHub item.
+
+    First searches by remote link URL, then falls back to title search.
+
+    :param jira: authenticated Jira client.
+    :param github_item: the GitHub :py:class:`Item` to search for.
+    :returns: a tuple of the matching Jira issue (or ``None``) and
+        a comma-separated string of duplicate keys (or ``None``).
+    """
+
+    def pick_item(results: list[Issue]) -> tuple[Optional[Issue], Optional[str]]:
+        """
+        Pick the oldest item, identify duplicates
+        """
+
+        if not results:
+            return None, None
+
+        duplicates = None
+        if len(results) > 1:
+            duplicates = ", ".join(result.key for result in results)
+
+        return min(results, key=lambda result: int(result.key.split("-")[-1])), duplicates
+
+    # Search by remote link URL (works only if already indexed)
+    query = f'project = "{JIRA_PROJECT}" AND remoteLinkUrl in ("{github_item.url}")'
+    results = cast(list[Issue], jira.search_issues(query, maxResults=False))
+    if results:
+        return pick_item(results)
+
+    # Fall back to title search and verify via remote links
+    title = github_item.title.replace('"', '')
+    query = f'project = "{JIRA_PROJECT}" AND summary ~ "{title}"'
+    candidates = cast(list[Issue], jira.search_issues(query, maxResults=False))
+
+    matched = [
+        candidate
+        for candidate in candidates
+        if get_upstream_url(jira, candidate) == github_item.url
+    ]
+    return pick_item(matched)
+
+
+def sync_sprint(
+    jira: JIRA,
+    sprint_id: int,
+    jira_items: list[Issue],
+    github_items: list[Item],
+    dry: bool = False,
+) -> None:
+    """
+    Sync Jira sprint with GitHub sprint items.
+
+    Removes Jira issues not present in the GitHub sprint and adds
+    missing GitHub items into the Jira sprint.
+
+    :param jira: authenticated Jira client.
+    :param sprint_id: ID of the sprint.
+    :param jira_items: list of Jira issues currently in the sprint.
+    :param github_items: list of GitHub :py:class:`Item` instances.
+    :param dry: if ``True``, only show what would be done without
+        making any changes.
+    """
+
+    github_items_by_url = {item.url: item for item in github_items}
+    github_urls = set(github_items_by_url.keys())
+    matched_urls: set[str] = set()
+
+    def show_item(
+        key: str,
+        decision: str,
+        summary: str,
+        url: Optional[str] = None,
+        duplicates: Optional[str] = None,
+    ) -> None:
+        """
+        Render a single item, color decisions, remove prefix
+        """
+
+        click.echo(
+            ITEM_TEMPLATE.render(
+                key=f"{key:<8}",
+                decision=click.style(f"{decision:<8}", fg=DECISION_COLORS.get(decision)),
+                summary=re.sub(r'^\[teemtee/[^\]]*\]\s*', '', summary),
+                url=url,
+                duplicates=duplicates,
+            )
+        )
+
+    # Step 1: Check existing Jira items — keep or remove
+    for jira_item in jira_items:
+        url = get_upstream_url(jira, jira_item)
+
+        if url and url in github_urls:
+            matched_urls.add(url)
+            show_item(jira_item.key, "keep", jira_item.fields.summary, url)
+        else:
+            show_item(jira_item.key, "drop", jira_item.fields.summary, url)
+            if not dry:
+                jira.move_to_backlog([jira_item.key])
+
+    # Step 2: Add missing GitHub items
+    missing_urls = github_urls - matched_urls
+    if not missing_urls:
+        return
+
+    for url in missing_urls:
+        github_item = github_items_by_url[url]
+        jira_item, duplicates = find_jira_issue(jira, github_item)
+
+        if jira_item:
+            show_item(jira_item.key, "add ", jira_item.fields.summary, url, duplicates)
+            if not dry:
+                jira.add_issues_to_sprint(sprint_id, [jira_item.key])
+        else:
+            show_item("???", "miss", github_item.title, url)
+
+
+@click.command()
+@click.option(
+    "--sprint",
+    metavar="NAME",
+    required=True,
+    help="Name of the sprint, e.g. 'TMT Sprint 11'.",
+)
+@click.option(
+    "--dry",
+    is_flag=True,
+    default=False,
+    help="Only show what would be done without making any changes.",
+)
+@click.argument("yaml_file", required=False, default=None)
+def main(sprint: str, dry: bool, yaml_file: Optional[str]) -> None:
+    """
+    Sync a Jira sprint with GitHub sprint items.
+
+    Reads GitHub sprint items in YAML format (from sprint-overview --yaml)
+    from YAML_FILE or stdin. Example usage:
+
+    \b
+        ./sprint-overview --sprint 'Sprint 11' --yaml | ./sprint-sync --sprint 'TMT Sprint 11'
+
+    Authentication is configured via environment variables:
+
+    \b
+        JIRA_SERVER ... Jira server URL, https://redhat.atlassian.net/ by default
+        JIRA_EMAIL .... User email for authentication
+        JIRA_TOKEN .... Personal access token
+    """
+
+    click.echo(f"\nSyncing sprint: {sprint}{' (dry mode)' if dry else ''}")
+
+    jira = connect_to_jira()
+
+    sprint_id = find_sprint_id(jira, sprint)
+    if not sprint_id:
+        click.echo(f"Error: sprint '{sprint}' not found.")
+        sys.exit(1)
+
+    github_items = load_github_items(yaml_file)
+    jira_items = fetch_jira_items(jira, sprint)
+    sync_sprint(jira, sprint_id, jira_items, github_items, dry=dry)
+
+
+if __name__ == "__main__":
+    main()
--- a/scripts/sprint-sync
+++ b/scripts/sprint-sync
@@ -117,7 +117,7 @@ def get_upstream_url(jira: JIRA, issue: Issue) -> Optional[str]:
     return None
 
 
-def find_sprint_id(jira: JIRA, sprint_name: str) -> Optional[int]:
+def find_sprint_id(jira: JIRA, sprint_name: str) -> int:
     """
     Find the sprint ID by name on the TMT board.
 
@@ -130,7 +130,8 @@ def find_sprint_id(jira: JIRA, sprint_name: str) -> Optional[int]:
         if sprint.name == sprint_name:
             return sprint.id
 
-    return None
+    click.echo(f"Sprint '{sprint_name}' not found.")
+    sys.exit(1)
 
 
 def find_jira_issue(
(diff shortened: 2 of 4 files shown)


psss:
Here's an example output:

sprint-sync

scripts/sprint-sync
```diff
@@ -0,0 +1,311 @@
...
+        """
+
+        if not results:
+            return None, None
+
+        duplicates = None
+        if len(results) > 1:
+            duplicates = ", ".join(result.key for result in results)
+
+        return min(results, key=lambda result: int(result.key.split("-")[1])), duplicates
```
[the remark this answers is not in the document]
psss: Ok, why not.

scripts/sprint-overview
```diff
@@ -24,6 +24,7 @@ from typing import Any, Optional
 import click
 import jinja2
 import requests
+from common import Item
```
happz: Hm, would it be better to use a relative import, `.common`? As simple `common`, positioned between `ruamel` and `requests`, it looks like a 3rd party package.

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+            show_item("???", "miss", github_item.title, url)
+
+
+@click.command()
+@click.option(
+    "--sprint",
+    metavar="NAME",
+    required=True,
+    help="Name of the sprint, e.g. 'TMT Sprint 11'.",
+)
```
LecrisUT, line 281: Uph, the separation in the naming conventions used is not nice. If it is well defined format, can the jira name be automated? That would also allow us to use `@current-X` aliases

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+@click.option(
+    "--dry",
+    is_flag=True,
+    default=False,
+    help="Only show what would be done without making any changes.",
+)
+@click.argument("yaml_file", required=False, default=None)
```
LecrisUT: Let's use click's built-in handling of stdin https://click.palletsprojects.com/en/stable/handling-files/#file-arguments

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+    """
+
+    click.echo(f"\nSyncing sprint: {sprint}{' (dry mode)' if dry else ''}")
+
+    jira = connect_to_jira()
+
+    sprint_id = find_sprint_id(jira, sprint)
+    if not sprint_id:
+        click.echo(f"Error: sprint '{sprint}' not found.")
+        sys.exit(1)
```
LecrisUT: Can be moved inisde the `find_sprint_id`

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+    """
+
+    if path:
+        with open(path) as f:
+            data = YAML().load(f)
+    else:
+        data = YAML().load(sys.stdin)
+
+    if not data:
+        return []
```
LecrisUT: Should error out if this is the case

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+    return items
+
+
+def load_github_items(path: Optional[str] = None) -> list[Item]:
+    """
+    Load GitHub sprint items in sprint-overview --yaml format.
+
+    :param path: path to the YAML file, or ``None`` to read from stdin.
+    :returns: a list of :py:class:`Item` instances.
```
LecrisUT:
```suggestion
    :returns: a list of :py:class:`common.Item` instances.
```
imports are not considered when referencing a sphinx object. Could also avoid this by not being that technical and word it as something like "list of sprint items"

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+    Fetch all items from the given sprint.
+
+    :param jira: authenticated Jira client.
+    :param sprint_name: name of the sprint.
+    :returns: list of Jira items
+    """
+
+    query = f'sprint = "{sprint_name}" ORDER BY key ASC'
+    items = cast(list[Issue], jira.search_issues(query, maxResults=False))
+    click.echo(f"Fetched {len(items)} items.\n")
```
LecrisUT, line 120: Why the various new-lines?

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+    """
+    Sync Jira sprint with GitHub sprint items.
+
+    Removes Jira issues not present in the GitHub sprint and adds
+    missing GitHub items into the Jira sprint.
+
+    :param jira: authenticated Jira client.
+    :param sprint_id: ID of the sprint.
+    :param jira_items: list of Jira issues currently in the sprint.
+    :param github_items: list of GitHub :py:class:`Item` instances.
```
LecrisUT: docstring reads weird. Maybe better "list of Github project sprint items"?

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+        click.echo(
+            ITEM_TEMPLATE.render(
+                key=f"{key:<8}",
+                decision=click.style(f"{decision:<8}", fg=DECISION_COLORS.get(decision)),
+                summary=re.sub(r'^\[teemtee/[^\]]*\]\s*', '', summary),
+                url=url,
+                duplicates=duplicates,
+            )
+        )
```
LecrisUT: The definitions are very scattered. Would be better to consolidate them into a single class/dataclass

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+    for jira_item in jira_items:
+        url = get_upstream_url(jira, jira_item)
+
+        if url and url in github_urls:
+            matched_urls.add(url)
+            show_item(jira_item.key, "keep", jira_item.fields.summary, url)
+        else:
+            show_item(jira_item.key, "drop", jira_item.fields.summary, url)
+            if not dry:
+                jira.move_to_backlog([jira_item.key])
```
LecrisUT: This seems wrong. There should be 3 cases:
- upstream url exists and it's in sprint: update
- upstream url exists and it's not in sprint: drop from sprint, move to backlog, something else
- upstream url does not exist: unclear, do manual action?

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+    :param sprint_id: ID of the sprint.
+    :param jira_items: list of Jira issues currently in the sprint.
+    :param github_items: list of GitHub :py:class:`Item` instances.
+    :param dry: if ``True``, only show what would be done without
+        making any changes.
+    """
+
+    github_items_by_url = {item.url: item for item in github_items}
+    github_urls = set(github_items_by_url.keys())
+    matched_urls: set[str] = set()
```
LecrisUT: Can be combined to a single `urls_to_process`

scripts/sprint-sync
LecrisUT: The logic done here does not match the description "Sync a Jira sprint with GitHub sprint items.". It does not really sync the sate. We do have the other sync to jira service and I am not sure how these two will integrate with each other. Some clearer description would be appreciated.

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+        """
+        Pick the oldest item, identify duplicates
+        """
+
+        if not results:
+            return None, None
+
+        duplicates = None
+        if len(results) > 1:
+            duplicates = ", ".join(result.key for result in results)
```
LecrisUT: Would prefer this to be handled in the jinja and not here

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+        return min(results, key=lambda result: int(result.key.split("-")[-1])), duplicates
```
LecrisUT: This whole function seems excessive. Could be handled in `sync_sprint` also

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+    # Search by remote link URL (works only if already indexed)
+    query = f'project = "{JIRA_PROJECT}" AND remoteLinkUrl in ("{github_item.url}")'
+    results = cast(list[Issue], jira.search_issues(query, maxResults=False))
+    if results:
+        return pick_item(results)
```
LecrisUT: Can you elaborate the "if already indexed"? When does it fail?

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+        duplicates = None
+        if len(results) > 1:
+            duplicates = ", ".join(result.key for result in results)
+
+        return min(results, key=lambda result: int(result.key.split("-")[-1])), duplicates
+
+    # Search by remote link URL (works only if already indexed)
+    query = f'project = "{JIRA_PROJECT}" AND remoteLinkUrl in ("{github_item.url}")'
+    results = cast(list[Issue], jira.search_issues(query, maxResults=False))
```
LecrisUT: The `cast` are unnecessary, they have `overload` that should be handling it

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+    First searches by remote link URL, then falls back to title search.
+
+    :param jira: authenticated Jira client.
+    :param github_item: the GitHub :py:class:`Item` to search for.
+    :returns: a tuple of the matching Jira issue (or ``None``) and
+        a comma-separated string of duplicate keys (or ``None``).
+    """
+
+    def pick_item(results: list[Issue]) -> tuple[Optional[Issue], Optional[str]]:
```
happz: Can a list of duplicates be returned instead? Joining them into a string means it's encoding the presentation requirement into the value - keys of duplicates can (and should) be merged when being presented, not before. It's a list after all.

AthreyVinay:
@psss A very naive question: Is syncing GitHub sprints to Jira a problem other Red Hat teams face too? I'd imagine so — is there any company-wide effort to solve this, or is this intentionally a quick team-specific solution for now?

scripts/sprint-overview
psss: Hmmm, seems that relative imports do not work with standalone scripts:

    ImportError: attempted relative import with no known parent package

Any suggestions? Something like this would not be much nicer, right?

```python
_dir = os.path.dirname(os.path.abspath(__file__))
sys.path.insert(0, os.path.dirname(_dir))
__package__ = os.path.basename(_dir)
```

Or just move the line and add an `unsorted-imports` ignore?

happz: I would avoid `sys.path` manipulation. Can you add `__init__.py` and try once more? If it wouldn't help, I'd use the waiver then.

scripts/sprint-sync
psss: > Uph, the separation in the naming conventions used is not nice.

Yeah, unique name would be needed for the issue search.

> If it is well defined format, can the jira name be automated? That would also allow us to use @current-X aliases

I'm not sure this is worth investing time, `sprint-sync` does not access github at all so it knows nothing about the `@current` aliases.

scripts/sprint-overview
LecrisUT: Huh, something is funky, they should work just fine. Is it because the file name does not have a `.py` suffix?

psss: Tried adding `__init__.py`, did not help. Seems this is an intentional behavior:

* https://stackoverflow.com/questions/79427581/reason-for-no-relative-imports-in-python-scripts

So... we go with ignore?

LecrisUT: Alternatively it can be merged in `sprint` as a sub-command?

scripts/sprint-sync
LecrisUT: > > Uph, the separation in the naming conventions used is not nice.
> 
> Yeah, unique name would be needed for the issue search.

Unique name with respect to what namespace? What is the naming convention?

> > If it is well defined format, can the jira name be automated? That would also allow us to use @Current-X aliases
> 
> I'm not sure this is worth investing time, `sprint-sync` does not access github at all so it knows nothing about the `@current` aliases.

Sure, but extending it to support it is not difficult. The context around the naming convention, how tied these scripts are with the current project vs extending it to `fmf`, etc. does play into if it makes sense to support it or not.

scripts/sprint-sync
psss: If the sprint is empty, it should sync it to any empty one. I think it's a valid use case.

scripts/sprint-sync
psss: Just to make it more beautiful :)

scripts/sprint-sync
LecrisUT: The expected output format from 
```yaml
    if output_yaml:
        YAML().dump([dataclasses.asdict(item) for item in items], sys.stdout)
```
guarantees that the schema there is is at least a empty list
```python
>>> from ruamel.yaml import YAML
... import sys
... YAML().dump([item for item in items], sys.stdout)
[]
```
If that promise is broken then we have some big fish to tackle

scripts/sprint-sync
psss: > * upstream url exists and it's in sprint: update

This should be `keep`, we do not do any update.

> * upstream url exists and it's not in sprint: drop from sprint, move to backlog, something else

Covered by the `else` branch.

> * upstream url does not exist: unclear, do manual action?

Do we expect any manually added items? I'd say, that anything which does not have an associated github item url should be just dropped from the sprint.

LecrisUT: > > * upstream url does not exist: unclear, do manual action?
> 
> Do we expect any manually added items? I'd say, that anything which does not have an associated github item url should be just dropped from the sprint.

I expect it to at least scream at me and let me know that something unexpected was encountered.

scripts/sprint-sync
psss: Indexing all the issues on the Jira Cloud server can take several days. Last time they restarted it and it took about 7 days to complete.

scripts/sprint-sync
psss: Without the `cast` it fails with:

```
  /home/psss/git/tmt/scripts/sprint-sync:172:26 - error: Argument of type "dict[str, Any] | ResultList[Issue]" cannot be assigned to parameter "results" of type "list[Issue]" in function "pick_item"
    Type "dict[str, Any] | ResultList[Issue]" is not assignable to type "list[Issue]"
      "dict[str, Any]" is not assignable to "list[Issue]" (reportArgumentType)
```

happz: `cast()` is dangerous: is `Issue` an alias for `dict[str, Any]`? If not, you might be allowing unexpected type in, claiming it's `Issue`.

IIUIC, `search_issues()` will return `list[Issue]` only when called with `json_result=False`. It's the default, true, but maybe pyright needs a bit of help to narrow the types.

scripts/sprint-sync
```diff
@@ -0,0 +1,311 @@
...
+    for jira_item in jira_items:
+        url = get_upstream_url(jira, jira_item)
+
+        if url and url in github_urls:
+            matched_urls.add(url)
+            show_item(jira_item.key, "keep", jira_item.fields.summary, url)
+        else:
+            show_item(jira_item.key, "drop", jira_item.fields.summary, url)
+            if not dry:
+                jira.move_to_backlog([jira_item.key])
```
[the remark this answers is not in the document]
psss: Done in 8732cb3a8.

scripts/sprint-sync
```diff
@@ -0,0 +1,311 @@
...
+    for url in missing_urls:
+        github_item = github_items_by_url[url]
+        jira_item, duplicates = find_jira_issue(jira, github_item)
+
+        if jira_item:
+            show_item(jira_item.key, "add ", jira_item.fields.summary, url, duplicates)
+            if not dry:
+                jira.add_issues_to_sprint(sprint_id, [jira_item.key])
+        else:
+            show_item("???", "miss", github_item.title, url)
```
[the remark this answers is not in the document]
psss: Done in 8732cb3a8.

scripts/sprint-sync
LecrisUT: What jira version do you have installed locally? https://github.com/pycontribs/jira/blob/b07a89bc3401094fdc4c2c2c6f125855fde0a10e/jira/client.py#L3568 if any of the type-checker cannot work that out we are in big trouble

Edit: added in [3.9.0](https://github.com/pycontribs/jira/commit/9900396455e749de7c7ba97801cee024c8c1aeec)

scripts/sprint-sync
```diff
@@ -0,0 +1,309 @@
...
+
+    # Search by remote link URL (works only if already indexed)
+    query = f'project = "{JIRA_PROJECT}" AND remoteLinkUrl in ("{github_item.url}")'
+    results = cast(list[Issue], jira.search_issues(query, maxResults=False))
+    if results:
+        return pick_item(results)
+
+    # Fall back to title search and verify via remote links
+    title = github_item.title.replace('"', '')
+    query = f'project = "{JIRA_PROJECT}" AND summary ~ "{title}"'
```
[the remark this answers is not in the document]
psss: Done in c09a1b539.

scripts/sprint-overview
psss: I would go with c0e0c7d2b.

scripts/sprint-sync
psss: Updated in c55997cfb.

scripts/sprint-sync
psss: Moved: cde895e07

scripts/sprint-sync
psss: Done: 31bb03f8f

scripts/sprint-sync
psss: Done: 31bb03f

scripts/sprint-sync
psss: What do you mean by scattered definitions?

scripts/sprint-sync
psss: Done in 0e0b81cc6.

scripts/sprint-sync
psss: Done here: 56f1a09ce.

scripts/sprint-sync
psss: Updated `find_jira_issues()` to return a list of all found issues in 1d5190af1.

scripts/sprint-sync
psss: I've updated `find_jira_issues()` to return a list of all found issues in 1d5190a.

psss:
> Is syncing GitHub sprints to Jira a problem other Red Hat teams face too? I'd imagine so — is there any company-wide effort to solve this, or is this intentionally a quick team-specific solution for now?

Good question, @AthreyVinay! Yeah, this is basically just a quick hack for us. I don't know of any other team. I expect that most teams do the planning activities directly in Jira. But let's [check](https://redhat-internal.slack.com/archives/C058M9PBGEM/p1776700354535389)! Maybe there's more people with a similar use case and we could collaborate on this.

scripts/sprint-sync
psss: > What jira version do you have installed locally?

```
> pip freeze|grep jira
jira==3.10.5
```

This seems to make `pyright` happy:

```python
results = jira.search_issues(query, maxResults=False, json_result=False)
assert isinstance(results, list)
```

Or any better ideas?

AthreyVinay:
> Maybe there's more people with a similar use case and we could collaborate on this.

Makes sense - the ask looks good :)

scripts/sprint-sync
psss: Well, that is actually not the case because of:

https://github.com/teemtee/tmt/blob/b8f4602b2712a6e6daea9925dd167b5d9b5fb715/scripts/sprint-overview#L256-L258

We can change that though if you really think it's worth it.

scripts/sprint-sync
psss: Usage updated to be more specific: 80e9967ac

psss: Ok, let's report such issues as `lost` instead: 111d99407

scripts/sprint-sync
psss: What about just using consistent names from the command line: 87f115274 ?

scripts/sprint-sync
psss: With the above-mentioned asserts it would look like this: 7bd556bbd and `pyright` seems to be happy.

LecrisUT: How do you run `pyright`? From pre-commit? You might need to purge the pre-commit caches to pick up the new jira version. I tried to run pyright directly manually and it seemed fine. But a narrow-type assert is fine also.

scripts/sprint-sync
LecrisUT: Could you add that context in the comment? Wasn't this just an issue for the intial sync? This sounds like something that jira should handle by themselves efficiently behind the scene or be filed as a bug.

scripts/sprint-sync
LecrisUT: :+1: that was what I was referring in
> If it is well defined format, can the jira name be automated?

scripts/sprint-sync
LecrisUT: Well, isn't that even worse? Because that would output that string to stdout and process it as the yaml content :face_with_peeking_eye:. One way or another this path needs to be fixed, maybe either:
- `click.echo` to stderr only
- do not support piping, and force the usage of an actual file

scripts/sprint-sync
LecrisUT: Well, can at least make it consistent and only new-line at the end, not at the beginning?

scripts/sprint-sync
LecrisUT: Having to scroll back-and-forth from `show_item` in `sync_sprint` to `ITEM_TEMPLATE` and `DECSION_COLORS`. But these are logically connected to each other and can be combined, especially since this `show_item` does not need to be a nested function.

scripts/sprint-sync
psss: Ok, not needed at the beginning: 9c523517d

scripts/sprint-sync
psss: > Well, isn't that even worse? Because that would output that string to stdout and process it as the yaml content 🫣.

Worse? I was not suggesting it should be changed in this way. Just describing the current status. Ok, what about this: f80e5f40e ?

scripts/sprint-sync
psss: Added more context here: ab878d5e7

> Wasn't this just an issue for the intial sync?

Not sure, as it was apparently restarted several times. I would not rely on it.

scripts/sprint-sync
psss: @happz, this one should be covered.

scripts/sprint-overview
psss: @happz, is this one good enough?

scripts/sprint-sync
LecrisUT: By "worse" I mean, "aren't we in an even worse situation?". But that looks good enough workaround, thanks.

LecrisUT (approved):
Looks good enough to me now.

scripts/sprint-overview
happz: Yep, resolving.

happz (approved):

happz:
@psss "write the documentation" checkbox is not complete yet, is it something you plan to address?

scripts/sprint-sync
psss: Ok, understood. Added a `SyncItem` class in 275ce9327.

LecrisUT: Perfect, thanks :)

psss:
> @psss "write the documentation" checkbox is not complete yet, is it something you plan to address?

Ah, right, forgot. Thanks for reminding. Wanted to add a brief documentation mentioning the two sprint scripts. Done in aa3a4a0f2.

scripts/sprint-sync
```diff
@@ -0,0 +1,325 @@
...
+    Find the sprint ID by name on the TMT board.
+
+    :param jira: authenticated Jira client.
+    :param sprint_name: name of the sprint.
+    :returns: the sprint ID, or ``None`` if not found.
+    """
+
+    for sprint in jira.sprints(JIRA_BOARD_ID):
+        if sprint.name == sprint_name:
+            return sprint.id
```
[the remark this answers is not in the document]
psss: Yeah, forgot to update the docstring, fixed in 6796e44c0.

scripts/sprint-sync
```diff
@@ -0,0 +1,325 @@
...
+
+    server = os.environ.get("JIRA_SERVER", JIRA_SERVER)
+    email = os.environ.get("JIRA_EMAIL")
+    token = os.environ.get("JIRA_TOKEN")
+
+    if not email:
+        click.echo("Error: JIRA_EMAIL environment variable is not set.")
+        sys.exit(1)
+
+    if not token:
```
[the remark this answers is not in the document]
psss: Ack, makes sense to consistently use `stderr` for all error messages, changed in 6796e44c0.

scripts/sprint-sync
```diff
@@ -0,0 +1,325 @@
...
+    """
+
+    github_items_by_url = {item.url: item for item in github_items}
+    urls_to_process = set(github_items_by_url.keys())
+
+    items_to_drop: list[str] = []
+    items_to_add: list[str] = []
+
+    # Step 1: Check existing Jira items — keep or remove
+    for jira_item in jira_items:
```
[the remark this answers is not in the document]
psss: Why not, `sorted` in 6796e44c0.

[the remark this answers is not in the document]
LecrisUT: That is a no-op. `set` is always sorted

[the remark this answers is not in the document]
psss: Hm, that does not seem to be true:

https://docs.python.org/3/tutorial/datastructures.html#sets

    Because sets are unordered, iterating over them or printing
    them can produce the elements in a different order than you
    expect.

https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset

    Being an unordered collection, sets do not record element
    position or order of insertion. Accordingly, sets do not
    support indexing, slicing, or other sequence-like behavior.

AthreyVinay (approved):

scripts/sprint-sync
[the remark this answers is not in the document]
LecrisUT: That is likely a misread. It is unordered with respect to the original order that you constructed with, i.e. compared to the list ordering.
```python
>>> set_a = set([4,1])
... set_a.add(3)
... for i in set_a:
...     print(i)
... set_b = set([1,4])
... set_b.add(3)
... for i in set_b:
...     print(i)
1
3
4
1
3
4
```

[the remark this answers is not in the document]
happz: That might be just a happy accident or an implementation detail. Python explicitly states `set` is an unordered collection, order of items when iterating over the set is not guaranteed to be stable or reproducible: https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset. Your example will probably remain same next week, but it's most likely thanks to the implementation, items being presented in the order of their hash modulo set size, according to some comments. For the same reason, sets do not support indexing or slicing, there is no guaranteed order of items.

If we want to iterate over set items in some order, we really should enforce the ordering.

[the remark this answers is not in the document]
LecrisUT: The unordered collection refers to
> Being an unordered collection, sets do not record element position or order of insertion.

The `set` itself is ordered in the sorted sense. [cpp set](https://cppreference.com/cpp/container/set) uses a similar language distinguishing ordered vs sorted.

Edit: hmm maybe more complicated. They do use comparison [^1], but they *may* also use hashing. I am not sure what the explicit logic is there

[^1]: https://github.com/python/cpython/blob/f93834ff01128774532c101c574e47c6c0418540/Objects/setobject.c#L288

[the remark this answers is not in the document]
thrix: I am resolving this, thanks.

happz:
Script-only, no need for more tests. Merging.

PR #4804, opened 2026-04-17
Add `whatprovides` resolution to `verify-install` plugin

### LINKED ISSUE #4714: Link the Artifact Verification Plugin for discover only Providers

vaibhavdaren:
Closes  #4714 
Use whatprovides lookup for requirements:

* Artifact verification now correctly handles virtual provides and file paths (like `/usr/bin/make`.
* Added a `resolve_provides` method to the package manager interface.
* Implemented the resolution for DNF via `repoquery --whatprovides` (legacy YUM is not supported).
* Removed an outdated TODO about missing virtual provide support.
* Updated tests to use file paths instead of hardcoded package names to verify the fix.

### DIFF
--- a/tests/prepare/artifact/providers/koji-build/data/main.fmf
+++ b/tests/prepare/artifact/providers/koji-build/data/main.fmf
@@ -17,4 +17,4 @@
     duration: 5m
     summary: Verify package from koji build
     require:
-      - make
+      - /usr/bin/make
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -3,7 +3,6 @@ import configparser
 import enum
 import re
 import shlex
-from collections import defaultdict
 from collections.abc import Iterable, Iterator
 from typing import (
     TYPE_CHECKING,
@@ -15,6 +14,7 @@ from typing import (
     TypedDict,
     TypeVar,
     Union,
+    cast,
 )
 from urllib.parse import urlparse
 
@@ -423,7 +423,7 @@ class PackageManagerEngine(tmt.utils.Common):
     def resolve_provides(
         self,
         provides: Iterable[str],
-        repo_ids: Iterable[str],
+        repo_ids: Iterable[str] = (),
     ) -> ShellScript:
         """
         Resolves each provide to the NEVRAs of packages that provide it.
@@ -443,7 +443,8 @@ class PackageManagerEngine(tmt.utils.Common):
                   repo_id: 'fedora'
 
         :param provides: Provides to resolve.
-        :param repo_ids: Limit the query to these repository IDs.
+        :param repo_ids: Restrict the query to these repository IDs; searches all enabled
+            repositories when not provided.
         :returns: A shell script emitting the YAML mapping described above.
         :raises PrepareError: If the package manager does not support this query.
         """
@@ -584,19 +585,17 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
             result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
         return result
 
-    # String annotation avoids circular import: _rpm.py imports Version from this module.
-    # NOTE: this method is only meaningful for RPM-based package managers; non-RPM engines
-    # raise PrepareError from their engine.resolve_provides implementation.
     def resolve_provides(
         self,
         provides: Iterable[str],
-        repo_ids: Iterable[str],
-    ) -> 'dict[str, list[RpmVersion]]':
+        repo_ids: Iterable[str] = (),
+    ) -> dict[str, list['RpmVersion']]:
         """
         Map each provide to the :py:class:`RpmVersion` objects of packages that provide it.
 
         :param provides: Provides to resolve.
-        :param repo_ids: Limit the query to these repository IDs.
+        :param repo_ids: Restrict the query to these repository IDs; searches all enabled
+            repositories when not provided.
         :returns: Mapping from each provide to a list of :py:class:`RpmVersion` objects,
             each carrying the NEVRA and source repository. Every requested provide appears
             as a key; provides with no match map to an empty list.
@@ -604,25 +603,21 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
         from tmt.package_managers._rpm import RpmVersion
 
         provides_list = list(provides)
-        script = self.engine.resolve_provides(provides_list, repo_ids=repo_ids)
-        output = self.guest.execute(script)
+        output = self.guest.execute(self.engine.resolve_provides(provides_list, repo_ids=repo_ids))
 
-        result: defaultdict[str, list[RpmVersion]] = defaultdict(list)
+        result: dict[str, list[RpmVersion]] = {provide: [] for provide in provides_list}
 
-        provides_yaml: Optional[dict[str, Optional[list[_ResolvedEntry]]]] = tmt.utils.from_yaml(
-            output.stdout or ''
+        provides_yaml: dict[str, Optional[list[_ResolvedEntry]]] = tmt.utils.from_yaml(
+            cast(str, output.stdout)
         )
-        if provides_yaml is None:
-            return result
 
         for provide, nevras in provides_yaml.items():
             if nevras is None:
                 self.info(f"Nothing provides '{provide}'.")
                 continue
-            resolved: list[RpmVersion] = []
             for resolved_provide in nevras:
                 try:
-                    resolved.append(
+                    result[provide].append(
                         RpmVersion.from_nevra(
                             resolved_provide['nevra'], repo_id=resolved_provide['repo_id']
                         )
@@ -631,7 +626,6 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
                     raise PrepareError(
                         f"Cannot parse '{resolved_provide}' for provide '{provide}': {exc}"
                     ) from exc
-            result[provide] = resolved
 
         return result
 
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -14,7 +14,6 @@ from typing import (
     TypedDict,
     TypeVar,
     Union,
-    cast,
 )
 from urllib.parse import urlparse
 
@@ -607,8 +606,9 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
 
         result: dict[str, list[RpmVersion]] = {provide: [] for provide in provides_list}
 
+        assert output.stdout is not None  # narrow type
         provides_yaml: dict[str, Optional[list[_ResolvedEntry]]] = tmt.utils.from_yaml(
-            cast(str, output.stdout)
+            output.stdout
         )
 
         for provide, nevras in provides_yaml.items():
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -274,13 +274,6 @@ def find_package_manager(
     return plugin
 
 
-def is_file_or_virtual_provide(requirement: str) -> bool:
-    """
-    Return ``True`` if *requirement* is a file path or virtual provide rather than a plain package.
-    """
-    return requirement.startswith('/') or '(' in requirement
-
-
 def escape_installables(*installables: Installable) -> Iterator[str]:
     for installable in installables:
         yield shlex.quote(str(installable))
@@ -314,10 +307,6 @@ class Options:
 class PackageManagerEngine(tmt.utils.Common):
     command: Command
     options: Command
-
-    #: Sentinel line emitted by :py:meth:`resolve_provides` shell scripts to
-    #: separate NEVRA groups for consecutive provides.
-    RESOLVE_PROVIDES_SENTINEL: ClassVar[str] = '---'
 
     def __init__(self, *, guest: 'Guest', logger: tmt.log.Logger) -> None:
         super().__init__(logger=logger)
@@ -411,22 +400,24 @@ class PackageManagerEngine(tmt.utils.Common):
         raise NotImplementedError
 
     @abc.abstractmethod
-    def resolve_provides(self, provides: Iterable[str]) -> ShellScript:
+    def resolve_provides(self, provides: list[str]) -> ShellScript:
         """
         Resolve each provide to the NEVRAs of packages that provide it.
 
-        The script must emit zero or more NEVRA lines per provide followed by a
-        :py:attr:`PackageManagerEngine.RESOLVE_PROVIDES_SENTINEL` line, one group per input provide
-        in the same order as the input.  An empty group (sentinel only) means
-        nothing provides that capability.
+        The script must emit YAML mapping each provide string to a list of NEVRA
+        strings (or null when nothing provides it), e.g.::
+
+            '/usr/bin/cmake':
+            - 'cmake-0:3.31.6-4.fc43.x86_64'
+            'make':
+            - 'make-1:4.4.1-8.fc43.x86_64'
 
         :param provides: Provides to resolve, e.g. package names, file paths, or
             virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
-        :returns: A shell script that emits NEVRA lines grouped by provide, each
-            group terminated by :py:attr:`PackageManagerEngine.RESOLVE_PROVIDES_SENTINEL`.
-        :raises NotImplementedError: If the package manager does not support this query.
+        :returns: A shell script that emits a YAML mapping of provide → NEVRA list.
+        :raises PrepareError: If the package manager does not support this query.
         """
-        raise NotImplementedError
+        raise PrepareError("Package manager does not support provides resolution.")
 
     def create_repository(self, directory: Path) -> ShellScript:
         """
@@ -563,7 +554,7 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
             result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
         return result
 
-    def resolve_provides(self, provides: Iterable[str]) -> dict[str, set[Version]]:
+    def resolve_provides(self, provides: list[str]) -> dict[str, set[Version]]:
         """
         Map each provide to the :py:class:`Version` objects of packages that provide it.
 
@@ -576,30 +567,24 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
         """
         from tmt.package_managers._rpm import RpmVersion
 
-        provides_list = list(provides)
-        result: dict[str, set[Version]] = {p: set() for p in provides_list}
-        if not provides_list:
-            return result
+        if not provides:
+            return {}
 
-        script = self.engine.resolve_provides(iter(provides_list))
+        script = self.engine.resolve_provides(provides)
         stdout = self.guest.execute(script).stdout or ''
+        parsed: dict[str, Optional[list[str]]] = tmt.utils.from_yaml(stdout) or {}
 
-        groups: list[list[str]] = []
-        current: list[str] = []
-        for line in stdout.splitlines():
-            line = line.strip()
-            if line == PackageManagerEngine.RESOLVE_PROVIDES_SENTINEL:
-                groups.append(current)
-                current = []
-            elif line:
-                current.append(line)
-
-        for provide, nevras in zip(provides_list, groups):
-            for nevra in nevras:
+        result: dict[str, set[Version]] = {}
+        for provide, nevras in parsed.items():
+            versions: set[Version] = set()
+            for nevra_str in nevras or []:
                 try:
-                    result[provide].add(RpmVersion.from_nevra(nevra))
+                    versions.add(RpmVersion.from_nevra(nevra_str))
                 except ValueError:
-                    self.debug(f"Could not parse NEVRA for '{provide}': {nevra}")
+                    self.debug(f"Could not parse NEVRA for '{provide}': {nevra_str}")
+            result[provide] = versions
+
+        self.debug(f"Resolved provides mapping: {result}")
         return result
 
     def create_repository(self, directory: Path) -> CommandOutput:
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -3,7 +3,7 @@ import configparser
 import enum
 import re
 import shlex
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Sequence
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -421,7 +421,7 @@ class PackageManagerEngine(tmt.utils.Common):
 
     def resolve_provides(
         self,
-        provides: Iterable[str],
+        provides: Sequence[str],
         repo_ids: Iterable[str] = (),
     ) -> ShellScript:
         """
@@ -586,7 +586,7 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
 
     def resolve_provides(
         self,
-        provides: Iterable[str],
+        provides: Sequence[str],
         repo_ids: Iterable[str] = (),
     ) -> dict[str, list['RpmVersion']]:
         """
@@ -601,10 +601,11 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
         """
         from tmt.package_managers._rpm import RpmVersion
 
-        provides_list = list(provides)
-        output = self.guest.execute(self.engine.resolve_provides(provides_list, repo_ids=repo_ids))
+        if not provides:
+            return {}
+        output = self.guest.execute(self.engine.resolve_provides(provides, repo_ids=repo_ids))
 
-        result: dict[str, list[RpmVersion]] = {provide: [] for provide in provides_list}
+        result: dict[str, list[RpmVersion]] = {provide: [] for provide in provides}
 
         assert output.stdout is not None  # narrow type
         provides_yaml: dict[str, Optional[list[_ResolvedEntry]]] = tmt.utils.from_yaml(
@@ -622,10 +623,10 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
                             resolved_provide['nevra'], repo_id=resolved_provide['repo_id']
                         )
                     )
-                except ValueError as exc:
+                except ValueError as error:
                     raise PrepareError(
-                        f"Cannot parse '{resolved_provide}' for provide '{provide}': {exc}"
-                    ) from exc
+                        f"Cannot parse '{resolved_provide}' for provide '{provide}'."
+                    ) from error
 
         return result
 
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -1,5 +1,6 @@
 import abc
 import configparser
+import contextlib
 import enum
 import re
 import shlex
@@ -399,6 +400,22 @@ class PackageManagerEngine(tmt.utils.Common):
         """
         raise NotImplementedError
 
+    def resolve_capabilities(self, *capabilities: str) -> ShellScript:
+        """
+        Resolve each capability to the NEVRA of the installed package providing it.
+
+        The script must emit one line per capability in the same order as the input.
+        Each line is either a valid NEVRA string for a found capability, or an error
+        message (e.g. ``no package provides <cap>``) for one that is not provided by
+        any installed package.
+
+        :param capabilities: Capabilities to resolve — package names, file paths, or
+            virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
+        :returns: A shell script whose stdout contains one NEVRA line per capability.
+        :raises NotImplementedError: If the package manager does not support this query.
+        """
+        raise NotImplementedError
+
     def create_repository(self, directory: Path) -> ShellScript:
         """
         Create repository metadata for package files in the given directory.
@@ -532,6 +549,35 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
             package = parts[0]
             # Omitted origin field → unknown source repository.
             result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
+        return result
+
+    def resolve_capabilities(self, capabilities: Iterable[str]) -> 'dict[str, Optional[Version]]':
+        """
+        Map each capability to the :py:class:`Version` of the installed package providing it.
+
+        :param capabilities: Capabilities to resolve — package names, file paths, or
+            virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
+        :returns: Mapping from each capability to its providing package :py:class:`Version`,
+            or ``None`` when no installed package provides it.
+        :raises NotImplementedError: If the package manager does not support this query.
+        """
+        from tmt.package_managers._rpm import RpmVersion
+
+        caps = list(capabilities)
+        if not caps:
+            return {}
+
+        script = self.engine.resolve_capabilities(*caps)
+        try:
+            output = self.guest.execute(script)
+            stdout = output.stdout
+        except tmt.utils.RunError as exc:
+            stdout = exc.stdout
+
+        result: dict[str, Optional[Version]] = dict.fromkeys(caps, None)
+        for cap, line in zip(caps, (stdout or '').splitlines()):
+            with contextlib.suppress(ValueError):
+                result[cap] = RpmVersion.from_nevra(line.strip())
         return result
 
     def create_repository(self, directory: Path) -> CommandOutput:
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -1,6 +1,5 @@
 import abc
 import configparser
-import contextlib
 import enum
 import re
 import shlex
@@ -275,6 +274,13 @@ def find_package_manager(
     return plugin
 
 
+def is_file_or_virtual_provide(requirement: str) -> bool:
+    """
+    Return ``True`` if *requirement* is a file path or virtual provide rather than a plain package.
+    """
+    return requirement.startswith('/') or '(' in requirement
+
+
 def escape_installables(*installables: Installable) -> Iterator[str]:
     for installable in installables:
         yield shlex.quote(str(installable))
@@ -308,6 +314,10 @@ class Options:
 class PackageManagerEngine(tmt.utils.Common):
     command: Command
     options: Command
+
+    #: Sentinel line emitted by :py:meth:`resolve_provides` shell scripts to
+    #: separate NEVRA groups for consecutive provides.
+    RESOLVE_PROVIDES_SENTINEL: ClassVar[str] = '---'
 
     def __init__(self, *, guest: 'Guest', logger: tmt.log.Logger) -> None:
         super().__init__(logger=logger)
@@ -400,18 +410,20 @@ class PackageManagerEngine(tmt.utils.Common):
         """
         raise NotImplementedError
 
-    def resolve_capabilities(self, *capabilities: str) -> ShellScript:
+    @abc.abstractmethod
+    def resolve_provides(self, provides: Iterable[str]) -> ShellScript:
         """
-        Resolve each capability to the NEVRA of the installed package providing it.
+        Resolve each provide to the NEVRAs of packages that provide it.
 
-        The script must emit one line per capability in the same order as the input.
-        Each line is either a valid NEVRA string for a found capability, or an error
-        message (e.g. ``no package provides <cap>``) for one that is not provided by
-        any installed package.
+        The script must emit zero or more NEVRA lines per provide followed by a
+        :py:attr:`PackageManagerEngine.RESOLVE_PROVIDES_SENTINEL` line, one group per input provide
+        in the same order as the input.  An empty group (sentinel only) means
+        nothing provides that capability.
 
-        :param capabilities: Capabilities to resolve — package names, file paths, or
+        :param provides: Provides to resolve, e.g. package names, file paths, or
             virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
-        :returns: A shell script whose stdout contains one NEVRA line per capability.
+        :returns: A shell script that emits NEVRA lines grouped by provide, each
+            group terminated by :py:attr:`PackageManagerEngine.RESOLVE_PROVIDES_SENTINEL`.
         :raises NotImplementedError: If the package manager does not support this query.
         """
         raise NotImplementedError
@@ -551,33 +563,43 @@ class PackageManager(tmt.utils.Common, Generic[PackageManagerEngineT]):
             result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
         return result
 
-    def resolve_capabilities(self, capabilities: Iterable[str]) -> 'dict[str, Optional[Version]]':
+    def resolve_provides(self, provides: Iterable[str]) -> dict[str, set[Version]]:
         """
-        Map each capability to the :py:class:`Version` of the installed package providing it.
+        Map each provide to the :py:class:`Version` objects of packages that provide it.
 
-        :param capabilities: Capabilities to resolve — package names, file paths, or
+        :param provides: Provides to resolve, e.g. package names, file paths, or
             virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
-        :returns: Mapping from each capability to its providing package :py:class:`Version`,
-            or ``None`` when no installed package provides it.
-        :raises NotImplementedError: If the package manager does not support this query.
+        :returns: Mapping from each provide to a set of :py:class:`Version` objects.
+            An empty set means nothing provides that capability.  Returns an
+            empty mapping when ``provides`` is empty.
+        :raises NotImplementedError: if the package manager does not support this query.
         """
         from tmt.package_managers._rpm import RpmVersion
 
-        caps = list(capabilities)
-        if not caps:
-            return {}
+        provides_list = list(provides)
+        result: dict[str, set[Version]] = {p: set() for p in provides_list}
+        if not provides_list:
+            return result
 
-        script = self.engine.resolve_capabilities(*caps)
-        try:
-            output = self.guest.execute(script)
-            stdout = output.stdout
-        except tmt.utils.RunError as exc:
-            stdout = exc.stdout
+        script = self.engine.resolve_provides(iter(provides_list))
+        stdout = self.guest.execute(script).stdout or ''
 
-        result: dict[str, Optional[Version]] = dict.fromkeys(caps, None)
-        for cap, line in zip(caps, (stdout or '').splitlines()):
-            with contextlib.suppress(ValueError):
-                result[cap] = RpmVersion.from_nevra(line.strip())
+        groups: list[list[str]] = []
+        current: list[str] = []
+        for line in stdout.splitlines():
+            line = line.strip()
+            if line == PackageManagerEngine.RESOLVE_PROVIDES_SENTINEL:
+                groups.append(current)
+                current = []
+            elif line:
+                current.append(line)
+
+        for provide, nevras in zip(provides_list, groups):
+            for nevra in nevras:
+                try:
+                    result[provide].add(RpmVersion.from_nevra(nevra))
+                except ValueError:
+                    self.debug(f"Could not parse NEVRA for '{provide}': {nevra}")
         return result
 
     def create_repository(self, directory: Path) -> CommandOutput:
--- a/tmt/package_managers/__init__.py
+++ b/tmt/package_managers/__init__.py
@@ -3,8 +3,19 @@ import configparser
 import enum
 import re
 import shlex
+from collections import defaultdict
 from collections.abc import Iterable, Iterator
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Generic, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Generic,
+    Optional,
+    TypedDict,
+    TypeVar,
+    Union,
+)
 from urllib.parse import urlparse
 
 import tmt.log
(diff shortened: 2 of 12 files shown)


vaibhavdaren:
The test failures are unrelated to the change.  Flaky Test :  [/tests/execute/restart/with-reboot](https://artifacts.dev.testing-farm.io/a90bda9c-abb1-4ab4-ada6-5839f83711d0/#work-executemte1q0rr_plans-provision-virtual-execute_12_default-0) [🔎](https://tmt.testing-farm.io/?test-url=https://github.com/teemtee/tmt&test-ref=b8f4602b2712a6e6daea9925dd167b5d9b5fb715&test-name=/tests/execute/restart/with-reboot&format=html)

/packit retest-failed

tmt/package_managers/__init__.py
```diff
@@ -399,6 +400,22 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
         raise NotImplementedError
 
+    def resolve_capabilities(self, *capabilities: str) -> ShellScript:
+        """
+        Resolve each capability to the NEVRA of the installed package providing it.
+
+        The script must emit one line per capability in the same order as the input.
+        Each line is either a valid NEVRA string for a found capability, or an error
+        message (e.g. ``no package provides <cap>``) for one that is not provided by
+        any installed package.
```
LecrisUT: The `capabilities` name is neither generic enough, nor is it correct in the dnf context. For dnf context, `provides` could be a better fit for what this is.

tmt/package_managers/__init__.py
```diff
@@ -399,6 +400,22 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+
+        :param capabilities: Capabilities to resolve — package names, file paths, or
+            virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
```
LecrisUT: Why the em-dash. Personally, if one uses em-dash, it feels like they did not consider the wording at all.

tmt/package_managers/__init__.py
```diff
@@ -399,6 +400,22 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+        :returns: A shell script whose stdout contains one NEVRA line per capability.
```
LecrisUT: Again, too technical. The `whose stdout` is not relevant and too literal. See the other docstrings

tmt/package_managers/__init__.py
```diff
@@ -399,6 +400,22 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+        :raises NotImplementedError: If the package manager does not support this query.
```
LecrisUT: See the raises just below here. Be consistent.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +551,35 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
             result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
         return result
 
+    def resolve_capabilities(self, capabilities: Iterable[str]) -> 'dict[str, Optional[Version]]':
```
LecrisUT: Incorrect quotations of the return type

tmt/package_managers/__init__.py
```diff
@@ -534,6 +551,35 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        """
+        Map each capability to the :py:class:`Version` of the installed package providing it.
```
```suggestion
        Map each capability to the :py:class:`Version` of the packages providing it.
```
- The package is not installed when you need to call this
- Multiple packages can provide the same provides, particularly if the user did not add relevant constraints

tmt/package_managers/__init__.py
```diff
@@ -534,6 +551,35 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        caps = list(capabilities)
+        if not caps:
+            return {}
+
+        script = self.engine.resolve_capabilities(*caps)
+        try:
+            output = self.guest.execute(script)
+            stdout = output.stdout
+        except tmt.utils.RunError as exc:
+            stdout = exc.stdout
```
LecrisUT: This is just wrong isn't it? Can't figure out what you are trying to catch here.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +551,35 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+
+        result: dict[str, Optional[Version]] = dict.fromkeys(caps, None)
```
LecrisUT: Move this at the top and you can avoid the intermediate `caps`. I commented about this very recently didn't I?

tmt/package_managers/__init__.py
```diff
@@ -534,6 +551,35 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        for cap, line in zip(caps, (stdout or '').splitlines()):
+            with contextlib.suppress(ValueError):
+                result[cap] = RpmVersion.from_nevra(line.strip())
```
LecrisUT: No `contextlib.suppress`. There is a special case being handled, then document it properly

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,10 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
             )
         ).to_script()
 
+    def resolve_capabilities(self, *capabilities: str) -> ShellScript:
+        # Reuse the existing rpm --whatprovides script; output is one NEVRA per line,
+        return self._construct_presence_script(*[Package(c) for c in capabilities])
```
LecrisUT: Not a good approach, see the other comments. The repoquery might give you more flexibility on the format and allow you to get more information to process further like the repoid.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +551,35 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        script = self.engine.resolve_capabilities(*caps)
+        try:
+            output = self.guest.execute(script)
+            stdout = output.stdout
+        except tmt.utils.RunError as exc:
+            stdout = exc.stdout
+
+        result: dict[str, Optional[Version]] = dict.fromkeys(caps, None)
+        for cap, line in zip(caps, (stdout or '').splitlines()):
+            with contextlib.suppress(ValueError):
```
AthreyVinay: This supresses parsing errors  (and others too) - ambigous. Suggest adding a `debug()` log here: 
``` python
for cap, line in zip(caps, (stdout or '').splitlines()):
    try:
        result[cap] = RpmVersion.from_nevra(line.strip())
    except ValueError:
        self.debug(f"Could not resolve capability '{cap}': {line.strip()}")
```

tmt/package_managers/__init__.py
AthreyVinay: Sorry - just seeing this now, but prolly made a duplicate comment: https://github.com/teemtee/tmt/pull/4804#discussion_r3118200575

tmt/package_managers/__init__.py
vaibhavdaren: changed to provides.  
**context**:  for choosing `capability`.  We already have a provider.

tmt/package_managers/dnf.py
vaibhavdaren: We now , run repoquery for each of the packages in a script.  Which is actually easier to parse. Addressed in 0fb9b508aecc37f9bff58075c39200e41691217b.

LecrisUT: Try running `dnf repoquery --whatprovides=make,cmake` and see what you get

vaibhavdaren: @LecrisUT ,  When multiple arguments are passed to --whatprovides, unresolved items are silently ignored and there isn’t a flag to make the command fail or report them explicitly. It also returns a merged result set without preserving input-to-output mapping.

```bash
➜  tmt git:(vaibhav-discover-only-providers-whatprovides) ✗ dnf repoquery --whatprovides=make,cmake,pikachu
Updating and loading repositories:
Repositories loaded.
cmake-0:3.31.11-1.fc42.aarch64
cmake-0:3.31.6-2.fc42.aarch64
make-1:4.4.1-10.fc42.aarch64
```

Is there something clever we could do at this point?

LecrisUT: Check the queryformat and what it can do. `fedrq` does some complex handling also if you need more reference

tmt/package_managers/__init__.py
vaibhavdaren: thanks for the suggestion, addressed in f68411ae743055fca35b9b36a8b85385a3644706.

tmt/package_managers/__init__.py
```diff
@@ -399,6 +399,22 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
         """
         raise NotImplementedError
 
+    def resolve_provides(self, *provides: str) -> ShellScript:
```
happz: Should be decorated as `abstractmethod`.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +550,36 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
             result[package] = parts[1] if len(parts) == 2 else SpecialPackageOrigin.UNKNOWN
         return result
 
+    def resolve_provides(self, provides: Iterable[str]) -> dict[str, Optional[Version]]:
```
happz: What is the reason for different type of `provides` here and in the engine?

tmt/package_managers/__init__.py
```diff
@@ -534,6 +550,36 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+            or ``None`` when no installed package provides it.
+        :raises NotImplementedError: if the package manager does not support this query.
+        """
+        from tmt.package_managers._rpm import RpmVersion
+
+        result: dict[str, Optional[Version]] = {}
+
+        provides = list(provides)
+        if not provides:
+            return result
```
happz: Empty mapping is not mentioned as a possible return value. You could use `defaultdict` instead of `dict` so that all keys would be mapped to `None` by default.

tmt/steps/prepare/verify_installation.py
```diff
@@ -92,13 +92,22 @@ def go(
...
-        # queries and is incompatible with bootc mode.
+        # Resolve provides (paths, virtual provides) to actual package names.
+        # Unresolvable entries keep their original string so they surface as
+        # NOT_INSTALLED in the origin check below.
         try:
-            package_origins = guest.package_manager.get_package_origin(self.data.verify.keys())
+            dependency_to_version = guest.package_manager.resolve_provides(self.data.verify.keys())
+        except NotImplementedError:
+            dependency_to_version = {}
+        verify: dict[str, list[str]] = {}
```
happz: Related to the note about `defaultdict` above, `dependency_to_version` as an empty mapping does not match what `resolve_provides()` says it would return. `verify` can also be `defaultdict`, it would be easier to understand than `setdefault()` below.

tmt/package_managers/dnf.py
vaibhavdaren: @LecrisUT 🥇 , 
I looked into queryformat and fedrq, but doing a bulk dnf repoquery creates a couple of blockers since we're just executing a shell script on the guest. 
My findings, 
* Lost order: DNF merges and sorts the final list, which completely destroys the 1:1 input-to-output mapping we need to track things.
* Silently missing packages: If DNF can't resolve an item, it just drops it from the output without throwing an error. We're left with no way to know which specific capability failed.

tmt/steps/prepare/verify_installation.py
vaibhavdaren: Addressed in cdc2b24bd243936c18c9b0443233df8d079a4e11.

tmt/package_managers/__init__.py
vaibhavdaren: Added proper wording to the contract. Fixed in 2e05b00ec16e4dbb2e33e4323c2eb1d9ba9aa1b1.

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,28 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
             )
         ).to_script()
 
+    def resolve_provides(self, provides: Iterable[str]) -> ShellScript:
+        # Query each provide individually using the local RPM database.
+        # This guarantees O(1) alignment for the zip() pairing in the caller,
+        # while executing in milliseconds compared to looping 'dnf repoquery'.
+
+        qf = r"%{name}-%{epoch}:%{version}-%{release}.%{arch}\n"
```
AthreyVinay: `%{epoch}` can potentially output the literal string (none) under `rpm -q` for packages without an explicit epoch - no? I cannot remember correctly - in some provider while testing, if an epoch doesnt exist, `none` is returned @LecrisUT correct me please if Im wrong.

vaibhavdaren: @AthreyVinay thankyou.
 
 %{epoch} outputs the literal string (none) for packages without
  an explicit epoch (the majority of packages). With the old queryformat,
  from_nevra silently parsed (none) as part of the version field — so bash would
  appear to have version (none):5.2.37 instead of 5.2.37
  
  qf -> spills out (none).  Fixed in 7e978fc8c93e26502960e1f9bcecf861610f4712.

LecrisUT: `%{epoch}` is `0` if unset, confirmed it:
```console
$ dnf repoquery --queryformat="%{epoch}\n" tmt
Updating and loading repositories:
Repositories loaded.
0
```

tmt/package_managers/dnf.py
LecrisUT: > * Lost order: DNF merges and sorts the final list, which completely destroys the 1:1 input-to-output mapping we need to track things.

Note that there is never a 1:1 mapping, e.g. look at the output of `--whatprovides` cmake there having multiple items that provide it, and that is expected and we have to work around them. What should be done, I don't know.

A concrete example case are the `copr_repo` provider.

> * Silently missing packages: If DNF can't resolve an item, it just drops it from the output without throwing an error. We're left with no way to know which specific capability failed.

Yes certainly, that's why I mentioned looking in the queryformat tags. `fedrq` does provided the structured information when you do the multi-query, but I don't know how it does it if it's from the tags that repoquery provides or not. I will try to look on my end as well to see what I can gather

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,28 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+
+        qf = r"%{name}-%|epoch?{%{epoch}}:{0}|:%{version}-%{release}.%{arch}\n"
+        provides_str = ' '.join(escape_installables(*[Package(p) for p in provides]))
+        # Fallback logic for file paths:
+        # Because most RPMs don't list their internal files in the 'Provides' header,
+        # `--whatprovides` will miss them. If a query fails but looks like an absolute
+        # path, we query the file list directly using `rpm -qf` to identify the package.
+        return ShellScript(f"""
+        for _provide in {provides_str}; do
+            _result=$(rpm -q --whatprovides "$_provide" --queryformat "{qf}" 2>/dev/null | head -1)
```
LecrisUT: The `head -1` is incorrect here. It is perfectly fine for multiple packages to provide the same provide.

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,28 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
             )
         ).to_script()
 
+    def resolve_provides(self, provides: Iterable[str]) -> ShellScript:
+        # Query each provide individually using the local RPM database.
+        # This guarantees O(1) alignment for the zip() pairing in the caller,
+        # while executing in milliseconds compared to looping 'dnf repoquery'.
+
+        qf = r"%{name}-%|epoch?{%{epoch}}:{0}|:%{version}-%{release}.%{arch}\n"
```
LecrisUT: You are replicating `%{full_nevra}` here

tmt/package_managers/dnf.py
LecrisUT: Do you have a reason for hiding stderr?

tmt/package_managers/dnf.py
vaibhavdaren:
```bash
(dev) ➜  tmt git:(vaibhav-discover-only-providers-whatprovides) rpm -q tmt --queryformat "%{epoch}"
(none)%
(dev) ➜  tmt git:(vaibhav-discover-only-providers-whatprovides) dnf repoquery --queryformat="%{epoch}\n" tmt
Updating and loading repositories:
Repositories loaded.
0
(dev) ➜  tmt git:(vaibhav-discover-only-providers-whatprovides)
```

LecrisUT: `rpm -q` vs `dnf repoquery` are completely different things. Which one should we use :shrug:. `rpm -q` has more tags, but they provide less info

AthreyVinay: > %{epoch} is 0 if unset, confirmed it:

Thanks @LecrisUT , however, `rpm -q` is used here which outputs `(none)`

LecrisUT: Actually `rpm -q` is the **wrong tool** to use:
```console
$ rpm -q --whatprovides fish --queryformat="%{nevra}\n" 
no package provides fish
$ dnf repoquery --whatprovides=fish --queryformat="%{full_nevra}\n" 
Updating and loading repositories:
Repositories loaded.
fish-0:4.0.2-5.fc43.x86_64
fish-0:4.2.0-2.fc43.x86_64
```
It only queries what is already installed, but at this point we want to query what is available

AthreyVinay: No arguing that :+1:

vaibhavdaren: There are two scenarios to consider:

### 1. Before the RPM is installed

* `rpm -q` does not work at this stage.
* Only `dnf repoquery` can be used.
* Note: For `repoquery` to work, repositories such as `tmt-artifact-shared` (and any other required repos) must already be configured. Which is the case and hence we can go with repoquery.

### 2. After the RPM is installed
* Both `rpm -q` and `dnf repoquery` work as expected.

LecrisUT: > * Note: For `repoquery` to work, repositories such as `tmt-artifact-shared` (and any other required repos) must already be configured. Which is the case and hence we can go with repoquery.

Not really, we can point it to a repo with no `/etc/yum/repos.d`. The design would have to be changed to link this to `repository` to know which one to use if this is needed. I don't think it would be needed for now.

> 2. After the RPM is installed

But that is too late:
- If the package was not installed, then this is in a weird state
- The original issue is to resolve the `requires`/`recommends` that could use non-package names. This can either be used for constructing the verify step in the artifacts step or when handling the special case of downloaded rpm artifacts and re-installing pre-installed packages
- This could be used to allow users to use the arbitrary package names in a custom `prepare/verify` phase, but should we be giving them that much flexibility in the first place?

tmt/package_managers/dnf.py
vaibhavdaren: When I use repoquery 
```
➜  ~ dnf repoquery --whatprovides=cmake --queryformat="%{full_nevra}\n"
Updating and loading repositories:
Repositories loaded.
cmake-0:3.31.11-1.fc42.aarch64
cmake-0:3.31.6-2.fc42.aarch64
➜  ~ dnf repoquery --whatprovides=cmake --queryformat="%{full_nevra}\n"  2>/dev/null
cmake-0:3.31.11-1.fc42.aarch64
cmake-0:3.31.6-2.fc42.aarch64
```

LecrisUT: But the stdout and stderr are already separated when you run the script with tmt, see `CommandOutput`

tmt/package_managers/dnf.py
vaibhavdaren: Provides should return a set<nevras>.  did some refactoring in ad05ffb958b4e463dd2d7073ac51df3150a2f9f7.

LecrisUT: So I did not find any good ways to reduce the `--whatprovides` calls, so a for script is still the best so far. I'll comment with a better way to design the script inside there

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,19 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+        cmd = (
+            self.command + Command('repoquery', '--queryformat', qf, '--whatprovides')
+        ).to_script()
+        provides_str = ' '.join(escape_installables(*[Package(p) for p in provides]))
+        return ShellScript(f"""
+        for _provide in {provides_str}; do
+            {cmd} "$_provide" 2>/dev/null
+            echo ---
+        done
+        """)
```
LecrisUT: This is unnecessarily complicated:
- in `dnf repoquery` you already have `%{full_nevra}`
- let's be more clever in how the information is constructed to avoid special parsing
  - instead of `---`, just format it as yaml:
    ```yaml
    '/usr/bin/cmake':
    - 'cmake-0:3.31.11-1.fc43.i686'
    - 'cmake-0:3.31.11-1.fc43.x86_64'
    - 'cmake-0:3.31.6-4.fc43.i686'
    - 'cmake-0:3.31.6-4.fc43.x86_64'
    ```
    just with something like:
    ```shell
    for _provide in /usr/bin/cmake; do
        echo "'$_provide':"
        dnf repoquery --whatprovides="$_provide" --queryformat="- '%{full_nevra}'\n"
    done
    ```
  - you still should more explicitly handle the case case of no provides. Afaict `dnf repoquery` does not exit 1 in that case or output anything to stdout, so technically the above would just work because it will be an implicit `null`. You could account for if the above is ever changes, and I think just the yaml not parsing correctly or not having a list value should catch it
- again the stderr may contain useful information, do not hide it

tmt/package_managers/__init__.py
```diff
@@ -274,6 +274,13 @@ def find_package_manager(
     return plugin
 
 
+def is_file_or_virtual_provide(requirement: str) -> bool:
+    """
+    Return ``True`` if *requirement* is a file path or virtual provide rather than a plain package.
+    """
+    return requirement.startswith('/') or '(' in requirement
```
LecrisUT: We do not care for the specific file, virtual provide or non-virtual provide (not a thing afaict, but basically the package provides that are added by default). E.g. see the `Provides`, `Obsoletes` in this project, that must also be handled.
```console
$  dnf repoquery --queryformat="provides:\n%{provides}" tmt
Updating and loading repositories:
Repositories loaded.
provides:
tmt = 1.58.0-1.fc43
python-tmt = 1.58.0-1.fc43
python3-tmt = 1.58.0-1.fc43
python3.14-tmt = 1.58.0-1.fc43
python3.14dist(tmt) = 1.58
python3dist(tmt) = 1.58
```
`tmt` and `python*` are not accounted, but should be accounted

vaibhavdaren: Yup , removed this bit all packages go to provides.

tmt/package_managers/__init__.py
```diff
@@ -399,6 +399,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+            - 'cmake-0:3.31.6-4.fc43.x86_64'
+            'make':
+            - 'make-1:4.4.1-8.fc43.x86_64'
+
+        :param provides: Provides to resolve, e.g. package names, file paths, or
+            virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
+        :returns: A shell script that emits a YAML mapping of provide → NEVRA list.
+        :raises PrepareError: If the package manager does not support this query.
+        """
+        raise PrepareError("Package manager does not support provides resolution.")
```
LecrisUT, line 450: Is it an `abstractmethod` or do you have a default implementation (other than `NotImplementedError`). Pick one

tmt/package_managers/__init__.py
```diff
@@ -534,6 +554,39 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+    def resolve_provides(self, provides: list[str]) -> dict[str, set[Version]]:
+        """
+        Map each provide to the :py:class:`Version` objects of packages that provide it.
+
+        :param provides: Provides to resolve, e.g. package names, file paths, or
+            virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
+        :returns: Mapping from each provide to a set of :py:class:`Version` objects.
+            An empty set means nothing provides that capability.  Returns an
+            empty mapping when ``provides`` is empty.
+        :raises NotImplementedError: if the package manager does not support this query.
```
LecrisUT: Not true, and not really applicable since it is not this function that does the raising

tmt/package_managers/__init__.py
```diff
@@ -534,6 +554,39 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        parsed: dict[str, Optional[list[str]]] = tmt.utils.from_yaml(stdout) or {}
+
+        result: dict[str, set[Version]] = {}
+        for provide, nevras in parsed.items():
+            versions: set[Version] = set()
+            for nevra_str in nevras or []:
+                try:
+                    versions.add(RpmVersion.from_nevra(nevra_str))
+                except ValueError:
+                    self.debug(f"Could not parse NEVRA for '{provide}': {nevra_str}")
```
LecrisUT: Overly cautious for something that is not supposed to fail and then you just hide in `debug`. I encourage you to be more conscious of why do you branch or catch special cases, I see at least 4 questionable cases in this function

tmt/package_managers/__init__.py
```diff
@@ -534,6 +554,39 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+            result[provide] = versions
+
+        self.debug(f"Resolved provides mapping: {result}")
```
LecrisUT: The data is no displayed nicely here and you already have the information from the previous execute script, not worth logging imo.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +554,39 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+            empty mapping when ``provides`` is empty.
+        :raises NotImplementedError: if the package manager does not support this query.
+        """
+        from tmt.package_managers._rpm import RpmVersion
+
+        if not provides:
+            return {}
+
+        script = self.engine.resolve_provides(provides)
+        stdout = self.guest.execute(script).stdout or ''
```
LecrisUT: It is annoying for debugging if a variable is eagerly dropped (using `execute(script).stdout` instead of  saving `execute(script)`)

tmt/package_managers/__init__.py
```diff
@@ -534,6 +554,39 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        parsed: dict[str, Optional[list[str]]] = tmt.utils.from_yaml(stdout) or {}
+
+        result: dict[str, set[Version]] = {}
+        for provide, nevras in parsed.items():
+            versions: set[Version] = set()
+            for nevra_str in nevras or []:
```
LecrisUT: Ironically the special case when the nevras is `null` (nothing provides it) is not being logged here or later.

tmt/package_managers/__init__.py
```diff
@@ -399,6 +399,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def resolve_provides(self, provides: list[str]) -> ShellScript:
```
LecrisUT: `Itterable` or `*arg` to encourage immutability

tmt/package_managers/__init__.py
```diff
@@ -399,6 +399,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+        """
+        Resolve each provide to the NEVRAs of packages that provide it.
+
+        The script must emit YAML mapping each provide string to a list of NEVRA
+        strings (or null when nothing provides it), e.g.::
+
+            '/usr/bin/cmake':
+            - 'cmake-0:3.31.6-4.fc43.x86_64'
+            'make':
+            - 'make-1:4.4.1-8.fc43.x86_64'
```
LecrisUT: Make this a proper yaml code block, see how it is being rendered https://tmt--4804.org.readthedocs.build/en/4804/z_autodocs/tmt.package_managers.html#tmt.package_managers.PackageManagerEngine.resolve_provides

tmt/package_managers/__init__.py
```diff
@@ -399,6 +399,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def resolve_provides(self, provides: list[str]) -> ShellScript:
+        """
+        Resolve each provide to the NEVRAs of packages that provide it.
+
+        The script must emit YAML mapping each provide string to a list of NEVRA
+        strings (or null when nothing provides it), e.g.::
```
```suggestion
        strings (or null when nothing provides it), e.g.:
```
Not parsed properly, technically should be caught by the sphinx linter, but obviously it is not doing a good job 🙁

tmt/package_managers/__init__.py
```diff
@@ -399,6 +399,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+
+            '/usr/bin/cmake':
+            - 'cmake-0:3.31.6-4.fc43.x86_64'
+            'make':
+            - 'make-1:4.4.1-8.fc43.x86_64'
+
+        :param provides: Provides to resolve, e.g. package names, file paths, or
+            virtual provides (e.g. ``make``, ``/usr/bin/make``, ``pkgconfig(openssl)``).
+        :returns: A shell script that emits a YAML mapping of provide → NEVRA list.
```
LecrisUT: The NEVRA list is not correct since it can be null. But on the other hand you already have the details explained in the section above, why duplicate it. My preference is to keep the `param`, `returns`, etc. as minimal as possible and put the details in the body instead.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +554,39 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        if not provides:
+            return {}
+
+        script = self.engine.resolve_provides(provides)
+        stdout = self.guest.execute(script).stdout or ''
+        parsed: dict[str, Optional[list[str]]] = tmt.utils.from_yaml(stdout) or {}
+
+        result: dict[str, set[Version]] = {}
+        for provide, nevras in parsed.items():
+            versions: set[Version] = set()
```
LecrisUT: Why the `set`? It can happen that we have some NEVRA collisions. We would probably need to do some better handling in that case, but it is still a thing that happens not that infrequently especially when having a copr repo.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,11 +355,19 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
 
-        # Build package → set of valid repo_ids, filtering to only required packages.
+        # Resolve all requirements to canonical package names via whatprovides so
+        # they can be matched against artifact.version.name below.  This handles
+        # plain names, file paths, pkgconfig(...) and package-level provides
+        # Artifact repos are already configured on the guest at this
+        # point even though the packages themselves are not yet installed.
+        resolved = guest.package_manager.resolve_provides(list(pkg_names))
+
+        resolved_names = {version.name for versions in resolved.values() for version in versions}
```
LecrisUT, line 379: Why strip and use only the names? We do have "contamination" from repos that we are not interested in, and actually, looking more carefully, you can allow `resolve_provides` to take a list of `repo_id` to do that filtering early on.

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,11 +355,19 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
             for pkg in install_phase.data.package:  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
                 pkg_names.add(str(pkg))  # pyright: ignore[reportUnknownArgumentType]
 
-        # Build package → set of valid repo_ids, filtering to only required packages.
+        # Resolve all requirements to canonical package names via whatprovides so
+        # they can be matched against artifact.version.name below.  This handles
+        # plain names, file paths, pkgconfig(...) and package-level provides
+        # Artifact repos are already configured on the guest at this
+        # point even though the packages themselves are not yet installed.
+        resolved = guest.package_manager.resolve_provides(list(pkg_names))
```
LecrisUT: Converting to a `list` is superfluous

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Nice suggestion, tried this out.

#### Issue with `--repo` filtering in `dnf repoquery`

I tried filtering early by passing `repo_ids` into `resolve_provides` and adding `--repo=<id>` to:

```bash
dnf5 repoquery --whatprovides --repo=<id> <pkg_or_path>
````

This silently returns empty results for file-path provides:

```bash
for p in /usr/bin/docker /usr/bin/flock; do
  dnf5 repoquery --whatprovides \
    --repo=docker-ce-stable --repo=tmt-artifact-shared "$p"
done

# '/usr/bin/docker':
# Repositories loaded.
# '/usr/bin/flock':
# Repositories loaded.
```

### Cause

* `--repo` disables all other repositories
* In this mode, DNF5 does **not load `filelists` metadata**
* File-path provides (`/usr/bin/...`) fail silently
* Package-name provides still work

### Why not use `optional_metadata_types=filelists`?

```bash
--setopt=optional_metadata_types=filelists
```

* Works only on DNF5
* Not supported on DNF4 (RHEL 9)
* Not portable → not suitable for `tmt`

### Fix

Query all repos (no `--repo`), then filter in Python using `repo_id`

LecrisUT: What are you on about? Here's a quick test I did
```console
$ dnf repoquery --repo=mullvad-stable --whatprovides=/usr/bin/mullvad
Updating and loading repositories:
 Mullvad VPN                                                                                                                                                                   100% |  27.1 KiB/s |  12.8 KiB |  00m00s
Repositories loaded.
mullvad-vpn-0:2026.1-1.x86_64
```

Where what told you and why and how? Please do not take whatever an LLM tells you at face value and do a proper analysis of what it is saying before offloading the burden of proof onto others.

vaibhavdaren: You're right, and I sorry. I tested this incorrectly and should have verified it more carefully before posting that analysis. I'll redo the testing properly.

tmt/package_managers/dnf.py
vaibhavdaren: This should be addressed in 2e05b00ec16e4dbb2e33e4323c2eb1d9ba9aa1b1.

tmt/package_managers/__init__.py
vaibhavdaren: Checked now renders properly.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +584,57 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
         return result
 
+    # String annotation avoids circular import: _rpm.py imports Version from this module.
+    # NOTE: this method is only meaningful for RPM-based package managers; non-RPM engines
+    # raise PrepareError from their engine.resolve_provides implementation.
+    def resolve_provides(
+        self,
+        provides: Iterable[str],
+        repo_ids: Iterable[str],
+    ) -> 'dict[str, list[RpmVersion]]':
```
LecrisUT: Why the quotes around the `dict`?

tmt/package_managers/__init__.py
```diff
@@ -534,6 +584,57 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+                    resolved.append(
+                        RpmVersion.from_nevra(
+                            resolved_provide['nevra'], repo_id=resolved_provide['repo_id']
+                        )
+                    )
+                except ValueError as exc:
+                    raise PrepareError(
+                        f"Cannot parse '{resolved_provide}' for provide '{provide}': {exc}"
+                    ) from exc
+            result[provide] = resolved
```
LecrisUT: You can skip this assignment you do `resolved = results[provide]`

tmt/package_managers/__init__.py
```diff
@@ -534,6 +584,57 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        repo_ids: Iterable[str],
+    ) -> 'dict[str, list[RpmVersion]]':
+        """
+        Map each provide to the :py:class:`RpmVersion` objects of packages that provide it.
+
+        :param provides: Provides to resolve.
+        :param repo_ids: Limit the query to these repository IDs.
+        :returns: Mapping from each provide to a list of :py:class:`RpmVersion` objects,
+            each carrying the NEVRA and source repository. Every requested provide appears
+            as a key; provides with no match map to an empty list.
```
LecrisUT, line 600: Not true because you do not pre-populate all the `result` first. Which reality is it meant to be?

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+        repo_ids: Iterable[str],
+    ) -> ShellScript:
+        provides_str = ' '.join(escape_installables(*[Package(p) for p in provides]))
+        cmd = (
+            self.command
+            + Command(
+                'repoquery', *[f'--repo={repo_id}' for repo_id in repo_ids], '--whatprovides'
+            )
+        ).to_script()
+        qf = "- nevra: '%{full_nevra}'\\n  repo_id: '%{repoid}'\\n"
```
LecrisUT:
```suggestion
        qf = r"- nevra: '%{full_nevra}'\n  repo_id: '%{repoid}'\n"
```

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,11 +355,27 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
             for pkg in install_phase.data.package:  # pyright: ignore[reportUnknownMemberType,reportUnknownVariableType]
                 pkg_names.add(str(pkg))  # pyright: ignore[reportUnknownArgumentType]
 
-        # Build package → set of valid repo_ids, filtering to only required packages.
+        provider_repo_ids = {SHARED_REPO_NAME} | {
```
LecrisUT, line 358: Technically it is not known if `SHARED_REPO_NAME` should be added, but that's an issue with the current provider api.

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
...
+    def resolve_provides(
+        self,
+        provides: Iterable[str],
+        repo_ids: Iterable[str],
+    ) -> ShellScript:
+        provides_str = ' '.join(escape_installables(*[Package(p) for p in provides]))
+        cmd = (
+            self.command
+            + Command(
+                'repoquery', *[f'--repo={repo_id}' for repo_id in repo_ids], '--whatprovides'
```
LecrisUT: Why not add `--queryformat` in here as well?

tmt/package_managers/dnf.py
```diff
@@ -208,6 +209,26 @@ def get_package_origin(self, packages: Iterable[str]) -> ShellScript:
             )
         ).to_script()
 
+    def resolve_provides(
+        self,
+        provides: Iterable[str],
+        repo_ids: Iterable[str],
```
LecrisUT: What about defaulting to `None` more explicitly?

tmt/package_managers/__init__.py
```diff
@@ -534,6 +584,57 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        script = self.engine.resolve_provides(provides_list, repo_ids=repo_ids)
+        output = self.guest.execute(script)
+
+        result: defaultdict[str, list[RpmVersion]] = defaultdict(list)
+
+        provides_yaml: Optional[dict[str, Optional[list[_ResolvedEntry]]]] = tmt.utils.from_yaml(
+            output.stdout or ''
+        )
+        if provides_yaml is None:
+            return result
```
LecrisUT: Unnecessary branching?

LecrisUT (approved):
Logic lgtm, just some cleanups.

tmt/package_managers/__init__.py
LecrisUT: You misunderstood me. The unnecessary branching is that `output.stdout` is never empty and `from_yaml(...)` never results in `None`. If what I post is unclear, please ping me to clarify it.

tmt/package_managers/__init__.py
LecrisUT: The comment
```python
    # String annotation avoids circular import: _rpm.py imports Version from this module.
```
is unnecessary. The issue I had in the implementation above is that quoting the whole `dict` is incorrect, the quote is only supposed to be around `RpmVersion`. The reason for it is either obvious or irrelevant. Again if I am unclear, ping me for clarifications.

vaibhavdaren: Removed the comment.

tmt/package_managers/__init__.py
vaibhavdaren: One edge case though.
 from_yaml can return None when provides is empty (empty loop  ->  no output -> from_yaml("") →  None.
 Fixed by emitting {} in the shell script when the loop doesn't run.

tmt/package_managers/__init__.py
```diff
@@ -534,6 +585,50 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+        """
+        from tmt.package_managers._rpm import RpmVersion
+
+        provides_list = list(provides)
+        output = self.guest.execute(self.engine.resolve_provides(provides_list, repo_ids=repo_ids))
+
+        result: dict[str, list[RpmVersion]] = {provide: [] for provide in provides_list}
+
+        provides_yaml: dict[str, Optional[list[_ResolvedEntry]]] = tmt.utils.from_yaml(
+            cast(str, output.stdout)
```
AthreyVinay: Why use `cast()` here? This does not give any valuable hint about a `None` failure - why not use this?
https://github.com/teemtee/tmt/blob/2b81bbd75e0fe34b16a7ea41b7c620cc05ac1c56/tmt/package_managers/dnf.py#L301
so that we can give a clear error message?

tmt/package_managers/__init__.py
LecrisUT: Sure, but that should be handled either at the beginning of the function or as soon as you have `output`, my preference is the former.

If you use the former, then changing the type to `Sequence` also makes sense to avoid the casting to `list`. The difference between `Iterable` and `Sequence` is the `len` which would be relevant for this check

tmt/steps/prepare/artifact/__init__.py
```diff
@@ -355,11 +355,27 @@ def _inject_verify_phase(self, providers: list[ArtifactProvider], guest: Guest)
...
+            for repo_id in repo.repo_ids
+        }
+
+        # Resolve all requirements to canonical package names via whatprovides so
+        # they can be matched against artifact.version.name below.  This handles
+        # plain names, file paths, pkgconfig(...) and package-level provides
+        # Artifact repos are already configured on the guest at this
+        # point even though the packages themselves are not yet installed.
+
+        resolved = guest.package_manager.resolve_provides(pkg_names, provider_repo_ids)
```
AthreyVinay: Can be a bit more defensive about checking the existence of `pkg_names` and returning early here.

tmt/package_managers/__init__.py
vaibhavdaren: output is never none, script always returns a valid yaml. so this is a dead branch.

LecrisUT: `assert ... # narrow type`

AthreyVinay: Yep... just prove it with above!

tmt/package_managers/__init__.py
```diff
@@ -534,6 +585,50 @@ def get_package_origin(self, packages: Iterable[str]) -> 'dict[str, PackageOrigi
...
+            for resolved_provide in nevras:
+                try:
+                    result[provide].append(
+                        RpmVersion.from_nevra(
+                            resolved_provide['nevra'], repo_id=resolved_provide['repo_id']
+                        )
+                    )
+                except ValueError as exc:
+                    raise PrepareError(
+                        f"Cannot parse '{resolved_provide}' for provide '{provide}': {exc}"
```
happz: Drop the `: {exc}` bit, please. Chaining of exceptions will take care of reporting `exc`.

tests/prepare/artifact/providers/koji-build/data/main.fmf
```diff
@@ -17,4 +17,4 @@
     duration: 5m
     summary: Verify package from koji build
     require:
-      - make
+      - /usr/bin/make
```
happz, line 20: Isn't this change dropping coverage for package name? Would it be possible to use two dependencies, e.g. a package and a filepath, let's say `tree` and `/usr/bin/make`?

vaibhavdaren: Package name coverage is already handled by the [koji-nvr tests ](https://artifacts.dev.testing-farm.io/8e1b9e77-37d4-4049-ac8f-376fe77d4f3a/work-the-restuw3to8tq/plans/provision/container/the-rest/execute/data/guest/default-0/tests/prepare/artifact/providers/koji-nvr-10/output.txt) those use a plain package name and auto verification still kicks in (see koji-nvr test output).
 
The koji-build test uses /usr/bin/make (filepath) intentionally to keep each test focused on a distinct whatprovides case rather than duplicating coverage across tests.

tmt/package_managers/__init__.py
vaibhavdaren: Addressed in 103058f0f12db0835a2b4a4c4c15de6f21d25a80.

tmt/steps/prepare/artifact/__init__.py
vaibhavdaren: Addressed in 103058f0f12db0835a2b4a4c4c15de6f21d25a80.

tmt/package_managers/__init__.py
vaibhavdaren: Addressed in 09459fd8fbdabeb5442dfd4c8975fff773c3b445. @AthreyVinay Can you have a look today?

LecrisUT (approved):

AthreyVinay (approved):

PR #4807, opened 2026-04-20
Fix recipe loading when fmf tree is not present

therazix:
Recipe loading now works even when the current working directory does not contain an fmf tree, by creating a new tree from the saved fmf root path.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/spec/recipe.fmf
+++ b/spec/recipe.fmf
@@ -88,8 +88,8 @@ description: |
                     how: tmt
                     order: 50
                     ...
-                # String, relative path to the results file from the previous run
-                results-path: plans/name/execute/results.yaml
+                # String, absolute path to the results file from the previous run
+                results-path: /var/tmp/tmt/run-001/plans/name/execute/results.yaml
 
             report:
                 enabled: true
--- a/spec/recipe.fmf
+++ b/spec/recipe.fmf
@@ -88,8 +88,8 @@ description: |
                     how: tmt
                     order: 50
                     ...
-                # String, absolute path to the results file from the previous run
-                results-path: /var/tmp/tmt/run-001/plans/name/execute/results.yaml
+                # String, relative path to the results file from the previous run
+                results-path: plans/name/execute/results.yaml
 
             report:
                 enabled: true
--- a/tmt/base/core.py
+++ b/tmt/base/core.py
@@ -2924,7 +2924,7 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         self.recipe_manager = RecipeManager(logger)
         self.recipe = None
         if recipe_path is not None and self._tree is not None:
-            self.recipe = self.recipe_manager.load(self, recipe_path, self._tree.tree)
+            self.recipe = self.recipe_manager.load(self, recipe_path)
 
     @property
     def run_workdir(self) -> Path:
--- a/tmt/base/core.py
+++ b/tmt/base/core.py
@@ -2923,7 +2923,7 @@ class Run(HasRunWorkdir, HasEnvironment, tmt.utils.Common):
         self.policies = policies or []
         self.recipe_manager = RecipeManager(logger)
         self.recipe = None
-        if recipe_path is not None and self._tree is not None:
+        if recipe_path is not None:
             self.recipe = self.recipe_manager.load(self, recipe_path)
 
     @property
--- a/tmt/recipe.py
+++ b/tmt/recipe.py
@@ -564,7 +564,7 @@ class RecipeManager(Common):
 
     def load(self, run: 'Run', recipe_path: Path) -> Recipe:
         recipe = Recipe.from_spec(
-            cast(_RawRecipe, tmt.utils.yaml_to_dict(self.read(recipe_path), yaml_type='rt')),
+            cast(_RawRecipe, tmt.utils.yaml_to_dict(self.read(recipe_path))),
             self._logger,
         )
         self._update_tree(run, recipe)
@@ -585,9 +585,7 @@ class RecipeManager(Common):
             ),
             plans=[_RecipePlan.from_plan(plan) for plan in run.plans],
         )
-        self.write(
-            run.run_workdir / 'recipe.yaml', tmt.utils.to_yaml(recipe.to_spec(), yaml_type='rt')
-        )
+        self.write(run.run_workdir / 'recipe.yaml', tmt.utils.to_yaml(recipe.to_spec()))
 
     def tests(self, recipe: Recipe, plan_name: str) -> list[TestOrigin]:
         """
--- a/tmt/recipe.py
+++ b/tmt/recipe.py
@@ -564,7 +564,8 @@ class RecipeManager(Common):
 
     def load(self, run: 'Run', recipe_path: Path) -> Recipe:
         recipe = Recipe.from_spec(
-            cast(_RawRecipe, tmt.utils.yaml_to_dict(self.read(recipe_path))), self._logger
+            cast(_RawRecipe, tmt.utils.yaml_to_dict(self.read(recipe_path), yaml_type='rt')),
+            self._logger,
         )
         self._update_tree(run, recipe)
         # TODO: We should have a way to set which steps are enabled
--- a/tmt/recipe.py
+++ b/tmt/recipe.py
@@ -606,12 +606,12 @@ class RecipeManager(Common):
     @staticmethod
     def _update_tree(run: 'Run', recipe: Recipe) -> None:
         """
-        Create a new fmf tree from the recipe's plan specifications.
+        Create a new fmf tree from the recipe's run and plan specifications.
         """
         from tmt.base.core import Tree
 
         fmf_tree = fmf.Tree({plan.name: plan.to_fmf_spec() for plan in recipe.plans})
-        root = str(Path.cwd())
+        root = recipe.run.root or str(Path.cwd())
         fmf_tree.root = root
         for node in fmf_tree.climb():  # pyright: ignore[reportUnknownVariableType]
             if isinstance(node, fmf.Tree):
--- a/tmt/recipe.py
+++ b/tmt/recipe.py
@@ -611,11 +611,10 @@ class RecipeManager(Common):
         from tmt.base.core import Tree
 
         fmf_tree = fmf.Tree({plan.name: plan.to_fmf_spec() for plan in recipe.plans})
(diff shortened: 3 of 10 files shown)


tmt/base/core.py
```diff
@@ -2924,7 +2924,7 @@ def __init__(
         self.recipe_manager = RecipeManager(logger)
         self.recipe = None
         if recipe_path is not None and self._tree is not None:
```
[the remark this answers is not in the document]
therazix: Removed in 907d6c856f1028da7399065bf08a3f359b630487.

happz (approved):

tmt/recipe.py
```diff
@@ -584,7 +584,9 @@ def save(self, run: 'Run') -> None:
             ),
             plans=[_RecipePlan.from_plan(plan) for plan in run.plans],
         )
-        self.write(run.run_workdir / 'recipe.yaml', tmt.utils.to_yaml(recipe.to_spec()))
+        self.write(
+            run.run_workdir / 'recipe.yaml', tmt.utils.to_yaml(recipe.to_spec(), yaml_type='rt')
+        )
```
LecrisUT: For the `rt` to work, you must also force `rt` at the load part (line 567?)

spec/recipe.fmf
```diff
@@ -88,8 +88,8 @@ description: |
                     how: tmt
                     order: 50
                     ...
-                # String, relative path to the results file from the previous run
-                results-path: plans/name/execute/results.yaml
+                # String, absolute path to the results file from the previous run
+                results-path: /var/tmp/tmt/run-001/plans/name/execute/results.yaml
```
LecrisUT: Hmm, but does this work with the other assumptions of the locality of the paths? E.g. `recipe.yaml` seems like it would always overwrite the previous run's execution. What is the context for forcing this one to be absolute?

tmt/recipe.py
therazix: Changed in 5a0cee0c43ba3fb7b3c63069fbe6a7739795e8ec.

spec/recipe.fmf
therazix: Loading run from the recipe does not override previous run. It will create a new run. `results-path` stores the path to the `results.yaml` file for the run that generated the recipe. It's needed only if you want to filter the tests in the recipe based on their results, which is what I did with [this script](https://gitlab.cee.redhat.com/fvagner/tmt-recipe-tool/-/merge_requests/1). The reason why this is now an absolute path is that the `recipe.yaml` could be moved to a different directory and we would lose the path to the `results.yaml` file and we wouldn't be able to filter the tests.

LecrisUT: > The reason why this is now an absolute path is that the `recipe.yaml` could be moved to a different directory

Doesn't that apply to all other path-like objects then? Shouldn't the consumer account for this and extend the path from the run path? An associated question, is this run directory assumed to be static with respect to multiple executions that consume the recipe.yaml?

therazix: Yeah, that makes sense. Originally, I thought that the absolute path will be better, but I changed it back to relative path in d3effef1fbb2e4976de44bca5b36c2512721beca. When users decide to move the recipe, they should handle the path themselves.

tcornell-bus:
I tried to verify the fix by doing a basic tmt run via container.

Then I moved into a blank directory with no fmf tree and tried to run with the recipe file:
(tmt version: 1.71.0)
```
tcornell ~/tmp $ tmt run -vvv --recipe /var/tmp/tmt/run-063/recipe.yaml

No metadata found in the '.' directory. Use 'tmt init' to get started.

Cause number 1:

    Unable to find tree root for '/home/tcornell/tmp'.
```
(PR)
```
(dev) tcornell ~/tmp $  tmt run -vvv --recipe /var/tmp/tmt/run-063/recipe.yaml

[...]

    execute
        queued execute tasks
            #1: default-0 on default-0
        
        how: tmt
        order: 50
        ignore-duration: false
        exit-first: false
            test: Concise summary describing what the test does
                duration deadline: 300 seconds, at 16:23:03 2026-04-20 UTC
                cmd: ./test.sh
                stdout: ./tmt-test-wrapper-inner.sh-smoke-1-default-0-default-0: line 1: ./test.sh: No such file or directory
                00:00:00 errr /smoke (on default-0) [1/1]
                    Note: beakerlib: TestResults FileError

    
        summary: 1 test executed
```
Looks like it made the fake tree, but the tree does not have the supporting test files.
When I execute the same command in the original fmf root directory, it runs fine.

Am I reproducing this correctly?

tmt/steps/report/reportportal.py
tcornell-bus, line 1: Not sure if this PR is supposed to fix this or not, but I did a tmt run with `report --how reportportal --project <my_project>` and it made the recipe.yaml file with the key/values in the `to_spec()\to_minimal_spec()` functions. Then I tried loading that same recipe file (in the fmf root directory), and I got an exception:
```
/plans/all
summary: Run all tests

plan failed.

Cause number 1:

    cannot represent an object: {1: '<some-hash>'}
```
Where `{1: '<some-hash>'}` is the reportportal `test-uuids` value.

therazix: Thanks for catching that. The issue was caused by `yaml_type='rt'` during recipe loading. Reverted in 4bfdcdb5d8889753524640faa1d766624d196b99.

therazix: And the reportportal changes were moved to https://github.com/teemtee/tmt/pull/4809.

LecrisUT: > Thanks for catching that. The issue was caused by `yaml_type='rt'` during recipe loading. Reverted in [4bfdcdb](https://github.com/teemtee/tmt/commit/4bfdcdb5d8889753524640faa1d766624d196b99).

That sounds fishy :thinking:

therazix:
> Looks like it made the fake tree, but the tree does not have the supporting test files. When I execute the same command in the original fmf root directory, it runs fine.
> 
> Am I reproducing this correctly?

Yes, I would say that is expected. When users move the recipe or execute it from a different location, they should handle the necessary test files. Either move them to the new directory or rewrite the recipe with updated paths.

tcornell-bus:
I think I am misunderstanding the use case. What situations would a user execute the tmt command outside of the fmf tree? Could you provide an example or add a test case, please?

therazix:
> I think I am misunderstanding the use case. What situations would a user execute the tmt command outside of the fmf tree? Could you provide an example or add a test case, please?

In some cases, an fmf tree is not needed when the run is loaded from the recipe, unless the test requires some additional files. I ran into this while creating a script for re-executing tests. tmt would raise an exception due to a missing tree, even though we wouldn't need it.

But looking at it again, I completely missed that we actually save the fmf root path in the recipe, so we can just set it from there, and that would solve the issue. Changed in f2512fc9431429eb51249f120857435db57f33ca.

LecrisUT:
> But looking at it again, I completely missed that we actually save the fmf root path in the recipe, so we can just set it from there, and that would solve the issue. Changed in [f2512fc](https://github.com/teemtee/tmt/commit/f2512fc9431429eb51249f120857435db57f33ca).

:+1: to the idea, but can you double check that the tree in there is the user tree and not the work tree. If it is the latter, then it should be changed to the former imo.

> What situations would a user execute the tmt command outside of the fmf tree?

Not exactly the original context, but a situation that should probably be supported (or disallowed explicitly) is if one runs `tmt run ...` without a `fmf` root at all.

therazix:
> 👍 to the idea, but can you double check that the tree in there is the user tree and not the work tree. If it is the latter, then it should be changed to the former imo.

It is a user tree, e.g. `/home/user/tmt-plans`

tmt/recipe.py
```diff
@@ -603,12 +604,19 @@ def tests(self, recipe: Recipe, plan_name: str) -> list[TestOrigin]:
...
-        Load the plans from the recipe and update the given fmf tree with their specifications.
+        Create a new fmf tree from the recipe's run and plan specifications.
         """
-        tree.children.clear()
-        tree.update({plan.name: plan.to_fmf_spec() for plan in recipe.plans})
+        from tmt.base.core import Tree
+
+        fmf_tree = fmf.Tree({plan.name: plan.to_fmf_spec() for plan in recipe.plans})
+        root = recipe.run.root or str(Path.cwd())
+        fmf_tree.root = root
```
LecrisUT: Maybe make the root be `None` to be more explicit on what case this is. I think the case there `recipe.run.root` would only happen if the original run did not have a user tree? In that case we should not be injecting a new one either?

therazix: I agree. Changed in a95c1ee7f7fb25c2c15c84acc68eded5df9f73de.

LecrisUT (approved):
👍 from me, but I deffer to @tcornell-bus discussion on handling the case with no fmf trees

tcornell-bus (approved):
LGTM now. 
In an empty tmp dir with no fmf tree:
Old:
```
tcornell ~/tmp $  tmt run -vvv --recipe /var/tmp/tmt/run-063/recipe.yaml

No metadata found in the '.' directory. Use 'tmt init' to get started.

Cause number 1:

    Unable to find tree root for '/home/tcornell/tmp'.
```

PR:
```
(dev) tcornell ~/tmp $ tmt run -vvv --recipe /var/tmp/tmt/run-063/recipe.yaml
/var/tmp/tmt/run-095
[...]
Found 1 plan.

/plans/all
summary: Run all tests
    discover
        directory: /home/tcornell/tmt_testing
        how: fmf
        order: 50
        summary: 1 test selected
            /smoke
            
 [....]
 total: 1 test passed
```
It found the original fmf tree and successfully executed.

happz:
freedesktop.org is down, hence the linkcheck fail. The rest is green, merging.

spec/recipe.fmf
tcornell-bus: Ahh, I ran into this inconvenience when testing the `tmt-recipe-tool`. I wanted to pull a couple of recipes from my run directories and store them together, so I don't have to remember which run directory each recipe lived in. I had to change each `results-path` to be the absolute path in the recipe.yaml files. It would be hugely helpful if the path were already absolute. It doesn't seem like it would hurt anything to provide this convenience to users.

tcornell-bus: made an issue so we can discuss further there https://github.com/teemtee/tmt/issues/4849

PR #4809, opened 2026-04-21
Fix recipe generation for ReportPortalData

therazix:
Adds a missing `to_spec/to_minimal_spec` methods to the ReportPortal plugin.

Pull Request Checklist

* [x] implement the feature

### DIFF
--- a/tmt/steps/report/reportportal.py
+++ b/tmt/steps/report/reportportal.py
@@ -2,7 +2,7 @@ import datetime
 import os
 import re
 from re import Pattern
-from typing import TYPE_CHECKING, Any, Optional, Union, overload
+from typing import TYPE_CHECKING, Any, Callable, Optional, Union, cast, overload
 
 import requests
 import urllib3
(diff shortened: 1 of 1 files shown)


happz:
@therazix please, include a small test if possible. It does not need to run anything serious, the issue is with saving recipe when `reportportal` plugin is present.

therazix:
> @therazix please, include a small test if possible. It does not need to run anything serious, the issue is with saving recipe when `reportportal` plugin is present.

As discussed in the review meeting, testing would require a working ReportPortal instance, otherwise the recipe would not be generated. Skipping the test for now.

LecrisUT (approved):

tcornell-bus (approved):
Ran tmt with reportportal plugin and it created the recipe.yaml file, the report section of the yaml file looks fine. Re-ran tmt with the generated recipe file and that run completed succesfully and the recipe files match.
Minor concern, there is no indent for some sections like:
```
      upload-log-pattern:
      - avc\.txt
      - dmesg-.*\.txt
      - output\.txt
      - tmt-watchdog\.txt
```
and
```
      require:
      - vim
```
The recipe file load went fine though so I think the lack of indentation did not break anything.

happz (approved):

happz:
/packit test --identifier testing-farm:centos-stream-9-x86_64:full

PR #4813, opened 2026-04-21
Properly account for sub-tests in `modified-only`

### LINKED ISSUE #4626: Plan with `modified-only: true` might run a modified test multiple times due to an unanchored regex
Hello,

I recently came across a strange issue, where a newly added test that shares a prefix with another also modified test got executed twice in a plan with `modified-only: true`.

I was able to "minimize" it into a following reproducer:

```
$ mkdir /tmp/tmt-test
$ cd /tmp/tmt-test
$ git init .
$ tmt init

## Create the first test & the test plan and commit everything
$ tmt test create --template shell foo
$ mkdir plans
$ cat >plans/modified.fmf <<EOF
summary: Only modified tests
/:
  inherit: false
discover:
  how: fmf
  modified-only: true
execute:
  how: tmt
EOF
$ git add plans foo .fmf
$ git commit -m "Initial commit"

## Create a new branch & a new test that shares a prefix with the already existing test, and modify the already existing one
$ git checkout -b new
$ tmt test create --template shell foo-bar
$ echo 'duration: 1m' >>foo/main.fmf 
$ git add foo foo-bar
$ git commit -m "More tests"

## Run the discover phase of the "modified" plan
$ tmt -v run discover --how fmf --modified-ref main
/var/tmp/tmt/run-040
Found 1 plan.

/plans/modified
summary: Only modified tests
    discover
        directory: /tmp/tmt-test
        commit-hash: eced158
        how: fmf
        modified-ref: main
        modified-ref hash: 559d0aa
        summary: 3 tests selected
            /foo-bar
            /foo
            /foo-bar
```

Notice how the `foo-bar` test is about to get executed twice.

After a quick skim through the code I suspect the culprit here is a missing end-of-string anchor in the regex at https://github.com/teemtee/tmt/blob/5ba7562c4dd9acd61b0aefa99bfea93e6dce9d12/tmt/steps/discover/fmf.py#L682-L684

so the easiest solution would be, probably, to add the end-of-string anchor. But I have a sneaky feeling the anchor is not there on purpose, so that's going to break other things.

I tested this with tmt v1.67, but the code appears to be the same in the latest version as well.

LecrisUT:
The issue was that when the files in `/tests/foo` are changed, it would also match the `/tests/foo-bar` tests as being modified. On the other hand we do not just use `^$` because we want to allow for running the tests under `/tests/foo/var_1 ...`. Thus we explicitly check for either `$` or `/`

---

Fixes #4626

Pull Request Checklist

* [x] implement the feature
* [x] include a release note


tmt/steps/discover/fmf.py
```diff
@@ -793,7 +793,10 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
             if output.stdout:
                 directories = [Path(name).parent for name in output.stdout.split('\n')]
                 modified = {
-                    f"^/{re.escape(str(directory))}" for directory in directories if directory
+                    # ($|/): match end of plan name or `/` which would be any sub-plan
+                    f"^/{re.escape(str(directory))}($|/)"
+                    for directory in directories
+                    if directory
                 }
```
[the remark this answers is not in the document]
LecrisUT: No. Either not interested (`splitlines`), or plain wrong (`.` thingy)

tcornell-bus:
Is it feasible for a simple test to be added to show the fix?

LecrisUT:
> Is it feasible for a simple test to be added to show the fix?

Considered it initially. Not with modifying the current tests, we would need to make dedicated tests in https://github.com/teemtee/tests

tcornell-bus:
> 
>    >  Is it feasible for a simple test to be added to show the fix?
> 
> Considered it initially. Not with modifying the current tests, we would need to make dedicated tests in https://github.com/teemtee/tests

Please consider adding a test and opening a PR there. It seems this bug was caused by someone trying to fix the opposite problem, so some kind of test to verify the intended behavior would be preferable.

LecrisUT:
> Please consider adding a test and opening a PR there. It seems this bug was caused by someone trying to fix the opposite problem, so some kind of test to verify the intended behavior would be preferable.

Added and are pending on https://github.com/teemtee/tests/pull/24 -> https://github.com/teemtee/tests/pull/25 merges

tmt/steps/discover/fmf.py
```diff
@@ -793,7 +793,10 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
             if output.stdout:
                 directories = [Path(name).parent for name in output.stdout.split('\n')]
                 modified = {
-                    f"^/{re.escape(str(directory))}" for directory in directories if directory
+                    # ($|/): match end of plan name or `/` which would be any sub-plan
+                    f"^/{re.escape(str(directory))}($|/)"
+                    for directory in directories
+                    if directory
```
tcornell-bus, line 799: I followed the reproducer in the [issue](https://github.com/teemtee/tmt/issues/4626#issue-4010847415) to test this, and that works fine. Then I added 3 more tests, which I think are in the format described by this [comment](https://github.com/teemtee/tmt/issues/4626#issuecomment-4031438604).
```
(dev) tcornell /tmp/tmt-test (new) $ ls -R
.:
foo  foo-bar  food  plans

./foo:
main.fmf  test.sh

./foo-bar:
main.fmf  test.sh

./food:
main.fmf  py313.fmf  py314  test.sh

./food/py314:
main.fmf

./plans:
modified.fmf
```
`food/` is the parent test, then `food/py313` and `food/py314` are the children. I have them in two slightly different formats. Their `.fmf` files point to `food/test.sh` though.

This is the current output:
```
(dev) tcornell /tmp/tmt-test (new) $ tmt -v run discover --how fmf --modified-ref main
/var/tmp/tmt/run-135
Found 1 plan.

/plans/modified
summary: Only modified tests
    discover
        directory: /tmp/tmt-test
        commit-hash: 0478a48
        how: fmf
        modified-ref: main
        modified-ref hash: 9dddb9a
        summary: 6 tests selected
            /food/py313
            /food/py314
            /food/py314
            /foo
            /foo-bar
            /food/py313
```
Looks like they are getting listed twice and the parent `food/` test is missing.

LecrisUT: Do you have a link to the commit? I think I understand why the duplication of `/food/py314` occurs. If you modified both `food/test.sh` and `food/py314/main.fmf`, then the `directory` here is `food` and `food/py314` and when it gets to expanding the `tree.tests(names)` both match.

It should be a pre-existing issue though, and I can see some other failures like when the tree is not the git root. The way I think this could be fixed (finding the top-most level where the files were changed) involves fixing that part as well.

tcornell-bus: made a quick repo https://github.com/tcornell-bus/tmt-test, just checkout `new` and run the above discover command and that should repro

LecrisUT: Ok see https://github.com/teemtee/tmt/issues/4835 for the duplicate `/food/py314`. I am 99% sure it is a pre-existing issue, but would appreciate the double-check. Any other issue for this thread?

tcornell-bus:
https://github.com/teemtee/tests/pull/24 -> https://github.com/teemtee/tests/pull/25 
both approved

tmt/steps/discover/fmf.py
tcornell-bus: The issue description looks right to me, and it looks like it's due to the code path you state. Thanks for looking into it and filing the issue!

tests/discover/modified/plan.fmf
```diff
@@ -0,0 +1,23 @@
...
+execute:
+  how: tmt
+
+/remote:
+  discover+:
+    # TODO: Move to main repo after https://github.com/teemtee/tests/pull/25 is merged
+    url: https://github.com/LecrisUT/tmt-tests
+    ref: f429b61
+    modified-ref: f429b61^
+    filter: tag:remote
```
tcornell-bus, line 14: I don't see any `tag: remote` in the tests added in https://github.com/teemtee/tests/pull/24, just making a note for when you go to touch this after those two are merged.

tests/discover/modified/plan.fmf
```diff
@@ -0,0 +1,23 @@
+discover:
+  how: fmf
+  modified-only: true
+provision:
+    how: local
```
tcornell-bus:
```suggestion
  how: local
```
minor nitpick

tcornell-bus (commented):
LGTM :slightly_smiling_face:  just waiting for the two teemtee/tests PRs to be merged, and then this PR  can be updated and the tests run.

tests/discover/modified/plan.fmf
LecrisUT: https://github.com/teemtee/tests/blob/main/tests/main.fmf :wink:

happz (approved):

tests/discover/modified/plan.fmf
```diff
@@ -0,0 +1,23 @@
...
+execute:
+  how: tmt
+
+/remote:
+  discover+:
+    # TODO: Move to main repo after https://github.com/teemtee/tests/pull/25 is merged
+    url: https://github.com/LecrisUT/tmt-tests
+    ref: f429b61
```
thrix: Flagging here just to make sure this does not get merged until this is updated

thrix (approved):
LGTM, nothing besides existing comments found

tests/discover/modified/plan.fmf
LecrisUT: Done

tests/discover/modified/plan.fmf
```diff
@@ -0,0 +1,22 @@
...
+  how: fmf
+  modified-only: true
+provision:
+  how: local
+execute:
+  how: tmt
+
+/remote:
+  discover+:
+    url: https://github.com/teemtee/tmt-tests
```
tcornell-bus:
```suggestion
    url: https://github.com/teemtee/tests
```
Tried to run locally and I think this needs to be tests

LecrisUT: Oops, yeah, me blind

tests/discover/modified.sh
```diff
@@ -27,4 +38,8 @@ rlJournalStart
             plan -n fmf/empty-modified >/dev/null'
         rlAssertGrep 'summary: 0 tests selected' "$rlRun_LOG"
     rlPhaseEnd
+
+    rlPhaseStartCleanup
+        rlRun "rm -r $run" 0 "Remove the run directory"
```
tcornell-bus: I am getting prompted about this command:
```
rm: remove write-protected regular file '/tmp/tmp.tMA9lw3d0g/plan/remote/discover/default-0/tests/.git/objects/pack/pack-74354b83338170af8e1b3e0340f0b8c8c525df3e.pack'?
```

tcornell-bus (approved):

tests/discover/modified.sh
LecrisUT: Yep, we are not consistent with the `rm` commands, I hit it quite a lot as well. Should be good with be1b8cf

tcornell-bus:
Tests are failing due to the added commits to the teemtee/tests repo. Pinning the ref down fixes it.

[/tests/recipe/generate](https://artifacts.dev.testing-farm.io/6b24b344-9b41-49e3-a215-f9baf79e109f/#work-basicml5qjk7u_plans-features-basic_62_default-0)
```diff
diff --git a/tests/recipe/data/plans.fmf b/tests/recipe/data/plans.fmf
index 6cb8533b..40273ce9 100644
--- a/tests/recipe/data/plans.fmf
+++ b/tests/recipe/data/plans.fmf
@@ -51,6 +51,7 @@ environment:
         import:
             url: https://github.com/teemtee/tests
             name: /plans/remote
+            ref: 7ab3dab
 
 /minimal:
     summary: A minimal plan
```

[/tests/discover/archive](https://artifacts.dev.testing-farm.io/6b24b344-9b41-49e3-a215-f9baf79e109f/#work-discoverx784w1mu_plans-features-steps-discover_2_default-0)
```diff
diff --git a/tests/discover/data/plans.fmf b/tests/discover/data/plans.fmf
index ce03838e..53da10be 100644
--- a/tests/discover/data/plans.fmf
+++ b/tests/discover/data/plans.fmf
@@ -199,10 +199,10 @@ execute:
         summary: Test url-type archive
         discover:
             how: fmf
-            url: https://github.com/teemtee/tests/archive/refs/heads/main.tar.gz
+            url: https://github.com/teemtee/tests/archive/7ab3dabdfbe4f9e74421748d2950682f78bab6ae.tar.gz
             url-content-type: archive
             test: /tests
-            path: /tests-main
+            path: /tests-7ab3dabdfbe4f9e74421748d2950682f78bab6ae
 
     /dynamic-ref:
         /no-context:
```
and
```diff
diff --git a/tests/discover/archive.sh b/tests/discover/archive.sh
index 51033666..71424163 100755
--- a/tests/discover/archive.sh
+++ b/tests/discover/archive.sh
@@ -17,9 +17,9 @@ rlJournalStart
         rlAssertGrep "2 tests selected" $rlRun_LOG
         plan_path="$run$plan"
         step_workdir="$plan_path/discover/default-0"
-        rlAssertExists "$step_workdir/tests-main.tar.gz" 0 "Check that the archive is present"
-        rlAssertExists "$step_workdir/tests/tests-main" 0 "Check that the extracted archive is present"
-        rlAssertExists "$step_workdir/tests/tests-main/scripts/random_file.sh" 0 "Check that there is a script in the extracted archive"
+        rlAssertExists "$step_workdir/tests-7ab3dabdfbe4f9e74421748d2950682f78bab6ae.tar.gz" 0 "Check that the archive is present"
+        rlAssertExists "$step_workdir/tests/tests-7ab3dabdfbe4f9e74421748d2950682f78bab6ae" 0 "Check that the extracted archive is present"
+        rlAssertExists "$step_workdir/tests/tests-7ab3dabdfbe4f9e74421748d2950682f78bab6ae/scripts/random_file.sh" 0 "Check that there is a script in the extracted archive"
     rlPhaseEnd
 
     rlPhaseStartTest "Basic archive tests (shell)"
```

Seems like a better idea to pin these down then to keep having to change them when someone touches teemtee/tests :shrug:

LecrisUT:
Will try a combination of ref locking and adjusting the number of tests: https://github.com/teemtee/tmt/pull/4850. It seems like we should also do some reorganization of that repo

tmt/steps/discover/fmf.py
```diff
@@ -793,7 +793,10 @@ def do_the_discovery(self, path: Optional[Path] = None) -> list['tmt.base.core.T
             if output.stdout:
                 directories = [Path(name).parent for name in output.stdout.split('\n')]
                 modified = {
-                    f"^/{re.escape(str(directory))}" for directory in directories if directory
+                    # ($|/): match end of plan name or `/` which would be any sub-plan
```
```suggestion
                    # ($|/): match end of test name or `/` which would be any sub-test
```

We're checking for modified tests here in the `discover` step, not for plans, right?

LecrisUT: Hmm, right

psss (approved):
Except for the confusing comment looks good. Thanks!

psss:
Also pull request summary & description should be updated accordingly.

tmt/steps/discover/fmf.py
LecrisUT: Ok, adjusted along with the description

